PR #30, opened 2020-01-28
Merge from uber/cadence on 01/27/20

alexshtin:


internal/headers_test.go
```diff
@@ -82,7 +83,9 @@ func TestHeaderWriter(t *testing.T) {
 	}
 
 	for _, test := range tests {
+		test := test
```
alexshtin: Looks rudimentary, but actually required.

PR #37, opened 2020-02-20
Switch to serviceerror package

alexshtin:
1. Updated to latest `temporal-proto` module.
2. Created `workflowServiceErrorWrapper` which converts gRPC status error to service error.
3. Replaced `status.Status` with `serviceerror` everywhere in the code.

### DIFF
--- a/internal/common/metrics/service_wrapper.go
+++ b/internal/common/metrics/service_wrapper.go
@@ -25,12 +25,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/gogo/status"
 	"github.com/uber-go/tally"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-
+	"go.temporal.io/temporal-proto/serviceerror"
 	"go.temporal.io/temporal-proto/workflowservice"
+	"google.golang.org/grpc"
 )
 
 type (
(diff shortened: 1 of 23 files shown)


internal/common/metrics/service_wrapper.go
```diff
@@ -117,10 +119,14 @@ func (w *workflowServiceMetricsWrapper) getOperationScope(scopeName string) *ope
 func (s *operationScope) handleError(err error) {
 	s.scope.Timer(CadenceLatency).Record(time.Since(s.startTime))
 	if err != nil {
-		st := status.Convert(err)
-		if st.Code() == codes.NotFound || st.Code() == codes.InvalidArgument || st.Code() == codes.AlreadyExists {
+		switch err.(type) {
```
alexshtin: This is how it looked like in original version.

mfateev: Should we use errors.As(...)? 
https://blog.golang.org/go1.13-errors

alexshtin: I did consider it. I even have this link still opened. Honestly I am in doubts. `errors.As()` requires one more variable declaration and in most cases (like this one) we don't actually need error itself. Also I don't want to mix patterns. If we decide to switch to `errors.As()` we need to do it everywhere, I believe.

PR #41, opened 2020-02-26
Added WorkflowInterceptor and removed global registrations.

mfateev:
Unfortunately this PR includes a few changes that potentially could be broken into separate changes.

The changes:

1. Default workflow and activity type names are changed to the function name without path. The name that was for example `go.temporal.io/temporal/evictiontest.testReplayWorkflow` now is just `testReplayWorkflow`. This is done to allow moving activity and workflow functions between packages without breaking backwards compatibility.
2. Global registration functions workflow.Register and activity.Register are removed. All registration should be done with a Worker object. This was frequently requested feature that simplified unit and other types of testing.
3. Introduced `WorkflowInterceptor` and correspondent `WorkflowInterceptorFactory`. An interceptor chain is configured through the` worker.Options`. The interceptor executes in the context of a workflow. So it must be deterministic or filter side effects though IsReplaying(). The first intended usage for the interceptors is ability to specify activity options like timeouts and retry policies through configuration.
4. As registration is not global anymore introduced WorkflowReplayer and correspondent NewReplayer method.


internal/interceptors.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/uber-go/tally"
+	"time"
+)
+import "go.uber.org/zap"
+
+// WorkflowInterceptorFactory is used to create a single link in the interceptor chain
+type WorkflowInterceptorFactory interface {
+	// NewInterceptor creates an interceptor instance. The created instance must delegate every call to
+	// the next parameter for workflow code function correctly.
+	NewInterceptor(info *WorkflowInfo, next WorkflowInterceptor) WorkflowInterceptor
```
samarabbas: Should we pass workflow context as parameter?

internal/internal_workflow.go
```diff
@@ -94,6 +94,7 @@ type (
 	// Note that workflow.Context is used instead of context.Context to avoid use of raw channels.
 	workflow interface {
 		Execute(ctx Context, input []byte) (result []byte, err error)
+		WorkflowType() string
```
samarabbas: We represent WorkflowType and ActivityType as a struct on server.  Do you think we should expose similar struct here?

internal/interceptors.go
mfateev: This code doesn't run in the context of the workflow thread. So I don't want to give the context here.

internal/internal_workflow.go
mfateev: I just removed it here. It was a leftover from a previous design attempt.

internal/common/rpc/error_wrapper.go
```diff
@@ -40,7 +40,7 @@ var (
 )
 
 // NewWorkflowServiceErrorWrapper creates a new wrapper to WorkflowService that will convert gRPC status errors to service errors.
-func NewWorkflowServiceErrorWrapper(service workflowservice.WorkflowServiceClient) *workflowServiceErrorWrapper {
+func NewWorkflowServiceErrorWrapper(service workflowservice.WorkflowServiceClient) workflowservice.WorkflowServiceClient {
```
alexshtin: [This](https://github.com/golang/go/wiki/CodeReviewComments) says:

> The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring.

I found this rule useful and try to follow it whenever it is possible.

PR #42, opened 2020-02-26
enable buildkite integration to run all tests

samarabbas:


docker/buildkite/docker-compose.yml
```diff
@@ -49,7 +49,7 @@ services:
       - "SERVICE_ADDR=cadence:7933"
```
alexshtin: gRPC port is `7233`.

docker/buildkite/docker-compose.yml
```diff
@@ -22,8 +22,8 @@ services:
         aliases:
           - statsd
 
-  cadence:
-    image: ubercadence/server:master-auto-setup
+  temporal:
+    image: temporalio/temporal:master-auto-setup
     ports:
       - "7933:7933"
       - "7934:7934"
```
alexshtin: I don't think we need to expose history service to integration tests. If yes, this port need to be changed to `7234`.

PR #48, opened 2020-03-01
Fixed flaky tests and local activity cancellation testsuite handling

mfateev:
Fixed flaky tests in child workflow handling of parent completion.
Fixed testsuite handling of local activity cancellation.

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -1111,16 +1111,8 @@ func (env *testWorkflowEnvironmentImpl) RequestCancelLocalActivity(activityID st
 		env.logger.Debug("RequestCancelLocalActivity failed, LocalActivity not exists or already completed.", zap.String(tagActivityID, activityID))
 		return
 	}
-	activityInfo := env.getActivityInfo(activityID, getActivityFunctionName(env.registry, task.params.ActivityFn))
 	env.logger.Debug("RequestCancelLocalActivity", zap.String(tagActivityID, activityID))
-	delete(env.localActivities, activityID)
-	env.postCallback(func() {
-		lar := &localActivityResultWrapper{err: ErrCanceled, backoff: noRetryBackoff}
-		task.callback(lar)
-		if env.onLocalActivityCanceledListener != nil {
-			env.onLocalActivityCanceledListener(activityInfo)
-		}
-	}, true)
+	task.cancel()
 }
 
 func (env *testWorkflowEnvironmentImpl) handleActivityResult(activityID string, result interface{}, activityType string,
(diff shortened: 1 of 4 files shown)


internal/internal_workflow_testsuite.go
```diff
@@ -1111,16 +1111,8 @@ func (env *testWorkflowEnvironmentImpl) RequestCancelLocalActivity(activityID st
 		env.logger.Debug("RequestCancelLocalActivity failed, LocalActivity not exists or already completed.", zap.String(tagActivityID, activityID))
 		return
 	}
-	activityInfo := env.getActivityInfo(activityID, getActivityFunctionName(env.registry, task.params.ActivityFn))
 	env.logger.Debug("RequestCancelLocalActivity", zap.String(tagActivityID, activityID))
-	delete(env.localActivities, activityID)
-	env.postCallback(func() {
-		lar := &localActivityResultWrapper{err: ErrCanceled, backoff: noRetryBackoff}
-		task.callback(lar)
-		if env.onLocalActivityCanceledListener != nil {
```
alexshtin: What was this?

internal/internal_workflow_testsuite_test.go
```diff
@@ -1694,13 +1694,14 @@ func (s *WorkflowTestSuiteUnitTest) Test_LocalActivity() {
...
 func (s *WorkflowTestSuiteUnitTest) Test_WorkflowLocalActivityWithMockAndListeners() {
+	var localActivityFnCancelled bool
 	localActivityFn := func(ctx context.Context, name string) (string, error) {
 		return "hello " + name, nil
 	}
 
-	cancelledLocalActivityFn := func() error {
-		// It will be cancelled anyway.
-		time.Sleep(time.Hour)
+	cancelledLocalActivityFn := func(ctx context.Context) error {
```
alexshtin: This is the difference between "just fix" and "right fix. :-)

internal/internal_workflow_testsuite_test.go
```diff
@@ -1694,13 +1694,14 @@ func (s *WorkflowTestSuiteUnitTest) Test_LocalActivity() {
 
 // Flaky test. Rerun if failed.
```
alexshtin: You can remove this comment too.

test/integration_test.go
```diff
@@ -368,24 +368,37 @@ func (ts *IntegrationTestSuite) TestChildWFWithMemoAndSearchAttributes() {
...
 
-// Flaky test. Rerun if failed.
 func (ts *IntegrationTestSuite) TestChildWFWithParentClosePolicyTerminate() {
 	var childWorkflowID string
 	err := ts.executeWorkflow("test-childwf-parent-close-policy", ts.workflows.ChildWorkflowSuccessWithParentClosePolicyTerminate, &childWorkflowID)
 	ts.NoError(err)
-	resp, err := ts.libClient.DescribeWorkflowExecution(context.Background(), childWorkflowID, "")
-	ts.NoError(err)
-	ts.True(resp.WorkflowExecutionInfo.GetCloseTime() > 0, "CloseTime is not greater than zero but %d. Flaky test. Rerun if failed.", resp.WorkflowExecutionInfo.GetCloseTime())
+	for {
```
alexshtin: So, just retry. Good to know.

test/workflow_test.go
```diff
@@ -486,7 +478,11 @@ func (w *Workflows) childForMemoAndSearchAttr(ctx workflow.Context) (result stri
 }
 
 func (w *Workflows) sleep(ctx workflow.Context, d time.Duration) error {
-	ctx = workflow.WithActivityOptions(ctx, w.defaultActivityOptions())
+	ctx = workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		ScheduleToStartTimeout: 5 * time.Second,
+		ScheduleToCloseTimeout: 5*time.Second + d,
```
alexshtin: nit: fix spacing around `*`.

internal/internal_workflow_testsuite.go
mfateev: For some reason the cancellation path in the testEnv was different for local activity and the real one. I unified them. I think it is because the local activity was added later.

test/integration_test.go
mfateev: It relies on the eventual workflow timeout.

test/workflow_test.go
mfateev: This is how go fmt thinks about it.

alexshtin: You're right. This is weird though.

PR #53, opened 2020-03-03
Change WorkflowIdReusePolicy default value

alexshtin:
Proto enum are changed separately but client has its own corresponding constants.


internal/internal_workflow_testsuite_test.go
```diff
@@ -1693,14 +1694,14 @@ func (s *WorkflowTestSuiteUnitTest) Test_LocalActivity() {
 }
 
 func (s *WorkflowTestSuiteUnitTest) Test_WorkflowLocalActivityWithMockAndListeners() {
-	var localActivityFnCancelled bool
+	var localActivityFnCancelled atomic.Bool
```
alexshtin: I run test wit race detector and it found race here.

PR #58, opened 2020-03-12
Create service client internally based on passed hostPort option

alexshtin:
1. `client.NewClient`, `client.NewDomainClient`, and `worker.New` don't accept service wrapper (`workflowservice.WorkflowServiceClient`) anymore.
2. `client.NewClient`, `client.NewDomainClient` accept `ClientOption` (not `*ClientOption`) to be consistent with `worker.New` and simplify code.
3.  `client.NewClient`, `client.NewDomainClient`, and `worker.New` returns 2 values: object and error which can be not nil in case of gRPC error.
4. `HostPort` is added to both `ClientOptions` and `WorkerOptions`. It is used to connect to Temporal cluster (`localhost:7322`) by default.
5. `GRPCDialer` is added to both `ClientOptions` and `WorkerOptions`. It allows to override connection mehanizm. Use TLS, custom interceptors or share single connection between worker and client (connection needs to be stored somewhere in customer code).
6. `CloseConnection` is added to `Client`, `DomainClient`, and `Worker` and needs to be called to close underlying connection. `goleaks` in integration tests catches leaked connections also.


internal/internal_worker_test.go
```diff
@@ -484,9 +483,9 @@ func (s *internalWorkerTestSuite) TestWorkerStartFailsWithInvalidDomain() {
 		domainErr  error
 		isErrFatal bool
 	}{
-		{status.Error(codes.NotFound, ""), true},
```
alexshtin: No `status` error in the code anymore.

mocks/Client.go
```diff
@@ -513,3 +513,17 @@ func (_m *Client) TerminateWorkflow(ctx context.Context, workflowID string, runI
 
 	return r0
 }
+
```
alexshtin: Fixed mocks manually but I think they can be safely removed. These are generated with some 'mockery` framework and we use `gomock` everywhere else. I tried to use `mockgen` to generate mocks and it generate correct mocks. So I would allow customers to generate mocks themselves.

internal/client.go
```diff
@@ -470,64 +484,97 @@ const (
...
+		return nil, err
 	}
-	var contextPropagators []ContextPropagator
-	if options != nil {
-		contextPropagators = options.ContextPropagators
+
+	return newServiceClient(workflowservice.NewWorkflowServiceClient(connection), connection, domain, options), nil
+}
+
+func newServiceClient(workflowServiceClient workflowservice.WorkflowServiceClient, connectionCloser io.Closer, domain string, options ClientOptions) Client {
```
alexshtin: This signature is still required because it is used in all unit tests which pass service mock here.

internal/client.go
```diff
@@ -470,64 +484,97 @@ const (
...
-		dataConverter = getDefaultDataConverter()
+
+	connection, err := options.GRPCDialer(GRPCDialerParams{
+		HostPort:             options.HostPort,
+		RequiredInterceptors: requiredInterceptors(metricsScope),
+		DefaultServiceConfig: DefaultServiceConfig,
+	})
+
+	if err != nil {
+		return nil, err
```
alexshtin: I might need to wrap error here.

internal/grpc_dialer.go
```diff
@@ -0,0 +1,62 @@
...
+type (
+	// GRPCDialerParams are passed to GRPCDialer and must be used to create gRPC connection.
+	GRPCDialerParams struct {
+		HostPort             string
+		RequiredInterceptors []grpc.UnaryClientInterceptor
+		DefaultServiceConfig string
+	}
+
+	// GRPCDialer can be used to set custom gRPC connection creation logic.
+	GRPCDialer func(params GRPCDialerParams) (*grpc.ClientConn, error)
```
mfateev: Add comment with the sample.

internal/client.go
```diff
@@ -302,15 +308,20 @@ type (
...
 		//  - EntityNotExistError
 		DescribeTaskList(ctx context.Context, tasklist string, tasklistType enums.TaskListType) (*workflowservice.DescribeTaskListResponse, error)
+
+		// CloseConnection closes underlying gRPC connection.
+		CloseConnection() error
 	}
 
 	// ClientOptions are optional parameters for Client creation.
 	ClientOptions struct {
+		HostPort           string
```
mfateev: Document the format and that it is optional. And what happens when it is not specified.

mfateev:
Do you need to do something similar for the Worker?

alexshtin:
> Do you need to do something similar for the Worker?

I did.

internal/common/metrics/service_wrapper.go
```diff
@@ -26,7 +26,6 @@ import (
 	"time"
```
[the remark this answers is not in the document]
alexshtin: #60

mocks/Client.go
alexshtin: #61

internal/grpc_dialer.go
alexshtin: I added comment to options, not here.

LICENSE
```diff
@@ -1,20 +1,21 @@
...
-// all copies or substantial portions of the Software.
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
+The MIT License (MIT)
```
samarabbas: Why change license headers on this change?

internal/client.go
```diff
@@ -470,64 +532,97 @@ const (
...
 
 // NewDomainClient creates an instance of a domain client, to manager lifecycle of domains.
-func NewDomainClient(service workflowservice.WorkflowServiceClient, options *ClientOptions) DomainClient {
-	var identity string
-	if options == nil || options.Identity == "" {
-		identity = getWorkerIdentity("")
-	} else {
-		identity = options.Identity
+func NewDomainClient(options ClientOptions) (DomainClient, error) {
+	metricsScope := tagScope(options.MetricsScope, tagDomain, "domain-client", clientImplHeaderName, clientImplHeaderValue)
```
samarabbas: I don't think using 'tagDomain' makes sense here.  You should remove this tag.
Or use some constant value like 'EmptyDomain'

LICENSE
alexshtin: I played with it. Will revert, it is not part of this PR anyway.

internal/client.go
alexshtin: Remove.

PR #65, opened 2020-03-14
Reset SDK versions

alexshtin:


alexshtin:
@mfateev, can you update java-sdk also?

PR #68, opened 2020-03-17
Move domain to options

alexshtin:
For both `ClientOptions` and `WorkerOptions` domain is now optional with default value `default`.


go.mod
```diff
@@ -5,24 +5,38 @@ go 1.12
 require (
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/facebookgo/clock v0.0.0-20150410010913-600d898af40a
+	github.com/gogo/googleapis v1.3.2 // indirect
```
alexshtin: Updated all dependencies.

internal/internal_worker.go
```diff
@@ -122,6 +120,9 @@ type (
 
 	// workerExecutionParameters defines worker configure/execution options.
 	workerExecutionParameters struct {
+		// Domain name.
```
mfateev: Add info about the default value

test/test_utils.go
```diff
@@ -36,7 +34,7 @@ type Config struct {
 
 func newConfig() Config {
 	cfg := Config{
-		ServiceAddr: internal.LocalHostPort,
+		ServiceAddr: "localhost:7233",
```
mfateev: I created a public constant for this.

internal/internal_worker.go
alexshtin: This is internal struct and it doesn't have default value. Public structs have default value in comment.

PR #69, opened 2020-03-18
Worker uses client to make a connection

alexshtin:
`worker.NewWorker` now depends on `client.Client` interface. It doesn't create/close own gRPC connection but uses client connection instead.

I had to make few short term shortcuts which I described below.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -568,7 +568,7 @@ func NewClient(options ClientOptions) (Client, error) {
 }
 
 // NewServiceClient creates workflow client from workflowservice.WorkflowServiceClient. Must be used internally in unit tests only.
-func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClient, connectionCloser io.Closer, options ClientOptions) *workflowClient {
+func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClient, connectionCloser io.Closer, options ClientOptions) *WorkflowClient {
 	// DomainName can be empty in unit tests.
 	if len(options.DomainName) == 0 {
 		options.DomainName = DefaultDomainName
@@ -588,7 +588,7 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
 		options.Tracer = opentracing.NoopTracer{}
 	}
 
-	return &workflowClient{
+	return &WorkflowClient{
 		workflowService:    workflowServiceClient,
 		connectionCloser:   connectionCloser,
 		domain:             options.DomainName,
--- a/internal/client.go
+++ b/internal/client.go
@@ -564,10 +564,11 @@ func NewClient(options ClientOptions) (Client, error) {
 		return nil, err
 	}
 
-	return newServiceClient(workflowservice.NewWorkflowServiceClient(connection), connection, options), nil
+	return NewServiceClient(workflowservice.NewWorkflowServiceClient(connection), connection, options), nil
 }
 
-func newServiceClient(workflowServiceClient workflowservice.WorkflowServiceClient, connectionCloser io.Closer, options ClientOptions) Client {
+// NewServiceClient creates workflow client from workflowservice.WorkflowServiceClient. Must be used internally in unit tests only.
+func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClient, connectionCloser io.Closer, options ClientOptions) *workflowClient {
 	// DomainName can be empty in unit tests.
 	if len(options.DomainName) == 0 {
 		options.DomainName = DefaultDomainName
--- a/internal/client.go
+++ b/internal/client.go
@@ -544,7 +544,7 @@ func NewClient(options ClientOptions) (Client, error) {
 		options.DomainName = DefaultDomainName
 	}
 
-	metricsScope := tagScope(options.MetricsScope, tagDomain, options.DomainName, clientImplHeaderName, clientImplHeaderValue)
+	options.MetricsScope = tagScope(options.MetricsScope, tagDomain, options.DomainName, clientImplHeaderName, clientImplHeaderValue)
 
 	if len(options.HostPort) == 0 {
 		options.HostPort = LocalHostPort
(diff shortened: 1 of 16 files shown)


internal/internal_workflow_client.go
```diff
@@ -56,8 +56,8 @@ var (
 )
 
 type (
-	// workflowClient is the client for starting a workflow execution.
-	workflowClient struct {
+	// WorkflowClient is the client for starting a workflow execution.
+	WorkflowClient struct {
```
samarabbas: Why do we have to make concrete implementation public?  Can we just another API on client to retrieve Options?

internal/client.go
```diff
@@ -556,24 +556,24 @@ func NewClient(options ClientOptions) (Client, error) {
...
 		return nil, err
 	}
 
-	return newServiceClient(workflowservice.NewWorkflowServiceClient(connection), connection, options), nil
+	return NewServiceClient(workflowservice.NewWorkflowServiceClient(connection), connection, options), nil
 }
 
-func newServiceClient(workflowServiceClient workflowservice.WorkflowServiceClient, connectionCloser io.Closer, options ClientOptions) Client {
+// NewServiceClient creates workflow client from workflowservice.WorkflowServiceClient. Must be used internally in unit tests only.
+func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClient, connectionCloser io.Closer, options ClientOptions) *WorkflowClient {
```
alexshtin: Had to make it public because of eviction test.

internal/internal_workflow_client.go
alexshtin: Had to make this public because it is returned by public func `NewServiceClient` and lint complains, if it is private.

internal/internal_workflow_testsuite.go
```diff
@@ -120,13 +120,16 @@ type (
...
-		ctxProps     []ContextPropagator
-		mockClock    *clock.Mock
-		wallClock    clock.Clock
+		mock               *mock.Mock
+		service            workflowservice.WorkflowServiceClient
+		logger             *zap.Logger
+		metricsScope       *metrics.TaggedScope
+		contextPropagators []ContextPropagator
+		identity           string
+		tracer             opentracing.Tracer
```
alexshtin: `identity` and `tracer` moved out from `workerOptions`.

internal/internal_workflow_testsuite.go
```diff
@@ -178,6 +181,7 @@ type (
 		testError        error
 		doneChannel      chan struct{}
 		workerOptions    WorkerOptions
+		dataConverter    DataConverter
```
alexshtin: This also comes from `workerOptions`.
I initially added this to the shared struct above and it was the bug I was chasing this morning.

internal/worker.go
```diff
@@ -35,16 +32,6 @@ type (
 	// The current timeout resolution implementation is in seconds and uses math.Ceil(d.Seconds()) as the duration. But is
 	// subjected to change in the future.
 	WorkerOptions struct {
-
```
alexshtin: All these options exist in `ClientOptions`.

internal/worker.go
```diff
@@ -249,43 +179,18 @@ func IsReplayDomain(dn string) bool {
...
+	client Client,
 	taskList string,
 	options WorkerOptions,
-) (*AggregatedWorker, error) {
-
-	if options.MetricsScope == nil {
-		options.MetricsScope = tally.NoopScope
-		options.Logger.Info("No metrics scope configured for temporal worker. Use NoopScope as default.")
+) *AggregatedWorker {
+	workflowClient, ok := client.(*WorkflowClient)
```
alexshtin: This is another shortcut I plan to refactor later. I didn't touch `Client` interface.

Actually probably it is better for `NewClient` to return `WorkflowClient` struct not `Client` interface. And `NewWorker` should also depend on `WorkflowClient` not `Client`.

internal/internal_workflow_client.go
mfateev: Yes, I hate this linter check

internal/worker.go
```diff
@@ -95,39 +82,6 @@ type (
...
-		// 	Reporter:        reporter,
-		// 	SanitizeOptions: &_sanitizeOptions,
-		// }
-		// scope, _ := tally.NewRootScope(opts, time.Second)
-		// default: no metrics.
-		MetricsScope tally.Scope
-
 		// Optional: Logger framework can use to log.
 		// default: default logger provided.
 		Logger *zap.Logger
```
mfateev: Why logger is not moving?

internal/workflow_testsuite.go
```diff
@@ -176,13 +177,38 @@ func (t *TestActivityEnvironment) ExecuteLocalActivity(activityFn interface{}, a
...
-// use options of Identity, MetricsScope and BackgroundActivityContext on the WorkerOptions. Other options are ignored.
+// use options of BackgroundActivityContext, MaxConcurrentSessionExecutionSize, and WorkflowInterceptorChainFactories on the WorkerOptions.
+// Other options are ignored.
 // Note: WorkerOptions is defined in internal package, use public type worker.Options instead.
 func (t *TestActivityEnvironment) SetWorkerOptions(options WorkerOptions) *TestActivityEnvironment {
 	t.impl.setWorkerOptions(options)
 	return t
 }
 
+// SetDataConverter sets data converter.
```
mfateev: What about adding setClientOptions instead of all of these?

internal/worker.go
samarabbas: This is not ideal.  Let's open an issue to clean this up later.

internal/worker.go
alexshtin: I removed only those which exist in `ClientOptions` already. `Logger` is not there. Apparently, client doesn't need it.

internal/workflow_testsuite.go
alexshtin: Yes, it is possible. Not sure if it is better though.

alexshtin: I am actually thinking about replacing `SetWorkerOptions` with 3 setters. I believe `ClientOptions` and `WorkerOptions` should be used to create client and worker only.

PR #74, opened 2020-03-23
Fix README.md and lint errors

alexshtin:
Fixed MD files and `errcheck` and `staticcheck` errors.


test/workflow_test.go
```diff
@@ -422,12 +422,7 @@ func (w *Workflows) ConsistentQueryWorkflow(ctx workflow.Context, delay time.Dur
 		ScheduleToCloseTimeout: 5 * time.Second,
 	})
 
-	workflowInfo := internal.GetWorkflowInfo(laCtx)
-	if &workflowInfo.WorkflowType == nil {
```
alexshtin: It can't be `nil`.

PR #83, opened 2020-03-31
Remove service_wrapper and add interceptor tests

alexshtin:


internal/common/metrics/service_wrapper.go
```diff
@@ -1,392 +0,0 @@
...
-	scopeNameDescribeWorkflowExecution          = TemporalMetricsPrefix + "DescribeWorkflowExecution"
-	scopeNameResetStickyTaskList                = TemporalMetricsPrefix + "ResetStickyTaskList"
-	scopeNameGetSearchAttributes                = TemporalMetricsPrefix + "GetSearchAttributes"
-	scopeNameListTaskListPartitions             = TemporalMetricsPrefix + "ListTaskListPartitions"
-	scopeNameGetClusterInfo                     = TemporalMetricsPrefix + "GetClusterInfo"
-	scopeNameGetWorkflowExecutionRawHistory     = TemporalMetricsPrefix + "GetWorkflowExecutionRawHistory"
-)
-
-var (
-	_ workflowservice.WorkflowServiceClient = (*workflowServiceMetricsWrapper)(nil)
```
alexshtin: This line was a pain. It is the only place where SDK had a direct dependency on `workflowservice.WorkflowServiceClient` interface.

PR #85, opened 2020-04-03
Rearrange proto files

alexshtin:
Only package renames. Need to update SDK first.


samarabbas:
Can you describe the high level of refactor from proto perspective.  It makes is really easy to review as rest is purely mechanical.

I don't like moving WorkflowExecution to workflow.WorkflowExecution.

Let's put the entire proposal in one place first.

alexshtin:
Here is the design proposal: https://github.com/temporalio/temporal-proto/issues/18. I agree some of them might be misplaced. I believe you, @samarabbas, and @mfateev should review new layout and I will make corrections. Also I would love to remove `common` package completely.

PR #100, opened 2020-04-13
Broke channel into WriteChannel and ReadChannel interfaces

mfateev:
Native Go channels support read only and write only interface.
To match that break `Channel` into `WriteChannel` and `ReadChannel` (naming is up to discussion).
Change `getSignalChannel` to return `ReadChannel` as ability to write and close the signal channel is a common topic for confusion.

### DIFF
--- a/internal/context.go
+++ b/internal/context.go
@@ -238,10 +238,10 @@ func propagateCancel(parent Context, child canceler) {
 	} else {
 		go func() {
 			s := NewSelector(parent)
-			s.AddReceive(parent.Done(), func(c ReceiveChannel, more bool) {
+			s.AddReceive(parent.Done(), func(c ReadChannel, more bool) {
 				child.cancel(false, parent.Err())
 			})
-			s.AddReceive(child.Done(), func(c ReceiveChannel, more bool) {})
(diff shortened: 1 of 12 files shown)


internal/context.go
```diff
@@ -238,10 +238,10 @@ func propagateCancel(parent Context, child canceler) {
 	} else {
 		go func() {
 			s := NewSelector(parent)
-			s.AddReceive(parent.Done(), func(c Channel, more bool) {
+			s.AddReceive(parent.Done(), func(c ReadChannel, more bool) {
```
alexshtin, line 241: It seems `ReceiveChannel` and `SendChannel` are better names and they match go terminology.

internal/workflow.go
```diff
@@ -49,9 +49,20 @@ var (
 )
 
 type (
-	// Channel must be used instead of native go channel by workflow code.
-	// Use workflow.NewChannel(ctx) method to create Channel instance.
-	Channel interface {
+	// WriteChannel is a write only view of the Channel
+	WriteChannel interface {
+		// Send blocks until the data is sent.
+		Send(ctx Context, v interface{})
```
alexshtin, line 55: Internal methods are alread `Send*` and `Receive*`.

PR #102, opened 2020-04-14
Options validation removal

mfateev:
Service already validates all the passed options. Validating them on the client leads to situations when workflows fail due to a bad option when they could be recovered. And even worse it leads to situations when validation logic on the client gets out of sync with the service.

Also fixed TestWorkflowEnvironment to support mock testing of activities by their string name without any registration. Also removed initialization of a session worker in TestWorkflowEnvironment unless requested.


internal/internal_task_handlers.go
```diff
@@ -598,8 +598,8 @@ func (wth *workflowTaskHandlerImpl) createWorkflowContext(task *workflowservice.
 		return nil, errors.New("first history event is not WorkflowExecutionStarted")
 	}
 	taskList := attributes.TaskList
-	if taskList == nil {
-		return nil, errors.New("nil TaskList in WorkflowExecutionStarted event")
+	if taskList == nil || len(taskList.Name) == 0 {
```
alexshtin, line 601: I personally prefer `taslkList.Name == ""`.

internal/internal_workflow_testsuite.go
```diff
@@ -963,6 +1002,169 @@ func (env *testWorkflowEnvironmentImpl) ExecuteActivity(parameters executeActivi
 	return activityInfo
 }
 
+// Copy of the server function func (v *decisionAttrValidator) validateActivityScheduleAttributes
```
alexshtin, line 1005: We are getting more and more cases when server and SDK shares some low level code (i.e. `serviceerror`). Should't we create some shared library?

mfateev: This is for testing. And I believe we want to share practically the whole service in embedded mode here.

internal/internal_task_handlers.go
mfateev: Updated everywhere regexp could find: https://github.com/temporalio/temporal-go-sdk/pull/102/commits/1e9170d1f19b92a1627398174dc08bb7475c959c

PR #103, opened 2020-04-17
Migrate []byte to proto Payload

alexshtin:
All user payload in proto files (mainly `details`, `input`, and `result` fields) was replaced from using `[]byte` to `commonpb.Payload`. This type has extra metadata attached to actual data.
See details here: https://github.com/temporalio/temporal/issues/107.


internal/internal_task_handlers_test.go
```diff
@@ -996,10 +999,11 @@ func (t *TaskHandlersTestSuite) TestConsistentQuery_Success() {
 	t.NoError(err)
 	t.NotNil(response)
 	t.Len(response.Decisions, 0)
+	answer, _ := DefaultDataConverter.ToData(startingQueryValue)
 	expectedQueryResults := map[string]*querypb.WorkflowQueryResult{
 		"id1": {
 			ResultType: querypb.QueryResultType_Answered,
-			Answer:     []byte(fmt.Sprintf("\"%v\"\n", startingQueryValue)),
```
alexshtin, line 1002: I hope this quotes and `\n` are not the part of contract.

internal/internal_worker.go
```diff
@@ -829,22 +824,19 @@ type workflowExecutor struct {
 	interceptors []WorkflowInterceptorFactory
 }
 
-func (we *workflowExecutor) Execute(ctx Context, input []byte) ([]byte, error) {
+func (we *workflowExecutor) Execute(ctx Context, input *commonpb.Payload) (*commonpb.Payload, error) {
 	var args []interface{}
 	dataConverter := getWorkflowEnvOptions(ctx).dataConverter
 	fnType := reflect.TypeOf(we.fn)
-	if fnType.NumIn() == 2 && isTypeByteSlice(fnType.In(1)) {
-		// Do not deserialize input if workflow has a single byte slice argument (besides ctx)
```
alexshtin, line 837: Now always deserialize because raw bytes is special encoding type.

internal/internal_workflow_testsuite_test.go
```diff
@@ -1362,7 +1361,7 @@ func (s *WorkflowTestSuiteUnitTest) Test_MockUpsertSearchAttributes() {
 		s.NotNil(wfInfo.SearchAttributes)
 		valBytes := wfInfo.SearchAttributes.IndexedFields["CustomIntField"]
 		var result int
-		_ = NewValue(valBytes).Get(&result)
+		_ = json.Unmarshal(valBytes, &result)
```
alexshtin, line 1364: For 'SearchAttributes' bytes are still bytes.

internal/internal_workflow_testsuite_test.go
```diff
@@ -1741,10 +1767,16 @@ func (s *WorkflowTestSuiteUnitTest) Test_WorkflowLocalActivityWithMockAndListene
 
 	env.SetOnLocalActivityCompletedListener(func(activityInfo *ActivityInfo, result Value, err error) {
 		s.NoError(err)
-		var resultValue string
+		var resultValue *commonpb.Payload
```
alexshtin, line 1770: I am not sure if it is ok that `result` has `*commonpb.Payload` now. I don't actually understand how `string` worked here before. Was it implicitly converted from `[]byte`?

internal/internal_workflow_client_test.go
```diff
@@ -979,7 +979,7 @@ func (s *workflowClientTestSuite) SignalWithStartWorkflowWithMemoAndSearchAttr()
 		gomock.Any(), gomock.Any(), gomock.Any()).Return(startResp, nil).
 		Do(func(_ interface{}, req *workflowservice.SignalWithStartWorkflowExecutionRequest, _ ...interface{}) {
 			var resultMemo, resultAttr string
-			err := json.Unmarshal(req.Memo.Fields["testMemo"], &resultMemo)
```
alexshtin, line 982: Such tricks are not possible anymore.

internal/internal_workflow_testsuite_test.go
mfateev: It was using DataConverter in result.Get
This test should keep workfing without any changes.

internal/session.go
```diff
@@ -195,7 +196,7 @@ func CreateSession(ctx Context, sessionOptions *SessionOptions) (Context, error)
 // The main usage of RecreateSession is for long sessions that are splited into multiple runs. At the end of
 // one run, complete the current session, get recreateToken from sessionInfo by calling SessionInfo.GetRecreateToken()
 // and pass the token to the next run. In the new run, session can be recreated using that token.
-func RecreateSession(ctx Context, recreateToken []byte, sessionOptions *SessionOptions) (Context, error) {
+func RecreateSession(ctx Context, recreateToken *commonpb.Payload, sessionOptions *SessionOptions) (Context, error) {
```
mfateev, line 199: Token should stay as a binary payload. It is an opaque blob from the client point of view.

internal/encoded_test.go
```diff
@@ -93,29 +100,48 @@ func newTestDataConverter() DataConverter {
 	return &testDataConverter{}
```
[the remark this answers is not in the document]
alexshtin: I don't know why we have it at all.

internal/encoded.go
```diff
@@ -47,59 +62,165 @@ type (
...
+		if !ok {
+			nvp.Name = fmt.Sprintf("values[%d]", i)
+			nvp.Value = value
+		}
 
-	data, err := encoder.Marshal(r)
-	if err != nil {
-		return nil, err
+		var payloadItem *commonpb.PayloadItem
+		if bytes, isByteSlice := nvp.Value.([]byte); isByteSlice {
```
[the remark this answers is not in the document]
alexshtin: I believe converters should be for one single argument (`PayloadItem`) and that "composite" dat converter should be pluggable with some rules like "use raw for []byte and json for all the rest". This is one of the topic I want to discuss.

internal/session.go
alexshtin: `deserializeRecreateToken` at the line bellow calls `dataConverter`. It must be `Payload` or data converter should not be used (just old plain JSON instead).

internal/internal_workflow_testsuite_test.go
alexshtin: It seems like it is double encoded. Need to investigate.

internal/encoded.go
```diff
@@ -47,59 +62,165 @@ type (
...
+		ToData(value ...interface{}) (*commonpb.Payload, error)
 		// FromData implements conversion of an array of values of different types.
 		// Useful for deserializing arguments of function invocations.
-		FromData(input []byte, valuePtr ...interface{}) error
+		FromData(input *commonpb.Payload, valuePtr ...interface{}) error
 	}
 
-	// defaultDataConverter uses thrift encoder/decoder when possible, for everything else use json.
+	// defaultDataConverter uses JSON.
 	defaultDataConverter struct{}
```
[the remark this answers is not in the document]
alexshtin: It is not only `JSON`. I am leaving this as it is and will refactor this later.

internal/encoded.go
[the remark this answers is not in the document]
alexshtin: After discussion we decided to leave it for the future.

internal/encoded.go
```diff
@@ -47,59 +62,165 @@ type (
...
+		var payloadItem *commonpb.PayloadItem
+		if bytes, isByteSlice := nvp.Value.([]byte); isByteSlice {
+			payloadItem = &commonpb.PayloadItem{
+				Metadata: map[string][]byte{
+					metadataEncoding: []byte(metadataEncodingRaw),
+					metadataName:     []byte(nvp.Name),
+				},
+				Data: bytes,
+			}
+			// } else if protoValue, isProto := nvp.Value.(proto.Marshaler); isProto {
```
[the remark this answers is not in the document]
alexshtin: Removed this commented code.

internal/internal_activity.go
```diff
@@ -267,16 +267,19 @@ func validateFunctionAndGetResults(f interface{}, values []reflect.Value, dataCo
...
 	// Parse result
 	if resultSize > 1 {
 		retValue := values[0]
-		if retValue.Kind() != reflect.Ptr || !retValue.IsNil() {
-			result, err = encodeArg(dataConverter, retValue.Interface())
-			if err != nil {
-				return nil, err
+
+		var ok bool
+		if result, ok = retValue.Interface().(*commonpb.Payload); !ok {
```
alexshtin, line 277: I had to add this check because previously bytes just went though `encodeArg` and now it ends up to be double encoded.

internal/internal_event_handlers.go
```diff
@@ -928,13 +928,12 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessQuery(queryType string, que
 			return nil, err
 		}
 
-		resultSize := len(result)
-		if resultSize > queryResultSizeLimit {
+		if result.Size() > queryResultSizeLimit {
```
alexshtin, line 931: This includes metadata also.

internal/internal_workflow_testsuite_test.go
```diff
@@ -1390,53 +1388,90 @@ func (s *WorkflowTestSuiteUnitTest) Test_MockUpsertSearchAttributes() {
...
 	var actualValues []string
-	retVal := &executionpb.WorkflowExecution{WorkflowId: "retwID2", RunId: "retrID2"}
+	retVal := "retVal"
 
-	// Passing one argument
-	activitySingleFn := func(ctx context.Context, wf *executionpb.WorkflowExecution) (*executionpb.WorkflowExecution, error) {
-		actualValues = append(actualValues, wf.GetWorkflowId())
-		actualValues = append(actualValues, wf.GetRunId())
-		return retVal, nil
+	activitySingleFn := func(ctx context.Context, s1 string, s2 *string, s3 **string) (*string, error) {
```
alexshtin, line 1395: `json.Marshal` really has some magic inside.

test/integration_test.go
```diff
@@ -434,7 +434,7 @@ func (ts *IntegrationTestSuite) TestLargeQueryResultError() {
 	ts.Error(err)
 
 	ts.IsType(&serviceerror.QueryFailed{}, err)
-	ts.Equal("query result size (3000000) exceeds limit (2000000)", err.Error())
+	ts.Equal("query result size (3000046) exceeds limit (2000000)", err.Error())
```
alexshtin, line 437: This is because for metadata.

internal/encoded.go
```diff
@@ -25,7 +25,20 @@
...
+	"fmt"
 	"reflect"
+
+	commonpb "go.temporal.io/temporal-proto/common"
+)
+
+const (
+	metadataEncoding     = "encoding"
+	metadataEncodingRaw  = "raw"
+	metadataEncodingJSON = "json"
```
mfateev, line 39: What about adding primitive types? At least numbers and strings? Now we are wrapping them as Json.

test/replaytests/basic.json
```diff
@@ -13,7 +13,7 @@
         "taskList": {
           "name": "helloWorldGroup"
         },
-        "input": "IkNhZGVuY2UiCg==",
+        "input": null,
```
mfateev, line 16: ideally we should regenerate them. But don't bother for now.

internal/encoded.go
```diff
@@ -47,59 +60,139 @@ type (
...
+		if bytes, isByteSlice := nvp.Value.([]byte); isByteSlice {
+			payloadItem = &commonpb.PayloadItem{
+				Metadata: map[string][]byte{
+					metadataEncoding: []byte(metadataEncodingRaw),
+					metadataName:     []byte(nvp.Name),
+				},
+				Data: bytes,
+			}
+		} else {
+			data, err := json.Marshal(nvp.Value)
```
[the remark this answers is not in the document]
alexshtin: There is A REASON!

internal/internal_workflow_testsuite_test.go
```diff
@@ -1710,17 +1745,18 @@ func (s *WorkflowTestSuiteUnitTest) Test_WorkflowLocalActivityWithMockAndListene
 		ctx = WithLocalActivityOptions(ctx, s.localActivityOptions)
 		var result string
 		f := ExecuteLocalActivity(ctx, localActivityFn, "local_activity")
+
+		// Hack to avoid race condition. Never do anything similar in real production code
+		// TODO: Fix race condition somewhere in test environment
+		time.Sleep(100 * time.Millisecond)
```
alexshtin, line 1751: I will file an issue for this.

PR #111, opened 2020-04-28
Convert Header message to use Payload

alexshtin:
Last replacement of `bytes` with `Payload` according to https://github.com/temporalio/temporal/issues/107.

### DIFF
--- a/internal/tracer.go
+++ b/internal/tracer.go
@@ -28,25 +28,34 @@ import (
 	"context"
 
 	"github.com/opentracing/opentracing-go"
+	commonpb "go.temporal.io/temporal-proto/common"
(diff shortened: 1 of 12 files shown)


internal/tracer.go
```diff
@@ -28,25 +28,34 @@ import (
 	"context"
 
 	"github.com/opentracing/opentracing-go"
+	commonpb "go.temporal.io/temporal-proto/common"
 	"go.uber.org/zap"
 )
 
+// Must impement opentracing.TextMapReader
```
alexshtin, line 35: This is not obvious.

mfateev: May be add real check that would break compilation if not implemented?

PR #117, opened 2020-05-01
Use Payload for SearchAttributes, Memo, and Headers

alexshtin:
`SearchAttributes`, `Memo`, and `Headers` always have single encoded value. There is no reason to use `Payloads` there.

### DIFF
--- a/encoded/encoded.go
+++ b/encoded/encoded.go
@@ -56,7 +56,7 @@ func GetDefaultDataConverter() DataConverter {
 	return internal.DefaultDataConverter
 }
 
-// GetDefaultDataConverter return default data converter used by Temporal worker
+// GetDefaultPayloadConverter return default data converter used by Temporal worker
 func GetDefaultPayloadConverter() PayloadConverter {
 	return internal.DefaultPayloadConverter
 }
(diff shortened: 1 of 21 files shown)


encoded/encoded.go
```diff
@@ -46,9 +46,17 @@ type (
 	// Temporal support using different DataConverters for different activity/childWorkflow in same workflow.
 	//   2. Activity/Workflow worker that run these activity/childWorkflow, through worker.Options.
 	DataConverter = internal.DataConverter
```
alexshtin, line 48: This will be renamed to `PayloadsConverter` later.

internal/encoded.go
```diff
@@ -71,22 +71,32 @@ type (
...
-	defaultDataConverter struct{}
+	// PayloadConverter converts single value to/from payload.
+	PayloadConverter interface {
+		// ToData single value to payload.
+		ToData(value interface{}) (*commonpb.Payload, error)
+		// FromData single value from payload.
+		FromData(input *commonpb.Payload, valuePtr interface{}) error
+	}
+
+	defaultPayloadConverter struct{}
```
alexshtin, line 85: For now this is the only one implementation. And name is not good. I will refactor it and will make it plugable later.

PR #119, opened 2020-05-04
Fix integration tests

alexshtin:
Fix integration tests that were broken with recent refactorings.

### DIFF
--- a/test/activity_test.go
+++ b/test/activity_test.go
@@ -56,7 +56,7 @@ func newActivities() *Activities {
 }
 
 func (a *Activities) RetryTimeoutStableErrorActivity() error {
-	time.Sleep(time.Second * 3)
+	time.Sleep(1 * time.Second)
 	return errFailOnPurpose
 }
 
--- a/test/activity_test.go
+++ b/test/activity_test.go
@@ -56,6 +56,7 @@ func newActivities() *Activities {
 }
 
 func (a *Activities) RetryTimeoutStableErrorActivity() error {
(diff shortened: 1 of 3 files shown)


test/workflow_test.go
```diff
@@ -434,12 +434,14 @@ func (w *Workflows) ConsistentQueryWorkflow(ctx workflow.Context, delay time.Dur
...
+		ScheduleToStartTimeout: 1 * time.Second,
+		StartToCloseTimeout:    2 * time.Second,
+		ScheduleToCloseTimeout: 10 * time.Second,
 		RetryPolicy: &temporal.RetryPolicy{
-			InitialInterval:    time.Second,
+			InitialInterval:    1 * time.Second,
 			BackoffCoefficient: 1.0,
-			MaximumInterval:    time.Second,
+			MaximumInterval:    1 * time.Second,
+			MaximumAttempts:    3,
```
alexshtin, line 444: This is sort of workarounds to make `TestActivityRetryOnTimeoutStableError` test works. But it should fail on `ScheduleToCloseTimeout` not on `MaximumAttempts`. #120 is to fix it.

test/activity_test.go
```diff
@@ -56,7 +56,7 @@ func newActivities() *Activities {
 }
 
 func (a *Activities) RetryTimeoutStableErrorActivity() error {
-	time.Sleep(time.Second * 3)
+	time.Sleep(1 * time.Second)
```
mfateev, line 59: Add TODO with issue #

test/workflow_test.go
mfateev: Add comment with #120

PR #121, opened 2020-05-07
RequestCancelActivity decision sends scheduleID instead of activityID

samarabbas:
Moved logic for generation of ID to decisionsHelper which relies on
DecisionTaskStarted eventID as the offset and then predicts actual
history event ID using it.


internal/internal_decision_state_machine.go
```diff
@@ -114,6 +115,7 @@ type (
 	}
 
 	decisionsHelper struct {
+		counter          int64
```
mfateev, line 118: Can we rename this to something more meaningful? Isn't it "nextDecisionEventId"?

internal/internal_decision_state_machine.go
```diff
@@ -695,6 +701,18 @@ func newDecisionsHelper() *decisionsHelper {
 	}
 }
 
+func (h *decisionsHelper) resetCounterOffset(decisionTaskStartedEventID int64) {
```
mfateev, line 704: s//setCurrentDecisionStartedEventId?

PR #123, opened 2020-05-09
Remove GRPCDialer

shawnhathaway:
Removing GRPCDialer - we will later introduce an object that settings would be set on and we would apply against our defaults.

### DIFF
--- a/internal/grpc_dialer.go
+++ b/internal/grpc_dialer.go
@@ -36,8 +36,8 @@ import (
 )
 
 type (
-	// ConnectionParameters are passed to GRPCDialer and must be used to create gRPC connection.
-	ConnectionParameters struct {
+	// connectionParameters are passed to GRPCDialer and must be used to create gRPC connection.
+	connectionParameters struct {
 		HostPort             string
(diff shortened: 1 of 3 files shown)


internal/grpc_dialer.go
```diff
@@ -36,15 +36,12 @@ import (
 )
 
 type (
-	// GRPCDialerParams are passed to GRPCDialer and must be used to create gRPC connection.
-	GRPCDialerParams struct {
+	// ConnectionParameters are passed to GRPCDialer and must be used to create gRPC connection.
+	ConnectionParameters struct {
```
alexshtin, line 40: Why do we need this `struct` now? It was there because it was part of interface. Now it doesn't. I would just replace it with 3 params.

shawnhathaway: I was going to extend this with an additional client provided object in the next PR. Would it be better to make that a fourth param or keep this object and extend it?

alexshtin: It's up to you. Make it `private` at least.

PR #124, opened 2020-05-10
Fixes non-determministic issue for workflows using GetVersion

samarabbas:


mfateev:
Make sure that the following cases are covered by integration tests. All of the following done after the code has executed originally:

- getVersion added 
- getVersion removed 
- getVersion call with different changeID added just before the call that was present on the original execution.
- Multiple getVersion call with different changeIDs added just before the call that was present on the original execution.
- getVersion call is replaced by a getVersion call with a different changeId
- getVersion calls is replaced by multiple getVersion calls with different changeIds
- getVersion call is surrounded by getVersion calls with different changeIDs

PR #130, opened 2020-05-13
Add ConnectionOptions, Support TLS

shawnhathaway:
This change exposes a new object in ```ClientOptions``` to SDK consumers - ```ConnectionOptions```; this object will be used to support additional connection scenarios, rather than supporting direct creation control of grpc dependencies.

Initially, TLS configuration is the only supported option; this takes in a golang `tls.Config` leaving complete control up to users.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -374,8 +374,12 @@ type (
 		Tracer opentracing.Tracer
 
 		// Optional: Sets ContextPropagators that allows users to control the context information passed through a workflow
-		// default: no ContextPropagators
+		// default: nil
 		ContextPropagators []ContextPropagator
+
+		// Optional: Sets options for server connection that allow users to control features of connections such as TLS settings.
+		// default: nil
+		ConnectionOptions ConnectionOptions
 	}
 
 	// StartWorkflowOptions configuration parameters for starting a workflow execution.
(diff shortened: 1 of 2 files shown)


internal/grpc_dialer.go
```diff
@@ -53,8 +61,13 @@ const (
 )
 
 func dial(params connectionParameters) (*grpc.ClientConn, error) {
```
alexshtin, line 63: I believe it should be rather `dialParameters` or `connect()`.

internal/client.go
```diff
@@ -564,6 +564,15 @@ func NewClient(options ClientOptions) (Client, error) {
 	return NewServiceClient(workflowservice.NewWorkflowServiceClient(connection), connection, options), nil
 }
 
+func generateConnectionParams(options *ClientOptions) connectionParameters {
```
alexshtin, line 567:
```suggestion
func newConnectionParameters(options ClientOptions) connectionParameters {
```

internal/client.go
```diff
@@ -374,8 +374,12 @@ type (
 		Tracer opentracing.Tracer
 
 		// Optional: Sets ContextPropagators that allows users to control the context information passed through a workflow
-		// default: no ContextPropagators
+		// default: nil
 		ContextPropagators []ContextPropagator
+
+		// Optional: Sets options for server connection that allow users to control features of connections such as TLS settings.
+		// default: nil
```
alexshtin, line 381:
```suggestion
		// default: no extra options
```
It can't be `nil`.

internal/grpc_dialer.go
```diff
@@ -53,8 +61,13 @@ const (
...
+	grpcSecurityOptions := grpc.WithInsecure()
+	if params.UserOptions.TLS != nil {
+		grpcSecurityOptions = grpc.WithTransportCredentials(credentials.NewTLS(params.UserOptions.TLS))
+	}
```
alexshtin, line 67:
```suggestion
	var grpcSecurityOptions grpc.DialOption
	if params.UserOptions.TLS != nil {
		grpcSecurityOptions = grpc.WithTransportCredentials(credentials.NewTLS(params.UserOptions.TLS))
	} else {
		grpcSecurityOptions = grpc.WithInsecure()
	}

```

internal/grpc_dialer.go
shawnhathaway: Dial was its original name and whole file is named ```grpc_dialer```

PR #131, opened 2020-05-14
CloseConnection doesn't return error

alexshtin:
but log error instead. Had to move logger to `ClientOptions` from `WorkerOptions` for this.

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -35,6 +35,7 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/pborman/uuid"
 	"github.com/uber-go/tally"
+	"go.uber.org/zap"
 
 	commonpb "go.temporal.io/temporal-proto/common"
 	eventpb "go.temporal.io/temporal-proto/event"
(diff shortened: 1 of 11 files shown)


internal/internal_workflow_client.go
```diff
@@ -942,12 +945,13 @@ func (wc *WorkflowClient) DescribeTaskList(ctx context.Context, taskList string,
...
 
 // CloseConnection closes underlying gRPC connection.
-func (wc *WorkflowClient) CloseConnection() error {
+func (wc *WorkflowClient) CloseConnection() {
 	if wc.connectionCloser == nil {
-		return nil
+		return
+	}
+	if err := wc.connectionCloser.Close(); err != nil {
+		wc.logger.Error("unable to close connection", zap.Error(err))
```
mfateev, line 953: I would change to warning

internal/internal_workflow_client.go
```diff
@@ -1010,23 +1014,24 @@ func (dc *namespaceClient) Describe(ctx context.Context, name string) (*workflow
...
 // CloseConnection closes underlying gRPC connection.
-func (dc *namespaceClient) CloseConnection() error {
-	if dc.connectionCloser == nil {
-		return nil
+func (nc *namespaceClient) CloseConnection() {
+	if nc.connectionCloser == nil {
+		return
+	}
+	if err := nc.connectionCloser.Close(); err != nil {
+		nc.logger.Error("unable to close connection", zap.Error(err))
```
mfateev, line 1033: ditto

PR #134, opened 2020-05-15
Add tests for activity struct arg

alexshtin:
Add tests to be sure that struct arguments can be passed to activity.


internal/internal_worker_test.go
```diff
@@ -1117,6 +1129,12 @@ func testActivityReturnEmptyString() (string, error) {
 	return "", nil
 }
 
+type testActivityArg struct {
+	Index int
+	Name  string
+	Data  []byte
```
mfateev, line 1135: May be add a pointer field?

PR #138, opened 2020-05-20
Use Failure and FailureInfo for errors

alexshtin:
`string reason`/`common.Payloads details` pairs were replaced with one `failure.Failure failure` field which holds all information about errors on SDK side. Corresponding go errors were added.

More details: https://github.com/temporalio/temporal/issues/89


internal/session.go
```diff
@@ -288,17 +288,11 @@ func createSession(ctx Context, creationTasklist string, options *SessionOptions
...
 	retryPolicy := &RetryPolicy{
 		InitialInterval:    time.Second,
 		BackoffCoefficient: 1.1,
 		MaximumInterval:    time.Second * 10,
-		NonRetriableErrorReasons: []string{
-			"temporalInternal:Panic",
-			"temporalInternal:Generic",
-			"temporalInternal:Timeout TimeoutTypeStartToClose",
-			"temporalInternal:Timeout TimeoutTypeHeartbeat",
-		},
```
alexshtin, line 301: I don't understand why this block was there.

internal/internal_event_handlers.go
```diff
@@ -998,18 +1003,21 @@ func (weh *workflowExecutionEventHandlerImpl) handleActivityTaskTimedOut(event *
 		return nil
 	}
 
-	var err error
 	attributes := event.GetActivityTaskTimedOutEventAttributes()
-	if len(attributes.GetLastFailureReason()) > 0 && attributes.GetTimeoutType() == commonpb.TimeoutType_StartToClose {
-		// When retry activity timeout, it is possible that previous attempts got other customer timeout errors.
-		// To stabilize the error type, we always return the customer error.
-		// See more details of background: https://github.com/temporalio/temporal/issues/185
```
alexshtin, line 1006: I believe this is resolved now in a very elegant way.

internal/error.go
```diff
@@ -369,3 +437,228 @@ func newUnknownExternalWorkflowExecutionError() *UnknownExternalWorkflowExecutio
...
+func getErrorType(err error) string {
+	var t reflect.Type
+	for t = reflect.TypeOf(err); t.Kind() == reflect.Ptr; t = t.Elem() {
+	}
+
+	return t.Name()
+}
+
+// convertErrorToFailure converts error to failure.
+func convertErrorToFailure(err error, dc DataConverter) *failurepb.Failure {
```
alexshtin, line 528: These 2 `funcs` are the core of this PR.

internal/internal_event_handlers.go
```diff
@@ -985,8 +984,14 @@ func (weh *workflowExecutionEventHandlerImpl) handleActivityTaskFailed(event *ev
...
 
 	attributes := event.GetActivityTaskFailedEventAttributes()
-	err := constructError(attributes.Reason, attributes.Details, weh.GetDataConverter())
-	activity.handle(nil, err)
+	activityTaskErr := NewActivityTaskError(
+		attributes.GetScheduledEventId(),
+		attributes.GetStartedEventId(),
+		attributes.GetIdentity(),
+		convertFailureToError(attributes.GetFailure(), weh.GetDataConverter()),
+	)
```
alexshtin, line 992: Here and below, actual error is wrapped inside `NewActivityTaskError`/`NewChildWorkflowExecutionError`.

internal/internal_workflow_client.go
```diff
@@ -1121,14 +1121,21 @@ func (workflowRun *workflowRunImpl) Get(ctx context.Context, valuePtr interface{
...
 	default:
-		err = fmt.Errorf("unexpected event type %s when handling workflow execution result", closeEvent.GetEventType())
+		return fmt.Errorf("unexpected event type %s when handling workflow execution result", closeEvent.GetEventType())
 	}
+
+	err = NewWorkflowExecutionError(
+		workflowRun.workflowID,
+		workflowRun.currentRunID,
+		getWorkflowFunctionName(workflowRun.registry, workflowRun.workflowFn),
+		err)
```
alexshtin, line 1137: In case of workflow execution error, `WorkflowExecutionError` is returned with actual error inside.

internal/error.go
```diff
@@ -27,82 +27,98 @@ package internal
...
-		details Values
-	}
-
-	// GenericError returned from workflow/workflow when the implementations return errors other than from NewCustomError() API.
-	GenericError struct {
-		err string
+	// ApplicationError returned from activity implementations with message and optional details.
+	ApplicationError struct {
+		message      string
+		orginalType  string
```
samarabbas, line 112: nit: misspelled

internal/error.go
```diff
@@ -27,82 +27,98 @@ package internal
...
-	CustomError struct {
-		reason  string
-		details Values
-	}
-
-	// GenericError returned from workflow/workflow when the implementations return errors other than from NewCustomError() API.
-	GenericError struct {
-		err string
+	// ApplicationError returned from activity implementations with message and optional details.
+	ApplicationError struct {
```
samarabbas, line 110: What about deserialization errors?  Example unable to deserialize signal data.

internal/error.go
```diff
@@ -369,3 +479,260 @@ func newUnknownExternalWorkflowExecutionError() *UnknownExternalWorkflowExecutio
...
+		return data
+	case *EncodedValues:
+		return d.values
+	default:
+		panic(fmt.Sprintf("unknown error details type %T", details))
+	}
+}
+
+// IsRetryable returns if error retryable or not.
+func IsRetryable(err error, nonRetryableTypes []string) bool {
```
samarabbas, line 536: Can we have a unit test for retry logic?

internal/error.go
```diff
@@ -369,3 +479,260 @@ func newUnknownExternalWorkflowExecutionError() *UnknownExternalWorkflowExecutio
...
+func getErrorType(err error) string {
+	var t reflect.Type
+	for t = reflect.TypeOf(err); t.Kind() == reflect.Ptr; t = t.Elem() {
+	}
+
+	return t.Name()
+}
+
+// convertErrorToFailure converts error to failure.
+func convertErrorToFailure(err error, dc DataConverter) *failurepb.Failure {
```
samarabbas, line 594: Please add unit test for this logic.

internal/error.go
```diff
@@ -369,3 +479,260 @@ func newUnknownExternalWorkflowExecutionError() *UnknownExternalWorkflowExecutio
...
+
+	if causeErr := errors.Unwrap(err); causeErr != nil {
+		failure.Cause = convertErrorToFailure(causeErr, dc)
+	}
+
+	return failure
+}
+
+// convertFailureToError converts failure to error.
+func convertFailureToError(failure *failurepb.Failure, dc DataConverter) error {
```
samarabbas, line 680: Add unit tests for this API too.

internal/error.go
alexshtin: Deserialization of signal data is in user workflow/activity code, right? In case of any user error in workflow or activity code `ApplicationError` will be returned. Users can return:
1. `ApplicationError` with custom details and it will go through as is,
2. `MyApplicationError` (or any other error) which will be converted to `ApplicationError` and `"MyApplicationError"` string will be stored in `originalType` field,
3. `errors.New("my error")` which will be converted to `ApplicationError` and message will be stored in `message` field.

I think this should be enough to pass information to error handlers.

internal/error.go
```diff
@@ -27,82 +27,98 @@ package internal
...
-	code could check to determine what kind of error it was and take actions based on the reason. The details is encoded
-	[]byte which workflow code could extract strong typed data. Workflow code needs to know what the types of the encoded
-	details are before extracting them.
-2) *GenericError:
-	If activity implementation returns errors other than from NewCustomError() API, workflow code would receive *GenericError.
-	Use err.Error() to get the string representation of the actual error.
-3) *CanceledError:
-	If activity was canceled, workflow code will receive instance of *CanceledError. When activity cancels itself by
+If activity fails then *ActivityTaskError is returned to the workflow code. The error has important information about activity
+and actual error which caused activity failure. This internal error can be unrapped using errors.Unwrap() or checked using errors.As().
```
samarabbas, line 38: *unwrapped

internal/error.go
samarabbas: I don't think user will have any control over returning this error.  This needs to be handled by SDK and returned back to application.  Here are some scenarios:

1. Deserialization of Activity parameters
2. Serialization of Activity completion result or failures
3. Serialization of Heartbeat data
4. Deserialization of Workflow input parameters
5. Serialization of Workflow completion and failures
6. Deserialization of signal data.  Needs to be communicated through signal channel returning back error.

samarabbas: I think we should make is super straightforward to handle serialization/deserialization errors.

internal/error.go
```diff
@@ -27,82 +27,98 @@ package internal
...
+		nonRetryable bool
+		details      Values
 	}
 
 	// TimeoutError returned when activity or child workflow timed out.
 	TimeoutError struct {
-		timeoutType commonpb.TimeoutType
-		details     Values
+		timeoutType          commonpb.TimeoutType
+		lastErr              error
```
samarabbas, line 120: Can we include some documentation here?  
I assume error here could be of following types:
1. ActivityTaskError
2. ChildWorkflowTaskError

internal/error.go
```diff
@@ -369,3 +479,260 @@ func newUnknownExternalWorkflowExecutionError() *UnknownExternalWorkflowExecutio
...
+		failureInfo := &failurepb.ApplicationFailureInfo{
+			Type:         getErrorType(&PanicError{}),
+			NonRetryable: true,
+		}
+		failure.FailureInfo = &failurepb.Failure_ApplicationFailureInfo{ApplicationFailureInfo: failureInfo}
+		failure.StackTrace = err.StackTrace()
+	case *TimeoutError:
+		failureInfo := &failurepb.TimeoutFailureInfo{
+			TimeoutType:          err.timeoutType,
+			LastFailure:          convertErrorToFailure(err.lastErr, dc),
```
samarabbas, line 633: Can you confirm where lastErr is coming from?  Is this on the Timeout history event?  It would be great if we can add an integration test for Timeout error with lastFailure and heartbeat details.

internal/error.go
```diff
@@ -369,3 +479,260 @@ func newUnknownExternalWorkflowExecutionError() *UnknownExternalWorkflowExecutio
...
+			LastHeartbeatDetails: convertErrDetailsToPayloads(err.lastHeartbeatDetails, dc),
+		}
+		failure.FailureInfo = &failurepb.Failure_TimeoutFailureInfo{TimeoutFailureInfo: failureInfo}
+	case *TerminatedError:
```
samarabbas, line 637: In what situation we will get this failure from the application?  Looks like dead code to me.

internal/internal_event_handlers.go
```diff
@@ -1269,7 +1316,7 @@ func (weh *workflowExecutionEventHandlerImpl) handleRequestCancelExternalWorkflo
 	// for cancellation of external workflow, we have to use cancellation ID
 	attribute := event.GetRequestCancelExternalWorkflowExecutionInitiatedEventAttributes()
 	workflowID := attribute.WorkflowExecution.GetWorkflowId()
-	cancellationID := string(attribute.Control)
+	cancellationID := attribute.Control
```
samarabbas, line 1319: I think control is no longer needed.  At least we should create a separate issue to see if we can get rid of Control as the identifier.

internal/error.go
alexshtin: Yes, you right. This is SDK error. Not user one. In this case it still will be converted to `ApplicationError` because all `errors.New/fmt.Errorf` are `ApplicationError`s.

The main idea behind this PR though is to convert errors to protobuf messages every time when they need to be transmitted between SDK and server. From this prospective all these errors are `ApplicationError`s. I can create special types for serialization/deserialization errors to simplify error handling in workflow code but let's discuss it first and this gonna be separate PR.

internal/internal_event_handlers.go
alexshtin: We had this discussion before and decided to leave it as `string`. Here, I just removed redundant type conversion.

internal/error.go
alexshtin: Yes, you are probably right. But as soon as we have both `TerminatedFailureInfo` and `TerminatedError` I decided to leave this 1:1 mapping. Can `TerminatedError` be the cause for, let's say, child workflow failure? Then it might be called as part of `cause` conversion.

internal/error.go
alexshtin: It comes from `ActivityTaskTimedOutEventAttributes.LastFailure` inside `handleActivityTaskTimedOut` handler. This `LastFailure` is set on server, apparently, to the last failure from history. This logic hasn't changed in this PR but so far I don't see that it's getting set correctly. This is actually the part of my next task (and another PR).

internal/error.go
alexshtin: No, it is actual last error from activity itself (i.e. `ApplicationError` or another `TimeoutError`). I can't add it right now but I will do with in next timeout repated PR.

PR #142, opened 2020-05-25
Expose low level API to implement workflows asynchronously

mfateev:
Go SDK can be used as a library for implementing workflows in other languages. But the existing high level Go API is not suitable for such integrations. This PR exposes the low level async API appropriate for such integrations.


test/bindings_workflows_test.go
```diff
@@ -0,0 +1,140 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package test_test
+
+import (
+	"context"
+	commonpb "go.temporal.io/temporal-proto/common"
+	"go.temporal.io/temporal/encoded"
+	b "go.temporal.io/temporal/internalbindings"
```
alexshtin, line 31: Can you alias it to `bindings`. `b` is like a reciever var.

test/bindings_workflows_test.go
```diff
@@ -0,0 +1,140 @@
+// The MIT License
```
alexshtin, line 1: This file doesn't have any tests. It is more like `workflow_definition_factory.go`.

test/activity_test.go
```diff
@@ -22,7 +22,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package test
+package test_test
```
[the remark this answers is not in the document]
mfateev: I believe it makes it private to anyone who includes the package.

test/bindings_workflows_test.go
mfateev: removing _test prefix makes it public :(

PR #143, opened 2020-05-26
Log timed out activity result

alexshtin:
If activity timed out its result just ignored. This PR adds logging of that returned result/error.


internal/internal_task_handlers.go
```diff
@@ -1799,6 +1799,13 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *workflowservice.
 
 	dlCancelFunc()
 	if <-ctx.Done(); ctx.Err() == context.DeadlineExceeded {
+		ath.logger.Info("Activity compete after timeout.",
```
mfateev, line 1802: Compete with whom?

alexshtin: :facepalm:

PR #145, opened 2020-05-27
Move all public errors to temporal package

alexshtin:


error.go
```diff
@@ -27,12 +27,85 @@ package temporal
...
+
+Workflow code could handle errors based on different types of error. Below is sample code of how error handling looks like.
+
+_, err := workflow.ExecuteActivity(ctx, MyActivity, ...).Get(nil)
+if err != nil {
+	var applicationErr *ApplicationError
+	if errors.As(err, &applicationError) {
+		// handle activity errors (created via NewApplicationError() API)
+		if !applicationErr.NonRetryable() {
+			// manually retry activity
```
mfateev, line 69: I'm not sure why anyone would ever check this flag as this is mostly for the service to deal with it.

error.go
```diff
@@ -27,12 +27,85 @@ package temporal
...
+	}
+
+	var panicErr *PanicError
+	if errors.As(err, &panicErr) {
+		// handle panic, message and stack trace are available by panicErr.Error() and panicErr.StackTrace()
+	}
+}
+
+Errors from child workflow should be handled in a similar way, except that instance of *ChildWorkflowExecutionError is returned to
+workflow code. It will contains *ActivityTaskError, which in turn will contains on of the errors above.
```
mfateev, line 102: This is confusing. It might contain for example an application error that a child workflow returned directly

PR #150, opened 2020-06-04
Store original failure in error

alexshtin:
If error was created from failure this original failure will be stored inside error and then used to serialize back to failure. This will persist fields which go code is not aware about (like `StackTrace` for `ApplicationFailureInfo`).

### DIFF
--- a/internal/error.go
+++ b/internal/error.go
@@ -117,6 +117,7 @@ Workflow consumers will get an instance of *WorkflowExecutionError. This error w
 type (
 	// ApplicationError returned from activity implementations with message and optional details.
 	ApplicationError struct {
+		temporalError
 		message      string
 		originalType string
 		nonRetryable bool
@@ -125,6 +126,7 @@ type (
 
 	// TimeoutError returned when activity or child workflow timed out.
 	TimeoutError struct {
+		temporalError
 		timeoutType          commonpb.TimeoutType
 		lastErr              error
 		lastHeartbeatDetails Values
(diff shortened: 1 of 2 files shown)


internal/error.go
```diff
@@ -198,6 +206,15 @@ type (
 		workflowType string
 		cause        error
 	}
+
+	temporalError struct {
+		originalFailure *failurepb.Failure
+	}
```
alexshtin, line 212: This like a base error class.

internal/error.go
```diff
@@ -689,46 +720,48 @@ func convertFailureToError(failure *failurepb.Failure, dc DataConverter) error {
...
 
+	var err error
+
 	if failure.GetApplicationFailureInfo() != nil {
 		applicationFailureInfo := failure.GetApplicationFailureInfo()
 		details := newEncodedValues(applicationFailureInfo.GetDetails(), dc)
 		switch applicationFailureInfo.GetType() {
 		case getErrorType(&ApplicationError{}):
-			return NewApplicationError(failure.GetMessage(), applicationFailureInfo.GetNonRetryable(), details)
+			err = NewApplicationError(failure.GetMessage(), applicationFailureInfo.GetNonRetryable(), details)
```
mfateev, line 730: Can user create an application error that contains any other error as a cause?

internal/error_test.go
```diff
@@ -812,3 +832,35 @@ func Test_convertFailureToError_ServerFailure(t *testing.T) {
 	require.Equal("message", serverErr.Error())
 	require.Equal(true, serverErr.nonRetryable)
 }
+
```
mfateev, line 835: Add tests for the chained errors and failures.

internal/error.go
alexshtin: No (`ApplicationError` does't have `cause`), but I can easily add this. Is there a requirement for this?

internal/error_test.go
alexshtin: There is `Test_convertErrorToFailure_ActivityTaskError` and `Test_convertErrorToFailure_ChildWorkflowExecutionError`. I will add chains to new tests also.

internal/error.go
mfateev: For example It is possible to create application error because an activity failed.

PR #151, opened 2020-06-04
Refactor markers

alexshtin:
1. Add `ActivityType` and `ActivityId` to `ActivityTaskError`.
2. Move `Failure` and `Result` out of `localActivityMarkerData`.


internal/internal_event_handlers.go
```diff
@@ -1067,38 +1077,48 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
...
-	encodedValues := newEncodedValues(attributes.Details, weh.dataConverter)
 	switch attributes.GetMarkerName() {
 	case sideEffectMarkerName:
-		var sideEffectID int64
-		var result *commonpb.Payloads
-		_ = encodedValues.Get(&sideEffectID, result)
-		weh.sideEffectResult[sideEffectID] = result
+		for sideEffectID, result := range attributes.GetDetails() {
+			weh.sideEffectResult[sideEffectID] = result
+		}
```
alexshtin, line 1084: Before only one `sideEffectID` (and all other markers bellow) was supported. Now it copies all of them from `Details`. I can add `break` here to have only first item copied. There should be only one anyway.

internal/internal_event_handlers.go
```diff
@@ -1067,38 +1077,48 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
...
-func (weh *workflowExecutionEventHandlerImpl) handleLocalActivityMarker(markerData *commonpb.Payloads) error {
-	lamd := localActivityMarkerData{}
+func (weh *workflowExecutionEventHandlerImpl) handleLocalActivityMarker(details map[string]*commonpb.Payloads, failure *failurepb.Failure) error {
+	if details == nil {
+		return ErrMissingLocalActivityMarkerDetails
+	}
 
+	var markerData *commonpb.Payloads
+	var ok bool
+	if markerData, ok = details[localActivityMarkerDataDetailsName]; !ok {
```
mfateev, line 1117: What about failure? Are details always popluated in this case?

internal/internal_event_handlers.go
mfateev: The cleaner approach would be to have fixed keys: "result", "id" and put payload and id into values.

internal/internal_event_handlers.go
alexshtin: Yes, `result` will not be there, but `data` is always added.

internal/internal_event_handlers.go
```diff
@@ -1077,45 +1075,69 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
...
-	case mutableSideEffectMarkerName:
-		for fixedID, result := range attributes.GetDetails() {
-			weh.mutableSideEffect[fixedID] = result
+	var err error
+	if attributes.GetDetails() == nil {
+		err = ErrMissingMarkerDetails
+	} else {
+		switch attributes.GetMarkerName() {
+		case sideEffectMarkerName:
+			if sideEffectIDPayload, ok := attributes.GetDetails()[sideEffectMarkerIDName]; !ok {
```
mfateev, line 1084: consider moving all the cases into separate functions the same way handleLocalActivityMarker is already done.

alexshtin: I had this idea too but decided to leave it as is for now.

PR #152, opened 2020-06-04
Add cause to ApplicationError

alexshtin:
`ApplicationError` has `cause` field now. This field can contains any other error or error chain. It is converted to `Failure.Cause` field.


error.go
```diff
@@ -149,14 +149,14 @@ var ErrNoData = internal.ErrNoData
 
 // NewNonRetryableApplicationError creates new instance of non-retryable *ApplicationError with reason and optional details.
 // Use ApplicationError for any use case specific errors that cross activity and child workflow boundaries.
-func NewNonRetryableApplicationError(reason string, details ...interface{}) *ApplicationError {
-	return internal.NewApplicationError(reason, true, details...)
+func NewNonRetryableApplicationError(reason string, cause error, details ...interface{}) *ApplicationError {
```
mfateev, line 152: Add comment that cause can be nil

mfateev: Or better add Method NewApplicationErrorWithCause and remote cause from this one

mfateev: rename reason to type

PR #165, opened 2020-06-19
Increase timeout values to prevent test flakiness

alexshtin:
Test passes on my local machine but fails in CI/CD:

image

Adding more time to prevent flakiness.

### DIFF
--- a/internal/common/backoff/retry_test.go
+++ b/internal/common/backoff/retry_test.go
@@ -58,27 +58,27 @@ func TestRetrySuccess(t *testing.T) {
 
 func TestNoRetryAfterContextDone(t *testing.T) {
 	t.Parallel()
-	i := 0
+	retryCounter := 0
 	op := func() error {
-		i++
+		retryCounter++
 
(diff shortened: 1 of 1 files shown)


markmark206:
Great, thank you for the change, hopefully this will reduce flakiness somewhat.

More generally, I wonder if there is a way for us to wring performance-based flakiness out of our tests altogether. It's almost like we want two test execution modes:
1. *general tests*: running tests on a system whose performance characteristics are not guaranteed (e. g. a general ci/cd system). in this mode, we will probably want ridiculously long timeouts, and 
2. *performance sensitive tests*: running tests on a system whose performance characteristics are under "our" control. In this mode, we are concerned with (and test) the performance of the system.

alexshtin:
It is a good idea to have some env var like `TIMEOUT_COEEFICIENT` and use it for all timeout. Set it to 1 by default and to 10 or 100 on CI/CD. All these tests are not performance tests and they are just for functionality. @mfateev, what do you think?

PR #168, opened 2020-06-23
Fix session bugs

alexshtin:
Fix Session which is not working now.
Closes https://github.com/temporalio/temporal-go-samples/issues/30.


internal/session.go
```diff
@@ -353,7 +352,7 @@ func createSession(ctx Context, creationTasklist string, options *SessionOptions
 			return
 		}
 		var canceledErr *CanceledError
-		if errors.As(err, &canceledErr) {
+		if !errors.As(err, &canceledErr) {
```
alexshtin, line 355: 2nd bug. :facepalm:

internal/internal_decision_state_machine.go
```diff
@@ -955,6 +956,7 @@ func (h *decisionsHelper) requestCancelExternalWorkflowExecution(namespace, work
 		// targeting child workflow
 		decision := h.getDecision(makeDecisionID(decisionTypeChildWorkflow, workflowID))
 		decision.cancel()
+		// TODO: add h.nextDecisionEventID++ ???
```
alexshtin, line 959: Do I need to add `h.nextDecisionEventID++` here also?

internal/internal_decision_state_machine.go
```diff
@@ -1083,6 +1085,7 @@ func (h *decisionsHelper) startTimer(attributes *decisionpb.StartTimerDecisionAt
 func (h *decisionsHelper) cancelTimer(timerID string) decisionStateMachine {
 	decision := h.getDecision(makeDecisionID(decisionTypeTimer, timerID))
 	decision.cancel()
+	// TODO: h.nextDecisionEventID++ ???
```
alexshtin, line 1088: Do I need to add `h.nextDecisionEventID++` here also?

internal/internal_decision_state_machine.go
```diff
@@ -771,6 +771,7 @@ func (h *decisionsHelper) requestCancelActivityTask(activityID string) decisionS
 	id := makeDecisionID(decisionTypeActivity, activityID)
 	decision := h.getDecision(id)
 	decision.cancel()
+	h.nextDecisionEventID++
```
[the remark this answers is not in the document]
samarabbas: I don't think this fix would work in all cases.  There are situations where calling decision.cancel() might not result in new decision being sent to server and nextDecisionEventID does not need to be incremented.

internal/internal_decision_state_machine.go
```diff
@@ -375,6 +375,9 @@ func (d *decisionStateMachineBase) cancel() {
 		d.moveState(decisionStateCanceledBeforeInitiated, eventCancel)
```
alexshtin, line 375: This turned out to be a non-dead code. At lease unit tests has some dependency on it.

test/integration_test.go
```diff
@@ -126,6 +126,11 @@ func (ts *IntegrationTestSuite) SetupTest() {
 		DisableStickyExecution:            ts.config.IsStickyOff,
 		WorkflowInterceptorChainFactories: []interceptors.WorkflowInterceptorFactory{ts.tracer},
 	}
+
+	if strings.Contains(ts.T().Name(), "Session") {
```
alexshtin, line 130: Kinda hack but I don't want to create another test suite just for one test.

test/integration_test.go
```diff
@@ -404,6 +409,35 @@ func (ts *IntegrationTestSuite) TestActivityCancelRepro() {
...
+
+func (ts *IntegrationTestSuite) TestCancelChildWorkflow() {
+	var expected []string
+	err := ts.executeWorkflow("test-cancel-child-workflow", ts.workflows.CancelChildWorkflow, &expected)
+	ts.NoError(err)
+	ts.EqualValues(expected, ts.activities.invoked())
+}
+
+func (ts *IntegrationTestSuite) TestCancelActivityImmediately() {
+	ts.T().Skip(`Currently fails with "PanicError": "unknown decision internal.decisionID{decisionType:0, id:"5"}, possible causes are nondeterministic workflow definition code or incompatible change in the workflow definition`)
```
alexshtin, line 434: This test doesn't pass now. @mfateev, you can use it to test immediate cancelation.

internal/internal_decision_state_machine.go
```diff
@@ -577,6 +580,7 @@ func (d *childWorkflowDecisionStateMachine) cancel() {
 	switch d.state {
 	case decisionStateStarted:
 		d.moveState(decisionStateCanceledAfterStarted, eventCancel)
+		d.helper.incrementNextDecisionEventID()
```
alexshtin, line 583: Child workflow uses different state not covered in base `cancel()`.

PR #169, opened 2020-06-24
WorkflowInterceptor refactoring

mfateev:
Refactored workflow interceptor to match the interface of the Java SDK. Added ability to intercept workflow goroutine creation.

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -294,7 +294,7 @@ type workflowEnvironmentInterceptor struct {
 	outboundInterceptor WorkflowOutboundCallsInterceptor
 }
 
-func (wc *workflowEnvironmentInterceptor) NewCoroutine(ctx Context, name string, f func(ctx Context)) Context {
+func (wc *workflowEnvironmentInterceptor) Go(ctx Context, name string, f func(ctx Context)) Context {
 	return wc.dispatcher.NewCoroutine(ctx, name, f)
 }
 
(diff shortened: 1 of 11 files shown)


internal/internal_workflow.go
```diff
@@ -444,32 +456,42 @@ func newWorkflowContext(env WorkflowEnvironment, interceptors WorkflowIntercepto
...
 
-func newWorkflowInterceptors(env WorkflowEnvironment, factories []WorkflowInterceptorFactory) (WorkflowInterceptor, *workflowEnvironmentInterceptor) {
+func newWorkflowInterceptors(env WorkflowEnvironment, interceptors []WorkflowInterceptor) (*workflowEnvironmentInterceptor, error) {
 	envInterceptor := &workflowEnvironmentInterceptor{env: env}
-	var interceptor WorkflowInterceptor = envInterceptor
-	for i := len(factories) - 1; i >= 0; i-- {
-		interceptor = factories[i].NewInterceptor(env.WorkflowInfo(), interceptor)
+	var interceptor WorkflowInboundCallsInterceptor = envInterceptor
+	for i := len(interceptors) - 1; i >= 0; i-- {
+		interceptor = interceptors[i].InterceptWorkflow(env.WorkflowInfo(), interceptor)
```
samarabbas, line 463: This feels weird.  Why we would intercept workflow invocation when we are setting up interceptor chain?

internal/internal_workflow.go
```diff
@@ -444,32 +456,42 @@ func newWorkflowContext(env WorkflowEnvironment, interceptors WorkflowIntercepto
 	return rootCtx
 }
 
-func newWorkflowInterceptors(env WorkflowEnvironment, factories []WorkflowInterceptorFactory) (WorkflowInterceptor, *workflowEnvironmentInterceptor) {
+func newWorkflowInterceptors(env WorkflowEnvironment, interceptors []WorkflowInterceptor) (*workflowEnvironmentInterceptor, error) {
```
samarabbas, line 459: I think the factory pattern was better in this case as we are establishing an interceptor chain based on factories registered with the registry.

internal/internal_workflow.go
mfateev: Because we create a chain per workflow invocation. This is pretty similar to gRPC interceptor: https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientInterceptor.html

internal/internal_workflow.go
mfateev: Factory is confusing IMHO as we are intercepting workflow executions

PR #172, opened 2020-06-27
Fix workflow history replay

alexshtin:
`replayWorkflowHistory` did not return an error if replay failed.


internal/internal_worker.go
```diff
@@ -1276,6 +1276,10 @@ func (aw *WorkflowReplayer) replayWorkflowHistory(logger *zap.Logger, service wo
 		return err
 	}
 
+	if failedReq, ok := resp.(*workflowservice.RespondDecisionTaskFailedRequest); ok {
+		return fmt.Errorf("replay workflow failed with failure: %v", failedReq.GetFailure().GetMessage())
```
mfateev, line 1280: I believe failure contains more info like stack trace

alexshtin: Let's use entire `failure` there then.

PR #176, opened 2020-06-29
Add type argument to NewApplicationError constructor

alexshtin:
`NewApplicationError` now has `type` argument which allows to specify error type explicitly. Use `temporal.ApplicationErrorType` for default type.

Plus few renames:
1. Field `originalType` -> `errType`
2. Func `OriginalType()` -> `Type()`

### DIFF
--- a/error.go
+++ b/error.go
@@ -150,22 +150,26 @@ var (
 	ErrNoData = internal.ErrNoData
 )
 
-// NewApplicationError creates new instance of retryable *ApplicationError with message and optional details.
+// NewApplicationError creates new instance of retryable *ApplicationError with message, type, and optional details.
 // Use ApplicationError for any use case specific errors that cross activity and child workflow boundaries.
+// errType can be used to control if error is retryable or not. Add the same type in to RetryPolicy.NonRetryableErrorTypes
+// to avoid retrying of particular error types.
 func NewApplicationError(message, errType string, details ...interface{}) *ApplicationError {
 	return internal.NewApplicationError(message, errType, false, nil, details...)
 }
 
-// NewApplicationErrorWithCause creates new instance of retryable *ApplicationError with message, cause, and optional details.
+// NewApplicationErrorWithCause creates new instance of retryable *ApplicationError with message, type, cause, and optional details.
 // Use ApplicationError for any use case specific errors that cross activity and child workflow boundaries.
+// errType can be used to control if error is retryable or not. Add the same type in to RetryPolicy.NonRetryableErrorTypes
+// to avoid retrying of particular error types.
 func NewApplicationErrorWithCause(message, errType string, cause error, details ...interface{}) *ApplicationError {
 	return internal.NewApplicationError(message, errType, false, cause, details...)
 }
 
 // NewNonRetryableApplicationError creates new instance of non-retryable *ApplicationError with message, and optional cause and details.
 // Use ApplicationError for any use case specific errors that cross activity and child workflow boundaries.
-func NewNonRetryableApplicationError(message, errType string, cause error, details ...interface{}) *ApplicationError {
-	return internal.NewApplicationError(message, errType, true, cause, details...)
+func NewNonRetryableApplicationError(message, cause error, details ...interface{}) *ApplicationError {
+	return internal.NewApplicationError(message, "", true, cause, details...)
 }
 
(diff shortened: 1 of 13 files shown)


error.go
```diff
@@ -145,25 +145,30 @@ type (
...
 )
 
-// ErrNoData is returned when trying to extract strong typed data while there is no data available.
-var ErrNoData = internal.ErrNoData
+var (
+	// ErrNoData is returned when trying to extract strong typed data while there is no data available.
+	ErrNoData = internal.ErrNoData
+
+	// ApplicationErrorType is default error type for ApplicationError.
+	ApplicationErrorType = internal.ApplicationErrorType
```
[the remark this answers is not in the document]
alexshtin: After 2nd thought I realized that this const is not needed at all.

internal/error.go
```diff
@@ -223,24 +223,26 @@ type (
...
+	// activity method returns. Activity needs to be completed by Client.CompleteActivity() separately. For example, if an
+	// activity require human interaction (like approve an expense report), the activity could return activity.ErrResultPending
+	// which indicate the activity is not done yet. Then, when the waited human action happened, it needs to trigger something
+	// that could report the activity completed event to temporal server via Client.CompleteActivity() API.
+	ErrActivityResultPending = errors.New("not error: do not autocomplete, using Client.CompleteActivity() to complete")
+)
 
-// NewApplicationError create new instance of *ApplicationError with message and optional details.
-func NewApplicationError(message string, nonRetryable bool, cause error, details ...interface{}) *ApplicationError {
+// NewApplicationError create new instance of *ApplicationError with message, type, and optional details.
```
mfateev, line 241: Add the explanation that this type is used by retry options.

internal/error.go
```diff
@@ -406,9 +408,10 @@ func (e *ApplicationError) Error() string {
 	return e.message
 }
 
-// OriginalType returns original error type represented as string.
-func (e *ApplicationError) OriginalType() string {
-	return e.originalType
+// Type returns error type represented as string.
+// This type can be passed explicitly to ApplicationError constructor or will be set automatically for any other go errors.
```
mfateev, line 412: Explain how it is set for other go errors. or even better explain in which situation other errors are converted to this.

internal/error.go
```diff
@@ -606,48 +609,54 @@ func IsRetryable(err error, nonRetryableTypes []string) bool {
...
-		applicationErrOriginalType = applicationErr.originalType
-	}
-
 	var timeoutErr *TimeoutError
 	if errors.As(err, &timeoutErr) {
 		if timeoutErr.timeoutType != enumspb.TIMEOUT_TYPE_START_TO_CLOSE &&
 			timeoutErr.timeoutType != enumspb.TIMEOUT_TYPE_HEARTBEAT {
 			return false
 		}
+		return true
```
mfateev, line 618: Use the same style you used on the server of return ... == ... ||  ... == ...

internal/error.go
```diff
@@ -606,48 +609,54 @@ func IsRetryable(err error, nonRetryableTypes []string) bool {
 		return false
```
mfateev, line 609: IsRetryable is used only by backoff.Retry function which in turn I used only by API calls to the service. I don't believe it should ever receive any of these errors.

internal/error.go
alexshtin: Added this to public alias.

error.go
```diff
@@ -150,22 +150,26 @@ var (
...
+// NewApplicationErrorWithCause creates new instance of retryable *ApplicationError with message, type, cause, and optional details.
 // Use ApplicationError for any use case specific errors that cross activity and child workflow boundaries.
+// errType can be used to control if error is retryable or not. Add the same type in to RetryPolicy.NonRetryableErrorTypes
+// to avoid retrying of particular error types.
 func NewApplicationErrorWithCause(message, errType string, cause error, details ...interface{}) *ApplicationError {
 	return internal.NewApplicationError(message, errType, false, cause, details...)
 }
 
 // NewNonRetryableApplicationError creates new instance of non-retryable *ApplicationError with message, and optional cause and details.
 // Use ApplicationError for any use case specific errors that cross activity and child workflow boundaries.
```
mfateev, line 170: Add errType

PR #187, opened 2020-07-10
Add ToString to DataConverter

mastermanu:
Adds `ToString` to `DataConverter` interface and `defaultDataConverter` so that a caller can get a "prettified" version of the `Payloads` as `string`, which can be useful for things like looking at heartbeat details and logging.

### DIFF
--- a/internal/encoded.go
+++ b/internal/encoded.go
@@ -81,9 +81,6 @@ type (
 		ToPrettyStrings(input *commonpb.Payloads) ([]string, error)
 	}
 
-	// DataStringer can be used by framework to print human readable data
-	// This interface is currently not settable
-
 	defaultDataConverter struct {
 	}
 )
--- a/internal/encoded.go
+++ b/internal/encoded.go
@@ -25,6 +25,7 @@
 package internal
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


internal/encoded.go
```diff
@@ -166,38 +170,103 @@ func (dc *defaultDataConverter) ToPayload(value interface{}) (*commonpb.Payload,
...
+	case metadataEncodingRaw:
+		return decodeEncodingRaw(payload, valuePtr)
+	case metadataEncodingJSON:
+		return decodeEncodingJSON(payload, valuePtr)
+	default:
+		return fmt.Errorf("encoding %s: %w", encoding, ErrEncodingIsNotSupported)
+	}
+}
+
+func decodeEncodingRaw(payload *commonpb.Payload, valuePtr interface{}) error {
```
alexshtin, line 193: This is what will migrate to `PayloadConverter` later.

internal/encoded.go
```diff
@@ -75,6 +76,9 @@ type (
 		// FromPayloads implements conversion of an array of values of different types.
 		// Useful for deserializing arguments of function invocations.
 		FromPayloads(input *commonpb.Payloads, valuePtrs ...interface{}) error
+
+		// ToPrettyStrings converts the Payloads object into human readable strings
+		ToPrettyStrings(input *commonpb.Payloads) ([]string, error)
```
alexshtin, line 81: `Pretty` is an ugly word :-). What's about just `ToStrings`?

PR #195, opened 2020-07-16
Remove logic for timer cancellation failed event

samarabbas:


internal/internal_task_handlers_test.go
```diff
@@ -1516,8 +1516,7 @@ func Test_NonDeterministicCheck(t *testing.T) {
 			commandEventTypeCount++
 		}
 	}
-	// CancelTimer has 2 corresponding events.
```
alexshtin, line 1519: :+1:

PR #204, opened 2020-07-23
Extract InterruptCh func

### LINKED ISSUE #94: Worker Start/Stop behavior
I'm reopening uber-go/cadence-client#642

Listening to signals should be left to the caller.

Something closer to a work actor like this would be nice:

https://github.com/sagikazarmark/appkit/pull/7

Then again, this is probably not the highest priority right now.

alexshtin:
Following code block:
```
w := worker.New(c, "hello-world", worker.Options{})
err = w.Run()
```
becomes:
```
w := worker.New(c, "hello-world", worker.Options{})
```
and it will wait for `Ctrl+C` or
```
w := worker.New(c, "hello-world", worker.Options{})
err = w.Run(nil)
```
and it will wait for `w.Stop()` call.

Or you can pass any `<-chan interface{}` and `Run` will stop the worker as soon as there is a signal in the channel.

Closes #94.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -35,11 +35,13 @@ import (
 	"io"
 	"math"
 	"os"
+	"os/signal"
 	"reflect"
 	"runtime"
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/gogo/protobuf/jsonpb"
@@ -345,16 +347,6 @@ func (ww *workflowWorker) Start() error {
 	return nil // TODO: propagate error
 }
 
-func (ww *workflowWorker) Run() error {
-	err := verifyNamespaceExist(ww.workflowService, ww.executionParameters.Namespace, ww.worker.logger)
-	if err != nil {
-		return err
-	}
-	ww.localActivityWorker.Start()
-	ww.worker.Run()
-	return nil
-}
-
 // Stop the worker.
 func (ww *workflowWorker) Stop() {
 	close(ww.stopC)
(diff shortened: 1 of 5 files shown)


internal/internal_worker.go
```diff
@@ -1083,18 +1055,24 @@ func getBinaryChecksum() string {
...
+func (aw *AggregatedWorker) Run(ch <-chan interface{}) error {
 	if err := aw.Start(); err != nil {
 		return err
 	}
-	d := <-getKillSignal()
-	aw.logger.Info("Worker has been killed", zap.String("Signal", d.String()))
-	aw.Stop()
+	select {
+	case s := <-ch:
+		aw.logger.Info("Worker has been killed", zap.String("Signal", fmt.Sprintf("%v", s)))
```
mfateev, line 1064: No 100% sure if this message is helpful if user controlled channel is passed as an argument.

worker/worker.go
```diff
@@ -114,9 +114,12 @@ type (
 
 		// Start the worker in a non-blocking fashion.
 		Start() error
-		// Run the worker in a blocking fashion. Stop the worker when process is killed with SIGINT or SIGTERM.
+		// Run the worker in a blocking fashion. Stop the worker when ch receives data.
+		// Pass worker.InterruptCh() to stop the worker with SIGINT or SIGTERM.
```
mfateev, line 118: The InterruptCh listens explicitly to sigterm only. Does it handle sigterm as well?

alexshtin: it is:
```
signal.Notify(c, os.Interrupt, syscall.SIGTERM)
```
and `os.Interrupt` is just an alias to `SIGINT`.

internal/internal_worker.go
alexshtin: Well, it is still signal. Not `os.Signal` but "signal to stop worker". Rephrased message though.

PR #214, opened 2020-07-28
Add DisableHealthCheck to ConnectionOptions

### LINKED ISSUE #86: Add ability to call healthcheck from WorkflowClient
Customer request:

> does temporal client have something similar to Ping() ?
> c, err := client.NewClient(client.Options{
> 		HostPort: <WRONG_HOSTPORT>,
> 	})
> when I start a workflow worker with the above, no error is returned until when you do something — would be nice to have something similar to SELECT 1 in postgres

My proposal is to add ability to call grpc health check endpoint of the frontends.

alexshtin:
Closes #86.


alexshtin:
Integration tests should pass after https://github.com/temporalio/temporal/pull/614 is merged.

internal/client.go
```diff
@@ -384,6 +390,12 @@ type (
 		ConnectionOptions ConnectionOptions
 	}
 
+	// ConnectionOptions is provided by SDK consumers to control optional connection params.
+	ConnectionOptions struct {
```
alexshtin, line 394: I moved this `struct` from another file and added `DisableHealthCheck` field.

internal/client.go
```diff
@@ -53,6 +56,9 @@ const (
 	// QueryTypeOpenSessions is the build in query type for Client.QueryWorkflow() call. Use this query type to get all open
 	// sessions in the workflow. The result will be a list of SessionInfo encoded in the EncodedValue.
 	QueryTypeOpenSessions string = "__open_sessions"
+
+	healthCheckServiceName         = "temporal.api.workflowservice.v1.WorkflowService"
+	healthCheckTimeout             = 1 * time.Second
```
mfateev, line 61: I would increase it to something like 5 seconds

mfateev: Or better add the timeout to the connectionoptions

hazcod:
@alexshtin looking back into this, it seems that the healthcheck is package-internal and only ran on Client initialization.
Could that be made public? e.g. Docker healthchecks run every 5 seconds and I want temporal to be included.
I am doing a `DescribeTaskQueue` right now.

PR #215, opened 2020-07-28
Fixes race condition where parent workflow fails to send a cancel command to a recently started child workflow

mastermanu:
This was the bug:

1) Child Workflow start command would be sent, but not yet completed (hence currently in Initiated state).
2) Parent Workflow gets a cancellation request.
3) ChildWorkflow's CancellationCallback.fn is invoked. But because the Child Workflow hasn't started yet (and hence we don't have the Child Workflow ID), childWorkflowExecution is equal to nil.
4) Because childWorkflowExecution is nil, we never send the cancellation request to the child.
5) Parent workflow is stuck "running" because the child is never requested to cancel itself (meaning the child workflow actually has to complete)

This PR fixes the bug by only registering the cancellation after the Workflow has actually started. Other than that, identical behavior is maintained.

Writing an integration test to cover this is difficult since having the child workflow use a bogus task queue doesn't really reproduce the race condition described between 1-3. Added a general integration test to cover cascading cancels anyway.

Bench test reliably reproduces and proves the issue is fixed after the SDK is updated


internal/workflow.go
```diff
@@ -692,25 +690,24 @@ func (wc *workflowEnvironmentInterceptor) ExecuteChildWorkflow(ctx Context, chil
...
+				if ok || !more {
+					cancellationCallback.fn(nil, more)
+				}
+			}
 		}
+
 		executionSettable.Set(r, e)
 	})
 
-	if cancellable {
```
mastermanu, line 700: This was the bug:

1) Workflow start command would be sent, but not yet finished.
2) Workflow would be cancelled
3) CancellationCallback.fn would be invoked. But because the Workflow hasn't started yet, childWorkflowExection was equal to null.
4) As a result, we did not request a cancellation of the child workflow.
5) Parent workflow is stuck "running" because the child is never requested to cancel this.

test/workflow_test.go
```diff
@@ -677,36 +692,39 @@ func (w *Workflows) ActivityCompletionUsingID(ctx workflow.Context) ([]string, e
...
-	worker.RegisterWorkflow(w.ConsistentQueryWorkflow)
-	worker.RegisterWorkflow(w.BasicSession)
+	worker.RegisterWorkflow(w.SimplestWorkflow)
 	worker.RegisterWorkflow(w.WorkflowWithLocalActivityCtxPropagation)
-	worker.RegisterWorkflow(w.ActivityCompletionUsingID)
+
+	worker.RegisterWorkflow(w.child)
+	worker.RegisterWorkflow(w.childForMemoAndSearchAttr)
+	worker.RegisterWorkflow(w.sleep)
+	worker.RegisterWorkflow(w.timer)
```
alexshtin, line 727: :clap:

internal/workflow.go
```diff
@@ -692,25 +690,24 @@ func (wc *workflowEnvironmentInterceptor) ExecuteChildWorkflow(ctx Context, chil
 		}
 	}, func(r WorkflowExecution, e error) {
 		if e == nil {
-			childWorkflowExecution = &r
+			if cancellable {
```
samarabbas, line 693: What happens if initiation takes a long time?  Won't we still be stuck on cancellation and waiting for initiation to go through?

internal/workflow.go
```diff
@@ -692,25 +690,24 @@ func (wc *workflowEnvironmentInterceptor) ExecuteChildWorkflow(ctx Context, chil
...
+				}
+			}
 		}
+
 		executionSettable.Set(r, e)
 	})
 
-	if cancellable {
-		cancellationCallback.fn = func(v interface{}, more bool) bool {
-			if ctx.Err() == ErrCanceled && childWorkflowExecution != nil && !mainFuture.IsReady() {
```
samarabbas, line 702: Will this work if we just remove the `childWorkflowExecution != nil check`?

internal/workflow.go
mastermanu: added a comment explaining why we are waiting on initiation

internal/workflow.go
mastermanu: added a comment explaining why we are waiting on initiation

PR #218, opened 2020-07-28
Rename temporal.go to retry_policy.go

alexshtin:
As it has only `RetryPolicy` type inside.


workflow/workflow.go
```diff
@@ -22,7 +22,6 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// Package workflow contains APIs that workflow implementation code must use
```
alexshtin, line 25: This header is already defined in `doc.go` and it should be enough.

PR #221, opened 2020-07-30
Convert times to time.Time and durations to time.Duration

alexshtin:
Adopt recent [proto changes](https://github.com/temporalio/api/pull/79).

Integration tests are passing with updated server.


internal/internal_task_pollers.go
```diff
@@ -526,11 +526,11 @@ func (lath *localActivityTaskHandler) executeLocalActivityTask(task *localActivi
 		}
 	}()
 
-	timeout := task.params.ScheduleToCloseTimeoutSeconds
-	if task.params.StartToCloseTimeoutSeconds != 0 && task.params.StartToCloseTimeoutSeconds < timeout {
-		timeout = task.params.StartToCloseTimeoutSeconds
+	timeout := task.params.ScheduleToCloseTimeout
+	if task.params.StartToCloseTimeout != 0 && task.params.StartToCloseTimeout < timeout {
```
shawnhathaway, line 530: Do we nill check these?

internal/activity.go
```diff
@@ -247,11 +248,11 @@ func WithActivityTask(
 	tracer opentracing.Tracer,
 ) context.Context {
 	var deadline time.Time
-	scheduled := time.Unix(0, task.GetScheduledTimestamp())
```
shawnhathaway, line 250: How do we handle this change in behavior? If `ScheduledTime` is set to `0` value; previously behavior would make `scheduled` equal `time.Unix(0,0)` but it is now `&time.Time{}`? Is this a problem?

internal/common/time.go
```diff
@@ -25,55 +25,35 @@
...
-}
-
-// MinInt64 returns the smaller of two given int64
-func MinInt64(a, b int64) int64 {
-	if a < b {
-		return a
+// TimeValue dereference time in a safe way.
+func TimeValue(t *time.Time) time.Time {
+	if t == nil {
+		return time.Time{}
```
shawnhathaway, line 34: On server I am treating this as `time.Unix(0, 0)` - https://github.com/temporalio/temporal/blob/master/common/primitives/timestamp/time.go#L192

internal/internal_task_pollers.go
alexshtin: These come from options and they are not nillable.

internal/activity.go
alexshtin: In this particular case it is not a problem at all because these values can't be zero.

internal/common/time.go
alexshtin: Yes, I am changing it there also. I think we need to use `time.Time{}` everywhere as `nil` replacement.

PR #224, opened 2020-08-04
Updates GO SDK to align with fixed Temporal Server Behavior for Child Workflow retries and timeouts

mastermanu:
- Eliminates the setting of "2" as the default backoff co-efficient as the server does this.
- Updates the ChildWorkflowRetryOnTimeout test as retries now get exhausted when the WorkflowExecutionTimeout lapses (which was not occurring prior to the server fix)

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -36,7 +36,6 @@ import (
 	enumspb "go.temporal.io/api/enums/v1"
 
 	"go.temporal.io/sdk/converter"
-	"go.temporal.io/sdk/internal/common/backoff"
 	"go.temporal.io/sdk/log"
 )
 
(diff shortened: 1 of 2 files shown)


internal/workflow.go
```diff
@@ -1457,9 +1456,7 @@ func convertRetryPolicy(retryPolicy *RetryPolicy) *commonpb.RetryPolicy {
 	if retryPolicy == nil {
 		return nil
 	}
-	if retryPolicy.BackoffCoefficient == 0 {
-		retryPolicy.BackoffCoefficient = backoff.DefaultBackoffCoefficient
```
alexshtin, line 1461: `backoff.DefaultBackoffCoefficient` is it still used anywhere?

mastermanu: yes

mastermanu: NewExponentialRetryPolicy

PR #225, opened 2020-08-09
Fix regression in Go SDK where Batch Timer interval for heartbeats is overcalculated

mastermanu:
- As part of switching from variables representing time in seconds to time.Duration, we forgot to correctly update the line of code that calculates the interval for sending the next heartbeat. We were multiplying by an extra factor of one billion, which either caused an overflow (so that we were constantly sending heartbeats) or resulted in a large number (where we would never send a heartbeat after the first one).

Testing:
- Modified unit-test to validate that timer is fired and heartbeat is reported at least a second time.
- Added an integration test for a long-running activity with a smaller heartbeat timeout.

### DIFF
--- a/internal/internal_task_handlers_test.go
+++ b/internal/internal_task_handlers_test.go
@@ -1324,15 +1324,25 @@ func (t *TaskHandlersTestSuite) TestHeartBeat_NoError() {
 	heartbeatErr := temporalInvoker.Heartbeat(nil, false)
 	t.NoError(heartbeatErr)
 
+	select {
+	case <-invocationChannel:
+	case <-time.After(3 * time.Second):
+		t.Fail("did not get expected 1st call to record heartbeat")
+	}
+
 	heartbeatErr = temporalInvoker.Heartbeat(nil, false)
 	t.NoError(heartbeatErr)
 
-	for i := 0; i < 2; i++ {
-		select {
-		case <-invocationChannel:
-		case <-time.After(3 * time.Second):
-			t.Fail("did not get expected calls to record heartbeat")
-		}
+	select {
+	case <-invocationChannel:
+		t.Fail("got unexpected call to record heartbeat. 2nd call should come via batch timer")
+	default:
+	}
+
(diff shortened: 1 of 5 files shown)


test/workflow_test.go
```diff
@@ -137,6 +137,24 @@ func (w *Workflows) ActivityRetryOnTimeout(ctx workflow.Context, timeoutType enu
...
+	opts.ScheduleToCloseTimeout = time.Second * 12
+	opts.StartToCloseTimeout = time.Second * 12
+	opts.RetryPolicy = &internal.RetryPolicy{
+		MaximumAttempts: 1,
+	}
+	ctx = workflow.WithActivityOptions(ctx, opts)
+
+	err := workflow.ExecuteActivity(ctx, "LongRunningHeartbeat", 8*time.Second, 300*time.Millisecond).Get(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("expected activity to succeeded but it failed: %v", err)
```
mastermanu, line 152: fix bad english in error message

internal/internal_task_handlers_test.go
```diff
@@ -1306,19 +1306,34 @@ func (t *TaskHandlersTestSuite) TestLocalActivityRetry_WorkflowTaskHeartbeatFail
...
+		taskToken:        nil,
+		heartBeatTimeout: time.Second,
 	}
 
 	heartbeatErr := temporalInvoker.Heartbeat(nil, false)
+	t.NoError(heartbeatErr)
 
+	heartbeatErr = temporalInvoker.Heartbeat(nil, false)
 	t.NoError(heartbeatErr)
+
```
mastermanu, line 1329: assert that there is no invocation directly after the 2nd heartbeat call returns. this guarantees the 2nd invocation comes via the heartbeat callback

samarabbas:
Thanks @mastermanu for fixing this.  I think this issue is severe enough where we should do a Go SDK patch release.

PR #227, opened 2020-08-11
Rename cancelled to canceled for consistency

alexshtin:
Let's use US form everywhere: https://www.dictionary.com/e/canceled-vs-cancelled/.

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -610,9 +610,9 @@ func (e *TestWorkflowEnvironment) SetOnTimerFiredListener(listener func(timerID 
 	return e
 }
 
-// SetOnTimerCancelledListener sets a listener that will be called after a timer is cancelled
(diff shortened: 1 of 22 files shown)


internal/workflow_testsuite.go
```diff
@@ -610,9 +610,9 @@ func (e *TestWorkflowEnvironment) SetOnTimerFiredListener(listener func(timerID
 	return e
 }
 
-// SetOnTimerCancelledListener sets a listener that will be called after a timer is cancelled
-func (e *TestWorkflowEnvironment) SetOnTimerCancelledListener(listener func(timerID string)) *TestWorkflowEnvironment {
-	e.impl.onTimerCancelledListener = listener
+// SetOnTimerCanceledListener sets a listener that will be called after a timer is canceled
+func (e *TestWorkflowEnvironment) SetOnTimerCanceledListener(listener func(timerID string)) *TestWorkflowEnvironment {
```
alexshtin, line 614: This is sort of breaking change. Can we afford this?

mfateev: Let's do it.

PR #231, opened 2020-08-13
Use a separate counter for local activity

samarabbas:
We cannot use ID generator used for command processing as it is 
only incremented when command is actually generated.  In the case 
of local activity, command is generated when the local activity 
completes.  This results in multiple invocations of LocalActivity in the 
same workflow task processing to potentially use the same id.  This
results in only one local activity to be processed and all others are 
ignored.
Created a separate counter which is used for unique id generation 
for local activities.  There may be more places which needs to be
moved to this counter instead of the id generator used for command
processing.


samarabbas:
PR #229 already has a better fix.

PR #235, opened 2020-08-19
Add integration test for SignalWorkflow

alexshtin:
Adding integration tests to check if `SignalWorkflow` works for primitive types and proto objects.
https://community.temporal.io/t/sending-proto-messages-directly-over-signal-channels/441/3

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -797,7 +797,7 @@ func (c *channelImpl) assignValue(from interface{}, to interface{}) error {
 	err := decodeAndAssignValue(c.dataConverter, from, to)
 	// add to metrics
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/internal_workflow.go
```diff
@@ -1239,7 +1239,7 @@ func (w *WorkflowOptions) getSignalChannel(ctx Context, signalName string) Recei
 	if ch, ok := w.signalChannels[signalName]; ok {
 		return ch
 	}
-	ch := NewBufferedChannel(ctx, defaultSignalChannelSize)
+	ch := NewNamedBufferedChannel(ctx, signalName, defaultSignalChannelSize)
```
alexshtin, line 1242: I hope there are no side effects for this change. `name` is used for error message and it would be great to see something meaningful there rather then empty string.

PR #240, opened 2020-08-29
Fixes flaky integration-tests

mastermanu:
1) Address flakiness in ActivityHBTimeout Test:
The test was validating that the full 5 seconds of the schedule-to-close time had elapsed before the activity failed.
This is actually incorrect as we will abort the retry early if the current time + current_retry_interval exceeds the schedule-to-close deadline. As a result, this could cause flakiness as the activity would fail after 4+ seconds (not bother doing the last retry since that would exceed the overall deadline). Fixes this to change the threshold to remove the maximum interval

2) Addresses flakiness in overall integration test:
The overall integration test timeout was 2 minutes. We've added new tests since that time and now we are running up against the limit on certain environments. Increasing the timeout to 3 minutes to account for the new tests

### DIFF
--- a/test/workflow_test.go
+++ b/test/workflow_test.go
@@ -170,7 +170,7 @@ func (w *Workflows) ActivityRetryOnHBTimeout(ctx workflow.Context) ([]string, er
 	}
 
 	elapsed := workflow.Now(ctx).Sub(startTime)
-	if elapsed < opts.StartToCloseTimeout-opts.RetryPolicy.MaximumInterval {
+	if elapsed < opts.ScheduleToCloseTimeout-opts.RetryPolicy.MaximumInterval {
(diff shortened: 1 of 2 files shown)


test/workflow_test.go
```diff
@@ -170,7 +170,7 @@ func (w *Workflows) ActivityRetryOnHBTimeout(ctx workflow.Context) ([]string, er
 	}
 
 	elapsed := workflow.Now(ctx).Sub(startTime)
-	if elapsed < 5*time.Second {
+	if elapsed < opts.ScheduleToCloseTimeout-opts.RetryPolicy.MaximumInterval {
```
alexshtin, line 173: I would add description of this PR (may be with some mods) as comment here.

mastermanu: done

PR #249, opened 2020-09-17
grpc.WithConnectParams was added without set MinConnectTimeout 

linvon:
gRPC backoff was added to file `grpc_dialer.go` at version v0.30.0. Func `grpc.WithConnectParams` will set `option.minConnectTimeout` to the func with return value `ConnectParams.MinConnectTimeout`, which has default value `minConnectTimeout = 20 * time.Second`.
But now we are not passing `MinConnectTimeout` to the parameter used by `grpc.WithConnectParams`, so the **func is set** and **use zero value**.

In that way, `grpc.Dial()` is easy to fail. `ConnectParams.MinConnectTimeout` should be set to the old 20 seconds default.


alexshtin:
@linvon, thanks for reporting this. I slightly fix your description and created another PR to fix this.

PR #256, opened 2020-09-25
Make sure metric with same name is emitted with same tags

samarabbas:


internal/common/metrics/constants.go
```diff
@@ -74,3 +74,23 @@ const (
 
 	WorkflowActiveThreadCount = TemporalMetricsPrefix + "workflow_active_thread_count"
 )
+
+// Metric tag keys
+const (
+	NamespaceTagName        = "namespace"
+	ClientTagName           = "temporal_client_name"
```
alexshtin, line 81:
```suggestion
	ClientTagName           = "client_name"
```
I renamed header name also.

internal/common/metrics/interceptor_test.go
```diff
@@ -119,16 +119,16 @@ func TestMetricsInterceptor(t *testing.T) {
...
 	}
 
 }
 
-func assertMetrics(assert *assert.Assertions, reporter *CapturingStatsReporter, methodName, separator string, counterNames []string) {
+func assertMetrics(assert *assert.Assertions, reporter *CapturingStatsReporter, counterNames []string) {
 	assert.Equal(len(counterNames), len(reporter.counts))
 	for _, name := range counterNames {
-		counterName := TemporalMetricsPrefix + methodName + separator + name
+		counterName := name
```
alexshtin, line 131: I think you don't need this line at all now.

internal/internal_worker.go
```diff
@@ -230,10 +230,10 @@ func ensureRequiredParams(params *workerExecutionParameters) {
 // verifyNamespaceExist does a DescribeNamespace operation on the specified namespace with backoff/retry
 // It returns an error, if the server returns an EntityNotExist or BadRequest error
 // On any other transient error, this method will just return success
-func verifyNamespaceExist(client workflowservice.WorkflowServiceClient, namespace string, logger log.Logger) error {
+func verifyNamespaceExist(client workflowservice.WorkflowServiceClient, metricsScope tally.Scope, namespace string, logger log.Logger) error {
 	ctx := context.Background()
 	descNamespaceOp := func() error {
-		grpcCtx, cancel := newGRPCContext(ctx)
+		grpcCtx, cancel := newGRPCContext(ctx, grpcMetricsScope(metrics.GetEmptyRPCScope(metricsScope)))
```
alexshtin, line 236: Instead of passing `metrics.GetEmptyRPCScope(metricsScope)` every time you can just set it in `intenal_utils.go` line 130 as:
```
MetricsScope: metrics.GetEmptyRPCScope(metricsScope),
```
and it will be overridden every time you pass specific metrics scope.

samarabbas: The problem is GRPCContext is not aware of rootMetricsScope.

alexshtin: You're right. The only thing I don't like here is that the option which is passed every time is not an option but required parameter. Let's keep it as is and I might refactor it later.

alexshtin (approved):

PR #263, opened 2020-10-06
Fix custom tracer header propagation

alexshtin:
This PR fixes bug: when custom context propagators are set, custom tracer stops to work (stops to propagate span headers).
First, I added an integration test for two context propagators to make sure they work together. This test passed fine and didn't discover the problem. Then, after some debugging, I figured out that there was an error during deserialization of tracer headers from headers payload. The proper fix is to store entire tracing data as single header and let opentracing to deal with it content.

Note: because header key is changing, workflows started with previous SDK versions will not propagate trace to this version and vice versa.

Tested manually with local Jaeger installation.

### DIFF
--- a/internal/headers.go
+++ b/internal/headers.go
@@ -26,6 +26,8 @@ package internal
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
 	commonpb "go.temporal.io/api/common/v1"
 )
@@ -37,6 +39,7 @@ type HeaderWriter interface {
 
 // HeaderReader is an interface to read information from temporal headers
 type HeaderReader interface {
+	Get(string) (*commonpb.Payload, error)
 	ForEachKey(handler func(string, *commonpb.Payload) error) error
 }
 
@@ -72,6 +75,17 @@ func (hr *headerReader) ForEachKey(handler func(string, *commonpb.Payload) error
 		}
 	}
 	return nil
+}
+
+func (hr *headerReader) Get(key string) (*commonpb.Payload, error) {
+	if hr.header == nil {
+		return nil, errors.New("header is nil")
+	}
+	if value, ok := hr.header.Fields[key]; ok {
+		return value, nil
+	}
+
+	return nil, fmt.Errorf("key %s doesn't exist in header fields", key)
 }
 
 // NewHeaderReader returns a header reader interface
--- a/internal/headers.go
+++ b/internal/headers.go
@@ -85,12 +85,12 @@ func (hr *headerReader) Get(key string) (*commonpb.Payload, error) {
 		return value, nil
 	}
 
-	return nil, fmt.Errorf("key %s doesn't exist in header fields", key)
+	return nil, fmt.Errorf("key %q doesn't exist in header fields", key)
 }
 
 // NewHeaderReader returns a header reader interface
 func NewHeaderReader(header *commonpb.Header) HeaderReader {
-	return &headerReader{header}
+	return &headerReader{header: header}
 }
 
 type headerWriter struct {
@@ -109,5 +109,5 @@ func NewHeaderWriter(header *commonpb.Header) HeaderWriter {
 	if header != nil && header.Fields == nil {
 		header.Fields = make(map[string]*commonpb.Payload)
 	}
-	return &headerWriter{header}
+	return &headerWriter{header: header}
 }
--- a/internal/headers.go
+++ b/internal/headers.go
@@ -26,8 +26,6 @@ package internal
 
 import (
 	"context"
-	"errors"
-	"fmt"
 
 	commonpb "go.temporal.io/api/common/v1"
 )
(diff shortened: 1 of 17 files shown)


internal/tracer.go
```diff
@@ -119,17 +128,56 @@ func (t *tracingContextPropagator) InjectFromWorkflow(
...
 ) (Context, error) {
-	spanContext, err := t.tracer.Extract(opentracing.TextMap, tracingReader{hr})
+	spanContext, err := t.readSpanContextFromHeader(hr)
 	if err != nil {
-		// did not find a tracing span, just return the current context
 		return ctx, nil
 	}
 	return contextWithSpan(ctx, spanContext), nil
 }
+
```
alexshtin, line 145: These two methods below are core of this fix.

internal/tracer.go
```diff
@@ -119,17 +128,56 @@ func (t *tracingContextPropagator) InjectFromWorkflow(
...
+	tracerPayload, err := converter.GetDefaultDataConverter().ToPayload(tracerData)
+	if err != nil {
+		return err
+	}
+
+	hw.Set(tracerHeaderKey, tracerPayload)
+	return nil
+}
+
+func (t *tracingContextPropagator) readSpanContextFromHeader(hr HeaderReader) (opentracing.SpanContext, error) {
```
mfateev, line 163: This method should take tracerData map as a parameter and deserialization should happen in the caller

mfateev (changes requested):

internal/tracer.go
```diff
@@ -26,42 +26,50 @@ package internal
...
 
 	"github.com/opentracing/opentracing-go"
-	commonpb "go.temporal.io/api/common/v1"
 
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/log"
 )
 
+const (
+	tracerHeaderKey = "_tracer-data"
```
alexshtin, line 38: Is it a good name for header?

internal/internal_workflow_client_test.go
```diff
@@ -118,36 +119,34 @@ func (s *stringMapPropagator) InjectFromWorkflow(ctx Context, writer HeaderWrite
...
+			return ctx, err
 		}
-		return nil
-	}); err != nil {
-		return nil, err
+		var decodedValue string
+		err = converter.GetDefaultDataConverter().FromPayload(value, &decodedValue)
+		if err != nil {
+			return ctx, err
+		}
```
mfateev, line 148: Move duplicated code into a separate method?

internal/headers.go
```diff
@@ -74,6 +77,17 @@ func (hr *headerReader) ForEachKey(handler func(string, *commonpb.Payload) error
 	return nil
 }
 
+func (hr *headerReader) Get(key string) (*commonpb.Payload, error) {
+	if hr.header == nil {
+		return nil, errors.New("header is nil")
```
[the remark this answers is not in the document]
alexshtin: Agree!

internal/headers.go
```diff
@@ -37,6 +39,7 @@ type HeaderWriter interface {
 
 // HeaderReader is an interface to read information from temporal headers
 type HeaderReader interface {
+	Get(string) (*commonpb.Payload, error)
```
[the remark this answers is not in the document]
alexshtin: Yes, let's mock `map` here.

internal/internal_workflow_client_test.go
alexshtin: It is test code (I want it to be clear and simple) and there is a loop, and with all this error handling, extracting doesn't make it better.

internal/internal_workflow_client_test.go
```diff
@@ -1029,37 +1025,6 @@ func (s *workflowClientTestSuite) TestStartWorkflow() {
 	s.Equal(createResponse.GetRunId(), resp.RunID)
 }
 
-func (s *workflowClientTestSuite) TestStartWorkflow_WithContext() {
```
alexshtin, line 1032: This was useless test. I replaced it with real integration test.

test/test_utils_test.go
```diff
@@ -103,24 +101,21 @@ func WaitForTCP(timeout time.Duration, addr string) error {
 
 // stringMapPropagator propagates the list of keys across a workflow,
 // interpreting the payloads as strings.
-// BORROWED FROM 'internal' PACKAGE TESTS.
+// TODO: BORROWED FROM 'internal' PACKAGE TESTS.
+// TODO: remove code duplication.
```
mfateev, line 105: stringMapPropagator is a weird name after the refactoring. May be multiKeyPropagator?

mfateev (approved):

test/test_utils_test.go
alexshtin: I so get used to it and didn't pay attention.

PR #264, opened 2020-10-08
Improve error messages

alexshtin:
Remove duplicate information from failure messages and add more context to go error messages.

### DIFF
--- a/internal/error.go
+++ b/internal/error.go
@@ -291,7 +291,7 @@ func NewTimeoutError(msg string, timeoutType enumspb.TimeoutType, cause error, l
 
 // NewHeartbeatTimeoutError creates TimeoutError instance.
 func NewHeartbeatTimeoutError(details ...interface{}) *TimeoutError {
-	return NewTimeoutError("Heartbeat timeout", enumspb.TIMEOUT_TYPE_HEARTBEAT, nil, details...)
+	return NewTimeoutError("heartbeat timeout", enumspb.TIMEOUT_TYPE_HEARTBEAT, nil, details...)
 }
 
 // NewCanceledError creates CanceledError instance.
--- a/internal/error.go
+++ b/internal/error.go
@@ -121,7 +121,7 @@ type (
 	// ApplicationError returned from activity implementations with message and optional details.
 	ApplicationError struct {
 		temporalError
-		message      string
+		msg          string
 		errType      string
 		nonRetryable bool
 		cause        error
@@ -131,6 +131,7 @@ type (
 	// TimeoutError returned when activity or child workflow timed out.
 	TimeoutError struct {
 		temporalError
+		msg                  string
 		timeoutType          enumspb.TimeoutType
 		lastHeartbeatDetails converter.EncodedValues
 		cause                error
@@ -174,7 +175,7 @@ type (
 	// ServerError can be returned from server.
 	ServerError struct {
 		temporalError
-		message      string
+		msg          string
 		nonRetryable bool
 		cause        error
 	}
@@ -216,6 +217,7 @@ type (
 	}
 
 	temporalError struct {
+		messenger
 		originalFailure *failurepb.Failure
 	}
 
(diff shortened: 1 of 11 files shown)


internal/error.go
```diff
@@ -406,9 +420,20 @@ func NewContinueAsNewError(ctx Context, wfn interface{}, args ...interface{}) *C
...
+// Error from error interface.
 func (e *ApplicationError) Error() string {
-	return e.message
+	msg := e.message()
+	if e.errType != "" {
+		msg = fmt.Sprintf("%s (type: %s, retryable: %v)", msg, e.errType, !e.nonRetryable)
+	}
+	if e.cause != nil {
+		msg = fmt.Sprintf("%s: %v", msg, e.cause)
+	}
```
alexshtin, line 431: This will make error message looks like native error message in case if original error was  standart go error (i.e. `errors.New("message")`).

internal/error.go
```diff
@@ -216,16 +217,24 @@ type (
...
 	}
 
 	failureHolder interface {
 		setFailure(*failurepb.Failure)
 		failure() *failurepb.Failure
 	}
+
+	messenger interface {
+		message() string
+	}
```
wxing1292, line 231: this is the breaking change we talked about few days ago?

internal/error.go
```diff
@@ -241,9 +250,9 @@ var (
 )
 
 // NewApplicationError create new instance of *ApplicationError with message, type, and optional details.
-func NewApplicationError(message, errType string, nonRetryable bool, cause error, details ...interface{}) *ApplicationError {
+func NewApplicationError(msg, errType string, nonRetryable bool, cause error, details ...interface{}) *ApplicationError {
```
wxing1292, line 253: can we not reuse the var type? 
like `msg, errType string` vs `msg string, errType string`
former is hard to parse?

internal/error.go
```diff
@@ -769,7 +855,8 @@ func convertFailureToError(failure *failurepb.Failure, dc converter.DataConverte
 	} else if failure.GetTimeoutFailureInfo() != nil {
 		timeoutFailureInfo := failure.GetTimeoutFailureInfo()
 		lastHeartbeatDetails := newEncodedValues(timeoutFailureInfo.GetLastHeartbeatDetails(), dc)
-		err = NewTimeoutError(
+		err = newTimeoutError(
+			failure.GetMessage(),
```
alexshtin, line 859: `TimeoutFailure` may have message set on server (`WorkflowTask timeout`/`ActivityTask timeout`). This will propagate it to error.

internal/error.go
alexshtin: It was today morning :-) And actually not. This is just internal interface to simplify dealing with error messages.

wxing1292 (approved):

internal/internal_event_handlers.go
```diff
@@ -1228,10 +1227,22 @@ func (weh *workflowExecutionEventHandlerImpl) handleStartChildWorkflowExecutionF
...
+			0,
+			enumspb.RETRY_STATE_NON_RETRYABLE_FAILURE,
+			errors.New("workflow execution already started"),
+		)
+		childWorkflow.handle(nil, err)
+		return nil
+	}
 
-	return nil
+	return fmt.Errorf("unknown cause: %v", attributes.GetCause())
```
alexshtin, line 1245: I am not sure about this change.

internal/error.go
```diff
@@ -645,6 +718,10 @@ func getErrType(err error) string {
 	for t = reflect.TypeOf(err); t.Kind() == reflect.Ptr; t = t.Elem() {
 	}
 
+	if t.Name() == goErrType {
+		return ""
+	}
```
alexshtin, line 723: If it was standard error created with `errors.New` error type will be empty string. I think this is better for `Type` field.

mfateev (approved):

PR #273, opened 2020-10-21
Update broken link to Go quickstart

robholland:
There isn't an obvious replacement for the broken link so pointed it at the top-level page for the SDK docs.


alexshtin (approved):

alexshtin:
Thanks for fixing it.

PR #281, opened 2020-11-06
Relocate workflow and activity registration functions to separate interfaces

### LINKED ISSUE #280: Add a Registry/WorkerRegistry interface to the worker package
When registering workflows/activities in a worker, consumers often don't need the Start and Stop methods of the worker interface.

Define a `Registry` or a `WorkerRegistry` interface in the worker package with the registration methods and embed it in the worker interface. Consumers can use this interface instead of the `Worker` interface. Although in Go consumers usually own interfaces, in this case the interface and the use case is well-defined by the provider, so it makes sense to define it in the worker package.

sagikazarmark:
Fixes #280


mfateev (approved):

sagikazarmark:
Thanks!

PR #286, opened 2020-11-10
fix PanicError when future return value is pointer

ti:
This request fix the Panic Error bug:

How to trigger the bug:

When the activity result is a pointer and settable set a pointer, the code will be Panic.

```go
// Workflow is a Hello World workflow definition.
func Workflow(ctx workflow.Context, name string) (*Result, error) {
		ScheduleToStartTimeout: time.Minute,
		StartToCloseTimeout:    time.Minute,
	}
	future, settable := workflow.NewFuture(ctx)
	workflow.Go(ctx, func(ctx workflow.Context) {
		result := new(Result)
		err := workflow.ExecuteActivity(ctx, Activity, name).Get(ctx, result)
		if err != nil {
			settable.Set(nil, err)
			return
		}
		settable.Set(result, err)
	})
	var result Result
	err := future.Get(ctx, &result)
	if err != nil {
		return nil, err
	}
	return &result, nil
}
// Result the result object
type Result struct {
	Data string
}

// Activity renturn a pointer
func Activity(ctx context.Context, name string) (*Result, error) {
	return &Result{Data:"Hello " + name + "!"}, nil
}
```

> more info see the code changes

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -333,7 +333,12 @@ func (f *futureImpl) Get(ctx Context, valuePtr interface{}) error {
 
 	fv := reflect.ValueOf(f.value)
 	if fv.IsValid() {
-		rf.Elem().Set(fv)
+		// fix the PanicError when the return value is pointer
+		if fv.Kind() == reflect.Ptr {
+			rf.Elem().Set(fv.Elem())
+		} else {
+			rf.Elem().Set(fv)
+		}
 	}
 	return f.err
 }


ti (commented):
fix a simple bug

alexshtin:
This is expected behavior. `Get` should get a pointer to what it should set. In your case it is pointer to pointer:
```go
	var result *Result
	err := future.Get(ctx, &result)
	if err != nil {
		return nil, err
	}
	return result, nil
```

As you can see, this fix broke some tests also.

ti:
I understand， But if I want all apis like grpc method, 
Activity future will not panic, but the settable future will be panic.
See the code below： Will this cause cognitive conflicts?  @alexshtin 

  
```go
 future, settable := workflow.NewFuture(ctx)
	workflow.Go(ctx, func(ctx workflow.Context) {
		var result Result
                // 1: This Code Not Panic
		err := workflow.ExecuteActivity(ctx, Activity, name).Get(ctx, &result)
		if err != nil {
			settable.Set(nil, err)
			return
		}
		settable.Set(&result, err)
	})
	var result Result
	// 2. This Code Will be Be Panic in master for https://github.com/temporalio/sdk-go/pull/286/files
	err := future.Get(ctx, &result)
	if err != nil {
		return nil, err
	}
```

mfateev:
I agree with @ti that we are not consistent by not panicking on activity and panicking on the future created through NewFuture.

Issue #291, opened 2020-11-17
Expose previous run failure though workflow context API


samarabbas:
**Is your feature request related to a problem? Please describe.**
Current we expose getLastCompletionResult api used for communicating the result of
previous workflow run to next run. But we don't have similar api which exposes last
failure.

**Describe the solution you'd like**
Expose lastFailure api which exposes failure of previous run but we still did a
continueAsNew because of cron or retry policy. LastCompletionResult is still carried over
from earlier run even when previous run fails due to an error.

**Additional context**
https://community.temporal.io/t/bug-on-getlastcompletionresult-returns-null-if-previous-run-timeout-or-fail/997

PR #293, opened 2020-11-19
Implement GetLastFailure API

Sushisource:
Resolves https://github.com/temporalio/sdk-go/issues/291

Added unit & integ test

### DIFF
--- a/internal/interceptors.go
+++ b/internal/interceptors.go
@@ -29,8 +29,6 @@ import (
 
 	"github.com/uber-go/tally"
 
-	failurepb "go.temporal.io/api/failure/v1"
-
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/log"
 )
@@ -88,7 +86,7 @@ type WorkflowOutboundCallsInterceptor interface {
 	IsReplaying(ctx Context) bool
 	HasLastCompletionResult(ctx Context) bool
 	GetLastCompletionResult(ctx Context, d ...interface{}) error
-	GetLastFailure(ctx Context) *failurepb.Failure
+	GetLastFailure(ctx Context) error
 }
 
 var _ WorkflowOutboundCallsInterceptor = (*WorkflowOutboundCallsInterceptorBase)(nil)
@@ -222,6 +220,6 @@ func (t *WorkflowOutboundCallsInterceptorBase) GetLastCompletionResult(ctx Conte
 }
 
 // GetLastFailure forwards to t.Next
-func (t *WorkflowOutboundCallsInterceptorBase) GetLastFailure(ctx Context) *failurepb.Failure {
+func (t *WorkflowOutboundCallsInterceptorBase) GetLastFailure(ctx Context) error {
 	return t.Next.GetLastFailure(ctx)
 }
--- a/internal/interceptors.go
+++ b/internal/interceptors.go
@@ -25,6 +25,7 @@
 package internal
 
 import (
+	"go.temporal.io/api/failure/v1"
 	"time"
 
 	"github.com/uber-go/tally"
@@ -86,6 +87,8 @@ type WorkflowOutboundCallsInterceptor interface {
 	IsReplaying(ctx Context) bool
 	HasLastCompletionResult(ctx Context) bool
 	GetLastCompletionResult(ctx Context, d ...interface{}) error
+	HasLastFailure(ctx Context) bool
+	GetLastFailure(ctx Context) *failure.Failure
 }
 
 var _ WorkflowOutboundCallsInterceptor = (*WorkflowOutboundCallsInterceptorBase)(nil)
@@ -216,4 +219,14 @@ func (t *WorkflowOutboundCallsInterceptorBase) HasLastCompletionResult(ctx Conte
 // GetLastCompletionResult forwards to t.Next
 func (t *WorkflowOutboundCallsInterceptorBase) GetLastCompletionResult(ctx Context, d ...interface{}) error {
 	return t.Next.GetLastCompletionResult(ctx, d...)
+}
+
+// HasLastFailure forwards to t.Next
+func (t *WorkflowOutboundCallsInterceptorBase) HasLastFailure(ctx Context) bool {
+	return t.Next.HasLastFailure(ctx)
+}
+
+// GetLastFailure forwards to t.Next
+func (t *WorkflowOutboundCallsInterceptorBase) GetLastFailure(ctx Context) *failure.Failure {
+	return t.Next.GetLastFailure(ctx)
 }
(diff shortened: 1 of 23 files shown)


internal/interceptors.go
```diff
@@ -86,6 +87,8 @@ type WorkflowOutboundCallsInterceptor interface {
 	IsReplaying(ctx Context) bool
 	HasLastCompletionResult(ctx Context) bool
 	GetLastCompletionResult(ctx Context, d ...interface{}) error
+	HasLastFailure(ctx Context) bool
```
mfateev, line 90: I don't think HasLastFailure is needed. We can just return nil from GetLastFailure. It didn't work for the completion result as it required deserialization and didn't return anything.

internal/workflow_testsuite.go
```diff
@@ -32,6 +32,8 @@ import (
 	"testing"
 	"time"
 
+	"go.temporal.io/api/failure/v1"
```
alexshtin, line 35: There is our internal convention use `pb` suffix for all api(=proto) imports. So this should be `failurepb`. You have it in other places. Please fix it also here too and everywhere else (if any).

alexshtin (changes requested):

test/integration_test.go
```diff
@@ -748,13 +762,19 @@ func (ts *IntegrationTestSuite) executeWorkflowWithContextAndOption(
...
-	return client.StartWorkflowOptions{
+	var wfOptions = client.StartWorkflowOptions{
 		ID:                       wfID,
 		TaskQueue:                ts.taskQueueName,
 		WorkflowExecutionTimeout: 15 * time.Second,
 		WorkflowTaskTimeout:      time.Second,
 		WorkflowIDReusePolicy:    enumspb.WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE,
 	}
+	// REVIEW: Putting this here feels a bit lame, but plumbing arguments down through `executeWorkflow`
+	//  feels like a lot of change that almost all the tests don't need. Thoughts?
```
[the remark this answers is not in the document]
Sushisource: Yeah, I always remove these after reviews - thanks

internal/workflow_testsuite.go
Sushisource: Will do

test/integration_test.go
```diff
@@ -684,6 +684,20 @@ func (ts *IntegrationTestSuite) TestContextPropagator() {
 	}, propagatedValues)
 }
 
+const CronWorkflowID = "test-cron"
+
+func (ts *IntegrationTestSuite) TestCron() {
```
alexshtin, line 689: This sounds like we test cron scheduler here but we don't, right? `TestFailurePropogation` probably a better name here. Workflow itself might have a better name.

internal/internal_task_handlers.go
```diff
@@ -633,6 +633,7 @@ func (wth *workflowTaskHandlerImpl) createWorkflowContext(task *workflowservice.
 		Namespace:                wth.namespace,
 		Attempt:                  attributes.GetAttempt(),
 		lastCompletionResult:     attributes.LastCompletionResult,
+		lastFailure:              attributes.ContinuedFailure,
```
alexshtin, line 636: I have one conceptual concern here. We don't expose `Failure` from SDK now. We expose go `error`s. Shouldn't you follow same pattern and convert `Failure` to `error`. There is func for this already. @mfateev, what do you think about it?

test/integration_test.go
Sushisource: Indeed

internal/internal_task_handlers.go
mfateev: Yes, I missed it. Let's convert it to the error, failure is internal to the framework.

Sushisource: Fixed. Thanks guys!

internal/interceptors.go
```diff
@@ -86,6 +86,7 @@ type WorkflowOutboundCallsInterceptor interface {
 	IsReplaying(ctx Context) bool
 	HasLastCompletionResult(ctx Context) bool
 	GetLastCompletionResult(ctx Context, d ...interface{}) error
+	GetLastFailure(ctx Context) error
```
Sushisource, line 89: @mfateev Wanted to get your thoughts on the name of this method. Was talking with @alexshtin and he made a good point that this isn't really consistent with the rest of the SDK - should probably be `GetLastError`. However, that makes Java/Go somewhat inconsistent with each other.

Should we prefer internal consistency, or a consistency of terms? Perhaps we stay consistent internally for now, and try to be more cross-language consistent in a V2?

mfateev: I think GetLastError fits Go SDK better.

internal/internal_task_handlers.go
```diff
@@ -633,6 +633,7 @@ func (wth *workflowTaskHandlerImpl) createWorkflowContext(task *workflowservice.
 		Namespace:                wth.namespace,
 		Attempt:                  attributes.GetAttempt(),
 		lastCompletionResult:     attributes.LastCompletionResult,
+		lastFailure:              convertFailureToError(attributes.ContinuedFailure, wth.dataConverter),
```
mfateev, line 636: Let's keep Failure at this level. DataConverter doesn't really belong here.

test/integration_test.go
```diff
@@ -684,6 +684,20 @@ func (ts *IntegrationTestSuite) TestContextPropagator() {
 	}, propagatedValues)
 }
 
+const CronWorkflowID = "test-cron"
+
+func (ts *IntegrationTestSuite) TestFailurePropagation() {
+	var expected int
+	err := ts.executeWorkflow(CronWorkflowID, ts.workflows.CronWorkflow, &expected)
+	// Workflow finishes itself by cancelling
```
mfateev, line 692: It is kind of a mess the way it is implemented. But the whole idea of cancellation is that it comes from outside. The workflow finishing itself by canceling is a misnomer.

Sushisource: Sure, makes sense. Is there a better way I should do this, or just clarifying the comment?

internal/interceptors.go
Sushisource: Sounds good. Any thoughts on the overall point?

internal/internal_task_handlers.go
Sushisource: Fixed

internal/internal_workflow_testsuite.go
```diff
@@ -2232,6 +2232,10 @@ func (env *testWorkflowEnvironmentImpl) setLastCompletionResult(result interface
 	env.workflowInfo.lastCompletionResult = data
 }
 
+func (env *testWorkflowEnvironmentImpl) setLastError(failure error) {
```
alexshtin, line 2235:
```suggestion
func (env *testWorkflowEnvironmentImpl) setLastError(err error) {
```
`err` is a common name for variables of `error` type.

Sushisource: Thx, fixed.

alexshtin (approved):

PR #294, opened 2020-11-20
Fix GetWorkflow to provide latest run id if none was provided to it

Sushisource:
Resolves https://github.com/temporalio/sdk-go/issues/272

I'm not totally sure this is the way we want to go about this, though, so would very much appreciate feedback on the approach. I left some detail in a comment.

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -47,6 +47,7 @@ import (
 	"go.temporal.io/sdk/internal/common/backoff"
 	"go.temporal.io/sdk/internal/common/metrics"
 	"go.temporal.io/sdk/internal/common/serializer"
+	"go.temporal.io/sdk/internal/common/util"
 	"go.temporal.io/sdk/log"
 )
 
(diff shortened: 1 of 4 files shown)


internal/internal_workflow_client.go
```diff
@@ -293,12 +293,24 @@ func (wc *WorkflowClient) ExecuteWorkflow(ctx context.Context, options StartWork
...
 
 	iterFn := func(fnCtx context.Context, fnRunID string) HistoryEventIterator {
 		return wc.GetWorkflowHistory(fnCtx, workflowID, fnRunID, true, enumspb.HISTORY_EVENT_FILTER_TYPE_CLOSE_EVENT)
 	}
 
+	// The ID may not actually have been set - if not, we have to ask the server for info about the workflow execution
+	// and extract run id from there. This is definitely less efficient than it could be if there was a more specific
+	// rpc method for this, or if there were more granular history filters - in which case it could be extracted from
+	// the `iterFn` inside of `workflowRunImpl`
+	if runID == "" {
```
mfateev, line 306: I would move this code into workflowRunImpl.GetRunID to save the roundtrip for cases when runID is not being accessed.

Sushisource: Makes sense, initially I had tried doing this with the history iterator, but found the events wouldn't contain what I needed. I'll plumb in a way to access the client from inside `GetRunID` -- or perhaps make runID be lazily loaded.

Sushisource: Made this lazy-load.

mfateev (approved):

Issue #295, opened 2020-11-22
Data race on cache eviction


mfateev:
## Expected Behavior
No data races

## Actual Behavior

```
==================
WARNING: DATA RACE
Write at 0x00c000120629 by goroutine 124:
  go.temporal.io/sdk/internal.(*coroutineState).close()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_workflow.go:868 +0x5a
  go.temporal.io/sdk/internal.(*dispatcherImpl).NewCoroutine.func1()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_workflow.go:909 +0x148

Previous read at 0x00c000120629 by goroutine 61:
  go.temporal.io/sdk/internal.(*dispatcherImpl).Close()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_workflow.go:991 +0x174
  go.temporal.io/sdk/internal.(*syncWorkflowDefinition).Close()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_workflow.go:548 +0xb7
  go.temporal.io/sdk/internal.(*workflowExecutionEventHandlerImpl).Close()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_event_handlers.go:941 +0xb7
  go.temporal.io/sdk/internal.(*workflowExecutionContextImpl).clearState()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_task_handlers.go:570 +0x20b
  go.temporal.io/sdk/internal.(*workflowExecutionContextImpl).onEviction()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_task_handlers.go:534 +0xb8
  go.temporal.io/sdk/internal.getWorkflowCache.func1.1()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_task_handlers.go:433 +0x61

Goroutine 124 (running) created at:
  go.temporal.io/sdk/internal.(*dispatcherImpl).NewCoroutine()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_workflow.go:899 +0x1d8
  go.temporal.io/sdk/internal.(*workflowEnvironmentInterceptor).Go()
      /Users/maxim/temporal/temporal-go-sdk/internal/internal_workflow.go:298 +0xa9
  go.temporal.io/sdk/test_test.(*tracingOutboundCallsInterceptor).Go()
[...]

PR #296, opened 2020-11-23
Opened missing types from WorkflowEnvironment

mfateev:
These types need to be open for PHP integration through internalbindings.

Also fixes https://github.com/temporalio/sdk-go/issues/295


internal/internal_activity.go
```diff
@@ -51,13 +51,12 @@ type (
 
 	// ActivityID uniquely identifies an activity execution
 	ActivityID struct {
-		scheduleID int64
-		activityID string
+		id string
```
alexshtin, line 54: It seems like simple
```go
type ActivityID string
```
would be enough here. But current version is extensible. Let's leave it.

alexshtin (approved):

internal/internal_activity.go
mfateev: I considered this, but decided to stick with the struct for now.

PR #297, opened 2020-11-25
Fixes issue where a deadlock detected via a Workflow Execution resumed from a Local Activity Completion fails the Workflow

mastermanu:
Fixes regression introduced by Deadlock Detection PR:

1) A workflow task execution that exceeded 1 second causes a panic.
2) If the task execution was spawned by the completion of a local activity, that would always fail the Workflow.

This fix alters the code to respect the WorkflowPanicPolicy, even in the case where the panic occurs during Local Activity Result completion. It also adds an integration test to exercise this path.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -936,6 +936,14 @@ ProcessEvents:
 			panicError = err
 		}
 	}
+
+	return w.applyPanicPolicy(workflowTask, panicError)
+}
+
+func (w *workflowExecutionContextImpl) applyPanicPolicy(workflowTask *workflowTask, panicError error) (interface{}, error) {
+	task := workflowTask.task
+	eventHandler := w.getEventHandler()
+
 	if panicError == nil && w.err != nil {
 		if panicErr, ok := w.err.(*workflowPanicError); ok {
 			panicError = panicErr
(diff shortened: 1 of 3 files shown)


internal/internal_task_handlers.go
```diff
@@ -936,6 +936,14 @@ ProcessEvents:
 			panicError = err
 		}
 	}
+
+	return w.applyPanicPolicy(workflowTask, panicError)
+}
+
+func (w *workflowExecutionContextImpl) applyPanicPolicy(workflowTask *workflowTask, panicError error) (interface{}, error) {
+	task := workflowTask.task
+	eventHandler := w.getEventHandler()
```
alexshtin, line 946: Move this to line 972.

internal/internal_task_handlers.go
```diff
@@ -936,6 +936,14 @@ ProcessEvents:
 			panicError = err
 		}
 	}
+
+	return w.applyPanicPolicy(workflowTask, panicError)
+}
+
+func (w *workflowExecutionContextImpl) applyPanicPolicy(workflowTask *workflowTask, panicError error) (interface{}, error) {
```
alexshtin, line 944: `panicError` is more like `workflowErr`. No?

alexshtin (approved):

PR #300, opened 2020-11-27
Replaces time.After() usage with time.NewTimer() to reduce memory pressure due to allocated timers

mastermanu:
See this article for an example on when time.After may not be optimal for memory usage purposes: https://medium.com/@oboturov/golang-time-after-is-not-garbage-collected-4cbc94740082

For workflows with a lot of local activities, we will end up unnecessary allocating a whole bunch of timers rather than re-using the same heartbeat one on successive loops. This swaps out time.After() with time.NewTimer() and then explicitly stops the timer when we are done using it.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -794,8 +794,10 @@ processWorkflowLoop:
 			heartbeatLoop:
 				for {
 					if delayDuration <= 0 {
-						heartbeatTimer.Stop()
-						heartbeatTimer = nil
+						if heartbeatTimer != nil {
+							heartbeatTimer.Stop()
+							heartbeatTimer = nil
+						}
 
 						// force complete, call the workflow task heartbeat function
 						workflowTask, err = heartbeatFunc(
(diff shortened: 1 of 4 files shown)


internal/internal_task_handlers.go
```diff
@@ -786,6 +794,9 @@ processWorkflowLoop:
 			heartbeatLoop:
 				for {
 					if delayDuration <= 0 {
+						heartbeatTimer.Stop()
```
wxing1292, line 797: not sure if it is possible to see this line execute > 1 times consecutively
so perhaps
```go
if heartbeatTimer != nil {
	heartbeatTimer.Stop()
	heartbeatTimer = nil
}
```

BTW, this pile of code needs rewrite (3 for loops)

wxing1292 (approved):

internal/internal_task_handlers.go
mastermanu: good catch. fixed

PR #302, opened 2020-11-30
Fix cancellation of a command before it is sent to server causing panic

Sushisource:
Fixes the issue described here: https://community.temporal.io/t/cancelling-timer-inside-select-causes-worker-panic/1078


vitarb (approved):
Please confirm with Max that he is okay with this approach, since it is different from what we've discussed. Code looks good to me, and given that tests pass, I'm going to approve the change.

Sushisource:
@mfateev Simply decrementing seems to fix this with pretty minimal change. Vitaly and I discussed and it may make sense to merge this (assuming you think it's fine) and then subsequently eliminate manipulating IDs entirely and try to take the approach the Java SDK does. Thoughts?

samarabbas (commented):
I don't think this fix will work in all situations.  Try the following use case:
```
create timer
execute activity
cancel timer
```

Sushisource:
@samarabbas I added a test with that case. Appears to be working, but let me know if this is not what you were imagining.

internal/internal_decision_state_machine.go
```diff
@@ -454,16 +491,22 @@ func (d *commandStateMachineBase) String() string {
...
 		command := createNewCommand(enumspb.COMMAND_TYPE_SCHEDULE_ACTIVITY_TASK)
 		command.Attributes = &commandpb.Command_ScheduleActivityTaskCommandAttributes{ScheduleActivityTaskCommandAttributes: d.attributes}
 		return command
 	case commandStateCanceledAfterInitiated:
-		command := createNewCommand(enumspb.COMMAND_TYPE_REQUEST_CANCEL_ACTIVITY_TASK)
-		command.Attributes = &commandpb.Command_RequestCancelActivityTaskCommandAttributes{RequestCancelActivityTaskCommandAttributes: &commandpb.RequestCancelActivityTaskCommandAttributes{
-			ScheduledEventId: d.scheduleID,
-		}}
-		return command
+		println("Act cmd get: ", d.id.String(), " sched: ", d.scheduleID)
```
vitarb, line 499: We shouldn't use println, use logger instead.

Sushisource: These are all getting ripped out later. They're just in there until I fix the tests

internal/internal_decision_state_machine.go
```diff
@@ -371,14 +406,11 @@ func (d *commandStateMachineBase) cancel() {
...
-		d.moveState(commandStateCompleted, eventCancel)
+		d.moveState(commandStateCanceledBeforeSent, eventCancel)
 	case commandStateCommandSent:
-		d.moveState(commandStateCanceledBeforeInitiated, eventCancel)
+		d.moveState(commandStateCancellationCommandSent, eventCancel)
 	case commandStateInitiated:
 		d.moveState(commandStateCanceledAfterInitiated, eventCancel)
-		// cancel doesn't add new command, therefore addCommand is not called.
-		// But *CancelRequested event is still being added to the history, therefore counter needs to be incremented.
-		d.helper.incrementNextCommandEventID()
```
samarabbas, line 381: Will this break backwards compatibility?

internal/internal_decision_state_machine.go
```diff
@@ -371,14 +406,11 @@ func (d *commandStateMachineBase) cancel() {
 	case commandStateCompleted, commandStateCompletedAfterCancellationCommandSent:
 		// No op. This is legit. People could cancel context after timer/activity is done.
 	case commandStateCreated:
-		d.moveState(commandStateCompleted, eventCancel)
+		d.moveState(commandStateCanceledBeforeSent, eventCancel)
 	case commandStateCommandSent:
-		d.moveState(commandStateCanceledBeforeInitiated, eventCancel)
+		d.moveState(commandStateCancellationCommandSent, eventCancel)
```
samarabbas, line 411: Do we need to differentiate for cancellation to be before sent/after sent?

internal/internal_decision_state_machine.go
```diff
@@ -388,7 +420,7 @@ func (d *commandStateMachineBase) handleInitiatedEvent() {
 	switch d.state {
 	case commandStateCommandSent:
 		d.moveState(commandStateInitiated, eventInitiated)
-	case commandStateCanceledBeforeInitiated:
+	case commandStateCanceledBeforeInitiated, commandStateCanceledBeforeSent, commandStateCancellationCommandSent:
```
samarabbas, line 423: This is super confusing.  There are many state machines where we make requestCancel first and then it is accepted or rejected.  Like activity and child workflow cancellation.
I don't think following state transition is valid:
commandStateCancellationCommandSent -> commandStateCanceledAfterInitiated

It should be `commandStateRequestCancel`.

internal/internal_decision_state_machine.go
```diff
@@ -440,7 +472,7 @@ func (d *commandStateMachineBase) handleCancelFailedEvent() {
 
 func (d *commandStateMachineBase) handleCanceledEvent() {
 	switch d.state {
-	case commandStateCancellationCommandSent:
+	case commandStateCancellationCommandSent, commandStateCanceledAfterInitiated, commandStateCanceledBeforeSent:
```
samarabbas, line 475: How can state machine in the state `commandStateCanceledBeforeSent` when handling `CanceledEvent`.  Server cannot generate canceled event on its own.  The command has to be sent for the server to generate the event.

internal/internal_decision_state_machine.go
```diff
@@ -485,13 +511,34 @@ func (d *activityCommandStateMachine) handleCancelFailedEvent() {
 	d.failStateTransition(eventCancelFailed)
 }
 
+func (d *activityCommandStateMachine) cancel() {
+	switch d.state {
+	case commandStateCreated, commandStateCommandSent:
+		attribs := &commandpb.RequestCancelActivityTaskCommandAttributes{
+			ScheduledEventId: d.scheduleID,
+		}
+		cancelCmd := d.helper.newCancelActivityStateMachine(attribs)
```
samarabbas, line 520: Do we create `newCancelActivityStateMachine` state machine only when cancellation happens in the same workflow task?  If that is the case then we should not include `commandStateCommandSent` in the case statement.

internal/internal_decision_state_machine.go
Sushisource: It certainly is confusing, heh - to be clear you're suggesting adding a new state `commandStateRequestCancel` and using that rather than `commandStateCanceledAfterInitiated` here?

internal/internal_decision_state_machine.go
Sushisource: So the idea here was these new `cancelTimerStateMachine` and `cancelActivityStateMachine` machines are the ones that generate the command, and so the original machines then handle their events. That said, though, probably it means these states should not be in this base handler, but explicitly in the activity/timer handlers.

internal/internal_decision_state_machine.go
Sushisource: As it stands without that case we would fail to send the cancellation command if the activity command itself has already been sent. I'm definitely not clear on the implications of being in the same workflow task or not so let's discuss that on our call

internal/internal_decision_state_machine.go
Sushisource: Per our discussion, there is one change in `internal_worker_test`: https://github.com/temporalio/sdk-go/pull/302/files#diff-2c7ff428b906672420092993fc82f49bcee4deff1806b693e2bad3c9eb59a13a where the ordering of activity test scheduled & cancel requested have swapped. Would that constitute a breaking change?

internal/internal_decision_state_machine.go
Sushisource: Indeed at least here the `commandStateCanceledBeforeSent` can be removed without regressing anything so this one is better now.

internal/internal_decision_state_machine.go
Sushisource: There are no changes to that test now.

internal/internal_decision_state_machine_test.go
```diff
@@ -150,8 +149,9 @@ func Test_TimerCancelEventOrdering(t *testing.T) {
 	require.Equal(t, commandStateCanceledAfterInitiated, d.getState())
 	commands = h.getCommands(true)
 	require.Equal(t, 2, len(commands))
-	require.Equal(t, enumspb.COMMAND_TYPE_RECORD_MARKER, commands[0].GetCommandType())
-	require.Equal(t, enumspb.COMMAND_TYPE_CANCEL_TIMER, commands[1].GetCommandType())
+	// TODO: UUUUGH
```
samarabbas, line 152: We need to guarantee same order in which commands are generated.

internal/internal_decision_state_machine.go
```diff
@@ -905,6 +991,8 @@ func (h *commandsHelper) recordLocalActivityMarker(activityID string, details ma
 	// to correctly incrementing counter before adding the command.
 	h.getNextID()
 	h.addCommand(command)
+	// LA markers should be the first command executed
+	h.orderedCommands.MoveToFront(h.commands[command.getID()])
```
Sushisource, line 995: @samarabbas So, this line makes that unit test referenced in the other comment pass.

However, I'm not sure that what it's asserting is actually a real concern. To check this, I wrote the workflow here: https://github.com/temporalio/sdk-go/pull/302/files#diff-8118e44ac15fb5ab2e2ece2ab52fd23062a10b5cc9c55ee5b3620a1242cff27cR749

Then I ran it against `master`, and against my changes. Without my changes, `TimerCanceled` shows up in the history *before* `MarkerRecorded` -- even though the timer is canceled after the local activity is started. After my changes the order is more like what the unit test claims, but in either case the workflows complete just fine.

Any thoughts on this? My gut feel is the unit test maybe isn't actually asserting anything important by saying the marker must be recorded before the timer is cancelled (in the same workflow task), in which case we can get rid of this line and there's no more reshuffling.

samarabbas (approved):

PR #303, opened 2020-12-04
Fix spurious nil derefs on retrying local activities

Sushisource:
Depends on https://github.com/temporalio/sdk-go/pull/302 being merged first
Relates to https://www.notion.so/Go-SDK-Nil-Reference-Panic-in-Bench-Test-b240b5f0db3b4d8088e3486939a53e78

~It appears my fix in the MR this one depends on also fixes this item. I wasn't able to repro the issue after a variety of bench runs.~ Not quite. Will require a bit of extra fix.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -84,6 +84,9 @@ type (
 		historyIterator HistoryIterator
 		doneCh          chan struct{}
 		laResultCh      chan *localActivityResult
+		// This channel must be initialized with a one-size buffer and is used to indicate when
+		// it is time for a local activity to be retried
+		laRetryCh chan *localActivityTask
 	}
 
 	// activityTask wraps a activity task.
@@ -577,8 +580,7 @@ func (w *workflowExecutionContextImpl) createEventHandler() {
 		w.wth.contextPropagators,
 		w.wth.tracer,
 	)
-	w.mutex.Lock()
-	defer w.mutex.Unlock()
+
 	w.eventHandler = &eventHandler
 }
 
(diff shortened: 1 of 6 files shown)


internal/internal_task_handlers.go
```diff
@@ -494,15 +493,10 @@ func (w *workflowExecutionContextImpl) Unlock(err error) {
...
-	if eventHandler == nil {
+	if w.eventHandler == nil {
 		return nil
 	}
-	eventHandlerImpl, ok := eventHandler.(*workflowExecutionEventHandlerImpl)
-	if !ok {
-		panic("unknown type for workflow execution event handler")
-	}
-	return eventHandlerImpl
+	return (*w.eventHandler).(*workflowExecutionEventHandlerImpl)
```
wxing1292, line 499: do we need to grab a lock for this?

wxing1292: since previously this is stored in atomic.Value

Sushisource: Putting it here can cause deadlocks - generally functions calling into this one are doing the locking. Overall there's not much consistency around when the lock is taken. I'd be worth cleaning this up as part of the bigger refactorings I'll take on soon.

internal/internal_task_handlers_test.go
```diff
@@ -1214,6 +1215,97 @@ func (t *TaskHandlersTestSuite) TestWorkflowTask_PageToken() {
 	t.NotNil(response)
 }
 
+func (t *TaskHandlersTestSuite) TestLocalActivityRetry_Workflow() {
```
vitarb, line 1218: Should we add a test that validates scenario mentioned in https://github.com/uber-go/cadence-client/issues/915?

Sushisource: This one doesn't - but the existing one directly after it does, so I think the path is covered.

vitarb: Yeah, I see, there is `TestLocalActivityRetry_WorkflowTaskHeartbeatFail`. Looks good to me then.

vitarb (approved):

PR #304, opened 2020-12-05
Set namespace field in async task completion and heartbeat requests

sergeybykov:
Set namespace fields on async task completion and heartbeat requests, so that it could be compared on the server with namespace in the token and/or those calls could be authorized on a namespace basis.


go.sum
```diff
@@ -40,6 +42,7 @@ github.com/golang/protobuf v1.4.0-rc.2/go.mod h1:LlEzMj4AhA7rCAGe4KMBDvJI+AwstrU
 github.com/golang/protobuf v1.4.0-rc.4.0.20200313231945-b860323f09d0/go.mod h1:WU3c8KckQ9AFe+yFwt9sWVRKCVIyN9cPHBJSNnbL67w=
 github.com/golang/protobuf v1.4.0/go.mod h1:jodUvKwWbYaEsadDk5Fwe5c77LiNKVO9IDvqG2KuDX0=
```
mastermanu, line 43: we should run go mod tidy as there is no reason we should have all of these references here.

internal/internal_workflow_testsuite.go
```diff
@@ -61,6 +61,7 @@ const (
 	defaultTestRunID            = "default-test-run-id"
 	defaultTestWorkflowTypeName = "default-test-workflow-type-name"
 	workflowTypeNotSpecified    = "workflow-type-not-specified"
+	testNamesapce               = "default"
```
mastermanu, line 64: [Nit] fix typo

mastermanu (approved):

samarabbas (approved):

PR #308, opened 2020-12-10
Properly handle infinite activity timeout

wxing1292:


wxing1292:
Also see alternative: https://github.com/temporalio/temporal/pull/1073

samarabbas (approved):

PR #310, opened 2020-12-15
Stop using a global, infinite-life workflow cache

Sushisource:
Relates to https://www.notion.so/temporalio/sticky-cache-fix-resource-leaks-on-worker-Stop-3701a1d197f14046bd99182a86856257

This stops us from using a global infinite-lifetime cache. There are some open questions: (draft until they / TODOs resolved)

- Multiple workers in one process used to be able to share a cache. Do we still care about doing this?
- Why would someone run multiple workers in a process? Our docs explicitly recommend not doing this: https://docs.temporal.io/docs/go-workers 
- Fallout from that, what to do about the publicly exposed `SetStickyWorkflowCacheSize`? This keeps making sense if we have workers share one cache which disappears when there are no more refs to it. If we want a cache per-worker, it no longer makes sense.


mfateev:
> Multiple workers in one process used to be able to share a cache. Do we still care about doing this?

I believe a single shared cache leads to better resource utilization.

> Why would someone run multiple workers in a process? 

There are multiple scenarios when it is necessary. For example, if the user wants to throttle different activities separately. 

> Fallout from that, what to do about the publicly exposed SetStickyWorkflowCacheSize?

Let's keep the global cache.

internal/internal_worker_cache.go
```diff
@@ -0,0 +1,124 @@
...
+	stickyCacheLock.Lock()
+	defer stickyCacheLock.Unlock()
+	if workflowCache != nil {
+		panic("cache already created, please set cache size before worker starts.")
+	}
+	stickyCacheSize = cacheSize
+}
+
+// Returns a pointer to the workerCache, and increases the outstanding refcount by one. Instances of
+// workerCache decrement the refcounter as a hook to runtime.SetFinalizer
```
mfateev, line 67: I would rename it to something that indicates that it is not just get, but actually more like lease that should be returned.

internal/internal_worker_cache.go
```diff
@@ -0,0 +1,124 @@
...
+	workflowCache *cache.Cache
+}
+
+// Don't touch this except via methods on workerCache. Holds shared workflow cache.
+var workflowCache cache.Cache
+var stickyCacheSize = defaultStickyCacheSize
+var stickyCacheLock sync.Mutex
+
+// Holds the number of outstanding references to the cache so we can zero it out if all workers are dead
+var workerRefcount atomic.Int32
```
mfateev, line 48: It is confusing to have both mutex and atomic in the same functions. IMHO just using mutex and int counter is simpler and perf is not an issue in this code path.

internal/internal_worker_cache.go
```diff
@@ -0,0 +1,124 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
```
mfateev, line 24: Did you run go fmt over this file?

internal/internal_worker_test.go
```diff
@@ -1089,11 +1089,13 @@ func (s *internalWorkerTestSuite) testWorkflowTaskHandlerHelper(params workerExe
 }
 
 func (s *internalWorkerTestSuite) TestWorkflowTaskHandlerWithDataConverter() {
+	cache := getWorkerCache()
 	params := workerExecutionParameters{
 		Namespace:     testNamespace,
 		Identity:      "identity",
 		Logger:        getLogger(),
 		DataConverter: iconverter.NewTestDataConverter(),
+		cache:         cache,
```
mfateev, line 1098: I couldn't find the test that validates that cache is deleted after all workers are closed.

Sushisource: @mfateev Yeah, could use a pointer on that one. The problem is such a test needs to run independently of other tests, since otherwise concurrently running tests will mess up the worker count. I couldn't find a good way to say "run this test independently" without actually changing how we call `go test` -- any ideas there?

internal/internal_worker_cache.go
Sushisource: Yeah, it's set up to auto-run on everything.

internal/internal_worker_cache.go
Sushisource: I changed the semantics just a little to something that makes a bit more sense. Let me know what you think now.

internal/internal_worker_test.go
Sushisource: Fixed this by using another layer of indirection (always the solution :smile: ) to make things more testable.

internal/internal_worker_cache.go
```diff
@@ -55,55 +62,66 @@ var workerRefcount atomic.Int32
...
-		defer stickyCacheLock.Unlock()
-		newcache := cache.New(stickyCacheSize, &cache.Options{
+// Creates a new workerCache, and increases workerRefcount by one. Instances of workerCache decrement the refcounter as
+// a hook to runtime.SetFinalizer (ie: When they are freed by the GC). When there are no reachable instances of
+// workerCache, shared caches will be cleared
+func newWorkerCache() *workerCache {
+	return _newWorkerCache(sharedWorkerCachePtr, &sharedWorkerCacheLock)
+}
+
+// This private version allows us to test functionality without affecting the global shared cache
```
mfateev, line 80: I would call the above method NewWorkerCache and this one newWorkerCache as Go uses capitalization of the first letter to indicate private/public

mfateev (approved):

Issue #284, opened 2020-11-09
Add the ability to register local activities only


wolfy-j:
**Is your feature request related to a problem? Please describe.**
Sometimes it makes sense to isolate workflow workers from the rest of business logic. However, local activites do not fit this mix. It will be useful to be able to register local activities only without enabling task queue activity consumption.

**Describe the solution you'd like**
It will be useful to be able to register local activities only without enabling task queue activity consumption.

**Describe alternatives you've considered**
none

vitarb:
From what I understand your concern is that in order to invoke a local activity you have to register activity with the worker and that would result in worker polling on activity task queue? What problems do you see with the current approach? Can you provide a use case when current behavior would be problematic?

mfateev:
@vitarb It is needed for PHP SDK worker.

vitarb:
Makes sense, we can add an option, something like `worker.Options.DisableActivityPoller` which, if set, would override our default logic and disable activity poller.

wolfy-j:
That will work!

vitarb:
We should be able to do this in the next few weeks, stay tuned.

vitarb:
Change is merged in, it should go out in the next 1.0.4 release soon.

vitarb:
Addressed in both go and java SDKs, new `LocalActivityWorkerOnly` worker option is available, which if set to true would turn off regular activity polling.

PR #312, opened 2020-12-16
Adding an option to allow starting local activity worker only

vitarb:
Similarly to [this change in Java SDK](https://github.com/temporalio/sdk-java/pull/276) this PR addresses #284 adding ability to set `LocalActivityWorkerOnly` flag in the worker options in order to avoid starting activity and session workers.
This allows running local activity workers only without any overhead for activity polling.

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -150,6 +150,10 @@ func (ts *IntegrationTestSuite) SetupTest() {
 		options.EnableSessionWorker = true
 	}
 
+	if strings.Contains(ts.T().Name(), "LocalActivityWorkerOnly") {
+		options.LocalActivityWorkerOnly = true
+	}
+
 	ts.worker = worker.New(ts.client, ts.taskQueueName, options)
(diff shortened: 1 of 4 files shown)


test/integration_test.go
```diff
@@ -150,6 +150,10 @@ func (ts *IntegrationTestSuite) SetupTest() {
 		options.EnableSessionWorker = true
 	}
 
+	if strings.Contains(ts.T().Name(), "LocalActivityWorkerOnly") {
+		options.LocalActivityWorkerOnly = true
+	}
+
```
Sushisource, line 156: nit: This is already being done above, but it'd be nice if we could somehow pass this in from the test itself rather than have setup inspect the test name.

Sushisource (approved):

test/integration_test.go
vitarb: I agree, it looks hacky, but I don't see a cleaner way of doing this. Will probably leave it as is for now.

PR #315, opened 2020-12-21
Prototype of our own state machine interfaces

Sushisource:


internal/statemachines/activity.go
```diff
@@ -0,0 +1,152 @@
...
+		panic("Attempted to stringify unknown activity state. This should be impossible.")
+	}
+}
+
+// TODO: Replace XX. prefixes with different types / constants etc
+func NewActivityStateMachine() *ActivityStateMachine {
+	fsmDef := BuildStateMachine("ActivityStateMachine", stateCreated, []State{stateCompleted, stateFailed, stateTimedOut, stateCanceled})
+	fsmDef.add(stateCreated, ExplicitEventSchedule, stateScheduleCommandCreated)
+
+	fsmDef.add(stateScheduleCommandCreated, CommandEvent{enumspb.COMMAND_TYPE_SCHEDULE_ACTIVITY_TASK}, stateScheduleCommandCreated)
```
mfateev, line 99: Why not have addCommand and addEvent methods instead?

Sushisource: That totally works for me. I was trying to stick to the Java semantics, but the fact that I have to use these wrapper structs since there are no generics kinda defeats the point.

In fact, may as well have `addExplicitEvent` be separate too.

Sushisource:
Closing this since we're likely going straight to a Rust-based core.

PR #317, opened 2020-12-23
Ensure workflow already started/finished errors are propagated back to user

### LINKED ISSUE #220: Unable to get WorkflowExecutionAlreadyStartedError
After migration from Uber Cadence to Temporal.io my internal logics are broken. As I see `StartWorkflow` was removed from interface and `ExecuteWorkflow` consumes error `WorkflowExecutionAlreadyStartedError` and returns existing `WorkflowRun`.

## Expected Behavior

Being able to receive `WorkflowExecutionAlreadyStartedError` on `ExecuteWorkflow` if `ReusePolicy` is `WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE_FAILED_ONLY`

## Actual Behavior

Unable to receive `WorkflowExecutionAlreadyStartedError` on `ExecuteWorkflow` if `ReusePolicy` is `WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE_FAILED_ONLY`

  1. Execute workflow with `WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE_FAILED_ONLY`
  2. Execute workflow with `WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE_FAILED_ONLY` with the same workflow id again.
  

## Specifications

  - Version: v0.26.0
  - Platform: linux x86_64

Sushisource:
Closes #220 

But this could break compat, potentially for some users, so unclear if we actually want to do it.


mfateev:
I would try adding this feature without breaking the backward compatibility. My proposal is to add an option to the StartWorkflowOptions. When set this option would throw already started independently from the IdReusePolicy.

Sushisource:
> I would try adding this feature without breaking the backward compatibility. My proposal is to add an option to the StartWorkflowOptions. When set this option would throw already started independently from the IdReusePolicy.

We can do that. My concerns would be discoverability of that option, and inconsistency between the SDKs - but someone getting unexpectedly broken is probably worse. I'll make the change and try to make things clear in docstrings

mfateev:
We need a similar option for Java SDK as well.

samarabbas:
Do we already have an API which allows connecting to existing workflow execution without starting a new one?  We should also add that too if one does now exist.

Sushisource:
@samarabbas Existing function that does what you want is `Client.GetWorkflow`

internal/client.go
```diff
@@ -431,10 +431,16 @@ type (
 		WorkflowTaskTimeout time.Duration
 
 		// WorkflowIDReusePolicy - Whether server allow reuse of workflow ID, can be useful
-		// for dedup logic if set to RejectDuplicate.
+		// for dedupe logic if set to RejectDuplicate.
 		// Optional: defaulted to AllowDuplicate.
 		WorkflowIDReusePolicy enumspb.WorkflowIdReusePolicy
 
+		// When WorkflowExecutionErrorWhenAlreadyStarted is true, Client.ExecuteWorkflow will return an error if the
+		// workflow id has already been used and WorkflowIDReusePolicy would disallow a re-run.
```
mfateev, line 439: Document what happens when it is set to false.

internal/internal_workflow_client_test.go
```diff
@@ -442,75 +442,75 @@ func (s *workflowRunSuite) TestExecuteWorkflow_NoDup_RawHistory_Success() {
...
 	)
-	s.Nil(err)
-	s.Equal(workflowRun.GetID(), workflowID)
-	s.Equal(workflowRun.GetRunID(), runID)
-	decodedResult := time.Minute
-	err = workflowRun.Get(context.Background(), &decodedResult)
-	s.Nil(err)
-	s.Equal(workflowResult, decodedResult)
+	s.Error(err)
+	s.Equal(mockerr, err)
```
mfateev, line 461: I might be confused, but I believe that the previous version of this test was testing that workflow reconnects to the already completed one and returns its result. The new test tests "WorkflowExecutionErrorWhenAlreadyStarted: true" case, but doesn't test the false case anymore.

Sushisource: The same tests that existed before still exist below, I just factored out their common code.

internal/client.go
mfateev: When WorkflowExecutionErrorWhenAlreadyStarted  is true then WorkflowExecutionErrorWhenAlreadyStarted will be always thrown for a workflow that is open independently of WorkflowIdReusePolicy. And for the closed workflows it is going to throw if no duplicated workflow can be created according to the WorkflowIdReusePolicy.

When WorkflowExecutionErrorWhenAlreadyStarted is false then WorkflowExecutionErrorWhenAlreadyStarted is thrown for open workflows only for WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE and never for the closed ones.

mfateev (approved):

Issue #276, opened 2020-10-26
Reduce the severnity of "no activities/workflow registered" message by Temporal worker


wolfy-j:
**Is your feature request related to a problem? Please describe.**
We are running separate workers for activities and workflows (server/client model with PHP), seeing this message is pretty annoying and does not bring any value.

**Describe the solution you'd like**
Keep the message but use DEBUG level.

**Additional context**
http://prntscr.com/v7a9r4

PR #322, opened 2020-12-28
PHP-SDK: PurgeStickyWorkflowCache, worker.Identity, debug messages

### LINKED ISSUE #279: StartWorkflow and OnWorkflowTaskStarted triggered twice on the same worker on QueryWorkflow
## Expected Behavior
Temporal should either use existing workflow env or destroy the previous version before queriing it.

## Actual Behavior
Temporal Go SDK issues a second workflow version without destroying previous state which corrupts the workflow state if external memory used. 

We are using extension API (**go.temporal.io/sdk/internalbindings**).

wolfy-j:
This PR contains two modifications into worker and worker cache behavior:

1) worker.Identity
It is now possible to set worker.Identity separately from the client.Identity. This is needed since we can reload the codebase of the application (PHP, Python, any other scripting language) without closing the connection.

2) PurgeStickyWorkflowCache
The destruction of a worker in Temporal leaves the cache open which is causing invalid behavior when a new worker with a new codebase attempts to access the same workflow. This method clears the cached content.

3) Addresses #276, potentially fixes #279 (load testing required)

Please let me know if we have to modify this PR somehow.


Sushisource:
@wolfy-j Thanks for this! This all looks good to me. If there are no existing workers, the cache should get GC'd -- though it might take a little while depending on when Go decides to collect. Are you saying you never saw it cleared? Or simply that it didn't happen fast enough? Would be nice to get that figured out.

It might be worth adding a call to your new purge method here: https://github.com/temporalio/sdk-go/pull/322/files#diff-2bb1ea4d5241bc089fae7d48a7ada4e5c84c4854ce94e30d56727ba01bb427c0R127-R129

wolfy-j:
Hi,

I do not think such call is required for the normal flow in the SDK. The problem is that GC of the workflow env completes after closing the worker, which is fine in typical scenario. 

However, in PHP SDK we keep the state of the workflow in external process, this causes the GC completion to attempt to send the command after worker was closed, and, sometimes, when new worker was already created, which is creating a collision.

It does not cause any critical failure but rather panic in log since original worker can't complete (PHP process is long gone).

Sushisource (approved):
Ah, ok that makes sense to me. Looks good to me then! We'll get a release out the door soon. Thanks!

PR #323, opened 2020-12-29
Child workflow cancellation unusual event ordering bugfix

Sushisource:
Addresses https://www.notion.so/temporalio/Child-Workflow-Cancellation-FSM-Error-ee8c794428134c79a55345fca56e7f25

Issue here is sometimes the SDK would see events in this order:
```
StartChildWorkflowExecutionInitiated
ChildWorkflowExecutionStarted
RequestCancelExternalWorkflowExecutionInitiated
ChildWorkflowExecutionCanceled
ExternalWorkflowExecutionCancelRequested
```

When it was only capable of handling an ordering where the last two events are swapped. IE: The cancel is accepted (`ExternalWorkflowExecutionCancelRequested`) before the cancel happens. However in rare situations that acceptance can be slow and arrive after the cancel. This should be allowed, and now the SDK can handle it.


wxing1292 (approved):

internal/internal_decision_state_machine.go
```diff
@@ -1077,7 +1083,14 @@ func (h *commandsHelper) handleExternalWorkflowExecutionCancelRequested(initiate
...
 	if !isExternal {
 		command = h.getCommand(makeCommandID(commandTypeChildWorkflow, workflowID))
-		// no state change for child workflow, it is still in CancellationCommandSent
+		asChildWfCmd := command.(*childWorkflowCommandStateMachine)
+		if asChildWfCmd.getState() == commandStateCompletedAfterCancellationCommandSent {
+			// Now we're really done.
+			command.handleCompletionEvent()
+		} else {
+			// We're in the cancellation command sent stage - new state to indicate we have seen the cancel accepted
+			asChildWfCmd.moveState(commandStateCancellationCommandAccepted, eventExternalWorkflowExecutionCancelRequested)
```
samarabbas, line 1092: Can we have handleCancelRequested api on the child workflow state machine?  It does not looks right to call moveState directly from commandsHelper.

internal/internal_decision_state_machine_test.go
```diff
@@ -461,6 +461,39 @@ func Test_ChildWorkflowStateMachine_InvalidStates(t *testing.T) {
 	require.NotNil(t, err)
 }
 
+func Test_ChildWorkflow_UnusualCancelationOrdering(t *testing.T) {
```
samarabbas, line 464: Can you also add a replay test for child workflow cancellation?

PR #330, opened 2021-01-08
Adding debug mode

vitarb:
Similarly to [this change](https://github.com/temporalio/sdk-java/pull/296) in the Java SDK, we are adding debug mode, which can be activated by setting `TEMPORAL_DEBUG` environment variable.
In this mode deadlock detector will be disabled.

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -29,6 +29,8 @@ package internal
 import (
 	"errors"
 	"fmt"
+	"math"
+	"os"
 	"reflect"
 	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/internal_workflow.go
```diff
@@ -260,6 +262,7 @@ var _ WaitGroup = (*waitGroupImpl)(nil)
 var _ dispatcher = (*dispatcherImpl)(nil)
 
 var stackBuf [100000]byte
+var debugMode = os.Getenv("TEMPORAL_DEBUG_MODE") == "true"
```
Sushisource, line 265: Nit: The java side used parseBoolean which I think will handle odd capitalizations and some other truthy values. For flags like this I like the idiom of if they're set to anything, they're true, and false is the absence of them. But, maybe better to be consistent with Java, and just lowercase the string or something before comparing.

Sushisource (approved):

internal/internal_workflow.go
alexshtin: Yeah, I would check for non-empty value and treat it as `true`. Also I would name it `TEMPORAL_DEBUG`, but this is up to you.

vitarb: Both are good comments. I was thinking about doing an empty check but decided to do it this way for consistency reasons with java implementation, and in java it's that way also for consistency reasons with other flags. But I totally agree that doing an empty check would be more idiomatic. Let me change it to that in both go and java.

PR #333, opened 2021-01-12
Fix panic when calling `GetWorkflow`

Sushisource:
with no run id when that workflow has never run before


internal/internal_workflow_client.go
```diff
@@ -315,7 +315,14 @@ func (wc *WorkflowClient) GetWorkflow(ctx context.Context, workflowID string, ru
 			if err != nil {
 				wc.logger.Error("error while fetching workflow execution info", err)
```
mfateev, line 316: I think we should remove this. It is not really an error.

mfateev (approved):

PR #337, opened 2021-01-13
[Testing] Add IsContinueAsNewError For Checking Errors in Tests

shannontan-bolt:
This adds a helper method to check if it is a continue as new error. I don't see any use cases aside from testing, so happy to close - or happy to put elsewhere. Made this practically copy-paste of temporal.IsCanceledError but I see that this can't exist in `temporal` for cyclic dependency reasons.

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -25,6 +25,8 @@
 package workflow
 
 import (
+	"errors"
+
 	"github.com/uber-go/tally"
 
 	"go.temporal.io/sdk/converter"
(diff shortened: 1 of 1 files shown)


vitarb:
Should we put this in the [error.go](https://github.com/temporalio/sdk-go/blob/0914229c3db0aa3dd57833681ef2d5714c20209c/temporal/error.go) instead?

shannontan-bolt:
> Should we put this in the [error.go](https://github.com/temporalio/sdk-go/blob/0914229c3db0aa3dd57833681ef2d5714c20209c/temporal/error.go) instead?

@vitarb - it can't for the same reason that NewContinueAsNewError doesn't exist there - there's a cyclical dependency if it were to be there. At least according to the PR that moved the ContinueAsNewError to this new location.

vitarb:
Makes sense, alright, let's keep it the way you did then.

vitarb (approved):

shannontan-bolt:
@vitarb - I actually can't merge this PR since I don't have write access. Mind merging for me?

shannontan-bolt:
@vitarb or @alexshtin - can someone merge this PR for me? Thanks!

I just keep having to update with master lol

PR #349, opened 2021-01-24
Add unit test for replay with side effect

samarabbas:


vitarb (approved):

vitarb:
Need to address couple complains raised by the error checker:
```
internal/internal_worker_test.go:2189:7:	Sleep(ctx, time.Second)
internal/internal_worker_test.go:2196:19:	encodedRandom.Get(&random)
internal/internal_worker_test.go:2209:10:	a1F.Get(ctx, &A1Result)
```

Issue #356, opened 2021-02-11
Registering Activity Fails Silently When Function Doesn't Match Signature


shannontan-bolt:
## Expected Behavior
When registering an activity struct that has methods that don't conform to the expected signature, activity registration should either fail with an error (maybe panic/fail in tests) or skip over the function.

## Actual Behavior
When registering an activity struct that has methods that don't conform to the expected signature because they don't intend on being registered, activity registration fails silently without indication of what method it failed to bind as an activity.

https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker.go#L534

## Steps to Reproduce the Problem
```
type Activity struct {}

func (a *Activity) ActivityMethod(ctx context.Context) error {}
func (a *Activity) OtherMethod(ctx context.Context) {}

...

env.RegisterActivityWithOptions(activities, activity.RegisterOptions{Name: "AsyncInputsActivities_"}) // no failure, no indication
```

## Specifications

  - Version: 1.4.1

shannontan-bolt:
I think the behavior should be 1 or the other - but not silent failure with only methods processed up until that nonconforming activity registered.

PR #360, opened 2021-02-11
Allow Skipping of Invalid Activity Functions in Struct

shannontan-bolt:
This allows activities to be registered that may have some Temporal activities along with other functions.

I couldn't find a way to test the negative case without causing a failure in the test start code that would require a re-run.

Intended to address #356.

### DIFF
--- a/internal/internal_worker_test.go
+++ b/internal/internal_worker_test.go
@@ -87,6 +87,11 @@ func testInternalWorkerRegister(r *registry) {
 	r.RegisterActivity(testActivityReturnStructPtr)
 	r.RegisterActivity(testActivityReturnNilStructPtrPtr)
 	r.RegisterActivity(testActivityReturnStructPtrPtr)
+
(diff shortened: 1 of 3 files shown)


vitarb (approved):

vitarb (changes requested):
Please address build failure.

internal/internal_worker_test.go
```diff
@@ -1726,6 +1740,12 @@ func testActivityReturnStructPtrPtr() (**testActivityResult, error) {
 	return &r, nil
 }
 
+type testActivityStructWithFns struct{}
+
+func (t *testActivityStructWithFns) ValidActivity(ctx context.Context) error { return nil }
+
+func (t *testActivityStructWithFns) InvalidActivity(ctx context.Context) { return }
```
vitarb, line 1747: This `return` should be removed, as the code won't compile this way.

Sushisource:
@shannontan-bolt I've fixed the problems and made a new PR here. I'm closing this one in favor of that. Thanks!

shannontan-bolt:
Thanks!

PR #361, opened 2021-02-11
Panic when Invalid Struct is Registered

shannontan-bolt:
Without this, an invalid struct can be partially registered in tests or in workers, which can lead to weird failure when activities are missing but unknown as to why they are missing.

Combines with #360 for solving #356.


shannontan-bolt:
I don't have good tests for this. Where/how should I test this?

mfateev (approved):
Thanks!

Issue #353, opened 2021-02-01
Allow the injection of additional grpc client interceptors


jmcnevin:
**Is your feature request related to a problem? Please describe.**

This may be related to https://github.com/temporalio/sdk-go/issues/277.

In order to use a custom claims mapper / authorizer in the server, we need to be able to inject authorization metadata into the client request, but this isn't currently possible with the sdk.  I've been able to fork SDK to allow for the injection of a ` grpc.UnaryClientInterceptor` via the `ConnectionOptions` struct in order to do this, but I would like to see what the Temporal team's strategy is for doing this.

**Describe the solution you'd like**

Anything that will allow the injection of authorization metadata into the outgoing grpc context.

**Describe alternatives you've considered**

There are probably more elegant solutions than I've put together, including the creation of some sort of AuthTokenGetter interface.

jmcnevin:
FWIW: this is the entirety of my current solution:

```diff
diff --git a/internal/client.go b/internal/client.go
index 58c4fa6..6bfbcee 100644
--- a/internal/client.go
+++ b/internal/client.go
@@ -393,9 +393,10 @@ type (

        // ConnectionOptions is provided by SDK consumers to control optional connection params.
        ConnectionOptions struct {
-               TLS                *tls.Config
-               DisableHealthCheck bool
-               HealthCheckTimeout time.Duration
+               TLS                    *tls.Config
+               DisableHealthCheck     bool
+               HealthCheckTimeout     time.Duration
+               CredentialsInterceptor grpc.UnaryClientInterceptor
        }

        // StartWorkflowOptions configuration parameters for starting a workflow execution.
diff --git a/internal/grpc_dialer.go b/internal/grpc_dialer.go
index bf23eb7..382701a 100644
--- a/internal/grpc_dialer.go
+++ b/internal/grpc_dialer.go
@@ -64,7 +64,6 @@ func dial(params dialParameters) (*grpc.ClientConn, error) {
        if params.UserConnectionOptions.TLS != nil {
@@ -64,7 +64,6 @@ func dial(params dialParameters) (*grpc.ClientConn, error) {
        if params.UserConnectionOptions.TLS != nil {
                grpcSecurityOptions = grpc.WithTransportCredentials(credentials.NewTLS(params.UserConnectionOptions.TLS))
        }
-
        // gRPC maintains connection pool inside grpc.ClientConn.
        // This connection pool has auto reconnect feature.
[...]

PR #363, opened 2021-02-16
Ability to provide a token provider function

jmcnevin:
In order to use a custom claims mapper in the temporal server, using data from the "authorization" GRPC metadata, we need a way to inject a value into the outgoing GRPC context.  This a first cut at that kind of functionality.

May need some guidance here as to the best place to put tests for this.

This is one possible solution for #353.


Sushisource:
@jmcnevin Thanks for the PR! I think you should be able to add a fairly straightforward unit test in `grpc_dialer_test.go` which looks a bit like the existing ones there.

vitarb:
Implemented in #368

PR #368, opened 2021-02-23
Adding support for custom gRPC headers

vitarb:
Adds ability to set `HeadersProvider` on `ClientOptions` during `Client` creation, which will be invoked in the gRPC interceptor on all outgoing calls.

Users can now implement HeadersProvider interface:
```
	HeadersProvider interface {
		GetHeaders(ctx context.Context) (map[string]string, error)
	}
```
and set authorization or any other headers that backend expects to see.

This solution is a slight refinement of #363, in my opinion we should try to not expose gRPC types to the end user because other newer language SDKs (JavaScript and beyond) will not have direct access to the gRPC layer, which will be done in the core rust sdk.

Another option that I've considered was adding ability to set headers on the `ClientOptions` and adding ability to modify them from the `Client`. But I didn't like this approach as it introduces a function into the `Client` that would be an outlier in terms of the responsibilities (all other functions are wrappers around RPC calls).

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -389,6 +389,15 @@ type (
 		// Optional: Sets options for server connection that allow users to control features of connections such as TLS settings.
 		// default: no extra options
 		ConnectionOptions ConnectionOptions
+
+		// Optional: HeadersProvider is getting invoked on every outgoing gRPC request and gives user ability to
+		// set custom request headers. This can be used to set auth headers for example.
+		HeadersProvider HeadersProvider
+	}
+
+	// HeadersProvider returns a map of gRPC headers that should be used on every request.
+	HeadersProvider interface {
+		GetHeaders(ctx context.Context) (map[string]string, error)
 	}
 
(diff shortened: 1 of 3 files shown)


alexshtin (approved):

internal/client.go
```diff
@@ -389,6 +389,15 @@ type (
 		// Optional: Sets options for server connection that allow users to control features of connections such as TLS settings.
 		// default: no extra options
 		ConnectionOptions ConnectionOptions
+
+		// Optional: HeadersProvider is getting invoked on every outgoing gRPC request and gives user ability to
```
Sushisource, line 393: Rephrase: ".. will be invoked .."

Sushisource (approved):
LGTM, this is a nice way to do it.

internal/client.go
vitarb: done

sergeybykov (approved):

PR #371, opened 2021-02-26
Make keep alive configurable and opt-in

vitarb:
This approach should make it safer for our users and allow us time to test the feature before we enable it by default.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -405,7 +405,7 @@ type (
 		TLS                *tls.Config
 		DisableHealthCheck bool
 		HealthCheckTimeout time.Duration
-		// Enables keep alive ping from client to the server, which can help drop abruptly closed connections faster.
+		// Enables keep alive ping from client to the server, which can help detect abruptly closed connections faster.
 		EnableKeepAliveCheck bool
 		// After a duration of this time if the client doesn't see any activity it
 		// pings the server to see if the transport is still alive.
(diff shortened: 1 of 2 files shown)


internal/client.go
```diff
@@ -405,6 +405,20 @@ type (
 		TLS                *tls.Config
 		DisableHealthCheck bool
 		HealthCheckTimeout time.Duration
+		// Enables keep alive ping from client to the server, which can help drop abruptly closed connections faster.
+		EnableKeepAliveCheck bool
+		// After a duration of this time if the client doesn't see any activity it
+		// pings the server to see if the transport is still alive.
+		// If set below 10s, a minimum value of 10s will be used instead.
```
Sushisource, line 412: Just checking that 10s min is enforced, doesn't show up obviously in this diff.

alexshtin (approved):

Sushisource (approved):

internal/client.go
vitarb: It's a gRPC default.

alexshtin: It is enforced by gRCP itself.

mastermanu (approved):

internal/client.go
```diff
@@ -405,6 +405,20 @@ type (
 		TLS                *tls.Config
 		DisableHealthCheck bool
 		HealthCheckTimeout time.Duration
+		// Enables keep alive ping from client to the server, which can help drop abruptly closed connections faster.
```
mastermanu, line 408: help detect

vitarb: detect and drop? :)

PR #385, opened 2021-03-15
Error Nil, Not Nil

houseofcat:
Various versions of Golang fluctuate by how prevalent of an issue it is, or when dependencies change and those authors forget to use explicit `nil` on return when err is in fact, `nil`.

```
err != nil {

}
```

Is sometimes true when err points to `nil` in Golang. The only solution is to never use implicit or ambiguous `err`. Only return `err` when it is not `nil`, and only return keyword `nil` when err was nil. Never use maybe assigned/unassigned `var err error` or `*error`.

It is safe to `return func() error` but only when the underlying function is using explicit returns itself.

```
	if err != nil {
		return err
	}
	return nil
```

As it stands right now, SDK 1.5.0 and at least Golang 1.15.8 / 1.16.2 these are the results on some of the SDK calls, usually when the retry mechanism is being called. Sometimes if you are stepping through Delve and are lucky enough to see the `void Unicorn` you see the pointer pointing to `void`, instead of the bug that displays error as `nil`. I have demonstrated both in screenshots below.

NilNotNil0
NilNotNil
NilNotNil2

Other Sources  
https://yourbasic.org/golang/gotcha-why-nil-error-not-equal-nil/  
https://codefibershq.com/blog/golang-why-nil-is-not-always-nil#:~:text=Nil%20represents%20a%20zero%20value,of%20checking%20values%20for%20emptyness.


houseofcat:
Execute `make test` and `make unit-test`. They don't appear to be in a healthy state. Master doesn't pass nor does my branch.

I will assume there is some special instruction that I am missing. I am happy to run both integration/unit-tests, instead I manually ran exposed tests with success.

I am committing one tweak to a unit test that fails on `Windows`.

alexshtin:
Difference between nil interface and nil pointer totally makes sense but I think this PR is overkill.
1. I believe it is save to pass `err` from all gRPC wrappers. 
2. What is the point of this?
```
	if err != nil {
		return err
	}
	return nil
```
Just to make sure that our unit/integration tests will catch the case when `err` is nil pointer?
3. I see many cases where you just blindly added the above block, but it is easy to trace that nil pointer error is never returned from the callee func.
4. I am ok to fix this in cases where it is needed bu I am against blindly adding it everywhere.

Btw, what is the `<error(void)>` on the last screenshot?

houseofcat:
The `<error(void)>` is I believe literally the real value of some `err` that are `nil` even when the developer sees `nil`. If you mean this very instance of error(void), it is the return value from local Retry function in the first screenshot returns err (even when `nil`).   

1 & 3.) I guess for me it comes down to, the boilerplate is ugly and obviously repetitive but works in 100% of all golang scenarios. It's also the semantic pattern from the golang developers (not the only one). It also handles the scenario if the code that is called changes what _**how it returns err/nil.**_. You would be very hard pressed to find `return err` without a nil check in golang itself. I have dug into it a few times hoping to find a better pattern for error handling specifically.  

2.) I am not sure what you mean by that block of code without a specific line lookup, but the pattern itself ties into I just tried helping out but converting you to the standard semantic pattern everywhere I could immediately see it. I tried my best to find most of them for you guys I am sure there are a few I missed.  

3.) To this I would just respond its the standard golang pattern handles conditions and side effects of future code changes. Obviously anything I wrote that's wrong is just wrong. The approach I took was if you were using `return err` and you didn't check for `nil`, I merely checked it for you. Some of the code you I had to be careful with. We may just disagree on this. It's your SDK after all, but I would not mix and match your `err` handling.  

4.) Happy to oblige, meant to be strictly helpful as I have been through this ringer before. I am C#/NetCore dev in my heart, but I have had my share of `golang` shocks that made it all the way in prod including the WHAT THE HECK, THIS IS NIL! HOW ARE WE IN THE NON-NIL CODE! Then my `golang` resource had to sit me down explain to me what was happening as I mumbled how stupid this was.  

All in all good feedback, if you want me to kill this PR that's fine let me know.

alexshtin:
I was inspired by this PR (thanks!) and fixed all the places where concrete error type is returned instead of `error` interface: #386. Yes, we still might have some concrete types returned from 3rd party libraries but I don't see such calls in this PR except gRPC generated code which I do trust. If there are such calls please fix them as separate PR and I am closing this one.

houseofcat:
@alexshtin sounds good Alex!

PR #388, opened 2021-03-24
Support extracting workflow and activity options from context

jlegrone:
### What was changed

This PR adds three new public functions to the `workflow` package:

- `GetChildWorkflowOptions`
- `GetActivityOptions`
- `GetLocalActivityOptions`

### Motivation & Context

In interceptors and workflow/activity wrapper functions, it was previously not possible to set default option values that could be overridden by the parent context.

By extracting the existing workflow or activity options from the workflow context, we can now conditionally set defaults only when a value has not already been specified:

```go
type FooRequest struct {
	Bar string
}
type FooResponse struct {
	Baz string
}

// Helper function to be used by worker implementations.
func ExecuteChildWorkflowFoo(ctx workflow.Context, req FooRequest) (FooResponse, error) {
	// Set default options
	opts := workflow.GetChildWorkflowOptions(ctx)
	if opts.TaskQueue == "" {
		opts.TaskQueue = "foo"
	}
	if opts.WorkflowRunTimeout == 0 {
		opts.WorkflowRunTimeout = time.Minute
	}
	ctx = workflow.WithChildOptions(ctx, opts)

	// Execute workflow and return a result
	var resp FooResponse
	err := workflow.ExecuteChildWorkflow(ctx, "Foo", req).Get(ctx, &resp)
	return resp, err
}
```

Closes issue: N/A

## How has this been tested?

The changeset should be covered by unit tests.

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.


internal/workflow_test.go
```diff
@@ -0,0 +1,136 @@
...
+		MaximumInterval:        3,
+		MaximumAttempts:        4,
+		NonRetryableErrorTypes: []string{"my_error"},
+	}
+}
+
+// assertNonZero checks that every top level value, struct field, and item in a slice is a non-zero value.
+func assertNonZero(t *testing.T, i interface{}) {
+	_assertNonZero(t, i, reflect.ValueOf(i).Type().Name())
+}
```
jlegrone, line 109: Apologies if this is overwrought; I think it's nice to have peace of mind that tests would break if new fields are added but not accounted for in tests or conversion functions, but it's not essential to this changeset.

mfateev (approved):
Awesome! Thanks for the contribution!

mfateev:
To merge the license header has to be added:
```



go run ./internal/cmd/tools/copyright/licensegen.go --verifyOnly
  | copyright header check failed, err=internal/workflow_test.go missing license header
  | exit status 255

```

PR #393, opened 2021-03-28
Support for context-aware data converters.

robholland:
## What was changed:
A new API as added so that DataConverter implementations can receive Workflow/Activity context and return a stateful version of themselves.

## Why?
This can be used to vary behaviour of the data converter, for example to vary encryption keys used by a converter based on workflow context (which should also be propagated to the activity context).

## How has this been tested?
Tested manually using the cryptconverter sample in samples-go developed alongside this change: https://github.com/temporalio/samples-go/pull/87, tctl support tested using the samples-go PR and https://github.com/temporalio/temporal/pull/1419.


converter/data_converter.go
```diff
@@ -52,4 +52,18 @@ type (
 		// ToStrings converts payloads object into human readable strings.
 		ToStrings(input *commonpb.Payloads) []string
 	}
+
+	DataConverterWithContext interface {
+		DataConverter
```
mfateev, line 57: I don't think there is a need to embed DataConverter interface into DataConverterWithContext. 

I would rename it to something else as it is not a DataConverter but more like mixin.

converter/data_converter.go
```diff
@@ -52,4 +52,18 @@ type (
...
+
+		WithContext(interface{}) DataConverter
```
mfateev, line 59: I would rename to `WithValue` as it is similar to `context.WithValue`

mfateev (commented):
Interesting idea. I think it should work. I would try to get an end-to-end implementation working as a separate PR before landing this.

robholland:
It's tested and working with https://github.com/temporalio/samples-go/pull/87/files. Still tidying things up though.

converter/data_converter.go
```diff
@@ -52,4 +52,17 @@ type (
 		// ToStrings converts payloads object into human readable strings.
 		ToStrings(input *commonpb.Payloads) []string
 	}
+
+	stateful interface {
```
mfateev, line 56: I think it should be public to support externally contributed extensions.

converter/data_converter.go
```diff
@@ -52,4 +52,17 @@ type (
...
 		ToStrings(input *commonpb.Payloads) []string
 	}
+
+	Stateful interface {
+		WithValue(interface{}) DataConverter
+	}
 )
+
+// WithValue returns a new DataConverter with state recorded.
+func WithValue(dc DataConverter, value interface{}) DataConverter {
```
Sushisource, line 62: nit: Maybe call `value` state to match the comment

converter/data_converter.go
```diff
@@ -52,4 +52,17 @@ type (
 		// ToStrings converts payloads object into human readable strings.
 		ToStrings(input *commonpb.Payloads) []string
 	}
+
+	Stateful interface {
```
Sushisource, line 56: Probably worth adding a docstring to this

Sushisource (approved):
LGTM, probably need some doc updates indicating that the context is going to be attached to the dataconverter by the `getDataConverterFromXXXXCtx` methods now, whether in doc comment or actual docs.

converter/plugin.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"net/rpc"
+
+	"github.com/hashicorp/go-plugin"
+	commonpb "go.temporal.io/api/common/v1"
+)
+
+type DataConverterRPC struct {
+	client *rpc.Client
+}
```
mfateev, line 12: Why is this in the SDK code? I thought we only plan to use this for the `tctl`.
I don't think we want SDK to depend on `go-plugin` pacakge.

mfateev (commented):
Do we have an integration test for the Stateful DataConverter?

converter/plugin.go
robholland: It can move to the `temporal` package if we're sure we won't want it for other purposes.

converter/data_converter.go
```diff
@@ -52,4 +52,19 @@ type (
 		// ToStrings converts payloads object into human readable strings.
 		ToStrings(input *commonpb.Payloads) []string
 	}
+
+	// Stateful DataConverters can return a copy of themselves tailored to a Workflow or Activity context.
+	// This interface is optional.
```
mfateev, line 57: Maybe make it more explicit that this interface can only be used as an extension interface to a DataConverter implementation?

mfateev (approved):
Awesome!

PR #394, opened 2021-03-29
Updated ActivityOptions.ScheduleToStartTimeout comment

mfateev:
## What was changed:
Updated ActivityOptions.ScheduleToStartTimeout comment to match the current service implementation.

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -86,7 +86,7 @@ type (
 		// Optional: The default value is the sum of ScheduleToStartTimeout and StartToCloseTimeout
 		ScheduleToCloseTimeout time.Duration
 
-		// ScheduleToStartTimeout - The maximum time an activity task can stay in a task queue before picked up by a worker.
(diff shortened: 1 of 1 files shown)


internal/activity.go
```diff
@@ -86,8 +86,11 @@ type (
 		// Optional: The default value is the sum of ScheduleToStartTimeout and StartToCloseTimeout
 		ScheduleToCloseTimeout time.Duration
 
-		// ScheduleToStartTimeout - The queue timeout before the activity starts executed.
-		// Mandatory: No default.
+		// ScheduleToStartTimeout - The maximum time an activity task can stay in a task queue before picked up by a worker.
+		// Note that ScheduleToStartTimeout is not retryable as retry would return it back into the same task queue.
```
Sushisource, line 90:
```suggestion
		// ScheduleToStartTimeout - The maximum time an activity task can stay in a task queue before being picked up by a worker.
		// Note that ScheduleToStartTimeout is not retryable as retry would return it back into the same task queue.
```

Sushisource (approved):

PR #396, opened 2021-04-01
Added ContextAware DataConverter extension

mfateev:
## What was changed:
A new API was added so that DataConverter implementations can receive Workflow/Activity context and return a stateful version of themselves.

## Why?
This can be used to vary behavior of the data converter, for example, to vary encryption keys used by a converter based on workflow context (which should also be propagated to the activity context).

## How has this been tested?
UnitTested

### DIFF
--- a/internal/headers.go
+++ b/internal/headers.go
@@ -26,42 +26,53 @@ package internal
 
 import (
 	"context"
+	"go.temporal.io/sdk/converter"
 
 	commonpb "go.temporal.io/api/common/v1"
 )
 
 // HeaderWriter is an interface to write information to temporal headers
-type HeaderWriter interface {
-	Set(string, *commonpb.Payload)
-}
+type (
+	HeaderWriter interface {
+		Set(string, *commonpb.Payload)
+	}
 
-// HeaderReader is an interface to read information from temporal headers
-type HeaderReader interface {
-	Get(string) (*commonpb.Payload, bool)
-	ForEachKey(handler func(string, *commonpb.Payload) error) error
-}
+	// HeaderReader is an interface to read information from temporal headers
+	HeaderReader interface {
+		Get(string) (*commonpb.Payload, bool)
+		ForEachKey(handler func(string, *commonpb.Payload) error) error
+	}
 
-// ContextPropagator is an interface that determines what information from
-// context to pass along
-type ContextPropagator interface {
-	// Inject injects information from a Go Context into headers
-	Inject(context.Context, HeaderWriter) error
+	// ContextPropagator is an interface that determines what information from
+	// context to pass along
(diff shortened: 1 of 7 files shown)


internal/headers.go
```diff
@@ -106,3 +117,23 @@ func NewHeaderWriter(header *commonpb.Header) HeaderWriter {
...
+// the DataConverter implements the ContextAware interface. Otherwise the DataConverter is returned
+// as-is.
+func WithWorkflowContext(ctx Context, dc converter.DataConverter) converter.DataConverter {
+	if d, ok := dc.(ContextAware); ok {
+		return d.WithWorkflowContext(ctx)
+	}
+	return dc
+}
+
+// WithContext returns a new DataConverter tailored to the passed Activity context if
```
robholland, line 131: nit: https://github.com/temporalio/sdk-go/blob/c8ef862218b0928928d66fda749bbe75dce9df15/internal/internal_workflow_client.go#L174 uses `WithContext` with a Context for a Workflow (but not as a workflow.Context as we don't have one there). Should perhaps broaden the comment to mention it won't always be an Activity.

robholland (commented):
Looks good but please also include: https://github.com/temporalio/sdk-go/pull/393/files#diff-667e8429b7c253de286f7d41d2fe894f3eed726dd956c6ca05bd98b1aaa63623R44. I think we should guarantee only that the default data converter implements DataConverter, the specific type is an internal detail.

robholland:
I've started testing this PR with https://github.com/temporalio/samples-go/pull/87/files. Having the WithContext/WithWorkflowContext in internal means that composite data converters are not able to pass the context down to their descendants: https://github.com/temporalio/samples-go/pull/87/files#diff-e74298db0b15fa4477fb91652135a3660e8e06818f9fb45f75a7f3cd8ac24103R28. It would have to be: `internal.WithWorkflowContext(ctx, dc.dataConverter),` which go refuses to compile: `cryptconverter/data-converter.go:10:2: use of internal package go.temporal.io/sdk/internal not allowed`

robholland (changes requested):
I'm not sure having this in internal provides a suitable API for composite data converters.

robholland:
Example of the minor annoyance from us not returning `DataConverter` from `GetDefaultDataConverter()`:

`../temporal/tools/cli/app.go:249:19: cannot use raw.(converter.DataConverter) (type converter.DataConverter) as type *converter.CompositeDataConverter in assignment: need type assertion`

```
myConverter := GetDefaultDataConverter()
myConverter = NewCustomDataConverter()
```

should work without having to cast.

mfateev:
Implemented through https://github.com/temporalio/sdk-go/pull/393

PR #397, opened 2021-04-01
Support executing local activities by name

### LINKED ISSUE #354: Allow string activity names in `workflow.ExecuteLocalActivity`
We have a DSL use-case and without this functionality, it requires multiple files changes when adding an activity unless we implement the reflection-based solution ourselves.

Allow string `activity` parameter values for `workflow.ExecuteLocalActivity` the same way `workflow.ExecuteActivity` allows it currently.

We could implement the reflection based call ourselves but currently we make the calls in a switch statement.

**Additional context**

lminaudier:
## What was changed:

The change replicates the behavior of `ExecuteActivity` for `ExecuteLocalActivity` to allow executing local activities by supplying funcs or func names.

In the process, `ExecuteLocalActivity` is using the activity registry to lookup the activity function by name when needed.

~One question I have is that this could be potentially considered a breaking change because right now, it seems local activities executions don't seem to rely on the registry meaning that potentially, you don't have to register local activities in your worker? I still have to test if that is true and if so, if this is a documented behaviour.~ It seems that's indeed the case [when I look at tests](https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_handlers_test.go#L1237-L1239). 

## Why?

This PR attempts to fix #354 to allow executing local activities by using the activity name similarly to non local activities.

## Checklist

1. Closes issue: closes #354
2. How was this tested: for now manually by using my fork on my own internal project
3. Any docs updates needed?: I don't think there is any doc around that specific to local activities.

## Todo

- [x] Fix existing tests
- [x] Add tests for local activity execution by name

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -542,31 +542,35 @@ func ExecuteLocalActivity(ctx Context, activity interface{}, args ...interface{}
 	i := getWorkflowOutboundCallsInterceptor(ctx)
 	env := getWorkflowEnvironment(ctx)
 	activityType := getActivityFunctionName(env.GetRegistry(), activity)
-	ctx = WithValue(ctx, localActivityFnContextKey, activity)
 	return i.ExecuteLocalActivity(ctx, activityType, args...)
 }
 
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
This looks great! Thanks for the PR and keeping things backwards compatible.

PR #398, opened 2021-04-01
Sync with common.ValidateRetryPolicy

jlcheng:
## What was changed:
Synced validation code from https://github.com/temporalio/temporal/blob/v1.8.1/common/util.go#L445

## Why?
A workflow using this code fails during unit test

```
			ScheduleToStartTimeout: time.Minute,             // Timeout between activity being scheduled to starting execution
			StartToCloseTimeout:    TransmitActivityTimeout, // Timeout of the execution
				MaximumAttempts: 1,
			},
		})
```

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?


Sushisource:
@jlcheng Thanks for the contribution! Adding a unit test that covers the case you mentioned failed would be really useful to have as well.

vitarb:
@jlcheng could you add a unit test as Spencer suggested then we could approve/land this change.

internal/internal_workflow_testsuite_test.go
```diff
@@ -2534,6 +2534,47 @@ func (s *WorkflowTestSuiteUnitTest) Test_ActivityRetry() {
 	s.Equal(4, attempt2Count)
 }
 
+func (s *WorkflowTestSuiteUnitTest) Test_ActivityRetry_NoRetries() {
+	maxRetryAttempts := int32(1)
```
Sushisource, line 2538: This should probably just be named `maxAttempts`

vitarb (approved):

PR #407, opened 2021-04-15
Expose gRPC Authority in ConnectionOptions

FreeGuwop:
## What was changed:

Exposed the gRPC authority as a connection option.

## Why?

gRPC Authority is used for routing in certain scenarios. This is a safe change as gRPC enforces that authority can only be overridden in insecure mode.

## Checklist

1. Closes issue: 

N/A.

2. How was this tested:

Integration and unit tests.

3. Any docs updates needed?

N/A.


alexshtin:
Reimplemented in #409.

PR #409, opened 2021-04-15
Add gRPC Authority connection option

alexshtin:
## What was changed:

Add gRPC Authority connection option.

## Why?

gRPC Authority is used for routing in certain scenarios. This is a safe change as gRPC enforces that authority can only be overridden in insecure mode.

### How was this tested
Integration and unit tests.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -407,19 +407,35 @@ type (
 
 	// ConnectionOptions is provided by SDK consumers to control optional connection params.
 	ConnectionOptions struct {
-		TLS                *tls.Config
+		// TLS configures a connection level security credentials.
+		TLS *tls.Config
+
+		// Authority specifies the value to be used as the :authority pseudo-header.
+		// This value only works only when TLS is nil.
+		Authority string
+
+		// By default, after gRPC connection to server is created, client will make a request to
+		// health check endpoint to make sure that server is accessible.
+		// Set DisableHealthCheck to true to disable it.
 		DisableHealthCheck bool
+
+		// Specify how to long to wait while checking server connection when creating new client.
+		// Default: 5s.
 		HealthCheckTimeout time.Duration
+
 		// Enables keep alive ping from client to the server, which can help detect abruptly closed connections faster.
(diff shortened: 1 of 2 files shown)


wxing1292 (approved):

internal/client.go
```diff
@@ -407,19 +407,35 @@ type (
 
 	// ConnectionOptions is provided by SDK consumers to control optional connection params.
 	ConnectionOptions struct {
-		TLS                *tls.Config
+		// TLS configures a connection level security credentials.
+		TLS *tls.Config
+
+		// Authority specifies the value to be used as the :authority pseudo-header.
+		// This value only works only when TLS is nil.
```
flossypurse, line 414: "Only works" might imply an error might happen?

Is it "only used when TLS is nil" ?

internal/client.go
```diff
@@ -407,19 +407,35 @@ type (
...
+		Authority string
+
+		// By default, after gRPC connection to server is created, client will make a request to
```
flossypurse, line 417:
```suggestion
		// By default, after gRPC connection to  the Server is created, client will make a request to
```

internal/client.go
```diff
@@ -407,19 +407,35 @@ type (
...
+		// health check endpoint to make sure that server is accessible.
```
flossypurse, line 418:
```suggestion
		// health check endpoint to make sure the Server is accessible.
```

internal/client.go
```diff
@@ -407,19 +407,35 @@ type (
...
+		// Set DisableHealthCheck to true to disable it.
 		DisableHealthCheck bool
+
+		// Specify how to long to wait while checking server connection when creating new client.
```
flossypurse, line 422: Shouldn't the comment start with the Field name?

vitarb (approved):
Approved assuming Cully's concerns are addressed before landing.

internal/client.go
alexshtin: It doesn't have too. But here it is easy to fix.

PR #423, opened 2021-04-30
Adds PermissionDenied service error to list of non-retryables

eritscher:
## What was changed:
Adds `serviceerror.PermissionDenied` to list indicating if an error should not be retried. 

## Why?
Currently, PermissionDenied errors will be retried until the context times out and clients will not see the true surfaced error.


mfateev:
Looks good, what stops from promoting it from draft status?

eritscher:
Thanks, just promoted. No reason, just a mistake on my part

mfateev (approved):
Thanks!

Issue #427, opened 2021-05-04
Test that an activity wasn't called


ovargasmahisoft:
For a workflow in which I have few branches, I'd like to assert that an activity wasn't called

I found this entry

https://community.temporal.io/t/workflow-unit-tests-questions-why-use-dummy-activities-how-to-ensure-activity-was-not-called-how-to-create-an-instance-of-workflow-context/756

## Expected Behavior

```go
env.OnActivity("activity a", mock.Anything).Panic(":( this activity shouldn't  be called")
```

Should return an error if the activity was called and the assert should **pass** if the activity wasn't called.

## Actual Behavior

The assert is failing if the activity isn't called.

## Steps to Reproduce the Problem

  1. Create a Workflow with at least one conditional branch
  2. Call a different activity on each branch
  3. Create a test for one the branches adding an OnActivity(...).Panic(...) to test that the activity in the other branch is not called
  4. Execute the test
  
Here is an example

```go
package dummy

import (
	"github.com/stretchr/testify/mock"
	"github.com/stretchr/testify/suite"
	"go.temporal.io/sdk/activity"
	"go.temporal.io/sdk/testsuite"
	"go.temporal.io/sdk/workflow"
	"testing"
	"time"
)

const (
	ActivityA = "Activity_a"
	ActivityB = "Activity_b"
	ActivityC = "Activity_c"
)

type Result struct {
	Value string
}

func Process(ctx workflow.Context, b bool) error {

	var r Result

	aCtx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
		ScheduleToStartTimeout: time.Second,
		ScheduleToCloseTimeout: time.Second,
	})

	if b {
		if err := workflow.ExecuteActivity(aCtx, ActivityA, "foo").Get(ctx, &r); err != nil {
			return err
		}
	} else {
		if err := workflow.ExecuteActivity(aCtx, ActivityB, "bar").Get(ctx, &r); err != nil {
			return err
		}
	}

	return workflow.ExecuteActivity(aCtx, ActivityC, r).Get(ctx, nil)
}

type Suite struct {
	suite.Suite
	testsuite.WorkflowTestSuite

	env      *testsuite.TestWorkflowEnvironment
}

func TestSuite(t *testing.T) {
	suite.Run(t, new(Suite))
}

func (s *Suite) AfterTest(suiteName, testName string) {
	s.env.AssertExpectations(s.T())
}

func (s *Suite) SetupTest() {
[...]

PR #428, opened 2021-05-05
Ability to verify when an activity should not be called

ovargasmahisoft:
## What was changed:

Added to the `MockCallWrapper` a way to specify that we expect Zero calls to the activity.

## Why?

This can be helpful to test workflows with branches.

## Checklist

1. Closes issue: #427 

2. How was this tested: Unit test added

3. Any docs updates needed?

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -449,8 +449,8 @@ func (c *MockCallWrapper) Times(i int) *MockCallWrapper {
 	return c
 }
 
-// Zero indicates that the mock should not be called.
-func (c *MockCallWrapper) Zero() *MockCallWrapper {
+// Never indicates that the mock should not be called.
+func (c *MockCallWrapper) Never() *MockCallWrapper {
 	c.call.Maybe()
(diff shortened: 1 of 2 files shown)


internal/workflow_testsuite.go
```diff
@@ -449,6 +449,13 @@ func (c *MockCallWrapper) Times(i int) *MockCallWrapper {
 	return c
 }
 
+// Zero indicates that the mock should not be called.
+func (c *MockCallWrapper) Zero() *MockCallWrapper {
```
Sushisource, line 453:
```suggestion
// Never indicates that the mock should not be called.
func (c *MockCallWrapper) Never() *MockCallWrapper {
```

I think never might be a slightly more intuitive name here, and is fairly common in mocking libs.

Sushisource (approved):
Thanks for the contribution! A minor wording suggestion.

internal/workflow_testsuite.go
ovargasmahisoft: Done!!

PR #430, opened 2021-05-05
Fixes leak if client is able to open gRPC connection to server, but health check subsequently fails.

mastermanu:
## What was changed:
During creation of the client, if we successfully create a gRPC connection, but cannot validate the health check, we error out without actually closing the gRPC connection. This results in a temporary leak. This PR fixes that leak.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -603,6 +603,9 @@ func NewClient(options ClientOptions) (Client, error) {
 	}
 
 	if err = checkHealth(connection, options.ConnectionOptions); err != nil {
+		if err := connection.Close(); err != nil {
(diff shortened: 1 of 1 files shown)


Sushisource (approved):

internal/client.go
```diff
@@ -603,6 +603,9 @@ func NewClient(options ClientOptions) (Client, error) {
 	}
 
 	if err = checkHealth(connection, options.ConnectionOptions); err != nil {
+		if err := connection.Close(); err != nil {
```
mattkim, line 606: nit: should we wrap error?

mattkim (approved):

Issue #431, opened 2021-05-06
ActivityEndToEndLatency Incorrectly Recorded


shannontan-bolt:
## Expected Behavior
Activity end to end latency is from when the activity task was scheduled to when the activity task response was completed.

## Actual Behavior
Activity end to end latency is incorrect.

```
activityMetricsScope.Timer(metrics.ActivityEndToEndLatency).Record(time.Since(activityTask.pollStartTime))
```

pollStartTime is defined here: https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_pollers.go#L825, but that's the start time of when the poll was started that received the activity task, not the start time of the actual activity task. 

It should line up with how ScheduleToStart is calculated, which uses the GetStartTime() method on the returned task: https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_pollers.go#L851

## Specifications

  - Version: 1.6.0
  - Platform: N/A

shannontan-bolt:
CC @mastermanu

vitarb:
This issue should be resolved as PR has been merged.

PR #434, opened 2021-05-06
Record End to End Latency Using Activity Task Start Time

shannontan-bolt:
## What was changed:
Change Activity End to End Latency to record from task start time.

## Why?
This is so that we record the time from when the activity first entered the queue and not when the poll started that grabbed this activity

## Checklist

1. Closes issue: https://github.com/temporalio/sdk-go/issues/431

2. How was this tested: N/A


shannontan-bolt:
CC @mastermanu

vitarb (approved):

Issue #424, opened 2021-04-30
Invocation of struct local activities is broken.


mfateev:
## Expected Behavior
```go
	var a *Activities

	var greetResult string
	err := workflow.ExecuteLocalActivity(ctx, a.GetGreeting).Get(ctx, &greetResult)
```
should work the same way it works for normal activities.

## Actual Behavior
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x18d07a4]

goroutine 70 [running]:
github.com/temporalio/samples-go/greetings.(*Activities).GetGreeting(...)
        /Users/maxim/temporal/temporal-go-samples/greetings/activities.go:17
reflect.Value.call(0x19454c0, 0xc000165910, 0x13, 0x1a72c0b, 0x4, 0x0, 0x0, 0x0, 0x203000, 0x203000, ...)
        /usr/local/go/src/reflect/value.go:476 +0x8c7
reflect.Value.Call(0x19454c0, 0xc000165910, 0x13, 0x0, 0x0, 0x0, 0x100f5f0, 0xc000588790, 0x10)
        /usr/local/go/src/reflect/value.go:337 +0xb9
go.temporal.io/sdk/internal.(*activityExecutor).executeWithActualArgsWithoutParseResult(0xc0005c2720, 0x1b801c0, 0xc000598ae0, 0x0, 0x0, 0x0, 0xc0005c09c0, 0x1b801c0, 0xc000598ae0)
        /Users/maxim/go/pkg/mod/go.temporal.io/sdk@v1.6.0/internal/internal_worker.go:830 +0x24a
go.temporal.io/sdk/internal.(*activityExecutor).ExecuteWithActualArgs(0xc0005c2720, 0x1b801c0, 0xc000598ae0, 0x0, 0x0, 0x0, 0x20a66c0, 0x202c7a9, 0xb)
        /Users/maxim/go/pkg/mod/go.temporal.io/sdk@v1.6.0/internal/internal_worker.go:804 +0x70
go.temporal.io/sdk/internal.(*localActivityTaskHandler).executeLocalActivityTask.func2(0x1b801c0, 0xc000598ae0, 0xc0004340e0, 0xc0005aaf00, 0xc0005c2720, 0xc00058c0d0, 0xc000588770, 0x1b88180, 0xc0005a2600, 0x2540be400, ...)
[...]

PR #436, opened 2021-05-08
Added ability to call local activity using nil struct pointer

mfateev:
## What was changed:
Added ability to call local activity using nil struct pointer:
```go
	var activities *Activities
	err := workflow.ExecuteLocalActivity(ctx, activities.Function).Get(ctx, nil)
```
## Why?
It is already supported for normal activities and allows using struct implementation for the local ones.

1. Closes issue: 
https://github.com/temporalio/sdk-go/issues/424

2. How was this tested:
Unit tests updated to cover the legacy and the new way to call local activities.

3. Any docs updates needed?
Yes. The Go SDK needs to be updated to demonstrate this pattern for both normal and local activities.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1388,6 +1388,8 @@ func getFunctionName(i interface{}) (name string, isMethod bool) {
 		return fullName, false
 	}
 	fullName := runtime.FuncForPC(reflect.ValueOf(i).Pointer()).Name()
+	// Full function name that has a struct pointer receiver has the following format
+	// <prefix>.(*<type>).<function>
 	isMethod = strings.ContainsAny(fullName, "*")
 	elements := strings.Split(fullName, ".")
 	shortName := elements[len(elements)-1]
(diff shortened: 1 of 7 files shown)


internal/internal_worker.go
```diff
@@ -1383,11 +1383,12 @@ func isError(inType reflect.Type) bool {
...
 }
 
-func getFunctionName(i interface{}) string {
+func getFunctionName(i interface{}) (name string, isMethod bool) {
 	if fullName, ok := i.(string); ok {
-		return fullName
+		return fullName, false
 	}
 	fullName := runtime.FuncForPC(reflect.ValueOf(i).Pointer()).Name()
+	isMethod = strings.ContainsAny(fullName, "*")
```
vitarb, line 1391: Would be nice to add a comment that explains what "*" means.

vitarb (approved):

internal/internal_worker.go
Sushisource: Oh man, this is how you did it? Can we guarantee that `*` can never appear in a function name except as part of a receiver pointer? This feels pretty icky, but if it's the only reason then I suppose it's reasonable.

Definitely needs the comment for sure, otherwise could be deeply confusing to future readers

PR #441, opened 2021-05-13
Fix local activity retry max attempts being off by two

Sushisource:
## What was changed
Two different bugs contributed to this:
* Local activity attempts start at 0
* Max attempts check was exclusive rather than inclusive

## Why?
Local activities retried more times than they should've been

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_task_handlers_test.go
+++ b/internal/internal_task_handlers_test.go
@@ -1255,10 +1255,12 @@ func (t *TaskHandlersTestSuite) TestLocalActivityRetry_WorkflowTaskHeartbeatFail
 	workflowTaskStartedEvent := createTestEventWorkflowTaskStarted(3)
 	now := time.Now()
 	workflowTaskStartedEvent.EventTime = &now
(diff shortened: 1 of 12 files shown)


internal/internal_task_handlers_test.go
```diff
@@ -1255,10 +1255,12 @@ func (t *TaskHandlersTestSuite) TestLocalActivityRetry_WorkflowTaskHeartbeatFail
 	workflowTaskStartedEvent := createTestEventWorkflowTaskStarted(3)
 	now := time.Now()
 	workflowTaskStartedEvent.EventTime = &now
+	// Task timeout must be larger than local activity retry backoff
+	wftTimeout := 2 * time.Second
```
mfateev, line 1259: I'm not sure I understand the change. It should work with the task timeout being shorter due to task heartbeat.

mfateev (approved):

PR #448, opened 2021-05-25
Filter out DeadlineExceeded error in retry

alexshtin:
Usually, after number of retry attempts, last attempt fails with `DeadlineExceeded` error. It is not informative and actual error reason is in the error occurred on previous attempt. Return this `lastErr` instead of `DeadlineExceeded` error provides a better user experience.


Sushisource (approved):
Sweet :+1: . A test that verifies the last error gets captured w/ a mock or something might be nice.

PR #453, opened 2021-05-27
Health check retry

### LINKED ISSUE #438: Client creation doesn't have any retries if health check intermittently fails
Health check code:

```
func checkHealth(connection grpc.ClientConnInterface, options ConnectionOptions) error {
	if options.DisableHealthCheck {
		return nil
	}
	healthClient := healthpb.NewHealthClient(connection)
	request := &healthpb.HealthCheckRequest{
		Service: healthCheckServiceName,
	}
	healthCheckTimeout := options.HealthCheckTimeout
	if healthCheckTimeout == 0 {
		healthCheckTimeout = defaultHealthCheckTimeout
	}
	ctx, cancel := context.WithTimeout(context.Background(), healthCheckTimeout)
	defer cancel()
	resp, err := healthClient.Check(ctx, request)
	if err != nil {
		return fmt.Errorf("health check error: %w", err)
	}
	if resp.Status != healthpb.HealthCheckResponse_SERVING {
		return fmt.Errorf("health check returned unhealthy status: %v", resp.Status)
	}
	return nil
}
```

User's only option is to disable the health check or add retries on their application layer. For checkHealth calls on create client, we may want more sophisticated logic, or at least the ability to configure something via ConnectionOptions

vitarb:
Resolves #438 by adding configurable healthcheck timeout for the client.
Default value is set to 10 seconds, which means that now client creation would block sending health check requests in case if server is not available.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -58,9 +58,9 @@ const (
 	// sessions in the workflow. The result will be a list of SessionInfo encoded in the EncodedValue.
 	QueryTypeOpenSessions string = "__open_sessions"
 
-	healthCheckServiceName     = "temporal.api.workflowservice.v1.WorkflowService"
-	defaultHealthCheckTimeout  = 5 * time.Second
-	defaultHealthCheckRetryFor = 10 * time.Second
+	healthCheckServiceName           = "temporal.api.workflowservice.v1.WorkflowService"
+	defaultHealthCheckAttemptTimeout = 5 * time.Second
+	defaultHealthCheckTimeout        = 10 * time.Second
 )
 
 type (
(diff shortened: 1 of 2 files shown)


internal/client.go
```diff
@@ -423,6 +425,10 @@ type (
 		// Default: 5s.
 		HealthCheckTimeout time.Duration
 
+		// HealthCheckRetryFor defines maximum amount of time that health check should be retried for.
+		// Defaults to 10s, once this timeout is reached error will be propagated to the client.
+		HealthCheckRetryFor time.Duration
```
mastermanu, line 430: [Nit] I think we need better name here. OR, HealthCheckTimeout should represent the overall time including all attempts, and we have the new field specify the PerAttempt timeout (or have that hard-coded). I can see how this might cause some weirdness with breaking changes though...

mastermanu (approved):

internal/client.go
vitarb: We can slightly turn it and add `HealthCheckAttemptTimeout`, which would allow us to use `HealthCheckTimeout` as an aggregate one, what do you think?

mfateev (approved):

PR #454, opened 2021-05-28
Propagate cancellation to the client if server is unavailable

### LINKED ISSUE #437: Cancel activity context on service not being reachable
## Expected Behavior
Activity context is canceled (Done channel is closed) if it cannot reach the service for the duration of the heartbeat timeout.

## Actual Behavior
The error is logged and a new heartbeat is issued after the heartbeat timeout. This can create problems as the activity cannot rely on being notified about issues after the heartbeat timeout interval.

## Relevant forum post

https://community.temporal.io/t/best-practices-for-long-running-activities/934/6

vitarb:
Fix #437 (minus tests) by making `Unavailable` and other transient errors being propagated as a cancelation to the client code.
I've double checked using cancellation sample that this is indeed an error returned when the server goes down.
Adding an integration test for this scenario is a little tricky though.
I've also changed `cancelHandler` to be invoked on all transient errors that's been retried for the duration of heartbeat timeout.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1670,7 +1670,7 @@ func (i *temporalInvoker) internalHeartBeat(ctx context.Context, details *common
 	ctx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	err := recordActivityHeartbeat(ctx, i.service, i.metricsScope, i.identity, i.taskToken, details)
+	err := recordActivityHeartbeat(ctx, i.service, i.identity, i.taskToken, details)
 
 	switch err.(type) {
 	case *CanceledError:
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -1670,15 +1670,15 @@ func (i *temporalInvoker) internalHeartBeat(ctx context.Context, details *common
...
+	err := recordActivityHeartbeat(ctx, i.service, i.identity, i.taskToken, details)
 
 	switch err.(type) {
 	case *CanceledError:
 		// We are asked to cancel. inform the activity about cancellation through context.
 		i.cancelHandler()
 		isActivityCanceled = true
 
-	case *serviceerror.NotFound, *serviceerror.NamespaceNotActive:
+	case *serviceerror.NotFound, *serviceerror.NamespaceNotActive, *serviceerror.Unavailable:
```
vitarb, line 1681: This is an actual change.

internal/internal_task_handlers.go
```diff
@@ -1683,10 +1683,19 @@ func (i *temporalInvoker) internalHeartBeat(ctx context.Context, details *common
 		// later when we have setter on cancel handler.
 		i.cancelHandler()
 		isActivityCanceled = true
+	case nil:
+		// No error, do nothing.
+	default:
+		// Transient errors are getting retried for the duration of the heartbeat timeout.
```
Sushisource, line 1689:
```suggestion
		// Transient errors are retried for the duration of the heartbeat timeout.
```

Sushisource (approved):

PR #455, opened 2021-06-03
Aggregated request metrics and retry mechanism refactoring

### LINKED ISSUE #442: Reported Metrics around availability include initial retry failures even if final call succeeds
## Expected Behavior
If, due to some transient issue, the SDK fails on its first attempt to call Temporal, but succeeds on a retry, the metric reports 1 failure and 1 success, which could be misleading. This is because the metric emission happens on an interceptor that is triggered per-request:

```
// NewScopeInterceptor creates new metrics scope interceptor.
func NewScopeInterceptor(defaultScope tally.Scope) grpc.UnaryClientInterceptor {
	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
		scope, ok := ctx.Value(ScopeContextKey).(tally.Scope)
		if !ok || scope == nil {
			scope = defaultScope
		}
		isLongPoll, ok := ctx.Value(LongPollContextKey).(bool)
		if !ok {
			isLongPoll = false
		}
		rs := newRequestScope(scope, method, isLongPoll)
		rs.recordStart()
		err := invoker(ctx, method, req, reply, cc, opts...)
		rs.recordEnd(err)
		return err
	}
}
```

Ideal behavior would be to differentiate "overall" SLA vs SLA including retry failures.

vitarb:
## What was changed
Adding an aggregated metric for RPC requests that excludes retry attempts.
In order to achieve this we had two options:
1. Report metrics from the backoff retry function or its caller.
2. Refactor existing retry mechanism and move retry logic into the interceptor, this way we can have interceptors reporting metrics before and after retry interceptor in the chain.

This PR follows 2nd approach, as approach number 1 is messy and requires reporting similar metrics from the different levels in the call stack, also those places don't always have easy access to the metric scope.
In order to achieve what we need, I've used [grpc-retry](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/retry) adding retry interceptor as well as additional interceptor which creates retry configuration. Now instead of calling `backoff.Retry`, client code is expected to simply populate retry information into grpc context as a key/value pair and interceptor chain should translate it into retry options, which would trigger standard grpc retry mechanism defined in the `grpc_retry.UnaryClientInterceptor()`.
Also introduced a new concept, called `TrafficController` that can be provided in client options and will be called from the last interceptor in the grpc interceptor chain providing a way to approve/deny call to the server. This feature is used for testing grpc retries as we want to create artificial failures on some calls.

## Why?
We need meaningful aggregated API level metrics that exclude transient failures and retries.

## Checklist

1. Closes #442
2. How was this tested: not well enough yet, hence it's a draft.
3. Any docs updates needed? Documenting new metrics would be nice.


internal/grpc_dialer.go
```diff
@@ -109,7 +111,18 @@ func dial(params dialParameters) (*grpc.ClientConn, error) {
...
+		// Report aggregated metrics for the call, this is done outside of the retry loop.
+		metrics.NewRPCMetricsInterceptor(metricScope, "_total"),
+		// By default the grpc retry interceptor *is disabled*, preventing accidental use of retries.
+		// We add call options for retry configuration based on the values present in the context.
+		retry.NewRetryOptionsInterceptor(),
+		// Performs retries *IF* retry options are set for the call.
+		grpc_retry.UnaryClientInterceptor(),
+		// Report metrics for every call made to the server.
+		metrics.NewRPCMetricsInterceptor(metricScope, ""),
+	}
```
vitarb, line 125: This change in the interceptor chain is at the core of this PR.

internal/common/retry/interceptor.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+const (
+	// ConfigKey context key for GrpcRetryConfig
+	ConfigKey = contextKey("RetryConfig")
+)
+
+// NewRetryOptionsInterceptor creates a new gRPC interceptor that populates retry options for each call based on values
+// provided in the context.
+func NewRetryOptionsInterceptor() grpc.UnaryClientInterceptor {
```
vitarb, line 115: This interceptor creates retry configuration for the `grpc_retry.UnaryClientInterceptor()` that is down in the chain.

internal/internal_task_pollers.go
```diff
@@ -368,58 +368,52 @@ func (wtp *workflowTaskPoller) RespondTaskCompletedWithMetrics(
...
-						wtp.logger.Debug("RespondQueryTaskCompleted failed.", tagError, err1)
-					})
-				}
-			default:
-				// should not happen
-				panic("unknown request type from ProcessWorkflowTask()")
+	grpcCtx, cancel := newGRPCContext(ctx, grpcMetricsScope(
+		metrics.GetMetricsScopeForRPC(wtp.metricsScope, task.GetWorkflowType().GetName(),
+			metrics.NoneTagValue, metrics.NoneTagValue)),
+		grpcContextValue(retry.ConfigKey, createDynamicServiceRetryPolicy(ctx).IntoGrpcRetryConfig()))
```
vitarb, line 374: This is how we specify retry configuration per grpc call now.

internal/common/backoff/retrypolicy.go
```diff
@@ -28,25 +28,19 @@ import (
...
-	defaultMaximumAttempts    = noMaximumAttempts
+	done time.Duration = -1
 )
 
 type (
 	// RetryPolicy is the API which needs to be implemented by various retry policy implementations
 	RetryPolicy interface {
-		ComputeNextDelay(elapsedTime time.Duration, numAttempts int) time.Duration
+		ComputeNextDelay(elapsedTime time.Duration, attempt int) time.Duration
+		IntoGrpcRetryConfig() *retry.GrpcRetryConfig
```
Ardagan, line 43: GrpcRetryConfig()

internal/common/backoff/retrypolicy.go
```diff
@@ -136,27 +130,27 @@ func (p *ExponentialRetryPolicy) SetMaximumAttempts(maximumAttempts int) {
 }
 
 // ComputeNextDelay returns the next delay interval.  This is used by Retrier to delay calling the operation again
-func (p *ExponentialRetryPolicy) ComputeNextDelay(elapsedTime time.Duration, numAttempts int) time.Duration {
+func (p *ExponentialRetryPolicy) ComputeNextDelay(elapsedTime time.Duration, attempt int) time.Duration {
 	// Check to see if we ran out of maximum number of attempts
-	if p.maximumAttempts != noMaximumAttempts && numAttempts >= p.maximumAttempts {
+	if p.maximumAttempts != retry.NoMaximumAttempts && attempt >= p.maximumAttempts {
```
Ardagan, line 135: Should these be IgnoreMaximumAttempts?

internal/common/backoff/retrypolicy.go
```diff
@@ -28,25 +28,19 @@ import (
 	"math"
 	"math/rand"
 	"time"
+
+	"go.temporal.io/sdk/internal/common/retry"
 )
 
 const (
-	// NoInterval represents Maximim interval
-	NoInterval                      = 0
```
Ardagan, line 35: Technically, NoInterval and DefaultBackoffCoefficient are accessible by the user. Can this be a breaking change?

internal/common/metrics/interceptor.go
```diff
@@ -31,8 +31,8 @@ import (
 	"google.golang.org/grpc"
 )
 
-// NewScopeInterceptor creates new metrics scope interceptor.
-func NewScopeInterceptor(defaultScope tally.Scope) grpc.UnaryClientInterceptor {
+// NewRPCMetricsInterceptor creates new metrics scope interceptor.
+func NewRPCMetricsInterceptor(defaultScope tally.Scope, metricSuffix string) grpc.UnaryClientInterceptor {
```
Ardagan, line 35: NewGRPCMetricsInterceptor

internal/grpc_dialer.go
```diff
@@ -109,7 +111,18 @@ func dial(params dialParameters) (*grpc.ClientConn, error) {
 }
 
 func requiredInterceptors(metricScope tally.Scope, headersProvider HeadersProvider) []grpc.UnaryClientInterceptor {
-	interceptors := []grpc.UnaryClientInterceptor{metrics.NewScopeInterceptor(metricScope), errorInterceptor}
+	interceptors := []grpc.UnaryClientInterceptor{
+		errorInterceptor,
+		// Report aggregated metrics for the call, this is done outside of the retry loop.
+		metrics.NewRPCMetricsInterceptor(metricScope, "_total"),
```
Ardagan, line 117: Better create constant for "_total".
I would be careful using "_total" here since it implies aggregation over arbitrary subset of tags and it can become obscure if reused in a different location. For example, we already reuse "_total" to aggregate values across multiple namespaces on server.

In this case, my initial hunch was that we aggregate some metrics across all the different calls, while regular metrics will be per-something (like per API name).

It may be better to use empty suffix here, but add "_per_retry" on L124.

vitarb: Hm, swapping the two is an interesting idea. Let's do that. Then all users that are relying on existing metrics would get correct ones automatically.

internal/common/retry/interceptor.go
```diff
@@ -0,0 +1,146 @@
...
+
+	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
+	"github.com/grpc-ecosystem/go-grpc-middleware/util/backoffutils"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+)
+
+const (
+	// NoMaximumAttempts is a value that maximumAttempts can be set to if total number of attempts should be unlimited.
+	NoMaximumAttempts = 0
```
Sushisource, line 40: This is pretty confusing. It's saying "If you specify maximum attempts equal to anything that is the same as this constant, then max attempts is ignored?" 

A better name might be `IgnoreMaximumAttemptsSigil` with the docstring like "When maximum attempts is set to this special value, then the number of attempts is unlimited"

Same for `NoInterval`

But, the other problem here is negative values are nonsense. So perhaps this should be a `>=` check.

internal/common/backoff/retrypolicy.go
Sushisource: I see Rust naming schemes are getting to you :smile:  -- this would be `AsGrpcRetryConfig` in Rust-speak anyway, since the `RetryPolicy` struct isn't destroyed in the process. That name would be reasonable too -- but I agree the `Into` is a little confusing here.

internal/common/retry/interceptor.go
Sushisource: I also don't think these docstrings need to repeat the field name (same with a lot of the other ones in this file)

internal/grpc_dialer.go
Sushisource: Would it be possible to create an interceptor that just combines these four post-error interceptors and use just that here? That would help ensure this ordering is maintained and provide a nice encapsulation.

internal/common/retry/interceptor.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+			opts = append(opts, grpc_retry.WithBackoff(func(attempt uint) time.Duration {
+				next := float64(rc.initialInterval) * math.Pow(rc.backoffCoefficient, float64(attempt))
+				if rc.maximumInterval != NoInterval {
+					next = math.Min(next, float64(rc.maximumInterval))
+				}
+				return backoffutils.JitterUp(time.Duration(next), rc.jitter)
+			}))
+			if rc.maximumAttempts != NoMaximumAttempts {
+				grpc_retry.WithMax(uint(rc.maximumAttempts))
```
Sushisource, line 131: Maybe I'm reading this library wrong but this statement doesn't seem like it does anything? (Same with `WithCodes` below)? They return a `CallOption` but we're not using that? Seems like they need to append to `opts` like above?

internal/internal_task_pollers.go
```diff
@@ -964,32 +944,23 @@ func reportActivityCompleteByID(ctx context.Context, service workflowservice.Wor
...
-		reportErr = backoff.Retry(ctx,
-			func() error {
-				grpcCtx, cancel := newGRPCContext(ctx, grpcMetricsScope(rpcScope))
-				defer cancel()
-
-				_, err := service.RespondActivityTaskFailedById(grpcCtx, request)
-				return err
-			}, createDynamicServiceRetryPolicy(ctx), isServiceTransientError)
+		grpcCtx, cancel := newGRPCContext(ctx, grpcMetricsScope(rpcScope),
+			grpcContextValue(retry.ConfigKey, createDynamicServiceRetryPolicy(ctx).IntoGrpcRetryConfig()))
```
Sushisource, line 954: This line gets repeated a lot and could probably be factored out into a simpler one-function call.

Sushisource (commented):
Overall approach makes sense to me. A few mostly small things to address

internal/common/backoff/retrypolicy.go
vitarb: Yep, ideally you want just plain `Into`, but unfortunately go can't do that generically. Renamed to `GrpcRetryConfig`.

internal/common/backoff/retrypolicy.go
vitarb: I think the name is okay as it is, it's also consistent with other similar names in this package.

internal/common/backoff/retrypolicy.go
vitarb: We don't provide any guarantees about backwards compatibility within internal package, so this should be fine.

internal/common/metrics/interceptor.go
vitarb: Done

internal/common/backoff/retrypolicy.go
vitarb: Let's maybe call it `UnlimitedMaximumAttempts`?

internal/common/retry/interceptor.go
vitarb: As suggested above, let's rename it to `UnlimitedMaximumAttempts`. Updated docstrings too.
Go lint actually requires you to repeat the name, kind of silly, but it's a hard requirement, which a lot of people don't like.

vitarb: Also I was considering making default values -1 instead of 0, which would make more sense, as 0 technically is a valid value, but I didn't want to introduce regression into existing retrier, that also uses these constants.

internal/internal_task_pollers.go
vitarb: It's actually repeated a lot in slightly different variations. What we can do is extract `grpcContextValue(retry.ConfigKey, createDynamicServiceRetryPolicy(ctx).IntoGrpcRetryConfig())` into a helper function and call it something like defaultGrpcRetryContextValue`

internal/common/retry/interceptor.go
vitarb: Already fixed.

internal/grpc_dialer.go
vitarb: We could, but wouldn't it just be hiding details and ultimately moving the list from one place into another, where things can also be reordered?

Sushisource: I suppose that's true, the benefit is just that it groups this stuff that "needs to go together", but, not a big deal either way

internal/internal_task_pollers.go
Sushisource: Yeah, I think that would be useful

Sushisource (approved):

alexshtin (changes requested):
I am blocking this until we make a decision on retryable errors/error codes.

internal/internal_task_pollers.go
```diff
@@ -925,32 +912,22 @@ func reportActivityComplete(ctx context.Context, service workflowservice.Workflo
...
 	case *workflowservice.RespondActivityTaskCanceledRequest:
-		reportErr = backoff.Retry(ctx,
-			func() error {
-				grpcCtx, cancel := newGRPCContext(ctx, grpcMetricsScope(rpcScope))
-				defer cancel()
-
-				_, err := service.RespondActivityTaskCanceled(grpcCtx, request)
-				return err
-			}, createDynamicServiceRetryPolicy(ctx), isServiceTransientError)
+		grpcCtx, cancel := newGRPCContext(ctx, grpcMetricsScope(rpcScope),
```
alexshtin, line 915: This is beautiful! What blocks you from removing `isServiceTransientError` func? We have this logic in two places now.

internal/common/retry/interceptor.go
```diff
@@ -0,0 +1,156 @@
...
+	GrpcRetryConfig struct {
+		initialInterval    time.Duration
+		backoffCoefficient float64
+		maximumInterval    time.Duration
+		expirationInterval time.Duration
+		jitter             float64
+		maximumAttempts    int
+	}
+
+	contextKey string
```
alexshtin, line 67:
```suggestion
	contextKey struct{}
```

internal/common/retry/interceptor.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+func (ck contextKey) String() string {
+	return string(ck)
+}
```
alexshtin, line 72:
```suggestion
```

internal/common/retry/interceptor.go
```diff
@@ -0,0 +1,156 @@
...
+		expirationInterval: DefaultExpirationInterval,
+		jitter:             DefaultJitter,
+		maximumAttempts:    DefaultMaximumAttempts,
+	}
+}
+
+const (
+	// ConfigKey context key for GrpcRetryConfig
+	ConfigKey = contextKey("RetryConfig")
+)
```
alexshtin, line 116:
```suggestion
varr (
	retryConfigContextKey contextKey
)
```

alexshtin: This is encapsulation mehanizm. No one outside of this package won't be able to get value from context because key is unexported value of unexported type.

internal/common/retry/interceptor.go
```diff
@@ -0,0 +1,156 @@
...
+			} else {
+				opts = append(opts, grpc_retry.WithMax(math.MaxUint32))
+			}
+			// We have to deal with plain gRPC error codes instead of service errors here as actual error translation
+			// happens after invoker is called below and invoker must have correct retry options right away in order to
+			// supply them to the gRPC retrier.
+			// Retrying only transient error codes, following statuses are excluded: Ok, AlreadyExists, FailedPrecondition,
+			// InvalidArgument, NotFound, PermissionDenied, Unauthenticated, Unimplemented.
+			opts = append(opts, grpc_retry.WithCodes(codes.Aborted, codes.Canceled, codes.DataLoss, codes.DeadlineExceeded, codes.Internal,
+				codes.OutOfRange, codes.ResourceExhausted, codes.Unavailable, codes.Unknown))
```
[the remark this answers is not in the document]
vitarb: Let's remove them, I thought DataLoss could be transport error, but it looks like gRPC would never generate it by itself, so we should probably never retry it.

internal/internal_task_pollers.go
vitarb: There are 2 more places that require a little bit of thinking. Let me try to address them too.

vitarb: There were couple of places with custom error handling, that I left for later. Let me try clean them up too.

internal/common/retry/interceptor.go
vitarb: I would prefer to keep string value for nicer debugging.

internal/common/retry/interceptor.go
vitarb: Ok

vitarb:
Feedback has been addressed, I'm going to land this change, please let me know if you have any follow up questions.

PR #466, opened 2021-06-16
Add max message size for grpc messages

vitarb:
Adding max message size for grpc send/receive to be 64mb instead of default 4mb.
This should prevent users from running into a situation where too large payload or memo can result in a workflow being stuck.

Also made ResourceExhausted non retryable:
```
	// ResourceExhausted indicates some resource has been exhausted, perhaps
	// a per-user quota, or perhaps the entire file system is out of space.
	//
	// This error code will be generated by the gRPC framework in
	// out-of-memory and server overload situations, or when a message is
	// larger than the configured maximum size.
```

This allows us to see the error associated with large payload size e.g.:
`        	            	trying to send message larger than max (1048791 vs. 10240)`
Instead of a context timeout.


Sushisource (approved):
Did we wanna make this a configurable option?

internal/common/retry/interceptor.go
```diff
@@ -117,8 +117,7 @@ var (
 	// gRPC response codes that represent retryable errors.
 	// The following status codes are never generated by the library:
 	//    INVALID_ARGUMENT, NOT_FOUND, ALREADY_EXISTS, FAILED_PRECONDITION, ABORTED, OUT_OF_RANGE, DATA_LOSS
-	retryableCodes = []codes.Code{codes.Aborted, codes.DeadlineExceeded, codes.Internal,
-		codes.ResourceExhausted, codes.Unavailable, codes.Unknown}
+	retryableCodes = []codes.Code{codes.Aborted, codes.DeadlineExceeded, codes.Internal, codes.Unavailable, codes.Unknown}
```
samarabbas, line 120: @alexshtin can you confirm that we are not using ResourceExhausted error for server side throttling?  If that is the case than it looks good.

vitarb: This error is particularly messy. We do use it for throttling and for some non-retryable errors like `ErrNamespaceCountLimitServerBusy`, `ErrSizeExceedsLimit`.
My thinking was that in case of QPS errors, we probably don't want to bombard server with retries (since our default policy is pretty aggressive 20ms with 1.2x backoff), because in this case we would probably be getting the same error again. I'm not sure if we should keep this one retryable or not, let's have a discussion.
I think it's probably better to leave it as it is, the tradeoff is that we won't see a clear message in case if message size exceeds limits, but default limit is very high so nobody should be hitting it anyways. On the other hand removing retry may result in potential issues.

internal/client.go
```diff
@@ -457,6 +457,9 @@ type (
 		// when there are no active RPCs, Time and Timeout will be ignored and no
 		// keepalive pings will be sent.
 		KeepAlivePermitWithoutStream bool
+
+		// MaxPayloadSize is a number of bytes that gRPC would allow to travel to and from server. Defaults to 64 MB.
```
wxing1292, line 461: server will also do input size validation, so `travel to` is not really necessary
the main point is to allow SDK to successfully replay the workflow history

internal/grpc_dialer.go
```diff
@@ -98,6 +107,8 @@ func dial(params dialParameters) (*grpc.ClientConn, error) {
 	}
 
 	opts = append(opts, securityOptions...)
+	opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallSendMsgSize(maxPayloadSize)))
```
wxing1292, line 110: again, no need to override the send limitation

wxing1292 (approved):

internal/grpc_dialer.go
vitarb: I don't have a strong opinion on this one, did it for consistency reasons.

PR #467, opened 2021-06-16
Return activity failure when activity is not registered

### LINKED ISSUE #465: Activity stays in started state when ActivityType is not registered
## Expected Behavior
Worker should fail the activity immediately with "Activity type not registered" error if it picks up the task from task queue and not able to find registered type.

## Actual Behavior
If activity start_to_close timeout is large than activity stays stuck in the started state.  Basically worker picks up the task and silently fails it without reporting anything back to the service.

  1.  Start workflow which schedules an activity where type is not registered
  1.  Start worker which polls on the task queue
  1.  Worker picks up the activity task and fails silently

## Specifications

  - Version:
  - Platform:

vitarb:
## What was changed
Failing an activity in case if activity type is not registered. Also added a new metric `unregistered_activity_invocation` that indicates number of times activity execution has been attempted on unregistered activity type.

## Why?
Currently we would simply log an error and do nothing, this result in activity appear to be stuck and not retried until it gets timed out.

1. Closes #465

2. How was this tested:
Unit tests, integration tests.


internal/common/metrics/constants.go
```diff
@@ -42,12 +42,13 @@ const (
...
-	ActivityPollNoTaskCounter      = TemporalMetricsPrefix + "activity_poll_no_task"
-	ActivityScheduleToStartLatency = TemporalMetricsPrefix + "activity_schedule_to_start_latency"
-	ActivityExecutionFailedCounter = TemporalMetricsPrefix + "activity_execution_failed"
-	ActivityExecutionLatency       = TemporalMetricsPrefix + "activity_execution_latency"
-	ActivityEndToEndLatency        = TemporalMetricsPrefix + "activity_endtoend_latency"
-	ActivityTaskErrorCounter       = TemporalMetricsPrefix + "activity_task_error"
+	ActivityPollNoTaskCounter             = TemporalMetricsPrefix + "activity_poll_no_task"
+	ActivityScheduleToStartLatency        = TemporalMetricsPrefix + "activity_schedule_to_start_latency"
+	ActivityExecutionFailedCounter        = TemporalMetricsPrefix + "activity_execution_failed"
+	UnregisteredActivityInvocationCounter = TemporalMetricsPrefix + "unregistered_activity_invocation"
```
samarabbas, line 48: This is great idea.  Please make sure we have similar metric on Java SDK.

internal/error.go
```diff
@@ -222,6 +222,12 @@ type (
 		cause        error
 	}
 
+	// ActivityNotRegisteredError is returned if worker doesn't support activity type.
+	ActivityNotRegisteredError struct {
+		activityType   string
+		supportedTypes string
```
wxing1292, line 228: `supportedTypes string` -> `supportedTypes []string `?

internal/error.go
```diff
@@ -222,6 +222,12 @@ type (
 		cause        error
 	}
 
+	// ActivityNotRegisteredError is returned if worker doesn't support activity type.
+	ActivityNotRegisteredError struct {
```
samarabbas, line 226: Should we just use `ApplicationError` directly instead of creating a new error type.  I think there are more locations in the code where conversions to Error and Failures are happening.  Having `ActivityNotRegisteredError` as a top level might cause issues in other places.  For instance how do you control `IsRetry` behavior for this error class?

internal/workflow_testsuite_test.go
```diff
@@ -123,7 +123,7 @@ func TestUnregisteredActivity(t *testing.T) {
 	require.True(t, errors.As(err, &activityErr))
 
 	err = errors.Unwrap(activityErr)
-	var err1 *PanicError
```
samarabbas, line 126: Any custom error results in being wrapped up within a `PanicError`?  I'm surprised to see this as we were seeing different behavior in the case of real invocation where the activity fails silently and workflow gets stuck.  Looks like the behavior in the case of testEnv differs from real execution.

vitarb: Application error [is the default](https://github.com/temporalio/sdk-go/blob/a39bbe82b2bab1281d24d7f18a93a9813e0e622b/internal/error.go#L889).

internal/error.go
vitarb: It's an internal error that does not get exposed to the user, it will get translated to the `ApplicationError` down [the line](https://github.com/temporalio/sdk-go/blob/a39bbe82b2bab1281d24d7f18a93a9813e0e622b/internal/error.go#L889).

internal/error.go
vitarb: Done

Sushisource (approved):

mfateev (approved):

vitarb:
I'm going to land this PR, please let me know if you have further suggestions and we can adjust it afterwards.

PR #471, opened 2021-06-18
Add ParentClosePolicy to unit test

### LINKED ISSUE #470: Test to support ParentClosePolicy
Test framework does not respect ParentClosePolicy set on child workflow. For example, if set child workflow's ParentClosePolicy to PARENT_CLOSE_POLICY_ABANDON, which means when parent workflow closed, its child workflow should continue executing. 

The unit test framework should respect ParentClosePolicy. 

**Additional context**
Reported by user:
https://community.temporal.io/t/best-way-to-create-an-async-child-workflow/114/6

yiminc:
## What was changed
Add ParentClosePolicy to unit test

## Why?
Properly support ParentClosePolicy in unit test.

## Checklist

1. Closes #470 

2. How was this tested:
New unit tests

3. Any docs updates needed?

No.

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -190,13 +190,14 @@ type (
 		queryHandler          func(string, *commonpb.Payloads) (*commonpb.Payloads, error)
 		startedHandler        func(r WorkflowExecution, e error)
 
-		isTestCompleted bool
-		testResult      converter.EncodedValue
-		testError       error
-		doneChannel     chan struct{}
-		workerOptions   WorkerOptions
-		dataConverter   converter.DataConverter
-		runTimeout      time.Duration
+		isWorkflowCompleted bool
+		testResult          converter.EncodedValue
+		testError           error
+		doneChannel         chan struct{}
+		doneChannelClosed   bool
+		workerOptions       WorkerOptions
+		dataConverter       converter.DataConverter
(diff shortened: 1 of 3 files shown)


vitarb (approved):

internal/internal_workflow_testsuite.go
```diff
@@ -888,6 +915,21 @@ func (env *testWorkflowEnvironmentImpl) Complete(result *commonpb.Payloads, err
...
+	// cancel child workflows if their ParentClosePolicy request so.
+	env.maybeCancelChildWorkflows()
+}
+
+func (env *testWorkflowEnvironmentImpl) maybeCancelChildWorkflows() {
+	for _, handle := range env.runningWorkflows {
+		if handle.env.parentEnv != nil &&
+			env.workflowInfo.WorkflowExecution.ID == handle.env.parentEnv.workflowInfo.WorkflowExecution.ID {
+			// current env is parent workflow of handle's workflow
+			if handle.params.ParentClosePolicy == enumspb.PARENT_CLOSE_POLICY_REQUEST_CANCEL {
```
mfateev, line 928: What about TERMINATE? I know it doesn't make sense for the root parent workflow, but what about the of the child of the root?

yiminc: Good point. We need to handle that as well. Will do in a follow up PR.

PR #476, opened 2021-06-22
Make deadlock detection timeout configurable

### LINKED ISSUE #404: Setting deadlock detection timeouts?
We had code that was calling `rsa.GenerateKey(rand.Reader, 4096)` in a `workflow.SideEffect` call. Generating a 4096 RSA key sometimes takes >1 second, so the deadlock detection kills the workflow and starts it over.

It would be great to have a few functions to give hints to the deadlock detection so that I could do something like this before calling an API that takes a bit longer than a second:
```go
  ctx.SetDeadlockTimeout(seconds)
  defer ctx.ResetDeadlockTimeout()
  rsa.GenerateKey(rand.Reader, 4096)
```

Similarly, something like `ctx.Alive()` might be helpful in loops.

Currently we just use TEMPORAL_DEBUG for the service that makes this `rsa.GenerateKey` call... obviously this is not ideal.

vitarb:
## What was changed
Made deadlock detection timeout configurable via WorkerOptions. Default value hasn't changed and is equal to 1 second. `TEMPORAL_DEBUG` environment variable can still be used and will set unlimited deadlock detection timeout IF value is not provided in WorkerOptions.

## Why?
Initially we thought that there were no valid use cases when workflow tasks could take more than 1 second to run. However some users perform heavy, yet deterministic computations which may sometimes take longer. Making this timeout configurable allows for more flexibility on the user end.

## Checklist

1. Closes #404

2. How was this tested:
Unit tests

### DIFF
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -116,7 +116,7 @@ type (
 		// Application level code must be executed from this function only.
 		// Execute call as well as callbacks called from WorkflowEnvironment functions can only schedule callbacks
 		// which can be executed from OnWorkflowTaskStarted().
-		OnWorkflowTaskStarted()
+		OnWorkflowTaskStarted(timeout time.Duration)
 		// StackTrace of all coroutines owned by the Dispatcher instance.
 		StackTrace() string
 		Close()
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -116,7 +116,7 @@ type (
 		// Application level code must be executed from this function only.
 		// Execute call as well as callbacks called from WorkflowEnvironment functions can only schedule callbacks
 		// which can be executed from OnWorkflowTaskStarted().
-		OnWorkflowTaskStarted(timeout time.Duration)
+		OnWorkflowTaskStarted(deadlockDetectionTimeout time.Duration)
 		// StackTrace of all coroutines owned by the Dispatcher instance.
 		StackTrace() string
 		Close()
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -83,7 +83,7 @@ type (
 	dispatcher interface {
 		// ExecuteUntilAllBlocked executes coroutines one by one in deterministic order
 		// until all of them are completed or blocked on Channel or Selector or timeout is reached.
-		ExecuteUntilAllBlocked(timeout time.Duration) (err error)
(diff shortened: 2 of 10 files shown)


internal/internal_worker_base.go
```diff
@@ -116,7 +116,7 @@ type (
 		// Application level code must be executed from this function only.
 		// Execute call as well as callbacks called from WorkflowEnvironment functions can only schedule callbacks
 		// which can be executed from OnWorkflowTaskStarted().
-		OnWorkflowTaskStarted()
+		OnWorkflowTaskStarted(timeout time.Duration)
```
mfateev, line 119: rename to deadlockDetectionTimeout

internal/internal_workflow.go
```diff
@@ -84,8 +82,8 @@ type (
 	// Dispatcher is a container of a set of coroutines.
 	dispatcher interface {
 		// ExecuteUntilAllBlocked executes coroutines one by one in deterministic order
-		// until all of them are completed or blocked on Channel or Selector
-		ExecuteUntilAllBlocked() (err error)
+		// until all of them are completed or blocked on Channel or Selector or timeout is reached.
+		ExecuteUntilAllBlocked(timeout time.Duration) (err error)
```
mfateev, line 86: ditto

internal/internal_workflow.go
```diff
@@ -862,16 +859,21 @@ func (s *coroutineState) unblocked() {
...
 	s.unblock <- func(status string, stackDepth int) bool {
 		return false // unblock
 	}
 
-	deadlockDetectorTimeout := time.Second
-	if debugMode {
-		deadlockDetectorTimeout = math.MaxInt64
+	// Defaults are populated in the worker options during worker startup, but test environment
+	// may have no default value for the deadlock detection timeout, so we also need to set it here for
+	// backwards compatibility.
```
mfateev, line 869: This looks like a hack to me. Why to not initialize it in the test environment instead?

vitarb: Users might be overriding worker options after creation of the test environment and it would break as deadlock detection timeout would default to 0 in that case.

internal/internal_worker_base.go
vitarb: This was actually intentional, as I think timeout makes perfect sense in this context. However I'm fine changing it.

mfateev (approved):

Issue #483, opened 2021-07-02
Propagation of cancellations to child workflows can cause state machine panic


Sushisource:
## Expected Behavior
Child workflows all cancel successfully 

## Actual Behavior
One or more children can fail to terminate w/ `"BadRequestCancelActivityAttributes: invalid history builder state for action: add-activitytask-cancel-requested-event"`

## Steps to Reproduce the Problem
See https://community.temporal.io/t/error-in-child-workflow-when-parent-is-cancelled/2492

## Specifications
  - Version: 1.7.0
  - Platform:

It looks like this could be related to https://github.com/temporalio/sdk-go/issues/481 which produces the same error. Ideally both can be fixed at once.

PR #485, opened 2021-07-08
Do not allow starting child workflows from an already-cancelled context

Sushisource:
## What was changed
Previously, calls to start a child workflow using an already-cancelled context would still start the child and then immediately cancel it. Now, the child isn't started at all and the future immediately resolves w/ cancelled error.

## Why?
Starting child workflows with a cancelled context is a nonsense operation

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/483

2. How was this tested:
Added UT and IT

3. Any docs updates needed?

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -735,8 +735,8 @@ func (wc *workflowEnvironmentInterceptor) ExecuteChildWorkflow(ctx Context, chil
 		executionFuture:  executionFuture.(*futureImpl),
 	}
 
-	// Immediately return if the context is already cancelled without spawning the child workflow
-	if ctx.Err() == ErrCanceled {
+	// Immediately return if the context has an error without spawning the child workflow
+	if ctx.Err() != nil {
 		executionSettable.Set(nil, ctx.Err())
 		mainSettable.Set(nil, ctx.Err())
 		return result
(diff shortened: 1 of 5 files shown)


internal/workflow.go
```diff
@@ -734,6 +734,14 @@ func (wc *workflowEnvironmentInterceptor) ExecuteChildWorkflow(ctx Context, chil
 		decodeFutureImpl: mainFuture.(*decodeFutureImpl),
 		executionFuture:  executionFuture.(*futureImpl),
 	}
+
+	// Immediately return if the context is already cancelled without spawning the child workflow
+	if ctx.Err() == ErrCanceled {
```
yiminc, line 739: would check not nil be enough? I think we don't want schedule child workflow when ctx.Err() return any error.

Sushisource: That's a good point. Changed.

test/integration_test.go
```diff
@@ -360,6 +360,14 @@ func (ts *IntegrationTestSuite) TestCascadingCancellation() {
 	ts.NotNil(run)
 	ts.NoError(err)
 
+	// Need to give workflow time to start its child
+	for {
+		_, err := ts.client.DescribeWorkflowExecution(ctx, childWorkflowID, "")
+		if err == nil {
+			break
+		}
```
yiminc, line 368: Is it possible that this would keep the for loop for ever?

yiminc (approved):

test/integration_test.go
Sushisource: Yeah, I'll select against a timeout and fail the test in case

vitarb (approved):

PR #487, opened 2021-07-15
use GetScheduledTime to calculate ActivityEndToEndLatency

mattkim:
## What was changed
use GetScheduledTime to calculate ActivityEndToEndLatency

## Why?
Currently ActivityEndToEndLatency is referencing activity started as the start time, which does not include the time scheduled.

## Checklist

1. Closes

2. How was this tested:
unit tests

3. Any docs updates needed?
no

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -910,7 +910,7 @@ func (atp *activityTaskPoller) ProcessTask(task interface{}) error {
 		return reportErr
 	}
 
-	activityMetricsScope.Timer(metrics.ActivityEndToEndLatency).Record(time.Since(activityTask.pollStartTime))
+	activityMetricsScope.Timer(metrics.ActivityEndToEndLatency).Record(time.Since(common.TimeValue(activityTask.task.GetScheduledTime())))
 	return nil
 }
 


internal/internal_task_pollers.go
```diff
@@ -910,7 +910,7 @@ func (atp *activityTaskPoller) ProcessTask(task interface{}) error {
 		return reportErr
 	}
 
-	activityMetricsScope.Timer(metrics.ActivityEndToEndLatency).Record(time.Since(activityTask.pollStartTime))
+	activityMetricsScope.Timer(metrics.ActivityEndToEndLatency).Record(time.Since(common.TimeValue(activityTask.task.GetScheduledTime())))
```
mattkim, line 913: One question I have is if we should use "GetCurrentAttemptScheduledTime" instead?

underrun: maybe we want both?

internal/internal_task_pollers.go
```diff
@@ -910,7 +910,7 @@ func (atp *activityTaskPoller) ProcessTask(task interface{}) error {
 		return reportErr
 	}
 
-	activityMetricsScope.Timer(metrics.ActivityEndToEndLatency).Record(time.Since(activityTask.pollStartTime))
```
mattkim, line 913: Also, this does not seem to be the latest change. Not sure why it's shown this way in the PR.

mattkim:
Closing this for now, will sync with Bolt first to clarify the definition of this alert, and see what requirements they are looking for.

PR #493, opened 2021-07-20
Prevent panic from reflect

laniehei:
Prevent panic from reflect Interface() method. Prevents the symptom of the panic. This bug should probably handled further upstream.


Sushisource (approved):
LGTM. Maybe worth adding a small unit test.

Issue #503, opened 2021-07-31
Unhandled timer fired commands during workflow cancelations can cause SDK to issue cancel cmd for already-resolved timer


Sushisource:
## Expected Behavior
We should never send a timer cancel command for a timer that has already fired

## Actual Behavior
We do :(

## Steps to Reproduce the Problem
Repro history in associated PR

PR #504, opened 2021-07-31
Problems with workflow cancel & command resolutions in same WFT

Sushisource:
This PR fixes the specific case of https://github.com/temporalio/sdk-go/issues/503

But, it is not truly correct in all cases since it has to fiddle with event numbering in the classic go sdk fashion and that is gonna be error prone until we do away with it entirely.

Probably the best real way to fix this would be to buffer all cancel commands that would be issued as a result of workflow cancellation, and only *actually* add them if they aren't also resolved by the same workflow task. But that's also ~probably~ a breaking change in terms of not being consistent with the way workers without that change would generate histories.

Even better would be to eliminate ID counting entirely by porting Go to live on top of Core when it's ready

### DIFF
--- a/internal/internal_decision_state_machine.go
+++ b/internal/internal_decision_state_machine.go
@@ -862,6 +862,25 @@ func (h *commandsHelper) addCommand(command commandStateMachine) {
 	h.incrementNextCommandEventID()
 }
 
+// This really should not exist, but is unavoidable without totally redesigning the Go SDK to avoid
+// doing event number counting. EX: Because a workflow execution cancel requested event calls a callback
+// on timers that immediately cancels them, we will queue up a cancel timer command even though that timer firing
+// might be in the same workflow task. In practice this only seems to happen during unhandled command events.
+func (h *commandsHelper) removeCommand(commandId commandID) {
+	// This is also wrong, the problem this solves could technically happen to anything that becomes
+	// cancelled as a result of workflow execution cancellation, where there is also an event resolving
+	// that thing in the same logical WFT.
+	if commandId.commandType != commandTypeCancelTimer {
+		panic("remove command should only be called for cancel timer")
+	}
+	orderedCmdEl, ok := h.commands[commandId]
+	if ok {
+		delete(h.commands, commandId)
+		h.orderedCommands.Remove(orderedCmdEl)
+		h.commandsCancelledDuringWFCancellation--
+	}
+}
+
 func (h *commandsHelper) scheduleActivityTask(
 	scheduleID int64,
 	attributes *commandpb.ScheduleActivityTaskCommandAttributes,
@@ -1203,6 +1222,10 @@ func (h *commandsHelper) cancelTimer(timerID TimerID) commandStateMachine {
 
 func (h *commandsHelper) handleTimerClosed(timerID string) commandStateMachine {
 	command := h.getCommand(makeCommandID(commandTypeTimer, timerID))
+	// If, for whatever reason, we were going to send a timer cancel command, don't do that anymore
+	// since we already know the timer is resolved.
+	possibleCancelId := makeCommandID(commandTypeCancelTimer, timerID)
+	h.removeCommand(possibleCancelId)
 	command.handleCompletionEvent()
 	return command
 }
--- a/internal/internal_decision_state_machine.go
+++ b/internal/internal_decision_state_machine.go
@@ -866,12 +866,10 @@ func (h *commandsHelper) addCommand(command commandStateMachine) {
 // doing event number counting. EX: Because a workflow execution cancel requested event calls a callback
 // on timers that immediately cancels them, we will queue up a cancel timer command even though that timer firing
 // might be in the same workflow task. In practice this only seems to happen during unhandled command events.
-func (h *commandsHelper) removeCommand(commandID commandID) {
-	// This is also wrong, the problem this solves could technically happen to anything that becomes
-	// cancelled as a result of workflow execution cancellation, where there is also an event resolving
-	// that thing in the same logical WFT.
-	if commandID.commandType != commandTypeCancelTimer {
-		panic("remove command should only be called for cancel timer")
+func (h *commandsHelper) removeCancelOfResolvedCommand(commandID commandID) {
+	// Ensure this isn't misused for non-cancel commands
+	if commandID.commandType != commandTypeCancelTimer && commandID.commandType != commandTypeRequestCancelActivityTask {
+		panic("removeCancelOfResolvedCommand should only be called for cancel timer / activity")
 	}
 	orderedCmdEl, ok := h.commands[commandID]
 	if ok {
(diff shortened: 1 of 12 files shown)


Sushisource:
Relates to https://github.com/temporalio/sdk-go/issues/503

vitarb (approved):

test/replaytests/workflows.go
```diff
@@ -113,3 +113,13 @@ func helloworldActivity(ctx context.Context, name string) (string, error) {
 	logger.Info("helloworld activity started")
 	return "Hello " + name + "!", nil
 }
+
+// TimerWf starts a timer and always starts another timer at workflow end, even if cancelled
+func TimerWf(ctx workflow.Context) error {
+	defer func() {
+		// Produce another timer after being cancelled
+		newCtx, _ := workflow.NewDisconnectedContext(ctx)
```
yiminc, line 121: You could do defer func() { workflow.ExecuteActivity() } and we would run into the same problem, right? I don't think this fix is good enough. We need stop sending more command when context is cancelled/errored.

yiminc (changes requested):

test/replaytests/workflows.go
yiminc: I misunderstand the problem. Sending cancellation command in response to workflow cancellation is fine and expected. The problem is sending cancellation command when the corresponding timer/activity etc is closed in the same workflow task. I think the fix is reasonable.

internal/internal_decision_state_machine.go
```diff
@@ -1203,6 +1222,10 @@ func (h *commandsHelper) cancelTimer(timerID TimerID) commandStateMachine {
 
 func (h *commandsHelper) handleTimerClosed(timerID string) commandStateMachine {
 	command := h.getCommand(makeCommandID(commandTypeTimer, timerID))
+	// If, for whatever reason, we were going to send a timer cancel command, don't do that anymore
+	// since we already know the timer is resolved.
+	possibleCancelID := makeCommandID(commandTypeCancelTimer, timerID)
+	h.removeCommand(possibleCancelID)
```
yiminc, line 1228: I think this should not be handled here, but instead be generalized and handled when we remove command for any closed commands.

internal/internal_decision_state_machine.go
```diff
@@ -862,6 +862,25 @@ func (h *commandsHelper) addCommand(command commandStateMachine) {
 	h.incrementNextCommandEventID()
 }
 
+// This really should not exist, but is unavoidable without totally redesigning the Go SDK to avoid
+// doing event number counting. EX: Because a workflow execution cancel requested event calls a callback
+// on timers that immediately cancels them, we will queue up a cancel timer command even though that timer firing
+// might be in the same workflow task. In practice this only seems to happen during unhandled command events.
+func (h *commandsHelper) removeCommand(commandID commandID) {
```
yiminc, line 869: This should be generalized to handle other cases not just for timer:
strawman proposal:
```
func (h *commandsHelper) removeCommand(id commandID) bool {
	if elem, ok := h.commands[id]; ok {
		h.orderedCommands.Remove(elem)
		delete(h.commands, id)
		cmd := elem.Value.(commandStateMachine)
		// cmd is gone, now we need to make sure we don't try to cancel that cmd
		switch cmd := cmd.(type) {
		case *timerCommandStateMachine:
			cancelCmd := makeCommandID(commandTypeCancelTimer, cmd.getID().id)
			if h.removeCommand(cancelCmd) {
				h.commandsCancelledDuringWFCancellation--
			}
		case *activityCommandStateMachine:
			cancelCmd := makeCommandID(commandTypeRequestCancelActivityTask, cmd.getID().id)
			if h.removeCommand(cancelCmd) {
				h.commandsCancelledDuringWFCancellation--
			}
		// add other cases that make sense here.
		}
		return true
	}
	return false
}
```

Then, any where in this file that calls orderedCommands.Remove() should be replaced to call this removeCommand() method.

test/replaytests/workflows.go
Sushisource: Yeah that's very much what I say in the description and mentioned in the meeting. We know this fix doesn't really work properly, but it's also just not possible to fix without huge changes. Simply not sending more commands isn't going to work, because we still want to issue cancellations for things that *aren't* also resolved in the same workflow task.

Buffering them and sending them at the end of the task works, but causes backwards incompatible changes to the way history is created, which we talked about in the meeting too.

So unless there's a simpler fix here which wasn't on the table this cheap patch is what we agreed we'd be fine with, but if you've got an idea of a better way I'm all ears.

internal/internal_decision_state_machine.go
```diff
@@ -862,6 +862,25 @@ func (h *commandsHelper) addCommand(command commandStateMachine) {
...
+	// cancelled as a result of workflow execution cancellation, where there is also an event resolving
+	// that thing in the same logical WFT.
+	if commandID.commandType != commandTypeCancelTimer {
+		panic("remove command should only be called for cancel timer")
+	}
+	orderedCmdEl, ok := h.commands[commandID]
+	if ok {
+		delete(h.commands, commandID)
+		h.orderedCommands.Remove(orderedCmdEl)
+		h.commandsCancelledDuringWFCancellation--
```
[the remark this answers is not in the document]
Sushisource: Yes, it absolutely can, which is why this is a cheapo fix

internal/internal_decision_state_machine.go
Sushisource: I'm inclined to contain the blast of this to just the specific case that needs to be fixed since it's such a lame patch. Although this can technically happen to activities I'm worried making more use of it is just going to lead to more potential problems down the line when we really want to just re-do all of this. I can make it apply to both if we really want though

yiminc: I still think this should be handled here in general way, not in handleTimerClosed(). Agree that if you want to limit the case to only handle timer for now.

yiminc: Actually I think we should fix all cases here. Because each case would cause non-deterministic and block user and it is a time bomb waiting to exploded.

Sushisource: K, sounds good. Thanks for the sample code!

internal/internal_decision_state_machine.go
```diff
@@ -1251,7 +1281,7 @@ func (h *commandsHelper) getCommands(markAsSent bool) []*commandpb.Command {
 
 		// remove completed command state machines
 		if d.getState() == commandStateCompleted {
-			h.orderedCommands.Remove(curr)
+			h.removeCommand(d.getID())
 			delete(h.commands, d.getID())
```
yiminc, line 1285: remove this line.

internal/internal_decision_state_machine.go
```diff
@@ -380,8 +380,8 @@ func (d *commandStateMachineBase) moveState(newState commandState, event string)
 	d.history = append(d.history, newState.String())
 
 	if newState == commandStateCompleted {
-		if elem, ok := d.helper.commands[d.getID()]; ok {
-			d.helper.orderedCommands.Remove(elem)
+		if _, ok := d.helper.commands[d.getID()]; ok {
+			d.helper.removeCommand(d.getID())
 			delete(d.helper.commands, d.getID())
```
yiminc, line 385: remove this line.

yiminc (changes requested):

yiminc (approved):

internal/internal_decision_state_machine.go
Sushisource: So, turned out this didn't work since we only have the information that this should be done inside the closed event handlers

mfateev (approved):

PR #509, opened 2021-08-03
Improve error message for proto payload converters

alexshtin:


converter/proto_payload_converter.go
```diff
@@ -106,7 +110,7 @@ func (c *ProtoPayloadConverter) FromPayload(payload *commonpb.Payload, valuePtr
 		return fmt.Errorf("type: %T: %w", protoValue, ErrTypeNotImplementProtoMessage)
 	}
 
-	// If case if original value is nil, create new instance.
+	// In case if original value is nil, create new instance.
```
Sushisource, line 113:
```suggestion
	// If original value is nil, create new instance.
```

converter/proto_json_payload_converter.go
```diff
@@ -128,7 +132,7 @@ func (c *ProtoJSONPayloadConverter) FromPayload(payload *commonpb.Payload, value
 		return fmt.Errorf("type: %T: %w", protoValue, ErrTypeNotImplementProtoMessage)
 	}
 
-	// If case if original value is nil, create new instance.
+	// In case if original value is nil, create new instance.
```
Sushisource, line 135:
```suggestion
	// If original value is nil, create new instance.
```

Sushisource (approved):

Issue #515, opened 2021-08-13
Add Signal/Query interceptors


vitarb:
One of our users has [recently noticed](https://community.temporal.io/t/fine-grained-authorization/2646/6) that we don't have Query/Signal interceptors.

We should address this [TODO](https://github.com/temporalio/sdk-go/blob/c7aae265b40d85663431281d5959e0c61dbb58eb/internal/interceptors.go#L56) and add interceptors for Signal/Query.

vitarb:
We should be able to support open tracing for all calls, headers should be available on API calls too.
Discuss with @Spikhalskiy who has a lot of context in java.

vitarb:
Need to make sure that headers are passed on the server for signal/query.

vitarb:
I'm going to keep this issue focused on Signal/Query as it's more specific.

PR #522, opened 2021-08-27
Add inbound interceptors for Signal and Query

vitarb:
## What was changed
Added inbound interceptors for Signal and Query

## Why?
We lack ability to intercept signals and queries.

## Checklist

1. Partially resolves issue #515

2. How was this tested: unit tested

3. Any docs updates needed?
No

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -383,6 +383,15 @@ func NewFuture(ctx Context) (Future, Settable) {
 	return impl, impl
 }
 
+func (wc *workflowEnvironmentInterceptor) ProcessSignal(Context, string, interface{}) {
+	// no op
+}
+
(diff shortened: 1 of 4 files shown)


internal/workflow.go
```diff
@@ -383,6 +383,15 @@ func NewFuture(ctx Context) (Future, Settable) {
 	return impl, impl
 }
 
+func (wc *workflowEnvironmentInterceptor) ProcessSignal(Context, string, interface{}) {
+	// no op
```
Spikhalskiy, line 387: You sure? If it's the same as the root interceptor in Java, I would expect the logic that actually triggers user code processing the signal be around here. Or scheduling of this processing by creating a thread (to respect the ordering).

PR #523, opened 2021-08-30
Add support for errors.Is()

### LINKED ISSUE #403: Can't use `errors.Is` to check a returned error
(Moving https://github.com/temporalio/temporal/issues/1453 here, as I think I created it in the wrong place)

## Expected Behavior

If an activity returns an error from some custom error from my system, it's wrapped and I can check it using `errors.Is`.

## Actual Behavior

My error message is wrapped there, but I can't compare/check it.

```golang
var ErrNotFound = errors.New("not found!")

func TestWorkflowTestSuite(t *testing.T) {
	suite.Run(t, new(WorkflowTestSuite))
}

type WorkflowTestSuite struct {
	suite.Suite
	testsuite.WorkflowTestSuite
	env *testsuite.TestWorkflowEnvironment
}

func (s *WorkflowTestSuite) SetupTest() {
	s.env = s.NewTestWorkflowEnvironment()
}

func (s *WorkflowTestSuite) TearDownTestSuite() {
	s.env.AssertExpectations(s.T())
}

func (s *WorkflowTestSuite) Test_WorkflowErrorHandling() {
	activity := func(number int) error {
		return fmt.Errorf("error: %w", ErrNotFound)
	}
	s.env.RegisterActivity(activity)

	workflow := func(ctx workflow.Context, number int) error {
				MaximumAttempts:    3,
				BackoffCoefficient: 2,
			},
		})

		if err := workflow.ExecuteActivity(ctx, activity, number).Get(ctx, nil); err != nil {
			return err
		}

		return nil
	}
	s.env.RegisterWorkflow(workflow)
	s.env.ExecuteWorkflow(workflow, 123)

	s.True(s.env.IsWorkflowCompleted())
	err := s.env.GetWorkflowError()
	s.ErrorIs(err, ErrNotFound)
}
```

The result will be:

```
Error:      	Target error should be in err chain:
                    expected: "not found!"
                    in chain: "workflow execution error (type: func2, workflowID: default-test-workflow-id, runID: default-test-run-id): activity error (type: func1, scheduledEventID: 0, startedEventID: 0, identity: ): error: not found! (type: wrapError, retryable: true): not found!"
                        "activity error (type: func1, scheduledEventID: 0, startedEventID: 0, identity: ): error: not found! (type: wrapError, retryable: true): not found!"
                        "error: not found! (type: wrapError, retryable: true): not found!"
                        "not found!"
```

## Specifications

  - Version: v1.5.0; 1.6.0
  - Platform: macOS

laniehei:
Resolves #403


internal/error_test.go
```diff
@@ -441,6 +441,102 @@ func Test_CanceledError(t *testing.T) {
...
+	workflowErr := NewTimeoutError("timeout", enumspb.TIMEOUT_TYPE_START_TO_CLOSE, nil)
+	require.True(t, errors.Is(workflowErr, &TimeoutError{}), workflowErr)
+}
+
+// TestWorkflowPanicErrorIs tests whether
+// the errors.Is() method operates as
+// expected.
+func TestWorkflowPanicErrorIs(t *testing.T) {
+	workflowErr := newWorkflowPanicError("", "")
+	require.True(t, errors.Is(workflowErr, &workflowPanicError{}), workflowErr)
```
vitarb, line 481: You can instead call a little shorter form, which under the hood calls `errors.Is` anyways:
'''
require.ErrorIs(t, workflowErr, &CanceledError{})
'''

vitarb (approved):
Please address my comment, otherwise LGTM.

vitarb:
Also linter complains about missing comments (see build output).

internal/error_test.go
laniehei: This is specifically testing that errors.Is(), as the original issue stated. require.ErrorIs() would obfuscate the purpose of the test, no?

vitarb: I would prefer a shorter version given that it does the same thing, but I'll leave this up to you to decide.

laniehei: It appears the user's code originally failed on your more succinct version. Changed the tests!

PR #525, opened 2021-09-02
Revert "Allow pending goroutines exit after workflow completion (#520)"

vitarb:
This reverts commit 0c69423d6b98508c910f65ddad17fbe1967cd8d7.

We are observing some instability in tests after this change landed, reverting it for additional investigation.


Sushisource:
I'm thinking this might actually be due to changes server side related to Cron

vitarb:
Doesn't fix the build failure. No reverting is needed.

PR #527, opened 2021-09-02
Follow new execution chain for other event types also

dnr:
## What was changed

After https://github.com/temporalio/temporal/pull/1866, cron and retried workflow executions will appear with the "proper" event types in their history, instead of continued-as-new. To keep the existing behavior of `WorkflowRun.Get` blocking until the whole chain ends, it should follow the new executions in other event types.

## Why?

Preserving current behavior with new server version.

## Checklist

1. Closes 

2. How was this tested:

The existing test `TestIntegrationSuite/TestFailurePropagation` fails with the current server, this fixes it.

3. Any docs updates needed?


internal/internal_workflow_client.go
```diff
@@ -1082,6 +1087,11 @@ func (workflowRun *workflowRunImpl) Get(ctx context.Context, valuePtr interface{
 		return workflowRun.dataConverter.FromPayloads(attributes.Result, valuePtr)
 	case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_FAILED:
 		attributes := closeEvent.GetWorkflowExecutionFailedEventAttributes()
+		if attributes.NewExecutionRunId != "" {
+			curRunID := util.PopulatedOnceCell(attributes.NewExecutionRunId)
+			workflowRun.currentRunID = &curRunID
+			return workflowRun.Get(ctx, valuePtr)
+		}
```
Sushisource, line 1094: Can we extract this bit to its own little helper function? Also gives us the opportunity to comment it explaining the reason

dnr: Sure

Sushisource (approved):

PR #539, opened 2021-09-16
Start worker even when no workflow/activity registered

yiminc:
## What was changed
Do not skip worker start when registry is empty.

## Why?
There are uses cases where registry happen after worker is started, for example in integration tests that you want to start worker as part of test setup where you don't know what to register yet.
This check only make sense when the registry was done via golang's init() func. Now registry is per worker and can be done after worker is started. The current behavior of silently skip worker start is inconvenient.

### DIFF
--- a/evictiontest/workflow_cache_eviction_test.go
+++ b/evictiontest/workflow_cache_eviction_test.go
@@ -159,7 +159,7 @@ func (s *CacheEvictionSuite) TestResetStickyOnEviction() {
 	s.service.EXPECT().ResetStickyTaskQueue(gomock.Any(), gomock.Any()).DoAndReturn(mockResetStickyTaskQueue).Times(1)
 
 	client := internal.NewServiceClient(s.service, nil, internal.ClientOptions{})
-	workflowWorker := internal.NewAggregatedWorker(client, "taskqueue", worker.Options{})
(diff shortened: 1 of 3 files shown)


Sushisource (approved):

evictiontest/workflow_cache_eviction_test.go
```diff
@@ -159,7 +159,7 @@ func (s *CacheEvictionSuite) TestResetStickyOnEviction() {
 	s.service.EXPECT().ResetStickyTaskQueue(gomock.Any(), gomock.Any()).DoAndReturn(mockResetStickyTaskQueue).Times(1)
 
 	client := internal.NewServiceClient(s.service, nil, internal.ClientOptions{})
-	workflowWorker := internal.NewAggregatedWorker(client, "taskqueue", worker.Options{})
+	workflowWorker := internal.NewAggregatedWorker(client, "taskqueue", worker.Options{LocalActivityWorkerOnly: true})
```
yiminc, line 162: This test requires workflow worker only and the mock service is setup accordingly. If we start activity worker, the mock would complain. The test works because this test does not register any activity and as side effect, activity worker does not start. With the change in this PR, activity worker would start and this test failed. Fix is to explicitly disable activity worker.

PR #540, opened 2021-09-17
Fix local activity retry backoff

yiminc:
## What was changed
Fix local activity retry backoff calculation.

## Why?
Attempt used to start from 0, then it was changed to start from 1, but backoff calculation wasn't updated accordingly.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


vitarb (approved):
Is this relevant for local activity only, or should we change general retry logic here https://github.com/temporalio/sdk-go/blob/00ce39afa471d07fdcd34142f645d42fadc92c36/internal/common/retry/interceptor.go#L136 as well?

yiminc:
@vitarb , the backoff changed in this PR is only used by local activity and the unit test framework to calculate activity/workflow retry backoff in unit test. I'm not sure about the grpc_retry as I don't know what is the attempt used there. It is worth double check.

PR #550, opened 2021-09-24
Service interceptors

robholland:
This is a spike of the service interceptor idea to provide a plugin-safe API for encryption and compression. For use cases that don't require varying encoding based on workflow context this is working well. The missing piece is a way for interceptor's encode method to have sight of the Header in order to make decisions about which key it should use or similar.


Sushisource:
Here's maybe an alternative approach with less code, and I think maybe can be layered on in a backwards compat way.

One of the main missing pieces is giving information about the overall request/response being de/serialized. You've solved that here by intercepting the request or response, and then selectively exposing parts of that req/resp. My idea is similar, but a bit less verbose and tries to expand on our existing `DataConverter`

The idea is rather than having `DataConverter` operate on `Payload` or `Payloads` types, which completely lack any sort of context, it can operate on a `PayloadInContext` type (or, they could be passed as extra params to new functions, but I think I prefer a type) which looks something like:

```go
type PayloadInContext struct {
  // Exposed directly so user can do conversion. Or could be exposed as method.
  Data: *Payload
  // This is the request or response type which eventually contains the payload. 
  // IE: If the payload is nested within a `Header` inside of a `PollWorkflowTaskQueueResponse`,
  //  this field would be a pointer to the overall poll response.
  livesInside: interface{}
  // If applicable, pointer to wf / whatever else context
  context: Context
}
```

Then that struct can implement an interface (or directly expose, not sure) methods like `getHeaders` which would return the overall headers for the req/resp if they exist, `getContext` for the workflow/etc context, and anything else needed.

For compat, I'm not exactly sure what the best approach is to expose this in a compat way. Maybe we wrap incoming dataconverters with something that can expose the `PayloadInContext` struct? That way the wold way still works but you can get the upgraded version if needed?

WDYT?

internal/interceptors.go
```diff
@@ -242,3 +248,1210 @@ func (t *WorkflowOutboundCallsInterceptorBase) GetLastCompletionResult(ctx Conte
...
+)
+
+var HeadersCache = map[string]*commonpb.Header{}
+
+type BaseRequestResponseInterceptor struct{}
+
+var _ RequestResponseInterceptor = &BaseRequestResponseInterceptor{}
+
+func (i *BaseRequestResponseInterceptor) SetHeaderForWorkflowExecution(token string, header *commonpb.Header) {
+	HeadersCache[token] = header
```
cretz, line 427: What is the use case for this? This isn't safe for concurrent calls of course yet is updating a global var.

robholland:
I think we're heading in a different direction for this. Closing for now.

PR #559, opened 2021-09-25
Add a test which double checks some cancel-on exit behavior

Sushisource:
## What was changed
Added a test (that didn't repro anything)

## Why?
Protection against possible canceled-after-fired errors

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -59,6 +59,15 @@ import (
 	"go.temporal.io/sdk/workflow"
 )
 
+const (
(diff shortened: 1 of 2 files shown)


Sushisource:
Was not able to repro any cancelled-after-fired bug. Issue we saw may have been b/c of nondeterminism in workflow code.

vitarb (approved):

PR #561, opened 2021-09-28
Skip failing test on Windows

cretz:
## What was changed

Skipped failing Windows test

## Why?

Test sends manual interrupt signal which fails. Full suite now succeeds on Windows.

### DIFF
--- a/internal/internal_worker_test.go
+++ b/internal/internal_worker_test.go
@@ -30,6 +30,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"runtime"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


internal/internal_worker_test.go
```diff
@@ -1474,6 +1475,10 @@ func (s *internalWorkerTestSuite) TestCreateWorkerWithDataConverter() {
 }
 
 func (s *internalWorkerTestSuite) TestCreateWorkerRun() {
+	// Windows doesn't support signalling interrupt
```
alexshtin, line 1478:
```suggestion
	// Windows doesn't support signalling interrupt.
```

alexshtin (approved):

vitarb (approved):

PR #564, opened 2021-09-29
Allow errors in signal interceptors

vitarb:
## What was changed
Allowed errors in signal interceptors. 

## Why?
Previously there was no way to return an error from the signal interceptor, meaning that even if there is a problem signal way processed anyways and it was impossible to interrupt signal handling. This can be a problem in case if interceptor is doing some non-optional operations that have a chance to fail (e.g. network interactions).


test/integration_test.go
```diff
@@ -558,15 +561,39 @@ func (ts *IntegrationTestSuite) TestSignalWorkflow() {
...
+	ctx, cancel := context.WithTimeout(context.Background(), ctxTimeout)
+	defer cancel()
+
+	// Return error 3 times from the interceptor
+	ts.inboundSignalInterceptor.returnErrorTimes = 3
+	wfOpts := ts.startWorkflowOptions("test-signal-workflow-interceptor-error")
+	run, err := ts.client.ExecuteWorkflow(ctx, wfOpts, ts.workflows.SignalWorkflow)
+	ts.Nil(err)
+	// Let workflow task run and send signal after to ensure correct order.
+	<-time.After(time.Second)
```
Sushisource, line 574: Ideally let's get this a better way - don't want to introduce more flakes like the ones you're already fixing :)

test/integration_test.go
```diff
@@ -558,15 +561,39 @@ func (ts *IntegrationTestSuite) TestSignalWorkflow() {
...
+	err = ts.client.SignalWorkflow(ctx, "test-signal-workflow-interceptor-error", run.GetRunID(), "string-signal", "string-value")
+	ts.NoError(err)
+
+	wt := &commonpb.WorkflowType{Name: "workflow-type"}
+	err = ts.client.SignalWorkflow(ctx, "test-signal-workflow-interceptor-error", run.GetRunID(), "proto-signal", wt)
+	ts.NoError(err)
+
+	var protoValue *commonpb.WorkflowType
+	err = run.Get(ctx, &protoValue)
+	// Workflow should succeed after retries upon an error in the signal interceptor
```
Sushisource, line 584: Ideally we'd somehow verify the "signal is dropped" behavior we want here. It's not super obvious to me if/how this test is doing that

internal/internal_workflow_testsuite.go
```diff
@@ -2249,7 +2249,7 @@ func (env *testWorkflowEnvironmentImpl) signalWorkflow(name string, input interf
 		panic(err)
 	}
 	env.postCallback(func() {
-		env.signalHandler(name, data)
+		_ = env.signalHandler(name, data)
```
cretz, line 2252: I'm curious, this seems like a JetBrains-ism, is it? I don't think any traditional go tools have any problem with ignoring these returns.

vitarb: We have errorcheck linter, which complains if you ignore an error.

test/integration_test.go
vitarb: Yes, as it stands test verifies that we don't break existing behavior. It's a little tricky as we need to verify that workflow task didn't run, hence can't test it from the workflow itself. I thought about checking metrics but that proven to be flaky.

PR #566, opened 2021-09-29
Apply default retry policy for local activities

### LINKED ISSUE #563: missing default retry policy for local activity
## Expected Behavior
LocalActivity should be retried even when no retry policy is specified.

## Actual Behavior
LocalActivity is not retried if no retry policy is specified. Or if retry policy is specified with only MaxAttempt=3 and all other fields left as unspecified.

## additional
Need to add unit test AND integration test to verify both remote activity and local activity is retried when no retry policy is specified. Also, when retry policy is specified with only MaxAttempt = 3 and all other fields unspecified.

vitarb:
## What was changed
Added default retry policy for local activities.

## Why?
Previously local activities would rely on workflow task retries in case if policy hasn't been specified. Also since no defaults were applied invalid/partial policy would cause it to use no retries within workflow task.

## Checklist

1. Closes #563

2. How was this tested:
Integration tests

3. Any docs updates needed?
Updated javadocs, we should also update public documentation.

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -146,7 +146,7 @@ type (
 
 		// RetryPolicy specify how to retry activity if error happens.
 		// Optional: default is to retry according to the default retry policy up to ScheduleToCloseTimeout
-		// with 50ms initial delay between retries and 2x backoff.
+		// with 1sec initial delay between retries and 2x backoff.
 		RetryPolicy *RetryPolicy
 	}
 )
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -146,6 +146,7 @@ type (
 
 		// RetryPolicy specify how to retry activity if error happens.
 		// Optional: default is to retry according to the default retry policy up to ScheduleToCloseTimeout
+		// with 50ms initial delay between retries and 2x backoff.
 		RetryPolicy *RetryPolicy
 	}
 )
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -1542,7 +1542,7 @@ func applyRetryPolicyDefaultsForLocalActivity(policy *RetryPolicy) *RetryPolicy 
 		policy.BackoffCoefficient = 2
 	}
 	if policy.InitialInterval == 0 {
-		policy.InitialInterval = 50 * time.Millisecond
+		policy.InitialInterval = 1 * time.Second
 	}
 	if policy.MaximumInterval == 0 {
 		policy.MaximumInterval = policy.InitialInterval * 100
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -1530,11 +1530,11 @@ func WithLocalActivityOptions(ctx Context, options LocalActivityOptions) Context
 
 	opts.ScheduleToCloseTimeout = options.ScheduleToCloseTimeout
 	opts.StartToCloseTimeout = options.StartToCloseTimeout
-	opts.RetryPolicy = applyRetryPolicyDefaultsForLocalActivity(options.RetryPolicy, opts.ScheduleToCloseTimeout)
+	opts.RetryPolicy = applyRetryPolicyDefaultsForLocalActivity(options.RetryPolicy)
 	return ctx1
 }
 
-func applyRetryPolicyDefaultsForLocalActivity(policy *RetryPolicy, timeout time.Duration) *RetryPolicy {
+func applyRetryPolicyDefaultsForLocalActivity(policy *RetryPolicy) *RetryPolicy {
 	if policy == nil {
 		policy = &RetryPolicy{}
 	}
@@ -1543,9 +1543,6 @@ func applyRetryPolicyDefaultsForLocalActivity(policy *RetryPolicy, timeout time.
 	}
 	if policy.InitialInterval == 0 {
 		policy.InitialInterval = 50 * time.Millisecond
-	}
-	if policy.MaximumInterval == 0 {
-		policy.MaximumInterval = timeout
 	}
 	return policy
 }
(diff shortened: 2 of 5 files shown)


test/workflow_test.go
```diff
@@ -966,6 +966,50 @@ func (w *Workflows) WorkflowWithLocalActivityRetries(ctx workflow.Context) error
...
+	for _, fut := range futures {
+		err := fut.Get(ctx, nil)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (w *Workflows) WorkflowWithLocalActivityRetriesAndPartialRetryPolicy(ctx workflow.Context) error {
```
Sushisource, line 990: These test names appear to be swapped

Sushisource (commented):
We should add unit tests as well. This is one of those places where the unit/integ behavior diverges and causes confusion later.

test/workflow_test.go
Sushisource: Ah, you fixed that

internal/activity.go
```diff
@@ -146,6 +146,7 @@ type (
 
 		// RetryPolicy specify how to retry activity if error happens.
 		// Optional: default is to retry according to the default retry policy up to ScheduleToCloseTimeout
+		// with 50ms initial delay between retries and 2x backoff.
```
yiminc, line 149: The default value for RetryPolicy state the InitialInterval is 1s. I'm worried that this 50ms here would cause confusion.

test/workflow_test.go
```diff
@@ -966,6 +966,51 @@ func (w *Workflows) WorkflowWithLocalActivityRetries(ctx workflow.Context) error
 	return nil
 }
 
+func (w *Workflows) WorkflowWithLocalActivityRetriesAndDefaultRetryPolicy(ctx workflow.Context) error {
+	laOpts := w.defaultLocalActivityOptions()
+	// Don't set any retry policy
+	ctx = workflow.WithLocalActivityOptions(ctx, laOpts)
```
yiminc, line 972: Could you add a unit test using TestWorkflowEnvironment just to verify unit test works as well.

internal/workflow.go
```diff
@@ -1530,10 +1530,26 @@ func WithLocalActivityOptions(ctx Context, options LocalActivityOptions) Context
...
+		policy = &RetryPolicy{}
+	}
+	if policy.BackoffCoefficient == 0 {
+		policy.BackoffCoefficient = 2
+	}
+	if policy.InitialInterval == 0 {
+		policy.InitialInterval = 50 * time.Millisecond
+	}
+	if policy.MaximumInterval == 0 {
+		policy.MaximumInterval = timeout
```
yiminc, line 1548: I think this is not what you intended?

yiminc (changes requested):

internal/activity.go
vitarb: 1 seconds seems too slow for local activity since the whole point is that they run locally and are mostly used for low latency use cases.

internal/workflow.go
vitarb: Why? Maximum interval should be equal to schedule to close timeout. That's what our [documentation](https://github.com/temporalio/sdk-go/blob/0a5c65d55f693ea569b5ca7009eed9d4ea9a5f64/internal/activity.go#L148) was saying, which wasn't true. So this line is fixing it.

vitarb: Yes, you are right, we should follow our [documentation](https://github.com/temporalio/sdk-go/blob/e8455cbe8014d2587a2b72ce153ae490160d9b11/internal/client.go#L556) and default max interval to 100x initial interval.

internal/activity.go
yiminc: Then we need to update the comment/doc for RetryPolicy.InitialInterval as well to reflect that it is 50ms for local activity. @mfateev

internal/workflow.go
```diff
@@ -1530,10 +1530,26 @@ func WithLocalActivityOptions(ctx Context, options LocalActivityOptions) Context
 
 	opts.ScheduleToCloseTimeout = options.ScheduleToCloseTimeout
 	opts.StartToCloseTimeout = options.StartToCloseTimeout
-	opts.RetryPolicy = options.RetryPolicy
+	opts.RetryPolicy = applyRetryPolicyDefaultsForLocalActivity(options.RetryPolicy)
 	return ctx1
 }
 
+func applyRetryPolicyDefaultsForLocalActivity(policy *RetryPolicy) *RetryPolicy {
```
yiminc, line 1537: where do we enforce that it only retry up to ScheduleToCloseTimeout? Is the comment on LocalActivityOption accurate?

test/workflow_test.go
vitarb: Yes, will do.

internal/workflow.go
vitarb: It's at the LA [task level](https://github.com/temporalio/sdk-go/blob/0550203f48dd2e84f0db6d0374b1c58dcc9b33a5/internal/internal_event_handlers.go#L530), we calculate it right away and set it as a context deadline.

internal/activity.go
vitarb: Let's have a quick conversation about this tomorrow. I want to make sure we are all on the same page.

internal/activity.go
```diff
@@ -146,6 +146,7 @@ type (
 
 		// RetryPolicy specify how to retry activity if error happens.
 		// Optional: default is to retry according to the default retry policy up to ScheduleToCloseTimeout
```
yiminc, line 148: may need to clarify what 'up to ScheduleToCloseTimeout' means.

yiminc (approved):

internal/activity.go
vitarb: Looking at java we have [1 second default](https://github.com/temporalio/sdk-java/blob/24a55be08d8cfaf0b8bb2bb49aab38e8344ce08c/temporal-sdk/src/main/java/io/temporal/common/RetryOptions.java#L232) there. This convinced me to make it the same for now.

Sushisource (approved):

PR #567, opened 2021-09-30
Discard unnecessary commands when close workflow

yiminc:
## What was changed
Discard unnecessary commands when close workflow

## Why?
User code in defer function in workflow code could generate unwanted commands when workflow is closed. Those commands would not be handled anyway but they might cause validation error on server that result in workflow task to fail. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource:
I'm not sure I love this. This needs to be implemented in every SDK and seems like something that can easily be handled server side at little cost. If we're just silently ignoring them here, why not just silently ignore them server side?

Why bother with this at all?

PR #588, opened 2021-10-14
Fix command ordering issue on child workflow cancel

cretz:
## What was changed

Child workflow cancel command was sent in the order of when in started instead of when it was canceled. This moves the command to the end of the order once it's canceled.

## Why?

Child workflow cancellation was being recorded based on when the child started instead of when the child was actually canceled.

## For discussion

Two primary questions:

1. Is it acceptable that this fix is tailored to child workflow cancellation? It would be a large undertaking to revisit the entire state machine to make sure that "states" that issue new commands are moved to the end of the command list every time.
2. Can a `StartChildWorkflowExecution` command and a `RequestCancelExternalWorkflowExecution` for that same workflow ever occur in the same `RespondWorkflowTaskCompleted` server call? If so this fix of moving the cancel to the end would need to be changed to instead create an entirely new command/state on cancel leaving the create one alone. I do not believe this is the case since the state can only represent one "state" at a time.

### DIFF
--- a/evictiontest/workflow_cache_eviction_test.go
+++ b/evictiontest/workflow_cache_eviction_test.go
@@ -138,7 +138,7 @@ func (s *CacheEvictionSuite) TestResetStickyOnEviction() {
 	}
 	// pick 5 as cache size because it's not too big and not too small.
 	cacheSize := 5
-	internal.SetStickyWorkflowCacheSize(cacheSize - 1)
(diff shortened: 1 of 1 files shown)


Sushisource:
To answer 2 -- I think it *could*, but it shouldn't. The SDK should simply not send any commands in that circumstance.

Issue #601, opened 2021-10-25
Support workflow metadata query


bergundy:
See original issue temporalio/sdk-features#51

cretz:
See https://github.com/temporalio/features/issues/51

PR #610, opened 2021-11-01
Interceptor Redesign/Rewrite and OpenTelemetry Implementation

### LINKED ISSUE #265: Add ActivityInterceptor
Java SDK has [ActivityInterceptor](https://github.com/temporalio/sdk-java/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityInterceptor.java#L22) and Go SDK misses one.

### LINKED ISSUE #526: Add open tracing support
See related discussion in #515

### LINKED ISSUE #529: Redesign interceptors to match the Java / Node.js capabilities

With the current interceptors design, there's no way to change the default execution of some SDK operations.
For example, one might want to use interceptors for authz when starting / signaling / querying workflows.

Some interceptor methods do not capture the entire span of the operation and cannot be used for tracing which is one of the use-cases for interceptors.

It also looks like there [query interceptor](https://github.com/temporalio/sdk-go/blob/cc674e0e50ffe72b85694a330cb85bc9b93e29aa/internal/interceptors.go#L64) is not chainable because it states that the interceptor implementor MUST call the provided handler function.

Interceptors should have similar capabilities across the different SDKs, they should all be chainable, capture the entires span of the operation and allow modification of the default SDK behavior.

**Additional context**

- #526 - make sure that we support tracing for queries and signals (requires server changes to add headers to the protocol)
- There's been a community request to implement authz of signals / queries using interceptors

### LINKED ISSUE #602: Support OpenTelemetry tracing
Currently, Temporal SDK supports OpenTracing which is now reported to be superseded by OpenTelemetry

OpenTracing API needs to be replaced with OpenTelemetry API. 

Using OpenTracing created out of OpenTelemetry bridge does not work properly, probably due to some context propagation-related issues.

cretz:
## What was changed

See the proposal at https://github.com/temporalio/proposals/pull/45.

* All existing interceptor code was remove (in fact, entire `interceptors` package was removed)
* New `interceptor` package added with support for client, workflow, and activity interception on par with other SDKs
* OpenTelemetry support on top of new interceptors
* OpenTracing support on top of new interceptors
* Various tests and utilities for supporting new interceptors

The review looks larger than it is due to Go boilerplate.

## Why?

Go's existing interceptor design was limited and didn't support tracing, activities, clients, etc.

## Checklist

1. Closes #265 
2. Closes #526
3. Closes #529
4. Closes #602
5. There is a spot or two in the review that I'll point out that may warrant discussion
6. https://github.com/temporalio/documentation/issues/714 will track public interceptor documentation
6. https://github.com/temporalio/samples-go/issues/150


internal/client.go
```diff
@@ -436,6 +431,13 @@ type (
 		// Optional parameter that is designed to be used *in tests*. It gets invoked last in
 		// the gRPC interceptor chain and can be used to induce artificial failures in test scenarios.
 		TrafficController TrafficController
+
+		// Interceptors to apply to some calls of the client. Any interceptors that
+		// also implement Interceptor (meaning they implement WorkerInterceptor in
+		// addition to ClientInterceptor) will be used for worker interception as
+		// well. When worker interceptors are here and in worker options, the ones
+		// here wrap the ones in worker options.
+		Interceptors []ClientInterceptor
```
cretz, line 440: I should make it clear earlier interceptors wrap later ones in this slice (same for worker options).

interceptor/interceptor.go
```diff
@@ -0,0 +1,221 @@
...
+// calls will be intercepted by it. If an implementation of this interceptor is
+// provided via worker options, all worker calls will be intercepted by it.
+//
+// All implementations of this should embed InterceptorBase but are not required
+// to.
+type Interceptor = internal.Interceptor
+
+// InterceptorBase is a default implementation of Interceptor meant for
+// embedding. It simply embeds ClientInterceptorBase and WorkerInterceptorBase.
+type InterceptorBase = internal.InterceptorBase
```
cretz, line 46: Go lint is complaining about this stuttering and breaking the build. I am going to change the linter, not this (it is a stubborn linter, so I may not be able to easily).

internal/internal_activity.go
```diff
@@ -302,98 +302,132 @@ func isActivityContext(inType reflect.Type) bool {
...
+	in *ExecuteActivityInput,
+) (interface{}, error) {
+	// Remove header from context
+	ctx = contextWithoutHeader(ctx)
+
+	return executeFunctionWithContext(ctx, a.fn, in.Args)
+}
+
+func (a *activityEnvironmentInterceptor) GetInfo(ctx context.Context) ActivityInfo {
+	// TODO(cretz): Accept the costs of copying everytime? See https://github.com/temporalio/proposals/pull/45/files#r732945623
```
cretz, line 366: Need to remove this TODO, I think we accept this (it's a big copy, but meh, it was decided long before me in `activity.GetInfo` that copying this huge struct was ok and inconsistent with `workflow.GetInfo`).

internal/internal_workflow.go
```diff
@@ -517,24 +518,17 @@ func (d *syncWorkflowDefinition) Execute(env WorkflowEnvironment, header *common
...
-		ch := eo.getSignalChannel(d.rootCtx, name).(*channelImpl)
-		ok := ch.SendAsync(input)
-		if !ok {
-			panic(fmt.Sprintf("Exceeded channel buffer size for signal: %v", name))
-		}
-		return nil
+		// TODO(cretz): We decode ahead of time for HandleQuery but we don't here. Is
+		// this inconsistency acceptable? Is it ok to pretend that "Arg" is
+		// interface{} when we know it will always be *commonpb.Payloads?
+		// See https://github.com/temporalio/proposals/pull/45/files#r732871182
```
cretz, line 524: I would like some feedback/discussion here. We cannot decode ahead of time for signals, but like the old interceptors, we leave the `Arg` as `interface{}` even though we know it's always a payload. Shall we change its type since we know it's always a payload and we can't decode it?

internal/internal_workflow_client.go
```diff
@@ -149,98 +150,6 @@ type (
 	}
 )
 
-// StartWorkflow starts a workflow execution
```
cretz, line 152: While large swaths of code like this look like they were deleted, they were just moved to an interceptor implementation so these calls could be intercepted.

interceptor/interceptor.go
Sushisource: You can just disable individual lines with a comment I believe, too

interceptor/interceptor.go
```diff
@@ -0,0 +1,221 @@
...
+
+// ClientCancelWorkflowInput is input for
+// ClientOutboundInterceptor.CancelWorkflow.
+type ClientCancelWorkflowInput = internal.ClientCancelWorkflowInput
+
+// ClientTerminateWorkflowInput is input for
+// ClientOutboundInterceptor.TerminateWorkflow.
+type ClientTerminateWorkflowInput = internal.ClientTerminateWorkflowInput
+
+// ClientQueryWorkflowInput is input for
```
Sushisource, line 195: All these "is input for" phrasings maybe read a bit better as "is the input to"

But you can take or leave it, not at all a big deal.

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,476 @@
...
+	runIDTagKey      = "temporalRunID"
+)
+
+// Tracer is an interface for tracing implementations as used by
+// NewTracingInterceptor. Most callers do not use this directly, but rather use
+// the opentracing or opentelemetry packages.
+type Tracer interface {
+	// SpanContextKey provides a key to put a span on a context unrelated to how a
+	// span might otherwise be put on a context by ContextWithSpan. This should
+	// never be nil.
```
Sushisource, line 47: I think I get what this does from this comment, but I don't get why it exists. Maybe change this to be more about why it exists

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,476 @@
...
+func (t *tracingWorkflowOutboundInterceptor) ExecuteLocalActivity(
+	ctx workflow.Context,
+	activityType string,
+	args ...interface{},
+) workflow.Future {
+	// Start span writing to header
+	span, ctx, err := t.startNonReplaySpan(ctx, "StartActivity", activityType, true)
+	if err != nil {
+		return err
+	}
```
Sushisource, line 298: There should probably be something that distinguishes a local activity from a normal one. A tag or a different name or something

interceptor/interceptor.go
cretz: Not with golint. It is deprecated and the authors stubbornly and _very_ intentionally chose not to allow such a thing. The vast majority of projects have moved off of it (or at least don't error on it).  See https://github.com/golang/lint/issues/263, https://github.com/golang/go/issues/38968, etc.

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,476 @@
...
+func (t *tracingWorkflowOutboundInterceptor) startNonReplaySpan(
+	ctx workflow.Context,
+	operation string,
+	name string,
+	dependedOn bool,
+) (span TracerSpan, newCtx workflow.Context, futErr workflow.ChildWorkflowFuture) {
+	// Noop span if replaying
+	if workflow.IsReplaying(ctx) {
+		return nopSpan{}, ctx, nil
+	}
```
Sushisource, line 352: It's not clear to me how much this will no-op out but I don't think we want to squelch all spans ever when replaying. You lose a lot of info that can be helpful when debugging. At minimum you probably want a way to at least turn these on

interceptor/interceptor.go
cretz: Will change next commit

interceptor/tracing_interceptor.go
cretz: Will do. For clarity, it exists because spans are put on contexts and taken back off and they might want to control the key where that happens.

internal/internal_activity.go
Sushisource: Size wise it's honestly not all that huge. Just a few kinda-long strings, which are just pointers anyway

internal/internal_worker.go
```diff
@@ -767,56 +763,42 @@ func (ae *activityExecutor) GetFunction() interface{} {
...
+	result, resultErr := interceptor.ExecuteActivity(ctx, &ExecuteActivityInput{Args: args})
+	var serializedResult *commonpb.Payloads
+	if result != nil {
+		// As a special case, if the result is already a payload, just use it
+		var ok bool
+		if serializedResult, ok = result.(*commonpb.Payloads); !ok {
+			var err error
+			if serializedResult, err = encodeArg(dataConverter, result); err != nil {
+				return nil, err
+			}
```
Sushisource, line 798: Seems like this and the above bit (~ln 740) could still make use of a `serializeResults` helper function

interceptor/tracing_interceptor.go
cretz: I believe in both Node and Java this is just `StartActivity` with no special tags for local vs non-local (granted Node has no "local" concept yet). Is this something we want to ask all SDKs to differentiate?

interceptor/tracing_interceptor.go
cretz: This replicates the existing behavior in Go and the behavior at least as seen in Java. Workflow outbound spans are just not created during replay. Is this something we want to make configurable in all APIs?

internal/internal_worker.go
cretz: Unless I'm misunderstanding, it'd just be a couple line method to call `encodeArg`. I am not seeing the benefit of reuse. 

Also IIRC, to maintain functionality with existing code, I only allow activity result to bypass serialization if already the proto type, not workflow. Like many places in the code base, it is unfortunately inconsistent but the benefit of change is less than the potential cost of incompatibility.

internal/internal_activity.go
cretz: Agreed it's not that huge and usually I wouldn't bat an eye (I do tend to use pointers if I am not afraid of mutation and it's more than a couple words in size). However, could travel through many interceptor layers and is the type of call that may happen a lot in an activity. But there are definitely many other places for perf improvement before this...it's just something that always makes me do a double-take when writing.

interceptor/tracing_interceptor.go
Sushisource: OK if that's how it is already then I'm fine leaving it that way. This works a little differently in core but all the spans are already different there anyway so no biggie.

internal/internal_worker.go
Sushisource: Oh, ok I thought that bypass might be fine in both places. Nevermind then.

internal/internal_workflow.go
cretz: I am now thinking the parameter is best as a `converter.EncodedValue`.

internal/client.go
bergundy: Maybe just add a comment warning not to register the same interceptors twice, here and in the worker.

cretz: Will also do that

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,476 @@
...
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/sdk/activity"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	workflowIDTagKey = "temporalWorkflowID"
+	runIDTagKey      = "temporalRunID"
```
bergundy, line 38: We used different tags in Core / TS, these should be the same across all SDKs
CC @Sushisource

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,476 @@
...
+// Tracer is an interface for tracing implementations as used by
+// NewTracingInterceptor. Most callers do not use this directly, but rather use
+// the opentracing or opentelemetry packages.
+type Tracer interface {
+	// SpanContextKey provides a key to put a span on a context unrelated to how a
+	// span might otherwise be put on a context by ContextWithSpan. This should
+	// never be nil.
+	SpanContextKey() interface{}
+
+	// HeaderKey is the key name on the Temporal header to serialize the span to.
```
bergundy, line 50: I'd mention what the default is here and that it should probably not be changed.
It's the same header key for all SDKs.

cretz: I do mention that in each tracing implementations config where this is actually set. This is just the abstraction over OpenTracing and OpenTelemetry and I mention at the top of the interface this interface is not usually used directly.

internal/internal_worker.go
alexshtin: I think I added this "hack" when did a migration from `[]byte` to `Payloads`. It didn't make sense for me but w/o it something didn't work and I failed to track when this func is called two (or maybe more) times instead of 1. Basically, this is a protection against double encoding which needs to be fixed in the first place but I didn't have time for it.

internal/internal_workflow.go
alexshtin: :exploding_head:

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,476 @@
...
+)
+
+// Tracer is an interface for tracing implementations as used by
+// NewTracingInterceptor. Most callers do not use this directly, but rather use
+// the opentracing or opentelemetry packages.
+type Tracer interface {
+	// SpanContextKey provides a key to put a span on a context unrelated to how a
+	// span might otherwise be put on a context by ContextWithSpan. This should
+	// never be nil.
+	SpanContextKey() interface{}
```
cretz, line 48: After use of this in another PR, it has become this should be `Options() TracerOptions` that is invoked one time that has options instead of `SpanContextKey` and `HeaderKey` separate.

bergundy: In typescript, SpanContext key is tied to the baggage and tracing propagators.

cretz: Span context key relates to go contexts, and is tied to the opentracing and opentelemetry tracers.

internal/client.go
cretz: Clarified

interceptor/interceptor.go
cretz: Golint removed from `make check`

internal/internal_activity.go
cretz: TODO removed

internal/internal_workflow.go
cretz: Changed `Arg` type to a raw payload

interceptor/interceptor.go
cretz: Changed

interceptor/tracing_interceptor.go
cretz: Marking resolved, but open to general cross-SDK approach to saying whether span represents local activity or not

interceptor/tracing_interceptor.go
cretz: Marking resolved since it matches existing Go SDK. Open to cross-SDK naming consistency if/when available.

bergundy: Issue to track?

cretz: * https://github.com/temporalio/sdk-go/issues/619
* https://github.com/temporalio/sdk-java/issues/862
* https://github.com/temporalio/sdk-typescript/issues/360

bergundy (approved):
Approving even though I did not read this through.

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,491 @@
...
+	Error error
+}
+
+type tracingInterceptor struct {
+	InterceptorBase
+	tracer  Tracer
+	options TracerOptions
+}
+
+// NewTracingInterceptor creates a new interceptor for the given tracer. Most
```
Sushisource, line 132:
```suggestion
// NewTracingInterceptor creates a new interceptor using the given tracer. Most
```

internal/internal_workflow.go
Sushisource: I think you can get rid of the TODO now?

opentelemetry/tracing_interceptor.go
```diff
@@ -0,0 +1,202 @@
...
+	// Create context with parent
+	var parent trace.SpanContext
+	switch optParent := opts.Parent.(type) {
+	case nil:
+	case *tracerSpan:
+		parent = optParent.SpanContext()
+	case *tracerSpanRef:
+		parent = optParent.SpanContext
+	default:
+		panic(fmt.Sprintf("unrecognized parent type %T", optParent))
```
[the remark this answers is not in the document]
cretz: This would be a compilation error if Go had generics. I had originally assumed the `StartSpan` call would only return runtime errors maybe for missing options or something. But I can return error here if we want.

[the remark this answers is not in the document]
Sushisource: Ah, I see what you mean. Yeah, given this isn't an invariant that really *matters* to us a whole lot, I'm inclined to just return (or even just log) an error rather than exploding.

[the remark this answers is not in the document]
cretz: Done

interceptor/tracing_interceptor.go
cretz: Done

meiliang86:
@cretz Thanks for working with us! I'll take a look at this PR and follow up with you in case there's any further change needed.

internal/interceptor.go
```diff
@@ -0,0 +1,318 @@
...
+// Interceptor is a common interface for all interceptors. See documentation in
+// the interceptor package for more details.
+type Interceptor interface {
+	ClientInterceptor
+	WorkerInterceptor
+}
+
+// WorkerInterceptor is a common interface for all interceptors. See
+// documentation in the interceptor package for more details.
+type WorkerInterceptor interface {
```
yiminc, line 44: Is this used as a factory/provider to create activity/workflow interceptor? Should the name be InboundInterceptorProvider?

internal/interceptor_header.go
```diff
@@ -0,0 +1,134 @@
...
+// WorkflowHeader provides Temporal header information from the workflow context
+// for reading or writing during specific interceptor calls. See documentation
+// in the interceptor package for more details.
+func WorkflowHeader(ctx Context) map[string]*commonpb.Payload {
+	m, _ := ctx.Value(headerKey{}).(map[string]*commonpb.Payload)
+	return m
+}
+
+func workflowContextWithNewHeader(ctx Context) Context {
+	return WithValue(ctx, headerKey{}, map[string]*commonpb.Payload{})
```
yiminc, line 94: can use a const in place of headerKey{}?

internal/internal_workflow.go
```diff
@@ -1400,37 +1396,8 @@ func (h *queryHandler) execute(input *commonpb.Payloads) (result *commonpb.Paylo
...
-	if err != nil {
-		return nil, fmt.Errorf("unable to decode the input for queryType: %v, with error: %w", h.queryType, err)
-	}
-	args = append(args, decoded...)
-
-	// invoke the query handler with arguments.
-	fnValue := reflect.ValueOf(h.fn)
-	retValues := fnValue.Call(args)
-
 	// we already verified (in validateHandlerFn()) that the query handler returns 2 values
```
yiminc, line 1399: no longer need this comment

internal/internal_workflow_client.go
```diff
@@ -1170,3 +950,273 @@ func serializeSearchAttributes(input map[string]interface{}) (*commonpb.SearchAt
...
+
+	request := &workflowservice.SignalWorkflowExecutionRequest{
+		Namespace: w.client.namespace,
+		WorkflowExecution: &commonpb.WorkflowExecution{
+			WorkflowId: in.WorkflowID,
+			RunId:      in.RunID,
+		},
+		SignalName: in.SignalName,
+		Input:      input,
+		Identity:   w.client.identity,
```
yiminc, line 1070: we should set RequestId here, I realize that we didn't set it in old code either, but that is not right.

internal/internal_workflow_client.go
```diff
@@ -1170,3 +950,273 @@ func serializeSearchAttributes(input map[string]interface{}) (*commonpb.SearchAt
...
+}
+
+func (w *workflowClientInterceptor) CancelWorkflow(ctx context.Context, in *ClientCancelWorkflowInput) error {
+	request := &workflowservice.RequestCancelWorkflowExecutionRequest{
+		Namespace: w.client.namespace,
+		WorkflowExecution: &commonpb.WorkflowExecution{
+			WorkflowId: in.WorkflowID,
+			RunId:      in.RunID,
+		},
+		Identity: w.client.identity,
```
yiminc, line 1174: and the RequestId here as well.

internal/internal_workflow_client.go
```diff
@@ -1170,3 +950,273 @@ func serializeSearchAttributes(input map[string]interface{}) (*commonpb.SearchAt
...
+		Args:       in.Args,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return result.QueryResult, nil
+}
+
+// Required to implement ClientOutboundInterceptor
+func (*workflowClientInterceptor) mustEmbedClientOutboundInterceptorBase() {}
```
yiminc, line 1222: Where is ResetWorkflowExecution?

internal/workflow.go
```diff
@@ -383,35 +382,40 @@ func NewFuture(ctx Context) (Future, Settable) {
...
-		// []byte arguments are not serialized
-		switch arg.(type) {
-		case []byte:
-			args = append(args, reflect.ValueOf(arg))
-		default:
-			args = append(args, reflect.ValueOf(arg).Elem())
+func (wc *workflowEnvironmentInterceptor) HandleQuery(ctx Context, in *HandleQueryInput) (interface{}, error) {
+	eo := getWorkflowEnvOptions(ctx)
+	handler, ok := eo.queryHandlers[in.QueryType]
+	// Should never happen because its presence is checked before this call too
```
yiminc, line 398: if the key not present, what error does the check before here returns? we rely on the error message below to tell what are the supported query types in error message.

Makefile
```diff
@@ -115,7 +115,9 @@ fmt:
 clean:
 	rm -rf $(BUILD)
 
-check: lint vet errcheck staticcheck copyright bins
+# goint is intentionally not included in the standard check since it is
```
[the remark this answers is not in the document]
cretz: Will fix

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,491 @@
...
+	// Operation is the general operation name without the specific name.
+	Operation string
+
+	// Name is the specific activity, workflow, etc for the operation.
+	Name string
+
+	// DependedOn is true if the parent depends on this span or false if it just
+	// is related to the parent. In OpenTracing terms, this is true for "ChildOf"
+	// reference types and false for "FollowsFrom" reference types.
+	DependedOn bool
```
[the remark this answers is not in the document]
cretz: These are only for legacy OpenTracing support. This is unused for OpenTelemetry and I expect would be unused for any further tracing implementation. I do not think an enum would be better here.

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,491 @@
...
+
+	// Finish is called when the span is complete.
+	Finish(*TracerFinishSpanOptions)
+}
+
+// TracerFinishSpanOptions are options for TracerSpan.Finish.
+type TracerFinishSpanOptions struct {
+	// Error is present if there was an error in the code traced by this specific
+	// span.
+	Error error
```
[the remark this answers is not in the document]
cretz: I believe tracing span errors are only for the span itself, not inherited. UIs may show the full trace as errored when some nested child has an error, but the error is not placed on every span itself.

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,491 @@
...
+	root *tracingInterceptor
+}
+
+func (t *tracingClientOutboundInterceptor) ExecuteWorkflow(
+	ctx context.Context,
+	in *ClientExecuteWorkflowInput,
+) (client.WorkflowRun, error) {
+	// Start span and write to header
+	span, ctx, err := t.root.startSpanFromContext(ctx, &TracerStartSpanOptions{
+		Operation: "StartWorkflow",
```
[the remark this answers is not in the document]
cretz: It should be `StartWorkflow` and we are using `StartWorkflow` for outbound and `RunWorkflow` for inbound. This matches Java.

interceptor/tracing_interceptor.go
```diff
@@ -0,0 +1,491 @@
...
+	return t.Next.ExecuteActivity(ctx, activityType, args...)
+}
+
+func (t *tracingWorkflowOutboundInterceptor) ExecuteLocalActivity(
+	ctx workflow.Context,
+	activityType string,
+	args ...interface{},
+) workflow.Future {
+	// Start span writing to header
+	span, ctx, err := t.startNonReplaySpan(ctx, "StartActivity", activityType, true)
```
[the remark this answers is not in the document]
cretz: Can you clarify why? Why not match Java? If local or non-local is important, should it be a tag instead? My goal was to reach parity, and if we want to come up with a new common set of names, I think it should be done at a general level and not just in this review just for Go.

internal/interceptor.go
cretz: Why? It is an interceptor, conceptually, whether it provides other interceptors or directly intercepts. This matches other SDK naming _and_ existing Go SDK naming. This was intentionally changed from `WorkflowInterceptorFactory` to `WorkerInterceptor` over a year ago in #169 and this naming was approved in the proposal at https://github.com/temporalio/proposals/pull/45.

internal/interceptor_header.go
cretz: Why? Using strings as context keys is bad practice and we need to stop doing that in other places.

From https://pkg.go.dev/context#WithValue:

> The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context. 

From https://go.dev/blog/context:

> To avoid key collisions, userip defines an unexported type key and uses a value of this type as the context key

There are plenty of other articles on this and Golint even warns on it IIRC.

internal/internal_workflow.go
cretz: While it still applies for why we are not validating the function return types here, I will remove because `executeFunction` also validates.

internal/internal_workflow_client.go
cretz: I don't think this PR is the place for such fixes. There are at least 5 places where we set the `Identity` proto fields of requests as the `client.identity`. Can you open an issue detailing the problem and solution?

internal/internal_workflow_client.go
cretz: Also commented at https://github.com/temporalio/sdk-go/pull/610#discussion_r746670411. I think we need a new issue for this instead of inside this PR.

internal/internal_workflow_client.go
cretz: There are many client calls that are not intercepted. From https://github.com/temporalio/proposals/pull/45:

> We intentionally choose to only allow a subset of client calls to be mocked as these are the ones considered "workflow client" calls by the SDKs

Every other SDK has only a few high-level client calls in what they call a "workflow client" and then asks the user to use gRPC for the rest. But Go is different in that it hides gRPC adds many other high-level client abstractions. It was decided for consistency reasons to only intercept the client calls that all SDKs allow to be intercepted instead of intercepting every client call.

internal/workflow.go
cretz: See `RegisterQueryHandler` in `workflow_internal.go` which does this same logic and sends back this same error as it always has before calling the interceptors which trickle down to this call.

Also note that we are adding #601 at some point to stop using error messages for such things.

internal/internal_activity.go
```diff
@@ -302,98 +302,131 @@ func isActivityContext(inType reflect.Type) bool {
...
-
-	if resultSize < 1 || resultSize > 2 {
-		fnName, _ := getFunctionName(f)
-		return nil, fmt.Errorf(
-			"the function: %v signature returns %d results, it is expecting to return either error or (result, error)",
-			fnName, resultSize)
+func setActivityParametersIfNotExist(ctx Context) Context {
+	params := getActivityOptions(ctx)
+	var newParams ExecuteActivityOptions
+	if params != nil {
```
meiliang86, line 308: Curious, what's the reason that you want to clone the old option?

internal/internal_activity.go
```diff
@@ -302,98 +302,131 @@ func isActivityContext(inType reflect.Type) bool {
...
 
-func validateFunctionAndGetResults(f interface{}, values []reflect.Value, dataConverter converter.DataConverter) (*commonpb.Payloads, error) {
-	resultSize := len(values)
-
-	if resultSize < 1 || resultSize > 2 {
-		fnName, _ := getFunctionName(f)
-		return nil, fmt.Errorf(
-			"the function: %v signature returns %d results, it is expecting to return either error or (result, error)",
-			fnName, resultSize)
+func setActivityParametersIfNotExist(ctx Context) Context {
```
meiliang86, line 305: nit: s/Parameters/Options for consistency?
Same for the local var names

internal/internal_activity.go
```diff
@@ -302,98 +302,131 @@ func isActivityContext(inType reflect.Type) bool {
...
-					return nil, err
-				}
-			}
-		}
+func getActivityEnvironmentInterceptor(ctx context.Context) *activityEnvironmentInterceptor {
+	a := ctx.Value(activityEnvInterceptorContextKey)
+	if a == nil {
+		panic("getActivityEnvironmentInterceptor: Not an activity context")
 	}
+	return a.(*activityEnvironmentInterceptor)
```
meiliang86, line 339: need to check type here?

internal/internal_workflow_client.go
```diff
@@ -74,7 +73,9 @@ type (
 		identity           string
 		dataConverter      converter.DataConverter
 		contextPropagators []ContextPropagator
-		tracer             opentracing.Tracer
+		workerInterceptors []WorkerInterceptor
```
meiliang86, line 76: nit: remove extra line?

internal/internal_workflow_client.go
```diff
@@ -333,162 +229,63 @@ func (wc *WorkflowClient) GetWorkflow(ctx context.Context, workflowID string, ru
...
-	// Validate type and its arguments.
-	workflowType, input, err := getValidatedWorkflowFunction(workflowFunc, workflowArgs, dataConverter, wc.registry)
-	if err != nil {
-		return nil, err
-	}
-
-	memo, err := getWorkflowMemo(options.Memo, wc.dataConverter)
-	if err != nil {
-		return nil, err
+	// Default workflow ID
```
meiliang86, line 245: How could user not knowing the workflowID if the intention is to do SignalWithStart?

internal/internal_workflow_client.go
```diff
@@ -74,7 +73,9 @@ type (
 		identity           string
 		dataConverter      converter.DataConverter
 		contextPropagators []ContextPropagator
-		tracer             opentracing.Tracer
+		workerInterceptors []WorkerInterceptor
+
+		interceptor ClientOutboundInterceptor
```
meiliang86, line 78: nit: naming: Call this clientInterceptor?

internal/internal_workflow_client.go
```diff
@@ -1170,3 +950,273 @@ func serializeSearchAttributes(input map[string]interface{}) (*commonpb.SearchAt
 	}
 	return &commonpb.SearchAttributes{IndexedFields: attr}, nil
 }
+
+type workflowClientInterceptor struct{ client *WorkflowClient }
```
meiliang86, line 954: Sorry, I might miss something obvious. Is this new or old code that got moved from somewhere?
If it's new what's the purpose?

internal/workflow_test.go
```diff
@@ -108,13 +108,17 @@ func TestConvertRetryPolicy(t *testing.T) {
...
-	return newWorkflowContext(&workflowEnvironmentImpl{
+	_, ctx, err := newWorkflowContext(&workflowEnvironmentImpl{
 		dataConverter: converter.GetDefaultDataConverter(),
 		workflowInfo: &WorkflowInfo{
 			Namespace:     "default",
 			TaskQueueName: "default",
 		},
-	}, nil, nil)
+	}, nil)
+	if err != nil {
```
meiliang86, line 118: why not using assert?

opentelemetry/tracing_interceptor.go
```diff
@@ -0,0 +1,202 @@
...
+	return interceptor.TracerOptions{
+		SpanContextKey: t.options.SpanContextKey,
+		HeaderKey:      t.options.HeaderKey,
+	}
+}
+
+func (t *tracer) UnmarshalSpan(m map[string]string) (interceptor.TracerSpanRef, error) {
+	ctx := trace.SpanContextFromContext(t.options.TextMapPropagator.Extract(context.Background(), textMapCarrier(m)))
+	if !ctx.IsValid() {
+		return nil, nil
```
meiliang86, line 126: Shouldn't this return error?

opentelemetry/tracing_interceptor.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	return &tracerSpanRef{SpanContext: ctx}, nil
+}
+
+func (t *tracer) MarshalSpan(span interceptor.TracerSpan) (map[string]string, error) {
+	data := textMapCarrier{}
+	t.options.TextMapPropagator.Inject(trace.ContextWithSpan(context.Background(), span.(*tracerSpan).Span), data)
```
meiliang86, line 133: Could this return error?

opentelemetry/tracing_interceptor.go
meiliang86: or we choose to silently drop the span?

opentelemetry/tracing_interceptor.go
```diff
@@ -0,0 +1,202 @@
...
+	return map[string]string(data), nil
+}
+
+func (t *tracer) SpanFromContext(ctx context.Context) interceptor.TracerSpan {
+	span := trace.SpanFromContext(ctx)
+	if !span.SpanContext().IsValid() {
+		return nil
```
meiliang86, line 140: Add some debug log?

opentracing/interceptor.go
```diff
@@ -0,0 +1,174 @@
...
+	HeaderKey string
+
+	// SpanStarter is a callback to create spans. If not set, this creates normal
+	// OpenTracing spans calling Tracer.StartSpan.
+	SpanStarter func(t opentracing.Tracer, operationName string, opts ...opentracing.StartSpanOption) opentracing.Span
+}
+
+type spanContextKey struct{}
+
+const defaultHeaderKey = "_tracer-data"
```
meiliang86, line 55: Not sure if this can still be changed (backward compatibility?), but if we still could, consider adding "temporal" in the string?

meiliang86 (commented):
Thanks for cleaning this up, Chad!

internal/interceptor_header.go
```diff
@@ -0,0 +1,134 @@
...
+		var err error
+		if ctx, err = ctxProp.Extract(ctx, reader); err != nil {
+			return nil, fmt.Errorf("failed propagating header: %w", err)
+		}
+	}
+	return context.WithValue(ctx, headerKey{}, header.Fields), nil
+}
+
+func headerPropagated(ctx context.Context, ctxProps []ContextPropagator) (*commonpb.Header, error) {
+	header := &commonpb.Header{Fields: Header(ctx)}
```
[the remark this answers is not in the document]
cretz: Are you saying like this?

```go
fields := Header(ctx)
if fields == nil {
  return nil, fmt.Errorf("context missing header")
}
header := &commonpb.Header{Fields: fields}
```

Or you want me to call `Header` twice? (granted it's cheap, but probably not much benefit over how it sits now)

internal/internal_activity.go
cretz: This is not my code. This remained unchanged from further down in the file. I would not have written the coed like this. Unfortunately git diff is confused with all of the changes in here.

internal/internal_activity.go
cretz: This is not my code, it remains unchanged. Git diff is struggling with this file.

internal/internal_activity.go
cretz: No more than we do in any other context `Value` lookup where we control the key and the setting internally and therefore know the type

internal/internal_workflow_client.go
cretz: :+1: Will remove

internal/internal_workflow_client.go
cretz: I think this stutters, for the same reason we don't call the fields `clientIdentity` or `clientNamespace`.

internal/internal_workflow_client.go
cretz: This is existing logic. Nonetheless, if `SignalWithStart` = `Start` + `Signal`, and `Signal` uses `Start`'s workflow ID, and `Start` defaults to an auto-generated workflow ID, then one might expect `SignalWithStart` to work the way `Start` does wrt workflow ID.

I believe `SignalWithStart` can work just fine with auto-generated workflow ID just like if you manually wrote it to start then signal with the auto-generated ID from start.

internal/internal_workflow_client.go
cretz: This is a new type. Client interceptors are a new thing. This is the implementation of last outbound interceptor in the client chain.

internal/workflow_test.go
cretz: As an existing test helper function, I figured there was little value in changing existing callers to pass in a testing context for assertion and I also figured there was little value in changing existing callers to handle an error that should never happen. This is the simplest way for this test helper (I could just swallow the error I know will not happen, but we might as well panic to be future proof to other uses of the error).

opentelemetry/tracing_interceptor.go
cretz: There may be no serialized span. There is a difference between an unmarshal failure and a nil span ref because it doesn't exist. I can make this clearer on the `interceptor.Tracer.UnmarshalSpan` interface docs.

cretz: I had trouble deciding here. With opentracing I do error, so I can error here too.

opentelemetry/tracing_interceptor.go
cretz: No, see https://pkg.go.dev/go.opentelemetry.io/otel/propagation#TextMapPropagator

opentelemetry/tracing_interceptor.go
cretz: There can be plenty of normal cases where a context doesn't have a span yet. From `interceptor.Tracer.SpanFromContext` docs:

> SpanFromContext returns the span from the general Go context or nil if not present.

Absence of a span is considered normal.

opentracing/interceptor.go
cretz: I have chosen not to change it but instead allow it to be customized. We also want this to work across SDKs, and Java uses this same string.

Makefile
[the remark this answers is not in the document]
cretz: Fixed

internal/internal_workflow.go
cretz: Fixed

internal/internal_workflow_client.go
cretz: Removed

cretz:
Marking outstanding conversations sans response since yesterday as "resolved"

opentelemetry/tracing_interceptor.go
cretz: Changed to return error.

interceptor/tracing_interceptor.go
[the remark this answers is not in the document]
yiminc: Oh, didn't know that Java side uses StartActivity for local activity. Ideally, I would prefer to have the ability to differentiate between activity and local activity. There is significant latency difference between them.

internal/interceptor_header.go
yiminc: Not a string. I was referring to const headerKey = headerKey{}, but since it is a value type, so I guess it does not matter.

internal/internal_workflow_client.go
yiminc: https://github.com/temporalio/sdk-go/issues/627

interceptor/tracing_interceptor.go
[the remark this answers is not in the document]
cretz: This can be done as a tag too. But I think we should solve these things as cross-SDK-tracing consistency issues and not specific to this review.

yiminc (approved):

internal/interceptor_header.go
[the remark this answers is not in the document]
meiliang86: Yes I meant the above.

internal/internal_activity.go
meiliang86: ok thanks

meiliang86 (approved):

internal/interceptor_header.go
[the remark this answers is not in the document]
cretz: Adding a line and a field doesn't seem necessary to save the struct creation that I expect will always be needed (I don't expect error to ever happen). But if the way I have it doesn't feel right, I can change.

PR #612, opened 2021-11-01
Document ambiguous ID and test for mismatch on signal start

### LINKED ISSUE #603: Document/enforce `SignalWithStartWorkflow` does not use same workflow ID option as `ExecuteWorkflow`

`SignalWithStartWorkflow` uses the `workflowID` parameter like `SignalWorkflow` instead of the ID in options like `ExecuteWorkflow`. That means the function accepts ID in two ways and ignores one.

Document that the ID in options is disallowed and error if it is provided and is different than the `workflowID` parameter.

cretz:
## What was changed

Document different ways to give workflow ID on signal with start, and ensure the unused one doesn't mismatch.

## Why?

We ignore workflow IDs coming from options, so we make sure it's not expecting to be used.

## Checklist

1. Closes #603


internal/internal_workflow_client.go
```diff
@@ -367,6 +367,12 @@ func (wc *WorkflowClient) SignalWithStartWorkflow(ctx context.Context, workflowI
 		return nil, err
 	}
 
+	// Due to the ambiguous way to provide workflow IDs, if options contains an
+	// ID, it must match the parameter
+	if options.ID != "" && options.ID != workflowID {
+		return nil, fmt.Errorf("workflow ID from options not used, must be unset or match workflow ID parameter")
```
bergundy, line 373: This is technically a breaking change.
I think it's necessary but we should document this in the release notes and mark it as breaking.

bergundy (approved):

internal/internal_workflow_client.go
cretz: It's only a breaking change if you consider extra validation of (unbeknownst) improper param setting as a breaking change. It went from being an invalid param do the wrong thing to an invalid param error.

bergundy: I'm aware, but this still could result in a change of behavior in some rare cases.

cretz: :+1: Ok, I will make sure to document in the release notes

PR #613, opened 2021-11-02
Fix testsuite child cron job running forever with parent policy abandon

tminusplus:
## What was changed

The testsuite will force a child cron job with parent policy abandon to terminate as soon as the parent workflow terminates.

## Why?

If you start a parent workflow, which creates a child workflow as a cronjob with parent policy abandon it spins forever. This happens because the timer auto-skips forever, which in-turn means the child cron-job will run forever.

## How was this tested
- [x] Ran all testsuites, including the newly added one to test this scenario

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -919,7 +919,11 @@ func (env *testWorkflowEnvironmentImpl) handleParentClosePolicy() {
 
 			switch handle.params.ParentClosePolicy {
 			case enumspb.PARENT_CLOSE_POLICY_ABANDON:
-				// noop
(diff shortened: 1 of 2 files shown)


tminusplus:
👋 I wanted to put up a PR to show an issue I encountered when using the testsuite. Will be happy if we need to rework the solution, or if it ends up that there is a good way to handle this case in a test.

tminusplus:
Wires got crossed, this does not demonstrate / fix the described issue as the added test isn't set with PARENT_CLOSE_POLICY_ABANDON. Additionally the fix doesn't work as intended. I'm going to open up an issue with this as an example and we can discuss there :).

PR #620, opened 2021-11-08
Header support for signals and queries

cretz:
## What was changed

Support headers in interceptors for signals and queries. This requires https://github.com/temporalio/api/pull/127 to be merged to compile and https://github.com/temporalio/temporal/pull/2149 to be merged to actually work instead of silently not work.

## Why?

Header metadata is propagated via headers from where workflows/activities are started and where they are ran. This does the same for signals and queries.

## How was this tested:

This was tested not only with several integration tests, but also confirming running this version of the SDK on an older server without support would be safe (but the functionality silently won't work).


cretz:
This PR is a draft and looks bigger than it is because #610 isn't merged and this builds off that branch.

go.mod
```diff
@@ -3,28 +3,34 @@ module go.temporal.io/sdk
...
 	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac
-	google.golang.org/grpc v1.40.0
+	google.golang.org/grpc v1.41.0
 	google.golang.org/protobuf v1.27.1
+	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+// TODO(cretz): Temporary, remove and update tag in require section before merge
+replace go.temporal.io/api => ../temporal-api-go
```
cretz, line 36: Do not merge this until https://github.com/temporalio/api/pull/127 is merged/tagged.

cretz:
I am actively fixing newly appearing unit test failures since my latest master merge...

cretz:
I am closing this draft and opening legit PR

PR #625, opened 2021-11-10
State machine issue after activity cancellation

### LINKED ISSUE #623: Activity cancellation in certain scenarios causes event handling failure
## Expected Behavior

Activity cancellation should work.

## Actual Behavior

Activity cancellation has an event problem when unwinding. This was introduced in v1.10.0 in #504.

I have not dug _too_ deeply, but https://github.com/temporalio/sdk-go/compare/master...cretz:sdk-history-bug shows the bug (FYI, branch may be deleted later). It shows the commented out lines make the test succeed. This was noticed in a couple scenarios, but we only replicated it when https://github.com/temporalio/sdk-go/issues/481 was updated by a user.

cretz:
## What was changed

A state machine failure is occurring when anything is done in a workflow after the workflow was canceled which caused activity cancellation. This essentially reverts the activity part of #504 while keeping the timer part in place.

## Why?

In cases where extra things happened after activity cancellation, a state machine bug occurred.

**While this is not believed to have any effect on history compatibility across versions, reviewers are requested to take a close look and give their own opinion.**

## Checklist

1. Closes #623

### DIFF
--- a/test/workflow_test.go
+++ b/test/workflow_test.go
@@ -1356,7 +1356,15 @@ func (w *Workflows) SleepForDuration(ctx workflow.Context, d time.Duration) erro
 	return workflow.Sleep(ctx, d)
 }
 
-func (w *Workflows) ActivityPostCancellation(ctx workflow.Context) error {
+type AdvancedPostCancellationInput struct {
+	PreCancelActivity  bool
+	PostCancelActivity bool
+	PreCancelTimer     bool
+	PostCancelTimer    bool
+}
+
+func (w *Workflows) AdvancedPostCancellation(ctx workflow.Context, in *AdvancedPostCancellationInput) error {
+	// Setup query to tell caller we're waiting for cancel
 	waitingForCancel := false
 	err := workflow.SetQueryHandler(ctx, "waiting-for-cancel", func() (bool, error) {
 		return waitingForCancel, nil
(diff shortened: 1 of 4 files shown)


test/workflow_test.go
```diff
@@ -1356,6 +1356,55 @@ func (w *Workflows) SleepForDuration(ctx workflow.Context, d time.Duration) erro
...
+	if err := fut.Get(ctx, nil); err != nil {
+		return fmt.Errorf("activity did not gracefully cancel: %w", err)
+	}
+
+	// Run another activity after cancel
+	ctx, _ = workflow.NewDisconnectedContext(ctx)
+	return workflow.ExecuteActivity(ctx, a.Sleep, 1*time.Millisecond).Get(ctx, nil)
+}
+
+func (w *Workflows) TimerPostCancellation(ctx workflow.Context) error {
```
Sushisource, line 1387: I think another useful test to add would be both activity and timer with this pattern.

Possibly one with more than one of each type as well

cretz: I have added all the permutations. It does add 5 seconds onto integration tests unfortunately.

Sushisource (approved):
Awesome! Thanks Chad

yiminc (approved):

PR #632, opened 2021-11-16
Add ability to not wait on child workflows in test suite.

### LINKED ISSUE #614: Testing cron child workflow with parent policy abandon runs forever
## Expected Behavior
Testing a cron child workflow with PARENT_CLOSE_POLICY_ABANDON should either close with the parent workflow for testsuite or have a method to shutdown the child workflow for the test.

## Actual Behavior
The cron child workflow runs forever due to the auto-fire timer and then panics when the test times out.

Originally produced while updating go-sdk in a repo. There is a test below for internal_workflow_testsuite_test.go which can reproduce the issue
```
func (s *WorkflowTestSuiteUnitTest) Test_CronChildWorkflowWithParentPolicyAbandon() {
	failedCount, successCount, lastCompletionResult := 0, 0, 0
	cronWorkflow := func(ctx Context) (int, error) {
		info := GetWorkflowInfo(ctx)
		var result int
		if HasLastCompletionResult(ctx) {
			_ = GetLastCompletionResult(ctx, &result)
		}
		_ = Sleep(ctx, time.Second*3)
		if info.Attempt == 1 {
			failedCount++
			return 0, errors.New("please-retry")
		}

		successCount++
		result++
		lastCompletionResult = result
		return result, nil
	}

	testWorkflow := func(ctx Context) error {
		ctx1 := WithChildWorkflowOptions(ctx, ChildWorkflowOptions{
			WorkflowRunTimeout: time.Minute * 10,
			RetryPolicy: &RetryPolicy{
				MaximumAttempts:        5,
				InitialInterval:        time.Second,
				MaximumInterval:        time.Second * 10,
				BackoffCoefficient:     2,
				NonRetryableErrorTypes: []string{"bad-bug"},
			},
			ParentClosePolicy: enumspb.PARENT_CLOSE_POLICY_ABANDON,
			CronSchedule:      "0 * * * *", // hourly
		})

		cronFuture := ExecuteChildWorkflow(ctx1, cronWorkflow) // cron never stop so this future won't return

		timeoutTimer := NewTimer(ctx, time.Hour*3)
		selector := NewSelector(ctx)
		var err error
		selector.AddFuture(cronFuture, func(f Future) {
			err = errors.New("cron workflow returns, this is not expected")
		}).AddFuture(timeoutTimer, func(f Future) {
			// err will be nil
		}).Select(ctx)

		return err
	}

	env.RegisterWorkflow(cronWorkflow)
	env.RegisterWorkflow(testWorkflow)

	startTime, _ := time.Parse(time.RFC3339, "2018-12-20T16:30:00+08:00")
	env.SetStartTime(startTime)
	env.ExecuteWorkflow(testWorkflow)

	s.True(env.IsWorkflowCompleted())
	err := env.GetWorkflowError()
	s.NoError(err)

	s.Equal(4, failedCount)
	s.Equal(4, successCount)
	s.Equal(4, lastCompletionResult)
}
```

Then you can run it with `go test -run "TestUnitTestSuite/Test_CronChildWorkflowWithParentPolicyAbandon"` in `sdk-go/internal`

## Specifications

  - Version: v1.11.0
  - Platform: Mac v11.6

cretz:
## What was changed

Added `TestWorkflowEnvironment.SetDetachedChildWait` that lets test environments not wait on detached (i.e. abandoned or request-cancel) children before finishing workflow.

## Why?

In #470, we introduced a change that causes a parent `ExecuteWorkflow` to not return until all of its children are completely done. This caused issues for people testing interminable child workflows such as ones started on cron because they never finish.

Alternatively, we could have done something like expose child test workflow environments so people could cancel children themselves. This seems more straightforward however.

## Checklist

1. Closes #614

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -132,12 +132,13 @@ type (
 
 		taskQueueSpecificActivities map[string]*taskQueueSpecificActivity
 
-		mock               *mock.Mock
-		service            workflowservice.WorkflowServiceClient
-		logger             log.Logger
-		metricsScope       tally.Scope
-		contextPropagators []ContextPropagator
-		identity           string
+		mock                      *mock.Mock
+		service                   workflowservice.WorkflowServiceClient
+		logger                    log.Logger
+		metricsScope              tally.Scope
+		contextPropagators        []ContextPropagator
+		identity                  string
+		detachedChildWaitDisabled bool
 
 		mockClock *clock.Mock
 		wallClock clock.Clock
@@ -337,6 +338,7 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(param
 	childEnv.workerOptions = env.workerOptions
 	childEnv.dataConverter = params.DataConverter
 	childEnv.registry = env.registry
+	childEnv.detachedChildWaitDisabled = env.detachedChildWaitDisabled
 
 	if params.TaskQueueName == "" {
 		return nil, serviceerror.NewWorkflowExecutionAlreadyStarted("Empty task queue name", "", "")
@@ -411,6 +413,10 @@ func (env *testWorkflowEnvironmentImpl) setContextPropagators(contextPropagators
 
 func (env *testWorkflowEnvironmentImpl) setWorkerStopChannel(c chan struct{}) {
 	env.workerStopChannel = c
+}
+
+func (env *testWorkflowEnvironmentImpl) setDetachedChildWaitDisabled(detachedChildWaitDisabled bool) {
+	env.detachedChildWaitDisabled = detachedChildWaitDisabled
 }
 
 func (env *testWorkflowEnvironmentImpl) setActivityTaskQueue(taskqueue string, activityFns ...interface{}) {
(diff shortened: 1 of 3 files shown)


internal/workflow_testsuite.go
```diff
@@ -539,6 +539,19 @@ func (e *TestWorkflowEnvironment) SetIdentity(identity string) *TestWorkflowEnvi
 	return e
 }
 
+// SetDetachedChildWait, if true, will make ExecuteWorkflow wait on all child
+// workflows to complete even if their close policy is set to abandon or request
+// cancel. If false, ExecuteWorkflow will wait on all children to complete
+// regardless of detached child. In situations with interminable cron-based
```
bergundy, line 545: This sentence is unclear to me:
> regardless of detached child
Can you please clarify?

internal/workflow_testsuite.go
```diff
@@ -539,6 +539,19 @@ func (e *TestWorkflowEnvironment) SetIdentity(identity string) *TestWorkflowEnvi
...
+// child workflows, setting this to false will allow the child to run once but
```
bergundy, line 546: Also the phrasing here is confusing, what does "allow" mean? Will it always wait for at least a single execution?

bergundy (commented):
Is this the right approach?
Wouldn't an explicit shutdown call when all test assertions are fulfilled be a better experience?

internal/workflow_testsuite.go
cretz: It means, regardless of whether the child is detached or not, `ExecuteWorkflow` will wait for all to complete. This is the default behavior today and the problematic one.

cretz:
>  Wouldn't an explicit shutdown call when all test assertions are fulfilled be a better experience?

The `ExecuteWorkflow` call never completes because there are child workflows that never complete. They used to complete before #470 was merged.

internal/workflow_testsuite.go
tszucs-stripe: It is a tricky thing to describe. Perhaps something like this might be clearer?
> If false, ExecuteWorkflow will block until all attached child workflows have completed and all detached child workflows have ran only once. This is useful when testing endless cron-based child workflows, as without it a detached child workflow would run forever.

tszucs-stripe (approved):
Not a code owner but this makes sense, I think enables users to test both situations, and does so in a simple manner with a bool flag.

internal/workflow_testsuite.go
cretz: Clarified

internal/workflow_testsuite.go
cretz: Clarified.

cretz:
I have now learned there is no guarantee the child will even run, it depends on order of selector and other things. Going to make it clear in the docs.

internal/internal_workflow_testsuite.go
```diff
@@ -675,15 +681,18 @@ func (env *testWorkflowEnvironmentImpl) startMainLoop() {
...
 
-		if !handle.handled && (handle.params.ParentClosePolicy == enumspb.PARENT_CLOSE_POLICY_ABANDON ||
-			handle.params.ParentClosePolicy == enumspb.PARENT_CLOSE_POLICY_REQUEST_CANCEL) {
-			return false
+			if !handle.handled && (handle.params.ParentClosePolicy == enumspb.PARENT_CLOSE_POLICY_ABANDON ||
+				handle.params.ParentClosePolicy == enumspb.PARENT_CLOSE_POLICY_REQUEST_CANCEL) {
+				return false
+			}
 		}
 	}
```
Sushisource, line 697: Can this ever return true?

Sushisource: Oh nevermind github just cut off the last line

Sushisource (approved):

PR #635, opened 2021-11-16
Dependency abstraction

cretz:
## What was changed

Move OpenTracing, OpenTelemetry, and Tally to their own modules. For the latter, this required a full abstraction of our use of metrics.

Breaking changes:

* Interceptors in master already cause breaking changes with OpenTracing
* This will cause people to use a separate dependency if they want `Tally`. So instead of `ClientOptions.MetricsScope = myTallyScope`, they'll need `ClientOptions.MetricsHandler = tally.NewMetricsHandler(myTallyScope)` and they'll have to `go get go.temporal.io/sdk/contrib/tally`.
* Any code using `GetMetricsScope` will now need `GetMetricsHandler`

## Why?

Forcing users to use certain dependencies causes issues. This abstraction allows them to choose what to use.

This is just a draft and is intentionally missing docs and tests. I am looking for feedback.


internal/common/metrics/handler.go
```diff
@@ -0,0 +1,103 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package metrics
+
+import "time"
+
+type Handler interface {
+	WithTags(map[string]string) Handler
```
yiminc, line 28: why not keep the names: Handler -> Scope, WithTags -> Tagged?

cretz: Why? Those are Tally'isms. `WithX` is common nomenclature as is a handler. I don't think `Scope` and `Tagged` apply.

client/client.go
```diff
@@ -403,6 +404,13 @@ type (
 	}
 )
 
+type MetricsHandler = metrics.Handler
+type MetricsCounter = metrics.Counter
+type MetricsGauge = metrics.Gauge
+type MetricsTimer = metrics.Timer
```
mmcshane, line 410: I had understood that we do not expect client code to use our internal instrument types. I.e. the  expectation is that they instrument and export their own metrics with whatever tools they prefer. Do I have that right?

contrib/opentelemetry/go.mod
```diff
@@ -0,0 +1,13 @@
+module go.temporal.io/sdk/contrib/opentelemetry
+
+go 1.16
+
+require (
+	github.com/stretchr/testify v1.7.0
+	go.opentelemetry.io/otel v1.1.0
```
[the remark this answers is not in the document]
cretz: We can definitely grab the latest. And this is the type of thing that this change facilitates! We can upgrade third party libs without forcing users of the primary library to do so.

internal/common/metrics/capturing_handler.go
```diff
@@ -0,0 +1,146 @@
...
+
+package metrics
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// This file contains test helpers only. They are not private because they are used by other tests.
```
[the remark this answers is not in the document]
cretz: I had considered that, but it wasn't worth it for this tiny thing that mimics what was there (mostly). If it were exposed I'd consider it. I did this with `internal/interceptortest` so it's not completely inconsistent with best practice. But arranging of internal packages isn't subject to compatibility rules, so I can change later.

client/client.go
cretz: I believe that is correct. This abstraction is very specific to our uses. This is not a general-purpose metrics abstraction. We definitely expect their metric use to leverage whatever they back the abstraction with (e.g. use the Tally scope directly), not the abstraction itself.

Technically they _could_ use our abstraction if they're uses were as limited as ours (e.g. no histograms). We do expose `activity.GetMetricsHandler` and `workflow.GetMetricsHandler` for this purpose which does have replay protection and some tags built-in, but unless they really need those things, I'd discourage it. Assuming the team is OK with this PR, I'll make all of this clear in the docs.

cretz:
Closing in preparation for the real thing

PR #637, opened 2021-11-17
Add task pool slots remaining metric

### LINKED ISSUE #594: Metric for determining task pool status

No way to determine via metrics alone that all internal workers are busy and therefore not accepting new tasks.

Need at least a metric showing how many workers are busy/running (consider a label per worker type). If possible, include a metric showing that the system is blocked.

**Additional context**

Note, while a deadlock error does exist for workflows at https://github.com/temporalio/sdk-go/blob/6bbdd9eeef901bf5b7782114724d5ad906ba7f2e/internal/internal_workflow.go#L890 with a timeout, there may not be one for activities and we can consider such an option (keeping in mind that activity tasks are known to be potentially long running).

cretz:
## What was changed

Added worker metric called `temporal_worker_task_slots_available` that shows the remaining available slots for concurrent tasks. This is a "worker" metric as in the internal "worker" concept, not Temporal Workers. This is similar to two other "worker" metrics, `temporal_worker_start` and `temporal_poller_start` in that it has a `worker_type` tag that is either `WorkflowWorker`, `ActivityWorker`, or `LocalActivityWorker`. There are no other meaningful per-worker tags.

## Why?

Requested by users wanting to see if they are using up their max concurrent tasks and waiting.

## Checklist

1. Closes #594

### DIFF
--- a/internal/common/metrics/capturing_stats_reporter.go
+++ b/internal/common/metrics/capturing_stats_reporter.go
@@ -38,7 +38,7 @@ import (
 func NewMetricsScope(isReplay *bool) (tally.Scope, io.Closer, *CapturingStatsReporter) {
 	reporter := &CapturingStatsReporter{}
 	opts := tally.ScopeOptions{Reporter: reporter}
-	scope, closer := tally.NewRootScope(opts, time.Second)
+	scope, closer := tally.NewRootScope(opts, 100*time.Millisecond)
 	return WrapScope(isReplay, scope, &realClock{}), closer, reporter
 }
 
--- a/internal/common/metrics/capturing_stats_reporter.go
+++ b/internal/common/metrics/capturing_stats_reporter.go
@@ -58,7 +58,7 @@ func (c *realClock) Now() time.Time {
 
 // CapturingStatsReporter is a reporter used by tests to capture the metric so we can verify our tests.
 type CapturingStatsReporter struct {
-	sync.Mutex
+	sync.RWMutex
 	counts                   []CapturedCount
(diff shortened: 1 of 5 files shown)


internal/common/metrics/capturing_stats_reporter.go
```diff
@@ -38,7 +38,7 @@ import (
 func NewMetricsScope(isReplay *bool) (tally.Scope, io.Closer, *CapturingStatsReporter) {
 	reporter := &CapturingStatsReporter{}
 	opts := tally.ScopeOptions{Reporter: reporter}
-	scope, closer := tally.NewRootScope(opts, time.Second)
+	scope, closer := tally.NewRootScope(opts, 100*time.Millisecond)
```
Sushisource, line 41: Is this changing reporting / aggregation frequency or something? 10x is a lot, just want to make sure this won't mean we're overloading things

cretz: This is only for a test utility. From a few lines above:

> // This file contains test helpers only. They are not private because they are used by other tests.

Having it here with the rest of the code is a bit confusing, but :shrug:. Maybe I'll take the advice from https://github.com/temporalio/sdk-go/pull/635#discussion_r751457854 and move it off to another package if I'm allowed to do that refactor.

Sushisource: Ah, got it. Makes sense.

Sushisource (approved):

bergundy (approved):
lgtm

PR #643, opened 2021-11-19
Cleaning up inline documentation

flossypurse:
## What was changed

- updates to doc.go pages
- updates to inline doc comments
- README edits, and adding list of "how-to" guides

## Why?

We want a consistent source of truth for term definitions and developer guides.
Updating the comments to be minimalistic and instead point to the sources of truth in our docs site.

## Checklist

1. Closes 

N/A

2. How was this tested:

N/A

3. Any docs updates needed?

N/A (these are docs changes)


README.md
```diff
@@ -1,27 +1,42 @@
...
-or
+Or clone the Go SDK repo to your preferred location:
 
 ```bash
-go get -u go.temporal.io/sdk
+git clone git@github.com:temporalio/sdk-go.git
 ```
+**Are there executable code samples?**
+
+You can find a complete list of executable code samples in the [samples library](/docs/samples-library/#go), which includes Temporal Go SDK code samples from the [temporalio/samples-go](https://github.com/temporalio/samples-go) repo.
```
cretz, line 26: The `samples library` link shouldn't be a relative path here

README.md
```diff
@@ -1,27 +1,42 @@
...
-You can also find the API documentation [here](https://pkg.go.dev/go.temporal.io/sdk).
+
+**Links**
+
+- [Contribution Guidelines](CONTRIBUTING.md)
+- [Temporal Server repository](https://github.com/temporalio/temporal)
+- [Temporal Documentation site](https://docs.temporal.io)
+- [MIT License](LICENSE)
 
 ## Using a custom logger
```
cretz, line 41: Wonder if you want to move this somewhere else. It's strange to just have this one sample in the README and not the dozens of other custom things you could create.

worker/worker.go
```diff
@@ -22,7 +22,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// Package worker contains functions to manage lifecycle of a Temporal client side worker.
+// Package worker contains APIs to manage the lifecycle of Temporal Client-side Worker Entities.
```
cretz, line 25: This should be removed if you're gonna have a `doc.go`

README.md
```diff
@@ -1,27 +1,42 @@
...
+[![Build Status](https://badge.buildkite.com/ce6df3b1a8b375270261ae70fb2d2756af298fef3a0dac4d20.svg?theme=github&branch=master)](https://buildkite.com/temporal/temporal-go-client) [![Coverage Status](https://coveralls.io/repos/github/temporalio/temporal-go-sdk/badge.svg?branch=master)](https://coveralls.io/github/temporalio/temporal-go-sdk?branch=master) [![PkgGoDev](https://pkg.go.dev/badge/go.temporal.io/sdk)](https://pkg.go.dev/go.temporal.io/sdk)
 
-[Temporal](https://github.com/temporalio/temporal) is a distributed, scalable, durable, and highly available orchestration engine used to execute asynchronous long-running business logic in a scalable and resilient way.
+The Temporal Go SDK provides a framework for Temporal Application development in the Go language.
+The SDK contains the following tools:
 
-"Temporal Go SDK" is the framework for authoring workflows and activities using Go language.
+- A Temporal Client to communicate with a Temporal Cluster
+- APIs to use within your Workflows
+- APIs to create and manage Worker Entities and Worker Processes
```
cretz, line 9: What are "entities" and "processes" in this context?

worker/worker.go
```diff
@@ -39,22 +39,23 @@ import (
...
+		// Start the Worker in a non-blocking fashion.
 		Start() error
 
-		// Run the worker in a blocking fashion. Stop the worker when interruptCh receives signal.
-		// Pass worker.InterruptCh() to stop the worker with SIGINT or SIGTERM.
-		// Pass nil to stop the worker with external Stop() call.
-		// Pass any other `<-chan interface{}` and Run will wait for signal from that channel.
-		// Returns error only if worker fails to start.
+		// Run the Worker in a blocking fashion.
+		// The Worker stops when interruptCh receives a signal.
```
cretz, line 51: FWIW, and these docs were written way before me, I think Godoc is clearer to read if it's just capped at 80 chars a line and the sentences come one after another. Adding a newline for each sentence does not affect its output (that whitespace is trimmed).

worker/worker.go
```diff
@@ -39,22 +39,23 @@ import (
...
+		// Pass worker.InterruptCh() to stop the Worker with SIGINT or SIGTERM.
+		// Pass nil to stop the Worker with an external Stop() call.
+		// Pass any other `<-chan interface{}` and Run will wait for a signal from that channel.
```
cretz, line 54: > will wait for a signal from that channel.

Pedantic: "will wait for a receive on that channel."

worker/worker.go
```diff
@@ -39,22 +39,23 @@ import (
...
 
-		// Run the worker in a blocking fashion. Stop the worker when interruptCh receives signal.
-		// Pass worker.InterruptCh() to stop the worker with SIGINT or SIGTERM.
-		// Pass nil to stop the worker with external Stop() call.
-		// Pass any other `<-chan interface{}` and Run will wait for signal from that channel.
-		// Returns error only if worker fails to start.
+		// Run the Worker in a blocking fashion.
+		// The Worker stops when interruptCh receives a signal.
+		// Pass worker.InterruptCh() to stop the Worker with SIGINT or SIGTERM.
+		// Pass nil to stop the Worker with an external Stop() call.
```
cretz, line 53: > Pass nil to stop the Worker with an external Stop() call.

Can make people think they couldn't use `Stop()` _unless_ they pass nil. Really you don't have to "Pass nil to stop the Worker with an external Stop() call", you can stop the worker w/ such a call always.

flossypurse:
> @flossypurse - Can you revisit this and update or close if out of date?

Yes, will get back to this soon.

README.md
flossypurse: Will remove it in this PR, and add it to the docs in this PR: https://github.com/temporalio/documentation/pull/866

flossypurse:
How to use a custom logger is being added to docs here: https://github.com/temporalio/documentation/pull/866

README.md
```diff
@@ -1,54 +1,69 @@
...
+
+## How-to guides
+
+- [How to develop a Workflow Definition](https://docs.temporal.io/docs/go/how-to-develop-a-workflow-definition-in-go)
+- [How to develop an Activity Definition](https://docs.temporal.io/docs/go/how-to-develop-an-activity-definition-in-go)
+- [How to develop a Worker Program](https://docs.temporal.io/docs/go/how-to-develop-a-worker-program-in-go)
+- [How to spawn a Workflow Execution](https://docs.temporal.io/docs/go/how-to-spawn-a-workflow-execution-in-go)
+- [How to spawn an Activity Execution](https://docs.temporal.io/docs/go/how-to-spawn-an-activity-execution-in-go)
+- [How to spawn a Child Workflow Execution](https://docs.temporal.io/docs/go/how-to-spawn-a-child-workflow-execution-in-go)
+- [How to execute a Side Effect](https://docs.temporal.io/docs/go/how-to-execute-a-side-effect-in-go)
```
flossypurse, line 41:
```suggestion
```
duplicate

cretz (approved):
Looks good to me! Want to get @bergundy's opinion before merging though to confirm what he might like/dislike with this pattern as it might pertain to other SDKs.

README.md
```diff
@@ -1,54 +1,68 @@
...
+**Are there executable code samples?**
+
+You can find a complete list of executable code samples in the [samples library](https://docs.temporal.io/docs/samples-library/#go), which includes Temporal Go SDK code samples from the [temporalio/samples-go](https://github.com/temporalio/samples-go) repo.
+Additionally, each of the Go SDK Tutorials in the [Temporal docs site](https://docs.temporal.io/docs/go) is backed by a fully executable template application.
+
+**Where is the Go SDK technical reference?**
+
+The [Temporal Go SDK API reference](https://pkg.go.dev/go.temporal.io/sdk) is published on [pkg.go.dev](https://pkg.go.dev/go.temporal.io/sdk)
+
+## How-to guides
```
bergundy, line 33: I don't love that we're putting this list here.
Let's keep it in the docs (for now at least).

worker/worker.go
```diff
@@ -39,22 +38,22 @@ import (
 )
 
 type (
-	// Worker hosts workflow and activity implementations.
-	// Use worker.New(...) to create an instance.
+	// Worker is the entity that hosts and executes Workflow Tasks and Activity Tasks.
```
bergundy, line 41: What does it mean to host tasks?
Why do we need to call it an entity? Feels like we're complicating things.

workflow/doc.go
```diff
@@ -384,31 +392,6 @@ workflow function should terminate by returning the special ContinueAsNewError e
 
 For a complete example implementing this pattern please refer to the Cron example.
 
-SideEffect API
```
bergundy, line 387: Looks like there's a lot of useful information here that's not found in our docs (e.g. that a side-effect can be executed more than once).

bergundy (commented):
I don't love that we're taking useful docs away from the code.
This is the best place for docs because it shows up in IDEs.

Let's talk about it and see if we want to go this direction.

README.md
```diff
@@ -1,54 +1,68 @@
...
-	// feed this logger into Temporal
-	logger := logur.LoggerToKV(logrusadapter.New(logrus.New()))
-}
+git clone git@github.com:temporalio/sdk-go.git
 ```
-
-Most of the popular logging solutions have existing adapters in logur. If you're curious about which adapters are available, here is a helpful [link](https://github.com/logur?q=adapter-).
+**Are there executable code samples?**
+
+You can find a complete list of executable code samples in the [samples library](https://docs.temporal.io/docs/samples-library/#go), which includes Temporal Go SDK code samples from the [temporalio/samples-go](https://github.com/temporalio/samples-go) repo.
```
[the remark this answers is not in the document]
flossypurse: > TBH I don't know why we have the list of samples in our docs, seems like it's good enough to put it in the samples repo.

Because internal engineers asked for it. There was also some positive support for this from external devs.

> Are we at least snipsyncing from the samples repo README?

Actually yes, we were. But Snipsync-ing content with clickable links means that docs builds break if the samples repo README links are wrong (makes it harder to update docs urls) - So I recently stopped doing that for the Go samples. We have redirects in docs to handle URL changes instead.

README.md
flossypurse: So you would prefer a single link?

bergundy: SGTM

cretz:
@flossypurse - Are you still wanting to merge this? If so, we can look into solving the conflicts and addressing any comments.

flossypurse:
Thanks for following up @cretz ! No, I think I will close this and circle back with something fresh at another time.

Appreciate it!

PR #650, opened 2021-11-23
Add a little detail on when heartbeat is seen from the activity side

### LINKED ISSUE #638: Document about and warn when attempting to use activity cancellation/heartbeats without HeartbeatTimeout
Currently we silently do not cancel the activity context even when regularly heartbeating and a cancel occurs from the outside unless `HeartbeatTimeout` is set. We should document that this value _must_ be set for standard cancellation (i.e. context-based cancellation) to work with activities and we should warn if heartbeating is attempted on an activity with no heartbeat timeout set.

cretz:
## What was changed

Make clear an why activity context doesn't see its cancellation immediately without documenting too many details.

## Why?

It is otherwise unclear when an activity can see its cancellation.

## Checklist

1. Closes #638


workflow/workflow.go
```diff
@@ -106,7 +106,9 @@ type (
 // *ApplicationError, *TimeoutError, *CanceledError, or *PanicError.
 //
 // You can cancel the pending activity using context(workflow.WithCancel(ctx)) and that will fail the activity with
-// *CanceledError set as cause for *ActivityError.
+// *CanceledError set as cause for *ActivityError. The context in the activity only becomes aware of the cancellation
+// when a heartbeat is sent to the server. Since heartbeats may be batched internally, this could take up to the
+// HeartbeatTimeout to appear or several minutes by default if that value is not set.
```
cretz, line 111: Technically it's only 80% of the timeout and technically the default is 10 minutes, but we leave that implementation detail out of the docs.

bergundy (approved):

Issue #570, opened 2021-10-02
Temporal Client send signal to different namespace


elb3k:
**Is your feature request related to a problem? Please describe.**
I have to send signal to different namespaces, but temporal client only allows to send signal to workflows in same namespace

**Describe the solution you'd like**
[SignalWorkflow](https://github.com/temporalio/sdk-go/blob/6580cbe0aa41a8b515791f95c2c15bb37db1dab1/internal/internal_workflow_client.go#L341-L365) just passes the namespace of client. May be we can encode the namespace to context (WtihWorkflowNamespace, just like workflow version), and use namespace in SignalWorkflow function (if exists).

**Describe alternatives you've considered**
Current solution to send signal to different namespace using custom workflow, where I send the signal to different namespace.

**Additional context**
It is possible to send the signal to different namespaces within workflow, using [WithWorkflowNamespace](https://github.com/temporalio/sdk-go/blob/6580cbe0aa41a8b515791f95c2c15bb37db1dab1/workflow/workflow_options.go#L39-L42). We could have same functionality for client.

elb3k:
One simple function to change namespace property of `Client` would be just fine too.

cretz:
With #641, you will be able to use the gRPC service directly to send signals in any way to any namespace. I will update when implemented.

cretz:
This has been implemented. You can now use the gRPC service directly to make whatever calls you want. Does this solve your need?

elb3k:
Yes @cretz.

PR #651, opened 2021-11-23
Expose gRPC

### LINKED ISSUE #641: Expose gRPC
We have traditionally hidden the gRPC client and settings from the user, but the benefit of hiding gRPC no longer outweighs the cost of decreased user flexibility.

Two changes:

1. Add `WorkflowService() workflowservice.WorkflowServiceClient` to `Client`
2. Add `GRPCDialOptions []grpc.DialOption` to `ConnectionOptions` and apply our dial options first so any explicitly provided dial options _could_ override our own if desired.

Questions:

1. Maybe add some form of `GetGRPCContext` so that they can use the same context we use for making a gRPC call?

cretz:
## What was changed

Added new `Client.WorkflowService` method that returns the underlying gRPC workflow service. Also expose gRPC dial options.

## Why?

People want to do advanced things that they cannot do with only our high-level wrapper (e.g. #570). This is exposed in other SDKs so should be here as well.

## Checklist

1. Closes #641


mocks/Client.go
```diff
@@ -58,6 +57,11 @@ func (_m *Client) CancelWorkflow(ctx context.Context, workflowID string, runID s
 	return r0
 }
 
+// Close provides a mock function with given fields:
```
Sushisource, line 60: This comment seems incomplete?

Sushisource: Oh, nevermind I guess these are autogenerated? Weird.

Sushisource (approved):

mocks/Client.go
cretz: Yes, `mockery` is rubbish and were I to have built this project I'd have never used it (callers should use the mocking library of their choice, we should not provide mocks IMO).

Sushisource: Yeah, I totally agree with that philosophy

PR #653, opened 2021-11-24
Dependency Abstraction and Multi-Module Repository

cretz:
## What was changed

* `opentelemetry/` was moved to `contrib/opentelemetry/` and made its own Go module
* `opentracing/` was moved to `contrib/opentracing/` and made its own Go module
* `client.MetricsScope` and all uses of metrics throughout the codebase were abstracted to use a `client.MetricsHandler` instead of having Tally hardcoded everywhere
* A `contrib/tally/` Go module was created with a Tally implementation of `client.MetricsHandler`
* `test/` was made its own Go module

## Why?

* By moving OpenTelemetry, OpenTracing, and Tally to their own Go modules, now users must opt-in to using those third party libraries. Today we force them on people and we have seen with both Tally and OpenTelemetry the cost of their incompatibilities.
* This mimics the abstraction we did with logging
* While this change is backwards incompatible (e.g. the `client.Options.MetricsScope` option is now `client.Options.MetricsHandler` with a different type), so is the interceptors change in the same way in master that is pending release
* Putting the integration tests in their own Go module means that future needs of the integration tests don't infect the rest of the SDK
  * This sets the stage for use of an embedded temporalite-type server
  * This sets the stage for import of an external SDK features-like project for centralized e2e/integration testing without circular dependency concerns
* People wanting to use, say, https://github.com/prometheus/client_golang can now do so without the Tally abstraction middleman

## Checklist

* Need to check whether everyone is ok with this `contrib/` folder living here or if we want some kind of `github.com/temporalio/sdk-go-contrib` repo. I personally like it here because:
  * I don't expect to accept every integration under the sun, just these few that we used to hardcode
  * This mimics what Java SDK does with OpenTracing and TypeScript SDK does with OpenTelemetry where they put the integrations directly in the repo
* Will need samples-go and other repo updates of course (part of normal release anyways)

## Vanity Import and Tagging

The vanity URLs do need to be updated with these three "subs" as https://github.com/leighmcculloch/vangen/ calls them

The tags and versions are different _for each nested module_. From https://go.dev/ref/mod#vcs-version:

> If a module is defined in a subdirectory within the repository, that is, the module subdirectory portion of the module path is not empty, then each tag name must be prefixed with the module subdirectory, followed by a slash. For example, the module golang.org/x/tools/gopls is defined in the gopls subdirectory of the repository with root path golang.org/x/tools. The version v0.4.0 of that module must have the tag named gopls/v0.4.0 in that repository.

See how https://github.com/hashicorp/vault names its tags for, say, it's `github.com/hashicorp/vault/api` nested module. Therefore when releasing this, we will make the normal tag `v1.12.0` for the primary module, but we will need 3 other tags: `contrib/opentelemetry/v0.1.0`, `contrib/opentracing/v0.1.0`, and `contrib/tally/v0.1.0`. Is this acceptable? Is the initial versions of `v0.1.0` for each acceptable?


cretz:
I know this PR _looks_ big, but it's really not that big, it just takes a lot of little changes to rip out Tally from where we hardcoded it everywhere.

Makefile
```diff
@@ -41,20 +39,20 @@ unit-test: $(BUILD)/dummy
...
-		go test "$$dir" $(TEST_ARG) -coverprofile=$(COVER_ROOT)/"$$dir"/cover.out || exit 1; \
+		(cd "$$dir" && go test . $(TEST_ARG) -coverprofile=$(COVER_ROOT)/"$$dir"/cover.out) || exit 1; \
 		cat $(COVER_ROOT)/"$$dir"/cover.out | grep -v "mode: atomic" >> $(UT_COVER_FILE); \
 	done;
 
 integration-test-zero-cache: $(BUILD)/dummy
 	@mkdir -p $(COVER_ROOT)
 	@for dir in $(INTEG_TEST_DIRS); do \
-		WORKFLOW_CACHE_SIZE=0 go test $(TEST_ARG) "$$dir" -coverprofile=$(INTEG_ZERO_CACHE_COVER_FILE) -coverpkg=./... || exit 1; \
+		(cd "$$dir" &&WORKFLOW_CACHE_SIZE=0 go test $(TEST_ARG) . -coverprofile=$(INTEG_ZERO_CACHE_COVER_FILE) -coverpkg=./...) || exit 1; \
```
robholland, line 49: Missing space after `&&`

robholland (approved):
Looks good to me.

Makefile
cretz: Thanks, will fix next commit.

mmcshane (approved):

Sushisource (approved):
Makes a lot of sense to me

cretz:
By request I am making a server draft PR to show what would change.

cretz:
A draft PR for server side has been created: https://github.com/temporalio/temporal/pull/2243

internal/common/metrics/handler.go
```diff
@@ -0,0 +1,138 @@
...
+import "time"
+
+// Handler is a handler for metrics emitted by the SDK. This interface is
+// intentionally limited to only what the SDK needs to emit metrics and is not
+// built to be a general purpose metrics abstraction for all uses.
+//
+// A common implementation is at
+// go.temporal.io/sdk/contrib/tally.NewMetricsHandler. The NopHandler is an
+// noop handler.
+type Handler interface {
```
Ardagan, line 34: I think we should give user an option to provide histogram metrics as well.

cretz: > unless user casts Handler to implementation and exposes functionality there.

Even this isn't safe because we wrap the handler in a "replay aware handler".

I have now exposed the scope from the Tally side if they want to use histograms. After internal discussion, since there isn't yet a good abstraction for histograms for OpenTelemetry and Tally, we are just exposing this value and asking them to be careful of replay. If a good histogram abstraction can be made, we can add it.

contrib/tally/handler.go
```diff
@@ -69,6 +69,36 @@ func NewMetricsHandler(scope tally.Scope) client.MetricsHandler {
...
+// Raw use of the scope is discouraged but may be used for Histograms or other
+// advanced features. This scope does not skip metrics during replay like the
+// metrics handler does. Therefore the caller should check replay state, for
+// example:
+// 	scope := tally.NoopScope
+// 	if !workflow.IsReplaying(ctx) {
+// 		scope = ScopeFromHandler(workflow.GetMetricsHandler(ctx))
+// 	}
+// 	scope.Histogram("my_histogram", nil).RecordDuration(5 * time.Second)
+func ScopeFromHandler(handler client.MetricsHandler) tally.Scope {
```
yiminc, line 86: if the handler is replay-aware, why would user need to check for IsReplaying flag?

cretz: This handler is not replay aware. The replay aware handler wraps this one (and any metrics handler they may create, e.g. for OpenTelemetry). The raw scope is accessible at any time in replay or not.

PR #655, opened 2021-11-29
Add GetLogger method to Tracer interface

jlegrone:
## What was changed

Adds a `GetLogger` method to the `Tracer` interface which allows tracer implementations to customize the logger used by workflow code.

There does not appear to be an `ActivityOutboundInterceptor` implementation for the `tracingInterceptor` struct, so customizing the activity logger was left out of scope for now.

This is likely to not be the last time a new method is needed on the `Tracer` interface, so a new `BaseTracer` struct has been added to allow this sort of change to be non-breaking in the future.

## Why?

Some tracing backends can correlate logs with spans to enhance observability. This is really useful in the Temporal context, where the worker logs associated with each workflow and activity can be directly linked to a trace.

More info about this capability in Datadog is available here: https://docs.datadoghq.com/tracing/connect_logs_and_traces/

## Checklist

1. Closes 

2. How was this tested:

Not tested. ~Will move this out of draft state when more validation has been performed.~

3. Any docs updates needed?


interceptor/tracing_interceptor.go
```diff
@@ -41,6 +42,9 @@ const (
 // Tracer is an interface for tracing implementations as used by
 // NewTracingInterceptor. Most callers do not use this directly, but rather use
 // the opentracing or opentelemetry packages.
+//
+// All implementations must embed BaseTracer to safely
+// handle future changes.
 type Tracer interface {
```
jlegrone, line 48: By the way, we really appreciate this being a public interface. I think this is going to significantly simplify our Datadog tracing implementation (and make us more confident in open sourcing a supported Temporal/Datadog integration if we can ensure that the module in which this interface is defined follows semver 😉)

cretz: :+1: Will keep it exported then.

cretz:
This looks good to me.

cretz (approved):
Note, #653 moves the opentracing/opentelemetry files around a bit, so if that lands before this we may have to make slight changes.

cretz:
Gonna go ahead and merge this when done building

PR #657, opened 2021-11-30
Add debug log for all errors received from RecordActivityHeartbeat

meiliang86:
## What was changed

Add debug log for all record activity heartbeat errors.

## Why?

For investigating heartbeat failures.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_activity.go
+++ b/internal/internal_activity.go
@@ -394,7 +394,7 @@ func (a *activityEnvironmentInterceptor) RecordHeartbeat(ctx context.Context, de
 	}
 	var data *commonpb.Payloads
 	var err error
-	// We would like to be a able to pass in "nil" as part of details(that is no progress to report to)
+	// We would like to be able to pass in "nil" as part of details(that is no progress to report to)
 	if len(details) > 1 || (len(details) == 1 && details[0] != nil) {
 		data, err = encodeArgs(getDataConverterFromActivityCtx(ctx), details)
 		if err != nil {
@@ -402,11 +402,8 @@ func (a *activityEnvironmentInterceptor) RecordHeartbeat(ctx context.Context, de
 		}
 	}
 
-	err = a.env.serviceInvoker.Heartbeat(ctx, data, false)
-	if err != nil {
-		log := GetActivityLogger(ctx)
-		log.Debug("RecordActivityHeartbeat with error", tagError, err)
-	}
+	// Heartbeat error is logged inside ServiceInvoker.internalHeartBeat
+	a.env.serviceInvoker.Heartbeat(ctx, data, false)
 }
 
 func (a *activityEnvironmentInterceptor) HasHeartbeatDetails(ctx context.Context) bool {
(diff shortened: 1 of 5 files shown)


internal/internal_task_handlers.go
```diff
@@ -1694,6 +1695,10 @@ func (i *temporalInvoker) internalHeartBeat(ctx context.Context, details *common
 		}
 	}
 
+	if err != nil {
```
cretz, line 1698: Hrmm, I wonder if it's too noisy to log canceled errors which seem like a normal thing to have happen. I guess canceled errors were logged before anyways?

internal/internal_activity.go
```diff
@@ -394,19 +394,16 @@ func (a *activityEnvironmentInterceptor) RecordHeartbeat(ctx context.Context, de
...
 		data, err = encodeArgs(getDataConverterFromActivityCtx(ctx), details)
 		if err != nil {
 			panic(err)
 		}
 	}
 
-	err = a.env.serviceInvoker.Heartbeat(ctx, data, false)
-	if err != nil {
-		log := GetActivityLogger(ctx)
-		log.Debug("RecordActivityHeartbeat with error", tagError, err)
```
cretz, line 408: IIRC, using `GetActivityLogger` from the context has all sorts of tags about the specific activity. I am concerned the new logging statement is not activity specific. Can you confirm what this log looks like before and after this PR?

internal/internal_task_handlers.go
meiliang86: This is debug log, and I don't expect customers to enable them by default.
For us, we should record all errors. I don't worry too much about the volume here because long running activity and heartbeat is not a large scale operation.

internal/internal_activity.go
meiliang86: I can use GetActivityLogger method. I was looking at the activityTaskHandler and it seems that it uses its own logger and add tags.

internal/internal_task_handlers.go
cretz: FWIW, I think debug log is on by default. See https://github.com/temporalio/sdk-go/blob/master/internal/log/default_logger.go where there are no conditionals on the level. I do agree we shouldn't probably have it that way, but it currently is.

internal/internal_activity.go
cretz: :+1: I haven't walked the code in detail, but whatever gets all those activity-specific tags on the logger I think is most reasonable.

internal/internal_task_handlers.go
meiliang86: IMO it's ok to log it at the debug level. People should adjust the log level it they feel this becomes too vervose.
RecordHeartbeat is already async/best effort. So I want to make sure we record all errors so we can investigate potential issues.

cretz: :+1:

internal/internal_task_handlers.go
```diff
@@ -1694,6 +1694,11 @@ func (i *temporalInvoker) internalHeartBeat(ctx context.Context, details *common
...
+		logger := GetActivityLogger(ctx)
```
cretz, line 1698: Do we need any protection for this codepath potentially not being an activity context?

cretz (approved):

internal/internal_task_handlers.go
meiliang86: Not from actual user code. Heartbeat can only be called from activity context.

PR #660, opened 2021-12-03
Support configurable/default heartbeat default/max intervals.

### LINKED ISSUE #656: Heartbeat throttling interval should be configurable and default interval should be less than 10 minutes
Go SDK to implement this behavior: https://temporalio.notion.site/Activity-heartbeats-inconsistent-behavior-between-SDKs-32d81da0748044a8932481d2c9864791

cretz:
## What was changed

Currently in Go the frequency pending heartbeats are relayed to the server are 80% of the given heartbeat timeout or if no heartbeat timeout, 8 minutes (80% of 10 minutes). This changes that to be 80% of given heartbeat timeout or 30 seconds and caps it at a configurable max, or if the max is not configured, 60 seconds.

This does mean users who previously had heartbeats way less frequent, e.g. every few minutes, will now have them more frequent (default no less frequent than 60s). We do not believe going down from the large default is particularly harmful to running users though it technically could cause more traffic.

## Why?

It was decided all SDKs should behave the same with regards to heartbeat frequency. See #656 and its associated links.

## Checklist

1. Closes #656
2. How was this tested: In the same way the 8 minute heartbeat was not tested in an integration setting, neither is this. Without a time-skipping server, timeout-based tests are not very feasible. But of course logic to derive the timeout based on configuration is tested.


internal/internal_task_handlers.go
```diff
@@ -1866,6 +1861,27 @@ func (ath *activityTaskHandlerImpl) getRegisteredActivityNames() (activityNames
...
+		heartbeatThrottleInterval = ath.defaultHeartbeatThrottleInterval
+	} else {
+		heartbeatThrottleInterval = defaultDefaultHeartbeatThrottleInterval
+	}
+	// Limit interval to a max
+	if ath.maxHeartbeatThrottleInterval > 0 && heartbeatThrottleInterval > ath.maxHeartbeatThrottleInterval {
+		heartbeatThrottleInterval = ath.maxHeartbeatThrottleInterval
+	} else if ath.maxHeartbeatThrottleInterval == 0 && heartbeatThrottleInterval > defaultMaxHeartbeatThrottleInterval {
+		heartbeatThrottleInterval = defaultMaxHeartbeatThrottleInterval
+	}
```
bergundy, line 1880: I find this a bit complicated to read.

It would have been easier to only have to consider `ActivityOptions.heartbeatTimeout`, `WorkerOptions.{default,max}HeartbeatThrottleInterval` and have defaults for those options provided outside of this method.

internal/worker.go
```diff
@@ -159,6 +159,17 @@ type (
...
+		// Optional: The maximum amount of time between sending each pending heartbeat to the server. Regardless of
+		// heartbeat timeout, no pending heartbeat will wait longer than this amount of time to send.
+		// default: 60 seconds
+		MaxHeartbeatThrottleInterval time.Duration
+
+		// Optional: The default amount of time between sending each pending heartbeat to the server. This is used if the
+		// ActivityOptions do not provide a HeartbeatTimeout. Otherwise, the interval becomes a value a bit smaller than the
+		// given HeartbeatTimeout.
+		// default: 30 seconds
+		DefaultHeartbeatThrottleInterval time.Duration
```
bergundy, line 171: Maybe mention that if heartbeatTimeout is provided than the throttle interval is 80% of that timeout?

bergundy (approved):
Overall LGTM.

internal/internal_task_handlers.go
cretz: They are provided outside this method too. But there are a couple of ways this is called in test code, so I wanted to do it here just to be safe.

internal/worker.go
cretz: > Otherwise, the interval becomes a value a bit smaller than the given HeartbeatTimeout.

To clarify, are you wanting me to change "a bit smaller than" to "80% of"? I intentionally left off the exact amount as an implementation detail, but can add it if I must.

bergundy: Scratch that, we don't have to document the exact value.

internal/internal_task_handlers.go
bergundy: So I'd split it into 2 methods, one that sets the default options and one that does the calculation to reduce the complexity.

cretz: Those methods would be a few lines and each called only once.

bergundy: Can we do something like this?

In `func newActivityTaskHandlerWithCustomProvider`
do:
```go
defaultHeartbeatThrottleInterval:
  providedOrDefaultInterval(params.DefaultHeartbeatThrottleInterval, defaultDefaultHeartbeatThrottleInterval),
maxHeartbeatThrottleInterval:
  providedOrDefaultInterval(params.MaxHeartbeatThrottleInterval, defaultMaxHeartbeatThrottleInterval),
```

And make `getHeartbeatThrottleInterval` something like:
```go
if heartbeatTimeout > 0 {
		heartbeatThrottleInterval = time.Duration(0.8 * float64(heartbeatTimeout))
} else {
		heartbeatThrottleInterval = ath.defaultHeartbeatThrottleInterval
}

if heartbeatThrottleInterval > ath.maxHeartbeatThrottleInterval {
		heartbeatThrottleInterval = ath.maxHeartbeatThrottleInterval
}
return heartbeatThrottleInterval
```

If we're missing the initialization in tests just make sure ActivityTaskHandler is properly initialized.

PR #668, opened 2021-12-11
Add function Context in logger for customization

tranngoclam:
## What was changed

- Add a new function Context returning Logger in the Logger interface.

## Why?

- This helps the logger customization can use the key value pairs which are attached in the `context.Context` easily.

## Checklist

1. Closes 

2. How was this tested:

- Unit test

3. Any docs updates needed?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ default: check test
 BUILD := .build
 
 TEST_TIMEOUT := 3m
-TEST_ARG ?= -race -v -timeout $(TEST_TIMEOUT)
+TEST_ARG ?= -timeout $(TEST_TIMEOUT)
 
 INTEG_TEST_ROOT := ./test
 COVER_ROOT := $(abspath $(BUILD)/coverage)
--- a/internal/log/default_logger.go
+++ b/internal/log/default_logger.go
@@ -25,6 +25,7 @@
 package log
 
 import (
+	"context"
 	"fmt"
 	golog "log"
 	"os"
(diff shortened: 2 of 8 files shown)


cretz:
Unless I'm misunderstanding the intent, I think this is more clearly done with interceptors. With the recently introduced interceptors, you can intercept calls to `GetLogger` from workflows and activities via `interceptor.WorkflowOutputInterceptor.GetLogger` and  `interceptor.ActivityOutboundInterceptor.GetLogger` respectively and both are passed a context.

You have to wire this up via `WorkerInterceptor` that returns either/both `WorkflowInboundInterceptor` and `ActivityInboundInterceptor`, then in their `Init` with the given outbound interceptor, call the next in-chain's `Init` with your own wrapper. I can show an example of this if unclear.

tranngoclam:
Thanks @cretz
Could you please share some samples to me ?

cretz:
@tranngoclam - Sure, I will add this as part of https://github.com/temporalio/samples-go/issues/150 soon.

hungdinhts:
> @tranngoclam - Sure, I will add this as part of [temporalio/samples-go#150](https://github.com/temporalio/samples-go/issues/150) soon.

Could you give us example how to extract some data from context and include to log message as [this example](https://github.com/logur/logur/blob/master/logger_context.go)? Our use case is to inject a distributed request/action id to every log message. @cretz

cretz:
The example has been added.

PR #669, opened 2021-12-13
Expose ProtoJSONPayloadConverter marshalers and unmarshalers options

alexandrevilain:
## What was changed

Protojson and gogo jsonpb marshaler and umarshaler options are now public fields of the ProtoJSONPayloadConverter struct.

## Why?

It allows end-users more control on the payload converter behavior. 

I had a workflow failing caused by proto message update adding a new field. The workflow starter had the most recent version of the proto message, but not the workflow implementation. Protojson returned an `unknown field X` error which can be skipped using the `DiscardUnknown` boolean.

## Checklist

1. How was this tested:

Added unit test to ensure it does not break the default behavior of the `ProtoJSONPayloadConverter`.

### DIFF
--- a/converter/payload_converter_test.go
+++ b/converter/payload_converter_test.go
@@ -73,6 +73,35 @@ func TestProtoJsonPayloadConverter_Gogo(t *testing.T) {
 	assert.Equal(t, `{"eventId":"1978","eventType":"WorkflowTaskTimedOut","workflowTaskTimedOutEventAttributes":{"scheduledEventId":"2","timeoutType":"ScheduleToStart"}}`, s)
 }
 
+func TestProtoJsonPayloadConverter_Gogo_Custom(t *testing.T) {
+	pc := NewProtoJSONPayloadConverter()
+	pc.GogoMarshaler.EmitDefaults = true
+	pc.GogoUnmarshaler.AllowUnknownFields = true
+
+	wt := &historypb.HistoryEvent{
+		EventId:   1978,
+		EventType: enumspb.EVENT_TYPE_WORKFLOW_TASK_TIMED_OUT,
+		Attributes: &historypb.HistoryEvent_WorkflowTaskTimedOutEventAttributes{WorkflowTaskTimedOutEventAttributes: &historypb.WorkflowTaskTimedOutEventAttributes{
+			ScheduledEventId: 2,
+			TimeoutType:      enumspb.TIMEOUT_TYPE_SCHEDULE_TO_START,
+		}}}
+
+	payload, err := pc.ToPayload(wt)
+	require.NoError(t, err)
+
+	s := pc.ToString(payload)
+	assert.Equal(t, `{"eventId":"1978","eventTime":null,"eventType":"WorkflowTaskTimedOut","version":"0","taskId":"0","workflowTaskTimedOutEventAttributes":{"scheduledEventId":"2","startedEventId":"0","timeoutType":"ScheduleToStart"}}`, s)
+
+	payload2 := &commonpb.Payload{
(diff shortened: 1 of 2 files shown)


cretz:
I am afraid this is a bit dangerous to customize as these are default implementations that are expected to match all SDKs across all languages. I think if you need to customize the proto converters, you should create a wholly separate `PayloadConverter` with a separate encoding. This should be very easy; I can show an example if you want.

alexandrevilain:
Hi @cretz :) 

Thanks for the review, I understand your concerns. 

IMHO, this PR isn't changing the default behavior of the `ProtoJSONPayloadConverter`, it's only providing a way for advanced users to customize the behavior without having to duplicated the whole converter logic. 

As it's instantiating empty `protojson.MarshalOptions` and `protojson.UnmarshalOptions`, it keeps the same behavior as the default `Marshal` and `Unmarshal` functions (see: https://github.com/protocolbuffers/protobuf-go/blob/v1.27.1/encoding/protojson/encode.go#L38).

cretz:
Ability to customize the behavior is the concern. Would example code of how to make your own payload converter w/ proto options suffice instead of altering the default?

alexandrevilain:
Thanks for your proposal. I already did my own to quick-fix the issue I had with unknown fields. 
But it's basically a copy-past of the `ProtoJSONPayloadConverter` with customizations on protojson.

I thought that it would be useful for others :)

cretz:
> But it's basically a copy-past of the ProtoJSONPayloadConverter with customizations on protojson.

Mostly, but a tad simpler if you can assume which proto library. Here's a quick thing I threw together untested:

```go
package myconverter

import (
	"encoding/base64"
	"fmt"

	"go.temporal.io/api/common/v1"
	"google.golang.org/protobuf/encoding/protojson"
	"google.golang.org/protobuf/proto"
)

type MyProtoConverter struct {
	// MetadataEncoding is required
	MetadataEncoding string
	MarshalOptions   protojson.MarshalOptions
	UnmarshalOptions protojson.UnmarshalOptions
}

var _ converter.PayloadConverter = &MyProtoConverter{}

func (m *MyProtoConverter) ToPayload(value interface{}) (*common.Payload, error) {
	// No encoding, no conversion
	if m.MetadataEncoding == "" {
		return nil, nil
	}

	// Assert type, then marshal
	p, _ := value.(proto.Message)
	if p == nil {
		return nil, nil
	}
	b, err := m.MarshalOptions.Marshal(p)
	if err != nil {
		return nil, fmt.Errorf("%w: %v", converter.ErrUnableToEncode, err)
	}
	return &common.Payload{
		Metadata: map[string][]byte{converter.MetadataEncoding: []byte(m.MetadataEncoding)},
		Data:     b,
	}, nil
}

func (m *MyProtoConverter) FromPayload(payload *common.Payload, valuePtr interface{}) error {
	// No encoding or wrong encoding, no conversion
	if m.MetadataEncoding == "" {
		return converter.ErrEncodingIsNotSet
	} else if m.MetadataEncoding != string(payload.Metadata[converter.MetadataEncoding]) {
		return converter.ErrEncodingIsNotSupported
	}

	// Assert type, then unmarshal
	p, _ := valuePtr.(proto.Message)
	if p == nil {
		return converter.ErrTypeNotImplementProtoMessage
	}
	return m.UnmarshalOptions.Unmarshal(payload.Data, p)
}

func (*MyProtoConverter) ToString(payload *common.Payload) string {
	return base64.RawStdEncoding.EncodeToString(payload.GetData())
}

func (m *MyProtoConverter) Encoding() string {
	return m.MetadataEncoding
}
```

PR #673, opened 2021-12-16
Support pointer-to-pointer in channel receive

### LINKED ISSUE #665: Allow channel to convert to same type set if pointer

Right now, a channel `Receive` must use a pointer to the type used with `Send`. With futures (ref #593) and in other cases, we allow you to use a pointer to convert from a pointer instead of requiring a pointer to a pointer.

Support regular pointers when converting pointers in channels instead of always requiring pointers to pointers.

cretz:
## What was changed / Why?

See #665. Basically calling `Receive` with the same pointer type as `Send` was used with did not work,

## Checklist

1. Closes #665

### DIFF
--- a/internal/encode_args.go
+++ b/internal/encode_args.go
@@ -121,11 +121,18 @@ func decodeAndAssignValue(dc converter.DataConverter, from interface{}, toValueP
 	} else if fv := reflect.ValueOf(from); fv.IsValid() {
 		fromType := fv.Type()
 		toType := reflect.TypeOf(toValuePtr).Elem()
-		assignable := fromType.AssignableTo(toType)
-		if !assignable {
-			return fmt.Errorf("%s is not assignable to  %s", fromType.Name(), toType.Name())
+		// If the value set was a pointer and is the same type as the wanted result,
+		// instead of panicking because it is not a pointer to a pointer, we will
+		// just set the pointer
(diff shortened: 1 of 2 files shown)


internal/encode_args.go
```diff
@@ -121,11 +121,18 @@ func decodeAndAssignValue(dc converter.DataConverter, from interface{}, toValueP
...
-			return fmt.Errorf("%s is not assignable to  %s", fromType.Name(), toType.Name())
+		// If the value set was a pointer and is the same type as the wanted result,
+		// instead of panicking because it is not a pointer to a pointer, we will
+		// just set the pointer
+		if fv.Kind() == reflect.Ptr && fromType.Elem() == toType {
+			reflect.ValueOf(toValuePtr).Elem().Set(fv.Elem())
+		} else {
+			assignable := fromType.AssignableTo(toType)
+			if !assignable {
+				return fmt.Errorf("%s is not assignable to  %s", fromType.Name(), toType.Name())
```
Sushisource, line 132:
```suggestion
				return fmt.Errorf("%s is not assignable to %s", fromType.Name(), toType.Name())
```

Sushisource (approved):

internal/encode_args.go
cretz: I took the message verbatim what it was, but I don't suspect anyone has any conditionals on it, :+1:

PR #681, opened 2022-01-03
Debug mode does not set the DeadlockDetectionTimeout

marvinkite:
## What was changed

The debug mode now sets correctly the DeadlockDetectionTimeout after fixing coding error and it uses the same variable to detect the debug mode.
## Why?

It's a bug and second the package private debugMode variable is set when the program is loaded while the other condition which reads after the new test environment can be inconsistent  if the test calls the `os.SetEnv` just before.
## Checklist

1. Closes 
N/A
2. How was this tested:
Manually.
3. Any docs updates needed?
N/A

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1468,8 +1468,9 @@ func setWorkerOptionsDefaults(options *WorkerOptions) {
 	if options.DeadlockDetectionTimeout == 0 {
 		if debugMode {
 			options.DeadlockDetectionTimeout = unlimitedDeadlockDetectionTimeout
+		} else {
+			options.DeadlockDetectionTimeout = defaultDeadlockDetectionTimeout
 		}
-		options.DeadlockDetectionTimeout = defaultDeadlockDetectionTimeout
 	}
 	if options.DefaultHeartbeatThrottleInterval == 0 {
 		options.DefaultHeartbeatThrottleInterval = defaultDefaultHeartbeatThrottleInterval
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -28,7 +28,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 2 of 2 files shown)


cretz:
Hrmm, that does indeed look like a bug. I was wondering why `TEMPORAL_DEBUG` was working for me though. After following the code, it appears that the `DeadlockDetectionTimeout` option just ends up being used inside the coroutine call where it forcefully overrides that value anyways when the env var is set,

Can you help me understand how the buggy `DeadlockDetectionTimeout` override w/ `TEMPORAL_DEBUG` issue that you fixed manifested for you? I'm showing `debugMode` would have overridden it later anyways. Or is this just a case of fixing the bug while there and the testsuite env-var-re-eval was the real problem?

cretz (approved):

marvinkite:
In previous SDK version I didn't use the `TEMPORAL_DEBUG` I just simply set the `workflowEnv.SetTestTimeout(5000 * time.Second)` but after I migrated to the new `v1.12.0` my unit test failed and when I tried to debug the same way It panic so I tried to investigate why the old  way not working. Next I called just `os.SetEnv("TEMPORAL_DEBUG", "1")` before I called the `NewTestWorkflowEnvironment()` function but it still resulted in panic. Then I noticed the  `DeadlockDetectionTimeout` so I set the environment `TEMPORAL_DEBUG=1` to make sure DeadlockDetectionTimeout is set because my unit test was reaching that buggy line which resets to 1s.  Overall the debug didn't work as It did before. Now Looking the code without debugging, I can only spot one possible thing. It's the explicit worker option.

```golang
// os.SetEnv("TEMPORAL_DEBUG", "1")
testSuite := &testsuite.WorkflowTestSuite{}
workflowEnv = testSuite.NewTestWorkflowEnvironment()
workflowEnv.SetWorkerOptions(worker.Options{
    // Here the worker option is not nil so maybe the temporal didn't use the default 
			BackgroundActivityContext: contextWithTestClient(context.Background()),
    // Now I realised I could have set it here 
    //			DeadlockDetectionTimeout: "long enough",
})
workflowEnv.ExecuteWorkflow(workflow.ProcessOrderWorkflowV1, workflowParam)
```

cretz:
Thanks! Merged.

PR #683, opened 2022-01-04
Set Fossa to run non-blocking in buildkite

mcbryde:
## What was changed
Updated buildkite config to run Fossa non-blocking (only on merge to master)

## Why?
Fossa has too many false positives to run in blocking mode


cretz (approved):

cretz:
@mcbryde - Will merge after CLA complete (or I can go make this change separately if you prefer).

mcbryde:
@cretz Sounds good to me, thanks!

PR #689, opened 2022-01-06
Add ability to disable workflow worker

### LINKED ISSUE #636: Add WorkerOption to disable workflow worker
Currently, there is no way to start worker for activity only. We need a flag in worker.Options to explicitly disable workflow worker.

cretz:
## What was changed

Added option to disable workflow worker.

## Why?

For activity-only workers (such as those used by https://github.com/temporalio/samples-go/pull/166), we need to be able to disable unnecessary workflow worker polling and such.

## Checklist

1. Closes #636

### DIFF
--- a/internal/internal_worker_test.go
+++ b/internal/internal_worker_test.go
@@ -2632,3 +2632,14 @@ func TestIsNonRetriableError(t *testing.T) {
 		require.Equal(t, test.expected, isNonRetriableError(test.err))
 	}
 }
+
+func TestWorkerRegisterDisabledWorkflow(t *testing.T) {
+	// Expect panic
+	var recovered interface{}
+	func() {
+		defer func() { recovered = recover() }()
+		worker := NewAggregatedWorker(&WorkflowClient{}, "some-task-queue", WorkerOptions{DisableWorkflowWorker: true})
+		worker.RegisterWorkflow(testReplayWorkflow)
+	}()
+	require.Equal(t, "workflow worker disabled, cannot register workflow", recovered)
+}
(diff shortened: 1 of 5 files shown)


internal/internal_worker_test.go
```diff
@@ -2632,3 +2632,14 @@ func TestIsNonRetriableError(t *testing.T) {
...
+
+func TestWorkerRegisterDisabledWorkflow(t *testing.T) {
+	// Expect panic
+	var recovered interface{}
+	func() {
+		defer func() { recovered = recover() }()
+		worker := NewAggregatedWorker(&WorkflowClient{}, "some-task-queue", WorkerOptions{DisableWorkflowWorker: true})
+		worker.RegisterWorkflow(testReplayWorkflow)
+	}()
+	require.Equal(t, "workflow worker disabled, cannot register workflow", recovered)
```
mmcshane, line 2644: Don't know if you've seen it  but `require.PanicsWithValue` is slightly less work than this. Not a big deal.

mmcshane (approved):

internal/worker.go
```diff
@@ -147,6 +147,11 @@ type (
 		// default: 1000
 		MaxConcurrentSessionExecutionSize int
 
+		// Optional: If set to true, a workflow worker is not started for this
+		// worker and workflows cannot be registered with this worker.
```
Sushisource, line 151:
```suggestion
		// Optional: If set to true, a workflow worker is not started for this
		// worker and workflows cannot be registered with this worker.
		// Use this if you only want your worker to execute activities.
```

Sushisource (approved):

internal/worker.go
cretz: Updated

PR #690, opened 2022-01-06
Only respond with activity cancellation when requested by server

### LINKED ISSUE #686: Manually returned activity cancel error without receiving a server-side error

Today for activities, when you return `temporal.CanceledError` or `context.Canceled`, we send back `RespondActivityTaskCanceledRequest` regardless of whether there was a cancellation in flight. See https://github.com/temporalio/sdk-go/blob/4d6762570b52a77fe6df5a644ca024a1191bc407/internal/internal_task_pollers.go#L974-L987.

* Make sure to only handle those cancelled errors as actually sending a cancel to the server if the server sent us a cancel
* Make sure to wrap `temporal.CanceledError` and `context.Canceled` into a normal retryable `ApplicationFailure` in other cases
* Make sure other SDKs do the same
* Write a test confirming what happens in these situations
* Write a test confirming what happens if an activity is not heartbeating to receive a cancel and returns a failure
* To check potential workflow behavior incompatibility, write a test to confirm what workflow does when we return a canceled error (or go canceled error) when a cancellation was requested and, most importantly, when it was not.

cretz:
## What was changed

Do not send `RespondActivityTaskCanceledRequest` on return cancel error unless the server initiated the cancel. Also added tests confirming behavior for activities and workflows when manually returning cancellation errors.

## Why?

See #686. Basically the server has to be expecting the cancellation to handle it.

## Checklist

1. Closes #686

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -955,8 +955,15 @@ func reportActivityCompleteByID(
 	return reportErr
 }
 
-func convertActivityResultToRespondRequest(identity string, taskToken []byte, result *commonpb.Payloads, err error,
-	dataConverter converter.DataConverter, namespace string) interface{} {
+func convertActivityResultToRespondRequest(
+	identity string,
+	taskToken []byte,
+	result *commonpb.Payloads,
+	err error,
+	dataConverter converter.DataConverter,
+	namespace string,
+	cancelAllowed bool,
+) interface{} {
 	if err == ErrActivityResultPending {
 		// activity result is pending and will be completed asynchronously.
 		// nothing to report at this point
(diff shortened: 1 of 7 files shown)


internal/internal_task_pollers.go
```diff
@@ -1011,24 +1036,32 @@ func convertActivityResultToRespondRequestByID(identity, namespace, workflowID,
...
-	if errors.As(err, &canceledErr) {
-		return &workflowservice.RespondActivityTaskCanceledByIdRequest{
-			Namespace:  namespace,
-			WorkflowId: workflowID,
-			RunId:      runID,
-			ActivityId: activityID,
-			Details:    convertErrDetailsToPayloads(canceledErr.details, dataConverter),
-			Identity:   identity}
+	// Only respond with canceled if allowed
+	if cancelAllowed {
```
bergundy, line 1040: Is there a way to de-dupe this code?

bergundy (approved):

internal/internal_task_pollers.go
cretz: Not really. There isn't as much duplicate code as it might appear (the created structs are different).

internal/internal_task_pollers.go
```diff
@@ -1011,24 +1036,32 @@ func convertActivityResultToRespondRequestByID(identity, namespace, workflowID,
...
-		return &workflowservice.RespondActivityTaskCanceledByIdRequest{
-			Namespace:  namespace,
-			WorkflowId: workflowID,
-			RunId:      runID,
-			ActivityId: activityID,
-			Identity:   identity}
+	// If a canceled error is returned but it wasn't allowed, we have to wrap in
+	// an unexpected-cancel application error
+	if _, isCanceledErr := err.(*CanceledError); isCanceledErr {
+		err = fmt.Errorf("unexpected activity cancel error: %w", err)
```
yiminc, line 1064: When people return CanceledError, they won't expect to see this error. It would be great to give more clear message that is self explanatory so they don't need to seek advise and know exactly what to fix.

test/integration_test.go
```diff
@@ -1819,6 +1819,82 @@ func (ts *IntegrationTestSuite) TestTallyScopeAccess() {
...
+	ts.Equal(enumspb.WORKFLOW_EXECUTION_STATUS_FAILED, resp.GetWorkflowExecutionInfo().GetStatus())
+
+	// Activity using Go canceled error after cancel should return normal cancel
+	// error
+	fromActivity, waitForCancel, goCancelError = true, true, true
+	err = ts.executeWorkflow(wfIDPrefix+"4", ts.workflows.ReturnCancelError, nil,
+		fromActivity, rawActivityError, waitForCancel, goCancelError)
+	ts.Error(err)
+	ts.NotContains(err.Error(), "context canceled")
+	ts.Contains(err.Error(), "canceled")
```
yiminc, line 1869: add assert that it does not contain "unexpected"

yiminc (approved):

test/integration_test.go
cretz: Unexpected only relates to temporal canceled errors not go canceled errors. Notice I don't have an assert that temporal canceled errors don't contain "context canceled". They are separate things.

No harm in adding it I guess.

internal/internal_task_pollers.go
cretz: Suggestion?

test/integration_test.go
cretz: Added

PR #693, opened 2022-01-11
Add metadata.messageType to protobuf payloads

lorensr:
Matches this PR in the Java SDK: https://github.com/temporalio/sdk-java/pull/942

I'm new to Go—feedback welcome! 🤗


converter/proto_helpers.go
```diff
@@ -0,0 +1,53 @@
...
+
+func GetProtoMessageName(protoMessage *protoreflect.ProtoMessage) string {
+	return string((*protoMessage).ProtoReflect().Descriptor().FullName())
+}
+
+func checkMessageType(payload *common.Payload, valueType string) {
+	messageTypeBytes, ok := payload.Metadata[MetadataMessageType]
+	messageType := string(messageTypeBytes)
+	if (ok && messageType != valueType) {
+		fmt.Println(`WARN Encoded protobuf message type "` + messageType	+ `" does not match value type "` + valueType + `"`);
```
lorensr, line 43: I'm guessing this should be using the `log` package? Perhaps a `Logger` instance that the user configures when setting up the Client or Worker? I'm wondering how to get that instance.

CONTRIBUTING.md
```diff
@@ -2,22 +2,29 @@
...
-  - Install on OS X with `brew install go`.
-  - Install on Ubuntu with `sudo apt install golang`.
+  - Ubuntu: `sudo apt install golang`.
+  - OS X: `brew install go` and add this to your `.bashrc`:
+
+```
+export GOPATH=$HOME/go
+export GOROOT="$(brew --prefix go)/libexec"
+export PATH="$PATH:${GOPATH}/bin:${GOROOT}/bin"
+```
```
cretz, line 17: Code code block is not properly nested under the bullet (see https://gist.github.com/clintel/1155906)

converter/payload_converter.go
```diff
@@ -42,6 +42,11 @@ type PayloadConverter interface {
 	Encoding() string
 }
 
+type ProtoPayloadConverterInterface interface {
```
cretz, line 45: I don't think this needs to be exported

converter/proto_helpers.go
cretz: There's no available log here. I am ok not checking this. Unfortunately unlike Java we have to pass around log instances.

converter/proto_helpers.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+
+	gogoproto "github.com/gogo/protobuf/proto"
+	"go.temporal.io/api/common/v1"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+func GetProtoMessageName(protoMessage *protoreflect.ProtoMessage) string {
```
cretz, line 35: I don't think this stuff needs to be public API

converter/proto_json_payload_converter.go
```diff
@@ -41,6 +41,11 @@ import (
 type ProtoJSONPayloadConverter struct {
 	gogoMarshaler   gogojsonpb.Marshaler
 	gogoUnmarshaler gogojsonpb.Unmarshaler
+	options         ProtoJSONPayloadConverterOptions
+}
+
+type ProtoJSONPayloadConverterOptions struct {
+	excludeProtobufMessageTypes bool
```
cretz, line 48: Nobody can set this value because it is not exported

converter/proto_json_payload_converter.go
```diff
@@ -41,6 +41,11 @@ import (
 type ProtoJSONPayloadConverter struct {
 	gogoMarshaler   gogojsonpb.Marshaler
 	gogoUnmarshaler gogojsonpb.Unmarshaler
+	options         ProtoJSONPayloadConverterOptions
+}
+
+type ProtoJSONPayloadConverterOptions struct {
```
cretz, line 47: All exported items should have documentation that starts with its name, e.g. `// ProtoJSONPayloadConverterOptions represents options for `NewProtoJSONPayloadConverterWithOptions`.

converter/payload_converter.go
lorensr: Ah, thanks. I was using capitals for using it outside of the file, but it's for outside of the package.

cretz (approved):
Looks good! Seems CI failed in a spurious way, I have restarted it. Once complete successfully, feel free to merge.

lorensr:
Thanks! I can't merge (don't have write access) or view buildkite (first says "login", then says "page not found", I assume due to permissions). `make check` and my `make test` with `MallocNanoZone=0` both exit with code 0.

cretz:
:+1: I will look into the build errors then do the merge

PR #706, opened 2022-01-24
Replace health checking with server-capability getter and disable retries on internal errors

### LINKED ISSUE #701: Implement server capabilities and replace health check

Server capabilities are needed for some things like knowing whether to retry certain classes of error

* Remove health checking and options (including removing the ability to disable health check)
  * This was a subject of much debate, but lazy capabilities lookup is difficult and having a way to opt-out of getting server capabilities is troublesome
  * This causes those that were traditionally using the disable-health-check option to have lazy client connectivity to rework their approach
* Upon client creation, instead of health check, call `getServerInfo` to obtain server capabilities
  * Make sure 404's are handle gracefully as though none of the capabilities were set
  * Errors besides 404, such as server not available, cause client creation to fail just like default health checks do. The difference is we don't have a client-side backoff capability, we fail fast.
* Leverage the `internal_error_differentiation` capability to, if true, stop gRPC retrying internal errors

cretz:
## What was changed

* Remove health checking
* On every connection, call `WorkflowService.GetServerInfo` to eagerly obtain the capabilities of the server (or ignore if call is not there, it was added in https://github.com/temporalio/temporal/pull/2309)
* If the capabilities say we don't need to retry internal errors, don't

## Why?

We need to know from the server what it supports. Specifically in this case, we need to stop retrying internal errors if the server says it differentiates between them.

## Checklist

1. Closes #701

# ⚠️ WARNING - Breaking API Change

This makes the following backwards incompatible changes:

* Removes `client.ConnectionOptions.DisableHealthCheck`, `client.ConnectionOptions.HealthCheckAttemptTimeout`, and `client.ConnectionOptions.HealthCheckTimeout`. This means users of those options will have to change code on upgrade.
  * Since on-new-client health checking (default performed, but opt-out via `DisableHealthCheck`) is replaced by obtaining these capabilities, we removed the configuration options.
  * We now offer no option to disable eager connection to the server. This was chosen as the better option to allowing callers to disable our ability to obtain server capabilities.
    * This will require reworking for anyone using `DisableHealthCheck` (which includes our own server), to not create the client until the server is running.
    * Arguably one should not use a client that cannot connect to a server, and connecting later instead of fail-fast makes troubleshooting tougher.
  * It was decided to remove the options instead of deprecate them since they wouldn't be deprecated, but rather no longer used.
* Removes health checking from `NewNamespaceClient`.

The removal of health checking and the removal of eager-connect opt-out was discussed internally with @bergundy, @alexshtin, and @yiminc. We can of course take other approaches if there are concerns here.

@mfateev and @samarabbas - Please review.


client/client.go
```diff
@@ -431,12 +431,17 @@ type MetricsTimer = metrics.Timer
...
+// available.
 func NewClient(options Options) (Client, error) {
 	return internal.NewClient(options)
 }
 
-// NewNamespaceClient creates an instance of a namespace client, to manage lifecycle of namespaces.
+// NewNamespaceClient creates an instance of a namespace client, to manage
+// lifecycle of namespaces. This will not attempt to connect to the server
+// eagerly and therefore may not fail for an unreachable server until a call is
+// made.
```
bergundy, line 444: Maybe explain that you can use `WithBlock` to eagerly connect? Is that even possible?

bergundy (approved):

client/client.go
cretz: It is possible since we exposed gRPC options in #651. I will add the comment next commit (or right before merge if there are no reasons for another commit).

cretz:
Note, I am having CI failures I am struggling to replicate locally. Pardon the repeated forced pushes.

samarabbas (approved):

cretz:
I will be waiting for after next server release and after next SDK release before coming back to this

client/client.go
cretz: Done

cretz:
This has now been updated to remove `namespace` from the call and is ready for review/inclusion.

bergundy (approved):

PR #708, opened 2022-01-27
Log poll failures at a higher level

### LINKED ISSUE #707: Increase visbiility for poll failures

Right now, poll failures are logged at `DEBUG` level and even then, only after https://pkg.go.dev/go.temporal.io/sdk/worker#EnableVerboseLogging is set.

Confirm they aren't noisy then add them at WARN level. If at all possible, only do this on connection failure.

cretz:
## What was changed

Changed poll failures to not be behind a flag and raised level to `WARN`.

## Why?

These failures were being hidden. I have confirmed that during normal long poll, server sends an empty response after 1m, so this should only happen on real server error. I also added a simple check to make sure we don't log the same error over and over.

This will log each unique error for each worker. For example, if a server crashes during a worker, this is the default log output:

```
2022/01/27 08:50:36 WARN  Failed to poll for task. Namespace default TaskQueue task-queue-b173c2bd-5ebd-450e-864d-aa1e0a3b2fcd WorkerID 18592@cretz-laptop@ WorkerType ActivityWorker Error error reading from server: EOF
2022/01/27 08:50:36 WARN  Failed to poll for task. Namespace default TaskQueue task-queue-b173c2bd-5ebd-450e-864d-aa1e0a3b2fcd WorkerID 18592@cretz-laptop@ WorkerType WorkflowWorker Error error reading from server: EOF
2022/01/27 08:50:36 WARN  Failed to poll for task. Namespace default TaskQueue task-queue-b173c2bd-5ebd-450e-864d-aa1e0a3b2fcd WorkerID 18592@cretz-laptop@ WorkerType ActivityWorker Error last connection error: connection closed before server preface received
2022/01/27 08:50:36 WARN  Failed to poll for task. Namespace default TaskQueue task-queue-b173c2bd-5ebd-450e-864d-aa1e0a3b2fcd WorkerID 18592@cretz-laptop@ WorkerType WorkflowWorker Error last connection error: connection closed before server preface received
2022/01/27 08:50:40 WARN  Failed to poll for task. Namespace default TaskQueue task-queue-b173c2bd-5ebd-450e-864d-aa1e0a3b2fcd WorkerID 18592@cretz-laptop@ WorkerType ActivityWorker Error last connection error: connection error: desc = "transport: Error while dialing dial tcp [::1]:7233: connectex: No connection could be made because the target machine actively refused it."
2022/01/27 08:50:40 WARN  Failed to poll for task. Namespace default TaskQueue task-queue-b173c2bd-5ebd-450e-864d-aa1e0a3b2fcd WorkerID 18592@cretz-laptop@ WorkerType WorkflowWorker Error last connection error: connection error: desc = "transport: Error while dialing dial tcp [::1]:7233: connectex: No connection could be made because the target machine actively refused it."
```

That can seem noisy because it's workflow and activity workers each separately reporting: 1) the poll call getting interrupted, 2) the connection pool failing to get a connection, and 3) the literal connection attempt failing. But after these logs, it stops there until reconnected.

## Checklist

1. Closes #707


internal/internal_worker_base.go
```diff
@@ -304,10 +307,13 @@ func (bw *baseWorker) pollTask() {
...
 	if bw.pollLimiter == nil || bw.pollLimiter.Wait(bw.limiterContext) == nil {
 		task, err = bw.options.taskWorker.PollTask()
-		if err != nil && enableVerboseLogging {
-			bw.logger.Debug("Failed to poll for task.", tagError, err)
-		}
 		if err != nil {
+			// Log the error as warn if it doesn't match the last error seen
+			if err.Error() != bw.lastPollTaskErrMessage {
+				bw.logger.Warn("Failed to poll for task.", tagError, err)
+				bw.lastPollTaskErrMessage = err.Error()
```
mmcshane, line 314: Just want to verify that we're definitely single-threaded through here?

mmcshane (approved):
Just the one question. I expect the answer is that it's fine.

internal/internal_worker_base.go
bergundy: It might make sense to log every few attempts to increase visibility otherwise it might seem like the process is stuck.

cretz: > Just want to verify that we're definitely single-threaded through here?

A good question and definitely yes. It was the reason for the comment:

> Only used during in pollTask to keep from spamming logs

Though I should have said "in pollTask non-concurrently"

cretz: > It might make sense to log every few attempts to increase visibility otherwise it might seem like the process is stuck.

Not sure such a message counter mechanism is that warranted but if you think it is needed I can add it

bergundy: I would be a bit more noisy here to notify users that the SDK is still retrying

cretz: Ok, I will add the counter mechanism. I am not a fan of noise or extra code/logic for these conditions, but I don't have a strong enough opinion.

cretz: I added a lock (turns out it was needed, sorry @mmcshane) and suppressed duplicate error logs for a certain period

mmcshane (approved):

Sushisource (approved):

PR #713, opened 2022-02-03
Return workflow trace logger

jlegrone:
## What was changed

Fixes a bug in https://github.com/temporalio/sdk-go/pull/655 where a custom tracing logger was instantiated but never used. Realized my silly mistake when we cut over our tracing implementation and logs disappeared from APM.

This has now been tested e2e for workflow logs. I'll open a second patch to implement GetLogger for activities.


jlegrone:
Closing in favor of #714

PR #714, opened 2022-02-03
Support trace logger in workflows & activities

jlegrone:
## What was changed

This PR enables the injection of tracing metadata in the activity logger. This builds on #713 and #655.

## Why?

This allows tracing implementations to correlate logs for multiple activities and workflows serving a single request.

## Checklist

Tested e2e with our internal tracer implementation.

### DIFF
--- a/interceptor/tracing_interceptor.go
+++ b/interceptor/tracing_interceptor.go
@@ -291,9 +291,27 @@ func (t *tracingClientOutboundInterceptor) QueryWorkflow(
 	return val, err
 }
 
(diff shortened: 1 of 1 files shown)


cretz (approved):
Looks like this includes #713 if you want to close that one.

Ideally there'd be a test case here just confirming it's invoked inside the workflow/activity, but you'd have to make your own fake/mock tracer impl to test it.

I'll give it a bit to see if anyone else wants to review before merging.

cretz:
Aight, people have had enough time :-)

PR #715, opened 2022-02-03
Add workflow/activity start time and activity ID to tracing interceptor spans

jlegrone:
## What was changed

* Added a `Time` field to `TracerStartSpanOptions`. This allows tracing implementations to set correct durations for spans that cross process boundaries.
* Added a `temporalActivityID` tag to `RunActivity` spans. This allows filtering for spans that are retries of the same activity within a workflow execution.

## Why?

After https://github.com/temporalio/sdk-go/pull/714 these should be the last two changes needed for us to cut over our internal tracing implementation to sdk-go's `interceptor` package.

## Checklist

Tested end to end in our environment, but I'm also happy to write some unit tests for this functionality.

### DIFF
--- a/interceptor/tracing_interceptor.go
+++ b/interceptor/tracing_interceptor.go
@@ -25,8 +25,10 @@ package interceptor
 import (
 	"context"
 	"fmt"
+	"time"
 
 	commonpb "go.temporal.io/api/common/v1"
+
 	"go.temporal.io/sdk/activity"
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/converter"
@@ -113,6 +115,12 @@ type TracerStartSpanOptions struct {
 
 	// Name is the specific activity, workflow, etc for the operation.
 	Name string
+
+	// Time indicates the start time of the span.
+	//
+	// For RunWorkflow and RunActivity operation types, this will match workflow.Info.WorkflowStartTime and
+	// activity.Info.StartedTime respectively. All other operations use time.Now().
+	Time time.Time
 
 	// DependedOn is true if the parent depends on this span or false if it just
 	// is related to the parent. In OpenTracing terms, this is true for "ChildOf"
@@ -207,6 +215,7 @@ func (t *tracingClientOutboundInterceptor) ExecuteWorkflow(
 		Name:      in.WorkflowType,
 		Tags:      map[string]string{workflowIDTagKey: in.Options.ID},
 		ToHeader:  true,
+		Time:      time.Now(),
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


interceptor/tracing_interceptor.go
```diff
@@ -374,6 +388,7 @@ func (t *tracingWorkflowInboundInterceptor) HandleSignal(ctx workflow.Context, i
 			runIDTagKey:      info.WorkflowExecution.RunID,
 		},
 		FromHeader: true,
+		Time:       time.Now(),
```
jlegrone, line 391: Will we need a directive comment here to tell the determinism analysis plugin that this can be ignored?

cretz: Hrmm, I doubt it because I think the code traversal for analysis doesn't walk past the "temporal boundary" which interceptors sit behind. May be worth checking though.

cretz (approved):
Nice. I am happy this doesn't have any backwards incompatibilities.

> I'm also happy to write some unit tests for this functionality.

I would appreciate it. I think you can probably just add an assertion somewhere in the existing contrib tests for each impl to confirm the time is set on the span (probably not centrally at `internal/interceptortest/tracing.go` since each impl represents start time its own way). If it's too much hassle though, no worries.

I'll leave this open just a bit for other feedback, if any, before merging.

jlegrone:
> add an assertion somewhere in the existing contrib tests for each impl to confirm the time is set on the span

I took a crack at this, but realized that both opentracing and opentelemetry set span start time to `time.Now()` when passed a zero value. I ended up writing a full `tracer.Tracer` instead in order to capture the start span options instead. It's a bit of boilerplate but I think it'll be useful for testing other stuff like span tags as well.

interceptor/tracing_interceptor_test.go
```diff
@@ -0,0 +1,63 @@
...
+func (t *testTraceInterceptor) ContextWithSpan(ctx context.Context, span interceptor.TracerSpan) context.Context {
+	return ctx
+}
+
+func (t *testTraceInterceptor) StartSpan(options *interceptor.TracerStartSpanOptions) (interceptor.TracerSpan, error) {
+	// Require start time to be set
+	if options.Time.IsZero() {
+		switch options.Operation {
+		case "RunWorkflow", "RunActivity":
+		// Do nothing; the test env doesn't set these at the moment.
```
jlegrone, line 52: I haven't had time to look into this more, but it seems that workflow and activity start times aren't being set in workflow and activity info structs passed to interceptors in the test environment.

cretz:
Aight, people have had enough time :-) I hope you don't mind I updated your branch with master.

PR #717, opened 2022-02-05
Add a helper for serving a payload encoder over HTTP.

robholland:
This will be used to replace the data converter plugin infrastructure with a more flexible approach.

## What was changed

A helper was added to implement the ServeHTTP interface for a payload encoder.
A new data converter type is provided to consume the encoder endpoint.

## Why?

This will allow tctl to encode/decode payloads without needing to use plugins, just a flag to configure the remote encoder endpoint. Currently tctl is not able to encode payloads at all, only decode, due to issues with the plugin interface.

## Checklist

2. How was this tested:

Tested by patching tctl include this dataconverter and enabling it when a `--remote-data-encoder ...` flag is given.

3. Any docs updates needed?

Yes, the deployment and configuration of this setup will need to be documented.

### DIFF
--- a/converter/encoding_data_converter.go
+++ b/converter/encoding_data_converter.go
@@ -25,8 +25,12 @@ package converter
 import (
 	"bytes"
 	"compress/zlib"
+	"encoding/json"
 	"io/ioutil"
+	"net/http"
+	"strings"
 
+	"github.com/gogo/protobuf/jsonpb"
 	"github.com/gogo/protobuf/proto"
 	commonpb "go.temporal.io/api/common/v1"
 )
@@ -243,4 +247,63 @@ func partiallyClonePayload(p *commonpb.Payload) *commonpb.Payload {
 		ret.Metadata[k] = v
 	}
 	return ret
+}
+
+type encoderHTTPHandler struct {
+	encoder PayloadEncoder
+}
+
+func (e *encoderHTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "POST" {
+		http.NotFound(w, r)
+		return
+	}
+
+	if !strings.HasSuffix(r.URL.Path, "/encode") &&
+		!strings.HasSuffix(r.URL.Path, "/decode") {
+		http.NotFound(w, r)
+		return
+	}
+
+	var p commonpb.Payload
+
+	if r.Body == nil {
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+		return
+	}
+
+	err := jsonpb.Unmarshal(r.Body, &p)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	switch {
+	case strings.HasSuffix(r.URL.Path, "/encode"):
+		err = e.encoder.Encode(&p)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	case strings.HasSuffix(r.URL.Path, "/decode"):
+		err = e.encoder.Decode(&p)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	default:
+		http.NotFound(w, r)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	err = json.NewEncoder(w).Encode(p)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+}
+
+func NewPayloadEncoderHTTPHandler(e PayloadEncoder) http.Handler {
+	return &encoderHTTPHandler{e}
 }
--- a/converter/remote_data_converter.go
+++ b/converter/remote_data_converter.go
@@ -1,0 +1,92 @@
+// The MIT License
+//
+// Copyright (c) 2022 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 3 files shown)


robholland:
Converted to draft while I line up other PRs to show the usage of this code.

robholland:
Full working example of sdk, tctl and ui with relevant changes: https://github.com/temporalio/background-checks/pull/30

converter/encoding_data_converter.go
```diff
@@ -244,3 +249,64 @@ func partiallyClonePayload(p *commonpb.Payload) *commonpb.Payload {
...
+
+	w.Header().Set("Content-Type", "application/json")
+	err = json.NewEncoder(w).Encode(p)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+}
+
+func NewPayloadEncoderHTTPHandler(e PayloadEncoder) http.Handler {
```
cretz, line 310: Need Godoc on here and other exposed things

converter/remote_data_converter.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		return nil
+	}
+
+	message, _ := io.ReadAll(response.Body)
+	return fmt.Errorf("%s: %s", http.StatusText(response.StatusCode), message)
+}
+
+func (rdc *remotePayloadConverter) Encode(p *commonpb.Payload) error {
+	return rdc.sendHTTP(rdc.options.Endpoint+ENCODE_PATH, p)
```
cretz, line 87: Consider either `path.Join` or stripping the trailing slash from endpoint on `NewRemoteEncoderDataConverter`

converter/remote_data_converter.go
```diff
@@ -0,0 +1,92 @@
...
+	"net/http"
+
+	"github.com/gogo/protobuf/jsonpb"
+	commonpb "go.temporal.io/api/common/v1"
+)
+
+const ENCODE_PATH = "/encode"
+const DECODE_PATH = "/decode"
+
+type RemoteEncoderDataConverterOptions struct {
```
cretz, line 39: Should add a non-pointer HTTP client (because it's default initialized) for two reasons: 1) they may have, say, custom TLS options, and 2) `http.DefaultTransport` may not be worth reusing.

cretz (commented):
Looks great! Will share with the team to get general feedback on approach.

converter/remote_data_converter.go
```diff
@@ -0,0 +1,92 @@
...
+	Endpoint string
+}
+
+type remotePayloadConverter struct {
+	options RemoteEncoderDataConverterOptions
+}
+
+func NewRemoteEncoderDataConverter(options RemoteEncoderDataConverterOptions) *EncodingDataConverter {
+	return NewEncodingDataConverter(
+		defaultDataConverter,
```
cretz, line 49: The underlying converter should be something the user can pass in

converter/encoding_data_converter.go
```diff
@@ -244,3 +252,127 @@ func partiallyClonePayload(p *commonpb.Payload) *commonpb.Payload {
...
+	return &remotePayloadEncoder{options}
+}
+
+func (rdc *remotePayloadEncoder) sendHTTP(endpoint string, p *commonpb.Payload) error {
+	payload, err := json.Marshal(p)
+	if err != nil {
+		return fmt.Errorf("unable to marshal payload: %w", err)
+	}
+
+	req, err := http.NewRequest(http.MethodPost, endpoint, bytes.NewReader(payload))
```
cretz, line 344: Technically could add a timeout to this, but it's probably ok for now.

cretz (approved):
Looks great!

Sushisource:
This general approach seems good to me. Why the switch from WS to HTTP out of curiosity?

robholland:
Websockets is only a thing currently because browsers are not permitted to send HTTP traffic to localhost via `fetch`. We're still debating whether websockets will be needed to allow local development or whether an alternative can be found, but for non-localhost interaction with the web UI a HTTP interface is more inline with norms and easier to deploy and code (on both client and server side).

cretz:
> This general approach seems good to me. Why the switch from WS to HTTP out of curiosity?

FWIW, I would support adding a `/ws` endpoint on this handler (though in Go you have to use a third party dependency for WS support which is ug) that upgraded the HTTP connection and handled encode/decode req/resp.

robholland:
Yeah I can do that in contrib maybe as we discussed. Don't wanna pull the dep into main SDK.

robholland:
Also not sure how easy ws is for other SDKs? We'd want to support this across all SDKs to rid us of the current requirement that the data converter that the web UI uses has to be written in Go.

Sushisource:
@robholland Right, OK that makes sense, I thought this was still all about running things locally.

If it's not, then this needs to be https as well, no? Otherwise if I'm running my dataconverter "server" somewhere and decoding encrypted traffic and sending it back over plain HTTP I've kinda defeated the purpose. Or is the idea you'd always just run this on the same box that's running the UI server or something?

Updating the desc. with info about how this is expected to be deployed would be a good idea IMO

robholland:
For https you'd just mount this handler onto a http server which is serving https (or behind ALB or blah blah). Nothing about the code forces http versus https. Client side you'd just make the endpoint https://blah

bergundy:
> Websockets is only a thing currently because browsers are not permitted to send HTTP traffic to localhost via `fetch`. We're still debating whether websockets will be needed to allow local development or whether an alternative can be found, but for non-localhost interaction with the web UI a HTTP interface is more inline with norms and easier to deploy and code (on both client and server side).

I agree that HTTP request / response is easier to implement than websockets (not by much though).
Whatever solution we come up with it needs to work on localhost too, as long as we satisfy that I'm fine with changing the current websocket based solution.

On a more general note, moving away from tctl plugin for the remote data converter is definitely a step in the right direction.

robholland:
Running further tests it seems that we don't need to keep the websockets workaround. This HTTP system would work with both the UI and the encoder on localhost or both deployed with HTTPS. On Chrome it would also work with a UI deployed on HTTPS and an encoder running on localhost, however Safari will block that traffic.

robholland:
@Sushisource @bergundy Any concerns with merging this now? Once merged I can follow up with extending the encryption sample to include a remote data encoder server alongside the plugin and then work on some documentation with the docs team.

Sushisource:
@robholland No problems here. My only concern is a higher level one that we'll  need to make this work for every language users might be writing encoders in. Luckily it's not exactly a ton of work. Even if we had users write converters in JS/WASM, they'd still need a way to fetch & use encryption keys which is what this is all really about, and that wouldn't solve that problem.

bergundy:
@robholland are we assuming that users won't care about Safari support? Are there other browsers that won't support this approach?

All I care about is that we have a solution that's flexible enough, if we see that this approach is limited I vote against it.

cretz:
@bergundy - Safari works fine for localhost-to-localhost and https-to-https, it's https-to-localhost or vice-versa that they balk at due to them violating the spec by not considering localhost a secure origin. See https://bugs.webkit.org/show_bug.cgi?id=171934.

robholland:
Safari does not support these cases for websockets either, so sticking with websockets would not help here.

bergundy (approved):

PR #725, opened 2022-02-11
Address rare custom resolver test race

mmcshane:
## What was changed

Ensure that the test code doesn't race against grpc connection
establishment in a way that results in test failures.

## Why?

Test failures are bad

## Checklist

1. Closes 

2. How was this tested:

I was able to reproduce a failure in as few as 12 test runs but sometimes it took more
than 2000 runs to reproduce. That said, this test failed in two of the
last four CI runs so it definitely has an impact.

3. Any docs updates needed?

### DIFF
--- a/internal/grpc_dialer_test.go
+++ b/internal/grpc_dialer_test.go
@@ -220,8 +220,10 @@ func TestCustomResolver(t *testing.T) {
 	var peerOut peer.Peer
 	for len(connected) < 2 {
 		req.RequestId = uuid.New()
-		client.WorkflowService().SignalWorkflowExecution(context.Background(), &req, grpc.Peer(&peerOut))
-		connected[peerOut.Addr] = struct{}{}
+		_, err := client.WorkflowService().SignalWorkflowExecution(context.Background(), &req, grpc.Peer(&peerOut))
+		if err == nil {
+			connected[peerOut.Addr] = struct{}{}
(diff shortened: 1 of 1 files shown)


internal/grpc_dialer_test.go
```diff
@@ -199,31 +202,57 @@ func TestCustomResolver(t *testing.T) {
...
+	// https://github.com/grpc/grpc-go/blob/bd7076973b45b81e37a45eb761efb789e2001618/balancer/roundrobin/roundrobin_test.go#L196-L212
+	connected := map[net.Addr]struct{}{}
+	req := workflowservice.SignalWorkflowExecutionRequest{
+		WorkflowExecution: &common.WorkflowExecution{WorkflowId: "workflowid", RunId: "runid"},
+		SignalName:        "signal",
+		Namespace:         DefaultNamespace,
+		Identity:          t.Name(),
+	}
+	var peerOut peer.Peer
+	for len(connected) < 2 {
```
cretz, line 221: `connected` is always empty here, right?

mmcshane: It starts as empty but the body of the loop inserts into it

cretz: Sorry, I misread! Thanks!

cretz (approved):

PR #727, opened 2022-02-14
Disallow MaxConcurrentWorkflowTaskPollers to be set to 1

### LINKED ISSUE #719: Worker silently doing nothing if MaxConcurrentWorkflowTaskPoller set to 1
Set MaxConcurrentWorkflowTaskPoller to 1 and start the worker, worker will start with no error. But if you try to start a workflow, it won't make any progress. This is because worker prefer sticky queue, so it always pick from sticky queue if max concurrent workflow task poller is limit to 1.
We either need to be smarter and poll from normal queue from time to time, or we need to fail the worker and document that we need at least 2 workflow task poller. 

Current silently doing nothing is not acceptable.

cretz:
## What was changed

Disallow MaxConcurrentWorkflowTaskPollers to be set to 1

## Why?

Because worker polling is built to require at least 2 due to how stickiness preference is implemented.

## Checklist

1. Closes #719

### DIFF
--- a/internal/worker.go
+++ b/internal/worker.go
@@ -85,7 +85,9 @@ type (
 
 		// Optional: Sets the maximum number of goroutines that will concurrently poll the
 		// temporal-server to retrieve workflow tasks. Changing this value will affect the
-		// rate at which the worker is able to consume tasks from a task queue.
+		// rate at which the worker is able to consume tasks from a task queue. Due to
+		// internal logic where pollers alternate between stick and non-sticky queues, this
+		// value cannot be 1 and will panic if set to that value.
 		// default: 2
 		MaxConcurrentWorkflowTaskPollers int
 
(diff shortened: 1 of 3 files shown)


internal/worker.go
```diff
@@ -85,7 +85,9 @@ type (
 
 		// Optional: Sets the maximum number of goroutines that will concurrently poll the
 		// temporal-server to retrieve workflow tasks. Changing this value will affect the
-		// rate at which the worker is able to consume tasks from a task queue.
+		// rate at which the worker is able to consume tasks from a task queue. Due to
+		// internal logic where pollers alternate between stick and non-sticky queues, this
+		// value cannot be 1.
```
Sushisource, line 90: Should mention we'll panic if it's set to 1. 

Or, consider saying "we'll just set it to 2" and not panicking

cretz: I will say we panic. I want to panic in this situation where they explicitly set the wrong value (I'd return an error if the API supported it).

Sushisource (approved):

PR #738, opened 2022-02-24
Rh encoding interceptor

robholland:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


client/generate_proxy.go
```diff
@@ -0,0 +1,92 @@
+// +build ignore
```
cretz, line 1: You'll need license headers on these files (for the generated code you might need to alter `/internal/cmd/tools/copyright/licensegen.go` if you don't want to add a header there)

converter/interceptor.go
```diff
@@ -0,0 +1,215 @@
+// The MIT License
```
cretz, line 1: Probably best named as `grpc_interceptor.go` to avoid overloading the term as we otherwise use it

converter/interceptor.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (s *serviceInterceptor) processCommands(commands []*command.Command) error {
+	var err error
+	for _, c := range commands {
+		switch c.CommandType {
```
cretz, line 97: A type switch is a bit safer. A command type set without the attributes will cause a panic here. Same for event type later.

converter/interceptor.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	return nil
+}
+
+// NewPayloadEncoderGRPCServerInterceptor returns a GRPC Server Interceptor that will mimic the encoding
+// that the SDK system would perform when configured with a matching EncodingDataConverter.
+// Note: This approach does not support use cases that rely on the ContextAware DataConverter interface as
+// workflow context is not available at the GRPC level.
+func NewPayloadEncoderGRPCServerInterceptor(encoders ...PayloadEncoder) grpc.UnaryServerInterceptor {
```
cretz, line 153: What is the use case for an encoding server interceptor?

converter/interceptor.go
```diff
@@ -0,0 +1,215 @@
...
+
+		return s.processResponse(resp), err
+	}
+}
+
+// NewPayloadEncoderGRPCClientInterceptor returns a GRPC Client Interceptor that will mimic the encoding
+// that the SDK system would perform when configured with a matching EncodingDataConverter.
+// Note: This approach does not support use cases that rely on the ContextAware DataConverter interface as
+// workflow context is not available at the GRPC level.
+func NewPayloadEncoderGRPCClientInterceptor(encoders ...PayloadEncoder) grpc.UnaryClientInterceptor {
```
cretz, line 175: Might want to take an options struct here like you did w/ the proxy for future proofing (and return an error).

converter/interceptor.go
```diff
@@ -0,0 +1,215 @@
...
+		if err != nil {
+			return err
+		}
+
+		return s.processResponse(response)
+	}
+}
+
+func encodePayloads(payloads *commonpb.Payloads, encoders ...PayloadEncoder) error {
+	for _, payload := range payloads.Payloads {
```
cretz, line 194: You need to check whether `payloads` is nil here or call `payloads.GetPayloads()` instead

client/generate_proxy.go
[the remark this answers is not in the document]
robholland: Used generateproxy

converter/interceptor.go
robholland: There are no types are there? afaik it's one type shared by all possible types of commands.

converter/interceptor.go
robholland: It feels as relevant as the client side one to me. For the proxy use case we can use either but I don't know if others would need server specifically.

converter/interceptor.go
cretz: There are, see the docs for https://pkg.go.dev/go.temporal.io/api/command/v1#Command. But I'm fine with what you have here, but may want to change `.Input` to `.GetInput()` which is safe from `GetXAttributes()` coming back nil.

converter/interceptor.go
cretz: Server makes less sense in my head since I don't expect many people to run a server, but it's pretty cheap to add here, so I don't mind.

bergundy (commented):
Looks like failure and header payloads aren't converted here

robholland:
Headers are not touched as the SDK doesn't use a data converter on the headers, the header related APIs expect the user to convert to/from Payload themselves (through whichever data converter they prefer). I'll check on Failure. I think the SDK may touch the Payloads for Details, but not any Payload outside of a Payloads wrapper.

bergundy:
> Headers are not touched as the SDK doesn't use a data converter on the headers, the header related APIs expect the user to convert to/from Payload themselves (through whichever data converter they prefer). I'll check on Failure. I think the SDK may touch the Payloads for Details, but not any Payload outside of a Payloads wrapper.

Yeah but the whole point of the proxy is so worker and client deployment doesn't get access to encryption secrets, I'm imagining users will use the default data converter here.

bergundy:
One of the issues with this PR's approach is that if there are any proto changes we'd miss some fields.
This is obviously fine for an initial PR and we don't have to solve it now but we should have either static analysis or comprehensive testing.

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,268 @@
...
+		return s.encodePayloads(r.Details)
+	case *workflowservice.RecordActivityTaskHeartbeatByIdRequest:
+		return s.encodePayloads(r.Details)
+	}
+
+	return nil
+}
+
+func (s *serviceInterceptor) processResponse(response interface{}) error {
+	switch r := response.(type) {
```
bergundy, line 107: `QueryWorkflowResponse` is missing here.
Anything with a header is missing too.

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,268 @@
...
+	return nil
+}
+
+func (s *serviceInterceptor) encodeFailure(failure *failure.Failure) error {
+	if failure == nil {
+		return nil
+	}
+
+	err := s.encodePayloads(
+		failure.GetApplicationFailureInfo().GetDetails(),
```
bergundy, line 125: Missing `TimeoutFailureInfo.last_heartbeat_details`

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,268 @@
...
+		return err
+	}
+
+	return s.encodeFailure(failure.GetCause())
+}
+
+func (s *serviceInterceptor) processCommands(commands []*command.Command) error {
+	var err error
+	for _, c := range commands {
+		switch c.CommandType {
```
bergundy, line 156: Missing:
- `CancelWorkflowExecutionCommandAttributes.details`
- `RecordMarkerCommandAttributes`

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,268 @@
...
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func (s *serviceInterceptor) processRequest(req interface{}) error {
+	switch r := req.(type) {
```
bergundy, line 72: Missing `QueryWorkflowRequest.query.query_args` and `RespondWorkflowTaskCompletedRequest.query_results.*.answer`

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,268 @@
...
+
+	return nil
+}
+
+func (s *serviceInterceptor) processResponse(response interface{}) error {
+	switch r := response.(type) {
+	case *workflowservice.GetWorkflowExecutionHistoryResponse:
+		return s.processEvents(r.History.GetEvents())
+	case *workflowservice.PollWorkflowTaskQueueResponse:
+		return s.processEvents(r.History.GetEvents())
```
bergundy, line 111: Missing `query` and `queries`

bergundy (commented):
Nice work @robholland.

I found a few missing conversions.

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,268 @@
...
+
+	return s.encodeFailure(failure.GetCause())
+}
+
+func (s *serviceInterceptor) processCommands(commands []*command.Command) error {
+	var err error
+	for _, c := range commands {
+		switch c.CommandType {
+		case enumspb.COMMAND_TYPE_SCHEDULE_ACTIVITY_TASK:
+			err = s.encodePayloads(c.GetScheduleActivityTaskCommandAttributes().Input)
```
cretz, line 158: Might want to use `GetInput()` on these just in case for whatever reason the attributes are nil

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+type PayloadEncoderGRPCServerInterceptorOptions struct {
```
cretz, line 218: Need docs on all exported items

robholland:
I've got an auto-generated version of the interceptor nearly working which would use an opt-out approach to exclude Payload(s) we don't want to be touched. Chad has suggested a test which will re-walk the API definitions and check if the list on which the interceptor is built has drifted from the current API version so that we can ensure decisions are made about new Payload(s) as they are introduced (or removed for that matter).

bergundy:
Which payloads do we not want to touch?

cretz:
When https://github.com/temporalio/api/pull/153 lands, it'll be a good example of whether the test will fail when we upgrade the API dependency but don't re-gen code.

bergundy:
> I've got an auto-generated version of the interceptor nearly working which would use an opt-out approach to exclude Payload(s) we don't want to be touched. Chad has suggested a test which will re-walk the API definitions and check if the list on which the interceptor is built has drifted from the current API version so that we can ensure decisions are made about new Payload(s) as they are introduced (or removed for that matter).

Nice! That's what I was going for

converter/interceptor.go
robholland: This is no longer relevant.

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,1037 @@
...
+// Currently this is just the list of encoders to use.
+type PayloadEncoderGRPCClientInterceptorOptions struct {
+	Encoders []PayloadEncoder
+}
+
+// NewPayloadEncoderGRPCClientInterceptor returns a GRPC Client Interceptor that will mimic the encoding
+// that the SDK system would perform when configured with a matching EncodingDataConverter.
+// Note: This approach does not support use cases that rely on the ContextAware DataConverter interface as
+// workflow context is not available at the GRPC level.
+func NewPayloadEncoderGRPCClientInterceptor(options PayloadEncoderGRPCClientInterceptorOptions) (grpc.UnaryClientInterceptor, error) {
```
cretz, line 1021: Nit - I'd move this and the options to the top of the file

internal/cmd/generateinterceptor/main.go
```diff
@@ -0,0 +1,412 @@
...
+					{{end}}
+				); err != nil { return err }
+{{end}}
+		}
+	}
+
+	return nil
+}
+
+func (s *serviceInterceptor) encode(objs ...interface{}) error {
```
cretz, line 168: Don't really need these if you're willing to pass the bool from inside the interceptor

internal/cmd/generateinterceptor/main.go
```diff
@@ -0,0 +1,412 @@
...
+			result[typ] = record
+		}
+	}
+
+	return result
+}
+
+// walk iterates the methods on a type and returns whether any of them can eventually lead to Payload(s)
+// The return type for each method on this type is walked recursively to decide which methods can lead to Payload(s)
+func walk(desired []types.Type, skip []types.Type, typ types.Type) bool {
```
cretz, line 296: You'll want to do a few things to make this reusable:

* Probably accept a stateful object as a param that you can build up instead of just bools
* Move this function outside of cmd (but still in a child package of internal) so it can be used by tests
* During walk, collect a set of payload paths that we do want to handle
* Serialize that set of payload paths into a variable in the generated code
* Write a test that does the walk and compares the resulting payload paths to the serialized variable

Also, there are technically other ways you could code-gen based on these paths (e.g. a map of collectors by message type that add payloads to a large set to be encoded/decoded at the end), but what you have is fine too.

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,1037 @@
...
+				if err := s.encodePayload(o); err != nil {
+					return err
+				}
+			} else {
+				if err := s.decodePayload(o); err != nil {
+					return err
+				}
+			}
+		case *commonpb.Payloads:
+			if encode {
```
cretz, line 99: Could just loop here like you do all other collections instead of having special methods for encode/decode payload, but doesn't matter much

converter/grpc_interceptor.go
```diff
@@ -0,0 +1,1037 @@
...
+			}
+		case map[string]*commonpb.Payload:
+			for _, x := range o {
+				if err := s.process(encode, x); err != nil {
+					return err
+				}
+			}
+
+		case *commandpb.CancelWorkflowExecutionCommandAttributes:
+			if o == nil {
```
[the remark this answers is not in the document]
robholland: Discussed on Slack. Due to the use of interfaces here the type matches do fire for nil pointers.

robholland:
`make check` now verifies via `make generatorcheck` that files generated based on the API are not stale.

cretz (approved):

cretz:
LGTM! Ignore the coveralls complaint (can squash/merge anyways once you get @bergundy's approval)

bergundy (approved):
Just approved.
I didn't go over the code again but I understand the new approach which is the right way to go IMO.

PR #739, opened 2022-02-24
Fix session retry and activity cancellation bug

### LINKED ISSUE #722: CreateSession improperly retrying errors it should not

At https://github.com/temporalio/sdk-go/commit/a88dec51b1767a1bab6426a602980d25a51d40eb#diff-5db8d6c4288666da4ff606f745645fe946e504370e93bbceb25402e0a520deb0 we inadvertently removed intentional retry-prevention to prevent session activity creation from retrying in all but the "too many sessions" case.

Do not have session creation activity failure retry anything but "too many sessions".

cretz:
## What was changed

* Disable retries on session creation activity
* Remove session creation's manual checking of max-session limit and instead reuse the worker activity limit
* Fix additional bug on #726 that surfaced during a session cancellation test

## Why?

See #722

Session creation activity (which is really "session run activity" since it spans the life of the session) was originally developed to retry only on a "too many sessions" error it manually returned when checking session count. It was even meant to not retry timeouts in case of worker death. When Temporal removed the ability to specifically disable retries for timeouts, this affected sessions.

Now, instead of the only reason for sessions to retry being a max-limit error, we disable retries on sessions and move max-limit to the activity worker itself. This allows us to remove in-session-creation-activity limit checks and other code that prevented polling during too many sessions.

Also, during development of an integration test, session cancellation without workflow cancellation was causing activity cancellation to have out-of-order command state machine issues similar to those seen in #726. #726 decremented a counter for all activity cancellations even though it incremented the counter only when the workflow was being cancelled. We added a check to make sure we only decrement that counter in the same scenario it was incremented.

## Checklist

1. Closes #722

### DIFF
--- a/internal/internal_decision_state_machine.go
+++ b/internal/internal_decision_state_machine.go
@@ -910,16 +910,20 @@ func (h *commandsHelper) removeCancelOfResolvedCommand(commandID commandID) {
 		delete(h.commands, commandID)
 		command := h.orderedCommands.Remove(orderedCmdEl)
 		// Sometimes commandsCancelledDuringWFCancellation was incremented before
-		// it was reset and sometimes not. We use the reset counter to see if we're
-		// still on the same iteration where we may have incremented it before.
-		switch command := command.(type) {
-		case *cancelActivityStateMachine:
-			if command.cancelledOnEventIDResetCounter == h.nextCommandEventIDResetCounter {
-				h.commandsCancelledDuringWFCancellation--
-			}
-		case *cancelTimerCommandStateMachine:
-			if command.cancelledOnEventIDResetCounter == h.nextCommandEventIDResetCounter {
-				h.commandsCancelledDuringWFCancellation--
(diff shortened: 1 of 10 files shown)


bergundy (commented):
Do we have integration test coverage for session recreation?
Does it still work and respect concurrent limits with this new approach?

cretz:
> Do we have integration test coverage for session recreation?
> Does it still work and respect concurrent limits with this new approach?

No, we never have (session creation and session recreation call the same internal code and the small part that is unique to recreation has unit test coverage). I will add it for peace of mind though.

cretz:
After further discussion, this incorrectly removes limits on session recreation by removing the session bucket. We are having internal discussions on this.

I am likely going to open the activity cancellation fix as a separate PR.

cretz:
I am closing this PR. This approach is not safe for session recreation. Will update #722 with the path forward.

PR #382, opened 2021-03-11
Fix an event ordering bug possible with outstanding commands during WF cancel

Sushisource:
This fixes #378 

The fix is... a bit suspicious. The Go state machine management in general desperately needs love. @vitarb and I want to replace it's guts with Rust core as soon as that's viable.

I wrote a replay test for the specific issue, and added an integ test that should account for a more elaborate but similar situation.

PR #741, opened 2022-02-24
Fix activity cancellation race when not cancelling workflow

cretz:
## What was changed

Only decrement the in-process cancelled counter during workflow cancel.

## Why?

In order to keep the event ID expectations correct, we keep a cancellation counter for things cancelled during workflow execution. This cancellation counter, which has been there for timers and reinstated in #726 for activities, is incremented upon cancel and decremented upon removal of the command. But while the increment was done only during workflow cancellation, the decrement was being done always. This makes sure that decrement is only done during workflow cancellation, when the increment occurs.

Whether we should be limiting the increment to only during workflow cancel done way back in #382, I am unsure.

The reason we saw this issue is during session cancellation which is _not_ workflow cancellation was decrementing that counter too far. This was noticed during #739. So regression tests have been added for them.

### DIFF
--- a/internal/internal_decision_state_machine.go
+++ b/internal/internal_decision_state_machine.go
@@ -910,16 +910,20 @@ func (h *commandsHelper) removeCancelOfResolvedCommand(commandID commandID) {
 		delete(h.commands, commandID)
 		command := h.orderedCommands.Remove(orderedCmdEl)
 		// Sometimes commandsCancelledDuringWFCancellation was incremented before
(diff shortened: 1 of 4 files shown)


Sushisource:
> Whether we should be limiting the increment to only during workflow cancel done way back in https://github.com/temporalio/sdk-go/pull/382, I am unsure.

This PR is doing that - right? Are you questioning if the change in the PR is the right thing to do?

Sushisource (approved):

PR #744, opened 2022-02-28
Add Sagas to workflows

sunny-b:
## What was changed

This PR adds Sagas to the sdk. This allows Go-SDK users to create [compensating transactions](https://en.wikipedia.org/wiki/Compensating_transaction). In essence, it allows users to easily specify rollback steps in their workflows that can be executed in the case that the workflow fails.

## Why?

Sagas already exist in the [Java-SDK](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/workflow/Saga.java) and this PR is would make the Go-SDK have better feature parity with the Java-SDK. Additionally, there are [third-party libraries](https://github.com/Courtsite/temporal-go-helpers/tree/master/saga) that exist for this already so there is desire for this to be a native feature.

## Checklist

1. This doesn't close a specific issue. However, I asked about it in the [Temporal Slack](https://temporalio.slack.com/archives/CTDTU3J4T/p1645585906354699) and it seemed like there was interest to create this.

2. How was this tested:

I created a pretty thorough [test suite](https://github.com/sunny-b/sdk-go/blob/d4b8fc5de367c6cd013d12e1d5bc0495e9d2ab66/internal/internal_saga_test.go) with a test workflow that tests various edge cases (if compensation fails, if saga is cancelled, if the workflow is cancelled, parallel execution, etc)

3. Any docs updates needed?

I didn't add anything to the README, though there is a Saga example in the [Java Examples repo](https://github.com/temporalio/samples-java/blob/main/src/main/java/io/temporal/samples/hello/HelloSaga.java) and it might be good to add an example to the Go examples repo as well.

### DIFF
--- a/internal/internal_saga.go
+++ b/internal/internal_saga.go
@@ -1,0 +1,172 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package internal
+
+// All code in this file is private to the package.
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+)
+
+const sagaOptionsContextKey = "sagaOptions"
+
+func getSagaOptions(ctx Context) *SagaOptions {
+	so := ctx.Value(sagaOptionsContextKey)
+	if so == nil {
+		return nil
+	}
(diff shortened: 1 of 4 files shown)


cretz:
Hrmm, I wonder if this is best at https://github.com/temporalio/samples-go/. We often keep even utilities there that are not directly needed by the core of the project (encryption converters, DSL-based workflows, Prometheus metrics, JWT-based auth, etc).

sunny-b:
@cretz If the consensus is that it's better in the samples repo, I can move it there. I mainly added it to the `sdk-go` to keep feature parity with the Java SDK.

internal/internal_saga.go
```diff
@@ -0,0 +1,190 @@
...
+type Saga struct {
+	ctx               Context
+	options           *SagaOptions
+	compensationStack []*compensationOp
+	once              sync.Once
+	cancelFunc        CancelFunc
+
+	// since the cancelFunc is non-blocking, we use a canceled flag and lock
+	// to ensure compensation steps aren't run after the cancelFunc is called.
+	canceled bool
```
sunny-b, line 72: I ran into a strange edge case in one of my tests where some compensation steps were executed even after the Saga context was canceled. The activity would return a `canceled` error but for some reason the activity was still executed. I'm not sure if that's expected or not. It was so not consistent either because sometimes the activity wouldn't execute but sometimes it did.

It didn't seem like a great user experience so I added this flag to prevent `Compensate` from running after `Cancel` is called.

internal/internal_saga.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// Saga implements the logic to execute compensation operations
+// https://en.wikipedia.org/wiki/Compensating_transaction that is
+// often required in Saga applications.
+type Saga struct {
+	ctx               Context
+	options           *SagaOptions
+	compensationStack []*compensationOp
+	once              sync.Once
```
cretz, line 67: We cannot use Go synchronization primitives in workflows.

internal/internal_saga.go
```diff
@@ -0,0 +1,190 @@
...
+	cancelFunc        CancelFunc
+
+	// since the cancelFunc is non-blocking, we use a canceled flag and lock
+	// to ensure compensation steps aren't run after the cancelFunc is called.
+	canceled bool
+	lock     sync.RWMutex
```
cretz, line 73: Same here, Go sync primitives cannot be used (and is never required anyways because all workflow execution is single-threaded)

cretz:
Thanks! Yeah, there are some regrets about the Java one being part of the SDK since it's so simple.

Still discussing internally, but we might want to move this to samples-go. Cleanup actions are a pretty basic concept (though many may not recognize the name "saga") and of course supported in Go with defer and disconnected context and what not. And of course multi-activity execution (sequential and parallel) and error management on them are not unique to cleanup.

I'll try to update soon.

cretz:
After some internal discussion, it was determined that the Java Saga implementation and this one don't add much over someone just looping themselves. But we may add more full-featured "Saga" support in the future to SDKs, but we need consensus on exactly what to do for all SDKs (see https://github.com/temporalio/sdk-features/issues/50).

Closing for now.

PR #746, opened 2022-03-01
Disable timeout retries during session creation

### LINKED ISSUE #722: CreateSession improperly retrying errors it should not

cretz:
## What was changed

* Disable retries of timeout errors during session creation
* Limit session creation worker by max session size
* Add test to confirm session creation across workers
* Made Godoc note that sessions do not work across workers before server version v1.15.1 (will add to public docs too)

## Why?

See #722. Basically [we inadvertently removed](https://github.com/temporalio/sdk-go/commit/a88dec51b1767a1bab6426a602980d25a51d40eb#diff-5db8d6c4288666da4ff606f745645fe946e504370e93bbceb25402e0a520deb0) the disabling of retry on timeouts for session creation. So when a session creation was replayed on another worker (i.e. after original worker crash/stop), the session creation would hang. Now that https://github.com/temporalio/temporal/pull/2524 was cherry picked for v1.15.1, we can stop retrying those errors again.

In #739 we had originally tried fixing this by removing session retries altogether, since they're only needed for max-session limit error, and just putting a limit on the session creation worker, but this didn't allow the max limit number to stretch across session recreation too. We abandoned that approach since it didn't work for session recreation

## Checklist

1. Closes #722


internal/session.go
```diff
@@ -290,10 +290,11 @@ func createSession(ctx Context, creationTaskqueue string, options *SessionOption
...
 	retryPolicy := &RetryPolicy{
-		InitialInterval:    time.Second,
-		BackoffCoefficient: 1.1,
-		MaximumInterval:    time.Second * 10,
-		MaximumAttempts:    0,
+		InitialInterval:        time.Second,
+		BackoffCoefficient:     1.1,
+		MaximumInterval:        time.Second * 10,
+		MaximumAttempts:        0,
+		NonRetryableErrorTypes: []string{"TemporalTimeout:StartToClose", "TemporalTimeout:Heartbeat"},
```
Sushisource, line 297: Updating this comment at the top to explain why the non retryable types are the way they are would be good

test/integration_test.go
```diff
@@ -1998,6 +2004,43 @@ func (ts *IntegrationTestSuite) TestMaxConcurrentSessionExecutionSizeForRecreati
...
+	ts.registerWorkflowsAndActivities(nextWorker)
+	ts.NoError(nextWorker.Start())
+	defer nextWorker.Stop()
+
+	// Get the result of the workflow run now
+	err = run.Get(ctx, nil)
+	// We expect the activity to be cancelled. Before the issue that was fixed
+	// when this test was written, this would hang because sessions would
+	// inadvertently retry.
+	ts.Error(err)
```
Sushisource, line 2040: This is cancelled because it timed out, right? Mentioning that would help clarify what the test does

internal/session.go
```diff
@@ -288,12 +288,16 @@ func createSession(ctx Context, creationTaskqueue string, options *SessionOption
 	}
 
 	taskqueueChan := GetSignalChannel(ctx, sessionID) // use sessionID as channel name
-	// Retry is only needed when creating new session and the error returned is NewApplicationError(errTooManySessionsMsg)
+	// Retry is only needed when creating new session and the error returned is
+	// NewApplicationError(errTooManySessionsMsg). Therefore we make sure to
```
yycptt, line 292: Please mark this error as retryable on L530.

cretz: Thanks! I may need a test here

cretz: I have fixed this and added a test to confirm. I inadvertently learned that we can't call `workflow.CompleteSession` in a separate `workflow.Go` coroutine.

Sushisource (approved):

yycptt (approved):

test/integration_test.go
```diff
@@ -1976,26 +1982,87 @@ func (ts *IntegrationTestSuite) TestMaxConcurrentSessionExecutionSize() {
...
+			RecreateAtIndex:        1,
+		})
+	ts.NoError(err)
+	// Wait until sessions created
+	ts.waitForQueryTrue(run2, "sessions-created-equals-and-pending", 1)
+
+	// Now let the activities complete which lets run1 complete and free up
+	// sessions for run2
+	manualCancel()
+	ts.NoError(run2.Get(ctx, nil))
```
bergundy, line 2027: Is it worth checking that run1 completes?
We usually verify all workflows finish in Core/TS tests for good measure.

bergundy (approved):

internal/internal_worker.go
```diff
@@ -362,6 +362,9 @@ func newSessionWorker(service workflowservice.WorkflowServiceClient, params work
 
 	params.MaxConcurrentActivityTaskQueuePollers = 1
 	params.TaskQueue = creationTaskqueue
+	// Although we have session token bucket to limit session size across creation
+	// and recreation, this is just an added limit for creation only just in case
+	params.ConcurrentActivityExecutionSize = maxConcurrentSessionExecutionSize
```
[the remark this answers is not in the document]
cretz: Does it limit retries? The poller doesn't poll if the session bucket limit has been reached. Either way, remove this part of the comment.

test/integration_test.go
cretz: I will add the check.

internal/internal_worker.go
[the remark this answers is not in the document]
cretz: Done.

test/integration_test.go
cretz: Done

workflow/session.go
```diff
@@ -99,6 +99,8 @@ var ErrSessionFailed = internal.ErrSessionFailed
 //        // Handle activity error
 //    }
 //    ... // execute more activities using sessionCtx
+//
+// NOTE: Sessions may not work properly across worker fail/crash before Temporal server version v1.15.1.
```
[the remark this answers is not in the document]
cretz: Added.

PR #748, opened 2022-03-08
Add mock call assertions to TestWorkflowEnvironment

Sikian:
## What was changed
Added `AssertCalled`, `AssertNotCalled` and `AssertNumberOfCalls` to `TestWorkflowEnvironment` to enable a straightforward way to assert for activity calls within workflows. 

They have been implemented following the same approach as https://pkg.go.dev/github.com/stretchr/testify/mock#Mock.

## Why?
Asserting how an activity has been called from within a workflow (e.g. storing/retrieve data from a db) can be cumbersome without assertions on how mocks have been called. 

The current best approach are:
- storing the arguments in variables to later assert them 
- asserting within the mocks handler function (which leads to very long stacks when the assertion fails)

The testify library provides `AssertCalled`, `AssertNotCalled` and `AssertNumberOfCalls` for this purpose, making the assertions straight-forward to understand and debug.

## Checklist

1. How was this tested:
Two unit tests have been implemented to test these cases.

3. Any docs updates needed?
In https://docs.temporal.io/docs/go/how-to-test-workflow-definitions-in-go/, `Test_SimpleWorkflow_ActivityParamCorrect` could be updated to read
```go
func (s *UnitTestSuite) Test_SimpleWorkflow_ActivityParamCorrect() {
        s.env.OnActivity(SimpleActivity, mock.Anything, mock.Anything).Return("", nil)
        s.env.ExecuteWorkflow(SimpleWorkflow, "test_success")

        s.env.AssertCalled(s.T(), SimpleActivity, mock.Anything, "test_success")
        s.env.AssertNumberOfCalls(s.T(), SimpleActivity, 1)
        s.True(s.env.IsWorkflowCompleted())
        s.NoError(s.env.GetWorkflowError())
}
```


cretz:
@Sikian - This looks good to me, thanks! Mark it as ready for review instead of draft and I'll gladly approve and merge it.

cretz (approved):

cretz:
Thanks! I approved. I'll leave it open for a bit to see if anyone else wants to look.

PR #749, opened 2022-03-09
Rename PayloadEncoder -> PayloadCodec.

robholland:
## What was changed

Renamed PayloadEncoder to PayloadCodec. Adjust the API to deal with slices of Payload rather than individual Payload or using a Payloads.

## Why?
This is to be more consistent with the naming and interfaces used in other SDKs.


converter/codec_test.go
```diff
@@ -138,47 +138,20 @@ func assertEncodingDataConverter(t *testing.T, data interface{}) {
...
-		converter.NewZlibEncoder(converter.ZlibEncoderOptions{}),
+		converter.NewZlibCodec(converter.ZlibCodecOptions{}),
 	)
 	require.NoError(t, err)
 	compUnderMinPayload, err := zlibIgnoreMinConv.ToPayload(data)
 	require.NoError(t, err)
 	require.True(t, proto.Equal(uncompPayload, compUnderMinPayload))
 }
 
-func TestEncodingDataConverterClone(t *testing.T) {
```
robholland, line 151: I have removed the cloning from `ToPayload(s)` because I believe the clone is only required for the `FromPayload(s)` cases where we must not mutate the `Payload` passed in. I believe it is safe to mutate `Payload` we receive from our parent `DataConverter` as those would not yet have been seen by any other code. Let me know if this is not correct.

converter/codec.go
```diff
@@ -39,151 +39,163 @@ import (
...
 // compression.
-type PayloadEncoder interface {
-	// Encode optionally encodes the given payload which is guaranteed to never
+// These can be used (and even chained) in NewCodecDataConverter.
+type PayloadCodec interface {
+	// Encode optionally encodes the given payloads which are guaranteed to never
 	// be nil. The byte slices of the payload's metadata or data should never be
 	// mutated directly, but they can be referenced or replaced.
-	Encode(*commonpb.Payload) error
+	Encode([]*commonpb.Payload) error
```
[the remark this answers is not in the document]
robholland: I'm a bit nervous about changing the payload count, but I agree that returning a slice for these APIs would make them easier to use safely.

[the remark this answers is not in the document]
robholland: Updated now.

converter/codec.go
```diff
@@ -0,0 +1,488 @@
...
+// NewZlibCodec creates a PayloadCodec for use in NewCodecDataConverter
+// to support zlib payload compression.
+//
+// While this serves as a reasonable example of a compression encoder, callers
+// may prefer alternative compression algorithms for lots of small payloads.
+func NewZlibCodec(options ZlibCodecOptions) PayloadCodec { return &zlibCodec{options} }
+
+func (z *zlibCodec) Encode(payloads []*commonpb.Payload) ([]*commonpb.Payload, error) {
+	result := make([]*commonpb.Payload, len(payloads))
+	for i, p := range payloads {
```
bergundy, line 79: We can now leverage the fact that this operates on multiple payloads and compress them all together.

robholland: That's not something I'm comfortable doing in this PR.

converter/codec.go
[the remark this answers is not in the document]
cretz: We don't need to change the payload count for our use. We might even should document that we require "at least one" payload response everytime.

converter/codec.go
cretz: That is a bit scary, but it could be done, sure. But I'm ok with not doing it here. Really this is an example and I'd be fine removing this codec altogether, heh.

cretz (approved):

PR #762, opened 2022-03-24
Add cloud test to CI

cretz:
## What was changed

Allow integration tests to use env-var-based cert/namespace and add cloud test to CI

## Why?

We need to ensure at least the most basic workflows work on cloud always


cretz:
Closing because you can't access secrets in a fork

PR #765, opened 2022-03-28
workflow.NewUUID helper

cretz:
## What was changed

Added `workflow.NewUUID`

## Why?

Requested

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -31,6 +31,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	failurepb "go.temporal.io/api/failure/v1"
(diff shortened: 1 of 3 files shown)


internal/workflow.go
```diff
@@ -1728,3 +1729,16 @@ func convertFromPBRetryPolicy(retryPolicy *commonpb.RetryPolicy) *RetryPolicy {
 func GetLastCompletionResultFromWorkflowInfo(info *WorkflowInfo) *commonpb.Payloads {
 	return info.lastCompletionResult
 }
+
+// NewUUID is a shortcut for:
+// 	workflow.SideEffect(ctx, func(workflow.Context) interface{} {
+// 		return uuid.NewString()
+// 	})
+func NewUUID(ctx Context) string {
```
bergundy, line 1737: Uuid in Java and ts use deterministic random

cretz: This was requested to use this approach.

And UUID libraries in Go use cryptographically secure randomization for UUIDs, so not doing this could give the false sense of security for how the UUIDs could be used. Node/browser, Python, Java, etc all use cryptographically secure randomization for UUIDv4 generation from what I can see.

In fact, I am unaware of any commonly used UUIDv4 generator that does not use secure randomization.

Anyone can create their own deterministic random seeded by the workflow start time or something if they want to opt-out of UUIDs safe for secure use. But I am not sure we should normalize such a thing. Not only is it very bad for collisions, it's very bad for any kind of commonly secure use (e.g. a password reset token).

bergundy: In TS we document it as being insecure:
```
Generate an RFC compliant V4 uuid.
Uses the workflow's deterministic PRNG making it safe for use within a workflow.
This function is cryptographically insecure.
```

Maybe we should rename to `insecureUUID4()`.

This is what Java has to say about it:

```
public static java.util.UUID randomUUID()
Replay safe way to generate UUID.
Must be used instead of UUID.randomUUID() which relies on a random generator, thus leads to non-deterministic code which is prohibited inside a workflow.
```

I'd be more worried about collisions than deterministic generation of the UUIDs.

bergundy: If we're just wrapping the go uuid function with a side-effect I don't think there's that much value in exposing it at all.

cretz: I'm starting to agree with that. I think this little helper is unneeded. I could see `NewRandom` as needed and then they could https://pkg.go.dev/github.com/google/uuid#NewRandomFromReader or something.

cretz:
I am closing this. We don't have a random safe for use `workflow` in Go and the docs at https://pkg.go.dev/go.temporal.io/sdk/workflow#hdr-SideEffect_API have long-since suggested to use side effect. This adds nothing of value.

PR #770, opened 2022-03-31
Fix histogram naming in prometheus naming scope for tally

naormatania:


naormatania:
> @naormatania - Thanks! I think the CLA needs to be signed before I merge this.

How do I sign it ? I can't find here a link where I can sign it

cretz (approved):

PR #771, opened 2022-04-11
SetOnActivityHeartbeatListener on TestActivityEnvironment

### LINKED ISSUE #768: Add TestActivityEnvironment.SetOnActivityHeartbeatListener
## Expected Behavior

Should be able to have heartbeat listener in activity environment same as in workflow environment

## Actual Behavior

Not able to set this

cretz:
## What was changed

Added `SetOnActivityHeartbeatListener` on `TestActivityEnvironment` same as `TestWorkflowEnvironment`

## Why?

Need to be able to mock heartbeat calls to server

## Checklist

1. Closes #768

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -199,6 +199,9 @@ type (
 
 		workerStopChannel  chan struct{}
 		sessionEnvironment *testSessionEnvironmentImpl
+
+		// True if this was created only for testing activities not workflows.
+		activityEnvOnly bool
 	}
 
 	testSessionEnvironmentImpl struct {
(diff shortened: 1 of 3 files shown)


bergundy (approved):
What's the added value of this callback over using activity interceptors?

cretz:
> What's the added value of this callback over using activity interceptors?

This is for the testing framework. We already have it for `TestWorkflowEnvironment` so it's just inconsistent not to have it for `TestActivityEnvironment`. That we could replace many of our mocks/callbacks with interceptors I think is a different discussion.

bergundy:
> > What's the added value of this callback over using activity interceptors?
> 
> This is for the testing framework. We already have it for `TestWorkflowEnvironment` so it's just inconsistent not to have it for `TestActivityEnvironment`. That we could replace many of our mocks/callbacks with interceptors I think is a different discussion.

Sure, just with interceptors you wouldn't have to add the comment below which of course will come as a surprise to many users.
I'd at least add a sentence stating that interceptors can be used to track each heartbeat call.

```
// Note: Due to internal caching by the activity system, this may not get called
// for every heartbeat recorded. This is only called when the heartbeat would be
// sent to the server (periodic batch and at the end only on failure).
```

cretz:
@bergundy - Added the comment. Agreed the behavior is confusing, but it is already in use that way.

PR #772, opened 2022-04-11
Allow data converters to be created without deadlock detection

### LINKED ISSUE #769: Move some data conversion out of the deadlock detection scope

Some advanced use cases of workflow data converters do not want to be bound to Temporal workflow deadlock rules

For as much as is reasonable, try to do the data conversion outside of the inline workflow and nearer to the commands/events. Even if I try to take a simple approach (e.g. shipping Go object + data converter around instead of payloads), I fear this is quite a daunting challenge. Also, I will have to confirm we don't expect data conversion errors to be able to be handled inline.

cretz:
## What was changed

Added `workflow.DataConverterWithoutDeadlockDetection` to create data converters without deadlock detection. Also reworked deadlock detection internally so it can be paused and resumed.

## Why?

Some data converters need to do advanced things

## Checklist

1. Closes #769


cretz:
Added simple tests for the converter outside of workflow

Issue #774, opened 2022-04-12
Possible memory leak when loading a workflow with a query


ovargasmahisoft:
## Expected Behavior

When querying a workflow, the worker should add the workflow to the internal cache.

## Actual Behavior

When the Worker starts and the first interaction with the workflow is a Query, the workflow is not cached, you can notice this behavior because the events are replayed from scratch every time.

The memory leak occurs because every time the query is handled, a new instance of the workflow is allocated by the worker. It only happens when the workflow stays awaiting new signals.

## Steps to Reproduce the Problem

Consider a workflow like

```go
package workflow

import (
	"go.temporal.io/sdk/workflow"
)

func ReproduceML(ctx workflow.Context, param string) error {
	log := workflow.GetLogger(ctx)
	log.Info("Starting new Workflow")
	defer func() {
		log.Info("Exit from workflow")
	}()
	state := param
	var receivedSignals []string

	err := workflow.SetQueryHandler(ctx, "getState", func() (string, error) {
		return state, nil
	})

	if err != nil {
		log.Error("SetQueryHandler getState failed", "Error", err)
		return err
	}

	channel := workflow.GetSignalChannel(ctx, "ml-signals")
	for {

		if state == "complete" {
			break
		}

		var signal string
		_ = channel.Receive(ctx, &signal)
		// Do something special with the signal and calculate the state
		state = signal

		// Forcing memory growing, so it can be noticed it in the metrics
		receivedSignals = append(receivedSignals, signal)
	}

	log.Info("Workflow finished")
	return nil
}

```

1. Start the worker
2. Create a workflow
3. Signal the workflow around 500 times to generate some data.
4. Restart the worker
5. ONLY query the workflow several times and notice how the memory grows.

Verifying:

- In the logs, you can notice an entry "Starting new Workflow" for every query and never the "Exit from workflow" meaning the workflow stayed allocated.
- If we disable the cache with `worker.SetStickyWorkflowCacheSize(0)`, the logs show "Starting new Workflow" and "Exit from workflow" with every query execution and the memory stops growing.
- With the cache enabled, If after the 4th step, the first interaction is a signal, the "Starting new Workflow" is shown only once, meaning the cache was properly set, and queries to that workflow will be returned faster and the memory won't grow as before.

Final note: with a workflow with a similar structure as the above, the memory increased to 6Gb+, and without modifying code after doing one of the last two bullet points the memory stays below 30 Mb.

## Specifications

- Version:
[...]

PR #779, opened 2022-04-14
Fix goroutine/memory leak on query-only poll responses

cretz:
## What was changed

Made sure that the workflow execution context state was cleared on non-cached workflows

## Why?

Workflows can start many coroutines which may hang waiting for server interaction. Currently we only clear that up on workflow completion or in the worker finalizer on cache eviction. But query-only poll responses (i.e. where the legacy `PollWorkflowTaskQueueResponse.query` is non-nil) do not cache. So they were not exiting the context which didn't clean up goroutines and other items.

Before this fix, you could query a started-but-not-finished workflow over and over and the memory would continually grow.

## Checklist

1. Relates to #774

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -416,7 +416,7 @@ func (w *workflowExecutionContextImpl) Lock() {
 	w.mutex.Lock()
 }
 
-func (w *workflowExecutionContextImpl) Unlock(err error) {
+func (w *workflowExecutionContextImpl) Unlock(contextCached bool, err error) {
 	if err != nil || w.err != nil || w.isWorkflowCompleted ||
 		(w.wth.cache.MaxWorkflowCacheSize() <= 0 && !w.hasPendingLocalActivityWork()) {
 		// TODO: in case of closed, it asumes the close command always succeed. need server side change to return
@@ -428,6 +428,10 @@ func (w *workflowExecutionContextImpl) Unlock(err error) {
 			// sticky is disabled, manually clear the workflow state.
 			w.clearState()
 		}
+	} else if !contextCached {
+		// Clear the state if we never cached the workflow so coroutines can be
+		// exited
+		w.clearState()
 	}
 
 	w.mutex.Unlock()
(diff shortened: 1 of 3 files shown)


internal/internal_task_handlers.go
```diff
@@ -635,13 +640,14 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
 
 		if wth.cache.MaxWorkflowCacheSize() > 0 && task.Query == nil {
 			workflowContext, _ = wth.cache.putWorkflowContext(runID, workflowContext)
+			cached = true
 		}
 		workflowContext.Lock()
 	}
```
Sushisource, line 647: Looks like maybe you could just add `cached := workflowContext != nil` here rather than needing to set it in two of the branches above?

Sushisource (approved):

internal/internal_task_handlers.go
cretz: GitHub does a bad job of indenting tab-based languages. Sometimes `workflowContext` is created without being cached. I only set `cached` as true if it was successfully retrieved or successfully put in cache, not if it just happens to be present.

Sushisource: Got it

internal/internal_task_handlers.go
```diff
@@ -703,13 +709,13 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 			tagPreviousStartedEventID, task.GetPreviousStartedEventId())
 	})
 
-	workflowContext, err := wth.getOrCreateWorkflowContext(task, workflowTask.historyIterator)
+	contextCached, workflowContext, err := wth.getOrCreateWorkflowContext(task, workflowTask.historyIterator)
```
yiminc, line 712: can we put isCached into the workflowContext itself so do don't need to know it here.

cretz: I think it's only needed for this area, I am hesitant to add yet more values into the context (values in context are basically untyped dynamic fields, whereas this param explicitly lets the reader/caller know that it is required to move through these calls).

yiminc: I mean a strong typed field in workflowExecutionContextImpl.

cretz: I see. The field would still have no value except as a return here and a param to the Unlock. There is no other use of the field, so it's seems strange to add it there. Is there a good reason to add it even though its only use will be a few lines below?

yiminc: I think that is some implementation detail that can be hide away and caller does not need to know about it.

internal/internal_task_handlers.go
```diff
@@ -635,13 +640,14 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
 
 		if wth.cache.MaxWorkflowCacheSize() > 0 && task.Query == nil {
 			workflowContext, _ = wth.cache.putWorkflowContext(runID, workflowContext)
+			cached = true
```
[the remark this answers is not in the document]
cretz: That makes sense. I didn't think of the case of multiple instances for a single run ID, but totally possible. Fixing.

cretz:
@yiminc - I moved `cached` to a field. Can you take another look and see if it looks right?

yiminc (approved):

Issue #778, opened 2022-04-13
Potential issue with activity mocking and cancellation


cretz:
A couple of issues have potentially been observed with the workflow/activity test env:

1. A mocked activity may not return a value if the context is cancelled
2. An activity called on the cancelled context for cleanup is immediately returning and still running the activity

Need standalone replication, then debugging/fix if applicable.

awsaba:
> 2. An activity called on the cancelled context for cleanup is immediately returning and still running the activity

Our use case is not so much clean up as multiple steps of bookeeping. Putting a `switch` after every return to distinguish between an `ActivityError` and cancellation seems like the obvious but wrong solution since it would have to added after most activities.

The workflows and test:
```go
package cancelwork

import (
	"context"
	"fmt"
	"time"

	"go.temporal.io/sdk/activity"
	"go.temporal.io/sdk/workflow"
)

// CancelMe is long-running activity that is the most likely to be running
// when a cancel is received
func CancelMe(ctx context.Context, _ string) (string, error) {
	time.Sleep(1 * time.Second)
	return "long-running-result", nil
}

// SaveStatus just logs the status, but could do some database operation
func SaveStatus(ctx context.Context, value string) error {
	logger := activity.GetLogger(ctx)
	time.Sleep(3 * time.Second)
	logger.Error(fmt.Sprintf("from SaveStatus, value is: \"%s\"", value))
	return nil
}

// SomeWorkflow calls a long running op to do something
func SomeWorkflow(ctx workflow.Context) error {
	logger := workflow.GetLogger(ctx)
	ao := workflow.ActivityOptions{
		StartToCloseTimeout:    10 * time.Minute,
		ScheduleToCloseTimeout: 24 * time.Hour,
		WaitForCancellation:    true,
	}
	ctx = workflow.WithActivityOptions(ctx, ao)

	defer func() {
		logger.Error("some other log before return")
	}()

	var val string
	err := workflow.ExecuteActivity(ctx, CancelMe, "").Get(ctx, &val)
	if err != nil {
		logger.Error(fmt.Sprintf("error in CancelMe: %s, is there a way to get val: \"%s\"", err.Error(), val))
		perr := workflow.ExecuteActivity(ctx, SaveStatus, val).Get(ctx, nil)
		if perr != nil {
			logger.Error(fmt.Sprintf("if cancelled, this should fail with cancel and not execute SaveStatus: \"%s\"", perr.Error()))
		}
		return err
	}

	return nil
}
```

[...]

PR #780, opened 2022-04-18
Fixes for activities in test suite

cretz:
## What was changed

* Fix issue where child workflow execution info was not propagated into activity info in test suite
* Fix issue where activity heartbeats were not handling cancellation in test suite
* Fix issue where activity was executing even if `ExecuteActivity` was called with cancelled context in test suite
* Suppress large output on `workflowcheck` test (unrelated to rest of issue)

## Why?

Multiple issues with how activities interact with cancellation and child workflows. The first of which was discovered as part of #778.

While our unit tests pass, this technically changes (fixes) some things about activity info and heartbeats that could break existing tests that rely on the broken behavior. Also, this relies on our tests having decent coverage of our own test suite.

### DIFF
--- a/contrib/tools/workflowcheck/determinism/determinism_test.go
+++ b/contrib/tools/workflowcheck/determinism/determinism_test.go
@@ -43,7 +43,7 @@ func Test(t *testing.T) {
 		determinism.NewChecker(determinism.Config{
 			IdentRefs:         identRefs,
 			SkipFiles:         []*regexp.Regexp{regexp.MustCompile(`.*/should_skip\.go`)},
-			Debug:             true,
+			Debug:             false, // Set to true to see details
 			DebugfFunc:        t.Logf,
 			EnableObjectFacts: true,
 		}).NewAnalyzer(),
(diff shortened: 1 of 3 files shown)


contrib/tools/workflowcheck/determinism/determinism_test.go
```diff
@@ -43,7 +43,7 @@ func Test(t *testing.T) {
 		determinism.NewChecker(determinism.Config{
 			IdentRefs:         identRefs,
 			SkipFiles:         []*regexp.Regexp{regexp.MustCompile(`.*/should_skip\.go`)},
-			Debug:             true,
+			Debug:             false, // Set to true to see details
```
cretz, line 46: This is unrelated to the rest of the issue, but is too noisy to see other test output

internal/internal_workflow_testsuite.go
```diff
@@ -312,14 +312,15 @@ func newTestWorkflowEnvironmentImpl(s *WorkflowTestSuite, parentRegistry *regist
 		return nil
 	}
 
-	em := mockService.EXPECT().RecordActivityTaskHeartbeat(gomock.Any(), gomock.Any(), gomock.Any()).
-		Return(&workflowservice.RecordActivityTaskHeartbeatResponse{CancelRequested: false}, nil)
-	em.Do(func(ctx context.Context, r *workflowservice.RecordActivityTaskHeartbeatRequest, opts ...grpc.CallOption) {
-		// TODO: The following will hit a data race in the gomock code where the Do() action is executed outside
```
cretz, line 318: I could not find any information on what this warning relates to. The tests were succeeding for me and I have uncommented this. We need the heartbeat result to know whether the activity is cancelled. I will keep an eye out to see if what this is referring to resurfaces.

yiminc (approved):

Sushisource (approved):

PR #785, opened 2022-04-20
Add task queue to worker metrics

### LINKED ISSUE #784: Task queue not recorded on temporal_worker_task_slots_available metric
## Expected Behavior

Expect worker task queue to be on `temporal_worker_task_slots_available` but it is not

## Actual Behavior

Task queue is not set

cretz:
## What was changed

Add `task_queue` label/tag to worker metrics, specifically `temporal_poller_start`, `temporal_worker_start`, and `temporal_worker_task_slots_available`.

## Why?

https://docs.temporal.io/docs/reference/sdk-metrics shows them available from other languages. Also, with gauges like `temporal_worker_task_slots_available` which are set to a fixed value on worker start, not separating per task queue means they can overwrite each other.

⚠️ Is this change acceptable? This will add cardinality to these 3 metrics.

Also, is it ok that `temporal_worker_task_slots_available` doesn't include some kind of worker identity? That means multiple workers on the same task queue sharing the same metric scope/collector will be overwriting each others' value for that metric since it's a fixed value update. I assume that multiple workers for the same task queue sharing the same metric scope is rare.

## Checklist

1. Closes #784


Sushisource (approved):

Sushisource:
>  Is this change acceptable? This will add cardinality to these 3 metrics.

Low cardinality, I don't think it's an issue.

>Also, is it ok that temporal_worker_task_slots_available doesn't include some kind of worker identity? That means multiple workers on the same task queue sharing the same metric scope/collector will be overwriting each others' value for that metric since it's a fixed value update. I assume that multiple workers for the same task queue sharing the same metric scope is rare.

It would be super weird to have multiple workers on the same TQ in the same process. No reason to be concerned about that IMO. (Honestly, if we want, we could throw errors in that situation. It just makes no sense.)

PR #791, opened 2022-04-25
Support disabling run following for workflow results

### LINKED ISSUE #789: Allow getting workflow result to not follow runs

People want to see results for that execution, but if `WorkflowExecutionCompletedEventAttributes.new_execution_run_id` is set, we automatically follow to the next one. This sucks for things like wanting to get an older cron result.

Allow users to opt-out of "follow runs" like we do in other languages. It is still up in the air on how best to provide that option to the `client.WorkflowRun.Get` call without breaking backwards compat (new call adds method to the iface, context value is ugly but may be best here).

cretz:
## What was changed

* Add `WorkflowRun.GetWithOptions` that supports an option to disable following runs
* Return continue as new error if follow runs disabled for continue as new

## Why?

Currently we always follow to the last run in the "chain". Some people don't want this, especially for crno (see #789).

## Checklist

1. Closes #789


internal/internal_workflow_client.go
```diff
@@ -91,21 +91,38 @@ type (
...
+	}
+
+	// WorkflowRunGetOptions are options for WorkflowRun.GetWithOptions.
+	WorkflowRunGetOptions struct {
+		// DisableFollowingRuns, if true, will not follow execution chains to the
+		// latest run. By default when this is false, getting the result of a
+		// workflow may not use the literal run ID but instead follow to later runs
+		// if the workflow returned a ContinueAsNewError, has a later cron, or is
+		// retried on failure.
+		DisableFollowingRuns bool
```
cretz, line 125: Should I fail if this was set to true but a run ID was not provided? Is there a use case for not providing a run ID but still not following to the next run?

bergundy: We don't in TS. I can't think of a reason not to allow it, saves looking up the last run ID for a workflow ID.

bergundy: Can we change this setting name to `FollowRuns` and still default to `true`? It's easier to follow.

bergundy (approved):
Took a brief look and left a comment inline for your consideration.

internal/internal_workflow_client.go
cretz: > Can we change this setting name to FollowRuns and still default to true? It's easier to follow.

I am afraid not in Go. You can't default fields, their default is always the zero value which for booleans is false. This is why for all Go structs, you specifically build the field where false is the reasonable default.

bergundy: Yeah, I get it. I thought that we could check if the options struct is null and default to true but we want empty struct to also default to true.

cretz: This options struct is never null inside the outer options (it's not a pointer). And even if it could be nullable, we want to support people creating it and just setting the fields they care about instead of trying to differentiate between unset and explicit false.

bergundy: Cool, I agree this is the best approach

PR #795, opened 2022-04-28
Deprecate NewClient, add Dial and NewLazyClient and CheckHealth

### LINKED ISSUE #595: Expose health check on WorkflowClient
Temporal SDK should expose a method like WorkflowServiceStubs#healthCheck or WorkflowServiceStubs#isHealthy that allows users to check if the WorkflowClient/WorkflowServiceStubs underlying channel / connection to Temporal server is healthy.
There were quite a bit of user requests for such a method from users for all SDKs.
Having this method on Stubs is preferable over WorkflowClient, because it's the lowest level abstraction on top of the channel and WorkflowClient can just fall back on the WorkflowServiceStubs method. But if the language doesn't have a wrapper for stubs (like Java does), this method may be on WorkflowClient.

Issues for other SDKs:
https://github.com/temporalio/sdk-java/issues/828
https://github.com/temporalio/sdk-node/issues/312

### LINKED ISSUE #753: Allow lazy connection on client
The ability to lazy connect to the server was removed in v1.14.0 with the addition of the getCapabilities call on client creation.

Ideally the SDK would still offer the option to connect lazily. Fail fast in a critical application with multiple concerns (i.e. Temporal access plus at least one other concern) is a non-starter. It can block and bring an entire service down during a deploy, scale up, or other lifecycle event even if only a small portion of the service is not operational.

Each call on the client can check whether or not the connection has been created, and use a mutex to lock and create it before allowing any calls to proceed.

One option would be to duplicate the entire interface in our application to wrap every call with the solution described above. From a maintenance perspective it's not great. Would prefer to have this option in the upstream client.

A client factory can help, but the factory has side-effects in that case. This is an anti-pattern, so additional warnings are then needed to prevent a developer from calling the factory somewhere it could block (e.g. in the instantiation of a module during dependency injection or similar). To better mitigate the problem an application can implement a factory that creates the config and then an additional `GetConnection()` call off of the returned struct. This signals to the developer that the call will connect to the server and should not be used during application start.

**Additional context**
See this thread for more discussion: https://temporalio.slack.com/archives/CTRCR8RBP/p1645050691089599

### LINKED ISSUE #793: Deprecate NewClient, create Dial, NewLazyClient, and CheckHealth calls and support laziness

Today we have `NewClient` which, possibly unexpectedly to users, does more than create a client, it also connects to the server to confirm all connectivity. We previously had the ability to disable this but we removed that in #706.

* (re)Add `client.Options.ConnectionOptions.DisableEagerConnection bool` that defaults to false for `NewClient`
* Add `client.Dial` call that force-sets `client.Options.ConnectionOptions.DisableEagerConnection` to `false` (so basically it's the same as `NewClient` today)
* Add `client.NewLazyClient` call that force-sets `client.Options.ConnectionOptions.DisableEagerConnection` to `true`
* Deprecate `NewClient` in favor of the `Dial` and `NewLazyClient` calls
* Add a client `ensureServerCapabilitiesFetched` that atomically lazily obtains the server capabilities if not fetched already and are memoized on success
* Update all client calls to have `ensureServerCapabilitiesFetched` call to support lazy connectivity
* Add a `CheckHealth` call on the client that uses the gRPC health API (note, it too will lazily fetch capabilities)

cretz:
## What was changed

Mostly what's at #793 w/ slight changes.

* Add `client.Dial` which eagerly connects like `NewClient` does
* Add `client.NewLazyClient` which doesn't connect until you make a call
* Deprecate `NewClient`
* Make sure server capabilities are lazily fetched before each high-level call and before worker start if they are not already fetched
* Add `client.Client.CheckHealth` call

## Why?

People want lazy-connecting clients but we still need capabilities checks. People also wanted explicit health checking.

## Checklist

1. Closes #793
2. Closes #753
3. Closes #595


internal/client.go
```diff
@@ -467,6 +466,12 @@ type (
 		// grpc.WithUnaryInterceptor option can be added since grpc.WithUnaryInterceptor is prepended to chains set with
 		// grpc.WithChainUnaryInterceptor.
 		DialOptions []grpc.DialOption
+
+		// Hidden for use by client overloads.
+		disableEagerConnection bool
+
+		// If not present, created during service client creation.
```
Spikhalskiy, line 473: May you add more javadoc about what it actually defines, not just how it's initialized? The name is not descriptive enough for me.

Spikhalskiy (approved):

internal/client.go
cretz: Sure. This name has already been in use, I just re-added it here and this is internal btw, so the customer doesn't see it. But I will add documentation on what it is.

internal/internal_workflow_client.go
```diff
@@ -190,6 +202,9 @@ func (wc *WorkflowClient) ExecuteWorkflow(ctx context.Context, options StartWork
 // The current timeout resolution implementation is in seconds and uses math.Ceil(d.Seconds()) as the duration. But is
 // subjected to change in the future.
 func (wc *WorkflowClient) GetWorkflow(ctx context.Context, workflowID string, runID string) WorkflowRun {
+	// We intentionally don't "ensureIntialized" here because there is no direct
+	// error return path. Rather we let GetWorkflowHistory do it.
+
```
alexshtin, line 207: I'd avoid words "we" and "I" in comments. Something more "dry" is better here:
```suggestion
	// "ensureIntialized" is not called here intentionally because there is no direct
	// error return path. GetWorkflowHistory will do it.

```

client/client.go
```diff
@@ -431,9 +441,24 @@ type MetricsTimer = metrics.Timer
 // MetricsNopHandler is a noop handler that does nothing with the metrics.
 var MetricsNopHandler = metrics.NopHandler
 
+// Dial creates an instance of a workflow client. This will attempt to connect
+// to the server eagerly and will return an error if the server is not
+// available.
+func Dial(options Options) (Client, error) {
+	return internal.DialClient(options)
```
alexshtin, line 448:
```suggestion
	return internal.Dial(options)
```
?

alexshtin (approved):

client/client.go
cretz: Just `Dial` in a `client` package makes sense, but just `Dial` in an internal package where you could be dialing other things does not. `worker.New` -> `internal.NewWorker`, `client.Options` -> `internal.ClientOptions`, etc, etc.

client/client.go
```diff
@@ -431,9 +441,24 @@ type MetricsTimer = metrics.Timer
 // MetricsNopHandler is a noop handler that does nothing with the metrics.
 var MetricsNopHandler = metrics.NopHandler
 
+// Dial creates an instance of a workflow client. This will attempt to connect
+// to the server eagerly and will return an error if the server is not
+// available.
+func Dial(options Options) (Client, error) {
```
mmcshane, line 447: Curous if you considered taking this opportunity to move to function-style optional args?

mmcshane (approved):
I'm okay with it because the functionality is there but I had different expectations.

I had understood that all SDKs would move to become lazy by default with an option to be eager and thus I expected Dial to be lazy and NewClient to remain eager (to be kind to existing callers who might expect certain behavior). Not sure that NewLazyClient pulls its own weight in terms of utility versus adding to the API surface area.

client/client.go
cretz: For most of the Go SDK, the struct options is good enough to still give us future proofing. There's little benefit of a `func Dial(options ...DialOption) (Client, error)` over struct-based if you 1) have no other params, 2) don't need to preserve backwards compatibility, and 3) can reasonably copy the struct.

I could be leaving off an important benefit, so if there is a good enough argument I'd definitely reconsider.

PR #796, opened 2022-04-28
Update max gRPC message size to 128 MB

### LINKED ISSUE #794: Change default gRPC max message size to 128MB
Matches other SDKs

cretz:
## What was changed

Changed default max gRPC message size to 128 MB from 64 MB.

## Why?

To align with other SDKs and the server

## Checklist

1. Closes #794


Spikhalskiy (approved):

internal/client.go
```diff
@@ -455,7 +455,7 @@ type (
 		// keepalive pings will be sent.
 		KeepAlivePermitWithoutStream bool
 
-		// MaxPayloadSize is a number of bytes that gRPC would allow to travel to and from server. Defaults to 64 MB.
+		// MaxPayloadSize is a number of bytes that gRPC would allow to travel to and from server. Defaults to 128 MB.
```
Spikhalskiy, line 458: Do we want to deprecate the setter?

cretz: I had considered it, but I was having trouble justifying it. Are we hurt by this option being available and/or is it really important to us to remove the option? I definitely wouldn't add it if it wasn't already there, but if I just slap a "deprecated - please don't" it doesn't really do anything for us unless we're actually having trouble with it.

PR #799, opened 2022-04-29
Use NamespaceNotFound error

alexshtin:
## What was changed

`NamespaceNotFound` is used instead of general `NotFound` in case of missing namespace.

See https://github.com/temporalio/temporal/pull/2785 and https://github.com/temporalio/api/pull/179 for details.

## Why?

To react on `NamespaceNotFound` differently.


cretz (approved):

bergundy (commented):
@alexshtin In the future, this sort of change should go through an issue in the sdk-features repo to make sure this is handled in all of our SDKs.

PR #801, opened 2022-05-05
Handle FAILED_CAUSE_NAMESPACE_NOT_FOUND

alexshtin:
## What was changed

Handle `FAILED_CAUSE_NAMESPACE_NOT_FOUND`.

## Why?

New failed cause was added on server. SDK needs to handle it properly.
See https://github.com/temporalio/temporal/pull/2785 for details.


bergundy (approved):
LGTM

internal/error.go
```diff
@@ -222,6 +222,9 @@ type (
 	// already started.
 	ChildWorkflowExecutionAlreadyStartedError struct{}
 
+	// NamespaceNotFoundError is set as the cause when failure is due namespace not found.
+	NamespaceNotFoundError struct{}
```
cretz, line 226: Could consider reusing `go.temporal.io/api/serviceerror.NamespaceNotFound` instead of adding this, but I don't mind this either (and maybe you don't have the namespace to put on the error at error time).

cretz (approved):

internal/error.go
alexshtin: I wouldn't mix these too and prefer to follow "the pattern".

PR #807, opened 2022-05-13
Add update-go-sum target

dnr:
## What was changed
Add `make update-go-sum` to propagate go.sum changes into other modules. It just does `go get ./...` in each module.

Also have the for-each-module loops in the Makefile echo the directory they're running in, which would have saved me a few minutes of confusion.

## Why?
developer experience

## Checklist

1. Closes 

2. How was this tested:
manually

3. Any docs updates needed?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,19 +76,24 @@ cover_ci: $(COVER_ROOT)/cover.out
 
 vet: $(ALL_SRC)
 	@for dir in $(MOD_DIRS); do \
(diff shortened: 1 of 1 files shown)


Sushisource (approved):
Thanks!

cretz (approved):
Looks good!

(I personally avoid Makefiles like the plague as they are a poor platform-dependent alternative for a real language. But the Makefile decision predates me.)

PR #809, opened 2022-05-17
Make mutable side effects work during replay

### LINKED ISSUE #803: MutableSideEffect doesn't seem to work
## Expected Behavior
MutableSideEffect works as described in the docs.

## Actual Behavior
This code:
```
	// load default number
	get := func(ctx workflow.Context) interface{} { return defaultNumber }
	eq := func(a, b interface{}) bool { return a.(int) == b.(int) }
	var number int
	if err := workflow.MutableSideEffect(ctx, "defaultNumber", get, eq).Get(&number); err != nil {
		panic("can't decode number:" + err.Error())
	}
```
works at first, but if I kill and restart the worker to force replay, I get:
```
2022/05/09 18:16:00 ERROR Workflow panic Namespace default TaskQueue hello-world WorkerID 2450384@pinglin@ WorkflowType Workflow WorkflowID hello_world_workflowID RunID 408db9e3-7885-4ca5-be59-0bce212c08eb Attempt 1 Error can't decode number:payload item 0: unable to decode: json: cannot unmarshal string into Go value of type int StackTrace coroutine root [panic]:
github.com/temporalio/samples-go/helloworld.Workflow({0xdfce40, 0xc00043a580}, {0xc0004360f8, 0x8})
        /home/dnr/t/samples-go/helloworld/helloworld.go:21 +0x23d
reflect.Value.call({0xbe9b60?, 0xd38768?, 0x30?}, {0xcd73c3, 0x4}, {0xc00040d1d0, 0x2, 0x18?})
        /nix/store/f2n8g9nbrxaqwyd9b3grynvwhyz3qlrw-go-1.18/share/go/src/reflect/value.go:556 +0x845
reflect.Value.Call({0xbe9b60?, 0xd38768?, 0xc000407d28?}, {0xc00040d1d0, 0x2, 0x2})
        /nix/store/f2n8g9nbrxaqwyd9b3grynvwhyz3qlrw-go-1.18/share/go/src/reflect/value.go:339 +0xbf
go.temporal.io/sdk/internal.(*workflowEnvironmentInterceptor).ExecuteWorkflow(0xc00042a730?, {0xdfce40?, 0xc00043a580?}, {0xc00042e490?, 0xc0005a86c0?}, {0xc00042e4a0?, 0x1?, 0xc0001030c8?})
        /home/dnr/t/go/pkg/mod/go.temporal.io/sdk@v1.6.0/internal/workflow.go:393 +0x1e5
go.temporal.io/sdk/internal.(*workflowExecutor).Execute(0xc00043a500, {0xdfce40, 0xc00043a580}, 0x25?)
        /home/dnr/t/go/pkg/mod/go.temporal.io/sdk@v1.6.0/internal/internal_worker.go:762 +0x2c3
go.temporal.io/sdk/internal.(*syncWorkflowDefinition).Execute.func1({0xdfcfc8, 0xc00040d170})
        /home/dnr/t/go/pkg/mod/go.temporal.io/sdk@v1.6.0/internal/internal_workflow.go:494 +0xcd

```

Another weird thing I noticed was what was encoded in the marker:

```
{
  "data": {
    "payloads": [
      {
        "metadata": {"encoding": "anNvbi9wbGFpbg=="},
        "data": "ImRlZmF1bHROdW1iZXIi"
      },
      {
        "metadata": {"encoding": "anNvbi9wcm90b2J1Zg=="},
        "data": "eyJwYXlsb2FkcyI6W3sibWV0YWRhdGEiOnsiZW5jb2RpbmciOiJhbk52Ymk5d2JHRnBiZz09In0sImRhdGEiOiJNVGs9In1dfQ=="
      }
    ]
  },
  "side-effect-id": {
    "payloads": [
      {
        "metadata": {"encoding": "anNvbi9wbGFpbg=="},
        "data": "ImRlZmF1bHROdW1iZXIi"
      }
    ]
  }
}
```
Decoding all of that base64 to strings just to make things clear (pseudo-json):
```
{
  "data": {
    "payloads": [
      {
        "metadata": {"encoding": "json/plain"},
        "data": `"defaultNumber"`
      },
      {
        "metadata": {"encoding": "json/protobuf"},
        "data": `{"payloads":[{"metadata":{"encoding":"anNvbi9wbGFpbg=="},"data":"MTk="}]}`
      }
    ]
  },
  "side-effect-id": {
    "payloads": [
      {
        "metadata": {"encoding": "json/plain"},
        "data": `"defaultNumber"`
      }
    ]
  }
}
```
That is, the actual data is double-base64-encoded.

Repro here: https://github.com/dnr/samples-go/tree/mutablesideeffect/helloworld

1. Run the worker
2. Run the starter
3. Kill the worker
4. Start the worker again

## Specifications

  - Version: tested go sdk 1.6.0 and also 1.14.0. server is a branch somewhere near 1.16.0. go 1.18
  - Platform: linux

cretz:
## What was changed

* Fixed mutable side effect marker value extraction during replay event
* Fixed mutable side effect marker command generation during replay

## Why?

Mutable side effects have been broken during replay in two ways (one of which is since their inception), see #803.

* Mutable side effect marker format was changed in #151, but the part that reads that out during replay wasn't and no tests were there to properly test it
* Replay wasn't generating the same commands as non-replay causing command counter mismatch

Question: What if two mutable side effects of the same ID are run during the same WFT? We don't differentiate by command, only by user-defined marker ID.

## Checklist

1. Closes #803


internal/internal_task_handlers.go
```diff
@@ -860,6 +859,8 @@ ProcessEvents:
 		} else {
 			w.workflowInfo.BinaryChecksum = binaryChecksum
 		}
+		// Reset the mutable side effect markers recorded
+		eventHandler.mutableSideEffectsRecorded = nil
```
mmcshane, line 863: Not sure I understand this one - guessing the eventHandler gets reused and this is why we need to lazy-instantiate the map above?

mmcshane (commented):
Fine at a code level - just a _why_ question

internal/internal_event_handlers.go
```diff
@@ -1132,17 +1141,37 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
...
+				err = weh.dataConverter.FromPayloads(sideEffectDataPayload, &sideEffectDataID, &sideEffectDataContents)
+			}
+			if err == nil && sideEffectID != sideEffectDataID {
+				err = fmt.Errorf("mutable side effect ID does not match embedded ID")
+			}
+			if err == nil {
+				weh.mutableSideEffect[sideEffectID] = &sideEffectDataContents
+				// We must mark that it is recorded so we can know whether a command
+				// needs to be generated during replay
+				if weh.mutableSideEffectsRecorded == nil {
```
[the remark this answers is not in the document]
cretz: Why? It's rarely used and only in this spot non-concurrently. Why instantiate every time if I only need to mutate in one place?

internal/internal_task_handlers.go
```diff
@@ -289,7 +289,6 @@ func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, marker
 	}
 	return result, markers, checksum, err
 }
-
 func (eh *history) hasMoreEvents() bool {
```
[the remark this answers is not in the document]
cretz: That is strange to me too. I have confirmed that `go fmt` does indeed leave this alone. Regardless, this is an inadvertent blank line removal and I have put it back.

internal/internal_task_handlers.go
cretz: Correct. Specifically, the event handler is mostly just a wrapper around the embedded `*workflowEnvironmentImpl` which is for the life of the locally cached workflow. But the specific pieces I am recording here are only for the life of the task.

mmcshane: Gotcha. Feels like there might be a piece missing to represent that scope but that wouldn't be for this PR.

mmcshane (approved):

internal/internal_task_handlers.go
cretz: :+1: There definitely are missing pieces all throughout the internals. It is quite a hodgepodge of structs referencing structs. With issues such as these, I try to limit the surface area of my alterations compared to a refactoring PR.

internal/internal_event_handlers.go
```diff
@@ -690,6 +694,11 @@ func (wc *workflowEnvironmentImpl) MutableSideEffect(id string, f func() interfa
 	if result, ok := wc.mutableSideEffect[id]; ok {
 		encodedResult := newEncodedValue(result, wc.GetDataConverter())
 		if wc.isReplay {
+			// During replay, we only generate a command if there was a known marker
+			// recorded on the next task
+			if wc.mutableSideEffectsRecorded[id] {
```
yiminc, line 699: should we clear that flag here? we probably only want to generate the command once, even if there are 2 workflow.MutableSideEffect() calls with same id in this workflow task.

cretz: I don't think it helps. This relates to a concern I brought up in Slack an in the description:

> Question: What if two mutable side effects of the same ID are run during the same WFT? We don't differentiate by command, only by user-defined marker ID.

The issue is that if you ran 20 of the exact same mutable side effect calls (say with timer creation in between each) in a single WFT, it would result in only one command and the other 19 are no-ops. If I cleared this each time, it'd result in 19 on replay if I understand correctly. There is no way to get replay of multiple mutable side effects w/ the same ID right since the only way you know whether a command should be generated is either via look ahead (we don't differentiate one mutable marker from another for the same ID) or actually running the callback.

I may be misunderstanding though.

yiminc (approved):

PR #811, opened 2022-05-23
CI Updates 

cretz:
## What was changed

* Remove BuildKite (we have reached full parity in GH actions)
* Add coveralls to GH actions
* Remove a couple other seemingly unnecessary files

## Why?

* GH actions gives us more control and is clearer to our open source users
* Other SDKs are using GH actions too


.travis.yml
```diff
@@ -1,36 +0,0 @@
-dist: trusty
```
alexshtin, line 1: How did this file manage to survive that long?

alexshtin (approved):

alexshtin:
You are in the loop now. You can't merge it until you remove it.

.travis.yml
cretz: :shrug: Sometimes I find remnants of the fork here and there.

PR #812, opened 2022-05-23
Fix test/go.mod

cretz:
## What was changed

* Ran `go mod tidy` in all modules

## Why?

* `go mod tidy` not run in `/test` which is causing Go 1.16 builds to fail since it was missing gRPC dep


cretz:
This was solved in #809

PR #830, opened 2022-06-15
Encode nil query result

alexshtin:
## What was changed

If query handler returns `nil` it should be encoded as regular value. `nil` `Payloads` doesn't pass server validation.

## Why?

Fix bug. Other SDKs already doing the right thing.


cretz:
@alexshtin - thanks! I noticed you also changed the name of a couple of pretty important files here. Was that by intention? While the names may make sense, I am not sure there is any real value in changing them here.

cretz:
Yeah, the file move doesn't seem related.

cretz (approved):

PR #831, opened 2022-06-16
Update API dep & expose worker versioning API

Sushisource:
## What was changed
Added client methods for new APIs

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource:
Note: Not publicly exposing the APIs on client yet. That'll happen later.

go.mod
```diff
@@ -16,8 +15,8 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/objx v0.3.0 // indirect
-	github.com/stretchr/testify v1.7.1
-	go.temporal.io/api v1.8.0
+	github.com/stretchr/testify v1.7.2
```
cretz, line 18: Usually when updating, we prefer a specific `go get go.temporal.io/api@v1.9.0` so it doesn't affect other deps. But this is fine (also that `make update-go-sum` target was unfortunately created using go get instead of go mod tidy).

cretz (approved):

PR #836, opened 2022-06-27
Eager activity execution

cretz:
## What was changed

Added eager activity execution. Specifically:

* Add `worker.Options.DisableEagerActivities` to completely opt-out of eager activity execution (enabled by default)
* Add `worker.Options.MaxConcurrentEagerActivityExecutionSize` to maintain a maximum number of eager activities (unlimited by default, but still bound by `MaxConcurrentActivityExecutionSize` of course)
* Add `workflow.ActivityOptions.DisableEagerExecution` to opt-out of eager activity execution for a specific activity
* Unit tests to confirm counting is done properly
* Docker compose override to enable eager activities in CI

## Why?

Support upcoming eager activity feature in server. SDK core already did at https://github.com/temporalio/sdk-core/pull/296


internal/internal_eager_activity.go
```diff
@@ -0,0 +1,157 @@
...
+func (e *eagerActivityExecutor) handleResponse(
+	resp *workflowservice.RespondWorkflowTaskCompletedResponse,
+	amountReserved int,
+) {
+	// Ignore disabled or none present
+	if e == nil || e.activityWorker == nil || e.disabled || (len(resp.ActivityTasks) == 0 && amountReserved == 0) {
+		return
+	} else if len(resp.ActivityTasks) > amountReserved {
+		panic(fmt.Sprintf("Unexpectedly received %v eager activities though we only requested %v",
+			len(resp.ActivityTasks), amountReserved))
```
Sushisource, line 111: I don't think panic helps much here. Definitely a server bug, and worth printing an error, but in Core I print the error and then drop them on the floor. They'll run into their timeout eventually. This will result in the same thing, but also the worker crashes.

Sushisource: Should probably also print errors if the response is trying to eagerly execute stuff but any of this part of the clause is true:

`e == nil || e.activityWorker == nil || e.disabled`

internal/internal_eager_activity.go
```diff
@@ -0,0 +1,157 @@
...
+		// Like other parts that push onto this channel, we assume there is room
+		// because we took it, so we do a blocking send
+		e.activityWorker.worker.pollerRequestCh <- struct{}{}
+	}
+
+	// Start each activity asynchronously
+	for _, activity := range resp.ActivityTasks {
+		// Before starting the goroutine we have to increase the wait group counter
+		// that the poller would have otherwise increased
+		e.activityWorker.worker.stopWG.Add(1)
```
Sushisource, line 135: IMO encapsulate this behind a function in activity worker

test/integration_test.go
```diff
@@ -266,7 +266,8 @@ func (ts *IntegrationTestSuite) TestBasic() {
 	ts.assertMetricCount("temporal_request", 1, "operation", "StartWorkflowExecution")
 	ts.assertMetricCountAtLeast("temporal_request", 1, "operation", "RespondWorkflowTaskCompleted")
 	ts.assertMetricCountAtLeast("temporal_workflow_task_queue_poll_succeed", 1)
-	ts.assertMetricCountAtLeast("temporal_long_request", 3, "operation", "PollActivityTaskQueue")
+	// We cannot check PollActivityTaskQueue metric because eager activities
+	// affect poll count
```
Sushisource, line 270: I think that's gotta get fixed - it makes the metric not representative of what it says on the tin

internal/internal_eager_activity.go
bergundy: I'd actually argue that it's better to panic and that we make sure we run integration tests with pre-release server versions to avoid server bugs popping up.

cretz: Yeah, this is basically an assertion that server doesn't do something we didn't ask it too. I think it's reasonable to fail like a never-should-fail assertion.

test/integration_test.go
cretz: The metric is representative though. We have a metric for how many activity polls we call on the server and we asserted on that count in this test. We no longer do that assertion because we don't call poll as many times anymore w/ eager activities.

internal/activity.go
```diff
@@ -131,6 +131,11 @@ type (
 		// To disable retries set MaximumAttempts to 1.
 		// The default RetryPolicy provided by the server can be overridden by the dynamic config.
 		RetryPolicy *RetryPolicy
+
+		// If true, will not request eager execution regardless of worker settings.
+		// If false, eager execution may still be disabled at the worker level or
+		// eager execution may not be requested due to lack of available slots.
```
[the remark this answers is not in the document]
cretz: I can add it. We don't usually define terms above options that affect them (e.g. we don't say what a workflow task is on workflow task timeout, or what a heartbeat is on heartbeat timeout), but meh.

internal/internal_eager_activity.go
Sushisource: Panic only during test would make a lot of sense too, I do that in a handful of places

cretz: Whether you are testing is not a knowable thing in Go (they don't have an `assert` like many other languages). Maybe I'll just log and ignore and give the slots back.

internal/internal_eager_activity.go
cretz: Agreed line count matters little (most things can be one line if you try hard enough), it's just that this is internal code with internal knowledge of its siblings, there's little value in abstraction. I'll go ahead and comment in the other spot where the `Add` is done on poll response that this is done in eager activity also.

internal/internal_eager_activity.go
```diff
@@ -0,0 +1,157 @@
...
+	// Go over every command checking for activities that can be eagerly executed
+	for _, command := range req.Commands {
+		if attrs := command.GetScheduleActivityTaskCommandAttributes(); attrs != nil {
+			// If not present, disabled, or on a different task queue, we must mark as
+			// explicitly disabled
+			if e == nil || e.disabled || e.activityWorker == nil || e.taskQueue != attrs.TaskQueue.GetName() {
+				attrs.RequestEagerExecution = false
+			} else if attrs.RequestEagerExecution {
+				// If it has been requested, attempt to reserve one pending
+				attrs.RequestEagerExecution = e.reserveOnePendingSlot()
```
bergundy, line 69: nit: could count the number requested and reserve all at once to reduce the number of times you need to lock

internal/internal_eager_activity.go
bergundy: I vote for keeping the panic just so we have visibility for bad server behavior

internal/internal_eager_activity.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	// We can request, so increase the pending count
+	e.pendingCount++
+	return true
+}
+
+func (e *eagerActivityExecutor) handleResponse(
+	resp *workflowservice.RespondWorkflowTaskCompletedResponse,
+	amountReserved int,
```
bergundy, line 104: nit: I'd call this `amountActivitySlotsReserved` to clarify

internal/internal_eager_activity.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+// eagerActivityExecutor is a worker-scoped executor for eager activities that
+// are returned from workflow task completion responses.
+type eagerActivityExecutor struct {
+	eagerActivityExecutorOptions
+
+	activityWorker *activityWorker
+	pendingCount   int // Only access under countLock
+	executingCount int // Only access under countLock
```
bergundy, line 39: Why is it important to make a distinction between pending and executing?
Doesn't it complicate the code unnecessarily?
Unless I'm missing something, you could replace these 2 ints and the lock with an atomic int.

internal/internal_eager_activity.go
cretz: Yes, but then that'd require a bit more complication because it'd have to return how many it actually could reserve (possible it could only reserve half the activities for example). So basically it'd require multiple times through the command loop, but I may do that.

.github/workflows/docker/dynamic-config-custom.yaml
```diff
@@ -0,0 +1,5 @@
+system.forceSearchAttributesCacheRefreshOnRead:
+  - value: true # Dev setup only. Please don't turn this on in production.
+    constraints: {}
+system.enableActivityLocalDispatch:
+  - value: true
```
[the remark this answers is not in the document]
cretz: I will confirm by updating the docker compose file I use.

internal/internal_eager_activity.go
cretz: It is not important now that I am keeping reservation and execution close together here. I may combine them.

> and the lock with an atomic int.

Can't really do this as the actual act of reserving is to obtain a value from a channel which needs to be in the same critical section as check of these values against max concurrent and update of these values.

bergundy: > Can't really do this as the actual act of reserving is to obtain a value from a channel which needs to be in the same critical section as check of these values against max concurrent and update of these values.

You can consume the permits from the counter before taking from the channel and return the unused permits when read from channel would block.

internal/internal_worker.go
```diff
@@ -1361,6 +1363,11 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 		DefaultHeartbeatThrottleInterval:      options.DefaultHeartbeatThrottleInterval,
 		MaxHeartbeatThrottleInterval:          options.MaxHeartbeatThrottleInterval,
 		cache:                                 cache,
+		eagerActivityExecutor: newEagerActivityExecutor(eagerActivityExecutorOptions{
+			disabled:      options.DisableEagerActivities,
+			taskQueue:     taskQueue,
+			maxConcurrent: options.MaxConcurrentEagerActivityExecutionSize,
```
bergundy, line 1369: I'd take the min of MaxConcurrentEagerActivityExecutionSize and MaxConcurrentActivityExecutionSize to avoid unnecessarily reading from the channel

bergundy: You could also consider warning if eager max is greater than general max

cretz: The channel is already bound by max concurrent activity execution size

.github/workflows/docker/dynamic-config-custom.yaml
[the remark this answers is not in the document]
cretz: Confirmed it is not yet enabled

internal/internal_eager_activity.go
cretz: Learned it's actually in 3 other places. Basically `stopWG` is used throughout to indicate "process that must complete before stop completes". So it's a known shared piece.

internal/internal_eager_activity.go
cretz: Left it there for now

internal/internal_eager_activity.go
cretz: Unnecessary IMO since this is already in an eager activity context, but I changed anyways

internal/internal_eager_activity.go
Sushisource: Cool, IMO that's worth encapsulating with a clear name and comment, but up to you.

PR #840, opened 2022-06-28
Helper for loading history from JSON

### LINKED ISSUE #808: Add ReplayWorkflowHistoryFromReader (or similar)

I'd like to use go:embed to access replay history files in the repository for testing. However, the current replay API seems to require me parsing the contents into a history object myself or not using go:embed and using a static file name instead.

Add `ReplayWorkflowHistoryFromReader(logger log.Logger, jsonFile io.Reader) error` and `ReplayPartialWorkflowHistoryFromReader(logger log.Logger, jsonfile io.Reader, lastEventID int64) error` (or similar) to the `WorkflowReplayer` interface and implementations. 

**Additional context**
Adding these APIs could also enable other future uses where replay histories were stored on a remote server, or stored gzipped, etc.

cretz:
## What was changed

Added `client.HistoryFromJSON` helper to load history

## Why?

Other forms of obtaining history JSON need to be supported beyond what's in the replayer

## Checklist

1. Closes #808

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -31,9 +31,11 @@ package client
 
 import (
 	"context"
+	"io"
 
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
+	historypb "go.temporal.io/api/history/v1"
 	"go.temporal.io/api/workflowservice/v1"
 
 	"go.temporal.io/sdk/converter"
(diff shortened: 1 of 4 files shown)


bergundy (approved):
Nice that you have the last event ID option, could be useful in other SDKs

cretz:
I only added that option because our replayer has overloads that accept that option

Issue #832, opened 2022-06-17
temporal SDK not compatible with temporal API v1.9.0


gregorhollmig:
## Expected Behavior

I cannot build my service using temporal anymore.

With release v1.9.0 of https://github.com/temporalio/api-go/releases/tag/v1.9.0 the latest temporal SDK is not compatible anymore

## Actual Behavior

When building the service I get the message. Checking the latest code hier in master this is already adapted and I think just a release is required.

```
go.temporal.io/sdk/client/service_proxy.go:50:9: cannot use &workflowServiceProxyServer{…} (value of type *workflowServiceProxyServer) as type workflowservice.WorkflowServiceServer in return statement:
	*workflowServiceProxyServer does not implement workflowservice.WorkflowServiceServer (missing GetWorkerBuildIdOrdering method)
```

## Steps to Reproduce the Problem

1. Build application with latest temporal API&SDK

## Specifications

  - Version:  v1.15.0
  - Platform: MacOS

cretz:
Yes, you have to use the dependency specified in the go.mod of the SDK version you are using. These are not meant to be upgraded independently.

(if you must use `v1.9.0` of the API dependency, you can use the latest `master`, but of course we always suggest a properly tagged release)

twmb:
Technically, api-go's change was a breaking change, as evidenced here -- if I `go get -u ./...`, my compilation should not break.

There isn't really a great solution for this AFAICT since these are separate repos. In my own projects, once I've frozen an interface with 1.0, I've either added extension interfaces or bumped the major version.

cretz:
> if I go get -u ./..., my compilation should not break.

In theory yes, however with many Go libraries these days, doing things such as adding a method to an interface in a point release is common sure to major version change hurtles.

> or bumped the major version

This is difficult in Go due to being forced to change everyone's import path (e.g. `/v2`)

> Are you already planning to release a new version of the go sdk with support for api v.1.9.0?

Yes, next release will have that. In the meantime, please continue to use the version required in go.mod or use `master` (not recommended of course).

Tarang:
I'm receiving this error across multiple projects. I upgraded from temporal 1.12 to temporal 1.15 while building my project.

```
# go.temporal.io/sdk/client
[...]

PR #849, opened 2022-07-08
Forward compatibility for service proxy

cretz:
## What was changed

Embedded `UnimplementedWorkflowServiceServer` for our service proxy server implementation

## Why?

Most gRPC server impls embed the unimplemented form so that they can be forward compatible with RPC additions. We intentionally did not because we wanted to make sure, at compile time, we manually handled every gRPC call including new ones that came about.

However, this is breaking users that are independently upgrading the API from the SDK (see #832). So it has been decided that instead of breaking them, now those using service proxy (mostly used for MITM'ing Temporal for centralized encryption) will just get gRPC failures at runtime when attempting to make newer service calls. This is not that harmful since they would have gotten a gRPC error anyways for unknown RPC call, we just now no longer break at compile time on independent API upgrade (but CI check will still catch if we didn't regen the proxy).


bergundy:
As long as there's some process to verify we cover all methods this is fine

bergundy (approved):

cretz:
> As long as there's some process to verify we cover all methods this is fine

It only verifies in CI if we upgrade the API. A user can independently upgrade the API and we may not be covering all the methods (the genesis of this issue).

Issue #628, opened 2021-11-11
Workflow API to check if ContinueAsNew is required


rylandg:
When writing infinitely running Workflows, handling ContinueAsNew becomes an immediate requirement in the design. For applications that have multiple infinitely running Workflow types, managing the correct ContinueAsNew timing/behavior becomes tedious and frustrating. 

To solve this problem, I personally resorted to wrapping the Workflow API in a struct. This essentially functions as an unofficial interceptor and maintains a counter of "actions" (transitions basically) by incrementing the count each time a stateful API is called (ExecuteActivity, SignalWorkflow etc). Then each infinitely running Workflow is required to regularly check the counter and see if its exceeded the safe threshold. If it has, we then ContinueAsNew.

It would be ideal if this was provided out of the box. Ideally on the Client it would be possible to specify the safe limits for a given Workflow type such that the API can be dynamically configured for each specific case. At minimum it would be great to know the number of events and history size for a Workflow.

bergundy:
This was already requested for TS https://github.com/temporalio/sdk-typescript/issues/357.
Once we implement it there we'll port to the other SDKs.

robholland:
The history size is now available in WorkflowInfo protobuf so this should be easy to add.

PR #851, opened 2022-07-08
Add workflow.GetInfo(ctx).GetCurrentHistoryLength()

cretz:
## What was changed

Added `workflow.GetInfo(ctx).GetCurrentHistoryLength()` and a helper to set during unit tests

## Why?

So people can use it to make continue-as-new determinations

## Checklist

1. Part of #628

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -826,6 +826,8 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	case enumspb.EVENT_TYPE_WORKFLOW_TASK_STARTED:
 		// Set replay clock.
 		weh.SetCurrentReplayTime(common.TimeValue(event.GetEventTime()))
+		// Set history length as this event's ID
+		weh.workflowInfo.currentHistoryLength = int(event.EventId)
 		// Reset the counter on command helper used for generating ID for commands
 		weh.commandsHelper.setCurrentWorkflowTaskStartedEventID(event.GetEventId())
 		weh.workflowDefinition.OnWorkflowTaskStarted(weh.deadlockDetectionTimeout)
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -338,6 +338,10 @@ func (env *testWorkflowEnvironmentImpl) setStartTime(startTime time.Time) {
 	env.workflowInfo.WorkflowStartTime = env.mockClock.Now()
 }
 
+func (env *testWorkflowEnvironmentImpl) setCurrentHistoryLength(length int) {
+	env.workflowInfo.currentHistoryLength = length
+}
+
 func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(params *ExecuteWorkflowParams, callback ResultHandler, startedHandler func(r WorkflowExecution, e error)) (*testWorkflowEnvironmentImpl, error) {
 	// create a new test env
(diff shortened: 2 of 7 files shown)


test/integration_test.go
```diff
@@ -2425,6 +2425,32 @@ func (ts *IntegrationTestSuite) registerNamespace() {
 	}
 }
 
+func (ts *IntegrationTestSuite) TestHistoryLength() {
```
bergundy, line 2428: To get some extra coverage, I'd evict this workflow from cache (fail a task?) to verify that the history length is correct during replay too.

bergundy (approved):
Approved with single comment

test/integration_test.go
cretz: Maybe I can complete the workflow then run a query for the same value I'm returning

cretz: Note, our integration tests run in no-cache mode too

bergundy: That should be enough then, I'd just maybe mention that we get the coverage from that

internal/internal_event_handlers.go
```diff
@@ -826,6 +826,8 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	case enumspb.EVENT_TYPE_WORKFLOW_TASK_STARTED:
 		// Set replay clock.
 		weh.SetCurrentReplayTime(common.TimeValue(event.GetEventTime()))
+		// Set history length as this event's ID
+		weh.workflowInfo.currentHistoryLength = int(event.EventId)
```
Sushisource, line 830: I don't know if this is sufficient. Based on what's going on here: https://cs.github.com/temporalio/sdk-go/blob/2f76f4de20f0171cb67a6881853c6969f2324761/internal/internal_task_handlers.go#L910

Event ids can be processed out-of-order when looking at markers, which makes sense, but also means this line should probably be more like:
`weh.workflowInfo.currentHistoryLength = max(weh.workflowInfo.currentHistoryLength, int(event.EventId))

test/integration_test.go
Sushisource: Definitely add some tests using local activities with and without replay too, per the ordering thing I mention in my other comment

internal/internal_event_handlers.go
cretz: This is just the event ID of workflow task started. While we do process change the order of history during processing, I believe we only do it one task at a time. So at the top of that loop iteration from your link there is `reorderedHistory.NextCommandEvents()`. Arguably I could just put this value right after we obtain that, but I think the effect is the same.

I may have to re-read the code to confirm, it is a bit hard to follow.

internal/internal_workflow_testsuite.go
```diff
@@ -338,6 +338,10 @@ func (env *testWorkflowEnvironmentImpl) setStartTime(startTime time.Time) {
 	env.workflowInfo.WorkflowStartTime = env.mockClock.Now()
 }
 
+func (env *testWorkflowEnvironmentImpl) setCurrentHistoryLength(length int) {
+	env.workflowInfo.currentHistoryLength = length
+}
```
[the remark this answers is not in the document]
Sushisource: Got it

test/integration_test.go
cretz: Updated the integration test to also use local activities and added comment making it clear we use the zero-cache integration test to confirm replay (and have also manually confirmed that the zero-cache does replay properly).

PR #852, opened 2022-07-13
Pull in v1.10.0 of go.temporal.io/api

mmcshane:
## What was changed

- Update v1.9.0 to v1.10.0 of go.temporal.io/api
- Regenerate proxy and interceptor code
- Add new event <-> command mappings

## Why?

Need the Update API to compile OSS server and eventually implement Update handlers in the SDK

## Checklist

1. Closes 
n/a

2. How was this tested:

Existing tests 

3. Any docs updates needed?

No


cretz (approved):
Looks good.

On the API side, not the biggest fan of tagging unimplemented API (same for the versioning stuff) since we can learn API-affecting things during implementation often. We know we'll at least need one more RPC to fetch update response asynchronously. But so long as we're not afraid to fix any of the APIs based on what we learn, no problem.

mmcshane:
> On the API side, not the biggest fan of tagging unimplemented API ...

I feel you but I'm not sure what to do otherwise? We could depend on a hash but that's functionally equivalent but uglier in go.mod. We definitely expect to discover mistakes and re-tag. Maybe tagging without creating the GH release would be a better middle-ground.

cretz:
> I feel you but I'm not sure what to do otherwise?

Basically the solution would be to do all dev outside of the associated repos until you're ready to land all of it together. But I can understand that is a bit rough and doesn't feel very incremental. I think what's here is fine, let's just not be afraid to break API compat if needed since nobody is using it yet.

bergundy:
We should mark all of these new APIs as experimental so users know not to rely on them.
@rylandg another thing for you to consider.

PR #863, opened 2022-07-21
Support really short heartbeat throttling interval

### LINKED ISSUE #859: Cannot set really short max heartbeat throttle interval
## Expected Behavior

Expect to be able to set this to, like, 1 nanosecond and it work

## Actual Behavior

For whatever reason in #660, I put `context.WithTimeout(ctx, i.heartbeatThrottleInterval)` which means that context times out immediately

cretz:
## What was changed

Stop timing out the RPC context for recording heartbeat before it can even record

## Why?

Really short timeouts were preventing calls from being made properly

## Checklist

1. Closes #859

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1724,7 +1724,16 @@ func (i *temporalInvoker) Heartbeat(ctx context.Context, details *commonpb.Paylo
 
 func (i *temporalInvoker) internalHeartBeat(ctx context.Context, details *commonpb.Payloads) (bool, error) {
 	isActivityCanceled := false
-	ctx, cancel := context.WithTimeout(ctx, i.heartbeatThrottleInterval)
+	// We don't want the recording of the heartbeat to keep retrying the RPC
+	// longer than the throttle interval. However, sometimes the interval is so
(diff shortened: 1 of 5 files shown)


internal/internal_task_handlers.go
```diff
@@ -1724,7 +1724,16 @@ func (i *temporalInvoker) Heartbeat(ctx context.Context, details *commonpb.Paylo
...
-	ctx, cancel := context.WithTimeout(ctx, i.heartbeatThrottleInterval)
+	// We don't want the recording of the heartbeat to keep retrying the RPC
+	// longer than the throttle interval. However, sometimes the interval is so
+	// small that the context is cancelled before it even starts the call.
+	// Therefore, we'll make sure not to timeout the context faster than the
+	// minimum RPC timeout.
+	recordTimeout := i.heartbeatThrottleInterval
+	// if recordTimeout < minRPCTimeout {
+	// 	recordTimeout = minRPCTimeout
+	// }
```
Sushisource, line 1735: Seems like this should not be commented out

cretz: Fixed.

Sushisource (approved):

PR #867, opened 2022-07-25
WIP Update handling

mmcshane:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -64,7 +64,6 @@ type (
 		handleInitiatedEvent()
 
 		handleCommandSent()
-
 		setData(data interface{})
 		getData() interface{}
 	}
@@ -130,9 +129,13 @@ type (
 		*naiveCommandStateMachine
 	}
 
-	upsertSearchAttributesCommandStateMachine struct {
+	completeOnSendStateMachine struct {
 		*naiveCommandStateMachine
 	}
+
+	completeWorkflowUpdateCommandStateMachine = completeOnSendStateMachine
+	acceptWorkflowUpdateCommandStateMachine   = completeOnSendStateMachine
+	upsertSearchAttributesCommandStateMachine = completeOnSendStateMachine
 
 	commandsHelper struct {
 		nextCommandEventID int64
@@ -184,6 +187,8 @@ const (
 	commandTypeUpsertSearchAttributes    commandType = 6
 	commandTypeCancelTimer               commandType = 7
 	commandTypeRequestCancelActivityTask commandType = 8
+	commandTypeAcceptWorkflowUpdate      commandType = 9
+	commandTypeCompleteWorkflowUpdate    commandType = 10
 )
 
 const (
@@ -262,6 +267,10 @@ func (d commandType) String() string {
 		return "CancelTimer"
 	case commandTypeRequestCancelActivityTask:
 		return "RequestCancelActivityTask"
+	case commandTypeAcceptWorkflowUpdate:
+		return "AcceptWorkflowUpdate"
+	case commandTypeCompleteWorkflowUpdate:
+		return "CompleteWorkflowUpdate"
 	default:
 		return "Unknown"
 	}
@@ -281,6 +290,65 @@ func (h *commandsHelper) newCommandStateMachineBase(commandType commandType, id 
 		state:   commandStateCreated,
 		history: []string{commandStateCreated.String()},
 		helper:  h,
+	}
+}
+
+func (h *commandsHelper) acceptWorkflowUpdate(updateID string) {
+	sm := h.newAcceptWorkflowUpdateStateMachine(updateID)
+	h.addCommand(sm)
+}
+
+func (h *commandsHelper) completeWorkflowUpdate(
+	updateID string,
+	success *commonpb.Payloads,
+	failure *failurepb.Failure,
+	durabilityPref enumspb.UpdateWorkflowRejectionDurabilityPreference,
+) {
+	sm := h.newCompleteWorkflowUpdateStateMachine(updateID, success, failure, durabilityPref)
+	h.addCommand(sm)
+}
+
+func (h *commandsHelper) newCompleteWorkflowUpdateStateMachine(
+	updateID string,
+	success *commonpb.Payloads,
+	failure *failurepb.Failure,
+	durabilityPref enumspb.UpdateWorkflowRejectionDurabilityPreference,
+) commandStateMachine {
+	attrs := &commandpb.Command_CompleteWorkflowUpdateCommandAttributes{
+		CompleteWorkflowUpdateCommandAttributes: &commandpb.CompleteWorkflowUpdateCommandAttributes{
+			UpdateToken:          []byte(updateID), // todo: this will change to UpdateId
+			DurabilityPreference: durabilityPref,
+		},
+	}
+	if failure != nil {
+		attrs.CompleteWorkflowUpdateCommandAttributes.Result = &commandpb.CompleteWorkflowUpdateCommandAttributes_Failure{
+			Failure: failure,
+		}
+	} else {
+		attrs.CompleteWorkflowUpdateCommandAttributes.Result = &commandpb.CompleteWorkflowUpdateCommandAttributes_Success{
+			Success: success,
+		}
+	}
+	return &completeWorkflowUpdateCommandStateMachine{
+		h.newNaiveCommandStateMachine(commandTypeCompleteWorkflowUpdate, updateID, &commandpb.Command{
+			CommandType: enumspb.COMMAND_TYPE_COMPLETE_WORKFLOW_UPDATE,
+			Attributes:  attrs,
+		}),
+	}
+}
+
+func (h *commandsHelper) newAcceptWorkflowUpdateStateMachine(
+	updateID string,
+) commandStateMachine {
+	return &acceptWorkflowUpdateCommandStateMachine{
+		h.newNaiveCommandStateMachine(commandTypeAcceptWorkflowUpdate, updateID, &commandpb.Command{
+			CommandType: enumspb.COMMAND_TYPE_ACCEPT_WORKFLOW_UPDATE,
+			Attributes: &commandpb.Command_AcceptWorkflowUpdateCommandAttributes{
+				AcceptWorkflowUpdateCommandAttributes: &commandpb.AcceptWorkflowUpdateCommandAttributes{
+					UpdateToken: []byte(updateID), // todo: this will change to UpdateId
+				},
+			},
+		}),
 	}
 }
 
@@ -820,7 +888,7 @@ func (d *markerCommandStateMachine) handleCommandSent() {
 	}
 }
 
-func (d *upsertSearchAttributesCommandStateMachine) handleCommandSent() {
+func (d *completeOnSendStateMachine) handleCommandSent() {
 	// This command is considered as completed once command is sent.
 	switch d.state {
 	case commandStateCreated:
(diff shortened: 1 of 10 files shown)


internal/error.go
```diff
@@ -842,79 +869,95 @@ func ConvertErrorToFailure(err error, dc converter.DataConverter) *failurepb.Fai
...
-		failure.FailureInfo = &failurepb.Failure_ApplicationFailureInfo{ApplicationFailureInfo: failureInfo}
-		failure.StackTrace = err.StackTrace()
-	case *workflowPanicError:
-		failureInfo := &failurepb.ApplicationFailureInfo{
-			Type:         getErrType(&PanicError{}),
-			NonRetryable: true,
+	originalError := err
+
+Loop:
+	for ; err != nil; err = errors.Unwrap(err) {
```
cretz, line 875: This is gonna swallow errors. What would have happened before if I had `*MyError` that wrapped `*MyOtherError` that wrapped `*ApplicationError` that had details? Before it'd be a `MyError` w/ cause `MyOtherError` with cause `ApplicationFailure` of whatever type and details. Now, this would be moving those details to the top-level error improperly.

Why do we have to change this function at all? If you are concerned about unwrapping your update error specifically, unwrap it specifically where you handle it.

internal/error.go
```diff
@@ -273,6 +277,29 @@ var (
...
+	return e.error
+}
+
+func (e durabilityBypassError) PreferBypassDurability() bool {
+	return true
+}
+
+func extractDurabilityBypassPreference(err error) enumspb.UpdateWorkflowRejectionDurabilityPreference {
+	var intf interface{ PreferBypassDurability() bool }
+	if errors.As(err, &intf) && intf.PreferBypassDurability() {
```
cretz, line 297: I don't think you should use `errors.As` because that walks the chain, and you don't care what weird thing they did 10 errors deep. The previously authors of the Go SDK made this mistake far too many times and it has cost us with confusion when checking error types. A simple `bypassErr, _ := err.(interface { PreferBypassDurability() bool }); bypassErr != nil && bypassErr.PreferBypassDurability()` I think would work and be clear that that has to be the actual returned error to not persist.

internal/internal_command_state_machine.go
```diff
@@ -284,6 +293,65 @@ func (h *commandsHelper) newCommandStateMachineBase(commandType commandType, id
 	}
 }
 
+func (h *commandsHelper) acceptWorkflowUpdate(updateID string) {
```
cretz, line 296: While we don't _necessarily_ need to do it here, often their pattern was to create one of these command state machines for an update, and then have several methods on the _same_ state machine. So a single update state machine could have `accept`, `complete`, `reject`, etc.

This can help with cancellation which may not matter here, but also may keep things a bit cleaner. A multi-interaction thing like an update may be clearer as a single state machine with actions that can be called on it.

internal/internal_workflow.go
```diff
@@ -1392,26 +1434,94 @@ func setQueryHandler(ctx Context, queryType string, handler interface{}) error {
...
+		dataConverter: getDataConverterFromWorkflowContext(ctx),
+	}
+	return nil
+}
+
+func validateSameParams(fn1, fn2 interface{}) error {
+	fn1Type := reflect.TypeOf(fn1)
+	fn2Type := reflect.TypeOf(fn2)
+
+	for _, fnType := range [...]reflect.Type{fn1Type, fn2Type} {
```
cretz, line 1465: Meh, two if statements ain't that bad, heh

internal/internal_workflow.go
```diff
@@ -1436,6 +1546,29 @@ func (h *queryHandler) execute(input []interface{}) (result interface{}, err err
 	return executeFunction(h.fn, input)
 }
 
+func (h *updateHandler) validate(input []interface{}) (err error) {
+	defer func() {
+		if p := recover(); p != nil {
```
cretz, line 1551: I wonder if we should have some worker-level option of whether to treat validator panics as persisted or not. We have `worker.Options.WorkflowPanicPolicy` that defines how to treat workflow panics, so an `UpdateValidatorPanicPolicy` may make sense. I could argue for either way on what the default should be though.

Actually, now that I think about it, I guess we can keep this (default durable) and if they don't like that, they can recover themselves and make it non-durable.

internal/internal_workflow.go
```diff
@@ -1436,6 +1546,29 @@ func (h *queryHandler) execute(input []interface{}) (result interface{}, err err
...
+			st := getStackTraceRaw("update validator [panic]:", 7, 0)
+			err = fmt.Errorf("update validator panic: %v, stack trace: %v", p, st)
```
cretz, line 1553: I think you should reuse the `PanicError` in this package, if not for other reasons, just to have that `PanicError` type on the application failure info same as we do with other panics.

internal/workflow.go
```diff
@@ -297,6 +297,27 @@ type (
...
+		// Validator is an optional (i.e. can be left nil) func with exactly the
+		// same type signature as the required update handler func but returning
+		// only a single value of type error. The implementation of this
+		// function MUST be safe and MUST NOT alter workflow state in any way.
+		// If the returned error can be cast to `interface{
+		// PreferBypassDurability() bool }` then the preference indicated by
+		// invoking PreferBypassDurability() will be passed upstream.  A panic
+		// from this function will be treated as equivalent to returning an
+		// error with a positive durability bypass preference.
+		Validator interface{}
```
cretz, line 314: I was thinking about this a bit. Question - do we think non-durable errors will be a fairly popular thing or rare?

If we do think they would be popular, I wonder if we want to do another approach to make it more explicit. There are basically two options here: 1) Return two errors and have first be non-durable (not good though, easy to get switched up), or 2) return a custom `ValidationResolution` type which would be an interface w/ provide three helpers/impls for. With the latter, you can have `func ValidationSuccess() ValidationResolution`, `func ValidationDurableError(error) ValidationResolution`, and `func ValidationNonDurableError(error) ValidationResolution`.

> If the returned error can be cast to `interface{ PreferBypassDurability() bool }` then the preference indicated by invoking PreferBypassDurability() will be passed upstream. 
While we should keep this doc, we should provide a helper for this, e.g. `func WrapAsNonDurableUpdateError(err) error` or something.

internal/workflow.go
```diff
@@ -297,6 +297,27 @@ type (
...
+	}
+)
+
+const (
+	UpdateDurabilityPreferenceStore  = 0
+	UpdateDurabilityPreferenceBypass = 1
```
cretz, line 320: Are these consts used anywhere?

internal/error.go
mmcshane: The idea was for those cases where one of these well-known error types gets wrapped up in a bypassDurabilityError, that error wouldn't become the "top" Failure. But as you see with the integration test failures, there are other places that are depend on the structure returned here and how it maps to the concrete error types so I'll need to revert this bit.

internal/error.go
mmcshane: I'm roughly ok with that. Just means that the top-level error returned MUST implement that interface if the workflow author wants bypasss. One last fmt.Errorf with %w would hide the intent. Do you foresee a situation where the bypassError is somewhere in the error cause chain but we _wouldn't_ want to honor it?

internal/internal_workflow.go
mmcshane: lol fair point. I was fussing with test coverage. hazards of reviewing a wip ;-)

internal/internal_workflow.go
mmcshane: There's also the errors that come from the inability to deser commonpb.Payloads or to invoke the registered validator/executor funcs .... the user doesn't have any control over those and they default to bypass. I'm not sure whether we need an option to control that bypass/nobypass or not.

internal/workflow.go
mmcshane: Nope. Detritus.

internal/error.go
cretz: Yeah, I think we can leave this function alone and you can just unwrap your specific bypass-durability error when handling update validation result

internal/error.go
cretz: While I don't foresee that situation, I think clearly requiring that the error is that type instead of allowing it anywhere in the chain is much easier to comprehend by users. If they wrap it, they lost it. Same way https://github.com/grpc/grpc-go/blob/v1.48.0/status/status.go#L91 works for gRPC instead of walking the chain (though they have more reason to suspect people might unintentionally wrap than we do).

It almost makes me want to use a custom result type as I mentioned in another comment. When adding what is essentially a tertiary return type such as we're doing here, that's hard to model by type of error.

internal/internal_command_state_machine.go
mmcshane: I think I'm misunderstanding the commandStateMachine abstraction. I had believed it to be a SM _for commands_. Like, every single command gets an appropriate SM instance. Accept/Complete being distinct commands would thus be distinct SM types. But it seems like you're saying that these command SMs are really more general and are intended to be used throughout event processing?

internal/internal_workflow.go
cretz: In retrospect, I agree. If they want panics to not be persisted, they can recover from them themselves. Panics shouldn't happen anyways.

internal/internal_command_state_machine.go
cretz: > I had believed it to be a SM _for commands_

Kinda. It's basically for every "state" a "thing" can go through. The result is commands, but it's really for "things", which can result in multiple states which we don't have many of. Currently:

* An activity state machine can send schedule or cancel (though this state machine starts a cancel state machine for unknown reasons rather than sending the cancel itself)
* A child workflow state machine that can send a start or cancel
* A timer state machine that can send a start or cancel (though like activity, this itself starts a separate cancel state machine which I think is silly)

This will be the first "thing" that can have three commands that I know of. It can accept (leaving it around), complete (removing itself), and reject (removing itself).

internal/workflow.go
mmcshane: I honestly don't have a clue how popular it will be so not much help there, sorry.

Both approaches fulfill the abstract requirement of the feature so I'd be happy with either and I prefer to leave these design details to the sdk author. As someone who has written a fair amount of Go I have my own intuition on which is preferrable, but It think it's better as your choice.

If we do stick with just error, I agree on a helper func. I would export the bypassDurability func from errors.go

internal/internal_workflow.go
mmcshane: No problem :+1:

internal/internal_workflow.go
mmcshane: In a sense you want to bypass for those so that someone/something can't flood you with garbage input as a DoS. But then again it might be useful in a development setting to capture that info for debugging. This duality is what let me toward a workflow option. If we don't want an option then forcing bypass is probably safer.

cretz: > This duality is what let me toward a workflow option. If we don't want an option then forcing bypass is probably safer.

Agreed. You could consider a deserialization failure as a panic from the "unexpected input" POV. Otherwise, if you craft a panic-causing input, the DoS is still a thing.

internal/error.go
mmcshane: Resolution: will remain unchanged.

internal/error.go
mmcshane: Current direction is to move away from distinguishing between bypass and non-bypass errors. Any error coming from a validator will be implicitly prefer-bypass; errors from execution function remain implicitly non-bypass.

internal/internal_workflow.go
mmcshane: Current direction is that any error or panic from validation is implicitly prefer-bypass so at this point a preference would not be useful.

internal/workflow.go
mmcshane: Current direction: return error from validator, validation error or panic will always be treated as prefer-bypass since that's the only reason you created a validation func in the first place. Otherwise it would just be inline checking in the execution func.

PR #879, opened 2022-08-09
Doc fixes

### LINKED ISSUE #855: Panic with double stop on worker
## Expected Behavior
Ignore the second stop

## Actual Behavior
panic: close of closed channel

https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker.go#L1051
Can it check if the channel is closed using assertNotStopped()?

Call worker.stop twice

### LINKED ISSUE #868: panic if `Run` and `Stop` race
`internal_worker.AggregatedWorker#Run` calls `aw.Start`, which asserts that the worker has not yet been stopped.

`Run` is a blocking function, it must be executed in a goroutine if the user wants to stop the worker independently. The user may then wait for an input signal and `Stop` the worker. It is possible that `Run` has not yet started and `Stop` executes before `Run`. In this tight sequence of events, the code will panic: https://github.com/temporalio/sdk-go/blob/dea0cfa509b981c05ea4e9c9ad2f841845997574/internal/internal_worker.go#L955

Since `Run` returns an error, I'd expect this code to just return `ErrClosed` or something similar.

### LINKED ISSUE #870: Documentation fixes for error handling

https://pkg.go.dev/go.temporal.io/sdk/workflow#hdr-Error_Handling talks about `error.CustomError()` which is not a thing.

Make sure none of the error references are stale. Also update workflow run's `Get` to clarify that result and error are mutually exclusive and you'll never get a result and an error.

cretz:
## What was changed

Doc fixes

## Checklist

1. Closes #868
1. Closes #855
1. Closes #870

### DIFF
--- a/workflow/doc.go
+++ b/workflow/doc.go
@@ -282,52 +282,11 @@ pattern, extra care needs to be taken to ensure the child workflow is started be
 
 Error Handling
 
-Activities and child workflows can fail. You could handle errors differently based on different error cases. If the
-activity returns an error as errors.New() or fmt.Errorf(), those errors will be converted to error.GenericError. If the
-activity returns an error as error.NewCustomError("err-reason", details), that error will be converted to
-*error.CustomError. There are other types of errors like error.TimeoutError, error.CanceledError and error.PanicError.
-So the error handling code would look like:
+Activities and child workflows can fail. Activity errors are *temporal.ActivityError and errors during child workflow
+execution are *temporal.ChildWorkflowExecutionError. The cause of the errors may be types like
+*temporal.ApplicationError, *temporal.TimeoutError, *temporal.CanceledError, and *temporal.PanicError.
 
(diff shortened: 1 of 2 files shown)


workflow/doc.go
```diff
@@ -282,52 +282,11 @@ pattern, extra care needs to be taken to ensure the child workflow is started be
 
 Error Handling
 
-Activities and child workflows can fail. You could handle errors differently based on different error cases. If the
-activity returns an error as errors.New() or fmt.Errorf(), those errors will be converted to error.GenericError. If the
-activity returns an error as error.NewCustomError("err-reason", details), that error will be converted to
-*error.CustomError. There are other types of errors like error.TimeoutError, error.CanceledError and error.PanicError.
-So the error handling code would look like:
-
-	err := workflow.ExecuteActivity(ctx, YourActivityFunc).Get(ctx, nil)
```
alexshtin, line 291: image

alexshtin (approved):

PR #881, opened 2022-08-09
Share connection across different clients

### LINKED ISSUE #861: Support creating new client from existing client's connection

People need to create a client from an existing connection but just change the non-connection options

Options:

* Accept something a `WorkflowService` in client options which means all connection options are ignored (this same will have to be done for operator service if/when we get there)
* Accept a grpc.ClientConn in client options which means all connection options are ignored. We'd also have to expose this connection from the client somehow (really hoping not to add a new interface method)
* Make a `Clone(existingClient Client, newOptions Options) (Client, error)` call

cretz:
## What was changed

* Added `NewClientFromExisting` function to create a new client with the same connection but different client options
* Made sure to support the different namespaces in the gRPC interceptor metrics

## Why?

Users want to use the same connection for multiple namespaces

## Checklist

1. Closes #861

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -468,6 +468,16 @@ func NewClient(options Options) (Client, error) {
 	return internal.NewClient(options)
 }
 
+// NewClientFromExisting creates a new client using the same connection as the
+// existing client. This means all options.ConnectionOptions are ignored and
+// options.HostPort is ignored. The existing client must have been created from
+// this package and cannot be wrapped. Currently, this always attempts an eager
+// connection even if the existing client was created with NewLazyClient and has
+// not made any calls yet.
+func NewClientFromExisting(existingClient Client, options Options) (Client, error) {
+	return internal.NewClientFromExisting(existingClient, options)
+}
+
 // NewNamespaceClient creates an instance of a namespace client, to manage
 // lifecycle of namespaces. This will not attempt to connect to the server
 // eagerly and therefore may not fail for an unreachable server until a call is
--- a/internal/common/metrics/grpc.go
+++ b/internal/common/metrics/grpc.go
@@ -59,9 +59,17 @@ func NewGRPCInterceptor(defaultHandler Handler, suffix string) grpc.UnaryClientI
 
 		// Only take method name after the last slash
 		operation := method[strings.LastIndex(method, "/")+1:]
-		handler = handler.WithTags(map[string]string{OperationTagName: operation})
+		tags := map[string]string{OperationTagName: operation}
+
+		// Since this interceptor can be used for clients of different name, we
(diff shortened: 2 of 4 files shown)


client/client.go
```diff
@@ -468,6 +468,16 @@ func NewClient(options Options) (Client, error) {
 	return internal.NewClient(options)
 }
 
+// NewClientFromExisting creates a new client using the same connection as the
+// existing client. This means all options.ConnectionOptions are ignored and
+// options.HostPort is ignored. The existing client must have been created from
+// this package and cannot be wrapped. Currently, this always attempts an eager
+// connection even if the existing client was created with NewLazyClient and has
+// not made any calls yet.
+func NewClientFromExisting(existingClient Client, options Options) (Client, error) {
```
Sushisource, line 477: It might make sense to make a narrower options type here with just the stuff that won't be ignored.

If that can be re-used in a backwards compat way inside the existing options, that's gravy. If not it might still be worth it just to make this less potentially confusing, but up to you.

Sushisource (approved):

client/client.go
dnr: In theory I like splitting the struct into "connection options" and "other options" but I don't think it's worth it for this feature that most people won't ever use

internal/common/metrics/grpc.go
```diff
@@ -59,9 +59,17 @@ func NewGRPCInterceptor(defaultHandler Handler, suffix string) grpc.UnaryClientI
 
 		// Only take method name after the last slash
 		operation := method[strings.LastIndex(method, "/")+1:]
-		handler = handler.WithTags(map[string]string{OperationTagName: operation})
+		tags := map[string]string{OperationTagName: operation}
+
+		// Since this interceptor can be used for clients of different name, we
+		// attempt to extract the namespace out of the request. All namespace-based
+		// requests have been confirmed to have a top-level namespace field.
```
dnr, line 66: Does this deserve a comment in the proto file? I mean I wouldn't do a PR just for that, but bundle it into your next api PR? :)

dnr (approved):

client/client.go
cretz: > It might make sense to make a narrower options type here with just the stuff that won't be ignored.

This would entail just copying all but two fields just for this method and hoping I properly keep them in sync when adding more. I think it's reasonable to reuse options explaining which are unused.

internal/common/metrics/grpc.go
cretz: Yeah, if I remember on my next one I'll add something like "Some clients expect all namespace-based RPC calls to have a single `namespace` string field in the request".

PR #883, opened 2022-08-15
Add UpsertMemo API

rodrigozhou:
## What was changed

Adding an API for `UpsertMemo`. It merges with current memo and creates a command `ModifyWorkflowProperties` with the updated memo.
This change depends on https://github.com/temporalio/api/pull/218.

## Why?

Feature request: https://github.com/temporalio/sdk-features/issues/119

## Checklist

1. How was this tested:
   - Created new tests.

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -393,11 +393,11 @@ func (wc *workflowEnvironmentImpl) UpsertMemo(memoMap map[string]interface{}) er
 
 	changeID := wc.GenerateSequenceID()
 	wc.commandsHelper.modifyProperties(changeID, memo)
-	wc.updateWorkflowInfoWithProperties(memo) // this is for getInfo correctness
+	wc.updateWorkflowInfoWithMemo(memo) // this is for getInfo correctness
 	return nil
 }
 
-func (wc *workflowEnvironmentImpl) updateWorkflowInfoWithProperties(memo *commonpb.Memo) {
+func (wc *workflowEnvironmentImpl) updateWorkflowInfoWithMemo(memo *commonpb.Memo) {
 	wc.workflowInfo.Memo = mergeMemo(wc.workflowInfo.Memo, memo)
 }
 
@@ -1579,7 +1579,7 @@ func (weh *workflowExecutionEventHandlerImpl) handleWorkflowPropertiesModified(
 	event *historypb.HistoryEvent,
 ) {
 	attributes := event.GetWorkflowPropertiesModifiedEventAttributes()
-	weh.updateWorkflowInfoWithProperties(attributes.UpsertedMemo)
+	weh.updateWorkflowInfoWithMemo(attributes.UpsertedMemo)
 }
 
 func (weh *workflowExecutionEventHandlerImpl) handleRequestCancelExternalWorkflowExecutionInitiated(event *historypb.HistoryEvent) error {
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -384,6 +384,51 @@ func validateAndSerializeSearchAttributes(attributes map[string]interface{}) (*c
 	return attr, nil
 }
 
+func (wc *workflowEnvironmentImpl) UpsertMemo(memoMap map[string]interface{}) error {
+	// This has to be used in WorkflowEnvironment implementations instead of in Workflow for testsuite mock purpose.
+	memo, err := validateAndSerializeMemo(memoMap, wc.dataConverter)
+	if err != nil {
+		return err
+	}
+
+	changeID := wc.GenerateSequenceID()
+	wc.commandsHelper.modifyProperties(changeID, memo)
+	wc.updateWorkflowInfoWithProperties(memo) // this is for getInfo correctness
+	return nil
+}
+
+func (wc *workflowEnvironmentImpl) updateWorkflowInfoWithProperties(memo *commonpb.Memo) {
+	wc.workflowInfo.Memo = mergeMemo(wc.workflowInfo.Memo, memo)
+}
+
+func mergeMemo(current, upsert *commonpb.Memo) *commonpb.Memo {
+	if current == nil {
+		if upsert == nil || len(upsert.Fields) == 0 {
+			return nil
+		}
+		current = &commonpb.Memo{
+			Fields: make(map[string]*commonpb.Payload),
+		}
+	}
+
+	fields := current.Fields
+	for k, v := range upsert.Fields {
+		fields[k] = v
+	}
+	return current
+}
+
+func validateAndSerializeMemo(memoMap map[string]interface{}, dc converter.DataConverter) (*commonpb.Memo, error) {
+	if len(memoMap) == 0 {
+		return nil, errMemoNotSet
+	}
+	memo, err := getWorkflowMemo(memoMap, dc)
+	if err != nil {
+		return nil, err
+	}
+	return memo, nil
+}
+
 func (wc *workflowEnvironmentImpl) RegisterCancelHandler(handler func()) {
 	wrappedHandler := func() {
 		wc.commandsHelper.workflowExecutionIsCancelling = true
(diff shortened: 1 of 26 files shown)


internal/internal_event_handlers.go
```diff
@@ -384,6 +384,53 @@ func validateAndSerializeSearchAttributes(attributes map[string]interface{}) (*c
 	return attr, nil
 }
 
+func (wc *workflowEnvironmentImpl) UpsertMemo(memo_map map[string]interface{}) error {
```
cretz, line 387:
```suggestion
func (wc *workflowEnvironmentImpl) UpsertMemo(memoMap map[string]interface{}) error {
```

Here and elsewhere

internal/internal_event_handlers.go
```diff
@@ -384,6 +384,53 @@ func validateAndSerializeSearchAttributes(attributes map[string]interface{}) (*c
...
+	wc.updateWorkflowInfoWithProperties(memo)
+	return nil
+}
+
+func (wc *workflowEnvironmentImpl) updateWorkflowInfoWithProperties(memo *commonpb.Memo) {
+	wc.workflowInfo.Memo = memo
+}
+
+func mergeMemo(current, upsert *commonpb.Memo) *commonpb.Memo {
+	if current == nil || len(current.Fields) == 0 {
```
cretz, line 407: Could also just `len(current.GetFields()) == 0`

test/integration_test.go
```diff
@@ -1453,6 +1453,7 @@ func (ts *IntegrationTestSuite) TestInterceptorCalls() {
 		"WorkflowOutboundInterceptor.RequestCancelExternalWorkflow": {},
 		"WorkflowOutboundInterceptor.SignalExternalWorkflow":        {},
 		"WorkflowOutboundInterceptor.UpsertSearchAttributes":        {},
+		"WorkflowOutboundInterceptor.UpsertMemo":                    {},
```
cretz, line 1456: Can you provide a full integration test here? I think a good test would be to start a workflow with some memo, assert that memo, then upsert that memo (and set a query saying you've done so), then describe the workflow via the client and confirm the memo was upserted.

If this is not available in the server yet for an integration test, I don't think we should merge the SDK functionality until it is.

bergundy:
Didn't look at the code but something I wish we'd done with search attributes in TS is delay sending the command until the end of an "activation", so if a user calls `UpsertMemo({ a: 1 }); UpsertMemo({ b: 2 }); UpsertMemo({ a: 2 })` the SDK would only issue 1 `UpsertMemo({ a: 2, b: 2})` command.

bergundy:
@rodrigozhou note that I moved the original issue to sdk-features and marked this PR as not closing for that issue, we need to implement in all SDKs.

internal/internal_event_handlers.go
```diff
@@ -384,6 +384,53 @@ func validateAndSerializeSearchAttributes(attributes map[string]interface{}) (*c
...
+
+	changeID := wc.GenerateSequenceID()
+	mergedMemo := mergeMemo(wc.workflowInfo.Memo, memo)
+	wc.commandsHelper.modifyProperties(changeID, mergedMemo)
+	// this is for getInfo correctness
+	wc.updateWorkflowInfoWithProperties(memo)
+	return nil
+}
+
+func (wc *workflowEnvironmentImpl) updateWorkflowInfoWithProperties(memo *commonpb.Memo) {
```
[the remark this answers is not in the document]
rodrigozhou: I'm naming it generic because so it can be extended later when new attributes are added.
I can make it specific, but I suppose new attributes would require creating individual functions to update each of them.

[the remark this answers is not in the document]
cretz: The parameter is very specific though, so it's obviously not capable of future attributes without altering the signature anyways. So might as well not have a generic name w/ one specific param for the one specific use case.

internal/internal_event_handlers.go
```diff
@@ -384,6 +384,53 @@ func validateAndSerializeSearchAttributes(attributes map[string]interface{}) (*c
...
+	wc.workflowInfo.Memo = memo
+}
+
+func mergeMemo(current, upsert *commonpb.Memo) *commonpb.Memo {
+	if current == nil || len(current.Fields) == 0 {
+		if upsert == nil || len(upsert.Fields) == 0 {
+			return nil
```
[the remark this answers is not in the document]
rodrigozhou: In theory, the new memo cannot be nil or an empty map. See `validateAndSerializeMemo` which is executed before merging the memo.

rodrigozhou:
The CI failure is because the CI used the docker-compose repo to start the temporal server, which doesn't include `UpsertMemo` yet. Locally, all tests passes.

test/integration_test.go
```diff
@@ -2515,6 +2517,170 @@ func (ts *IntegrationTestSuite) TestHeartbeatThrottleDisabled() {
 	ts.assertReportedOperationCount("temporal_request_failure", "RecordActivityTaskHeartbeat", 0)
 }
 
+func (ts *IntegrationTestSuite) TestUpsertMemoFromNil() {
+	// TODO(rodrigozhou): unskip after UpsertMemo is released
+	ts.T().Skip("Not implemented in server yet")
```
cretz, line 2522: My approval of the PR is pending @bergundy and @yiminc and others coming to a conclusion on whether we want a capability boolean so users can check whether their server version supports this.

workflow/workflow.go
```diff
@@ -470,6 +470,10 @@ func UpsertSearchAttributes(ctx Context, attributes map[string]interface{}) erro
 	return internal.UpsertSearchAttributes(ctx, attributes)
 }
 
+func UpsertMemo(ctx Context, memo map[string]interface{}) error {
```
cretz, line 473: Every exported function should have Godoc, especially this one :-)

test/integration_test.go
rodrigozhou: Updated to check capability.

test/integration_test.go
```diff
@@ -2516,6 +2518,191 @@ func (ts *IntegrationTestSuite) TestHeartbeatThrottleDisabled() {
...
+	var memo *commonpb.Memo
+	err = run.Get(ctx, &memo)
+	ts.NoError(err)
+
+	// Wait a little bit for ES to update
+	time.Sleep(2 * time.Second)
+
+	// Query ES for memo
+	var nextPageToken []byte
+	resp, err := ts.client.ListWorkflow(ctx, &workflowservice.ListWorkflowExecutionsRequest{
```
cretz, line 2565: I think a simple `DescribeWorkflowExecution` is much better here than listing. But I'm gonna mark this approved anyways, change only if you want.

cretz (approved):

workflow/workflow.go
```diff
@@ -470,6 +470,34 @@ func UpsertSearchAttributes(ctx Context, attributes map[string]interface{}) erro
...
+//	}
+//
+// The workflow memo will eventually be:
+//
+//	map[string]interface{}{
+//		"Key1": 2,
+//		"Key2": true,
+//		"Key3": "seattle",
+//	}
+func UpsertMemo(ctx Context, memo map[string]interface{}) error {
```
cretz, line 497: We should probably make it clear not all servers support this (and maybe say which server version does)

internal/internal_task_handlers.go
```diff
@@ -1441,6 +1442,17 @@ func isCommandMatchEvent(d *commandpb.Command, e *historypb.HistoryEvent, strict
...
+
+	case enumspb.COMMAND_TYPE_MODIFY_WORKFLOW_PROPERTIES:
+		if e.GetEventType() != enumspb.EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED {
+			return false
+		}
+		eventAttributes := e.GetWorkflowPropertiesModifiedEventAttributes()
+		commandAttributes := d.GetModifyWorkflowPropertiesCommandAttributes()
+		if strictMode && !isMemoMatched(eventAttributes.UpsertedMemo, commandAttributes.UpsertedMemo) {
+			return false
+		}
```
yiminc, line 1454: StrictMode was there for historical reason. But I don't think they are needed anymore. We probably should remove strictMode. @cretz 
(No change requested for this PR.)

PR #885, opened 2022-08-19
Spawn update coro directly through dispatcher

mmcshane:
## What was changed

Spawn update coro directly through dispatcher

## Why?

Calling Go/GoNamed with a context from outside the workflow will fail
because the context doesn't have a handle to dispatcher state.

## Checklist

1. Closes 

n/a

2. How was this tested:

Defect found via sdk-features test for update which will land soon

3. Any docs updates needed?

No

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -494,7 +494,7 @@ func defaultUpdateHandler(
 	serializedArgs *commonpb.Payloads,
 	header *commonpb.Header,
 	callbacks UpdateCallbacks,
-	goNamed func(Context, string, func(Context)),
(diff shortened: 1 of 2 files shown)


cretz (approved):
Makes sense. Technically could have used the root context to start the go named I think, but I think direct dispatcher coroutine creation is the same.

mmcshane:
> [...] could have used the root context to start the go named

Do you mean the workflow definition's rootCtx? That's what I was previously using but didn't work. I'd prefer to use GoNamed if you think we can figure out the right Context to use.

cretz:
My mistake, I thought rootCtx was the same ctx we gave to the user. Ignore me.

PR #888, opened 2022-08-22
Add versioning API methods to client

Sushisource:
## What was changed
Added top-level methods for the versioning API to the client.
Added worker flag to opt in to the feature

Go fmt also seemed to get a little more... aggressive? Some changes b/c of that.

## Why?
So we can use them in `tctl` and eventually release

## Checklist
Relates to https://github.com/temporalio/tctl/issues/250

1. Closes 

2. How was this tested:

3. Any docs updates needed?
Will be once ready for general consumption


client/client.go
```diff
@@ -371,11 +371,24 @@ type (
...
 
-		// ResetWorkflowExecution reset an existing workflow execution to WorkflowTaskFinishEventId(exclusive).
+		// ResetWorkflowExecution resets an existing workflow execution to WorkflowTaskFinishEventId(exclusive).
 		// And it will immediately terminating the current execution instance.
 		// RequestId is used to deduplicate requests. It will be autogenerated if not set.
 		ResetWorkflowExecution(ctx context.Context, request *workflowservice.ResetWorkflowExecutionRequest) (*workflowservice.ResetWorkflowExecutionResponse, error)
 
+		// UpdateWorkerBuildIdOrdering **IN DEVELOPMENT**
+		// Allows you to update the worker-build-id based version graph for a particular task queue. This is used in
+		// conjunction with workers who specify their build id and thus opt into the feature. For more, see: <doc link>
```
bergundy, line 381: Put a TODO here to add the link?

bergundy (approved):
Would be nice if your IDE didn't automatically reformat all of the docstrings

internal/worker_version_graph.go
```diff
@@ -0,0 +1,76 @@
...
+package internal
+
+import (
+	taskqueuepb "go.temporal.io/api/taskqueue/v1"
+	"go.temporal.io/api/workflowservice/v1"
+)
+
+// WorkerBuildIDVersionGraph is the response for Client.GetWorkerBuildIdOrdering and represents the graph
+// of worker build id based versions.
+type WorkerBuildIDVersionGraph struct {
```
cretz, line 32: May be able to cut down on the wording and/or simplify, e.g.:

* `WorkerBuildIDVersionGraph` to `WorkerVersionGraph`
* `WorkerVersionIDNode` to `WorkerVersion`
* `CompatibleLeaves` to `CompatibleVersions`
* `WorkerBuildID` to `BuildID`

internal/worker_version_graph.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return &WorkerVersionIDNode{
+		WorkerBuildID:        node.GetVersion().GetWorkerBuildId(),
+		PreviousCompatible:   fromProtoNode(node.PreviousCompatible),
+		PreviousIncompatible: fromProtoNode(node.PreviousIncompatible),
+	}
+}
+
+func fromProtoNodes(nodes []*taskqueuepb.VersionIdNode) []*WorkerVersionIDNode {
+	if nodes == nil {
```
cretz, line 68:
```suggestion
	if len(nodes) == 0 {
```

Catches nil and empty which is often the intent for things like this in Go

internal/worker_version_graph.go
```diff
@@ -0,0 +1,76 @@
...
+	// A pointer to the previous incompatible version (previous major version)
+	PreviousIncompatible *WorkerVersionIDNode
+}
+
+func fromProtoResponse(response *workflowservice.GetWorkerBuildIdOrderingResponse) *WorkerBuildIDVersionGraph {
+	if response == nil {
+		return nil
+	}
+	return &WorkerBuildIDVersionGraph{
+		CurrentDefault:   fromProtoNode(response.CurrentDefault),
```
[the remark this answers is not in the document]
Sushisource: It doesn't

client/client.go
```diff
@@ -371,11 +377,26 @@ type (
...
 		ResetWorkflowExecution(ctx context.Context, request *workflowservice.ResetWorkflowExecutionRequest) (*workflowservice.ResetWorkflowExecutionResponse, error)
 
+		// UpdateWorkerBuildIDOrdering **IN DEVELOPMENT**
+		// Allows you to update the worker-build-id based version graph for a particular task queue. This is used in
+		// conjunction with workers who specify their build id and thus opt into the feature. For more, see: <doc link>
+		//  - workerBuildId is required and indicates the build id being added to the version graph.
+		//  - previousCompatible may be empty, and if set, indicates an existing version the new id should be considered
+		//   compatible with.
+		//  - If becomeDefault is true, this new id will become the default version for new workflow executions.
+		UpdateWorkerBuildIDOrdering(ctx context.Context, taskQueue string, workerBuildId string, previousCompatible string, becomeDefault bool) error
```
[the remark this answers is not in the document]
Sushisource: It's not necessarily just for adding, as it can rewire existing nodes. I want to keep the names close to the actual API names as well, since we do that everywhere else

internal/worker_version_graph.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	taskqueuepb "go.temporal.io/api/taskqueue/v1"
+	"go.temporal.io/api/workflowservice/v1"
+)
+
+// UpdateWorkerBuildIDOrderingOptions is the input to Client.UpdateWorkerBuildIDOrdering.
+type UpdateWorkerBuildIDOrderingOptions struct {
+	// The task queue to update the version graph of.
+	taskQueue string
```
cretz, line 33: Users cannot set these unexported fields

cretz:
Per off-PR discussion, we're gonna see if we can get an integration test or two in here w/ real server impl. If the server isn't released w/ the impl fast enough we can feature-flag it off. Not much value in merging features that can't work.

test/worker_versioning_test.go
```diff
@@ -0,0 +1,87 @@
...
+func (ts *WorkerVersioningTestSuite) TearDownSuite() {
+	ts.Assertions = require.New(ts.T())
+	ts.client.Close()
+}
+
+func (ts *WorkerVersioningTestSuite) SetupTest() {
+	ts.taskQueueName = taskQueuePrefix + "-" + ts.T().Name()
+}
+
+func (ts *WorkerVersioningTestSuite) TestManipulateVersionGraph() {
```
cretz, line 58: I think the matching side of the test would be valuable. Specifically 3 versions of a worker+workflow that just return different strings, and then manipulate build IDs and make invocations and confirm proper workflow is reached based on default, compatibility, etc.

cretz:
Thanks for extracting some of those shared utilities out of the common integration test. The ultimate goal would be to use sdk-features for our integration tests but obviously we're not there yet. I just had one comment about another integration test that would be valuable.

FWIW I am getting these same 3+ minute CI timeouts in my other PR. I am researching, but I wonder if things are just slower now.

test/test_utils_test.go
```diff
@@ -207,3 +217,117 @@ func (s *keysPropagator) ExtractToWorkflow(ctx workflow.Context, reader workflow
...
 }
+
+func (ts *ConfigAndClientSuiteBase) InitConfigAndClient() error {
+	ts.config = NewConfig()
+	var err error
+	err = WaitForTCP(time.Minute, ts.config.ServiceAddr)
+	if err != nil {
+		return err
+	}
+	ts.client, err = client.Dial(client.Options{
```
cretz, line 228: This is causing a goroutine leak w/ an unclosed client in the integration test which I think is causing your test timeout. You are calling this in `SetupSuite` which creates this client in the field at the suite level, but that never happened before. The client is meant to be created _per test_ which it is in `SetupTest` and closed in `TearDownTest`.

Don't set this field in suite-level init, only in test-level init. And always make sure created clients are closed.

test/worker_versioning_test.go
Sushisource: Of course, but, that has to actually be implemented first :)

test/test_utils_test.go
Sushisource: Thanks

Sushisource:
Closing in favor of https://github.com/temporalio/sdk-go/pull/920

The changes are pretty much all the same, just slightly tweaked for the newer API

PR #889, opened 2022-08-23
Sync batch operations api

rodrigozhou:
## What was changed

Batch operations were added in API in https://github.com/temporalio/api/pull/197.
This PR syncs generated files affected by API changes.

## Why?

Unblock other changes that depends on recent API changes.

## Checklist

1. How was this tested:

`make`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,10 +15,13 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.8.0
-	go.temporal.io/api v1.11.0
+	go.temporal.io/api v1.11.1-0.20220804232755-77e12eef4f2d
 	go.uber.org/atomic v1.9.0
+	golang.org/x/net v0.0.0-20220822230855-b0a4917ee28c // indirect
+	golang.org/x/sys v0.0.0-20220818161305-2296e01440c6 // indirect
 	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac
 	golang.org/x/tools v0.1.10
+	google.golang.org/genproto v0.0.0-20220822174746-9e6da59bd2fc // indirect
 	google.golang.org/grpc v1.48.0
 	google.golang.org/protobuf v1.28.1
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
--- a/internal/cmd/generateinterceptor/main.go
+++ b/internal/cmd/generateinterceptor/main.go
@@ -76,7 +76,7 @@ import (
 
 	"google.golang.org/grpc"
 	
-	workflowservicepb "go.temporal.io/api/workflowservice/v1"
+	batchpb "go.temporal.io/api/batch/v1"
 	commonpb "go.temporal.io/api/common/v1"
 	historypb "go.temporal.io/api/history/v1"
 	commandpb "go.temporal.io/api/command/v1"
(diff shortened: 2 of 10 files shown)


internal/cmd/generateinterceptor/main.go
```diff
@@ -76,7 +76,7 @@ import (
 
 	"google.golang.org/grpc"
 	
-	workflowservicepb "go.temporal.io/api/workflowservice/v1"
+	batchpb "go.temporal.io/api/batch/v1"
```
rodrigozhou, line 79: I added the new API module here for the generator.

go.mod
```diff
@@ -15,10 +15,13 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.8.0
-	go.temporal.io/api v1.11.0
+	go.temporal.io/api v1.11.1-0.20220804232755-77e12eef4f2d
```
cretz, line 18: Is this implemented in the server? If not, I think we should wait until then to merge this. Also, I wonder if once implemented in the server we should tag the API since this is such a big change.

rodrigozhou: The server has dummy implementations returning `Unimplemented` error.

cretz: Why not wait until implemented there to add here? Can reference a local SDK perfect path during development.

rodrigozhou: I added those to unblock myself since I made changes to the API (same with this PR). Cc: @yux0

yiminc: The server implementation PR (for batch API) will be land soon, but we should not block this PR just because server currently return unimplemented as this API is not used anywhere anyway. Server also want to use UpsertMemo feature which is currently blocked on this PR.

rodrigozhou: Just to be clear, this PR only updates the generated files (`client/service_proxy.go` and `converter/grpc_interceptor.go`). It does not create/expose an API that call those unimplemented API endpoints.

cretz: > but we should not block this PR just because server currently return unimplemented

Why not? This is a user-facing API update to a non-tagged API version.

> Server also want to use UpsertMemo feature which is currently blocked on this PR.

This should not block development. I can show how to use `go.mod` replace clauses to reference a local form of this while you develop. If server wants to use `UpsertMemo`, that's blocked by #883 where I've also requested we have real tests to even confirm it works. All of this work can be done locally.

rodrigozhou: I meant that this PR blocks merging the `UpsertMemo` PR. I've done all the  local tests, and also almost done with writing the integration tests there.

cretz: :+1: Yeah, integration tests means things are working which works for me. I'd happily approve/merge. Curious, why does this PR block that one?

rodrigozhou: The changes to API related to batch operations affect the generated files `grpc_interceptor.go` and `service_proxy.go`. This PR just re-generates those files. It has nothing to do with `UpsertMemo`, but since my changes also affect those files, I thought it would better to have a PR doing an update for batch related stuff first instead of altogether.

cretz: Sure, you're call, updating API is updating API, could do in same PR or other.

So I can merge this. My concern is that we are giving new public API to end users that is not tagged and is not implemented. But as we've done with other API changes, we can just hope they don't try to use it yet.

cretz (approved):
Approving though I am not a fan of untagged, unimplemented API in SDK `master`, but :shrug:. You'll need to update the PR w/ latest master and then can merge after CI runs.

PR #893, opened 2022-08-29
Handle Client.Close on cloned clients

### LINKED ISSUE #886: independent lifecycles for cloned clients
## Expected Behavior

When using `NewClientFromExisting` from #881, the "parent" and "child" `Client`s have independent lifecycles from the point of view of client code, even though they share some resources. That is, I can call `Close` on only one or both of them, and resources will be cleaned up when appropriate.

This makes it easier to use in large programs since we don't have to essentially ban use of `Close`, we can let anyone who gets a cloned client treat it as if it were a fully-independent one.

## Actual Behavior

Calling `Close()` on either the parent or child will close the connection for both of them immediately.

  1.
  1.
  1.

## Specifications

  - Version: unreleased post 1.16
  - Platform:

cretz:
## What was changed

Use a counter to make sure only the last `Close()` of a shared-connection set of clients applies.

## Checklist

1. Closes #886

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -29,8 +29,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"reflect"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/pborman/uuid"
@@ -83,6 +85,10 @@ type (
 		excludeInternalFromRetry *uberatomic.Bool
 		capabilities             *workflowservice.GetSystemInfoResponse_Capabilities
 		capabilitiesLock         sync.RWMutex
+
+		// The pointer value is shared across multiple clients. If non-nil, only
+		// access/mutate atomically.
+		unclosedClients *int32
 	}
 
 	// namespaceClient is the client for managing namespaces.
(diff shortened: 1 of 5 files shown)


internal/internal_workflow_client.go
```diff
@@ -942,11 +948,25 @@ func (wc *WorkflowClient) ensureInitialized() error {
 
 // Close client and clean up underlying resources.
 func (wc *WorkflowClient) Close() {
-	if wc.conn == nil {
-		return
+	// If there's a set of unclosed clients, we have to decrement it and then
+	// set it to a new pointer of max to prevent decrementing on repeated Close
+	// calls to this client. If the count has not reached zero, this close call is
+	// ignored.
+	if wc.unclosedClients != nil {
```
dnr, line 955: this shouldn't ever be nil, right?

internal/internal_workflow_client.go
```diff
@@ -942,11 +948,25 @@ func (wc *WorkflowClient) ensureInitialized() error {
...
+		wc.unclosedClients = &maxUnclosedClients
+		// If there are any remaining, do not close
+		if remainingUnclosedClients > 0 {
+			return
+		}
 	}
-	if err := wc.conn.Close(); err != nil {
-		wc.logger.Warn("unable to close connection", tagError, err)
+
+	if wc.conn != nil {
```
dnr, line 966: when is this nil?

internal/internal_workflow_client_test.go
```diff
@@ -1402,3 +1404,45 @@ func serializeEvents(events []*historypb.HistoryEvent) *commonpb.DataBlob {
...
+	// Create primary client
+	server, err := startTestGRPCServer()
+	require.NoError(t, err)
+	defer server.Stop()
+	client, err := DialClient(ClientOptions{HostPort: server.addr})
+	require.NoError(t, err)
+	workflowClient := client.(*WorkflowClient)
+
+	// Confirm there is 1 unclosed client
+	require.EqualValues(t, 1, *workflowClient.unclosedClients)
```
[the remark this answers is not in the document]
cretz: This is in a test case. I can trust my lack of race issues in my own test case.

internal/internal_workflow_client.go
cretz: I cannot trust all places that create a client will set this. That includes tests and other code paths.

internal/internal_workflow_client.go
cretz: Test cases w/ mocks. This logic has not changed, I just moved the condition from the top of the function to down here.

dnr (approved):
sounds good to me

cretz:
This is reliably causing integration tests to go from 2 mins to 3. I am investigating...

internal/internal_workflow_client_test.go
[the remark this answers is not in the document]
Sushisource: I would do it just for consistency's sake. It's supposed to be an atomic int, it should be manipulated atomically even if Go lets you not do that. Trivial amount of extra typing.

Sushisource (approved):

internal/internal_workflow_client_test.go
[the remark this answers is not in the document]
cretz: I also access the pointers directly in a non-atomic way to compare they point to the same spot. But I will make the change.

PR #895, opened 2022-08-30
Added Channel.Empty and Channel.ReceiveWithTimeout

mfateev:
## What was changed
Added Len and ReceiveWithTimeout to the Channel interface

## Why?
Most applications listen to a signal with a timeout. This will significantly reduce the boilerplate code needed for this.

2. How was this tested:

Unit test included.

4. Any docs updates needed?

Samples and docs need to be updated.


internal/internal_workflow.go
```diff
@@ -763,6 +763,23 @@ func (c *channelImpl) Receive(ctx Context, valuePtr interface{}) (more bool) {
 
 }
 
+func (c *channelImpl) ReceiveWithTimeout(ctx Context, timeout time.Duration, valuePtr interface{}) (more bool, timedOut bool) {
```
cretz, line 766: This looks pretty easy to write as a user, and there's no equivalent for this in the Go spec for channels like there is for the other functions here. I might have used timer + selector, but it's basically the same thing.

But if this helper is worth changing the core interface, ok. Alternatively we could just have a package-level function helper for this so the interface is not changed, but it doesn't matter much.

internal/internal_workflow.go
```diff
@@ -784,6 +801,10 @@ func (c *channelImpl) ReceiveAsyncWithMoreFlag(valuePtr interface{}) (ok bool, m
 	}
 }
 
+func (c *channelImpl) Empty() bool {
```
cretz, line 804: I think a `Len() int` and a `Cap() int` here is more in line with https://pkg.go.dev/builtin#len and https://pkg.go.dev/builtin#cap respectively.

internal/internal_workflow.go
```diff
@@ -784,6 +801,10 @@ func (c *channelImpl) ReceiveAsyncWithMoreFlag(valuePtr interface{}) (ok bool, m
...
+	return c.closed || (len(c.buffer) == 0 && len(c.blockedSends) == 0 && c.recValue == nil)
```
cretz, line 805: In Go, technically a closed channel with items still in the buffer has a len > 0 and is not considered empty. Didn't check if `closed` field here represents when a user calls close or after the user has called closed _and_ all items are gone from the buffer.

internal/internal_workflow.go
```diff
@@ -763,6 +763,23 @@ func (c *channelImpl) Receive(ctx Context, valuePtr interface{}) (more bool) {
 
 }
 
+func (c *channelImpl) ReceiveWithTimeout(ctx Context, timeout time.Duration, valuePtr interface{}) (more bool, timedOut bool) {
+	// Channel operations are not interrupted by cancellation
```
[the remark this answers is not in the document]
mfateev: I wanted to be consistent with all other Channel operations which are not cancellable to be consistent with the native Go channels. But I agree that in this case as the timeout is involved cancellation does make sense. I'm going to fix this.

internal/internal_workflow.go
cretz: I don't have a strong opinion here, this works for me (though would prefer `Empty` to become `Len` if possible).

PR #911, opened 2022-09-15
Provide better custom logger example in README

ryanwalls:
## What was changed

Updated the README with a more helpful logging example.

## Why?

The current example did not make it clear how to "feed the logger" into Temporal.  Grabbed the better example from https://docs.temporal.io/go/how-to-set-clientoptions-in-go/#logger.

## Checklist

1. Closes 
n/a

2. How was this tested:

n/a

3. Any docs updates needed?

Done

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Although the Go SDK does not support most third-party logging solutions natively
 package main
 import (
 	"github.com/sirupsen/logrus"
+	"go.temporal.io/sdk/client"
 	logrusadapter "logur.dev/adapter/logrus"
 	"logur.dev/logur"
 )
@@ -32,6 +33,11 @@ import (
 func main() {
 	// feed this logger into Temporal
 	logger := logur.LoggerToKV(logrusadapter.New(logrus.New()))
+	clientOptions := client.Options{
+		Logger: logger,
+	}
+	temporalClient, err := client.Dial(clientOptions)
+	// ...
 }
 ```
 


README.md
```diff
@@ -25,13 +25,19 @@ Although the Go SDK does not support most third-party logging solutions natively
 package main
 import (
 	"github.com/sirupsen/logrus"
+	"go.temporal.io/sdk/client"
```
cretz, line 28: I wonder if we should just remove this section (notice there aren't other examples/docs here in the README) and that way people will look at https://github.com/temporalio/samples-go instead (specifically https://github.com/temporalio/samples-go/tree/main/zapadapter).

Thoughts?

cretz (approved):

README.md
cretz: > and a very austere README may send the wrong message.

The README is already very tiny compared to docs.

> Perhaps some of the sections can be replaced with links to other docs, but keep the section titles?

There aren't really any sections of note besides this one.

I can merge this and we can just table the idea of altering the README more drastically. Sound good?

ryanwalls: Merging for now is fine.  Or you can just replace the one section we are talking about with a link to the docs for the logger.  I'm good either way.

cretz: Can merge. Can you update the branch w/ master? and get CI passing?

cretz:
Merged, thanks!

PR #915, opened 2022-09-20
Do not call ResetSticky on force eviction

yiminc:
There are 2 benefits of calling ResetSticky: 
1) so that next workflow task will contain full history, and worker won’t need to poll for the missing history separately. 
2) there will be no 5s delay if no sticky worker on that sticky queue.

For #1, to gain that benefit of one less GetHistory call, we pay one extra call to ResetSticky which is a CAS operation and is much expensive than GetHistory.
For #2, matching service will return NoStickyWorkerAvailable and there will be no 5s delay if sticky worker is gone.


cretz (approved):
I admit to not understanding the manual reset sticky, so I trust your judgement here.

PR #917, opened 2022-09-22
Support backing off more for resource exhausted on pollers

cretz:
## What was changed

Added backoff mimicking server's logic at https://github.com/temporalio/temporal/blob/v1.18.0/common/backoff/retry.go#L198-L208 for resource exhausted. It's obviously not a good algorithm to optionally trigger stateful backoffs and is _not_ the algorithm used by TypeScript at https://github.com/temporalio/sdk-typescript/pull/879. But the goal of this issue was to solve the immediate need without much refactoring. Also to that point, the pollers were not written to have the backoff times easily testable and it was not worth rewriting pollers, so the tests only confirm the secondary retrier works, not whether the poller part does.

## Why?

Asked to in https://github.com/temporalio/sdk-features/issues/96

If we wanted to do this on non-poll gRPC calls, it would be a completely different (they use the gRPC retrier)

### DIFF
--- a/internal/common/backoff/retry.go
+++ b/internal/common/backoff/retry.go
@@ -45,8 +45,10 @@ type (
 	// requests due to out-of-quota or server busy errors.
 	ConcurrentRetrier struct {
 		sync.Mutex
-		retrier      Retrier // Backoff retrier
-		failureCount int64   // Number of consecutive failures seen
+		retrier                 Retrier // Backoff retrier
+		secondaryRetrier        Retrier
+		failureCount            int64 // Number of consecutive failures seen
+		includeSecondaryRetrier bool
 	}
 )
 
@@ -63,6 +65,16 @@ func (c *ConcurrentRetrier) throttleInternal(doneCh <-chan struct{}) time.Durati
 	c.Lock()
 	if c.failureCount > 0 {
 		next = c.retrier.NextBackOff()
+		// If secondary is included, use the greatest of the two (which also means
+		// if one is "done", which is -1, the one that's not done is chosen)
+		if c.includeSecondaryRetrier {
+			c.includeSecondaryRetrier = false
+			if c.secondaryRetrier != nil {
+				if secNext := c.secondaryRetrier.NextBackOff(); secNext > next {
+					next = secNext
+				}
+			}
+		}
 	}
 	c.Unlock()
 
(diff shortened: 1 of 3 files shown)


internal/common/backoff/retry.go
```diff
@@ -63,6 +65,16 @@ func (c *ConcurrentRetrier) throttleInternal(doneCh <-chan struct{}) time.Durati
 	c.Lock()
 	if c.failureCount > 0 {
 		next = c.retrier.NextBackOff()
+		// If secondary is included, use the greatest of the two (which also means
+		// if one is "done", which is -1, the one that's not done is chosen)
+		if c.includeSecondaryRetrier {
+			c.includeSecondaryRetrier = false
+			if c.secondaryRetrier != nil {
+				if secNext := c.secondaryRetrier.NextBackOff(); secNext > next {
```
bergundy, line 73: This ended up being exactly like the server code where the attempt isn't shared between the 2 retriers.
I think this is fine, it's just different from TS.

internal/common/backoff/retry.go
```diff
@@ -63,6 +65,16 @@ func (c *ConcurrentRetrier) throttleInternal(doneCh <-chan struct{}) time.Durati
...
+					next = secNext
+				}
+			}
+		}
```
bergundy, line 77: This turned out hard to parse IMHO

bergundy: Should the secondary retrier attempt be incremented even if it wasn't used? This would be consistent with the behavior defined in the issue and what I did in TS.

I don't think it's critical so I'm approving the PR.

bergundy (approved):

internal/common/backoff/retry.go
cretz: > Should the secondary retrier attempt be incremented even if it wasn't used?

I specifically asked at https://github.com/temporalio/sdk-features/issues/96#issuecomment-1201201389:

> Can you clarify the logic here when, after resource exhausted retry policy is set, what we do when we receive non-resource-exhausted error? Should we completely reset the retry policy? Same question if it's a non-resource-exhausted error and then a resource exhausted error.

And was pointed to https://github.com/temporalio/temporal/blob/3182dc2cde75e3e49fccb1de2905db5d30db58ab/common/backoff/retry.go#L204-L206 which I emulated here.

bergundy: My comment describes different logic.
As mentioned, I don't think any of this matters that much as long as the SDK backoffs "more" on resource exhausted.
Just wanted to keep the logic consistent between the SDKs.

cretz: Your approach would require larger Go alterations and is in conflict with @yiminc's. If between y'all your approach is the one wanted for Go, we can apply it.

PR #921, opened 2022-09-28
Datadog tracing implementation

MarcosCela:
## What was changed

Added a new tracing implementation that makes use of datadog. This makes use of the field `options.IdempotencyKey` to generate the `SpanID` to be consistent across replays.

## Other details

To be able to pass the tests, we need to change how we originally implemented this:
```diff
- s := tracer.StartSpan("temporal."+options.Operation, startOpts...)
+ s := tracer.StartSpan(options.Operation+":"+options.Name, startOpts...)
```

Our original implementation makes traces look like this:

image

In addition to that, we also replace tags:
```go
// Add tags to start options
	for k, v := range options.Tags {
		// Display Temporal tags in a nested group in Datadog APM
		tagKey := strings.Replace(k, "temporal", "temporal.", 1)
		startOpts = append(startOpts, tracer.Tag(tagKey, v))
	}
```

This is because Datadog will group these tags if they are dot-delimited, and group them together, which makes things more compact and easier to find.


contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func (tracerImpl) StartSpan(options *interceptor.TracerStartSpanOptions) (interceptor.TracerSpan, error) {
+	startOpts := []tracer.StartSpanOption{
+		tracer.ResourceName(options.Name),
+		tracer.StartTime(options.Time),
+	}
+
+	// Set a deterministic span ID for workflows which are long-running and cross process boundaries
+	if options.Operation == "RunWorkflow" {
```
MarcosCela, line 89: This might break on operation renames or if operations get a prefix.

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil {
+			return nil, err
+		}
+		startOpts = append(startOpts, tracer.ChildOf(parentTrace))
+	}
+
+	// Add tags to start options
+	for k, v := range options.Tags {
+		// Display Temporal tags in a nested group in Datadog APM
+		tagKey := strings.Replace(k, "temporal", "temporal.", 1)
```
MarcosCela, line 126: We do this because datadog groups by a "dot". This way all tags that share the "temporal." prefix are grouped in the interface. It could break if temporal ever adds its own "grouping"

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+		startOpts = append(startOpts, tracer.Tag(tagKey, v))
+	}
+
+	// Start and return span
+	s := tracer.StartSpan("temporal."+options.Operation, startOpts...)
```
MarcosCela, line 131: This is currently not compliant with what the tests expect, should be make the tests more generic so this is supported?

contrib/datadog/tracing/interceptor.go
jlegrone: Perhaps we could define operation names (or at least this one) with a public const in the `go.temporal.io/sdk/interceptor` package?

contrib/datadog/tracing/interceptor.go
jlegrone: We'd also want to avoid adding the `temporal` prefix if the interceptor ever adds support for "custom" user-defined operations.

jlegrone: Perhaps we could add an optional `CoreOperationPrefix() string` method to `interceptor.Tracer` to be automatically added to any Temporal specific spans? I think it would be nicest if all tracers used the same operation names, but I'm pretty sure adding a prefix everywhere would be considered a breaking change.

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+	case *tracerSpan:
+		parent = opParent.Context()
+	case *tracerSpanCtx:
+		parent = opParent
+	default:
+		return nil, fmt.Errorf("unrecognized parent type %T", opParent)
+	}
+	if parent != nil {
+		// Convert the parent context into a full trace (required, otherwise
+		// the implementation of StartSpan() will NOT respect ChildOf()
```
jlegrone, line 113: Nit: missing closing `)`. A new sentence avoids any parens.

```suggestion
		// Convert the parent context into a full trace.
		// This is required, otherwise the implementation of StartSpan()
		// will NOT respect ChildOf().
```

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+	// Set a deterministic span ID for workflows which are long-running and cross process boundaries
+	if options.Operation == "RunWorkflow" {
+		rid, ok := options.Tags["temporalRunID"]
+		if !ok {
+			return nil, fmt.Errorf("missing required tag `temporalRunID` on `RunWorkflow` span")
+		}
+		h := fnv.New64()
+		// Write() always writes all bytes and never fails; the count and error result are for implementing io.Writer.
+		_, _ = h.Write([]byte(fmt.Sprintf("workflow:%s", rid)))
+		startOpts = append(startOpts, tracer.WithSpanID(h.Sum64()))
```
jlegrone, line 97: We should add a test to ensure that the resulting span ID here is stable. It's a breaking change if the way we compute the span ID ever changes since child spans will become orphaned.

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// Set a deterministic span ID for workflows which are long-running and cross process boundaries
+	if options.Operation == "RunWorkflow" {
+		rid, ok := options.Tags["temporalRunID"]
+		if !ok {
+			return nil, fmt.Errorf("missing required tag `temporalRunID` on `RunWorkflow` span")
+		}
+		h := fnv.New64()
+		// Write() always writes all bytes and never fails; the count and error result are for implementing io.Writer.
+		_, _ = h.Write([]byte(fmt.Sprintf("workflow:%s", rid)))
```
jlegrone, line 96: Should we take workflow id & namespace into account here as well?

cretz:
Let me know when y'all want me to review and if/when you want me to approve the CI run

contrib/datadog/tracing/interceptor.go
MarcosCela: That would depend on the generation of the `runId`. I am guessing it is being randomly generated somewhere so collision is unlikely, but adding `hash(workflowId+namespace+runId)` would guarantee that we are unique even if the `runId` generation changes.

contrib/datadog/tracing/interceptor.go
MarcosCela: Added a test, let me know what you think!

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,205 @@
...
+		return nil
+	}
+	return &tracerSpan{Span: span}
+}
+
+func (tracerImpl) ContextWithSpan(ctx context.Context, span interceptor.TracerSpan) context.Context {
+	return tracer.ContextWithSpan(ctx, span.(*tracerSpan).Span)
+}
+
+func genSpanID(runId string) uint64 {
```
jlegrone, line 82: Just in case there end up being multiple spans which require deterministic IDs, and also to avoid user code for custom spans accidentally colliding, I think we should include a bit more than just the run ID as input to the hashing function. Perhaps something like `<workflow_namespace>:<workflow_id>:<workflow_run_id>:<trace_operation_name>` would be more future proof?

contrib/datadog/tracing/interceptor.go
jlegrone: This can be reworked to not inspect the operation name now that #931 is merged.

contrib/datadog/tracing/interceptor.go
jlegrone: This is also no longer a concern as a result of #931.

MarcosCela: Updated to use it!

contrib/datadog/tracing/interceptor.go
MarcosCela: We still have to inspect the operation, otherwise test fail with stack overflow

cretz:
@MarcosCela - is this ready for review and CI run?

MarcosCela:
> 



> @MarcosCela - is this ready for review and CI run?

Discussion is still open, but changes should be minimal from this point.

cretz (commented):
Looks great! Minor comment or two. I will approve to run.

Also, while it is not required, you are more than welcome to add an integration test in the `test/integration_test.go` file if you want. We only have one for OpenTelemetry not OpenTracing and this definitely doesn't need one, I just wanted to let it be known you are allowed to if you wanted.

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+
+	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/log"
+	"go.temporal.io/sdk/workflow"
+)
+
+var (
+	_ interceptor.Tracer     = tracerImpl{}
+	_ interceptor.TracerSpan = &tracerSpan{}
+)
```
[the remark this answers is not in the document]
MarcosCela: I will remove them, it's a very personal thing of mine to show that a given struct is implementing a given interface

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+	return &tracerSpan{Span: s}, nil
+}
+
+func (tracerImpl) GetLogger(logger log.Logger, ref interceptor.TracerSpanRef) log.Logger {
+	spanRef, ok := ref.(*tracerSpan)
+	if !ok {
+		logger.Error(fmt.Sprintf("Error injecting TraceID in GetLogger: TracerSpanRef is type %T, expected *tracerSpan", ref))
+		return logger
+	}
+	return log.With(logger, "dd.trace_id", spanRef.TraceID(), "dd.span_id", spanRef.SpanID())
```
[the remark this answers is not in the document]
MarcosCela: Yup, we can definitely make it opt-out, but this is the way that datadog can link traces with logs.

contrib/datadog/tracing/go.mod
```diff
@@ -0,0 +1,10 @@
+module go.temporal.io/sdk/contrib/datadog/tracing
```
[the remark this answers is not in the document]
MarcosCela: We will most likely add `go.temporal.io/sdk/contrib/datadog/metrics` in the future!

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+var (
+	_ interceptor.Tracer     = tracerImpl{}
+	_ interceptor.TracerSpan = &tracerSpan{}
+)
+
+// NewInterceptor creates an interceptor for setting on client options
+// that implements Datadog tracing for workflows.
+func NewInterceptor() interceptor.Tracer {
```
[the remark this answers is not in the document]
MarcosCela: Added options!

contrib/datadog/tracing/go.mod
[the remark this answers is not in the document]
cretz: Right, so you may want to move `go.mod` up a directory so all `datadog` integration is one module. Unless there's a good dependency reason for not doing so?

contrib/datadog/tracing/interceptor.go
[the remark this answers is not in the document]
cretz: It's a common thing for many. What I mean about these not being needed is not that you don't want this assertion, it's that the assertion is done by `return tracerImpl{` and `return &tracerSpan{`.

(unresolving because adding comments on resolved threads are often not read)

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// DisableLogTraceLinking can be set to disable the automatic addition of "dd.trace_id" and "dd.span_id" fields to the logger
+	// provided by this interceptor.
+	DisableLogTraceLinking bool
+}
+
+// NewInterceptor creates an interceptor for setting on client options
+// that implements Datadog tracing for workflows.
+func NewInterceptor(opts InterceptorOptions) interceptor.Tracer {
+	return tracerImpl{
```
cretz, line 58: Doesn't really matter for a few bools, but no benefit in not making this a pointer and changing your receivers to pointers. While you don't keep mutable state now, you're not saving much w/ copies for each call.

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,247 @@
...
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/log"
+	"go.temporal.io/sdk/workflow"
+)
+
+// InterceptorOptions are options provided to NewInterceptor
+type InterceptorOptions struct {
```
cretz, line 43: I think you should also let the user customize the header key and context key like our other tracer implementations

cretz (commented):
Looks great, just one request to add a couple more options and I think we're ready to merge (also commented on whether tracer can be a pointer, but that change is optional).

contrib/datadog/tracing/interceptor.go
MarcosCela: Done!

cretz (approved):

cretz:
@MarcosCela - one more thing, can you merge `master` into your branch?

contrib/datadog/tracing/go.mod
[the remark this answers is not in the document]
MarcosCela: Moved the file, you are absolutely right!

MarcosCela:
@cretz Updated from requested changes again, many thanks!!! 🙇🏼

cretz (approved):
Looks great! Will wait for CI and @jlegrone's approval before merging.

cretz:
@MarcosCela - Can you update the branch?

MarcosCela:
@cretz updated!

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,262 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package tracing provides DataDog tracing utilities
```
jlegrone, line 23: Branding guidelines nit 😉 

```suggestion
// Package tracing provides Datadog tracing utilities
```

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,262 @@
...
+	"go.temporal.io/sdk/workflow"
+)
+
+// InterceptorOptions are options provided to NewInterceptor
+type InterceptorOptions struct {
+	// DisableSignalTracing can be set to disable signal tracing.
+	DisableSignalTracing bool
+
+	// DisableQueryTracing can be set to disable query tracing.
+	DisableQueryTracing bool
```
jlegrone, line 48: Nit: I would rather have a generic `TracerOptions` struct in common across all the tracing implementations since we're ultimately only passing these through to the generic Tracer in the end and don't act on them at all in Datadog specific implementation.

Or better yet, perhaps we could add the ability to customize tracer options from https://pkg.go.dev/go.temporal.io/sdk@v1.17.0/interceptor#NewTracingInterceptor and simply return defaults from the Datadog tracer's `Options` method.

This comment isn't a blocker though; the pattern here seems consistent with the existing tracers.

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// DisableLogTraceLinking can be set to disable the automatic addition of "dd.trace_id" and "dd.span_id" fields to the logger
+	// provided by this interceptor.
+	DisableLogTraceLinking bool
```
jlegrone, line 52: Nit: I think we should remove this option; if a customer is both emitting logs and using traces, I think these attributes are likely to be cheap enough and useful enough to be relevant in all cases. Even if you're using a non-Datadog logging product, including the trace ID is important to being able to look up the request in Datadog APM.

I'm happy to re-evaluate adding this option later however based on customer feedback.

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// SpanContextKey is the context key used for internal span tracking. If not
+	// set, this defaults to an internal key (recommended).
+	SpanContextKey interface{}
+
+	// HeaderKey is the Temporal header field key used to serialize spans. If
+	// empty, this defaults to the one used by all SDKs (recommended).
+	HeaderKey string
```
jlegrone, line 60: What is the use case for making these configurable? I think we can always add higher level fields to extract/inject spans from carriers if necessary.

I would prefer not to include these options, since this makes it difficult/impossible to ensure tracing compatibility between language SDKs or even differently configured worker processes.

jlegrone: Discussed with @cretz on Slack, and we're good to remove these options for now.

The use case was for handling custom spans in workflow code. I would like to tackle this with a higher level API that generates deterministic span IDs to make it safe to trace long-running pieces of workflow code.

contrib/datadog/tracing/interceptor.go
cretz: I think that would assume a bit too much similarity amongst tracers (even if that does seem to be the case today). Granted that is the case now, but from a user-facing POV having DataDog-specific values in here can work too in the future when you need them.

Specifically for `DisableSignalTracing` and `DisableQueryTracing` they were added as a safety measure when that tracing was turned on by default. I'd be ok if y'all didn't want to expose these.

I would ask though, even if y'all get rid of every option, to have an empty struct for future proofing (we prefer the struct-as-config model over the options-type vararg pattern).

contrib/datadog/tracing/interceptor.go
cretz: I believe the use case was to support multiple tracing implementations at once. For example if you wanted to have OpenTracing and OpenTelemetry, you'd need to disambiguate the headers. But I have no problem with y'all removing this from these options.

MarcosCela:
Updated as per comments, what do you all think?

cretz (approved):
LGTM

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,237 @@
...
+
+	// Add tags to start options
+	for k, v := range options.Tags {
+		// Display Temporal tags in a nested group in Datadog APM
+		tagKey := "temporal." + strings.TrimPrefix(k, "temporal")
+		startOpts = append(startOpts, tracer.Tag(tagKey, v))
+	}
+
+	// Start and return span
+	s := tracer.StartSpan(options.Operation+":"+options.Name, startOpts...)
```
jlegrone, line 168: We don't include `options.Name` in the span's operation type in our internal implementation. Was there a reason to do this here? It's already set as the span's resource name at the top of the function.

Internally we also manually add a `temporal.` prefix in order to contextualize spans (names like `ExecuteWorkflow` are very generic when paired next to tens or hundreds of other span types).

```suggestion
	s := tracer.StartSpan("temporal."+options.Operation, startOpts...)
```

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,237 @@
...
+	// Add parent span to start options
+	var parent ddtrace.SpanContext
+	switch opParent := options.Parent.(type) {
+	case nil:
+	case *tracerSpan:
+		parent = opParent.Context()
+	case *tracerSpanCtx:
+		parent = opParent
+	default:
+		return nil, fmt.Errorf("unrecognized parent type %T", opParent)
```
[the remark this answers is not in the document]
MarcosCela: I think this is because of how the datadog library handles a parent trace depending if it comes from network or it comes from memory (somewhere else in the same code).
You are absolutely right for the default case, we should just log it and have a `nil` parent in that case.

contrib/datadog/tracing/interceptor.go
MarcosCela: The only reason is that the current testing suite expects the operation names to follow a given pattern. I do agree that we should detach the naming pattern or at least make it configurable somehow, and the same goes for adding a prefix (that should default to temporal).

contrib/datadog/tracing/interceptor.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+		startOpts = append(startOpts, tracer.ChildOf(parentTrace))
+	}
+
+	// Add tags to start options
+	for k, v := range options.Tags {
+		// Display Temporal tags in a nested group in Datadog APM
+		tagKey := "temporal." + strings.TrimPrefix(k, "temporal")
+		startOpts = append(startOpts, tracer.Tag(tagKey, v))
+	}
```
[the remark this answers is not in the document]
cretz: The `Tags` here was only ever meant to be tags set by Temporal from the abstraction, not user tags. In fact, I was not expecting end users to ever have to reference that tracing abstraction at all.

We need to clarify/improve how users can start their own spans. But the idea was to use library-native ways. This is what happens in activities (e.g. https://pkg.go.dev/github.com/opentracing/opentracing-go#StartSpanFromContext works just fine I believe). We just don't have anything for workflows.

I wonder if each contrib impl would have its own `StartSpanFromWorkflowContext` or similar. This lets each implementation support its own options. I'd accept API options for how custom spans in a workflow might look, but today I guess they're kinda absent in the Go SDK. I think in the meantime you can assume `Tags` here is only from the abstraction not from users.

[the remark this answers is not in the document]
jlegrone: @MarcosCela mind just adding a TODO or XXX comment here to warn about not including the `temporal.` prefix if/when custom span support is added?

contrib/datadog/tracing/interceptor.go
jlegrone: I think we'll need to update the test suite then, because separately indexing span types like `temporal.RunWorkflow` and [dynamic resources](https://docs.datadoghq.com/tracing/glossary/#resources) like `MyWorkflowName` is key to how we expect customers to set up generic monitoring for things like per-workflow-type error rate or execution latency.

Let's see if we can customize the tests to enforce the current behavior for the other tracing implementations, but allow a custom prefix and span names for Datadog. We'll need another review from @cretz on that of course.

cretz: :+1: I would support altering the shared test in a way that lets this work while still letting the test control what spans its asserting presence of and letting the other tests still pass.

MarcosCela:
Hey @cretz @jlegrone I did some work on the issue of being able to use custom naming without breaking the test, I am happy to do another PR JUST for that if you think this is good. My first approach was in:
https://github.com/temporalio/sdk-go/pull/921/commits/ba2bc7587a3f86059321094ae95b02022f610820
But I am not really a fan of how convoluted it is,  so I iterated (see last commit) an ultimately ended up adding a method to the Tracer interface. Let me know what you think.

interceptor/tracing_interceptor.go
```diff
@@ -73,6 +73,9 @@ type Tracer interface {
 	// GetLogger returns a log.Logger which may include additional fields in its
 	// output in order to support correlation of tracing and log data.
 	GetLogger(log.Logger, TracerSpanRef) log.Logger
+	// SpanName can be used to give a custom name to a Span according to the input TracerStartSpanOptions,
+	// or the decision can be deferred to the BaseTracer implementation.
+	SpanName(options *TracerStartSpanOptions) string
```
cretz, line 78: :+1: This works for me

MarcosCela:
This has reached a point where we are all happy with it, we will test it internally to verify that everything is 👌🏼 , and then I believe we are good to merge!  Thanks a ton for all the feedback.

cretz:
@MarcosCela - Thanks! I will wait for @jlegrone's stamp of approval.

cretz:
@MarcosCela - sorry this has sat and may need a master merge. Also may have to notify @jlegrone to approve.

MarcosCela:
> @MarcosCela - sorry this has sat and may need a master merge. Also may have to notify @jlegrone to approve.

Yep, we still need to verify it locally to ensure that everything is OK. We have been a little bit slower due to needing to update first & vacations around the holidays. 😢

MarcosCela:
Hi again, we have tested this locally and everything looks fine! If it's fine for you, we can move forward with the merge.

jlegrone (approved):
lgtm, sorry for the delay! We have tested and are prepared to switch over to this implementation internally. 🎉

cretz:
@MarcosCela - Looks like a go.sum update needed in the contrib.

cretz:
@MarcosCela - waiting for CI, but after that I think we're ready to merge. That ok with you?

MarcosCela:
> @MarcosCela - waiting for CI, but after that I think we're ready to merge. That ok with you? We can then make a `contrib/datadog/v0.1.0` tag.

Yep, we can live without the tag for a while and wait for it to be GA if you prefer. Thanks again for babysitting this for so long!

cretz:
@MarcosCela - No prob! Just hit the merge-master button to get your branch up to date since there were no conflicts. I will merge if/when CI passes. Thanks!

cretz:
@MarcosCela - I am sorry to say there's another one of those `go.sum` issues. Can you pull your latest branch (it'll have changes since I merged master into it) and update go.sum in this contrib folder (I think `make update-go-sum` will also do it). Thanks.

MarcosCela:
> @MarcosCela - I am sorry to say there's another one of those `go.sum` issues. Can you pull your latest branch (it'll have changes since I merged master into it) and update go.sum in this contrib folder (I think `make update-go-sum` will also do it). Thanks.

Done!

cretz:
Merged, thanks!

MarcosCela:
> Merged, thanks! Will look into adding a tag at some point soon (notify us if we forget :-))
> 
> Also, we have to add a vanity URL update to our generator. Doing that shortly.

Thank you very much!! 🙇🏼

PR #931, opened 2022-10-11
Add IdempotencyKey start span option

jlegrone:
## What was changed

This PR adds a new `IdempotencyKey` field to `TracerStartSpanOptions` and sets this field on spans created from the `WorkflowInboundInterceptor` tracing implementation.

## Why?

This option can be used to compute a deterministic span ID automatically for tracing implementations which support specifying an explicit span ID. This is being proposed as an alternative to the way we're computing the `RunWorkflow` span ID in #921.

This approach might also be used to provide a safe interface for handling custom spans created by end users in workflow code.

## Checklist

1. Closes

N/A

3. How was this tested:

See new unit test for creating idempotency keys. There are no existing tracing implementations in tree which support explicit spans, but we may be able to add additional tests in #921 if this is merged.

4. Any docs updates needed?

No

### DIFF
--- a/interceptor/tracing_interceptor.go
+++ b/interceptor/tracing_interceptor.go
@@ -139,6 +139,17 @@ type TracerStartSpanOptions struct {
 	// ToHeader is used internally, not by tracer implementations, to determine
 	// whether the span should be placed on the Temporal header.
 	ToHeader bool
+
+	// IdempotencyKey may optionally be used by tracing implementations to generate
+	// deterministic span IDs.
+	//
+	// This is useful in workflow contexts where spans may need to be "resumed" before
+	// ultimately being reported. Generating a deterministic span ID ensures that any
+	// child spans created before the parent span is resumed do not become orphaned.
+	//
+	// IdempotencyKey is not guaranteed to be set for all operations; Tracer
+	// implementations MUST therefore ignore zero values for this field.
+	IdempotencyKey string
 }
 
 // TracerSpanRef represents a span reference such as a parent.
@@ -197,7 +208,7 @@ func (t *tracingInterceptor) InterceptWorkflow(
 	ctx workflow.Context,
 	next WorkflowInboundInterceptor,
 ) WorkflowInboundInterceptor {
-	i := &tracingWorkflowInboundInterceptor{root: t}
+	i := &tracingWorkflowInboundInterceptor{root: t, info: workflow.GetInfo(ctx)}
 	i.Next = next
 	return i
 }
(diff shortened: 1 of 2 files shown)


interceptor/tracing_interceptor.go
```diff
@@ -357,7 +368,20 @@ func (t *tracingActivityInboundInterceptor) ExecuteActivity(
...
+	root        *tracingInterceptor
+	spanCounter uint16
+	info        *workflow.Info
+}
+
+// newIdempotencyKey returns a new idempotency key by incrementing the span counter and interpolating
+// this new value into a string that includes the workflow namespace/id/run id and the interceptor type.
+func (t *tracingWorkflowInboundInterceptor) newIdempotencyKey() string {
+	t.spanCounter++
+	return fmt.Sprintf("WorkflowInboundInterceptor:%s:%s:%s:%d",
```
jlegrone, line 380: We might want to consider a shared counter across all workflow interceptors instead. However including the "interceptor type" prefix ensures that we can add an independent counter to `tracingWorkflowOutboundInterceptor` later if we want to without worrying about idempotency key collisions.

interceptor/tracing_interceptor.go
```diff
@@ -139,6 +139,17 @@ type TracerStartSpanOptions struct {
 	// ToHeader is used internally, not by tracer implementations, to determine
 	// whether the span should be placed on the Temporal header.
 	ToHeader bool
+
+	// IdempotencyKey may optionally be used by tracing implementations to generate
+	// deterministic span IDs.
+	//
+	// This is useful in workflow contexts where spans may need to be "resumed" before
+	// ultimately being reported. Generating a deterministic span ID ensures that any
+	// child spans created before the parent span is resumed do not become orphaned.
```
jlegrone, line 148: Open to ideas on wording here 😅

interceptor/tracing_interceptor.go
cretz: Yeah, I think what is present now is reasonable. It is much easier to customize/expose this later if this abstraction is doubly used than to overthink it now.

interceptor/tracing_interceptor.go
```diff
@@ -357,7 +368,20 @@ func (t *tracingActivityInboundInterceptor) ExecuteActivity(
...
+}
+
+// newIdempotencyKey returns a new idempotency key by incrementing the span counter and interpolating
+// this new value into a string that includes the workflow namespace/id/run id and the interceptor type.
+func (t *tracingWorkflowInboundInterceptor) newIdempotencyKey() string {
+	t.spanCounter++
+	return fmt.Sprintf("WorkflowInboundInterceptor:%s:%s:%s:%d",
+		t.info.Namespace,
+		t.info.WorkflowExecution.ID,
+		t.info.WorkflowExecution.RunID,
```
cretz, line 383: Should be noted that technically this value changes if a workflow is reset to a previous task. However, not only do we not appear to provide the "original run ID", but I think this is probably the correct behavior anyways.

interceptor/tracing_interceptor_internal_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+func TestIdempotencyKeyGeneration(t *testing.T) {
```
cretz, line 11: Since this key is not used currently, I support no integration test with it. May want one to confirm counter is workflow-specific and replay safe though in the impl that does use it.

cretz (approved):

interceptor/tracing_interceptor.go
cretz: LGTM

interceptor/tracing_interceptor.go
jlegrone: Yep I like to think of workflow reset as a fancy continue-as-new. No problem with having a new trace for a new workflow run.

bergundy: You can end up with orphaned spans this way if a wrapping workflow span crosses a reset point

PR #941, opened 2022-10-31
Adopt changes in api repo for upcomming sync workflow update functionality

alexshtin:
- Update workflow
- Update api

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


alexshtin:
Wow, this is something very old.

Issue #871, opened 2022-07-30
High level API for schedules


bergundy:
Not sure what the Go client has now, I guess some work was already done on this by @dnr since the server and tctl use the Go SDK client but we should review this API before handing it off to users.

PR #943, opened 2022-11-01
Add schedule API

Quinn-With-Two-Ns:
## What was changed
Add schedule API

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/871

2. How was this tested:

3. Any docs updates needed?


internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+
+		// Args - Arguments to pass to the workflow.
+		Args []interface{}
+
+		// WorkflowOptions - Options to use when launching the workflow.
+		WorkflowOptions StartWorkflowOptions
+	}
+
+	// ScheduleWorkflowActionDescription implements ScheduleAction to launch a workflow.
+	ScheduleWorkflowActionDescription struct {
```
Quinn-With-Two-Ns, line 814: We need a separate types for the creation and description of a workflow action because of the `Workflow `, `Args`, and the `Memo ` and `SearchAttributes ` of the workflow

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		// Workflow - The business identifier of the workflow execution.
+		WorkflowID string
+
+		// WorkflowType - The type of the wokrflow.
+		WorkflowType WorkflowType
+
+		// Args - Arguments of the workflow
+		Args *commonpb.Payloads
```
Quinn-With-Two-Ns, line 822: @cretz we talked about using encoded values for the description of the workflow arguments but I found they couldn't be converted back to payload for the `Update` function.

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		Memo map[string]interface{}
+
+		// SearchAttributes - Optional indexed info that can be used in query of List schedules APIs (only
+		// supported when Temporal server is using ElasticSearch). The key and value type must be registered on Temporal server side.
+		// Use GetSearchAttributes API to get valid key and corresponding value type.
+		SearchAttributes map[string]interface{}
+	}
+
+	// ScheduleWorkflowExecution contains details on a workflows execution stared by a schedule.
+	ScheduleWorkflowExecution struct {
```
Quinn-With-Two-Ns, line 917: I took this type from the proposed TS API. We could replace this type with `WorkflowExecution` . The difference is this type highlights the runID is the original run ID.

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+
+		// SchedulePolicies - this schedules policies
+		Policy *SchedulePolicies
+
+		// State - this schedules state
+		State *ScheduleState
+	}
+
+	// ScheduleUpdate describes the desired new schedule from ScheduleHandle.Update.
+	ScheduleUpdate struct {
```
Quinn-With-Two-Ns, line 1022: I added another struct around `Schedule` incase we wanted to allow updating other parts of a schedule in the future. I was specifically thinking the `Memo` and `SearchAttributes`

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		// Schedule - New schedule to replace the existing schedule with
+		Schedule *Schedule
+	}
+
+	// ScheduleUpdateOptions configure the parameters for updating a schedule.
+	ScheduleUpdateOptions struct{}
```
Quinn-With-Two-Ns, line 1028: This is empty right now because the server does not return a failure on conflict token mismatch. Once that is fixed we can add an option to fail on that case and retry.

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		CronExpressions []string
+
+		// Skip - Any matching times will be skipped.
+		//
+		// All aspects of the schedulepb.ScheduleCalendarSpec—including seconds—must match a time for the time to be skipped.
+		Skip []ScheduleCalendarSpec
+
+		// StartAt - Any times before `startAt` will be skipped. Together, `startAt` and `endAt` make an inclusive interval.
+		// Optional: Defaulted to the beginning of time
+		StartAt *time.Time
```
Quinn-With-Two-Ns, line 781: I don't love taking a pointer here, but the server does not seem to treat the zero time as the default here. Same for `EndAt`

dnr: you could just convert the zero time to nil when you copy into the pb struct?

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+
+		// EndAt - Any times after `endAt` will be skipped.
+		// Optional: Defaulted to the end of time
+		EndAt *time.Time
+
+		// Jitter - All times will be incremented by a random value from 0 to this amount of jitter.
```
dnr, line 787: is it worth noting that it's capped by the time until the next scheduled action? (i.e. jitter will never cause two actions to switch relative order)

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		// Optional: Defaulted to 0
+		Jitter time.Duration
+
+		// Timezone - IANA timezone name, for example `US/Pacific`.
```
dnr, line 791: not really a "name" if you're taking a `time.Location`. also, should it be a `*time.Location`? that's what `time.UTC` is and most `time` functions take/return

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		Action ScheduleAction
+
+		// Overlap - Controls what happens when an Action would be started by a Schedule at the same time that an older Action is still
+		// running. This can be changed after a Schedule has taken some Actions, and some changes might produce
+		// unintuitive results. In general, the later policy overrides the earlier policy.
+		//
+		// Optional: defaulted to SCHEDULE_OVERLAP_POLICY_SKIP
+		Overlap enumspb.ScheduleOverlapPolicy
+
+		// CatchupWindow - The Temporal Server might be down or unavailable at the time when a Schedule should take an Action.
```
dnr, line 874: ... and then what?

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		// Optional: defaulted to 1 minute
+		CatchupWindow time.Duration
+
+		// PauseOnFailure - When an Action times out or reaches the end of its Retry Policy.
```
dnr, line 878: ... and then what?

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+
+		// RemainingActions - limit the number of Actions to take.
+		//
+		// This number is decremented after each Action is taken, and Actions are not
+		// taken when the number is `0` (unless {@link ScheduleHandle.trigger} is called).
+		//
+		// Optional: defaulted to zero
+		RemainingActions int64
+
+		// TriggerImmediately - Trigger one Action immediately.
```
dnr, line 899: trigger one action immediately on creating the schedule

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		WorkflowID string
+
+		// FirstExecutionRunID - The Run Id of the original execution that was started by the Schedule. If the Workflow retried, did
+		// Continue-As-New, or was Reset, the following runs would have different Run Ids.
+		FirstExecutionRunID string
+	}
+
+	// ScheduleInfo describes other information about a schedule.
+	ScheduleInfo struct {
+		// NumActions - Which Action to take
```
dnr, line 928: comment is off?

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		RemainingActions int64
+	}
+
+	// Schedule describes a created schedule.
+	Schedule struct {
+		// Action - Which Action to take
+		Action ScheduleAction
+
+		// Schedule - Describes when Actions should be taken.
+		Spec *ScheduleSpec
```
dnr, line 1012: just curious, why pointers for some and not others?

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		// If update returns a nil response then no update will occur.
+		//
+		// NOTE: If two Update calls are made in parallel to the same Schedule there is the potential
+		// for a race condition.
+		Update(ctx context.Context, fn func(*ScheduleDescribeResponse) *ScheduleUpdate, options ScheduleUpdateOptions) error
+
+		// Describe fetches the Schedule's description from the Server
+		Describe(ctx context.Context) (*ScheduleDescribeResponse, error)
+
+		// Trigger an Action to be taken immediately. Will override the schedules default policy
```
dnr, line 1048: note that the override is optional.. you can leave it UNSPECIFIED and it'll use the policy in the schedule

dnr, line 1048: add Backfill too?

internal/client.go
```diff
@@ -626,6 +630,515 @@ type (
...
+		RecentActions []ScheduleActionResult
+
+		// NextActionTimes - Next 10 scheduled Action times.
+		NextActionTimes []time.Time
+
+		// Memo - Non-indexed user supplied information.
+		Memo *commonpb.Memo
+
+		// SearchAttributes - Indexed info that can be used in query of List schedules APIs (only
+		// supported when Temporal server is using ElasticSearch). The key and value type must be registered on Temporal server side.
```
dnr, line 1106: we probably shouldn't mention Elasticsearch in the comments here. we call it the feature set that includes search attributes "Advanced Visibility" elsewhere. also, it will soon be supported for all backends

internal/client.go
Quinn-With-Two-Ns: yes I'll add that

internal/client.go
Quinn-With-Two-Ns: Yeah this isn't a name, I think I missed this from the typescript API. I'll update the comment.  I'll update to take a pointer

internal/client.go
Quinn-With-Two-Ns: We could, but if some did write zero time as the `StartAt ` or `EndAt` time could cause the value to get changed. I don't see a use case for setting them to zero time so I am OK with that.

internal/client.go
Quinn-With-Two-Ns: Yeah I took the wording from `StartWorkflowOptions` but standardizing on "Advanced Visibility" is better, I think.

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+
+		// Backfill the schedule by going though the specified time periods and taking Actions as if that time passed by right now, all at once.
+		Backfill(ctx context.Context, backfill []ScheduleBackfill) error
+
+		// Update the Schedule. fn takes a description of the schedule and returns the new desired schedule.
+		// If update returns a nil response then no update will occur.
+		//
+		// NOTE: If two Update calls are made in parallel to the same Schedule there is the potential
+		// for a race condition.
+		Update(ctx context.Context, fn func(*ScheduleDescribeResponse) *ScheduleUpdate, options ScheduleUpdateOptions) error
```
cretz, line 1055:
```suggestion
		Update(ctx context.Context, fn func(*ScheduleUpdate) error) error
```

```go
type ScheduleUpdate struct {
  // Read-only, alterations ignored
  Description *ScheduleDescribeResponse
  // Prepopulated from the description
  Schedule *Schedule
}
```

Benefits:

* Allows for more than just the describe response if we need it in the future
* Should prepopulate `Schedule` from the description. Args and memo and search attributes and what not can be `internal.EncodedValue` for pass-through needs.
* Removes need for "options" as the unwieldy last parameter after a closure. If/when a user wants to set options, we can expose those on the `ScheduleUpdate` such as fail-on-conflict vs retry-on-conflict
* Allows people to return an error from the callback

If you want people to be able to not update, I would support in order of preference: 1) allow `ErrSkipScheduleUpdate` to be returned, 2) changing return to `(bool, error)`, 3) have a `SkipUpdate bool` field on the update, or 4) attempt to diff against description to see if any mutations were made.

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+
+		// CreatedAt -  When the schedule was created
+		CreatedAt time.Time
+
+		// LastUpdateAt - When a schedule was last updated
+		LastUpdateAt time.Time
+	}
+
+	// ScheduleDescribeResponse describes the current Schedule details from ScheduleHandle.Describe.
+	ScheduleDescribeResponse struct {
```
cretz, line 966:
```suggestion
	ScheduleDescription struct {
```

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+		// for a race condition.
+		Update(ctx context.Context, fn func(*ScheduleDescribeResponse) *ScheduleUpdate, options ScheduleUpdateOptions) error
+
+		// Describe fetches the Schedule's description from the Server
+		Describe(ctx context.Context) (*ScheduleDescribeResponse, error)
+
+		// Trigger an Action to be taken immediately. Will override the schedules default policy
+		// with the one specified here. If overlap is SCHEDULE_OVERLAP_POLICY_UNSPECIFIED the schedule
+		// policy will be used.
+		Trigger(ctx context.Context, overlap enumspb.ScheduleOverlapPolicy) error
```
cretz, line 1063:
```suggestion
		Trigger(ctx context.Context, options ScheduleTriggerOptions) error
```

We've been bitten by needing more params too often. Unfortunately this means `Pause` and `Unpause` need `PauseOptions` and `UnpauseOptions` respectively. I know it can seem unwieldy for a caller, but we'll pay that for forward compatibility.

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+	ScheduleActionResult struct {
+		// ScheduleTime - Time that the Action was scheduled for, including jitter.
+		ScheduleTime time.Time
+
+		// ActualTime - Time that the Action was actually taken.
+		ActualTime time.Time
+
+		// StartWorkflowResult - If action was ScheduleWorkflowAction, returns the
+		// ID of the workflow.
+		StartWorkflowResult ScheduleWorkflowExecution
```
cretz, line 1084:
```suggestion
		StartWorkflowResult *ScheduleWorkflowExecution
```

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+		// ScheduleID - The business identifier of the schedule.
+		ScheduleID string
+
+		// Spec - Describes when Actions should be taken.
+		Spec *ScheduleSpec
+
+		// Note - Informative human-readable message with contextual notes, e.g. the reason
+		// a Schedule is paused. The system may overwrite this message on certain
+		// conditions, e.g. when pause-on-failure happens.
+		Note string
```
cretz, line 1098: I have seen "note" in our Go code and "notes" in the proto API. Should we call this `Notes` everywhere instead? Why this mismatch?

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+	ScheduleListIterator interface {
+		// HasNext return whether this iterator has next value
+		HasNext() bool
+
+		// Next returns the next schedule and error
+		Next() (*ScheduleListEntry, error)
+	}
+
+	// Client for creating Schedules and creating Schedule handles
+	ScheduleClient interface {
```
cretz, line 1142: Maybe just `Create`, `List`, and `GetHandle` since you're already scoped to a schedule client

internal/interceptor.go
```diff
@@ -299,6 +299,22 @@ type ClientOutboundInterceptor interface {
 	mustEmbedClientOutboundInterceptorBase()
 }
 
+// ScheduleClientInterceptor provides a way to intercept
+// certain schedule-specific client calls from the SDK. See documentation in the
+// interceptor package for more details.
+type ScheduleClientInterceptor interface {
```
cretz, line 305: If you must have a separate interceptor, you should rename this to `ScheduleClientOutboundInterceptor` and add `InterceptScheduleClient(next ScheduleClientOutboundInterceptor) ScheduleClientOutboundInterceptor` to `ClientInterceptor` and invoke it with your underlying interceptor on instantiation. I would also support just putting `CreateSchedule` on the `ClientOutboundInterceptor`.

internal/internal_workflow_client.go
```diff
@@ -1515,3 +1542,701 @@ func (w *workflowClientInterceptor) QueryWorkflow(
...
+		Intervals: intervals,
+		Skip:      skip,
+		StartAt:   scheduleSpec.GetStartTime(),
+		EndAt:     scheduleSpec.GetEndTime(),
+		Jitter:    common.DurationValue(scheduleSpec.GetJitter()),
+		Timezone:  timezone,
+	}
+}
+
+func ScheduleDescriptionFromPB(describeResponse *workflowservice.DescribeScheduleResponse) (*ScheduleDescribeResponse, error) {
```
cretz, line 1989: No need to expose

internal/client.go
Quinn-With-Two-Ns: >we don't expose WorkflowExecution to users anyways, so this is fine.
We do in `WorkflowInfo` no?

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
 	}
+
+	// ScheduleRange represents a set of integer values, used to match fields of a calendar
+	// time in StructuredCalendarSpec. If end < start, then end is interpreted as
+	// equal to start. This means you can use a Range with start set to a value, and
+	// end and step unset (defaulting to 0) to represent a single value.
+	ScheduleRange = schedulepb.Range
+
+	// ScheduleRanges represents a list of ScheduleRange.
+	ScheduleRanges = []*ScheduleRange
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Sure, I wrapped it cause I found it was verbose writing the tests. When I wrap it i'll make it a value type so It will be a bit better.

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+
+		// WorkflowType - The type of the wokrflow.
+		WorkflowType WorkflowType
+
+		// Args - Arguments of the workflow
+		Args *commonpb.Payloads
+
+		// TaskQueueName -The workflow tasks of the workflow are scheduled on the queue with this name.
+		// This is also the name of the activity task queue on which activities are scheduled.
+		TaskQueueName string
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Oh it's `TaskQueueName` in https://pkg.go.dev/go.temporal.io/sdk@v1.17.0/internal#WorkflowOptions and https://pkg.go.dev/go.temporal.io/sdk@v1.17.0/internal#WorkflowInfo

internal/internal_workflow_client.go
```diff
@@ -1515,3 +1542,701 @@ func (w *workflowClientInterceptor) QueryWorkflow(
...
+			DayOfWeek:  c.DayOfWeek,
+			Comment:    c.Comment,
+		}
+	}
+
+	intervals := make([]*schedulepb.IntervalSpec, len(scheduleSpec.Intervals))
+	for i, interval := range scheduleSpec.Intervals {
+		intervalSpec := interval
+		intervals[i] = &schedulepb.IntervalSpec{
+			Interval: &intervalSpec.Every,
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: `Interval` is required per the GRPC, and `Phase` defaults to zero if unset

```
// IntervalSpec matches times that can be expressed as:
// epoch + n * interval + phase
// where n is an integer.
// phase defaults to zero if missing. interval is required.
// Both interval and phase must be non-negative and are truncated to the nearest
// second before any calculations.
// For example, an interval of 1 hour with phase of zero would match every hour,
// on the hour. The same interval but a phase of 19 minutes would match every
// xx:19:00. An interval of 28 days with phase zero would match
// 2022-02-17T00:00:00Z (among other times). The same interval with a phase of 3
// days, 5 hours, and 23 minutes would match 2022-02-20T05:23:00Z instead.
type IntervalSpec struct {
	Interval *time.Duration `protobuf:"bytes,1,opt,name=interval,proto3,stdduration" json:"interval,omitempty"`
	Phase    *time.Duration `protobuf:"bytes,2,opt,name=phase,proto3,stdduration" json:"phase,omitempty"`
}
```

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+		EndAt *time.Time
+
+		// Jitter - All times will be incremented by a random value from 0 to this amount of jitter, capped
+		// by the time until the next schedule. 
+		// Optional: Defaulted to 0
+		Jitter time.Duration
+
+		// Timezone - timezone to interpret all ScheduleCalendarSpec in.
+		// Optional: Defaulted to UTC
+		Timezone *time.Location
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Sure i'll change it, I always thought it was one word online that seems a bit controversial

internal/client.go
cretz: Yes, as `workflow.Execution`, but we should try to not require users use the `workflow` package outside of an actual workflow, and especially not for clients if we can help it. We messed up with `workflow.RegisterInfo` too. It is really hard for a user to discover from Go doc that some of this is visible from the `client` package and half from the `workflow` package.

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+		// supported when Temporal server is using advanced visibility). The key and value type must be registered on Temporal server side.
+		// Use GetSearchAttributes API to get valid key and corresponding value type.
+		SearchAttributes *commonpb.SearchAttributes
+	}
+
+	// ScheduleOptions configure the parameters for creating a schedule.
+	ScheduleOptions struct {
+		// ScheduleID - The business identifier of the schedule.
+		// Optional: defaulted to a uuid.
+		ScheduleID string
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Some other structs like `WorkflowOptions` called the workflow ID `WorkflowID` so I was trying to be consistent with them. I'd rather just use `ID`

internal/client.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: updating the memo of the workflow action? No I will add one

internal/client.go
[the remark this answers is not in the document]
cretz: This isn't wrapping it, it's aliasing it (not to be confused with `ScheduleRanges []*ScheduleRange` which actually does make a new type). All this is doing is saving you from typing two extra characters each time.

internal/client.go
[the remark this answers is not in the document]
cretz: Yup, we're pretty bad at consistency. But this is all client API so let's go w/ the client one.

internal/client.go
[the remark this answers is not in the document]
cretz: I think we should match the client for this client API.

internal/client.go
Quinn-With-Two-Ns: That is true. TS calls it `note` as well. I would prefer `note` because, generally, I think plurals should be used for lists. That being said I am in favor of changing to `notes` so we are consistent.

internal/client.go
Quinn-With-Two-Ns: it's only scoped to the interface, not the implementation. The implementation could still have collisions if it implemented multiple interfaces.

In this case probably not an issue, but that was my thought process.

internal/client.go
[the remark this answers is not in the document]
cretz: Thanks! It can help me see what updating a small part of the schedule looks like.

internal/client.go
Quinn-With-Two-Ns: Makes sense, should we propagate this back to typescript? I guess they can take variants so they can patch in a struct easier.

internal/client.go
cretz: I don't mind if we are consistent in our SDKs. It is unfortunate that we can't be consistent across the company, but oh well.

internal/client.go
cretz: > The implementation could still have collisions if it implemented multiple interfaces.

Would prefer to optimize for callers than implementers here. `myClient.ScheduleClient().Create()` seems more reasonable than `myClient.ScheduleClient().CreateSchedule()`.

internal/client.go
cretz: I do think TS should do this, yes, `options? { note?: string }` I think is clearer. But they don't need to as much as we do because they can add new params in a compat-safe manner a bit easier.

internal/client.go
Quinn-With-Two-Ns: Oh I see your point, OK I'll leave as is.

internal/internal_workflow_client.go
```diff
@@ -1515,3 +1542,701 @@ func (w *workflowClientInterceptor) QueryWorkflow(
...
+				PauseOnFailure: in.Options.PauseOnFailure,
+			},
+			State: &schedulepb.ScheduleState{
+				Notes:            in.Options.Note,
+				Paused:           in.Options.Paused,
+				LimitedActions:   in.Options.RemainingActions != 0,
+				RemainingActions: in.Options.RemainingActions,
+			},
+		},
+		InitialPatch: &schedulepb.SchedulePatch{
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Looking at the server code it's fine, but maybe slightly wasteful (since the server will trigger a patch that does nothing). It is very little work to leave out.

internal/client.go
Quinn-With-Two-Ns: >It's confusing that I can create a schedule with a description action.

it's a description of an action, I think it is pretty normal that a description of an object could be used to create a new one. A potential use case would be if you want to copy a schedule action with a different schedule spec or to a new namespace.

Another option I thought of is remove the type `ScheduleWorkflowAction` and replace with a function 
```
CreateScheduleWorkflowAction(Workflow interface{}, Args []interface{}, WorkflowOptions StartWorkflowOptions) ScheduleWorkflowActionDescription
```

`ScheduleWorkflowActionDescription` could then be renamed

internal/client.go
Quinn-With-Two-Ns: >Should prepopulate Schedule from the description. Args and memo and search attributes and what not can be internal.EncodedValue for pass-through needs.

Probably misunderstanding, this doesn't seem like a benefit just a description of what the API is? The 1st and 4th benefit are clear I don't get why we need the prepopulated ` Schedule *Schedule` field since `Description .Schedule` would have the same information?

>Removes need for "options" as the unwieldy last parameter after a closure. If/when a user wants to set options, we can expose those on the ScheduleUpdate such as fail-on-conflict vs retry-on-conflict

I agree the last parameter options is unwieldy, I am not sure about exposing them in `ScheduleUpdate` since we may want to know an option before we run `fn` also has the possibility the options may change between calls. Maybe something like
```go
Update(ctx context.Context, options UpdateOptions) error

type UpdateOptions struct {
  
  Fn func(*ScheduleUpdate) error
  // example
  RetryCount int
}
```

internal/client.go
cretz: Having two types of actions represent the same thing on the same interface is rough for general-purpose/helper code. We should think about possible alternatives. Definitely want simple to create and simple to update.

internal/client.go
```diff
@@ -626,6 +630,528 @@ type (
...
+
+		// ScheduleBackfill - Runs though the specified time periods and takes Actions as if that time passed by right now, all at once. The
+		// overlap policy can be overridden for the scope of the ScheduleBackfill.
+		ScheduleBackfill []ScheduleBackfill
+
+		// Memo - Optional non-indexed info that will be shown in list schedules.
+		Memo map[string]interface{}
+
+		// SearchAttributes - Optional indexed info that can be used in query of List schedules APIs (only
+		// supported when Temporal server is using advanced visibility). The key and value type must be registered on Temporal server side.
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Other places with `SearchAttributes` we mention the server needs Elastic or require advanced visibility so I was trying to be consistent. I think it is a simple grep to find all mentions and update/remove, but I wouldn't die on this hill if people feel strongly the other way.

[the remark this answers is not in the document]
cretz: Those places are going to be wrong too. But if we want to change them later, ok.

internal/client.go
cretz: > I don't get why we need the prepopulated Schedule *Schedule field since Description .Schedule would have the same information?

If that remains the case once we solve the problem of two actions meaning the same, no problem. But we still might want to copy the pointer to top level so that they don't think the otherwise-read-only description is expected to be mutated.

> since we may want to know an option before we run fn

Usually I support making the API unwieldy for options support, this may be a bit much. I think your `RetryCount int` would be more like `RetryAttempt int` on `ScheduleUpdate` and the user could bail.

But I'm ok with what you have suggested there. I'd call the field `DoUpdate` instead of `Fn`, but if this single-required-field struct works for you, it works for me too.

test/integration_test.go
```diff
@@ -2682,6 +2682,746 @@ func (ts *IntegrationTestSuite) TestUpsertMemoWithExistingMemo() {
...
+		ts.NoError(handle.Delete(ctx))
+	}()
+	for i := 0; i < 5; i++ {
+		ts.NoError(handle.Trigger(ctx, enumspb.SCHEDULE_OVERLAP_POLICY_ALLOW_ALL))
+		// schedule actions can only trigger once per second
+		time.Sleep(3 * time.Second)
+	}
+	description, err := handle.Describe(ctx)
+	ts.NoError(err)
+	ts.EqualValues(5, description.Info.NumActions)
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Done, the test now integrates through the recent actions and gets the workflow results as well

internal/internal_workflow_client.go
```diff
@@ -1515,3 +1542,701 @@ func (w *workflowClientInterceptor) QueryWorkflow(
...
+		Options: &options,
+	})
+}
+
+func (wc *WorkflowClient) ListSchedules(ctx context.Context, options ScheduleListOptions) (ScheduleListIterator, error) {
+	pageSize := options.PageSize
+	if options.PageSize <= 0 {
+		pageSize = 1000
+	}
+	paginate := func(nextToken []byte) (*workflowservice.ListSchedulesResponse, error) {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I found it cleaner code, but that's a bit subjective.

internal/internal_workflow_client.go
```diff
@@ -1515,3 +1542,701 @@ func (w *workflowClientInterceptor) QueryWorkflow(
...
+		defer cancel()
+		request := &workflowservice.ListSchedulesRequest{
+			Namespace:       wc.namespace,
+			MaximumPageSize: pageSize,
+			NextPageToken:   nextToken,
+		}
+		var response *workflowservice.ListSchedulesResponse
+		var err error
+		for {
+			response, err = wc.workflowService.ListSchedules(grpcCtx, request)
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: No shouldn't be needed I removed the loop

internal/client.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Added

interceptor/tracing_interceptor.go
```diff
@@ -221,6 +221,25 @@ type tracingClientOutboundInterceptor struct {
 	root *tracingInterceptor
 }
 
+func (t *tracingClientOutboundInterceptor) CreateSchedule(ctx context.Context, in *ScheduleClientCreateInput) (client.ScheduleHandle, error) {
+	// Start span and write to header
+	span, ctx, err := t.root.startSpanFromContext(ctx, &TracerStartSpanOptions{
+		Operation: "CreateSchedule",
+		Name:      in.Options.ID,
```
Quinn-With-Two-Ns, line 228: schedules don't really have a concept of a name so I used the ID.

internal/internal_schedule_client.go
```diff
@@ -0,0 +1,789 @@
...
+		recentActions[i] = ScheduleActionResult{
+			ScheduleTime:        common.TimeValue(a.GetScheduleTime()),
+			ActualTime:          common.TimeValue(a.GetActualTime()),
+			StartWorkflowResult: workflowExecution,
+		}
+	}
+	return recentActions
+}
+
+func encodeScheduleWorklowArgs(dc converter.DataConverter, args []interface{}) (*commonpb.Payloads, error) {
```
Quinn-With-Two-Ns, line 751: @cretz I did consider modifying the existing functions instead of creating new ones, but I didn't want to leak this visible change out of the experimental API

internal/internal_schedule_client.go
```diff
@@ -0,0 +1,789 @@
...
+	payloads := make([]*commonpb.Payload, len(args))
+	for i, arg := range args {
+		// arg is already encoded
```
Quinn-With-Two-Ns, line 754: I support a mix of serialized and not serialized Args. I found it was nice when writing integration tests.

internal/internal_schedule_client.go
```diff
@@ -0,0 +1,789 @@
...
+		if enc, ok := arg.(*commonpb.Payload); ok {
```
Quinn-With-Two-Ns, line 755: I chose to use *commonpb.Payload here instead of `EncodedValues` because it limited the amount of possible user error, was consistent with current memo and search attribute usage in other parts of the SDK

internal/internal_workflow_client.go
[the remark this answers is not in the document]
cretz: Hopefully you don't normally prefer anon funcs over methods :-)

internal/internal_schedule_client.go
cretz: This is probably ok, but technically this is a bit different than elsewhere where we'd call `DataConverter.ToPayloads` (plural) w/ all args if they wanted to, say, condense them or something. But most people treat that as the same as calling each individually.

You may want to document on args, search attributes, and memos that we just pass through payloads on create/update.

internal/internal_schedule_client.go
cretz: That's a fairly recent change to make those that way IIRC, but I'm fine w/ it.

cretz (approved):

internal/internal_schedule_client.go
```diff
@@ -0,0 +1,789 @@
...
+			Action: action,
+			Policies: &schedulepb.SchedulePolicies{
+				OverlapPolicy:  in.Options.Overlap,
+				CatchupWindow:  &in.Options.CatchupWindow,
+				PauseOnFailure: in.Options.PauseOnFailure,
+			},
+			State: &schedulepb.ScheduleState{
+				Notes:            in.Options.Note,
+				Paused:           in.Options.Paused,
+				LimitedActions:   in.Options.RemainingActions != 0,
```
mjameswh, line 136: In describe and update, the value of `RemainingActions` has to be interpreted in regard of `LimitedActions`. Assuming `LimitedActions == true`, then `RemainingActions == 0` would indicate that the schedule can't happens anymore (ie. it is implicitely paused), until remainingActions is reset to a non zero value.

Having only `RemainingActions` in create might make this semantic harder to understand for users. I'd suggest either adding `LimitedActions` to create options, or using a value other than 0 (ie `-1`) for "don't limit on action count", and using that same semantic in Describe and Update.

test/integration_test.go
```diff
@@ -2682,6 +2682,850 @@ func (ts *IntegrationTestSuite) TestUpsertMemoWithExistingMemo() {
...
+		ts.NoError(handle.Delete(ctx))
+	}()
+
+	description, err := handle.Describe(ctx)
+	ts.NoError(err)
+	// test spec
+	ts.Equal([]client.ScheduleCalendarSpec{
+		{
+			Second: []client.ScheduleRange{{}},
+			Minute: []client.ScheduleRange{{}},
```
mjameswh, line 2909: I must be missing something, but here, I would expect an array containing a single proto Range, with `{0, 0, 0}`... Same for `Second`. Isn't that the case?

mjameswh (approved):

PR #944, opened 2022-11-02
Adapt to interaction-based upstream API changes

mmcshane:
## What was changed

An implementation of sync update using v1.13 of the upstream temporal API which is based on the interaction concept. 

## Why?

This is a checkpoint get a working implementation before we refactor upstream toward the protocol/message concepts.

## Checklist

1. Closes 

2. How was this tested:

Unit tests and sdk-features

3. Any docs updates needed?


alexshtin (approved):

internal/internal_task_handlers.go
```diff
@@ -796,9 +800,27 @@ processWorkflowLoop:
...
 }
 
+// indexInvocations builds a map of the interaction invocations contained in a
+// workflow task where the index is the interaction's event level (i.e. the
+// event after which the interaction can execute) and the value is the
+// invocation itself. This function may return an empty map but it will not
+// return nil.
+func indexInvocations(workflowTask *workflowTask) map[int64][]*interactionpb.Invocation {
+	out := map[int64][]*interactionpb.Invocation{}
+	for _, ev := range workflowTask.task.GetHistory().GetEvents() {
```
cretz, line 814: Can you help me understand this a bit? Can you confirm that `GetEvents()` is _only_ events for this task?

Right now what I read is "every time you see a workflow task started, duplicate the interaction list on its meta event ID in the map" which seems a bit strange. Is the goal to see if the _most recent_ workflow task started matches the meta event ID (missing conditional?) and then only add that to the map?

As the algorithm sits now, why even loop over history events? Just make the map from the interactions.

cretz (approved):
LGTM, one question about the interaction map. Approving anyways because for me - no test failures and no public API changes means good enough.

(we may have to make a bit of haste on this to get SDK working with latest API tag)

internal/internal_task_handlers.go
mmcshane: Thank you for catching this. This code block is from when we were going to attach invocations to the WFTStartedEvent. We have switched to attaching invocations directly to the WFT. I included the wrong impl in this PR. Updated this function and added a UT.

PR #956, opened 2022-11-15
bugfix: consistently plumb failureConverter

asonawalla:
## What was changed
Literal definitions of `workflowRunImpl` now consistently plumb the `failureConverter` through, fixing a panic we've been observing since the implementation of failureConverter.

NOTE: some formatting nits included courtesy of `gofmt`

## Why?
The recent addition of a `failureConverter` added implicit not-nil assumption [here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_workflow_client.go#L1140), where we've been observing a nil-pointer panic.  Even though the temporal client itself initializes a default failure converter, before this change it was not passed to workflowRunImpl.

## Checklist
1. Closes
N/A
2. How was this tested:
By querying a failed workflow's status using the patched code. The unpatched code would panic, the patched code behaves as expected.
3. Any docs updates needed?
N/A

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -194,9 +194,11 @@ type (
 // reaches the end state, such as workflow finished successfully or timeout.
 // The user can use this to start using a functor like below and get the workflow execution result, as EncodedValue
 // Either by
-//     ExecuteWorkflow(options, "workflowTypeName", arg1, arg2, arg3)
-//     or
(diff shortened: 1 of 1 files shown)


bergundy (approved):
Not sure why the comments were reformatted but the fix LGTM.
Thanks!

cretz (approved):
> Not sure why the comments were reformatted but the fix LGTM.

https://golang.org/doc/comment was updated significantly in Go 1.19 and go fmt reflects it. Expect to see these more and more when these older files are touched.

Issue #958, opened 2022-11-15
Confusing interface match for client.ScheduleWorkflowAction


cretz:
In Go, a receiver that is a non-pointer matches pointer and non-pointer. Therefore most interfaces that are used like sum types, like our `ScheduleAction`, expect the concrete types to be pointers so the developer cannot have both pointer and non-pointer options.

Today, `ScheduleAction` can be set to `ScheduleWorkflowAction` _or_ `*ScheduleWorkflowAction`, but the latter fails at runtime. I think maybe we should always _require_ `*ScheduleWorkflowAction` by changing receiver to `func (*ScheduleWorkflowAction) isScheduleAction()`.

Quinn-With-Two-Ns:
closed in https://github.com/temporalio/sdk-go/pull/959

Issue #957, opened 2022-11-15
ScheduleOptions.ID says optional, but runtime requires it


cretz:
ScheduleOptions.ID says optional and will default with UUID, but runtime errors with "no schedule ID in options". Need to decide whether docs are wrong or runtime is wrong.

Also, please document that this may not be the actual workflow ID, it may have the timestamp appended.

Quinn-With-Two-Ns:
>Also, please document that this may not be the actual workflow ID, it may have the timestamp appended.

The schedule ID is not related to the workflow ID?

cretz:
> Typescript requires the ID, I think Go should be the same

Works for me, but it means https://github.com/temporalio/sdk-go/blob/v1.18.1/internal/schedule_client.go#L270 is wrong and needs to be updated.

> The schedule ID is not related to the workflow ID?

The schedule ID is related but not the same. I ran into this writing tests just now. From https://github.com/temporalio/api/blob/v1.12.0/temporal/api/schedule/v1/message.proto#L252-L253:

> The workflow id of the started workflow may not match this exactly, it may have a timestamp appended for uniqueness.

When I was trying to look up a scheduled workflow by the ID I put in the scheduled workflow action, it failed because there was a timestamp appended.

Quinn-With-Two-Ns:
yeah I can update the comment. 

>The schedule ID is related but not the same. I ran into this writing tests just now. From https://github.com/temporalio/api/blob/v1.12.0/temporal/api/schedule/v1/message.proto#L252-L253:

that's the ID of the schedule workflow action, not the schedule, but I agree we should clarify it won't be used verbatim.

cretz:
From the API there is no such thing as a schedule workflow action ID, only a `start_workflow` options for a workflow action. But the comment makes clear that ID may have something appended. We can make the same comment - meaning don't say "it won't be used verbatim", but rather "it may have timestamp appended". My tests at https://github.com/temporalio/sdk-features/pull/158 for instance rely on the prefix to disambiguate across runs.

Quinn-With-Two-Ns:
yeah I agree we should have the same comment 👍

Quinn-With-Two-Ns:
closed in https://github.com/temporalio/sdk-go/pull/959

PR #959, opened 2022-11-15
Schedule fixes

Quinn-With-Two-Ns:
Fix a few issues @cretz found in the schedules API:
- clean up documentation on schedule ID and workflow ID
- standardize on taking a pointer type for `ScheduleWorkflowAction`

closes:
* https://github.com/temporalio/sdk-go/issues/958
* https://github.com/temporalio/sdk-go/issues/957

Note: This is a breaking change since users can no longer pass a struct for `ScheduleWorkflowAction` to avoid this we could  allow users to pass a pointer or a struct.

### DIFF
--- a/internal/schedule_client.go
+++ b/internal/schedule_client.go
@@ -193,7 +193,7 @@ type (
 		EndAt time.Time
 
 		// Jitter - All times will be incremented by a random value from 0 to this amount of jitter, capped
-		// by the time until the next schedule. 
+		// by the time until the next schedule.
 		// Optional: Defaulted to 0
 		Jitter time.Duration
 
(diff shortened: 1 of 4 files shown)


internal/schedule_client.go
```diff
@@ -597,8 +599,7 @@ type (
 		// methods like ScheduleHandle.Describe() will return an error.
 		GetHandle(ctx context.Context, scheduleID string) ScheduleHandle
 	}
-
 )
 
-func (ScheduleWorkflowAction) isScheduleAction() {
+func (*ScheduleWorkflowAction) isScheduleAction() {
```
cretz, line 604: Note that this is a backwards incompatible change. I think it's ok because we're "experimental", just worth noting.

cretz (approved):

PR #963, opened 2022-11-21
Fix default catchupWindow

Quinn-With-Two-Ns:
Fix a mismatch between the server and SDK around `CatchupWindow`. The SDK uses zero `time.duration` to represent default but the server treated this duration as the minimum (10s). This causes schedule actions to fail more often under load because the catch up window is smaller.

### DIFF
--- a/internal/internal_schedule_client.go
+++ b/internal/internal_schedule_client.go
@@ -117,6 +117,13 @@ func (w *workflowClientInterceptor) CreateSchedule(ctx context.Context, in *Sche
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


bergundy (approved):
Can you check if this applies to TS SDK too?

Quinn-With-Two-Ns:
@bergundy TS should be fine because  `catchupWindow` is marked as optional and the undefined is passed through https://github.com/temporalio/sdk-typescript/blob/8103c004e56450a06654279e29731d4a29938503/packages/common/src/time.ts#L56

Issue #854, opened 2022-07-15
num_pollers metric


robholland:
**Is your feature request related to a problem? Please describe.**
There is no clear way to tell how many pollers were active at any given time when tuning a cluster or doing an outage post mortem.

**Describe the solution you'd like**
Please implement the num_pollers metric for the Go SDK: https://docs.temporal.io/references/sdk-metrics/#num_pollers

yiminc:
Need to be able to tell how many activity pollers and how many workflow pollers.

PR #965, opened 2022-11-28
Add num_pollers metric

Quinn-With-Two-Ns:
Closes https://github.com/temporalio/sdk-go/issues/854

Implements the `num_pollers` metric for Go

NOTE:
Once released we need to update the docs to include Go here
https://docs.temporal.io/references/sdk-metrics#num_pollers

### DIFF
--- a/internal/common/metrics/tags.go
+++ b/internal/common/metrics/tags.go
@@ -80,3 +80,10 @@ func WorkerTags(workerType string) map[string]string {
 		WorkerTypeTagName: workerType,
 	}
 }
+
+// PllerTags returns a set of tags for pollers.
+func PollerTags(pollerType string) map[string]string {
+	return map[string]string{
+		PollerTypeTagName: pollerType,
+	}
+}
--- a/internal/common/metrics/tags.go
+++ b/internal/common/metrics/tags.go
@@ -81,7 +81,7 @@ func WorkerTags(workerType string) map[string]string {
 	}
 }
 
-// PllerTags returns a set of tags for pollers.
+// PollerTags returns a set of tags for pollers.
 func PollerTags(pollerType string) map[string]string {
 	return map[string]string{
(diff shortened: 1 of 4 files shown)


internal/common/metrics/tags.go
```diff
@@ -80,3 +80,10 @@ func WorkerTags(workerType string) map[string]string {
 		WorkerTypeTagName: workerType,
 	}
 }
+
+// PllerTags returns a set of tags for pollers.
```
cretz, line 84:
```suggestion
// PollerTags returns a set of tags for pollers.
```

internal/internal_task_pollers.go
```diff
@@ -632,6 +644,23 @@ func (wtp *workflowTaskPoller) getNextPollRequest() (request *workflowservice.Po
 	}
 }
 
+// Poll the workflow task queue and update the num_poller metric
+func (wtp *workflowTaskPoller) pollWorkflowTaskQueue(ctx context.Context, request *workflowservice.PollWorkflowTaskQueueRequest) (*workflowservice.PollWorkflowTaskQueueResponse, error) {
+	if request.TaskQueue.GetKind() == enumspb.TASK_QUEUE_KIND_NORMAL {
+		wtp.numNormalPollersGauge.Update(float64(atomic.AddInt32(&wtp.numNormalPollers, 1)))
```
cretz, line 650: This is racy. What if `numNormalPollers` is atomically updated twice in one order but gauge `Update` is called twice in a different order. You could have:

```
numNormalPollers=20
numNormalPollers=21
gauge.Update(21)
gauge.Update(20)
```

If you want truly accurate numbers, you must make the gauge update atomic, not just the var. If there is a perf concern, we can discuss accuracy vs contention though if we're willing to accept inaccurate. But I think contention is low enough here, locking the gauge update is reasonable.

cretz (approved):

PR #966, opened 2022-11-28
Add local_activity_execution_failed metric

Quinn-With-Two-Ns:
add local_activity_execution_failed to align with java

for https://github.com/temporalio/sdk-features/issues/167

### DIFF
--- a/internal/common/metrics/constants.go
+++ b/internal/common/metrics/constants.go
@@ -50,13 +50,14 @@ const (
 	ActivitySucceedEndToEndLatency        = TemporalMetricsPrefix + "activity_succeed_endtoend_latency"
 	ActivityTaskErrorCounter              = TemporalMetricsPrefix + "activity_task_error"
 
-	LocalActivityTotalCounter           = TemporalMetricsPrefix + "local_activity_total"
-	LocalActivityCanceledCounter        = TemporalMetricsPrefix + "local_activity_canceled"
-	LocalActivityFailedCounter          = TemporalMetricsPrefix + "local_activity_failed" // Deprecated: Use LocalActivityExecutionFailedCounter instead.
-	LocalActivityExecutionFailedCounter = TemporalMetricsPrefix + "local_activity_execution_failed"
-	LocalActivityErrorCounter           = TemporalMetricsPrefix + "local_activity_error"
-	LocalActivityExecutionLatency       = TemporalMetricsPrefix + "local_activity_execution_latency"
-	LocalActivitySucceedEndToEndLatency = TemporalMetricsPrefix + "local_activity_succeed_endtoend_latency"
+	LocalActivityTotalCounter             = TemporalMetricsPrefix + "local_activity_total"
+	LocalActivityCanceledCounter          = TemporalMetricsPrefix + "local_activity_canceled" // Deprecated: Use LocalActivityExecutionCanceledCounter instead.
+	LocalActivityExecutionCanceledCounter = TemporalMetricsPrefix + "local_activity_execution_cancelled"
+	LocalActivityFailedCounter            = TemporalMetricsPrefix + "local_activity_failed" // Deprecated: Use LocalActivityExecutionFailedCounter instead.
+	LocalActivityExecutionFailedCounter   = TemporalMetricsPrefix + "local_activity_execution_failed"
+	LocalActivityErrorCounter             = TemporalMetricsPrefix + "local_activity_error"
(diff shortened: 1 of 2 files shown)


internal/common/metrics/constants.go
```diff
@@ -52,7 +52,8 @@ const (
 
 	LocalActivityTotalCounter           = TemporalMetricsPrefix + "local_activity_total"
 	LocalActivityCanceledCounter        = TemporalMetricsPrefix + "local_activity_canceled"
```
cretz, line 54: Should we also have `local_activity_execution_cancelled`?

Quinn-With-Two-Ns: yeah that's missing to I'll add in this PR

cretz (approved):
Saw that this came from @tsurdilo's SDK features request. Would like @bergundy's confirmation that this is an acceptable approach before merging. Adding extra metrics is ok with me.

Quinn-With-Two-Ns:
In go we are still missing https://docs.temporal.io/references/sdk-metrics#activity_execution_cancelled in the java sdk, but this PR should at least align the go sdk with what our docs say

internal/common/metrics/constants.go
```diff
@@ -50,12 +50,14 @@ const (
...
 
-	LocalActivityTotalCounter           = TemporalMetricsPrefix + "local_activity_total"
-	LocalActivityCanceledCounter        = TemporalMetricsPrefix + "local_activity_canceled"
-	LocalActivityFailedCounter          = TemporalMetricsPrefix + "local_activity_failed"
-	LocalActivityErrorCounter           = TemporalMetricsPrefix + "local_activity_error"
-	LocalActivityExecutionLatency       = TemporalMetricsPrefix + "local_activity_execution_latency"
-	LocalActivitySucceedEndToEndLatency = TemporalMetricsPrefix + "local_activity_succeed_endtoend_latency"
+	LocalActivityTotalCounter             = TemporalMetricsPrefix + "local_activity_total"
+	LocalActivityCanceledCounter          = TemporalMetricsPrefix + "local_activity_canceled" // Deprecated: Use LocalActivityExecutionCanceledCounter instead.
+	LocalActivityExecutionCanceledCounter = TemporalMetricsPrefix + "local_activity_execution_cancelled"
```
bergundy, line 55: Are we using 2 `l`s in Java?

internal/common/metrics/constants.go
```diff
@@ -50,12 +50,14 @@ const (
...
+	LocalActivityFailedCounter            = TemporalMetricsPrefix + "local_activity_failed" // Deprecated: Use LocalActivityExecutionFailedCounter instead.
+	LocalActivityExecutionFailedCounter   = TemporalMetricsPrefix + "local_activity_execution_failed"
+	LocalActivityErrorCounter             = TemporalMetricsPrefix + "local_activity_error"
```
bergundy, line 58: What's the difference between these 2?
I don't see `local_activity_error` in our docs.

internal/common/metrics/constants.go
bergundy: Yes: https://github.com/temporalio/sdk-java/blob/c5f0ebbe8a5cd52ee018809ecdd466379b5a7339/temporal-sdk/src/main/java/io/temporal/worker/MetricsType.java

bergundy:
Overall LGTM, I'm fine duplicating some metrics to keep backwards compatibility.
Let's add an issue to track removal of duplicates after a while (duration not defined at the moment).
Also, please don't mark the issue as closed, we need to fix the Core metrics too.
I'd want to have sdk-features validation before we close.

internal/common/metrics/constants.go
Quinn-With-Two-Ns: yes, go and java match

internal/common/metrics/constants.go
Quinn-With-Two-Ns: I didn't touch `local_activity_error ` in this PR. Looks like it counts how many times local activities panic

Quinn-With-Two-Ns:
Issue to deprecate in the future: https://github.com/temporalio/sdk-go/issues/968

internal/common/metrics/constants.go
bergundy: And what does `local_activity_execution_failed` track?

Quinn-With-Two-Ns: `local_activity_execution_failed ` tracks if the local activity fails. If the activity panics both are incremented

Issue #970, opened 2022-12-01
Search attributes and memos are kept from previous runs when resetting workflow context


cretz:
## Expected Behavior

When a workflow's context is reset due to cache staleness, all mutable aspects should be reset to original values

## Actual Behavior

When a workflow's context is reset due to cache staleness, the mutated search attribute map remains (and I presume memo too)

## Steps to Reproduce the Problem

This sample program will cause a non-determinism error due to cache reset not working right. (there is another issue concerning workers sharing cache that I may open another issue for)

```go
package main

import (
	"context"
	"fmt"
	"log"
	"time"

	"github.com/google/uuid"
	"go.temporal.io/sdk/client"
	"go.temporal.io/sdk/worker"
	"go.temporal.io/sdk/workflow"
)

func main() {
	if err := run(); err != nil {
		log.Fatal(err)
	}
}

func run() error {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	log.Printf("Creating client")
	c, err := client.NewClient(client.Options{})
	if err != nil {
		return fmt.Errorf("failed creating client: %w", err)
	}
	defer c.Close()

	log.Printf("Starting worker1")
	taskQueue := uuid.NewString()
	worker1 := worker.New(c, taskQueue, worker.Options{})
	worker1.RegisterWorkflow(MyWorkflow)
	if err := worker1.Start(); err != nil {
		return fmt.Errorf("failed starting worker client: %w", err)
	}

	log.Printf("Starting workflow")
	run, err := c.ExecuteWorkflow(ctx, client.StartWorkflowOptions{TaskQueue: taskQueue}, MyWorkflow)
	if err != nil {
		return fmt.Errorf("failed starting workflow: %w", err)
	}

	log.Printf("Sending signal in 2 seconds")
	time.Sleep(2 * time.Second)
	if err := c.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), "tick", nil); err != nil {
		return err
	}

	log.Printf("Stopping worker1 and starting worker2")
	worker1.Stop()
	worker2 := worker.New(c, taskQueue, worker.Options{})
	worker2.RegisterWorkflow(MyWorkflow)
	if err := worker2.Start(); err != nil {
		return fmt.Errorf("failed starting worker client: %w", err)
	}
	defer worker2.Stop()

	log.Printf("Sending signal in 2 seconds, then ending program in 10")
	time.Sleep(2 * time.Second)
[...]

PR #973, opened 2022-12-02
Reset SA and memos if execution context is stale

Quinn-With-Two-Ns:
Certain, previously constant now mutable fields, in the workflow info are not reset when a stale context is reset leading to them leaking into replays and causing non deterministic behaviour if those fields were used to make decision in the workflow that effected history.

closes https://github.com/temporalio/sdk-go/issues/970

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -666,6 +666,19 @@ func (w *workflowExecutionContextImpl) resetStateIfDestroyed(task *workflowservi
 				return err
 			}
 		}
+		if w.workflowInfo != nil {
+			// Reset the search attributes and memos from the WorkflowExecutionStartedEvent.
+			// The search attributes and memo may have been modified by calls like UpsertMemo
+			// or UpsertSearchAttributes. They must be reset to avoid non determinism on replay.
+			h := task.History
+			startedEvent := h.Events[0]
+			attributes := startedEvent.GetWorkflowExecutionStartedEventAttributes()
+			if attributes == nil {
+				return errors.New("first history event is not WorkflowExecutionStarted")
+			}
+			w.workflowInfo.SearchAttributes = attributes.SearchAttributes
+			w.workflowInfo.Memo = attributes.Memo
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


internal/internal_task_handlers.go
```diff
@@ -1156,10 +1155,30 @@ func (w *workflowExecutionContextImpl) ResetIfStale(task *workflowservice.PollWo
...
+		)
 		w.clearState()
-		return w.resetStateIfDestroyed(task, historyIterator)
+		err := w.resetStateIfDestroyed(task, historyIterator)
+		if err != nil {
+			return err
+		}
+		if w.workflowInfo != nil {
+			// Reset the search attributes and memos from the WorkflowExecutionStartedEvent.
+			// The search attributes and memo may have bben modified by calls like UpsertMemo
```
cretz, line 1167:
```suggestion
			// The search attributes and memo may have been modified by calls like UpsertMemo
```

cretz (approved):
LGTM, but would like @mfateev or @yiminc's review before merging to see if they can think of any compatibility concerns here.

Quinn-With-Two-Ns:
` sdk-features-test ` is unfortunately broken on go-sdk head because of an issue with the `schedules` API we had to change. I tested manually and all tests pass. Once we release a new version of the go-sdk I can fix it.

internal/internal_task_handlers.go
Quinn-With-Two-Ns: 😱

Quinn-With-Two-Ns: fixed

internal/internal_task_handlers.go
```diff
@@ -1156,10 +1155,30 @@ func (w *workflowExecutionContextImpl) ResetIfStale(task *workflowservice.PollWo
...
+			startedEvent := h.Events[0]
+			attributes := startedEvent.GetWorkflowExecutionStartedEventAttributes()
+			// The first event in the history has to be WorkflowExecutionStartedEvent because
+			// w.clearState() should "destroy" the state and w.resetStateIfDestroyed() will reset the
+			// history of the execution context if it is considered "destroyed".
+			if attributes == nil {
+				return errors.New("first history event is not WorkflowExecutionStarted")
+			}
+			w.workflowInfo.SearchAttributes = attributes.SearchAttributes
+			w.workflowInfo.Memo = attributes.Memo
```
yiminc, line 1179: I think this should be done in resetStateIfDestroyed() or even further in resetHistory().

Quinn-With-Two-Ns: Done

cretz: Can I get a bit more clarity as to "why" here? Specifically, what cases is it stale but not destroyed that we don't want to reset search attributes and memo?

Issue #971, opened 2022-12-01
Move most interceptor/proxy-based code gen to api-go


cretz:
**Is your feature request related to a problem? Please describe.**

Right now our SDK relies on every API message which is not cool

**Describe the solution you'd like**

We need to move the code gen for proxy and interceptor to https://github.com/temporalio/api-go.

Specifically in api-go we should add a `cmd/generateproxy` binary package that:

* Does what `internal/cmd/generateproxy` does today _and_ what `internal/cmd/generateinterceptor` does (can just be two
  methods called from top-level)
* Make the `generateinterceptor` not rely on known package set (if you have to use the file system to walk the known
  packages so be it). We want to be future proof for new packages.
* Give this its own `go.mod` so that it can depend on `golang.org/x/tools/go` without forcing api-go to add that
  directory.
* Runs after protoc everytime api-go is regenerated

That tool should generate/overwrite package `go.temporal.io/api/proxy` (so basically just a `proxy` dir in `api-go`). It
should generate the following Go public API (can be broken up across files if needed) for the proxy:

```go
package proxy

type WorkflowServiceProxyOptions struct {
  Client workflowservice.WorkflowServiceClient
}

func NewWorkflowServiceProxyServer(options WorkflowServiceProxyOptions) (workflowservice.WorkflowServiceServer, error) {
  // ...
}
```

This is exactly what's in `client` of the SDK today. As for the interceptor side, we need:

```go
package proxy

type VisitPayloadOptions struct {
  Visitor func(*common.Payload) error
  SkipSearchAttributes bool
}

func VisitPayloads(msg proto.Message, options VisitPayloadOptions) error {
  // ...
}

type PayloadVisitorInterceptorOptions struct {
  Outbound *VisitPayloadOptions
  Inbound *VisitPayloadOptions
}

func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grpc.UnaryClientInterceptor, error) {
  // ...
}
```

Do what is needed behind the scenes to make that work. You'll want the unexported recursive function to take a pointer to those `VisitPayloadOptions` btw so as not to copy each recursion.

Finally, just change the SDK to use this. The `client` package calls can just delegate to this. The `converter` stuff
can just delegate to this but make the payload codec encode/decode as the visitor.

**Describe alternatives you've considered**

Reflection, but decided against.

PR #974, opened 2022-12-05
Rewrite service proxy and interceptor to use new api-go code.

robholland:
## Why?

#971

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


client/service_proxy.go
```diff
@@ -22,253 +22,16 @@
...
-	return wh.client.UpdateSchedule(ctx, req)
-}
-
-func (wh *workflowServiceProxyServer) UpdateWorkerBuildIdOrdering(ctx context.Context, req *workflowservice.UpdateWorkerBuildIdOrderingRequest) (*workflowservice.UpdateWorkerBuildIdOrderingResponse, error) {
-	return wh.client.UpdateWorkerBuildIdOrdering(ctx, req)
-}
-
-func (wh *workflowServiceProxyServer) UpdateWorkflow(ctx context.Context, req *workflowservice.UpdateWorkflowRequest) (*workflowservice.UpdateWorkflowResponse, error) {
-	return wh.client.UpdateWorkflow(ctx, req)
+func NewWorkflowServiceProxyServer(options proxy.WorkflowServiceProxyOptions) (workflowservice.WorkflowServiceServer, error) {
```
cretz, line 35: Hrmm, technically a BC break. If we're gonna wrap the `proxy` package's `NewWorkflowSeriviceProxyServer` call, we might as well translate options.

cretz (approved):
Awesome. Let's wait on API go to be tagged of course

client/service_proxy.go
robholland: Yeah, couldn't decide here. Now put the client type back and casting.

cretz: > and casting.

Not cast, just transform

converter/grpc_interceptor.go
```diff
@@ -55,1360 +45,78 @@ type PayloadCodecGRPCClientInterceptorOptions struct {
...
-func (s *serviceInterceptor) decodePayloads(payloadspb *commonpb.Payloads) error {
-	var err error
-	payloads := payloadspb.Payloads
-	for _, codec := range s.codecs {
-		if payloads, err = codec.Decode(payloads); err != nil {
-			return err
-		}
+// NewFailureGRPCClientInterceptor returns a GRPC Client Interceptor that will mimic the encoding
+// that the SDK system would perform when configured with a FailureConverter with the EncodeCommonAttributes option set.
+func NewFailureGRPCClientInterceptor(options NewFailureGRPCClientInterceptorOptions) (grpc.UnaryClientInterceptor, error) {
```
cretz, line 91: May want to add doc that says "If using with NewPayloadCodecGRPCClientInterceptor, make sure this interceptor comes first in the chain"

converter/grpc_interceptor.go
```diff
@@ -55,1360 +45,78 @@ type PayloadCodecGRPCClientInterceptorOptions struct {
...
-	payloads := payloadspb.Payloads
-	for i := len(s.codecs) - 1; i >= 0; i-- {
-		if payloads, err = s.codecs[i].Encode(payloads); err != nil {
-			return err
-		}
-	}
-	payloadspb.Payloads = payloads
-	return nil
+type NewFailureGRPCClientInterceptorOptions struct {
+	// DataConverter ??
```
cretz, line 79: Hrmm, yes, we should accept this and use the default if not provided

converter/grpc_interceptor.go
```diff
@@ -55,1360 +45,78 @@ type PayloadCodecGRPCClientInterceptorOptions struct {
...
+	dc := GetDefaultDataConverter()
+	return proxy.NewFailureVisitorInterceptor(proxy.FailureVisitorInterceptorOptions{
+		Outbound: &proxy.VisitFailuresOptions{
+			Visitor: func(vpc *proxy.VisitFailuresContext, failure *failurepb.Failure) error {
+				var err error
+				failure.EncodedAttributes, err = dc.ToPayload(encodedFailure{
+					Message:    failure.Message,
+					StackTrace: failure.StackTrace,
+				})
+				return err
```
cretz, line 105: You also have to set the existing `Message` to `"Encoded failure"` and unset the `StackTrace`.

It may be best if, even though small, you expose a helper from `internal/failure_converter.go` that that and this call to encode/decode common attributes.

converter/grpc_interceptor.go
```diff
@@ -71,3 +74,49 @@ func NewPayloadCodecGRPCClientInterceptor(options PayloadCodecGRPCClientIntercep
...
+			},
+		},
+		Inbound: &proxy.VisitFailuresOptions{
+			Visitor: func(vpc *proxy.VisitFailuresContext, failure *failurepb.Failure) error {
+				var ea encodedFailure
+				err := dc.FromPayload(failure.EncodedAttributes, &ea)
```
cretz, line 111: Should not assume this will always be set. Some older workflow histories may come through this proxy without having this set. Just check if present and try to decode.

Take this logic from `internal/failure_decoder.go` which just ignores if it can't decode the attributes. (or even better, make a small helper you both share)

converter/failure_converter.go
```diff
@@ -34,3 +34,32 @@ type FailureConverter interface {
 	// FailureToError converts a Failure proto message to a Go Error.
 	FailureToError(failure *failurepb.Failure) error
 }
+
+type encodedFailure struct {
+	Message    string `json:"message"`
+	StackTrace string `json:"stack_trace"`
+}
+
+func EncodeCommonFailureAttributes(dc DataConverter, failure *failurepb.Failure) error {
```
cretz, line 43: Add Godoc to exported functions

internal/failure_converter.go
```diff
@@ -186,11 +181,7 @@ func (dfc *DefaultFailureConverter) FailureToError(failure *failurepb.Failure) e
 	}
 
 	if failure.GetEncodedAttributes() != nil {
-		var ea encodedFailure
```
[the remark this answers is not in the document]
robholland: It's not, removed now.

converter/grpc_interceptor_test.go
```diff
@@ -62,118 +58,6 @@ func payloadEncoding(payloads *commonpb.Payloads) string {
 	return string(payloads.Payloads[0].Metadata[MetadataEncoding])
 }
 
-func TestServiceInterceptorRequests(t *testing.T) {
```
cretz, line 65: Can we have some test to confirm that the failure conversion interceptor does as expected?

robholland: Yep sorry I missed that. Added now.

cretz (approved):
LGTM. As mentioned in other approval, will wait for api-go to be tagged and updated in go.mod here.

robholland:
https://github.com/temporalio/api-go/pull/100 now merged, just waiting on a new release before updating this.

cretz:
@robholland - https://github.com/temporalio/api-go has now been tagged as 1.14.0 whenever you want to update this

PR #979, opened 2022-12-08
Add EnableLoggingInReplay flag to WorkflowReplayerOptions for propagation

mattpercy-anz:
## What was changed
Added the flag EnableLoggingInReplay to the WorkflowReplayerOptions so that it can be passed through to the workerExecutionParameters. There is no way to set it (from what I can see) while running a workflow replay.

## Why?

Would like the ability to see more information around the workflow replay execution, which is currently hidden without this flag being set.

## Checklist

How was this tested:
Tested locally that with this flag set, we see additional workflow related logging


cretz (approved):
Makes perfect sense to want a replayer to log. Thanks for the contribution!

PR #981, opened 2022-12-12
Update sdk-go for latest interaction API

mmcshane:
## What was changed

- Adds UpdateWorkflowExecution to the client API and relevant interceptors
- Adopts expectation that interaction invocations will be attached to a PollWorkflowTaskQueueResponse (had previously been attached to the WFT Started event).
- Comment formatting for Go 1.19+

## Why?

Part of sync update project. Note that most of the relevant APIs are still in /internal/. Client api is made public only because there is an assertion that client.Client and internal.Client are compatible interfaces.

## Checklist

1. Closes 

2. How was this tested:

Unit tests & sdk-features

3. Any docs updates needed?

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -157,7 +157,7 @@ type (
 
 	// ScheduleDescription describes the current Schedule details from ScheduleHandle.Describe.
 	// NOTE: Experimental
-	ScheduleDescription = internal.ScheduleDescription 
+	ScheduleDescription = internal.ScheduleDescription
 
 	// Schedule describes a created schedule.
 	// NOTE: Experimental
@@ -186,6 +186,11 @@ type (
 	// ScheduleBackfillOptions configure the parameters for backfilling a schedule.
 	// NOTE: Experimental
 	ScheduleBackfillOptions = internal.ScheduleBackfillOptions
+
+	// UpdateWorkflowExecutionOptions encapsulates the optional parameters for
+	// sending an update to a workflow execution.
+	// NOTE: Experimental
+	UpdateWorkflowExecutionOptions = internal.UpdateWorkflowExecutionOptions
 
 	// Client is the client for starting and getting information about a workflow executions as well as
 	// completing activities asynchronously.
@@ -477,6 +482,11 @@ type (
 		// API. If the check fails, an error is returned.
 		CheckHealth(ctx context.Context, request *CheckHealthRequest) (*CheckHealthResponse, error)
 
+		// UpdateWorkflowExecution issues an update request to the specified
+		// workflow execution and returns the result synchronously.
+		// NOTE: Experimental
+		UpdateWorkflowExecution(ctx context.Context, workflowID string, updateName string, args []interface{}, opts UpdateWorkflowExecutionOptions) (converter.EncodedValue, error)
+
 		// WorkflowService provides access to the underlying gRPC service. This should only be used for advanced use cases
 		// that cannot be accomplished via other Client methods. Unlike calls to other Client methods, calls directly to the
 		// service are not configured with internal semantics such as automatic retries.
@@ -611,8 +621,9 @@ var _ internal.NamespaceClient = NamespaceClient(nil)
 // User had Activity.RecordHeartbeat(ctx, "my-heartbeat") and then got response from calling Client.DescribeWorkflowExecution.
 // The response contains binary field PendingActivityInfo.HeartbeatDetails,
 // which can be decoded by using:
-//   var result string // This need to be same type as the one passed to RecordHeartbeat
-//   NewValue(data).Get(&result)
+//
+//	var result string // This need to be same type as the one passed to RecordHeartbeat
+//	NewValue(data).Get(&result)
 func NewValue(data *commonpb.Payloads) converter.EncodedValue {
 	return internal.NewValue(data)
 }
@@ -621,9 +632,10 @@ func NewValue(data *commonpb.Payloads) converter.EncodedValue {
 // User had Activity.RecordHeartbeat(ctx, "my-heartbeat", 123) and then got response from calling Client.DescribeWorkflowExecution.
 // The response contains binary field PendingActivityInfo.HeartbeatDetails,
 // which can be decoded by using:
-//   var result1 string
-//   var result2 int // These need to be same type as those arguments passed to RecordHeartbeat
-//   NewValues(data).Get(&result1, &result2)
+//
+//	var result1 string
+//	var result2 int // These need to be same type as those arguments passed to RecordHeartbeat
+//	NewValues(data).Get(&result1, &result2)
 func NewValues(data *commonpb.Payloads) converter.EncodedValues {
 	return internal.NewValues(data)
 }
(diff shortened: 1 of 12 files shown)


client/client.go
```diff
@@ -477,6 +482,11 @@ type (
 		// API. If the check fails, an error is returned.
 		CheckHealth(ctx context.Context, request *CheckHealthRequest) (*CheckHealthResponse, error)
 
+		// UpdateWorkflowExecution issues an update request to the specified
+		// workflow execution and returns the result synchronously.
+		// NOTE: Experimental
+		UpdateWorkflowExecution(ctx context.Context, workflowID string, updateName string, args []interface{}, opts UpdateWorkflowExecutionOptions) (converter.EncodedValue, error)
```
cretz, line 488: Signal and query both accept can-be-empty `runID` option in the signature. I wonder if we should too.

Now I'm wondering, with first execution run ID being the only option that should appear in that struct (runID can be in signature and headers are interceptor only), and since cancel/terminate _also_ offer first execution run ID in their proto request and yet never allow a user to set it, is there any value for options here at all? We can then use varargs (so nice) and we will match signal, query, cancel, terminate, etc.

I think if we want to come back one day and add first execution run ID, we'd do it to multiple calls, not just this one.

So, despite our Slack convo and against known future proofing, I'm not on "team no options". Thoughts?

internal/client.go
```diff
@@ -441,6 +446,15 @@ type (
 		CheckCallAllowed(ctx context.Context, method string, req, reply interface{}) error
 	}
 
+	// UpdateWorkflowExecutionOptions encapsulates the optional parameters for
+	// sending an update to a workflow execution.
+	// NOTE: Experimental
+	UpdateWorkflowExecutionOptions struct {
+		FirstExecutionRunID string
+		RunID               string
+		Header              *commonpb.Header
```
cretz, line 455: In other calls, we keep the header hidden from non-interceptor API

client/client.go
```diff
@@ -611,8 +621,9 @@ var _ internal.NamespaceClient = NamespaceClient(nil)
 // User had Activity.RecordHeartbeat(ctx, "my-heartbeat") and then got response from calling Client.DescribeWorkflowExecution.
 // The response contains binary field PendingActivityInfo.HeartbeatDetails,
 // which can be decoded by using:
-//   var result string // This need to be same type as the one passed to RecordHeartbeat
-//   NewValue(data).Get(&result)
+//
+//	var result string // This need to be same type as the one passed to RecordHeartbeat
+//	NewValue(data).Get(&result)
```
Sushisource, line 626: Aren't these ones (and in a number of other places) supposed to be indented since they're code blocks? Won't this dedent not render them that way?

internal/internal_workflow_client.go
```diff
@@ -877,6 +878,42 @@ func (wc *WorkflowClient) ResetWorkflowExecution(ctx context.Context, request *w
...
+	argPayloads, err := wc.dataConverter.ToPayloads(args...)
+	if err != nil {
+		return nil, err
+	}
+
+	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
+	defer cancel()
+	resp, err := wc.workflowService.UpdateWorkflow(grpcCtx, &workflowservice.UpdateWorkflowRequest{
+		RequestId:         uuid.New(),
+		ResultAccessStyle: enumspb.WORKFLOW_UPDATE_RESULT_ACCESS_STYLE_REQUIRE_INLINE,
```
Sushisource, line 894: Worth adding a note about why this is currently fixed

client/client.go
mmcshane: It's hard to see but the "v" in `var` is under the "h" in `which` above. They're tab indented so they do end up rendering properly.

image

internal/client.go
```diff
@@ -441,6 +446,15 @@ type (
 		CheckCallAllowed(ctx context.Context, method string, req, reply interface{}) error
 	}
 
+	// UpdateWorkflowExecutionOptions encapsulates the optional parameters for
+	// sending an update to a workflow execution.
+	// NOTE: Experimental
+	UpdateWorkflowExecutionOptions struct {
+		FirstExecutionRunID string
```
bergundy, line 453: Where is update ID provided? Are you adding this later when we switch to the new protocol?

mmcshane: Yes, this PR is interaction based which was from the RequestID design. Next PR to API brings in client-specified UpdateID

client/client.go
mmcshane: Yeah, I was going go make a note of that. I don't like the difference but I also don't like optional fields so both have downsides and I'm not strongly attached to either approach. 

If we don't allow FirstExecutionRunID, what are the chances we regret it in the future? I'm not familiar enough with how often it's used in practice. I guess if we end with ...interface{} then we could also allow for well-known option types for extensibility in the future. Gross but at least it's possible.

internal/client.go
mmcshane: I took this from QueryWorkflowWithOptionsRequest which does have Header. But I'm not married to it if you'd prefer we hide it away.

client/client.go
cretz: > If we don't allow FirstExecutionRunID, what are the chances we regret it in the future?

We haven't regretted it yet on cancel or terminate

> I guess if we end with ...interface{} then we could also allow for well-known option types for extensibility in the future. Gross but at least it's possible.

Gross indeed. Another option of course is an overload, e.g. `UpdateWorkflowExecutionWithOptions`. This is what we had to do with `QueryWorkflow`. Note there is a `QueryWorkflowWithOptions` and it _only_ takes options and has everything inside of there. I'd support that. So maybe:

```go
UpdateWorkflow(ctx context.Context, workflowID string, runID string, updateName string, args ...interface{}) (converter.EncodedValue, error)

UpdateWorkflowWithOptions(ctx context.Context, req *UpdateWorkflowWithOptionsRequest) (*UpdateWorkflowWithOptionsResponse, error)
```

I think this, while admittedly ugly, is the best experience for users at the cost of just one overload.

Note I also changed `UpdateWorkflowExecution` to just `UpdateWorkflow` to match start, cancel, terminate, signal, and query (and deviating from describe and reset by intention, those are bad APIs that should never be exposed w/ that raw proto mess).

internal/client.go
cretz: Lets leave it if we take the `UpdateWorkflowWithOptions` approach we took here with query

client/client.go
mmcshane: I'm fine with the two function approach. I wish we had don functional args way back when.

FYI the gRPC API definition will be keeping the "Execution" part. But we already have prior art for diverging from the service's naming patterns with workflowservice.SignalWorkflowExecution vs. client.SignalWorkflow.

cretz: Definitely want "Execution" on the gRPC request call itself, that very much matches the others.

mmcshane:
I'm going to pull down this PR and rework it in favor of something that supports async update invocation from day 1

PR #987, opened 2022-12-25
Start version numbers at 0 in godoc

lorensr:
Since `DefaultVersion` is -1: https://pkg.go.dev/go.temporal.io/sdk@v1.19.0/internal#DefaultVersion

Matches https://github.com/temporalio/documentation/pull/1820


Quinn-With-Two-Ns (approved):

cretz (approved):
Had to update the issue title. It scared me at first :-)

PR #989, opened 2022-12-29
Client APIs for workflow update

mmcshane:
## What was changed

- Adds support for issuing workflow updates via the SDK client API.
- Adopts expectation that interaction invocations will be attached to a PollWorkflowTaskQueueResponse (had previously been attached to the WFT Started event).

## Why?

New update feature. Note that this PR is an extension of https://github.com/temporalio/sdk-go/pull/981 to support an async hande-based approach from the first release

## Checklist

1. Closes 

2. How was this tested:

unit tests and sdk-features

3. Any docs updates needed?


internal/internal_workflow_client.go
```diff
@@ -877,6 +927,69 @@ func (wc *WorkflowClient) ResetWorkflowExecution(ctx context.Context, request *w
 	return resp, nil
 }
 
+func (wc *WorkflowClient) UpdateWorkflowWithOptions(
```
bergundy, line 930: This should be interceptable too. I think it's probably better to make this the intercepted method and `UpdateWorkflow` call it.

bergundy: Hmm.. I see we did the same with `QueryWorkflow`, that seems like a bug.
Let's leave as is for this PR and discuss with @cretz / @Quinn-With-Two-Ns.

internal/internal_workflow_client.go
```diff
@@ -734,6 +735,55 @@ func (wc *WorkflowClient) QueryWorkflow(ctx context.Context, workflowID string,
...
+type WorkflowUpdateHandle interface {
+	// UpdateRef() *updatepb.UpdateRef //once API lands
+	Get(ctx context.Context, valuePtrs ...interface{}) error
+}
+
+// updateHandle is a dumb implementation of WorkflowExecutionUpdateHandle that
+// only supports the case where the output value is aready known at construction
+// time.
+type updateHandle struct {
+	value interface{} // converter.EncodedValue | error
```
bergundy, line 784: nit: I personally prefer a strongly typed version of this that saves the type switch but I don't have a strong opinion.

internal/internal_worker.go
```diff
@@ -1210,6 +1211,23 @@ func (aw *WorkflowReplayer) ReplayWorkflowExecution(ctx context.Context, service
...
 }
 
+// inferInvocations extracts the set of *interactionpb.Invocation objects that
+// should be attached to a workflow task (i.e. the
+// PollWorkflowTaskQueueResponse.Interactions) if that task were to carry the
+// provided slice of history events.
+func inferInvocations(events []*historypb.HistoryEvent) []*interactionpb.Invocation {
+	var invocations []*interactionpb.Invocation
+	for _, e := range events {
+		if attrs := e.GetWorkflowUpdateAcceptedEventAttributes(); attrs != nil {
```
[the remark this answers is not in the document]
mmcshane: An update is always accepted or rejected as part of the workflow task processing on the task that carries the initial request. Rejections are (for now) never made durable so the only evidence that an update was ever invoked is the AcceptedEvent. Since this function is inferring the existence of invocations that's what we check here.

internal/internal_workflow_client.go
```diff
@@ -734,6 +735,55 @@ func (wc *WorkflowClient) QueryWorkflow(ctx context.Context, workflowID string,
...
+	// this isn't upstream in API yet
+	// WaitFor enumspb.WorkflowExecutionUpdateWaitEvent
+}
+
+// WorkflowUpdateHandle is a handle to a workflow execution update
+// process. The update may or may not have completed so an instance of this type
+// functions simlarly to a Future with respect to the result value of the
+// update. If the update is rejected or returns an error, the Get function on
+// this type will return that error.
+type WorkflowUpdateHandle interface {
```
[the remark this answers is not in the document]
mmcshane: I don't think we're ready to require Go 1.18, are we? @cretz ?

internal/internal_workflow_client.go
```diff
@@ -734,6 +735,55 @@ func (wc *WorkflowClient) QueryWorkflow(ctx context.Context, workflowID string,
...
+	// UpdateRef() *updatepb.UpdateRef //once API lands
+	Get(ctx context.Context, valuePtrs ...interface{}) error
```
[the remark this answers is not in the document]
mmcshane: You're right - I forgot that I had enforced the update func return type to be either (1) just an error or (2) a serializable value and an error. I'll update.

client/client.go
```diff
@@ -477,6 +487,25 @@ type (
 		// API. If the check fails, an error is returned.
 		CheckHealth(ctx context.Context, request *CheckHealthRequest) (*CheckHealthResponse, error)
 
+		// UpdateWorkflow issues an update request to the specified
+		// workflow execution and returns the result synchronously. Calling this
+		// function is equivalent to calling UpdateWorkflowOptions with
+		// the same arguments and indicating that the RPC call should wait for
+		// completion of the update process.
+		// NOTE: Experimental
```
cretz, line 495:
```suggestion
		//
		// WARNING!! This API is experimental and subject to incompatible changes.
```

Wonder if we want to make this a little louder.

internal/interceptor.go
```diff
@@ -239,7 +239,8 @@ type WorkflowOutboundInterceptor interface {
 	// SetQueryHandler intercepts workflow.SetQueryHandler.
 	SetQueryHandler(ctx Context, queryType string, handler interface{}) error
 
-	SetUpdateHandler(ctx Context, updateName string, handler interface{}, opts UpdateOptions) error
+	// SetUpdateHandler intercepts workflow.SetUpdateHandler.
```
cretz, line 242: Let's toss experimental warnings on this stuff too

internal/internal_workflow_client.go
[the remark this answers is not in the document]
cretz: I mean, I think we are, but that's unrelated to @bergundy's point. He means "generic" I think in the English sense not the programmer sense :-)

@bergundy - I don't think this needs to be a generic future. There will be more update-specific calls here later for doing things like describing the update. We can abstract to a shared future interface later if there is value, but there is no value in doing it now. Do you have a specific reason for a generic future besides that it just feels right in your OO mind?

internal/internal_workflow_client.go
cretz: I agree, I think two mutually exclusive fields for success and failure here is much more reasonable.

internal/internal_workflow_client.go
cretz: It is interceptable. This is the overload that the simpler form calls. But the interceptor is called for the most advanced of the calls. In some cases where it makes sense we deviate from the client interface names in the interceptor interface In this case. If we added a `SignalWorkflowWithOptions` it'd go through `ClientOutboundInterceptor.SignalWorkflow` too.

EDIT: Actually I just realized we forgot to make `QueryWorkflowWithOptions` go through the interceptor. We should fix that. It is the intention. I have opened https://github.com/temporalio/sdk-go/issues/991.

internal/internal_workflow_client.go
```diff
@@ -877,6 +927,69 @@ func (wc *WorkflowClient) ResetWorkflowExecution(ctx context.Context, request *w
...
+			Header: header,
+			Name:   req.UpdateName,
+			Args:   argPayloads,
+		},
+	})
+	if err != nil {
+		return &updateHandle{value: err}, nil
+	}
+	if failure := resp.Output.GetFailure(); failure != nil {
+		return &updateHandle{value: wc.failureConverter.FailureToError(failure)}, nil
```
cretz, line 968: For queries, on the server side we differentiate all other errors from user-defined worker errors with a "query failed error" gRPC status details. See https://github.com/temporalio/api-go/blob/8a0bd676b97b61e2f8bd254389a3c9b96ed278b3/serviceerror/convert.go#L114. It seems here we are intentionally not reusing gRPC errors for update failed errors, which is fine, but I think we need to make sure somewhere that the caller can differentiate from update failure and all other server failures. For queries today they can check if it's `serviceerror.QueryFailed`. We need something similar for updates too (even if the error is defined SDK side instead of API side).

internal/internal_workflow_client.go
```diff
@@ -877,6 +927,69 @@ func (wc *WorkflowClient) ResetWorkflowExecution(ctx context.Context, request *w
...
+	}
+	argPayloads, err := wc.dataConverter.ToPayloads(req.Args...)
+	if err != nil {
+		return nil, err
+	}
+	header, _ := headerPropagated(ctx, wc.contextPropagators)
+
+	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
+	defer cancel()
+	resp, err := wc.workflowService.UpdateWorkflow(grpcCtx, &workflowservice.UpdateWorkflowRequest{
```
cretz, line 945: I think we made a mistake having query interceptor call query with options. All client query/update calls should go through the interceptor IMO. I opened https://github.com/temporalio/sdk-go/issues/991 for this just now on queries. For this, we should go ahead and move this logic to inside the interceptor and call the interceptor from here and just call this overload from the non-options overload.

workflow/workflow.go
```diff
@@ -360,37 +381,46 @@ func GetVersion(ctx Context, changeID string, minSupported, maxSupported Version
...
+//	}
 func SetQueryHandler(ctx Context, queryType string, handler interface{}) error {
 	return internal.SetQueryHandler(ctx, queryType, handler)
 }
 
+func SetUpdateHandler(ctx Context, updateName string, handler interface{}) error {
+	return SetUpdateHandlerOpts(ctx, updateName, handler, UpdateHandlerOptions{})
+}
+
+func SetUpdateHandlerOpts(ctx Context, updateName string, handler interface{}, opts UpdateHandlerOptions) error {
```
cretz, line 420: These two functions need docs including large experimental warnings. Also, I think we should call this `SetUpdateHandlerWithOptions`.

workflow/workflow.go
```diff
@@ -360,37 +381,46 @@ func GetVersion(ctx Context, changeID string, minSupported, maxSupported Version
...
+//	    return err
+//	  }
+//	  currentState = "done"
+//	  return nil
+//	}
 func SetQueryHandler(ctx Context, queryType string, handler interface{}) error {
 	return internal.SetQueryHandler(ctx, queryType, handler)
 }
 
+func SetUpdateHandler(ctx Context, updateName string, handler interface{}) error {
```
mmcshane, line 416: Oops - need godocs here and below. Working on it.

workflow/workflow.go
cretz, line 420: Question - about how far out are we from having an integration test or two for updates? Would totally support _only_ using SDK features if you'd like though that often requires release coordination so I understand just an integration test or two in here.

internal/internal_workflow_client.go
mmcshane: Switched to a product type

workflow/workflow.go
mmcshane: Server hasn't stabilized to the point where putting integration tests in this repo would be a good idea. There are sdk-feature tests but they can't land in main until this PR lands since they depend on public API stuff.

internal/internal_workflow_client.go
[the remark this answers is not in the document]
bergundy: I can see more methods being added here when we add an ability to get a handle for an existing update.
We'll probably want to be able to wait for acceptance for example.
Describing an update seems reasonable to me too TBH but probably not as useful on handles returned by the update-initiating APIs.

For the current functionality, I don't see a justification in returning the specialized handle.
My reason is more YAGNI (for this flow) than "doesn't feel right".

internal/internal_worker.go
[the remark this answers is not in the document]
bergundy: Okay, I went over the protocol and realized that error returned from the update handler ends up as completed with a failure result variant.

So I guess we could save the extra event when the handler completes immediately but that's just an optimization and is probably not worth it right now.

internal/internal_workflow_client.go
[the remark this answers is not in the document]
cretz: >  My reason is more YAGNI (for this flow) than "doesn't feel right".

"You Aren't Gonna Need It" where "It" = "generic future abstraction". The describing of an update is likely gonna be needed as is cancelling, a getter for workflow ID, etc. `WorkflowRun` isn't a generic future either. And since Go is duck typed, you can type this in your code:

```go
type Future interface {
  Get(context.Context, ...interface{}) error
}
```

And now you have a generic future for `WorkflowUpdateHandle` and `WorkflowRun`. But it's silly to make some generic thing in a non-generic situation. This is a update handle very simply.

client/client.go
cretz: Hrmm, I guess if there's so many we should at least do it with the entry points. Probably weren't loud enough there. But this is fine too, no need to change anything.

internal/internal_worker.go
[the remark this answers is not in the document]
mmcshane: Yes, in this discussion in the past we decided it was preferable to always have the accepted event, even if it could be implied in some instances by the completed event.

internal/internal_workflow_client.go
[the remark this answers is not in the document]
mmcshane: We need a getter for ID and WFRunID (because it's otherwise not knowable if it's not specified at invocation time). That's why there's a getter commented out on the line immediately under the line we're all commenting on. If the type was available in API today it would be included in this PR. It is known and understood to be required to re-hydrate update handles. We will need it to ship even a pre-release.

internal/internal_workflow_client.go
mmcshane: Updated to funnel both Update variants into the outbound interceptor and added a test to the update branch in sdk-features.

internal/interceptor.go
mmcshane: Yes, and added comments too - I missed these initially.

internal/internal_workflow_client.go
mmcshane: Mentioned elsewhere but yes, I just pushed this to the PR.

internal/internal_workflow_client.go
mmcshane: Not there yet with the server implementation but I take your point. :+1:

internal/internal_workflow_client.go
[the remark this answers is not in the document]
cretz: :+1: We could go ahead and add `GetID()` right now since you have it, but yeah no use in it except for logging yet until we have `GetUpdate` or whatever in the future.

internal/internal_workflow_client.go
cretz: Is the server impl expected to set unique error details that represent worker-side update failure? If so and we add a new type to https://pkg.go.dev/go.temporal.io/api/serviceerror to support that, no prob. But then why even have a `Output.GetFailure()` instead of just failing the gRPC call if you're always gonna fail the gRPC call in a specific way in this case? I don't mind either way and other SDKs will be making their own `WorkflowUpdateFailed` errors that differentiate.

cretz: But this isn't a gRPC failure, so the server side won't help you completely. You're not gonna add a new type in https://pkg.go.dev/go.temporal.io/api/serviceerror because that's only for gRPC failures right? Since this is not a gRPC failure, are you expecting to create a update-failure type in https://github.com/temporalio/api/blob/master/temporal/api/failure/v1/message.proto? Whether you do or not, https://pkg.go.dev/go.temporal.io/sdk/temporal will need a `WorkflowUpdateError` type.

Basically the caller of this call should know whether it was a worker-side update failure just like the caller of queries know it's a worker-side query failure. I don't mind how we implement it, but if you keep it as not a gRPC error, you're gonna need a type in this SDK repo to represent it.

internal/interceptor.go
cretz: Still missing at least a `NOTE: Experimental` or something on these.

cretz: Still missing at least a `NOTE: Experimental` or something on these.

cretz (approved):
LGTM, just a minor note on adding experimental note to the interceptor stuff

PR #992, opened 2023-01-03
Fix schedule docs

Quinn-With-Two-Ns:
Noticed a few typos in the schedule docs

### DIFF
--- a/internal/schedule_client.go
+++ b/internal/schedule_client.go
@@ -51,7 +51,7 @@ type (
 	// ScheduleCalendarSpec is an event specification relative to the calendar, similar to a traditional cron specification.
 	// A timestamp matches if at least one range of each field matches the
 	// corresponding fields of the timestamp, except for year: if year is missing,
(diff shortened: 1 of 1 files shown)


internal/schedule_client.go
```diff
@@ -181,7 +181,7 @@ type (
 
 		// Skip - Any matching times will be skipped.
 		//
-		// All aspects of the schedulepb.ScheduleCalendarSpec—including seconds—must match a time for the time to be skipped.
+		// All aspects of the ScheduleCalendarSpec—including seconds—must match a time for the time to be skipped.
```
dnr, line 184: maybe:
```suggestion
		// All fields of the ScheduleCalendarSpec—including seconds—must match a time for the time to be skipped.
```

dnr (approved):

cretz (approved):

internal/schedule_client.go
Quinn-With-Two-Ns: good point will fix

Issue #994, opened 2023-01-04
Cannot set workflow ID for replayer


Drahflow:
We have a workflow which derives various child workflow execution IDs by appending subtask names (`"-start-session"` in the example below) to the current workflow execution ID. Upon replaying those histories during CI to check for non-determinism problems, we see (re-wrapped for your convenience):
```
2023/01/04 13:33:12 ERROR Workflow panic
    WorkflowType Main
    WorkflowID ReplayId
    RunID 2bc48c60-ee48-4b19-99fb-0656c4d30a99
    Attempt 1 Error unknown command
    CommandType: ChildWorkflow, ID: merchant-consumer-workflow/ed472a6d-b793-501c-af06-a501158487b6-start-session,
    possible causes are nondeterministic workflow definition code or incompatible change in the workflow definition
    StackTrace process event for Merchant [panic]:
go.temporal.io/sdk/internal.panicIllegalState(...)
        /home/drahflow/go/pkg/mod/go.temporal.io/sdk@v1.19.0/internal/internal_command_state_machine.go:532
...
```
as the re-derived child workflow IDs will now (i.e. during replay) be based upon the `ReplayId` string used as the workflow execution ID here:
https://github.com/temporalio/sdk-go/blob/06e474c93e936b71dc4afcec973460b22c13986d/internal/internal_worker.go#L1235

In reality however, there is nothing non-deterministic about such a derivation, as the workflow execution ID is fixed for a given execution.

A possible solution would be to allow providing a custom execution ID (in practice derived from the recorded history filename) when replaying a history.

----

A workaround is relatively straightforward via recording the own execution ID in a side-effect, but that seems extremely unnecessary.

cretz:
Yeah workflow ID, unlike run ID, cannot currently be derived from history. Unlike our other SDKs, the Go SDK does not offer a way to set the workflow ID on the replayer. I have altered the title of this issue.

Issue #865, opened 2022-07-21
Replayer uses different run ID than runtime after reset


cretz:
## Expected Behavior

They use the same run ID

## Actual Behavior

Today, replayer uses `temporal.api.history.v1.WorkflowExecutionStartedEventAttributes.original_execution_run_id` while runtime uses `temporal.api.workflowservice.v1.workflow_execution.run_id`. This run ID affects the default child workflow ID. We should change the replayer (changing the runtime is not acceptable for compatibility reasons of course).

PR #998, opened 2023-01-05
Add option to override replayer execution info

Quinn-With-Two-Ns:
Add an option to the replayer to set workflow ID and run ID

* closes https://github.com/temporalio/sdk-go/issues/994

* closes https://github.com/temporalio/sdk-go/issues/865

### DIFF
--- a/internal/internal_worker_test.go
+++ b/internal/internal_worker_test.go
@@ -1379,6 +1379,24 @@ func testReplayWorkflowSideEffect(ctx Context) error {
 	return nil
 }
 
+func testReplayRunID(ctx Context) error {
+	info := GetWorkflowInfo(ctx)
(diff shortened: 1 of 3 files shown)


internal/internal_worker_test.go
```diff
@@ -1465,6 +1483,54 @@ func (s *internalWorkerTestSuite) TestReplayWorkflowHistory_SideEffect() {
 	require.NoError(s.T(), err)
 }
 
+func (s *internalWorkerTestSuite) TestReplayWorkflowHistory_Bleh() {
```
cretz, line 1486: Bleh? :-)

cretz (approved):
LGTM

internal/internal_worker_test.go
Quinn-With-Two-Ns: naming hard

Issue #947, opened 2022-11-06
Panic: Error adding duplicate command on duplicated child workflow


mfateev:
## Expected Behavior
An attempt to start a child workflow with the same ID as the one already running should result in `ChildWorkflowExecutionAlreadyStartedError`.

## Actual Behavior
```
2022/11/06 11:12:07 ERROR Workflow panic Namespace default TaskQueue child-workflow WorkerID 8968@Maxims-MacBook-Pro.local@ WorkflowType SampleParentWorkflow WorkflowID parent-workflow_ed5bbdc7-183d-43b7-936a-d2a9e4d9bd2f RunID b4a73c0b-a1d3-4ef9-ab29-8bfd34b4138d Attempt 1 Error adding duplicate command CommandType: ChildWorkflow, ID: ABC-SIMPLE-CHILD-WORKFLOW-ID, state=Created, isDone()=false, history=[Created] StackTrace coroutine root [panic]:
go.temporal.io/sdk/internal.panicIllegalState(...)
        /Users/maxim/go/pkg/mod/go.temporal.io/sdk@v1.17.0/internal/internal_command_state_machine.go:502
go.temporal.io/sdk/internal.(*commandsHelper).addCommand(0x14000554230, {0x1016656f0, 0x140000a3290})
        /Users/maxim/go/pkg/mod/go.temporal.io/sdk@v1.17.0/internal/internal_command_state_machine.go:998 +0x2f8
go.temporal.io/sdk/internal.(*commandsHelper).startChildWorkflowExecution(0x41b8?, 0x102180668?)
        /Users/maxim/go/pkg/mod/go.temporal.io/sdk@v1.17.0/internal/internal_command_state_machine.go:1228 +0x38
go.temporal.io/sdk/internal.(*workflowEnvironmentImpl).ExecuteChildWorkflow(_, {{{0x140001b21e0, 0xe}, 0x0, 0x0, 0x0, {0x1013796f7, 0x7}, {0x10138d56c, 0x1c}, ...}, ...}, ...)
        /Users/maxim/go/pkg/mod/go.temporal.io/sdk@v1.17.0/internal/internal_event_handlers.go:476 +0x3b0
go.temporal.io/sdk/internal.(*workflowEnvironmentInterceptor).ExecuteChildWorkflow(0x140005506c0?, {0x10165fff8?, 0x1400017e000}, {0x1017c363a, 0x13}, {0x140000a3200, 0x1, 0x1})
[...]

PR #999, opened 2023-01-05
Fix panic on starting child workflows with duplicate IDs

Sushisource:
## What was changed
Fix this panic. See the added docstring in the PR for more.

## Why?
Panic bad, yo.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/947

2. How was this tested:
Added integ test, existing tests.

3. Any docs updates needed?

### DIFF
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -164,6 +164,12 @@ type (
 	stateMachineIllegalStatePanic struct {
 		message string
 	}
+
+	// Error returned when a child workflow with the same id already exists and hasn't completed
+	// and been removed from internal state.
+	childWorkflowExistsWithId struct {
+		id string
+	}
 )
 
 const (
@@ -1253,10 +1259,18 @@ func (h *commandsHelper) recordMutableSideEffectMarker(mutableSideEffectID strin
 	return command
 }
 
-func (h *commandsHelper) startChildWorkflowExecution(attributes *commandpb.StartChildWorkflowExecutionCommandAttributes) commandStateMachine {
+// startChildWorkflowExecution can return an error in the event that there is already a child wf
+// with the same ID which exists as a command in memory. Other SDKs actually will send this command
+// to server, and have it reject it - but here the command ID is exactly equal to the child's wf ID,
+// and changing that without potentially blowing up backwards compatability is difficult. So we
+// return the error eagerly locally, which is at least an improvement on panicking.
+func (h *commandsHelper) startChildWorkflowExecution(attributes *commandpb.StartChildWorkflowExecutionCommandAttributes) (commandStateMachine, error) {
 	command := h.newChildWorkflowCommandStateMachine(attributes)
+	if h.commands[command.getID()] != nil {
+		return nil, &childWorkflowExistsWithId{id: attributes.WorkflowId}
+	}
 	h.addCommand(command)
-	return command
+	return command, nil
 }
 
 func (h *commandsHelper) handleStartChildWorkflowExecutionInitiated(workflowID string) {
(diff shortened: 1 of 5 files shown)


internal/internal_command_state_machine.go
```diff
@@ -1253,10 +1259,18 @@ func (h *commandsHelper) recordMutableSideEffectMarker(mutableSideEffectID strin
...
-func (h *commandsHelper) startChildWorkflowExecution(attributes *commandpb.StartChildWorkflowExecutionCommandAttributes) commandStateMachine {
+// startChildWorkflowExecution can return an error in the event that there is already a child wf
+// with the same ID which exists as a command in memory. Other SDKs actually will send this command
+// to server, and have it reject it - but here the command ID is exactly equal to the child's wf ID,
+// and changing that without potentially blowing up backwards compatability is difficult. So we
+// return the error eagerly locally, which is at least an improvement on panicking.
+func (h *commandsHelper) startChildWorkflowExecution(attributes *commandpb.StartChildWorkflowExecutionCommandAttributes) (commandStateMachine, error) {
 	command := h.newChildWorkflowCommandStateMachine(attributes)
+	if h.commands[command.getID()] != nil {
+		return nil, &childWorkflowExistsWithId{id: attributes.WorkflowId}
```
Sushisource, line 1270: I can't say I love this, but as is often the case in the Go SDK it seemed like the least offensive of some unfortunate options.

test/workflow_test.go
```diff
@@ -520,6 +520,29 @@ func (w *Workflows) ChildWorkflowCancelUnusualTransitionsRepro(ctx workflow.Cont
...
+	err := child1.GetChildWorkflowExecution().Get(ctx, &childWE)
+	if err != nil {
+		return err
+	}
+	workflow.SignalExternalWorkflow(ctx, childWE.ID, childWE.RunID, "unblock", nil)
+	if err != nil {
+		return err
+	}
+	err = workflow.ExecuteChildWorkflow(childCtx, w.childWorkflowWaitOnSignal).Get(ctx, nil)
+	if _, ok := err.(*internal.ChildWorkflowExecutionAlreadyStartedError); !ok {
```
Quinn-With-Two-Ns, line 540: Shouldn't we expose this error outside of internal?

internal/internal_command_state_machine.go
```diff
@@ -164,6 +164,12 @@ type (
 	stateMachineIllegalStatePanic struct {
 		message string
 	}
+
+	// Error returned when a child workflow with the same id already exists and hasn't completed
+	// and been removed from internal state.
+	childWorkflowExistsWithId struct {
```
Quinn-With-Two-Ns, line 170: Why create a new error instead of using `ChildWorkflowExecutionAlreadyStartedError`?

Sushisource: That's a proto - so it doesn't impl the `Error` interface and feels icky to use in this place

Quinn-With-Two-Ns:
Once we have sdk versioning could we align the go sdk with the other sdks since we will be able to make a backwards incompatible change like changing the ID

test/workflow_test.go
Sushisource: Yeah, so, it is one of the many things that's aliased. I can't refer to the `temporal` version of it here b/c circular import. It's aliased here: https://github.com/temporalio/sdk-go/blob/8ab62d9e211fe86c3f5a5723f6c5e79a65f083fd/temporal/error.go#L135

I don't track the whole aliasing stuff so if I need to somehow return it differently I'm all ears.

Sushisource:
> Once we have sdk versioning could we align the go sdk with the other sdks since we will be able to make a backwards incompatible change like changing the ID

At that point we'll want to just move to be on top of Core. I think we should invest the minimum amount of time possible into the state machine internals here.

test/workflow_test.go
Quinn-With-Two-Ns: How come other errors can be used from temporal https://github.com/temporalio/sdk-go/blob/master/test/workflow_test.go#L220 but this one can't?

bergundy (approved):
I agree this is better than panicking.
We'll change to reach the server if we ever use Core to power this SDK.

test/workflow_test.go
Sushisource: Hahah, derp. I didn't realize this comment was on the _test_ file. Fixed.

Quinn-With-Two-Ns (approved):

cretz:
Note, this still occurs with activity ID. Also, technically, one should be allowed to try to start a child workflow of an ID that has not yet finished in the workflow, but I think this is fine.

PR #1006, opened 2023-01-17
Convert SDK to protocol/messages approach

mmcshane:
For review, NOT merge. I want to get eyes on this so that when we merge the API change upstream we're immediately ready with these changes.  Ignore the go.mod/go.sum redirects and conflicts.

Some code moved around as I tried to consolidate update code into update.go and protocol tracking code into the internal/protocol package.


internal/update.go
```diff
@@ -0,0 +1,354 @@
...
+		name       string
+	}
+)
+
+// newUpdateResponder constructs an updateProtocolResponder instance to handle
+// update callbacks.
+func newUpdateProtocol(
+	protoInstanceID string,
+	scheduleUpdate func(name string, args *commonpb.Payloads, header *commonpb.Header, callbacks UpdateCallbacks),
+	env updateEnv) *updateProtocol {
```
cretz, line 83:
```suggestion
	env updateEnv,
) *updateProtocol {
```

A bit clearer IMO

internal/update.go
```diff
@@ -0,0 +1,354 @@
...
+	)
+	if err != nil {
+		callbacks.Reject(fmt.Errorf("unable to decode the input for update %q: %w", name, err))
+		return
+	}
+	input := UpdateInput{Name: name, Args: args}
+
+	spawn(ctx, name, func(ctx Context) {
+		envInterceptor := getWorkflowEnvironmentInterceptor(ctx)
+		if !IsReplaying(ctx) {
```
cretz, line 217: Reminder: We'll want a test in SDK features or somewhere to confirm that making a validation return failure on replay is ok because validator never called on replay. Not only am I unsure when this `IsReplaying` is set, but it tells other SDKs they have to make sure to support no-validate-on-replay too.

internal/update.go
```diff
@@ -0,0 +1,354 @@
...
+		reflect.TypeOf(handler.fn),
+		serializedArgs,
+	)
+	if err != nil {
+		callbacks.Reject(fmt.Errorf("unable to decode the input for update %q: %w", name, err))
+		return
+	}
+	input := UpdateInput{Name: name, Args: args}
+
+	spawn(ctx, name, func(ctx Context) {
```
cretz, line 215: Reminder: We'll want SDK features tests for panic in validation handler and panic in update. Obviously with Go code they are the same (wft failure), but it forces other SDKs to acknowledge their unexpected-error paths too.

cretz (approved):
I only have pedantic stuff. This LGTM. I take comfort knowing this is experimental and therefore if we missed something internally here we can fix it. I think this should remain experimental until implemented in other languages.

internal/internal_event_handlers.go
```diff
@@ -1686,3 +1643,20 @@ func (weh *workflowExecutionEventHandlerImpl) handleSignalExternalWorkflowExecut
 
 	return nil
 }
+
+func (weh *workflowExecutionEventHandlerImpl) protocolConstructorForMessage(
+	msg *protocolpb.Message,
+) (func() protocol.Instance, error) {
+	protoName, err := protocol.NameFromMessage(msg)
```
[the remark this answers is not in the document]
mmcshane: Fair point. The saving grace here is that it's all internal. The reason I put the registry in place (it's not strictly needed - the Update protocol instance could be moved into the Couroutine and execute to completion there) was to leave some aspect of a blueprint for the next protocol to come along if I'm not the implementer.

internal/update.go
```diff
@@ -0,0 +1,350 @@
+package internal
```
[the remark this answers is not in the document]
mmcshane: Argh. You're probably right. I don't care for it but yeah.

internal/update.go
mmcshane: Agreed. Not committing yours only because this whole thing is going to get wiped when the new upstream API comes in.

internal/update.go
mmcshane: FWIW I tested this by hand and it does work in the way we want. Doing it in features will require some finesse but I agree that it's worth it from a pseudo-documentation perspective.

internal/update.go
mmcshane: Sure. And FYI they're in the unit tests too.

internal/update.go
cretz: > Doing it in features will require some finesse

I think the best way may be to have a (bad) validator that increments some global counter and returns error if > 1, then after workflow complete, force a replay (e.g. query or something).

internal/internal_event_handlers.go
```diff
@@ -292,6 +286,21 @@ func (s *scheduledSignal) handle(result *commonpb.Payloads, err error) {
 	s.callback(result, err)
 }
 
+func (wc *workflowEnvironmentImpl) drainOutbox(sink *[]*protocolpb.Message) {
```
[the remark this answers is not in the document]
mmcshane: I think I'm going to keep this one. Chasing private member variables from outside a type - even within a package - gives me The Fear.

PR #1009, opened 2023-01-25
Update API with eager workflow dispatch definitions

bergundy:
NOTE: This PR does not implement eager workflow dispatch for the SDK, it simply updates the API module.


bergundy:
Looks like this was already updated in #1006

PR #1013, opened 2023-01-26
Prevent cache eviction race

mmcshane:
## What was changed

In some server tests the race detector noticed that SetPreviousStartedEventID can race with cache removal which happens on the cache eviction thread.

## Why?

Data races bad

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):

alexshtin (approved):

cretz:
Interesting, surprised it didn't come up before.

PR #1015, opened 2023-01-28
Deliver messages even when events are skipped

mmcshane:
## What was changed

Here we switch to delivering all messages dependent on events IDs less than or equal to the current event ID.

## Why?

We don't run WFTScheduledEvent instances through standard event handling so we were missing messages dependent on events of that type

## Checklist

1. Closes 

2. How was this tested:

Unit & features

3. Any docs updates needed?

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -815,19 +815,6 @@ processWorkflowLoop:
 	return
 }
 
-// indexInvocations builds a map of the interaction invocations contained in a
-// workflow task where the index is the interaction's event level (i.e. the
-// event after which the interaction can execute) and the value is the
-// invocation itself. This function may return an empty map but it will not
-// return nil.
-func indexMessages(workflowTask *workflowTask) map[int64][]*protocolpb.Message {
-	out := map[int64][]*protocolpb.Message{}
-	for _, msg := range workflowTask.task.Messages {
-		out[msg.GetEventId()] = append(out[msg.GetEventId()], msg)
-	}
-	return out
-}
-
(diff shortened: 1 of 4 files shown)


internal/internal_task_handlers.go
```diff
@@ -841,7 +828,7 @@ func (w *workflowExecutionContextImpl) ProcessWorkflowTask(workflowTask *workflo
 	var replayCommands []*commandpb.Command
 	var respondEvents []*historypb.HistoryEvent
 
-	msgs := indexMessages(workflowTask)
+	msgs := indexMessagesByEventID(workflowTask.task.GetMessages())
```
cretz, line 831: Would it have value to ask the server to always deliver these in event ID order?

Also, is the number of these so great that you really need to sort because of fear of the cost of `takeLTE`? I think you can just make `takeLTE` iterate over all messages every time and https://github.com/golang/go/wiki/SliceTricks#filter-in-place to keep the ones greater than while extracting the ones less than or equal.

So something like this (untested, just typed here in comments):

```go
func (emi *eventMsgIndex) takeLTE(eventID int64) (out []*protocolpb.Message) {
	n := 0
	for _, msg := range *emi {
		if msg.GetEventID() > eventID {
			(*emi)[n] = msg
			n++
		} else {
			out = append(out, msg)
		}
	}
	*emi = *emi[:n]
	return;
}
```

mmcshane: I'm fine with doing a little more work in takeLTE. I would expect the number of messages per WFT to be low (which to be fair, also makes sorting cheap) but there may be a substantial number when we rebundle into a bigger WFT for replay.

cretz (approved):

PR #1017, opened 2023-01-30
Correct updateHandle.Get when passed nil pointer

mmcshane:
## What was changed

Match the behavior of Future, which is to return the error value (which
is likely nil) in such cases and not to try to assign to nil.

## Why?

Consistent with Future and allows for updates that do not return a value. 

## Checklist

1. Closes 

2. How was this tested:

Unit test and features test

3. Any docs updates needed?

No

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -1663,11 +1663,7 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
 			},
 		},
 	})
-	handle := &updateHandle{
-		ref: &updatepb.UpdateRef{
-			WorkflowExecution: wfexec,
-			UpdateId:          in.UpdateID,
-		}}
+	handle := &updateHandle{ref: resp.UpdateRef}
 	if err != nil {
 		handle.err = err
 		return handle, nil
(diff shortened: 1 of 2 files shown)


internal/internal_workflow_client.go
```diff
@@ -1699,7 +1695,12 @@ func (uh *updateHandle) UpdateID() string {
 }
 
 func (uh *updateHandle) Get(ctx context.Context, valuePtr interface{}) error {
-	if uh.err != nil {
+	// implementation note: this is a broken implementation that assumes the
+	// update has already completed by the time Get is called. This is true for
+	// the current implementation of synchronous updates but will not be true in
+	// the future with async update.
+
+	if uh.err != nil || valuePtr == nil {
```
cretz, line 1703: To confirm, what you're saying here is that you acknowledge that in the future when I pass `nil` for `valuePtr`, you'll change this code to properly wait on the update? (what the `valuePtr` value is only applies _after_ completion)

mmcshane: Yes. The semantics should be exactly the same as Future.Get to make things as unsurprising as possible.

cretz: :+1: Will try to set a reminder for when non-sync update comes around

cretz (approved):

PR #1019, opened 2023-01-30
Add ContextPropagators option for replay tests

jlegrone:
## What was changed

This PR enables setting `ContextPropagators` on workflow replay options, which can currently only be set on `client.Options` passed to a worker.

## Why?

Context propagators can extract information from workflow headers and modify workflow contexts. This is important for workflow replay in some cases, because information in a `workflow.Context` may be used to modify workflow behavior, eg. workflow code that performs role based access control based on claims in workflow headers.

## Checklist

How was this tested:

This was tested in our internal environment where workflow replay was previously failing due to missing context propagators.


cretz (approved):
Awesome, thanks

Issue #996, opened 2023-01-05
Opt-in to having stack trace query use custom data converter


cretz:
**Describe the solution you'd like**

Make a new worker option like `StackTraceQueryUsesConverter bool`. If false, the `__stack_trace` query response should not be converted/encoded/etc, but if `true` it should be. We accept that this changes today's default behavior so we need to make it loud and clear on release notes.

cretz:
We need to discuss how this affects the proxy now in api-go. Do we need to have that stop encoding this specific query?

PR #1020, opened 2023-01-30
Disable encoding of built in queries by default

Quinn-With-Two-Ns:
Disable encoding of built in queries by default like java. Since go sdk also has the open sessions built in query I generalized the parameter and applied it to that as well since I couldn't see a reason to treat it any different.

Note: this is a breaking change as before stack traces always uses the user provided data converter.

resolves: https://github.com/temporalio/sdk-go/issues/996

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -133,9 +133,10 @@ type (
 		queryHandler    func(queryType string, queryArgs *commonpb.Payloads, header *commonpb.Header) (*commonpb.Payloads, error)
 		updateHandler   func(name string, args *commonpb.Payloads, header *commonpb.Header, callbacks UpdateCallbacks)
 
-		logger                log.Logger
-		isReplay              bool // flag to indicate if workflow is in replay mode
-		enableLoggingInReplay bool // flag to indicate if workflow should enable logging in replay mode
+		logger                     log.Logger
+		isReplay                   bool // flag to indicate if workflow is in replay mode
+		enableLoggingInReplay      bool // flag to indicate if workflow should enable logging in replay mode
+		enableEncodeBuiltInQueries bool // flag to indicate if built in queries like stacktrace queries should use the provided data converter or the default
 
 		metricsHandler           metrics.Handler
 		registry                 *registry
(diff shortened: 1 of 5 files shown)


internal/internal_event_handlers.go
```diff
@@ -1072,9 +1075,17 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessQuery(
 ) (*commonpb.Payloads, error) {
 	switch queryType {
 	case QueryTypeStackTrace:
-		return weh.encodeArg(weh.StackTrace())
+		if weh.enableEncodeBuiltInQueries {
```
cretz, line 1078: So I just asked the UI team how they know when to pass a payload to the remote codec. They said they pass all non-search-attributes to remote codecs. We ask people to pass through unencoded payloads in their codecs, but we don't necessarily require it.

Is there a concern that people whose remote codecs assume always-encoded and have worked in the UI today will now start breaking on stack traces? cc @bergundy, @Sushisource, and @Spikhalskiy. Just want confirmation that we can tell these users they need to fix their codec to pass through unencoded data. And if they come back with "but for security reasons we want to require encoding or fail" we are ok saying that they now need to restart their workers and opt-in with this boolean.

Quinn-With-Two-Ns: If they already have remote codecs setup then the easier fix is to enable this option no? Assuming they read release notes (big assumption to be fair) they shouldn't need to restart their workers.

cretz: > If they already have remote codecs setup then the easier fix is to enable this option no?

I agree, but it may not be obvious they need to enable this. So they'll go to the UI, see stack traces broken, and have to come back and redeploy all of their workers with this flag. I don't think it's that big of a deal, just want to get team confirmation here.

bergundy: I thought this was known that the codec should pass through anything it doesn't handle.
Could be worth documenting.

cretz: People intentionally don't want to do that in all cases. Some users want to intentionally fail if they see unencoded data as a way to catch unencoded data early. So they'll need to set this field. We need to be very loud about this field because there are users that are very concerned about encrypted data automatically becoming unencrypted (however unfounded). We have even been asked for manual/security reviews that all non-search payloads are properly encrypted.

Sushisource: I think the ideal situation would be able to set an option for the UI server like:

`payload_decoding: PassNonstandardToRemote | AlwaysPassToRemote` 

With the first being the default. This way users in the situation Chad talks about can choose to pass everything if they're paranoid (and can choose to do nothing if needed), but normally all known-decodable Payloads get handled locally.

Ofc, that's outside the scope of this change, but we could decide if we want to add that first to UI or not. Seems like it'd be an easy change.

bergundy: Another option would be to add a metadata field to this payload to mark that it is intentionally not encoded or have the UI provide more context to the remote codec server.

We need to make the proxy configurable to bypass stack trace queries which complicates this feature (granted we'll need to have it opt out for SDK generated metadata on WFT completion but the condition there is simpler).
Let's have a team discussion and see if we want to revert our decision to have this opt-in for encryption behavior.
Seems like we're adding a bunch of special cases where it might not be necessary.

Issue #1021, opened 2023-01-31
Expose session state to users


cretz:
**Is your feature request related to a problem? Please describe.**

People need a way to know whether session has failed, specifically if an activity has failed due to being cancelled by session failure.

**Describe the solution you'd like**

Can expose the already-existing session state on session info. We need to make sure that it is set to failed if an activity is failed because of session failure.

Quinn-With-Two-Ns:
How is this different then the error we already have in the SDK `ErrSessionFailed`?

>We need to make sure that it is set to failed if an activity is failed because of session failure.

How do you propose we tell an activity failed because it was the session that failed?

cretz:
> How is this different then the error we already have in the SDK ErrSessionFailed?

I believe based on docs that an already-in-progress activity on a session just gets cancelled, you can't tell from the error returned by the activity that it was due to session failure. (I could be wrong about this, did not test)

> How do you propose we tell an activity failed because it was the session that failed?

While there are technically some race conditions, if the activity failed due to cancellation the user can check if the session is failed and _may_ make reasonable a judgement that that may be the reason. It's about the best we can offer I think without changing error approaches.

Quinn-With-Two-Ns:
>While there are technically some race conditions, if the activity failed due to cancellation the user can check if the session is failed and may make reasonable a judgement that that may be the reason. It's about the best we can offer I think without changing error approaches.

I'd need to test but I think depending on worker and activity settings it could be pretty easy to have false positives and false negatives. I agree it is the best we can do so I am fine with this approach.

In your original comment you suggested we should mark the session failed if an activity is failed because of session failure. That I don't like because I don't think it is possible to reliably do and is a breaking change to the session API as we explicitly call out we don't fail the session on activity failure in the documentation.

cretz:
> In your original comment you suggested we should mark the session failed if an activity is failed because of session failure

I am not suggesting changing anything with how sessions are marked. I am just suggesting we expose session state/status. By "we need to make sure" I just mean a test to confirm that the session status is as expected when an in-flight activity cancel/failure is caused by session failure. Sorry for the wording confusion.

armoona:
From the user's perspective, it would be simplest if ExecuteActivity().Get() always returned ErrSessionFailed on session failure, whether it happened while an activity is in progress or otherwise.

Thinking out loud, I think the following hack would work now to reliably trigger session restarts in all cases. It uses a dummy activity that does nothing.

```
[...]

PR #1024, opened 2023-01-31
Expose session state

Quinn-With-Two-Ns:
Expose session state to users so they can understand if their activity failed because the session worker failed.

resolves:https://github.com/temporalio/sdk-go/issues/1021

### DIFF
--- a/workflow/session.go
+++ b/workflow/session.go
@@ -47,9 +47,20 @@ type (
 	SessionOptions = internal.SessionOptions
 )
 
-// ErrSessionFailed is the error returned when user tries to execute an activity but the
-// session it belongs to has already failed
-var ErrSessionFailed = internal.ErrSessionFailed
+var (
+	// ErrSessionFailed is the error returned when user tries to execute an activity but the
+	// session it belongs to has already failed
+	ErrSessionFailed = internal.ErrSessionFailed
+
+	// SessionStateOpen means the session worker is heartbeating and new activities will be schedule on the session host.
+	SessionStateOpen = internal.SessionStateOpen
+
+	// SessionStateClosed means the session was closed by the workflow and new activities will not be scheduled on the session host.
+	SessionStateClosed = internal.SessionStateClosed
+
+	// SessionStateFailed means the session worker was detected to be down and the session cannot be used to schedule new activities.
+	SessionStateFailed = internal.SessionStateFailed
+)
 
 // Note: Worker should be configured to process session. To do this, set the following
 // fields in WorkerOptions:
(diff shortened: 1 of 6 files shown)


workflow/session.go
```diff
@@ -47,9 +47,20 @@ type (
...
+	ErrSessionFailed = internal.ErrSessionFailed
+
+	// SessionStateOpen means the session worker is heartbeating and new activities will be schedule on the session host.
+	SessionStateOpen = internal.SessionStateOpen
+
+	// SessionStateClosed means the session was closed by the workflow and new activities will not be scheduled on the session host.
+	SessionStateClosed = internal.SessionStateClosed
+
+	// SessionStateFailed means the session worker was detected to be down and the session cannot be used to schedule new activities.
+	SessionStateFailed = internal.SessionStateFailed
```
cretz, line 62: How about the `SessionState` type itself?

cretz (commented):
For this one we might want a test that confirms that session state shows as failed if an in-flight activity fails due to session failure. My concern is something crashes the session

Not sure how such a test looks (an integration test or maybe a "features" test that shuts the session worker down).

workflow/session.go
Quinn-With-Two-Ns: hmmm yes I guess I should. Originally I didn't think I needed to, but If someone wants to make a function that takes the state as a parameter it would be useful to expose.

cretz: I just want to make sure this session state var is updated on worker crash before activity failure is returned. If not and the session appears open but activity fails, we may need to document that.

Quinn-With-Two-Ns:
>For this one we might want a test that confirms that session state shows as failed if an in-flight activity fails due to session failure. My concern is something crashes the session

I can make an integration test for this.

I will also note again it is pretty easy, I think, to create a scenario where a session worker fails, but the activity fails firsts so the workflow thinks the session is still open. I think we should document this behaviour.

cretz:
> I will also note again it is pretty easy, I think, to create a scenario where a session worker fails, but the activity fails firsts so the workflow thinks the session is still open. I think we should document this behaviour.

Hrmm, that a common scenario? I wonder if that makes this whole issue moot. If the user needs to know that their activity was cancelled due to session failure, it seems we can't do that. Sure we can show activity was cancelled due to session _complete_ but not session workflow failure.

Do we have to document that session `HeartbeatTimeout` has to be less than half that of an activity heartbeat timeout for this to be reported correctly? Or do we just tell users the session may still show open after activity failure due to session worker failure for the duration of the session heartbeat timeout? Or do we just scrap this issue due to these caveats and just document that knowing whether an activity failed due to session failure is unknowable?

cretz (approved):
Good enough for me, but we need to let that user know that they need to understand what they want to consider a "session failure" because it may be an activity failure on crash before the session reports failure.

PR #1026, opened 2023-02-01
Workaround for updates delivered in first WFT

mmcshane:
## What was changed

Adds a conditional yield point during update processing to allow the scheduler to do other things like registering update handlers.

## Why?

If the first WFT includes an update we can have a problem because the workflow function itself hasn't run yet to register update handlers and the update will be rejected. Adding this yield allows the scheduler to execute the workflow function up to the first blocking point, at which point the handler(s) will be registered by user code.

## Checklist

1. Closes 

2. How was this tested:

Unit test here and I was able to remove all the time.Sleep(1*time.Second) from the features tests.

3. Any docs updates needed?

No

### DIFF
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -232,7 +232,7 @@ func defaultUpdateHandler(
 		// because this update is part of the first workflow task and is being
 		// delivered before the workflow function itself has run and had a
 		// chance to register update handlers) then we yield control back to the
-		// scheduler to allow handler registration to occur. The sceduler will
+		// scheduler to allow handler registration to occur. The scheduler will
 		// resume this coroutine after others have run to a blocking point.
 		if len(eo.updateHandlers) == 0 {
 			scheduler.Yield(ctx, "yielding for initial handler registration")
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -65,6 +65,16 @@ type (
 		Complete(success interface{}, err error)
 	}
 
+	// UpdateScheduluer allows an update state machine to spawn coroutines and
+	// yield itself as necessary.
+	UpdateScheduler interface {
+		// Spawn starts a new named coroutine, executing the given function f.
+		Spawn(ctx Context, name string, f func(ctx Context)) Context
+
+		// Yield returns control to the scheduler.
+		Yield(ctx Context, status string)
+	}
+
 	// updateEnv encapsulates the utility functions needed by update protocol
 	// instance in order to implement the UpdateCallbacks interface. This
 	// interface is conveniently implemented by
@@ -207,7 +217,7 @@ func defaultUpdateHandler(
 	serializedArgs *commonpb.Payloads,
 	header *commonpb.Header,
 	callbacks UpdateCallbacks,
-	spawn func(Context, string, func(Context)) Context,
+	scheduler UpdateScheduler,
 ) {
 	env := getWorkflowEnvironment(rootCtx)
 	ctx, err := workflowContextWithHeaderPropagated(rootCtx, header, env.GetContextPropagators())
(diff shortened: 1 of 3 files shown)


mmcshane:
I don't love this - it's a little brittle and dependent on current scheduler design - but I haven't been able to figure out a better/cleaner way to achieve the same thing. Would love to hear some other suggestions.

internal/internal_update.go
```diff
@@ -207,37 +217,47 @@ func defaultUpdateHandler(
...
-		for k := range eo.updateHandlers {
-			keys = append(keys, k)
+	scheduler.Spawn(ctx, name, func(ctx Context) {
+		eo := getWorkflowEnvOptions(ctx)
+
+		// If we suspect that handler registration has not occurred (e.g.
+		// because this update is part of the first workflow task and is being
+		// delivered before the workflow function itself has run and had a
+		// chance to register update handlers) then we yield control back to the
+		// scheduler to allow handler registration to occur. The sceduler will
```
cretz, line 235:
```suggestion
		// scheduler to allow handler registration to occur. The scheduler will
```

internal/internal_update.go
```diff
@@ -207,37 +217,47 @@ func defaultUpdateHandler(
...
+		// resume this coroutine after others have run to a blocking point.
```
cretz, line 236: Hrmm. I am racking my brain trying to think of a downside to this. We will have to discuss what update-with-start looks like I think (just whether the update event coming before the workflow start event is even an issue). 

I can't think of any downsides to this off the top of my head. Even if there are multiple updates before workflow start I think a yield from each is safe. And I don't think yielding in every case of no handlers no matter how far into the workflow we are is harmful either (just pushing this to the end of the loop). Is there any change this yield could not get completed before next WFT? I know there's some "run until all yielded" code/logic before commands are sent as a WFT completion, so I want to confirm that this "yield" is not considered waiting on Temporal like the other yields.

I would like to request a update-before-worker-start integration test in the features repo (though all other languages offer pre-start handler registration unlike Go).

mmcshane: The downside for me is that this aspect of update delivery (admittedly an edge-case) becomes dependent on exactly how workflow execution happens. In particular on the fact that during startup the root coro yields exactly once before running the user-supplied workflow task.

As for the update-before-start integration test ... that's actually all of them! Prior to this PR, I always had to time.Sleep(1*time.Second) between launching the workflow and submitting the first update. Without that sleep, the first update sees an error returned "no registered handler for 'foo'" With this change, that sleep can be removed.

cretz: > The downside for me is that this aspect of update delivery (admittedly an edge-case) becomes dependent on the precise mechanism of how workflow execution happens and it does so in a somewhat invisible way.

This is the accepted cost of Go SDK's inability to predefine anything about a workflow before starting it. No worries (much cleaner in other langs)

> is not considered to have blocked [...] within the current call to ExecuteUntilAllBlocked 

Perfect, thanks

cretz (approved):

internal/internal_update.go
```diff
@@ -207,37 +217,47 @@ func defaultUpdateHandler(
...
+	scheduler.Spawn(ctx, name, func(ctx Context) {
+		eo := getWorkflowEnvOptions(ctx)
+
+		// If we suspect that handler registration has not occurred (e.g.
+		// because this update is part of the first workflow task and is being
+		// delivered before the workflow function itself has run and had a
+		// chance to register update handlers) then we yield control back to the
+		// scheduler to allow handler registration to occur. The scheduler will
+		// resume this coroutine after others have run to a blocking point.
+		if len(eo.updateHandlers) == 0 {
```
bergundy, line 237: I have a few of concerns here:

1. What are the delivery ordering semantics after adding this yield?
What if we have a signal-update-signal sequence?
With this solution, will the update necessarily be delivered after the signals?

2. What about the future async update guarantees?
Are we okay rejecting any updates if a handler is not registered by the end of a workflow task? This is different than signals, which are unidirectional and are okay to buffer.

3. What if user wants to run a local activity before registering a dynamic update handler?
Does using this one time yield mechanism prevent that case?

I'm leaning towards having a solution where we buffer updates and (at least with the fast path) we reject them at the end of workflow task processing.

PR #1027, opened 2023-02-01
Handle CanceledAfterStarted in CanceledEvent

Quinn-With-Two-Ns:
User reported this issue where I believe they tried to cancel a workflow that had a child workflow and it resulted in the parent workflow getting stuck panicing on invalid state transition on the `ChildWorkflowExecutionCanceled` because it didn't expect it to be in `CanceledAfterStarted`. I don't see why we would treat `CanceledAfterInitiated` differently than `CanceledAfterStarted` so I allowed the state transition.

Previously the SDK would consider this an invalid state transition and panic, now the SDK handles it as a canceled child workflow.

### DIFF
--- a/internal/internal_worker_test.go
+++ b/internal/internal_worker_test.go
@@ -1161,6 +1161,57 @@ func testReplayWorkflowCancelChildWorkflowUnusualOrdering(ctx Context) error {
 	return nil
 }
 
+func (s *internalWorkerTestSuite) TestReplayWorkflowHistory_ChildWorkflowCancellation_WhenWorkflowCanceled() {
+	taskQueue := "taskQueue1"
+	testEvents := []*historypb.HistoryEvent{
+		createTestEventWorkflowExecutionStarted(1, &historypb.WorkflowExecutionStartedEventAttributes{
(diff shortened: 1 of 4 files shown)


internal/internal_worker_test.go
```diff
@@ -1161,6 +1161,57 @@ func testReplayWorkflowCancelChildWorkflowUnusualOrdering(ctx Context) error {
 	return nil
 }
 
+func (s *internalWorkerTestSuite) TestReplayWorkflowHistory_ChildWorkflowCancellation_WhenWorkflowCanceled() {
```
cretz, line 1164: Can you replicate the failure in an integration test? Seems straightforward enough to wait for a child workflow to start and then cancel its parent and confirm the parent panicked/WFT-failed before this fix.

(sorry, I just trust integration tests much more than manual history creation)

Quinn-With-Two-Ns: No fair point I'll make an integration test, I will say this manual history is basically what the customers history looked like

cretz (approved):

PR #1029, opened 2023-02-02
Allow duplicate update requests

mmcshane:
## What was changed

These may be duplicated due to upstream retry behavior.

## Why?

If the specified update has already scheduled the request down toward the user workflow code, just drop any additional requests - they are by definition the same value or the client is being naughty.

## Checklist

1. Closes 

2. How was this tested:

Unit test modifications

3. Any docs updates needed?

No


bergundy (approved):

mmcshane:
Per @bergundy prefer to fail here

PR #1031, opened 2023-02-03
Pin staticcheck to v0.3.3

mmcshane:
## What was changed

Pin staticcheck to v0.3.3

## Why?

The latest version of staticcheck (0.4.0 as of 3 Feb 2023) scans dependencies (?) and incorrectly detects that the
proxy.WorkflowServiceProxyOptions.Client field is unused in api-go. We will pin to the previous version for now but be advised this version is known not to work with go1.20.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,12 @@ vet: $(ALL_SRC)
 	done;
 
 staticcheck: $(ALL_SRC)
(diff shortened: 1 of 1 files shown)


cretz (approved):
Approving with the expectation that we will move on from staticcheck soon

PR #1032, opened 2023-02-04
Add testsuite.StartDevServer that starts temporal CLI

bergundy:


testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	EnableUI bool
+	// Log format - defaults to "pretty"
+	LogFormat string
+	// Log level - defaults to "warn"
+	LogLevel string
+	// Additional arguments to the dev server.
+	ExtraArgs []string
+}
+
+type DevServer interface {
```
cretz, line 90: Godocs should be on all public constructs

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+
+	args := []string{
+		"server",
+		"start-dev",
+		"--ip", host, "--port", port,
+		"--namespace", clientOptions.Namespace,
+		"--dynamic-config-value", "frontend.enableServerVersionCheck=false",
+		"--dynamic-config-value", "system.enableEagerWorkflowStart=true",
+		"--dynamic-config-value", "system.enableEagerWorkflowStart=true",
```
cretz, line 136: Duplicate line

cretz, line 136: Why are we enabling eager workflow start on the dev server here? The dev server should match the non-dev server IMO. When we enable the non-dev server form of this we can enable the dev server form. They can opt in to this just like they do on the real server.

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func (opts *DevServerOptions) clientOptionsOrDefault() client.Options {
+	var out client.Options
+	if opts.ClientOptions != nil {
+		// Shallow copy the client options since we intend to overwrite some fields.
+		out = *opts.ClientOptions
+	} else {
+		out = client.Options{}
+	}
```
cretz, line 262:
```suggestion
	}
```

Redundant

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// Temporal CLI based DevServer
+type devServer struct {
+	cmd    *exec.Cmd
+	client client.Client
+}
+
+// StartDevServer starts a Temporal CLI dev server process. This may download the server if not already downloaded.
+func StartDevServer(ctx context.Context, options *DevServerOptions) (DevServer, error) {
```
cretz, line 104:
```suggestion
func StartDevServer(ctx context.Context, options DevServerOptions) (DevServer, error) {
```

And just override herein. It's normal for options structs to be by-value for copying (even if they are large).

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	// Accept nil because all options are "optional".
```
cretz, line 105: Even if all are optional, making the user pass in an instance is very Go like and matches our other SDK approaches.

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	if options == nil {
+		options = &DevServerOptions{}
+	}
+	clientOptions := options.clientOptionsOrDefault()
+
+	exePath, err := downloadIfNeeded(options, clientOptions.Logger)
```
cretz, line 111: Pass context here and use context when downloading so user can cancel

testsuite/devserver.go
cretz, line 90: Why an interface instead of just making the `devServer` exported? If this is potential reuse for test server, we need a better interface name, because `DevServerOptions` also uses "dev server" term but the args clearly wouldn't work for test server.

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	cmd := exec.Command(exePath, args...)
+	cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr
+	clientOptions.Logger.Info("Starting DevServer", "ExePath", exePath, "Args", args)
+	if err := cmd.Start(); err != nil {
+		return nil, fmt.Errorf("failed starting: %w", err)
+	}
+
+	returnedClient, err := waitServerReady(ctx, clientOptions)
+	clientOptions.Logger.Info("DevServer ready")
```
cretz, line 152: Should be after error check

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package testsuite contains unit testing framework for Temporal workflows and activities and a helper to download and
+// start a dev server.
```
cretz, line 24:
```suggestion
```

This only needs to be on one file

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	// Existing path on the filesystem for the executable.
+	ExistingPath string
+	// Download the executable if not already there.
+	CachedDownload CachedDownload
+}
+
+// Configuration for the dev server.
+type DevServerOptions struct {
+	// Path to executable or download info - defaults to cached download.
+	Exe EphemeralExe
```
cretz, line 72:
```suggestion
	// Existing path on the filesystem for the executable.
	ExistingExePath string
	// Download the executable if not already there. Ignored if ExistingExePath set.
	CachedDownload CachedDownload
```

No need for a separate `EphemeralExe` in Go since we can't enforce mutual exclusivity at runtime

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	// Client options used to create a client for the dev server.
+	// The provided Namespace or the "default" namespace is automatically registered on startup.
+	// If HostPort is provided, the host and port will be used to bind the server, otherwise the server will bind to
+	// localhost and obtain a free port.
+	ClientOptions *client.Options
+	// SQLite DB filename if persisting or non-persistent if none.
+	DBFilename string
```
cretz, line 79: I don't see this used below

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	// Whether to enable the UI.
+	EnableUI bool
```
cretz, line 81: I don't see this used below

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	// Log format - defaults to "pretty"
+	LogFormat string
+	// Log level - defaults to "warn"
+	LogLevel string
+	// Additional arguments to the dev server.
+	ExtraArgs []string
```
cretz, line 87: I don't see this used below

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	// Client options used to create a client for the dev server.
+	// The provided Namespace or the "default" namespace is automatically registered on startup.
+	// If HostPort is provided, the host and port will be used to bind the server, otherwise the server will bind to
+	// localhost and obtain a free port.
+	ClientOptions *client.Options
+	// SQLite DB filename if persisting or non-persistent if none.
+	DBFilename string
+	// Whether to enable the UI.
+	EnableUI bool
+	// Log format - defaults to "pretty"
```
cretz, line 82: Inconsistent with whether you use punctuation to end Godoc sentences.

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+		return "", fmt.Errorf("failed unmarshaling info: %w", err)
+	}
+
+	// Download and extract
+	logger.Info("Downloading temporal CLI", "Url", info.ArchiveURL, "ExePath", exePath)
+	resp, err = http.Get(info.ArchiveURL)
+	if err != nil {
+		return "", fmt.Errorf("failed downloading: %w", err)
+	}
+	defer resp.Body.Close()
```
cretz, line 221: After this line, check the HTTP call status before continuing

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	// Build info URL
+	platform := runtime.GOOS
+	if platform != "windows" && platform != "darwin" && platform != "linux" {
+		return "", fmt.Errorf("unsupported platform %v", platform)
+	}
+	arch := runtime.GOARCH
+	if arch != "amd64" && arch != "arm64" {
+		return "", fmt.Errorf("unsupported architecture %v", arch)
+	}
+	infoURL := fmt.Sprintf("https://temporal.download/cli/%v?platform=%v&arch=%v&sdk-name=sdk-go&sdk-version=%v", version, platform, arch, internal.SDKVersion)
```
cretz, line 194:
```suggestion
	infoURL := fmt.Sprintf("https://temporal.download/cli/%v?platform=%v&arch=%v&sdk-name=sdk-go&sdk-version=%v", url.QueryEscape(version), platform, arch, internal.SDKVersion)
```

(wasn't as much of a concern for our features repo as it is now)

testsuite/devserver.go
```diff
@@ -0,0 +1,356 @@
...
+	"time"
+
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/internal"
+	ilog "go.temporal.io/sdk/internal/log"
+	"go.temporal.io/sdk/log"
+)
+
+// Cached download of the dev server.
+type CachedDownload struct {
```
Quinn-With-Two-Ns, line 53: Why not just depend on temporalite and not need to do all this?

cretz: That adds a huge statically linked dependency on server. I do think in the future we will:

* Add a new "testing server" interface that this dev server impls
* Add a new testing server impl for the Java test server
* Add a https://pkg.go.dev/go.temporal.io/server/temporaltest that is like https://pkg.go.dev/github.com/temporalio/temporalite/temporaltest and also impls this testing server iface so users can choose to statically link server if they really want

testsuite/devserver.go
bergundy: I was debating this... sure.

testsuite/devserver.go
bergundy: Overlooked, thanks.

testsuite/devserver.go
bergundy: We agreed to return an interface to support `temporaltest`, Java test server, and CLI.

I might rename it though, just don't have a better name.

testsuite/devserver.go
bergundy: Yes, this was meant to enable eager activities but you're right that we should use the server defaults (except for the version check).

testsuite/devserver.go
bergundy: Yes, thanks

Issue #980, opened 2022-12-09
workflowcheck should allow fmt.Errorf and fmt.Sprintf and others


cretz:
## Expected Behavior

These are deterministic (or at least enough so these days that they should be ignored) and should not error

## Actual Behavior

Getting errors like:

```
calls non-deterministic function fmt.Errorf
    fmt.Errorf is non-deterministic, reason: calls non-deterministic function (*fmt.pp).doPrintf
      (*fmt.pp).doPrintf is non-deterministic, reason: calls non-deterministic function (*fmt.pp).printArg
        (*fmt.pp).printArg is non-deterministic, reason: calls non-deterministic function (*fmt.pp).fmtComplex
          (*fmt.pp).fmtComplex is non-deterministic, reason: calls non-deterministic function (*fmt.pp).badVerb
            (*fmt.pp).badVerb is non-deterministic, reason: calls non-deterministic function (*fmt.pp).printValue
              (*fmt.pp).printValue is non-deterministic, reason: calls non-deterministic function (reflect.Value).Bool
                (reflect.Value).Bool is non-deterministic, reason: calls non-deterministic function (reflect.Value).panicNotBool
                  (reflect.Value).panicNotBool is non-deterministic, reason: calls non-deterministic function (reflect.flag).mustBe
                    (reflect.flag).mustBe is non-deterministic, reason: calls non-deterministic function reflect.valueMethodName
                      reflect.valueMethodName is non-deterministic, reason: calls non-deterministic function (*runtime.Frames).Next
                        (*runtime.Frames).Next is non-deterministic, reason: calls non-deterministic function runtime.expandCgoFrames
                          runtime.expandCgoFrames is non-deterministic, reason: calls non-deterministic function runtime.gostring
                            runtime.gostring is non-deterministic, reason: calls non-deterministic function runtime.rawstring
                              runtime.rawstring is non-deterministic, reason: calls non-deterministic function runtime.mallocgc
                                runtime.mallocgc is non-deterministic, reason: calls non-deterministic function runtime.gcStart
                                  runtime.gcStart is non-deterministic, reason: calls non-deterministic function runtime.gcBgMarkStartWorkers
                                    runtime.gcBgMarkStartWorkers is non-deterministic, reason: starts goroutine
              (*fmt.pp).printValue is non-deterministic, reason: calls non-deterministic function internal/fmtsort.Sort
[...]

PR #1038, opened 2023-02-13
workflowcheck: mark some fmt functions as deterministic

Quinn-With-Two-Ns:
resolves: https://github.com/temporalio/sdk-go/issues/980

### DIFF
--- a/contrib/tools/workflowcheck/determinism/ident_refs.go
+++ b/contrib/tools/workflowcheck/determinism/ident_refs.go
@@ -38,10 +38,16 @@ var DefaultIdentRefs = IdentRefs{
 	// We mark these as deterministic since they give so many false positives
 	"(reflect.Value).Interface": false,
 	"runtime.Caller":            false,
+	"fmt.Append":                false,
+	"fmt.Appendf":               false,
+	"fmt.Appendln":              false,
 	"fmt.Errorf":                false,
 	"fmt.Sprintf":               false,
 	"fmt.Sprint":                false,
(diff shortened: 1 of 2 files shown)


contrib/tools/workflowcheck/determinism/ident_refs.go
```diff
@@ -38,6 +38,10 @@ var DefaultIdentRefs = IdentRefs{
 	// We mark these as deterministic since they give so many false positives
 	"(reflect.Value).Interface": false,
 	"runtime.Caller":            false,
+	"fmt.Errorf":                false,
+	"fmt.Sprintf":               false,
+	"fmt.Sprint":                false,
+	"fmt.Sprintln":              false,
```
cretz, line 44: May want to add each of these being called in a method in `contrib/tools/workflowcheck/determinism/testdata/src/a/calls.go` but not that big of a deal.

Also, may want to add the `Append` and `Sscan` funcs here too

cretz: Yeah, just for peace of mind, can we add the following to the bottom of `contrib/tools/workflowcheck/determinism/testdata/src/a/calls.go`:

```go
func SafeFmtCall() {
	fmt.Sprintf("foo bar")
}
```

Quinn-With-Two-Ns: sure

cretz (approved):

PR #1046, opened 2023-02-18
Local activity nonfirst attempt counts for metering

Sushisource:
## What was changed
Count non-first-WFT attempts during each WFT for each local activity

## Why?
In support of recent metering changes

## Checklist

1. Closes 

2. How was this tested:
Added integ test

3. Any docs updates needed?
Nothing Go specific


internal/internal_event_handlers.go
```diff
@@ -871,6 +874,32 @@ func (wc *workflowEnvironmentImpl) GetRegistry() *registry {
 	return wc.registry
 }
 
+// ResetLAWFTAttemptCounts resets the number of attempts in this WFT for all LAs to 0 - should be
+// called at the beginning of every WFT
+func (wc *workflowEnvironmentImpl) ResetLAWFTAttemptCounts() {
```
Quinn-With-Two-Ns, line 879: Can we write an integration test to cover this function as well? I think `TestSendsCorrectMeteringData` would still pass even if this function wasn't called.

test/integration_test.go
```diff
@@ -3693,6 +3693,53 @@ func (ts *IntegrationTestSuite) TestScheduleUpdateWorkflowActionMemo() {
...
+	ts.Nil(w.Start())
+
+	wfOpts := ts.startWorkflowOptions("test-sends-correct-metering-data")
+	wfOpts.WorkflowTaskTimeout = 2 * time.Second
+	ts.NoError(ts.executeWorkflowWithOption(wfOpts,
+		ts.workflows.WorkflowWithLocalActivityRetriesAndHeartbeat, nil))
+
+	ts.Equal(uint32(0), nonfirstLAAttemptCounts[0])
+	for i := 1; i < len(nonfirstLAAttemptCounts); i++ {
+		ts.True(nonfirstLAAttemptCounts[i] > 0)
```
Quinn-With-Two-Ns, line 3738: Can we not test the exact count?

Sushisource: We could, but there's a good chance it'd make it flaky I think. How many attempts happen per WFT is going to be somewhat timing sensitive.

internal/internal_event_handlers.go
Sushisource: It fails if the call is removed - otherwise the attempts which span a heartbeat are over-counted.

Quinn-With-Two-Ns (approved):

cretz (approved):

Issue #1040, opened 2023-02-15
ScheduleCalendarSpec should have a default for every time except year


Quinn-With-Two-Ns:
**Is your feature request related to a problem? Please describe.**
currently in `ScheduleCalendarSpec`
> For all fields besides year, at least one Range must be present to match anything.

This is difficult for users and is not what was implemented in other SDKs. The ask is to align with other SDKs like typescript and set unset ranges to the zero range when creating a schedule

PR #1047, opened 2023-02-19
Use sensible defaults for ScheduleCalendarSpec

Quinn-With-Two-Ns:
Align with python and typescript SDKs and use sensible defaults for ScheduleCalendarSpec to help users.

Defaults are taken from the python SDK.

Note: This is a breaking change because we are changing defaults, but schedules is marked as experimental and users have been asking for this

### DIFF
--- a/internal/internal_schedule_client.go
+++ b/internal/internal_schedule_client.go
@@ -719,9 +719,37 @@ func convertFromPBScheduleCalendarSpecList(calendarSpecPB []*schedulepb.Structur
 	return calendarSpec
 }
 
+func applyScheduleCalendarSpecDefault(calendarSpec *ScheduleCalendarSpec) {
+	if calendarSpec.Second == nil {
+		calendarSpec.Second = []ScheduleRange{{Start: 0}}
+	}
+
+	if calendarSpec.Minute == nil {
+		calendarSpec.Minute = []ScheduleRange{{Start: 0}}
+	}
+
+	if calendarSpec.Hour == nil {
+		calendarSpec.Hour = []ScheduleRange{{Start: 0}}
+	}
+
+	if calendarSpec.DayOfMonth == nil {
+		calendarSpec.DayOfMonth = []ScheduleRange{{Start: 1, End: 31}}
+	}
+
+	if calendarSpec.Month == nil {
+		calendarSpec.Month = []ScheduleRange{{Start: 1, End: 12}}
+	}
+
+	if calendarSpec.DayOfWeek == nil {
+		calendarSpec.DayOfWeek = []ScheduleRange{{Start: 0, End: 6}}
+	}
+}
+
 func convertToPBScheduleCalendarSpecList(calendarSpec []ScheduleCalendarSpec) []*schedulepb.StructuredCalendarSpec {
 	calendarSpecPB := make([]*schedulepb.StructuredCalendarSpec, len(calendarSpec))
 	for i, e := range calendarSpec {
+		applyScheduleCalendarSpecDefault(&e)
+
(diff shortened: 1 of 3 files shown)


Quinn-With-Two-Ns:
resolves: https://github.com/temporalio/sdk-go/issues/1040

internal/internal_schedule_client.go
```diff
@@ -719,9 +719,37 @@ func convertFromPBScheduleCalendarSpecList(calendarSpecPB []*schedulepb.Structur
...
+
+	if calendarSpec.DayOfWeek == nil {
+		calendarSpec.DayOfWeek = []ScheduleRange{{Start: 0, End: 6}}
+	}
+}
+
 func convertToPBScheduleCalendarSpecList(calendarSpec []ScheduleCalendarSpec) []*schedulepb.StructuredCalendarSpec {
 	calendarSpecPB := make([]*schedulepb.StructuredCalendarSpec, len(calendarSpec))
 	for i, e := range calendarSpec {
+		applyScheduleCalendarSpecDefault(&e)
```
Quinn-With-Two-Ns, line 751: currently I apply the default on schedule create and update, but [perhaps it should only be applied on update

cretz: What you have here makes the most sense to me. I think applying a default that the Godoc says it will apply to unset fields should be regardless of how the struct is used.

cretz (approved):

internal/schedule_client.go
```diff
@@ -54,25 +54,38 @@ type (
 	// that means all years match. For all fields besides year, at least one Range must be present to match anything.
```
dnr, line 54: this comment isn't accurate anymore?

internal/internal_schedule_client.go
```diff
@@ -719,9 +719,37 @@ func convertFromPBScheduleCalendarSpecList(calendarSpecPB []*schedulepb.Structur
 	return calendarSpec
 }
 
+func applyScheduleCalendarSpecDefault(calendarSpec *ScheduleCalendarSpec) {
+	if calendarSpec.Second == nil {
```
dnr, line 723: I guess it's okay that a non-nil but zero-length slice doesn't get defaulted here?

cretz: No, we should fix this IMO. Any time we're defaulting a slice we should use a len == 0 check.

Quinn-With-Two-Ns: That was an intentional choice. If someone set an empty array they explicitly set the value and we should not override it. This also aligns with out other SDKs

internal/schedule_client.go
Quinn-With-Two-Ns: It is correct since at least one range for each field must be present still. We just have a default now. If you have a better wording I welcome any suggestions.

internal/internal_schedule_client.go
cretz: Hrmm, this is not usually a good Go practice for field defaults. I've found expecting a user to differentiate between nil and empty because they have different behaviors is rough. We can keep it, but I wouldn't be surprised if you'll have to explain it away (e.g. "I load this from a config file and the unmarshaller sets this to empty array"). Other langs do not have a precedent of treating empty and unset the same like Go and Proto and some others do.

Quinn-With-Two-Ns: An empty slice and a nil slice are different things and it isn't unusual for APIs to treat them differently. `json.Marshal` treats them differently for example. The same confusion could also happen the other way if someone wants to specify and empty list (or a higher level API wrapping `ScheduleCalendarSpec` does) and we override it.

Issue #1014, opened 2023-01-28
workflow.MutableSideEffect not deterministic if called multiple times with the same ID in the same WFT


Quinn-With-Two-Ns:
This workflow
```
func MutableSideEffectWorkflow(ctx workflow.Context) error {
	i := 0
	encodedValue := workflow.MutableSideEffect(ctx, "id",
		func(ctx workflow.Context) interface{} { return i },
		func(a interface{}, b interface{}) bool {
			return a == b
		},
	)
	var sideEffectValue int
	err := encodedValue.Get(&sideEffectValue)
	if err != nil {
		return err
	}
	var future workflow.Future
	if sideEffectValue == 0 {
		ao := workflow.ActivityOptions{
			ScheduleToStartTimeout: time.Minute,
			StartToCloseTimeout:    time.Minute,
			HeartbeatTimeout:       time.Second * 20,
		}
		ctx = workflow.WithActivityOptions(ctx, ao)
		input := "hello"
		future = workflow.ExecuteActivity(ctx, helloworldActivity, input)
	}
	i += 1

	encodedValue = workflow.MutableSideEffect(ctx, "id",
		func(ctx workflow.Context) interface{} { return i },
		func(a interface{}, b interface{}) bool {
			return a == b
		},
	)
	err = encodedValue.Get(&sideEffectValue)
	if err != nil {
		return err
	}

	var output string
	err = future.Get(ctx, &output)
	if err != nil {
		return err
	}
	return nil
}
```
Is not deterministic because on the fist run `sideEffectValue` is equal to one 0 after the first `MutableSideEffect` and the activity starts. On replay `sideEffectValue` is equal to 1 and the activity is not launched. We don't document this [limitation](https://pkg.go.dev/go.temporal.io/sdk/workflow#MutableSideEffect) of mutable side effects and I believe it is a bug.

The problem is the Go SDK always uses the lastest value of a MutableSideEffect up to the current WFT for each ID so if `MutableSideEffect` returns a different value in the same WFT the last will be used on replay.

I found this while trying to fix workflow versioning in a loop, but @cretz also brought up this issue in his PR https://github.com/temporalio/sdk-go/pull/809

>Question: What if two mutable side effects of the same ID are run during the same WFT? We don't differentiate by command, only by user-defined marker ID.

  - Version: 1.20.0
  - Platform: Mac M1

Quinn-With-Two-Ns:
[...]

PR #1051, opened 2023-02-22
Track mutable side effect calls

Quinn-With-Two-Ns:
Track how many times a mutable side effect is called to distinguish multiple `MutableSIdeEffect` calls in the same WFT.

I'll note this PR wont help existing histories that executed multiple `MutableSIdeEffect` in the same WFT. If we also use the `CommandID` to help distinguish multiple calls. The could help certain histories replay correctly. The trade off is an even more complex implementation. There will always be some histories that we cannot save because the SDK did not write enough information to history

Resolves: https://github.com/temporalio/sdk-go/issues/1014

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -30,6 +30,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sort"
 	"sync"
 	"time"
 
@@ -109,7 +110,7 @@ type (
 		changeVersions             map[string]Version
 		pendingLaTasks             map[string]*localActivityTask
 		completedLaAttemptsThisWFT uint32
-		mutableSideEffect          map[string]*commonpb.Payloads
+		mutableSideEffect          map[string]map[int]*commonpb.Payloads
 		unstartedLaTasks           map[string]struct{}
 		openSessions               map[string]*SessionInfo
 
@@ -117,6 +118,10 @@ type (
 		// during replay to determine whether a command should be created. The keys
 		// are the user-provided IDs + "_" + the command counter.
 		mutableSideEffectsRecorded map[string]bool
+		// Records the number of times a mutable side effect was called per ID over the
+		// life of the workflow. Used to help distinguish multiple calls to MutableSideEffect in the same
+		// WorkflowTask.
+		mutableSideEffectCallCounter map[string]int
 
 		// LocalActivities have a separate, individual counter instead of relying on actual commandEventIDs.
 		// This is because command IDs are only incremented on activity completion, which breaks
@@ -196,22 +201,23 @@ func newWorkflowExecutionEventHandler(
 	deadlockDetectionTimeout time.Duration,
 ) workflowExecutionEventHandler {
 	context := &workflowEnvironmentImpl{
-		workflowInfo:             workflowInfo,
-		commandsHelper:           newCommandsHelper(),
-		sideEffectResult:         make(map[int64]*commonpb.Payloads),
-		mutableSideEffect:        make(map[string]*commonpb.Payloads),
-		changeVersions:           make(map[string]Version),
-		pendingLaTasks:           make(map[string]*localActivityTask),
-		unstartedLaTasks:         make(map[string]struct{}),
-		openSessions:             make(map[string]*SessionInfo),
-		completeHandler:          completeHandler,
-		enableLoggingInReplay:    enableLoggingInReplay,
-		registry:                 registry,
-		dataConverter:            dataConverter,
-		failureConverter:         failureConverter,
-		contextPropagators:       contextPropagators,
-		deadlockDetectionTimeout: deadlockDetectionTimeout,
-		protocols:                protocol.NewRegistry(),
+		workflowInfo:                 workflowInfo,
+		commandsHelper:               newCommandsHelper(),
+		sideEffectResult:             make(map[int64]*commonpb.Payloads),
+		mutableSideEffect:            make(map[string]map[int]*commonpb.Payloads),
+		changeVersions:               make(map[string]Version),
+		pendingLaTasks:               make(map[string]*localActivityTask),
+		unstartedLaTasks:             make(map[string]struct{}),
+		openSessions:                 make(map[string]*SessionInfo),
+		completeHandler:              completeHandler,
+		enableLoggingInReplay:        enableLoggingInReplay,
+		registry:                     registry,
+		dataConverter:                dataConverter,
+		failureConverter:             failureConverter,
+		contextPropagators:           contextPropagators,
+		deadlockDetectionTimeout:     deadlockDetectionTimeout,
+		protocols:                    protocol.NewRegistry(),
+		mutableSideEffectCallCounter: make(map[string]int),
 	}
 	context.logger = ilog.NewReplayLogger(
 		log.With(logger,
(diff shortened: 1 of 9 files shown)


Quinn-With-Two-Ns:
Marking as a draft because I manly want feedback on the counting approach

Quinn-With-Two-Ns:
Discussed offline and agreed to leave old broken workflow behavior as is

cretz:
:+1: Will wait for ready-to-review before reviewing in earnest

worker/worker.go
```diff
@@ -199,6 +199,12 @@ type (
 		// The logger is the only optional parameter. Defaults to the noop logger. The Run ID and Workflow ID used during replay are derived
 		// from execution.
 		ReplayWorkflowExecution(ctx context.Context, service workflowservice.WorkflowServiceClient, logger log.Logger, namespace string, execution workflow.Execution) error
+
+		// GetWorkflowResult gets the result from a succesfully replayed workflow that finished with WorkflowExecutionCompleted.
+		// WorkflowID is optional and defaults to ReplayId.
+		//
+		// Note: For all replay functions if no WorkflowID is given for replay the default is ReplayId.
+		GetWorkflowResult(workflowID string, valuePtr interface{}) error
```
Quinn-With-Two-Ns, line 207: We should have added something like this when we stopped checking the workflow execution result and we need it to test we didn't break old behaviour.

internal/internal_event_handlers.go
```diff
@@ -1311,7 +1356,20 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
 				err = weh.dataConverter.FromPayloads(sideEffectDataPayload, &sideEffectDataID, &sideEffectDataContents)
 			}
 			if err == nil {
-				weh.mutableSideEffect[sideEffectDataID] = &sideEffectDataContents
+				counterHintPayload, ok := attributes.GetDetails()[mutableSideEffectCallCounterName]
+				var counterHint int
+				if ok {
+					err = weh.dataConverter.FromPayloads(counterHintPayload, &counterHint)
+				} else {
+					// An old version of the SDK did not write the counter hint so we have to assume.
```
Quinn-With-Two-Ns, line 1364: Maybe if we detect this scenario we should not write `counterHints` to maintain behaviour in a given workflow?

internal/internal_event_handlers.go
```diff
@@ -782,15 +788,51 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() (*commonpb.Payloads, erro
 	wc.logger.Debug("SideEffect Marker added", tagSideEffectID, sideEffectID)
 }
 
+func (wc *workflowEnvironmentImpl) lookupMutableSideEffect(id string) (*commonpb.Payloads, bool) {
```
cretz, line 791: May need a comment. I am confused what this is doing at first glance. It's a mutable lookup?

cretz, line 791: If I understand your algorithm correctly, this may be simpler (untested):

```go
func (wc *workflowEnvironmentImpl) lookupMutableSideEffect(id string) *commonpb.Payloads {
	// Fail if ID not found
	callCountPayloads := wc.mutableSideEffect[id]
	if len(callCountPayloads) == 0 {
		return nil
	}

	// Find the most recent call at/before the current call count
	var payloads *commonpb.Payloads
	payloadIndex := -1
	for callCount, maybePayloads := range callCountPayloads {
		if callCount <= currentCallCount && callCount > payloadIndex {
			payloads = maybePayloads
			payloadIndex = callCount
		}
	}
	return payloads
}
```

* No need for multiple iterations and sorting and all of that
* No need for a bool when nullness can represent it just fine
* I don't understand why you "garbage collect" old entries. Maybe I don't understand this map. Is call count per task or something else I am missing? If it's to help Go mem, no need (and it's actually not helping Go mem anyways tbh), and if it's to save the amount of iterations next time through, also no need.

internal/internal_event_handlers.go
```diff
@@ -1311,7 +1356,20 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
 				err = weh.dataConverter.FromPayloads(sideEffectDataPayload, &sideEffectDataID, &sideEffectDataContents)
 			}
 			if err == nil {
-				weh.mutableSideEffect[sideEffectDataID] = &sideEffectDataContents
+				counterHintPayload, ok := attributes.GetDetails()[mutableSideEffectCallCounterName]
+				var counterHint int
+				if ok {
+					err = weh.dataConverter.FromPayloads(counterHintPayload, &counterHint)
```
cretz, line 1362: Would suggest eager return here on error (can use different `err` var name for shadowing concerned). Confusing to read code below being executed on error. At the very least add a comment saying you know logic after conditional will run in errored state.

internal/internal_event_handlers.go
cretz: I struggle to follow this code well enough to know, but so long as old behavior remains, any approach is fine by me

worker/worker.go
cretz, line 207: I am not quite sure we are ready for exposing workflow results from our replayers. We have intentionally avoided doing this in other replayers. I think we need more discussion around the API. For example, does this show errors from a workflow? How do I know if it didn't complete?

I do understand the desire for this and how it helps the replay tests. Options:

1. Get together as a team and decide that we want workflow results provided as a result of replay (includes success and fail instead of just wft fail like today)
2. Remove this from the interface and you can leave this on the internal struct for now and just cast to that internal struct in these tests
3. Assert in your workflow and panic on failure

internal/internal_event_handlers.go
```diff
@@ -782,15 +788,51 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() (*commonpb.Payloads, erro
...
+		for i := 0; i < foundIndex; i++ {
+			delete(payloadAtCallCount, calls[i])
+		}
+
+		return payload, true
+	}
+	return nil, false
+}
+
 func (wc *workflowEnvironmentImpl) MutableSideEffect(id string, f func() interface{}, equals func(a, b interface{}) bool) converter.EncodedValue {
```
cretz, line 824: Do we need to add some logic to `isCommandMatchEvent` to make sure the counters match? Or is that too strict of determinism checking.

test/replaytests/workflows.go
```diff
@@ -217,3 +217,39 @@ func SideEffectWorkflow(ctx workflow.Context, field string) error {
 func EmptyWorkflow(ctx workflow.Context, _ string) error {
 	return nil
 }
+
+func MutableSideEffectWorkflow(ctx workflow.Context) ([]int, error) {
```
cretz, line 221: Is it possible, maybe in addition to these replay tests, to replicate the error that caused this issue in the first place? Maybe this workflow will, haven't tested, but want to make sure we can verify new behavior is for new workflow run against server.

Also I wonder if we want to have an integration test proving non-deterministic mutable side effect order change. That may be asking too much though.

internal/internal_event_handlers.go
Quinn-With-Two-Ns: `MutableSideEffect` skip command match

worker/worker.go
Quinn-With-Two-Ns: > ready for exposing workflow results from our replayers.
Not sure what conditions would make us ready?

> show errors from a workflow
We already expose this on replay with the current replay API

> How do I know if it didn't complete?
What didn't complete the original workflow history or the replayed workflow? If the original workflow didn't complete you know that (you have the history) and the replayed workflow can't complete. If the original did, but the replayed didn't we fail.

test/replaytests/workflows.go
Quinn-With-Two-Ns: > Or is that what that small assertion addition in workflow_test.go already does?

Yes exactly

>Also I wonder if we want to have an integration test proving non-deterministic mutable side effect order change. That may be asking too much though.

not sure I understand, do you mean a test that shows changing the order of mutable side effects is not deterministic?

internal/internal_event_handlers.go
Quinn-With-Two-Ns: I'll test I think your code works and is simpler, thanks!

> Maybe I don't understand this map. Is call count per task or something else I am missing

It's call count per workflow execution. I have that documented here

>// Records the number of times a mutable side effect was called per ID over the
// life of the workflow. Used to help distinguish multiple calls to MutableSideEffect in the same
// WorkflowTask.
mutableSideEffectCallCounter map[string]int

But tbf it should be documented at `callCountPayloads` as well.

the other question is should it be per workflow execution or workflow task?

cretz: Per entire workflow execution makes sense. I was trying to understand why you were "GC'ing" map elements instead of just leaving them there. But if it was just an attempt to help Go out, no worries (but don't need to do it).

Quinn-With-Two-Ns: >But if it was just an attempt to help Go out, no worries (but don't need to do it)

Why do I not need to? If I don't the map will grow indefinitely and the memory will never be freed?

worker/worker.go
cretz: > Not sure what conditions would make us ready?

Just confirming group wide that we want to across SDKs and if there are any caveats here

> If the original workflow didn't complete you know that (you have the history)

By this logic you also have the history for workflow completion so we don't need to return the result. If we're gonna return result of a fully-complete workflow from the replayer, the user should be able to know the difference between workflow incomplete vs workflow complete successfully vs workflow complete unsuccessfully vs workflow task failure.

> Currently if a WFT fails on replay we expose the error. If the original workflow failed I think we just exit and don't check it matched with the replay which seems like a bug?

It's not a bug in today's approach because our replayer does not expose anything about workflow complete (success or failure). But since you are adding workflow complete information, we need to provide full workflow complete information, not only success.

internal/internal_event_handlers.go
cretz: Ah, I see that now, thanks!

worker/worker.go
Quinn-With-Two-Ns: >By this logic you also have the history for workflow completion so we don't need to return the result. If we're gonna return result of a fully-complete workflow from the replayer, the user should be able to know the difference between workflow incomplete vs workflow complete successfully vs workflow complete unsuccessfully vs workflow task failure.

Your missing the point of what I am saying. I am saying for most part they can. Except for workflow execution failed.

>It's not a bug in today's approach because our replayer does not expose anything about workflow complete (success or failure). But since you are adding workflow complete information, we need to provide full workflow complete information, not only success.

It isn't about exposing it is about. I am saying we don't check it at all. That isn't consistent with how we handle Workflow success or continue as new.

test/replaytests/workflows.go
cretz: > do you mean a test that shows changing the order of mutable side effects is not deterministic?

I do. Basically if I add a mutable side effect for the ID if it's non-deterministic. But I think this is a bit much, no need.

internal/internal_event_handlers.go
cretz: > Why do I not need to?

Because the workflow execution struct is automatically collected with all fields when it leaves cache. I highly doubt storing every mutable side effect ever created for a specific workflow is harmful.

Also, many don't know this but deleting map elements does _not_ reclaim space in Go yet. And it's not like you're removing the nested map from its parent when it's empty anyways.

> If I don't the map will grow indefinitely and the memory will never be freed?

It won't grow indefinitely because it'll be bounded by the number of mutable side effects the user runs right? That also affects command slice size and other things harmlessly IMO.

worker/worker.go
cretz: Ok, maybe I am misunderstanding. To ask it simply: Why are we exposing `GetWorkflowResult` in workflow replayer today in Go SDK and not in any other SDK? Also: Since we're now returning workflow result, how can I as a user differentiate replayer failure from workflow failure?

internal/internal_event_handlers.go
Quinn-With-Two-Ns: >Granted removing the reference to the payload pointer does allow it to be collected.

Yeah this is what I want to be collected because it could be larger

worker/worker.go
Quinn-With-Two-Ns: Discussed offline, will drop from the interface for know and revisit exposing more completely later

PR #1053, opened 2023-03-01
Fix getFunctionName functions with generic types

robpryan:
For a function with generic types `FuncForPC` will return [...] after the function name. So the result from `getFunctionName` would end up being `]` after the split. Fix this by removing `[...]` from the full name.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1578,9 +1578,11 @@ func getFunctionName(i interface{}) (name string, isMethod bool) {
 	if fullName, ok := i.(string); ok {
 		return fullName, false
 	}
-	fullName := runtime.FuncForPC(reflect.ValueOf(i).Pointer()).Name()
 	// Full function name that has a struct pointer receiver has the following format
 	// <prefix>.(*<type>).<function>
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
Hey @robpryan , Thanks for the contribution. 

One interesting this is that `FuncForPC` has no type information so could cause unintentional collision if people register use generic workflows or activities.

https://go.dev/play/p/mGOySR4RFZc

Also could you agree to the CLA?

cretz:
I believe we should disallow generic function registration for the following three reasons:

1. Go 1.19 doesn't support `runtime.FuncForPC`'s `Name()` well like Go 1.20 (just does `func1`, `func2`, etc) so you'd have to wait until 1.19 EOL at least (and we often don't disallow use of our SDK on a version even when Go EOLs the version)
2. One of the primary benefits of typed functions is for caller side use and type parameters make it hard for the caller to know what was registered
3. It is really easy for someone to write a fixed-type one-liner to delegate to their generic if they really need to

Note, it may be hard to disallow generic functions in Go 1.19 because it is not clear in reflection and not clear from the name. That's fine, but we need to document that they are disallowed.

robpryan:
Yeah that makes sense, would be good to disallow them to make it clear.

Issue #1044, opened 2023-02-16
SDK updates not being properly run


cretz:
## Steps to Reproduce the Problem

Here are two standalone examples that cause context deadline results for update.

Gets some proper events:
<details><summary>Replication 1</summary><p>

```go
package main

import (
	"context"
	"fmt"
	"log"
	"time"

	"github.com/google/uuid"
	"go.temporal.io/sdk/client"
	"go.temporal.io/sdk/worker"
	"go.temporal.io/sdk/workflow"
)

// Scenario:
// * Workflow waits for activity inside update
// * Workflow worker stops after activity started
// * Update sent
// * Activity completed
// * Workflow worker started back up

func main() {
	if err := run(); err != nil {
		log.Fatal(err)
	}
}

func run() error {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	log.Printf("Creating client")
	c, err := client.Dial(client.Options{})
	if err != nil {
		return fmt.Errorf("failed creating client: %w", err)
	}
	defer c.Close()

	log.Printf("Starting workflow worker")
	taskQueue := uuid.NewString()
	workflowWorker := worker.New(c, taskQueue, worker.Options{LocalActivityWorkerOnly: true})
	workflowWorker.RegisterWorkflow(MyWorkflow)
	if err := workflowWorker.Start(); err != nil {
		return fmt.Errorf("failed starting worker: %w", err)
	}

	log.Printf("Starting activity worker")
	activityWorker := worker.New(c, taskQueue, worker.Options{DisableWorkflowWorker: true})
	activityWorker.RegisterActivity(MyActivity)
	if err := activityWorker.Start(); err != nil {
		return fmt.Errorf("failed starting worker: %w", err)
	}
	defer activityWorker.Stop()

	log.Printf("Starting workflow")
	run, err := c.ExecuteWorkflow(ctx, client.StartWorkflowOptions{TaskQueue: taskQueue}, MyWorkflow)
	if err != nil {
		return fmt.Errorf("failed starting workflow: %w", err)
	}

	log.Printf("Waiting for activity to show as started")
	<-activityWaiting

	log.Printf("Starting update")
	// Send update in background
	updateResultCh := make(chan string, 1)
	updateErrCh := make(chan error, 1)
	go func() {
		var updateResult string
		handle, err := c.UpdateWorkflow(ctx, run.GetID(), run.GetRunID(), "MyUpdate")
		if err != nil {
			updateErrCh <- err
[...]

PR #1055, opened 2023-03-06
Lookahead to infer messages on all replay paths

mmcshane:
## What was changed

Include lookahead message inference on this path to workflow state rebuilding

## Why?

Without inference, updates are not replayed

## Checklist

1. Closes 

#1044 

2. How was this tested:

New unit test here.

3. Any docs updates needed?

No

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -665,6 +665,7 @@ func (w *workflowExecutionContextImpl) resetStateIfDestroyed(task *workflowservi
 				return err
 			}
 		}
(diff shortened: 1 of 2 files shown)


bergundy (approved):
Is there some regression test we can that tests this E2E?

mmcshane:
> Is there some regression test we can that tests this E2E?

Working on adding it to `features` but that work has to include

- dynamic skipping of tests based on sniffing namespace capabilities
- stop/start of worker

So it has been slower going.

Issue #1052, opened 2023-02-27
Fix getVersion search attribute bloat


Quinn-With-Two-Ns:
Java and Go accumulate values in the `TemporalChangeVersion` search attribute every time a `getVersion` call is made.

The search attribute value size is capped, which causes user workflows to get stuck in a WFT failure loop without an easy way of out of this situation. If we hit the size limit we should stop adding new versions to `TemporalChangeVersion` and log a clear error/warning when situation reached.

cretz:
> If we hit the size limit we should stop adding new versions to TemporalChangeVersion

Or remove the oldest from the list. But either way is fine, we just need to log a clear error/warning when situation reached.

Quinn-With-Two-Ns:
Yep good point, I noted the requirement to log as well

PR #1056, opened 2023-03-08
Reduce SA bloat on GetVersion

Quinn-With-Two-Ns:
Reduce SA bloat on GetVersion. This PR implements 3 new features to the Go SDK.

1. Adds capability visibility to workflow and activity task execution.
2. Adds sdk flags based off the core API to allow internal SDK versioning
3. Uses sdk flags to no longer upsert search attributes on `GetVersion` if that search attribute would exceed the default size limit

### DIFF
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -930,7 +930,7 @@ func (h *commandsHelper) incrementNextCommandEventIDIfVersionMarker() {
 		// is UpsertSearchableAttributes to keep track of executions using particular version of code.
 		delete(h.versionMarkerLookup, h.nextCommandEventID)
 		h.incrementNextCommandEventID()
-		// UpsertSearchableAttributes may not have been written if the search attriute was too large.
+		// UpsertSearchableAttributes may not have been written if the search attribute was too large.
 		if marker.searchAttrUpdated {
 			h.incrementNextCommandEventID()
 		}
@@ -1093,18 +1093,20 @@ func (h *commandsHelper) recordVersionMarker(changeID string, version Version, d
 		panic(err)
 	}
 
-	searchAttributeWasUpdatedPayload, err := dc.ToPayloads(searchAttributeWasUpdated)
-	if err != nil {
-		panic(err)
-	}
-
 	recordMarker := &commandpb.RecordMarkerCommandAttributes{
 		MarkerName: versionMarkerName,
 		Details: map[string]*commonpb.Payloads{
-			versionMarkerChangeIDName:         changeIDPayload,
-			versionMarkerDataName:             versionPayload,
-			versionSearchAttributeUpdatedName: searchAttributeWasUpdatedPayload,
+			versionMarkerChangeIDName: changeIDPayload,
+			versionMarkerDataName:     versionPayload,
 		},
+	}
+
+	if !searchAttributeWasUpdated {
+		searchAttributeWasUpdatedPayload, err := dc.ToPayloads(searchAttributeWasUpdated)
+		if err != nil {
+			panic(err)
+		}
+		recordMarker.Details[versionSearchAttributeUpdatedName] = searchAttributeWasUpdatedPayload
 	}
 
 	command := h.newMarkerCommandStateMachine(markerID, recordMarker)
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -140,6 +140,11 @@ type (
 		*naiveCommandStateMachine
 	}
 
+	versionMarker struct {
+		changeID          string
+		searchAttrUpdated bool
+	}
+
 	commandsHelper struct {
 		nextCommandEventID int64
 		orderedCommands    *list.List
@@ -148,7 +153,7 @@ type (
 		scheduledEventIDToActivityID          map[int64]string
 		scheduledEventIDToCancellationID      map[int64]string
 		scheduledEventIDToSignalID            map[int64]string
-		versionMarkerLookup                   map[int64]string
+		versionMarkerLookup                   map[int64]versionMarker
 		commandsCancelledDuringWFCancellation int64
 		workflowExecutionIsCancelling         bool
 
@@ -221,13 +226,14 @@ const (
 	localActivityMarkerName     = "LocalActivity"
 	mutableSideEffectMarkerName = "MutableSideEffect"
 
-	sideEffectMarkerIDName           = "side-effect-id"
-	sideEffectMarkerDataName         = "data"
-	versionMarkerChangeIDName        = "change-id"
-	versionMarkerDataName            = "version"
-	localActivityMarkerDataName      = "data"
-	localActivityResultName          = "result"
-	mutableSideEffectCallCounterName = "mutable-side-effect-call-counter"
+	sideEffectMarkerIDName            = "side-effect-id"
+	sideEffectMarkerDataName          = "data"
+	versionMarkerChangeIDName         = "change-id"
+	versionMarkerDataName             = "version"
+	versionSearchAttributeUpdatedName = "version-search-attribute-updated"
+	localActivityMarkerDataName       = "data"
+	localActivityResultName           = "result"
+	mutableSideEffectCallCounterName  = "mutable-side-effect-call-counter"
 )
 
 func (d commandState) String() string {
@@ -884,7 +890,7 @@ func newCommandsHelper() *commandsHelper {
 		scheduledEventIDToActivityID:          make(map[int64]string),
 		scheduledEventIDToCancellationID:      make(map[int64]string),
 		scheduledEventIDToSignalID:            make(map[int64]string),
-		versionMarkerLookup:                   make(map[int64]string),
+		versionMarkerLookup:                   make(map[int64]versionMarker),
 		commandsCancelledDuringWFCancellation: 0,
 	}
 }
@@ -917,15 +923,18 @@ func (h *commandsHelper) getNextID() int64 {
 }
 
 func (h *commandsHelper) incrementNextCommandEventIDIfVersionMarker() {
-	_, ok := h.versionMarkerLookup[h.nextCommandEventID]
+	marker, ok := h.versionMarkerLookup[h.nextCommandEventID]
 	for ok {
 		// Remove the marker from the lookup map and increment nextCommandEventID by 2 because call to GetVersion
-		// results in 2 events in the history.  One is GetVersion marker event for changeID and change version, other
+		// results in 1 or 2 events in the history.  One is GetVersion marker event for changeID and change version, other
 		// is UpsertSearchableAttributes to keep track of executions using particular version of code.
 		delete(h.versionMarkerLookup, h.nextCommandEventID)
 		h.incrementNextCommandEventID()
-		h.incrementNextCommandEventID()
-		_, ok = h.versionMarkerLookup[h.nextCommandEventID]
+		// UpsertSearchableAttributes may not have been written if the search attriute was too large.
+		if marker.searchAttrUpdated {
+			h.incrementNextCommandEventID()
+		}
+		marker, ok = h.versionMarkerLookup[h.nextCommandEventID]
 	}
 }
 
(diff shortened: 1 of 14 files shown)


Quinn-With-Two-Ns:
resolves: https://github.com/temporalio/sdk-go/issues/1052

internal/internal_event_handlers.go
```diff
@@ -728,8 +733,24 @@ func (wc *workflowEnvironmentImpl) GetVersion(changeID string, minSupported, max
...
 		version = maxSupported
-		wc.commandsHelper.recordVersionMarker(changeID, version, wc.GetDataConverter())
-		_ = wc.UpsertSearchAttributes(createSearchAttributesForChangeVersion(changeID, version, wc.changeVersions))
+		changeVersionSA := createSearchAttributesForChangeVersion(changeID, version, wc.changeVersions)
+		attr, err := validateAndSerializeSearchAttributes(changeVersionSA)
+		if err == nil {
+			// Server has a limit for the max size of a single search attribute value. If we exceed the default limit
+			// do not try to upsert as it will cause the workflow to fail.
+			updateSearchAttribute := true
+			if wc.sdkFlags.tryUse(LimitChangeVersionSASize, !wc.isReplay) && len(attr.IndexedFields[TemporalChangeVersion].GetData()) >= changeVersionSearchAttrSizeLimit {
```
Quinn-With-Two-Ns, line 742: I'll note this will change the behavior  for new WFT not new workflows

internal/internal_command_state_machine.go
```diff
@@ -1084,11 +1093,17 @@ func (h *commandsHelper) recordVersionMarker(changeID string, version Version, d
...
+	}
+
 	recordMarker := &commandpb.RecordMarkerCommandAttributes{
 		MarkerName: versionMarkerName,
 		Details: map[string]*commonpb.Payloads{
-			versionMarkerChangeIDName: changeIDPayload,
-			versionMarkerDataName:     versionPayload,
+			versionMarkerChangeIDName:         changeIDPayload,
+			versionMarkerDataName:             versionPayload,
+			versionSearchAttributeUpdatedName: searchAttributeWasUpdatedPayload,
```
Quinn-With-Two-Ns, line 1106: I need to adjust this to not send if it is not false. 

This flag isn't strictly needed but because the Go SDK counts command IDs and processes history out of order it is useful to have to avoid refactoring how the SDK replays commands  just for this edge case.

internal/internal_command_state_machine.go
```diff
@@ -917,15 +923,18 @@ func (h *commandsHelper) getNextID() int64 {
...
 	for ok {
 		// Remove the marker from the lookup map and increment nextCommandEventID by 2 because call to GetVersion
-		// results in 2 events in the history.  One is GetVersion marker event for changeID and change version, other
+		// results in 1 or 2 events in the history.  One is GetVersion marker event for changeID and change version, other
 		// is UpsertSearchableAttributes to keep track of executions using particular version of code.
 		delete(h.versionMarkerLookup, h.nextCommandEventID)
 		h.incrementNextCommandEventID()
-		h.incrementNextCommandEventID()
-		_, ok = h.versionMarkerLookup[h.nextCommandEventID]
+		// UpsertSearchableAttributes may not have been written if the search attriute was too large.
```
cretz, line 933:
```suggestion
		// UpsertSearchableAttributes may not have been written if the search attribute was too large.
```

internal/internal_event_handlers.go
```diff
@@ -728,8 +733,24 @@ func (wc *workflowEnvironmentImpl) GetVersion(changeID string, minSupported, max
 		// GetVersion for changeID is called first time (non-replay mode), generate a marker command for it.
 		// Also upsert search attributes to enable ability to search by changeVersion.
 		version = maxSupported
-		wc.commandsHelper.recordVersionMarker(changeID, version, wc.GetDataConverter())
-		_ = wc.UpsertSearchAttributes(createSearchAttributesForChangeVersion(changeID, version, wc.changeVersions))
+		changeVersionSA := createSearchAttributesForChangeVersion(changeID, version, wc.changeVersions)
+		attr, err := validateAndSerializeSearchAttributes(changeVersionSA)
```
cretz, line 737: I know we never cared about serialization failures before before (and it should never happen), but IMO we should either log it or truly ignore it and don't let it affect code path

cretz, line 737: Does this, or has this ever, altered the value in the user-visible SA map?

internal/internal_flags.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"go.temporal.io/api/workflowservice/v1"
+)
+
+// sdkFlag represents a flag used to help version the sdk internally to make breaking changes
+// in workflow logic.
+type sdkFlag uint32
+
+const (
+	Unset sdkFlag = iota
```
cretz, line 36: Since this is a big shared package, would like to prefix or suffix flag names, e.g. `SDKFlagUnset`

internal/internal_flags.go
```diff
@@ -0,0 +1,106 @@
...
+type sdkFlags struct {
+	capabilities *workflowservice.GetSystemInfoResponse_Capabilities
+	// Flags that have been recieved from the server
+	currentFlags map[sdkFlag]bool
+	// Flags that have been set this WFT that have not been sent to the server.
+	// Keep track of them sepratly so we know what to send to the server.
+	newFlags map[sdkFlag]bool
+}
+
+func newSdkFlags(capabilities *workflowservice.GetSystemInfoResponse_Capabilities) *sdkFlags {
```
cretz, line 68:
```suggestion
func newSDKFlags(capabilities *workflowservice.GetSystemInfoResponse_Capabilities) *sdkFlags {
```

(and other places where applicable)

internal/internal_task_handlers.go
```diff
@@ -846,11 +860,12 @@ func (w *workflowExecutionContextImpl) ProcessWorkflowTask(workflowTask *workflo
 	metricsTimer := metricsHandler.Timer(metrics.WorkflowTaskReplayLatency)
 
 	eventHandler.ResetLAWFTAttemptCounts()
+	eventHandler.sdkFlags.resetSdkFlags()
```
cretz, line 863: I am not sure you want to reset SDK flags every task. I think the idea behind many flags (including this one, though maybe not all in the future) is that they only need to be set on any workflow task to apply to all future workflow tasks.

internal/internal_event_handlers.go
cretz, line 742: Can you add an integration test that intentionally does enough versions to go over the SA size and then confirm the ones past the size are not serialized?

internal/internal_flags.go
Quinn-With-Two-Ns: I didn't name the struct `SDKFlag` because I did not want it to be exported outside internal, unless your suggesting just renaming this function?

internal/internal_task_handlers.go
Quinn-With-Two-Ns: >is that they only need to be set on any workflow task to apply to all future workflow tasks.
Yes  I agree, I'll rename this function since it's confusing

internal/internal_event_handlers.go
Quinn-With-Two-Ns: Yes I already do that is the `version-loop-workflow-236.json` history.

internal/internal_event_handlers.go
Quinn-With-Two-Ns: >  don't let it affect code path

not sure what you mean by this. If we can't serialize the SA we can't check the size or upsert so it will effect the code path. I am fine logging, panicing or skipping the upsert but still updating the marker personally.

internal/internal_flags.go
```diff
@@ -0,0 +1,106 @@
...
+
+// tryUse returns true if this flag may currently be used. If record is true, always returns
+// true and records the flag as being used.
+func (sf *sdkFlags) tryUse(flag sdkFlag, record bool) bool {
+	if !sf.capabilities.GetSdkMetadata() {
+		return false
+	}
+
+	if record {
+		sf.newFlags[flag] = true
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: oh yes that isn't intentional, i'll fix

internal/internal_event_handlers.go
cretz: Logging works for me. I just meant ignore because you know it will never fail, but I too prefer logging.

internal/internal_flags.go
cretz: I am suggesting just this function and any Go function where an abbreviation needs to be capitalized (i.e. non-exported but not the first word).

internal/internal_event_handlers.go
cretz: Test performance be damned, would you be willing to have an _integration_ test run a new workflow against a server triggering this behavior? It's mostly just to codify our expectations (e.g. server drops the max for SAs, though I don't think they ever will, this just makes us feel better).

Quinn-With-Two-Ns: Yep can do

cretz (approved):
LGTM, but would like either @Sushisource or @bergundy to review the SDK flag behavior just in case.

internal/internal_flags.go
```diff
@@ -0,0 +1,107 @@
...
+
+// sdkFlag represents a flag used to help version the sdk internally to make breaking changes
+// in workflow logic.
+type sdkFlag uint32
+
+const (
+	SDKFlagUnset sdkFlag = iota
+	// LimitChangeVersionSASize will limit the search attribute size of TemporalChangeVersion to 2048 when
+	// calling GetVersion. If the limit is exceeded the search attribute is not updated.
+	SDKFlagLimitChangeVersionSASize
```
Sushisource, line 39: I'd probably just use explicit numbering here. Reason being if flags are ever removed, you want to "reserve" those numbers a-la protobuf.

internal/internal_task_handlers.go
```diff
@@ -230,12 +232,21 @@ func (eh *history) IsNextWorkflowTaskFailed() (isFailed bool, binaryChecksum str
...
 		isFailed := nextEventType == enumspb.EVENT_TYPE_WORKFLOW_TASK_TIMED_OUT || nextEventType == enumspb.EVENT_TYPE_WORKFLOW_TASK_FAILED
 		var binaryChecksum string
+		var flags []sdkFlag
 		if nextEventType == enumspb.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
 			binaryChecksum = nextEvent.GetWorkflowTaskCompletedEventAttributes().BinaryChecksum
+			for _, flag := range nextEvent.GetWorkflowTaskCompletedEventAttributes().GetSdkMetadata().GetLangUsedFlags() {
+				f := sdkFlagFromUint(flag)
+				if !f.isValid() {
+					// If a flag is not recognized (value is too high or not defined), it must fail the workflow task
+					return false, "", nil, errors.New("could not recognize SDK flag")
```
Sushisource, line 242: Hmm, I'm actually not doing this currently and probably should.

internal/internal_task_handlers.go
```diff
@@ -274,20 +285,20 @@ func isCommandEvent(eventType enumspb.EventType) bool {
 
 // NextCommandEvents returns events that there processed as new by the next command.
 // TODO(maxim): Refactor to return a struct instead of multiple parameters
-func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, binaryChecksum string, err error) {
+func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, binaryChecksum string, sdkFlags []sdkFlag, err error) {
```
Sushisource, line 288: Might be a good time to do this TODO

Sushisource (approved):
Looking good

internal/internal_flags.go
Quinn-With-Two-Ns: yeah good point

internal/internal_task_handlers.go
Quinn-With-Two-Ns: oh that reminds me I wanted to add a test for this case, I'll do that quickly

internal/internal_task_handlers.go
Quinn-With-Two-Ns: I did try this and didn't find it to much cleaner because this code uses named return values heavily

Issue #1041, opened 2023-02-16
Workflow got hanged if spawn childWorkflow with same workflowId


raymondsze:
## Expected Behavior
I expect the parent workflow can be run successfully and only one child workflow is executed (because same workflow id). 

## Actual Behavior
It hanged in the second child workflow `childWorkflowFuture.GetChildWorkflowExecution().Get(ctx, &childWE)`. 

## Steps to Reproduce the Problem
I followed the documentation in https://legacy-documentation-sdks.temporal.io/go/spawn-a-child-workflow-execution.
If I spawn two childworkflow, both with the same workflowId. The second one will be hanged when calling `childWorkflowFuture.GetChildWorkflowExecution().Get(ctx, &childWE)`. What I want to do is, if the child workflow is already registered and running, then don't do anything. If the child workflow is completed or not yet registered, then register and run the child workflow without waiting. 

## Specifications

  - Version:
  - Platform:

raymondsze:
Sure.
```go

func ConsumerWorkflow(ctx workflow.Context) error {
	logger := workflow.GetLogger(ctx)
	for i := 0; i < 2; i++ {
		childWorkflowOptions := workflow.ChildWorkflowOptions{
			ParentClosePolicy:     enums.PARENT_CLOSE_POLICY_ABANDON,
			WorkflowID:            fmt.Sprintf("consumer-%v-%v", workflow.GetInfo(ctx).WorkflowExecution.ID, 0),
			WorkflowIDReusePolicy: enums.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE,
		}
		ctx = workflow.WithChildOptions(ctx, childWorkflowOptions)
		logger.Info("BEFORE EXECUTE!")
		childWorkflow := workflow.ExecuteChildWorkflow(ctx, ChildWorkflow, i)
		logger.Info("AFTER EXECUTE!")
		logger.Info("BEFORE GET EXECUTION!")
		_ = childWorkflow.GetChildWorkflowExecution().Get(ctx, nil)
		logger.Info("AFTER GET EXECUTION!")
	}
	return nil
}
```

Here is the log

Worker
```bash
2023/02/17 11:22:28 INFO  No logger configured for temporal client. Created default one.
2023/02/17 11:22:28 INFO  Started Worker Namespace default TaskQueue hello-world WorkerID 4822@Raymond@
[...]

PR #1058, opened 2023-03-11
Pass errors to the child workflow execution future

Quinn-With-Two-Ns:
Pass errors to the child workflow execution future. I had to version this because the change is technically non deterministic, I don't know if anyone would hit is, but I also don't understand why people launch duplicate child workflows so I am playing it safe.

resolves: https://github.com/temporalio/sdk-go/issues/1041

### DIFF
--- a/internal/internal_flags.go
+++ b/internal/internal_flags.go
@@ -37,7 +37,10 @@ const (
 	// LimitChangeVersionSASize will limit the search attribute size of TemporalChangeVersion to 2048 when
 	// calling GetVersion. If the limit is exceeded the search attribute is not updated.
 	SDKFlagLimitChangeVersionSASize = 1
-	SDKFlagUnknown                  = math.MaxUint32
+	// LimitChangeVersionSASize return errors to child workflow execution future if the child workflow would
+	// fail in the synchronous path.
+	SDKFlagChildWorkflowErrorExecution = 2
+	SDKFlagUnknown                     = math.MaxUint32
 )
 
 func sdkFlagFromUint(value uint32) sdkFlag {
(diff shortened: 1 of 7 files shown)


test/replaytests/workflows.go
```diff
@@ -261,3 +262,68 @@ func VersionLoopWorkflow(ctx workflow.Context, changeID string, iterations int)
 	}
 	return workflow.Sleep(ctx, time.Second)
 }
+
+func ChildWorkflowWaitOnSignal(ctx workflow.Context) error {
+	workflow.GetSignalChannel(ctx, "unblock").Receive(ctx, nil)
+	return nil
+}
+
+func DuplicateChildWorkflow(ctx workflow.Context) error {
```
Quinn-With-Two-Ns, line 271: This workflow will replay differently before and after my change, hence the use of SDK flags.

internal/internal_flags.go
```diff
@@ -37,7 +37,10 @@ const (
 	// LimitChangeVersionSASize will limit the search attribute size of TemporalChangeVersion to 2048 when
 	// calling GetVersion. If the limit is exceeded the search attribute is not updated.
 	SDKFlagLimitChangeVersionSASize = 1
-	SDKFlagUnknown                  = math.MaxUint32
+	// LimitChangeVersionSASize return errors to child workflow execution future if the child workflow would
```
mmcshane, line 40: Wrong flag name in the comment here? (Might be a GH rendering glitch on my end)

mmcshane (approved):
LGTM, Just the comment thing

internal/internal_flags.go
Quinn-With-Two-Ns: no good catch, just a typo

Issue #1057, opened 2023-03-09
Double Update Accepted Event Due to Replay


mmcshane:
## Expected Behavior
An update should be accepted (or rejected) exactly once

## Actual Behavior
If a workflow replays due to something like a worker crash while there is an update in-flight, we don't re-run the validator for the update but we _do_ call Accept() on the update state machine. That leads to an update.Acceptance message being sent on the WFTCompleted which is duly made durable into the history by the server.

Note that one might consider the fact that the server doesn't catch the mistake to _also_ be a bug. That will be covered with a separate issue. 

## Steps to Reproduce the Problem

There will be a test in temporalio/features for this (it's how the problem was found in the first place). The general flow is:

1. Start a workflow with separate activity and workflow workers
1. Start an update that itself launches an activity
1. Block in the activity
1. Stop or otherwise kill the workflow worker
1. Unblock the blocked activity
1. Start the workflow worker again
1. Observe that the workflow replays to rebuild state but that the incorrect extra UpdateAccepted event is added to the history. This history will not replay correctly.

## Specifications

  - Version: v1.21
  - Platform: all

mmcshane:
Closed by #1059

PR #1059, opened 2023-03-13
Suppress update messages during replay

mmcshane:
## What was changed

When replaying, don't send response messages (Accept/Reject/Complete).

## Why?

These messages are redundant during replay - they should only be sent when not replaying.

## Checklist

1. Closes 

#1057 

3. How was this tested:

Unit test & features test

4. Any docs updates needed?

No

### DIFF
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -83,7 +83,12 @@ type (
 		GetFailureConverter() converter.FailureConverter
 		GetDataConverter() converter.DataConverter
 		Send(*protocolpb.Message)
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
Are they merely redundant, or actively incorrect? Either way the change makes sense.

mmcshane:
> Are they merely redundant, or actively incorrect? Either way the change makes sense.

Actively incorrect works. Aggressively erroneous. Wicked busted. Pick your phrasing.

Issue #1068, opened 2023-03-23
LocalActivity Tunnel not reset on stale cache lookup


Quinn-With-Two-Ns:
## Expected Behavior
If a worker is restarted (but not the process) cached workflows should work...

## Actual Behavior
Cached workflows can no longer start local activities.

## Steps to Reproduce the Problem

  1. launch a workflow that waits on signal then starts a local activity
  2. restart the worker, but not the process
  3. signal the workflow

The workflow will now be stuck because the workflow execution was using the local activity tunnel for the worker before the restart

Quinn-With-Two-Ns:
On namespace migration since the workflow context is not only holding a stale local activity channel, but the workflow task handler is also stale.

PR #1069, opened 2023-03-23
Reset local activity tunnel on stale lookup

Quinn-With-Two-Ns:
The stale reset code strikes again.

resolves: https://github.com/temporalio/sdk-go/issues/1068

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -508,6 +508,7 @@ func (w *workflowExecutionContextImpl) clearState() {
 	w.previousStartedEventID = 0
 	w.newCommands = nil
 	w.newMessages = nil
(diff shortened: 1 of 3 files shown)


internal/internal_task_handlers.go
```diff
@@ -508,6 +508,7 @@ func (w *workflowExecutionContextImpl) clearState() {
 	w.previousStartedEventID = 0
 	w.newCommands = nil
 	w.newMessages = nil
+	w.laTunnel = nil
```
Quinn-With-Two-Ns, line 511: `clearState()`should be called if the worker is restarted and the workflow is in cache because the new worker will get the full history so it will consider the cached stale.

cretz (approved):

PR #1070, opened 2023-03-24
Track the worker for each workflow context

Quinn-With-Two-Ns:
If a different worker is trying to use a workflow context from a different worker discard and create a new context. We need to recreate the context because the context has references to the worker so if it came from a different worker it cannot be used. 

resolves: https://github.com/temporalio/sdk-go/issues/1068

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -618,20 +618,28 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
 	if task.Query == nil || (task.Query != nil && !isFullHistory) {
 		workflowContext = wth.cache.getWorkflowContext(runID)
 	}
-
+	// Verify the cached state is current and for the correct worker
 	if workflowContext != nil {
 		workflowContext.Lock()
-		if task.Query != nil && !isFullHistory {
+		if task.Query != nil && !isFullHistory && wth == workflowContext.wth {
 			// query task and we have a valid cached state
 			metricsHandler.Counter(metrics.StickyCacheHit).Inc(1)
-		} else if history.Events[0].GetEventId() == workflowContext.previousStartedEventID+1 {
+		} else if history.Events[0].GetEventId() == workflowContext.previousStartedEventID+1 && wth == workflowContext.wth {
 			// non query task and we have a valid cached state
 			metricsHandler.Counter(metrics.StickyCacheHit).Inc(1)
 		} else {
-			// non query task and cached state is missing events, we need to discard the cached state and rebuild one.
-			_ = workflowContext.ResetIfStale(task, historyIterator)
+			// possible another task already destroyed this context.
+			if !workflowContext.IsDestroyed() {
+				// non query task and cached state is missing events, we need to discard the cached state and build a new one.
+				wth.cache.removeWorkflowContext(runID)
+				workflowContext.clearState()
+			}
+			workflowContext.Unlock(err)
+			workflowContext = nil
 		}
-	} else {
+	}
+	// If the workflow was not cached or the cache was stale.
+	if workflowContext == nil {
 		if !isFullHistory {
 			// we are getting partial history task, but cached state was already evicted.
 			// we need to reset history so we get events from beginning to replay/rebuild the state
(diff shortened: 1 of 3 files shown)


internal/internal_task_handlers.go
```diff
@@ -618,20 +618,28 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
...
+		if task.Query != nil && !isFullHistory && wth == workflowContext.wth {
 			// query task and we have a valid cached state
 			metricsHandler.Counter(metrics.StickyCacheHit).Inc(1)
-		} else if history.Events[0].GetEventId() == workflowContext.previousStartedEventID+1 {
+		} else if history.Events[0].GetEventId() == workflowContext.previousStartedEventID+1 && wth == workflowContext.wth {
 			// non query task and we have a valid cached state
 			metricsHandler.Counter(metrics.StickyCacheHit).Inc(1)
 		} else {
-			// non query task and cached state is missing events, we need to discard the cached state and rebuild one.
-			_ = workflowContext.ResetIfStale(task, historyIterator)
```
Quinn-With-Two-Ns, line 632: I removed the reset and replaced with a delete and create because this has been a source of multiple bugs in the go SDK when we hit this branch.

internal/internal_task_handlers.go
```diff
@@ -618,20 +618,28 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
 	if task.Query == nil || (task.Query != nil && !isFullHistory) {
 		workflowContext = wth.cache.getWorkflowContext(runID)
 	}
-
+	// Verify the cached state is current and for the correct worker
 	if workflowContext != nil {
 		workflowContext.Lock()
-		if task.Query != nil && !isFullHistory {
+		if task.Query != nil && !isFullHistory && wth == workflowContext.wth {
```
Quinn-With-Two-Ns, line 624: Another option is to assign each workers a UUID. That seemed unnecessary, but if there is a good justification I am fine switching it

internal/internal_task_handlers.go
```diff
@@ -618,20 +618,28 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
...
+		} else if history.Events[0].GetEventId() == workflowContext.previousStartedEventID+1 && wth == workflowContext.wth {
 			// non query task and we have a valid cached state
 			metricsHandler.Counter(metrics.StickyCacheHit).Inc(1)
 		} else {
-			// non query task and cached state is missing events, we need to discard the cached state and rebuild one.
-			_ = workflowContext.ResetIfStale(task, historyIterator)
+			// possible another task already destroyed this context.
+			if !workflowContext.IsDestroyed() {
+				// non query task and cached state is missing events, we need to discard the cached state and build a new one.
+				wth.cache.removeWorkflowContext(runID)
```
cretz, line 634: I assume it's by intention that this removes the context even if it is for another worker in this process? (I know this has always existed, just want to make sure we aren't concerned about it)

internal/internal_task_handlers.go
```diff
@@ -618,20 +618,28 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
...
+				workflowContext.clearState()
```
cretz, line 635: I think the old code cleared state of this regardless of whether it was destroyed. Do we need to do the same?

internal/internal_task_handlers.go
cretz: Do we want to put back the "Cached state staled" (sic) log here?

Quinn-With-Two-Ns: Yeah that is good point. I'll add it back

internal/internal_task_handlers.go
Quinn-With-Two-Ns: Yes that's by intention. If two workers are both using a workflow context that is already broken in some ways today (local activities and data converters being two examples)

internal/internal_task_handlers.go
Quinn-With-Two-Ns: if `IsDestroyed()` is true then `clearState()` should be a noop so I don't think it is needed?

dnr (approved):
I'm not familiar with the code of course, but I can confirm this fixes the issue I was seeing

internal/internal_task_handlers.go
cretz: I admit lack of familiarity here, so if it's a no-op, no prob

cretz (approved):
I admit I am heavily relying on integration tests here to make me feel good about this change. I have a hard time manually confirming the lifecycle of this context without a lot more work.

PR #1071, opened 2023-03-24
Add FrontendHostPort method to DevServer

sevein:
## What was changed

Added a new `FrontendHostPort` method to `testsuite.DevServer` that returns the `host:port` to the dev server, similar to Temporalite ([here](https://github.com/temporalio/temporalite/blob/2df426ad3426fb6ecc91d3b2e03274018bf14721/server.go#L158-L164)).

## Why?

This would be useful when the client.Client can't be shared with the unit being tested, e.g. a Command struct that only accepts os.Args, environment strings or other form of configuration.

### DIFF
--- a/testsuite/devserver.go
+++ b/testsuite/devserver.go
@@ -82,8 +82,9 @@ type DevServerOptions struct {
 
 // Temporal CLI based DevServer
 type DevServer struct {
-	cmd    *exec.Cmd
-	client client.Client
+	cmd              *exec.Cmd
+	client           client.Client
+	frontendHostPort string
 }
 
 // StartDevServer starts a Temporal CLI dev server process. This may download the server if not already downloaded.
@@ -121,7 +122,11 @@ func StartDevServer(ctx context.Context, options DevServerOptions) (*DevServer, 
 		return nil, err
 	}
 	clientOptions.Logger.Info("DevServer ready")
-	return &DevServer{client: returnedClient, cmd: cmd}, nil
+	return &DevServer{
+		client:           returnedClient,
+		cmd:              cmd,
+		frontendHostPort: clientOptions.HostPort,
+	}, nil
 }
 
 func prepareCommand(options *DevServerOptions, host, port, namespace string) []string {
@@ -365,4 +370,9 @@ func (s *DevServer) Stop() error {
 // Get a connected client, configured to work with the dev server.
 func (s *DevServer) Client() client.Client {
 	return s.client
+}
+
+// FrontendHostPort returns the host:port for this server.
+func (s *DevServer) FrontendHostPort() string {
+	return s.frontendHostPort
 }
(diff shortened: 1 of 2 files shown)


cretz (approved):
LGTM, @Quinn-With-Two-Ns - work for you?

sevein:
@cretz @Quinn-With-Two-Ns: apologize for the late changes, but I've made some additional updates after discovering a couple of things:

1. The tests in `testsuite/devserver_test.go` were not running in CI. I've fixed this in 9f63d87 - I moved the file to `/test`, the root of the integration tests and confirmed that they run now when using `make integration-test-zero-cache` and `make integration-test-normal-cache`. Previously, the test was not included in `make unit-test` because `INTEG_TEST_ROOT` matched both `test/` and `testsuite/`.
2. The tests in `testsuite/devserver_test.go` were failing due to the way `defer` was being used to stop the server and how Go evaluates the statement. Fixed in 2eeb22b.

cretz:
> The tests in testsuite/devserver_test.go were not running in CI

I don't think we should move them, we should just do what is necessary to make them work with `make unit-test`

> The tests in testsuite/devserver_test.go were failing due to the way defer was being used to stop the server and how Go evaluates the statement

Can you clarify this a bit? We want `defer` to work and we want these tests to show a normal way of starting and `defer`-stopping the test server instead of using `Cleanup`. Do we need to add exit code error checking to `Stop()`?

sevein:
> I don't think we should move them, we should just do what is necessary to make them work with make unit-test

Cool. I can do that, I thought it was integration after all because it's downloading the temporal-cli binary and running it - but maybe that can be considered an unit test.

> Can you clarify this a bit? We want defer to work and we want these tests to show a normal way of starting and defer-stopping the test server instead of using Cleanup. Do we need to add exit code error checking to Stop()?

`defer server.Stop()` works, but `defer require.NoError(t, server.Stop())` doesn't because the function parameters are evaluated right away (https://go.dev/play/p/Gf0A8g7FCwU). We don't have to use t.Cleanup - it's a habit of mine when testing, but we'd need to combine defer with a function literal, e.g.: https://go.dev/play/p/ytn5y_6ygPA.

sevein:
> Do we need to add exit code error checking to Stop()?

Not really, maybe that's actually a form of leaking that should be avoided. But I thought it'd be nice to look up the actual error since that's what we shoudl see when using cmd.Wait().

sevein:
Thanks, @cretz! I'll get to it as soon as possible.

cretz:
Opened https://github.com/temporalio/cli/issues/175 for issue where exit code is not always 0 here

sevein:
Done.

Quinn-With-Two-Ns (approved):

cretz (approved):

cretz:
Awesome, thanks for the contribution.

PR #1073, opened 2023-03-28
Message inference for more WF ctx reset paths

mmcshane:
## What was changed

Infer messages on an additional workflow context rebuild path - namely, a cache miss on the worker side.

## Why?

Need to replay messages on all rebuild paths for correctness

## Checklist

1. Closes 

2. How was this tested:

Unit test

3. Any docs updates needed?

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -558,6 +558,7 @@ func (wth *workflowTaskHandlerImpl) createWorkflowContext(task *workflowservice.
 	if taskQueue == nil || taskQueue.Name == "" {
 		return nil, errors.New("nil or empty TaskQueue in WorkflowExecutionStarted event")
 	}
+	task.Messages = append(inferMessages(task.GetHistory().GetEvents()), task.Messages...)
 
 	runID := task.WorkflowExecution.GetRunId()
 	workflowID := task.WorkflowExecution.GetWorkflowId()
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -558,6 +558,7 @@ func (wth *workflowTaskHandlerImpl) createWorkflowContext(task *workflowservice.
 	if taskQueue == nil || taskQueue.Name == "" {
 		return nil, errors.New("nil or empty TaskQueue in WorkflowExecutionStarted event")
 	}
+	task.Messages = append(inferMessages(task.GetHistory().GetEvents()), task.Messages...)
```
Quinn-With-Two-Ns, line 561: Why do we need to append here? shouldn't `Messages` be empty?

mmcshane: If there was one or more messages on the WFT that initiated this whole thing then that message won't be represented by a WorkflowUpdateAccepted event in the history so we want to propagate it along with the inferred messages

mmcshane: For any given instance of Update, yes. But you could have three UpdateAccepted in the history _plus_ an update.Request message (not yet accepted) on the WFT. In this case you would want to synthesize update.Request messages for replay _and also_ preserve the newer as-yet-unaccepted request that does not exist in history.

Quinn-With-Two-Ns (approved):

PR #1079, opened 2023-04-11
Add message IDs to update messages

mmcshane:
## What was changed

Adds message IDs to update messages

## Why?

This was just missed earlier. The IDs will be necessary to refer to messages by ID in the future.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):

Quinn-With-Two-Ns (approved):
I'll use the same convention in Java as well

PR #1083, opened 2023-04-13
Prevent inferring messages twice

mmcshane:
## What was changed

Remove a place where we infer messages from an event set.

## Why?

With the recent fixes to infer messages on all context reset/reload paths, this call to inferMessages has become superfluous. In fact, this call is causing us to add messages to the phony WFT twice, resulting in replay failures.

Tests in the features repo confirm the brokenness and the fix. The reason this wasn't caught as part of the PR feature tests is that update is disabled by default so update tests skip themselves.

## Checklist

1. Closes 

2. How was this tested:

All update tests in the features repo

3. Any docs updates needed?

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1334,7 +1334,6 @@ func (aw *WorkflowReplayer) replayWorkflowHistory(logger log.Logger, service wor
 		WorkflowExecution:      execution,
 		History:                history,
 		PreviousStartedEventId: math.MaxInt64,
-		Messages:               inferMessages(history.Events),
 	}
 
 	iterator := &historyIteratorImpl{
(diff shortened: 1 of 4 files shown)


Quinn-With-Two-Ns:
Should we cut a `1.22.1` release for this fix?

mmcshane:
P.S. 
Shame on me for not running those features tests for my previous changes. I could have sworn I did but I must not have because they all fail without this fix. Maybe I botched the go.mod replacement.

mmcshane:
> Should we cut a `1.22.1` release for this fix?

Crap. Yes.

Quinn-With-Two-Ns:
To avoid this in the future can we add a test that replays a workflow history with messages?

Quinn-With-Two-Ns (approved):

PR #1088, opened 2023-04-14
Async Update Polling

mmcshane:
## What was changed

Adds support for async updates
- A WorkflowUpdateHandle implementation that runs a poll RPC as part of the call to Get
- Add PollWorkflowUpdate to the client and the interceptor layers
- The ability to "rehydrate" an update reference into a handle 

## Why?

Planned feature

## Checklist

1. Closes 

2. How was this tested:

Unit tests here and a succeeding feature test that is waiting on this PR and an server PR

3. Any docs updates needed?

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -172,10 +172,10 @@ type (
 	// NOTE: Experimental
 	WorkflowUpdateHandle = internal.WorkflowUpdateHandle
 
-	// WorkflowUpdateRef encapsulates the parameters needed to unambiguously
+	// GetWorkflowUpdateHandleOptions encapsulates the parameters needed to unambiguously
 	// refer to a Workflow Update
 	// NOTE: Experimental
-	WorkflowUpdateRef = internal.WorkflowUpdateRef
+	GetWorkflowUpdateHandleOptions = internal.GetWorkflowUpdateHandleOptions
 
 	// UpdateWorkerBuildIdCompatibilityOptions is the input to Client.UpdateWorkerBuildIdCompatibility.
 	// NOTE: Experimental
@@ -512,7 +512,7 @@ type (
 		// which can be polled for an outcome. Note that runID is optional and
 		// if not specified the most recent runID will be used.
 		// NOTE: Experimental
-		GetWorkflowUpdateHandle(ref WorkflowUpdateRef) WorkflowUpdateHandle
+		GetWorkflowUpdateHandle(ref GetWorkflowUpdateHandleOptions) WorkflowUpdateHandle
 
 		// WorkflowService provides access to the underlying gRPC service. This should only be used for advanced use cases
 		// that cannot be accomplished via other Client methods. Unlike calls to other Client methods, calls directly to the
--- a/client/client.go
+++ b/client/client.go
@@ -23,7 +23,6 @@
 // THE SOFTWARE.
 
 //go:generate mockgen -copyright_file ../LICENSE -package client -source client.go -destination client_mock.go
-//go:generate go run ../internal/cmd/generateproxy/main.go
 
 // Package client is used by external programs to communicate with Temporal service.
 // NOTE: DO NOT USE THIS API INSIDE OF ANY WORKFLOW CODE!!!
--- a/client/client.go
+++ b/client/client.go
@@ -172,8 +172,8 @@ type (
 	// NOTE: Experimental
 	WorkflowUpdateHandle = internal.WorkflowUpdateHandle
 
-	// WorkflowUpdateRef reflects the fields needed to refer to a workflow
-	// update.
+	// WorkflowUpdateRef encapsulates the parameters needed to unambiguously
+	// refer to a Workflow Update
 	// NOTE: Experimental
 	WorkflowUpdateRef = internal.WorkflowUpdateRef
 
@@ -508,10 +508,11 @@ type (
 		// NOTE: Experimental
 		UpdateWorkflowWithOptions(ctx context.Context, request *UpdateWorkflowWithOptionsRequest) (WorkflowUpdateHandle, error)
 
-		// WorkflowUpdateHandleFromRef creates a handle to the referenced update
-		// which can be polled for an outcome.
+		// GetWorkflowUpdateHandle creates a handle to the referenced update
+		// which can be polled for an outcome. Note that runID is optional and
+		// if not specified the most recent runID will be used.
 		// NOTE: Experimental
-		WorkflowUpdateHandleFromRef(ref WorkflowUpdateRef) WorkflowUpdateHandle
+		GetWorkflowUpdateHandle(ref WorkflowUpdateRef) WorkflowUpdateHandle
 
 		// WorkflowService provides access to the underlying gRPC service. This should only be used for advanced use cases
 		// that cannot be accomplished via other Client methods. Unlike calls to other Client methods, calls directly to the
--- a/client/client.go
+++ b/client/client.go
@@ -172,6 +172,11 @@ type (
 	// NOTE: Experimental
 	WorkflowUpdateHandle = internal.WorkflowUpdateHandle
 
+	// WorkflowUpdateRef reflects the fields needed to refer to a workflow
+	// update.
+	// NOTE: Experimental
+	WorkflowUpdateRef = internal.WorkflowUpdateRef
+
 	// UpdateWorkerBuildIdCompatibilityOptions is the input to Client.UpdateWorkerBuildIdCompatibility.
 	// NOTE: Experimental
 	UpdateWorkerBuildIdCompatibilityOptions = internal.UpdateWorkerBuildIdCompatibilityOptions
@@ -502,6 +507,11 @@ type (
 		// directly from this function call.
 		// NOTE: Experimental
 		UpdateWorkflowWithOptions(ctx context.Context, request *UpdateWorkflowWithOptionsRequest) (WorkflowUpdateHandle, error)
+
+		// WorkflowUpdateHandleFromRef creates a handle to the referenced update
+		// which can be polled for an outcome.
+		// NOTE: Experimental
+		WorkflowUpdateHandleFromRef(ref WorkflowUpdateRef) WorkflowUpdateHandle
 
 		// WorkflowService provides access to the underlying gRPC service. This should only be used for advanced use cases
 		// that cannot be accomplished via other Client methods. Unlike calls to other Client methods, calls directly to the
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -763,8 +763,8 @@ type UpdateWorkflowWithOptionsRequest struct {
 	// then the server will reject the update request with an error.
 	FirstExecutionRunID string
 
-	// this isn't upstream in API yet
-	// WaitFor enumspb.WorkflowExecutionUpdateWaitEvent
+	// How this RPC should block on the server before returning.
+	WaitPolicy *updatepb.WaitPolicy
 }
 
 // WorkflowUpdateHandle is a handle to a workflow execution update process. The
@@ -786,13 +786,29 @@ type WorkflowUpdateHandle interface {
 	Get(ctx context.Context, valuePtr interface{}) error
 }
 
-// updateHandle is a dumb implementation of WorkflowExecutionUpdateHandle that
-// only supports the case where the output value is aready known at construction
-// time.
-type updateHandle struct {
+type WorkflowUpdateRef interface {
+	GetWorkflowExecution() *commonpb.WorkflowExecution
+	GetUpdateId() string
+}
+
+type baseUpdateHandle struct {
+	ref *updatepb.UpdateRef
+}
+
+// completedUpdateHandle is an UpdateHandle impelementation for use when the outcome
+// of the update is already known and the Get call can return immediately.
+type completedUpdateHandle struct {
+	baseUpdateHandle
 	value converter.EncodedValue
 	err   error
-	ref   *updatepb.UpdateRef
+}
+
+// lazyUpdateHandle represents and update that is not known to have completed
+// yet (i.e. the associated updatepb.Outcome is not known) and thus calling Get
+// will poll the server for the outcome.
+type lazyUpdateHandle struct {
+	baseUpdateHandle
+	client *WorkflowClient
 }
 
 // QueryWorkflowWithOptionsRequest is the request to QueryWorkflowWithOptions
@@ -1000,6 +1016,34 @@ func (wc *WorkflowClient) UpdateWorkflowWithOptions(
 		Args:                req.Args,
 		RunID:               req.RunID,
 		FirstExecutionRunID: req.FirstExecutionRunID,
+		WaitPolicy:          req.WaitPolicy,
+	})
+}
+
+func (wc *WorkflowClient) WorkflowUpdateHandleFromRef(ref WorkflowUpdateRef) WorkflowUpdateHandle {
+	return &lazyUpdateHandle{
+		client: wc,
+		baseUpdateHandle: baseUpdateHandle{
+			ref: &updatepb.UpdateRef{
+				WorkflowExecution: ref.GetWorkflowExecution(),
+				UpdateId:          ref.GetUpdateId(),
+			},
+		},
+	}
+}
+
+// PollWorkflowUpdate sends a request for the outcome of the specified update
+// through the interceptor chain.
+func (wc *WorkflowClient) PollWorkflowUpdate(
+	ctx context.Context,
+	ref *updatepb.UpdateRef,
+) (converter.EncodedValue, error) {
+	if err := wc.ensureInitialized(); err != nil {
+		return nil, err
+	}
+	ctx = contextWithNewHeader(ctx)
+	return wc.interceptor.PollWorkflowUpdate(ctx, &ClientPollWorkflowUpdateInput{
+		UpdateRef: ref,
 	})
 }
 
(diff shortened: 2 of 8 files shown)


internal/internal_workflow_client.go
```diff
@@ -1708,45 +1747,90 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
 			},
 		},
 	})
-	handle := &updateHandle{ref: resp.GetUpdateRef()}
 	if err != nil {
-		handle.err = err
-		return handle, nil
+		return nil, err
```
mmcshane, line 1751: This change is a compatibility break and it actually breaks the update-enablement-test in the features repo but I think we should do it...

There are two ways to deal with the sync/async wait policies: one way is to spawn a goroutine to do the rpc and return the handle immediately with a future in it (in all cases). The second way is to not return the handle until the RPC returns and there's no background goro to manage.

In the first case, all errors have to be tunneled through the Handle.Get call, even early server errors like a security error (which is what happens when update is disabled on the server). In the second approach where the initial UpdateWorkflowExecution waits for the RPC to return, if the server returns an early error then we have the opportunity to return the error directly and never a Handle. 

I want to use the second approach and that is what is implemented here.  In addition to being simpler in concept and implementation it also sidesteps an awkward situation where the RunID is not necessarily known until the RPC returns so what should the handle do in that case? A second Future for the RunID? Tell clients spin on GetRunID()?

client/client.go
```diff
@@ -504,6 +508,11 @@ type (
 		// NOTE: Experimental
 		UpdateWorkflowWithOptions(ctx context.Context, request *UpdateWorkflowWithOptionsRequest) (WorkflowUpdateHandle, error)
 
+		// WorkflowUpdateHandleFromRef creates a handle to the referenced update
+		// which can be polled for an outcome.
+		// NOTE: Experimental
+		WorkflowUpdateHandleFromRef(ref WorkflowUpdateRef) WorkflowUpdateHandle
```
cretz, line 514: I think this should be `GetWorkflowUpdateHandle` and should just take workflow information and update ID easily created by the caller (I think requiring the caller to create/use protos directly is a bit much).

internal/internal_workflow_client.go
```diff
@@ -1708,45 +1747,90 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
...
+	wait := updatepb.WaitPolicy{
+		LifecycleStage: enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED,
+	}
+	pollReq := workflowservice.PollWorkflowExecutionUpdateRequest{
+		Namespace:  w.client.namespace,
+		UpdateRef:  in.UpdateRef,
+		WaitPolicy: &wait,
+		Identity:   w.client.identity,
+	}
+	resp, err := w.client.workflowService.PollWorkflowExecutionUpdate(grpcCtx, &pollReq)
```
cretz, line 1791: This needs to be in a loop and the server needs to wait no longer than 60s before sending back a "not yet, try again". This helps with our philosophy of promising to proxy creators and others that we have no call lasting longer than one minute.

internal/internal_workflow_client.go
```diff
@@ -1708,45 +1747,90 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
...
+}
+
+func (w *workflowClientInterceptor) PollWorkflowUpdate(
+	ctx context.Context,
+	in *ClientPollWorkflowUpdateInput,
+) (converter.EncodedValue, error) {
+	// header, _ = headerPropagated(ctx, w.client.contextPropagators)
+	//todo header not in PollWorkflowUpdate
+
+	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
```
cretz, line 1780:
```suggestion
	grpcCtx, cancel := newGRPCContext(ctx, grpcLongPoll(true), defaultGrpcRetryParameters(ctx))
```

Or you can have `grpcLongPoll(isLongPoll)` and somehow have the user specify whether they want to wait on completion or have a non-blocking poll.

internal/internal_workflow_client.go
```diff
@@ -1708,45 +1747,90 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
...
+	}
+	if err := ch.value.Get(valuePtr); err != nil {
+		return err
+	}
+	return nil
+}
 
-	if uh.err != nil || valuePtr == nil {
-		return uh.err
+func (luh *lazyUpdateHandle) Get(ctx context.Context, valuePtr interface{}) error {
```
cretz, line 1830: I remember discussions around being able to wait for a stage instead of completion, is that a future PR? Also, is there any way for me to just check whether completed in a non-poll/non-blocking way?

mmcshane: That would be subsequent to this, yes. This is just enough to get e2e working being async and blocking on ACCEPTED

client/client.go
mmcshane: I'm find with that. @Quinn-With-Two-Ns do you have a feeling here?

internal/internal_workflow_client.go
mmcshane: Server is just going to return a GRPC timeout status here, it won't (and IMO shouldn't) do the early-return trick. But I can detect that and retry if the parent context still has some life left in it.

internal/internal_workflow_client.go
mmcshane: Spoke w/Chad  - we both see TryGet as a likely near-term request. Less clear is some form of "describe" API to observe execution state. We can wait for some usecases for the latter.

internal/internal_workflow_client.go
cretz: I think we should do the early return "trick". I don't think it's a failing state to have a poll say "nothing yet". This is different than a timeout which the caller cannot differentiate between a _real_ timeout (client or server side based on context deadline) and this one.

Whatever we choose to do, I think it needs to never take longer than a worker poll. And I believe we are telling people with proxies to set their max RPC timeouts at 80s. And the client should retry forever (or as long as their caller is willing) in the scenario where it isn't done yet. I think when you write the if statement to have the client retry on gRPC code timeout, it'll become clear that reusing timeout here is not the best.

client/client.go
```diff
@@ -173,6 +172,11 @@ type (
 	// NOTE: Experimental
 	WorkflowUpdateHandle = internal.WorkflowUpdateHandle
 
+	// WorkflowUpdateRef reflects the fields needed to refer to a workflow
+	// update.
+	// NOTE: Experimental
```
Sushisource, line 177: Probably worth explaining why/how this is distinguished from a handle

client/client.go
Sushisource: I'd +1 that

internal/internal_workflow_client.go
Sushisource: I would say since the retrying is hidden from the user, whether or not the early-return thing is used isn't relevant except to make it easy for us to detect and retry (and possibly the proxy thing, but I don't know if we actually advise users that).

The early return does feel slightly lame, though I can't really pinpoint exactly why. And actual timeout feels more semantically appropriate to me, and we can distinguish which one is which by including a bit of metadata.

As much as it's not my favorite argument, I can kinda get sticking with the early return to keep things consistent w/ other long pollers

internal/internal_workflow_client.go
mmcshane: Ok - did this. Renamed as suggested and turned the interface into a 3-field struct to be passed by value.

client/client.go
mmcshane: Ok, did this. Renamed as suggested and turned the parameter into a 2-field struct

internal/internal_workflow_client.go
mmcshane: Fixed up to loop until the parent context .Err() returns non-nil. The server is definitely returning a GRPC timeout status right now (and that was an intentional design choice) but I can verify that we're happy with the approach again.

internal/internal_workflow_client.go
mmcshane: This ended being two calls because the defaultGrpcRetryParameters doesn't nest well but anyway the long-poll flag is in place and we use a 60s timeout.

client/client.go
```diff
@@ -504,6 +508,12 @@ type (
 		// NOTE: Experimental
 		UpdateWorkflowWithOptions(ctx context.Context, request *UpdateWorkflowWithOptionsRequest) (WorkflowUpdateHandle, error)
 
+		// GetWorkflowUpdateHandle creates a handle to the referenced update
+		// which can be polled for an outcome. Note that runID is optional and
+		// if not specified the most recent runID will be used.
+		// NOTE: Experimental
+		GetWorkflowUpdateHandle(ref WorkflowUpdateRef) WorkflowUpdateHandle
```
cretz, line 515:
```suggestion
		GetWorkflowUpdateHandle(options GetWorkflowUpdateOptions) WorkflowUpdateHandle
```

I know it's a bit wordier, but reads better with regards to other calls here

internal/internal_workflow_client.go
```diff
@@ -63,6 +66,8 @@ const (
 	defaultGetHistoryTimeout = 65 * time.Second
 
 	getSystemInfoTimeout = 5 * time.Second
+
+	pollUpdateTimeout = 60 * time.Second
```
cretz, line 70: For other polls, we make sure the server returns 10 seconds before this I believe. Can you confirm that will happen here?

internal/internal_workflow_client.go
```diff
@@ -1708,45 +1765,104 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
...
+			grpcTimeout(pollUpdateTimeout),
+		)
+		ctx = context.WithValue(
+			ctx,
+			retry.ConfigKey,
+			createDynamicServiceRetryPolicy(ctx).GrpcRetryConfig(),
+		)
+		resp, err := w.client.workflowService.PollWorkflowExecutionUpdate(ctx, &pollReq)
+		cancel()
+		if err == context.DeadlineExceeded || status.Code(err) == codes.DeadlineExceeded {
```
cretz, line 1819: How do you know this isn't a legitimate deadline-exceeded from the server? Or how do you know this isn't a user-defined context deadline that we should bubble out? I think we need an empty response from the server here before SDK timeout. We don't need error codes appearing on normal operation. SDK side timeout IMO should only occur on actual error.

internal/internal_workflow_client.go
mmcshane: As it stands, that will not happen. Server prefers to return DeadlineExceeded here.

internal/internal_workflow_client.go
mmcshane: I think I'm missing your intent here ... all DeadlineExceededs from the server are legitimate?

If it's the parent context that failed, we won't loop through again (see guard on for loop) - I added a test for the same.

client/client.go
mmcshane: If we're going that direction, wouldn't it be `GetWorkflowUpdateHandleOptions`? I'm not sure what you're suggesting we be consistent with since this is just a constructor. Arguably it should just be a free function.

Quinn-With-Two-Ns: I thought the suggestion was to rename `WorkflowUpdateRef` -> `GetWorkflowUpdateOptions`.

internal/internal_workflow_client.go
Quinn-With-Two-Ns: >all DeadlineExceededs from the server are legitimate?
Is there not way to tell the difference between the server not responding and the update not being complete yet?

client/client.go
mmcshane: That is the suggestion - But if the pattern is Get${THING} takes an arg of type ${THING}Options (and I _think_ that's pattern being referred to as "other calls here") then $THING here is WorkflowUpdateHandle, not WorkflowUpdate

internal/internal_workflow_client.go
mmcshane: If you're fetching the result of an update, I don't think there exists a meaningful difference.

client/client.go
Quinn-With-Two-Ns: ah I see your point. Personally I am fine with `GetWorkflowUpdateHandleOptions`

internal/internal_workflow_client.go
cretz: > Server prefers to return DeadlineExceeded here.

I don't think that's true on normal long polls. I think it returns an empty success and I think it should here. Helps with error code tracking and such. I don't think it's an error case for your update to not be done yet.

internal/internal_workflow_client.go
cretz: My suggestion is very simply when an update is not done yet, do not return an error, return some other thing that says not done yet. Then let all timeout errors bubble as they may (whether user-defined or defined by SDK as longer than the poll-success-but-empty-try-again time).

internal/internal_workflow_client.go
mmcshane: > I don't think that's true on normal long polls

This is correct. It was an intentional decision to use DEADLINE_EXCEEDED here and understood to be different than the queue polling calls.

internal/internal_workflow_client.go
mmcshane: So an api like

```
Handle {
    Get(ctx context.Context, outPtr interface{})(error, bool)
}
```

?

I have to say that for me that's too weirdly different from Future. I understand that this api isn't a future but a somewhat semantically symmetric api is worth something.

cretz: I am only talking about internal behavior here. Sorry for the confusion. Don't use gRPC errors for not-yet-keep-trying timeout from the server. The external user-facing Go API should remain the same. This comment is specifically on `if err == context.DeadlineExceeded || status.Code(err) == codes.DeadlineExceeded {` and I don't think we should always assume those mean retry.

mmcshane: Okay so how about this: given that the server is currently implemented to return DEADLINE_EXCEEDED and that this was specifically chosen by that team, I don't want to remove the error check. That said, I can add a check for `outcome == nil` here and _also_ treat that as "not yet" so that if the server changes in the future we'll have that escape hatch. Thoughts on a compromise?

cretz: > I don't want to remove the error check

So what if I call get update with my own context that has a deadline in 10s? You will not respect that. I think we need to revisit how it's currently implemented. I think it should be like all other poll calls IIUC in that it is not an error to not be done yet.

internal/internal_workflow_client.go
Quinn-With-Two-Ns: Yes, I am basically trying to make sure these expected timeouts won't count against our SLAs

internal/internal_workflow_client.go
Quinn-With-Two-Ns: > I think we need to revisit how it's currently implemented.

In general I agree with Chad that the server design in strange and should be looked at again, but I don't think this PR is the right forum to discuss a server design decision.

mmcshane: > This code will not respect that and retry

The ctx used in the grpc call is derived from the caller supplied context so I'm not sure where you get that idea? Unless something in the internals is discarding the parentCtx. One hopes not. But I can do another unit test to confirm.

mmcshane: I updated the unit test to confirm that if a caller submits a PollWorkflowUpdate with a context timeout of 50ms, that is indeed the timeout that is propagated through to the server and that if no user-supplied timeout is given, the default is used.

internal/internal_workflow_client.go
mmcshane: Yep, confirmed with the server team who requested this style in the first place.

internal/internal_workflow_client.go
cretz: I assume my `temporal_long_request_failure` metrics are gonna start going up if my update, say, takes 3 minutes?

mmcshane: (a) can we run the this-code-will-not-respect-the-passed-context argument to ground before shifting? I really don't see in the code where that's the case and am asking you to make that clear

(b) re: the failure metric - I have no idea and will look into it. But again, the server's intention is to use a DEADLINE_EXCEEDED here so we may need to make allowances like we've done with longPoll.

cretz: > can we run the this-code-will-not-respect-the-passed-context argument to ground before shifting? 

I think your context check in the loop is enough. So we can shift now. User context is no longer a concern. It just popped in my head as the first of many concerns about reusing errors for flow here.

> But again, the server's intention is to use a DEADLINE_EXCEEDED here so we may need to make allowances like we've done with longPoll.

Can we run the "server's intention" argument into the ground before shifting? Why is that the intention? Why can that not change? Do they do that on other long polls?

mmcshane: It can absolutely change but you're talking about something that was discussed and settled a few weeks ago so it's not the sort of thing that's going to happen in the context of this PR.

cretz: :+1: I vote for changing it. If it can't change now, ok, but I think it should. I think we'll be happier for it.

mmcshane: If it helps, the code change to treat a nil response.Outcome as a signal to retry (subject to the parent context, natch) means that we can change the server without breaking field SDKs.

cretz (approved):
Approving with the known caveat that we will change server to not use error codes to signify update not completed yet

PR #1090, opened 2023-04-15
Exposing WorkflowExecution Started Event attributes on the Workflow Info

### LINKED ISSUE #1087: Expose OriginalRunId and FirstRunId
They are already provided in Java SDK: https://github.com/temporalio/sdk-java/blob/4f91956597b6218f18de39509e19b61d01f9a08c/temporal-sdk/src/main/java/io/temporal/workflow/WorkflowInfo.java#L85

And also Cadence Golang SDK:
https://github.com/uber-go/cadence-client/blob/18c143aa5b28f24e59c8579a26a20db62f54e101/internal/workflow.go#L1097

edmondop:
## What was changed

This pull requests address issue #1087 to expose additional fields from the WorkflowExecutionStartedEvent into the SDK
 

## Why?
It was a missing feature that the Java SDK already implemented, and that is implemented in Cadence.

## Checklist

- [] Add unit tests

1. Closes #1087 

2. How was this tested:

Not tested

3. Any docs updates needed?

Comments describing the fields should be good

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -944,9 +944,11 @@ func (wc *workflowEnvironmentInterceptor) ExecuteChildWorkflow(ctx Context, chil
 
 // WorkflowInfo information about currently executing workflow
 type WorkflowInfo struct {
-	WorkflowExecution        WorkflowExecution
-	OriginalRunID            string // The original runID before resetting. Using it instead of current runID can make workflow decision determinstic after reset. See also FirstRunId
-	FirstRunID               string // The very first original RunId of the current Workflow Execution preserved along the chain of ContinueAsNew, Retry, Cron and Reset. Identifies the whole Runs chain of Workflow Execution.
+	WorkflowExecution WorkflowExecution
+	// The original runID before resetting. Using it instead of current runID can make workflow decision deterministic after reset. See also FirstRunId
+	OriginalRunID string
+	// The very first original RunId of the current Workflow Execution preserved along the chain of ContinueAsNew, Retry, Cron and Reset. Identifies the whole Runs chain of Workflow Execution.
+	FirstRunID               string
 	WorkflowType             WorkflowType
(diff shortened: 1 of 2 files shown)


edmondop:
@longquanzheng can you confirm that within the WithActivityContext, this information should not be available on the context?

When the activity context is created deserializing the Task here https://github.com/edmondop/sdk-go/blob/master/internal/activity.go#L278 the `FirstRunId` and `OriginalRunId` are not available since they are not in the protobuf scheme of https://github.com/temporalio/api/blob/ae312b0724003957b96fb966e3fe25a02abaade4/temporal/api/common/v1/message.proto#L78 and therefore they are not propagated here https://github.com/temporalio/api/blob/ae312b0724003957b96fb966e3fe25a02abaade4/temporal/api/workflowservice/v1/request_response.proto#L381

internal/workflow.go
```diff
@@ -945,6 +945,8 @@ func (wc *workflowEnvironmentInterceptor) ExecuteChildWorkflow(ctx Context, chil
 // WorkflowInfo information about currently executing workflow
 type WorkflowInfo struct {
 	WorkflowExecution        WorkflowExecution
+	OriginalRunID            string // The original runID before resetting. Using it instead of current runID can make workflow decision determinstic after reset. See also FirstRunId
+	FirstRunID               string // The very first original RunId of the current Workflow Execution preserved along the chain of ContinueAsNew, Retry, Cron and Reset. Identifies the whole Runs chain of Workflow Execution.
```
cretz, line 949: Should document fields above the field

edmondop: Fixed with https://github.com/temporalio/sdk-go/pull/1090/commits/b4d8ee53878cf0b410202c456872a5bab5827164

Quinn-With-Two-Ns (approved):
LGTM, thank you for the contribution

PR #1093, opened 2023-04-20
Error on Registering Activity with Workflow Context

### LINKED ISSUE #1061: Warn on calling an `ExecuteActivity` on a function that takes a `workflow.Context`
A issue new users run into is they define an activity that takes a `workflow.Context` as a parameter, often just mistyping `context.Context`. The error users get is confusing as it says they aren't passing enough parameters, but if they don't notice the types they expect the context to be injected.

```
reflect: Call with too few input arguments
```

The other issue with passing a `workflow.Context` to an activity is it indicated a potential error if the user did intend to pass the workflow version because it means they may be trying to invoke a workflow function outside a workflow

shannonjmtan:
## What was changed
Check for workflow context usage within activity arguments as this is usually developer error (accidental and misunderstanding). Note that this would likely fail later on anyways when executing the function.

## Why?
Prevent accidental errors.

## Checklist

1. Closes #1061 

2. How was this tested: Added unit tests for new behavior, existing tests should verify current behavior. 

3. Any docs updates needed? No as this is not stated as an allowed behavior (nor will it even work).

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -723,6 +723,15 @@ func validateFnFormat(fnType reflect.Type, isWorkflow bool) error {
 		if !isWorkflowContext(fnType.In(0)) {
 			return fmt.Errorf("expected first argument to be workflow.Context but found %s", fnType.In(0))
 		}
+	} else {
+		// For activities, check that workflow context is not accidentally provided
+		// Activities registered with structs will have their receiver as the first argument so confirm it is not
+		// in the first two arguments
+		for i := 0; i < fnType.NumIn() && i < 2; i++ {
+			if isWorkflowContext(fnType.In(i)) {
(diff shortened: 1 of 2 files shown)


shannontan94:
The original issue said to warn, but I figured this was equivalent as there was a flag to bypass failing functions. I've also not worked that closely with reflection so I'm open to alterations.

cretz (approved):
LGTM. Will let @Quinn-With-Two-Ns look too.

Quinn-With-Two-Ns:
LGTM, just need to fix the unit test

shannonjmtan:
Looking at the error, I think this is unrelated to my PR? Let me know if it I'm misreading.

Quinn-With-Two-Ns:
@shannonjmtan yep unrelated

PR #1101, opened 2023-05-11
Properly store the original failurepb on decode

Quinn-With-Two-Ns:
Fix a discrepancy between failure encoded and decoded. On decode the `failureHolder` stored the decoded failure, on encoded it assumed the `failureHolder` had an encoded failure.

### DIFF
--- a/internal/failure_converter.go
+++ b/internal/failure_converter.go
@@ -25,6 +25,7 @@ package internal
 import (
 	"errors"
 
+	"github.com/gogo/protobuf/proto"
 	commonpb "go.temporal.io/api/common/v1"
 	failurepb "go.temporal.io/api/failure/v1"
 	"go.temporal.io/sdk/converter"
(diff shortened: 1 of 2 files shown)


Sushisource (approved):

internal/failure_converter.go
```diff
@@ -174,7 +174,9 @@ func (dfc *DefaultFailureConverter) FailureToError(failure *failurepb.Failure) e
 	if failure == nil {
 		return nil
 	}
-
+	// Copy the original future to pass to the failureHolder
+	var orignalFailure failurepb.Failure
+	orignalFailure.XXX_Merge(failure)
```
cretz, line 179:
```suggestion
	originalFailure := proto.Clone(failure).(*failurepb.Failure)
```

This is the better proto cloning approach (`XXX_` methods should usually be avoided). Also, I fixed the spelling of `orignamFailure` (sic)

internal/failure_converter.go
```diff
@@ -174,7 +175,8 @@ func (dfc *DefaultFailureConverter) FailureToError(failure *failurepb.Failure) e
 	if failure == nil {
 		return nil
 	}
-
+	// Copy the original future to pass to the failureHolder
+	originalFailure := proto.Clone(failure).(*failurepb.Failure)
```
cretz, line 179: This is not free of course, but I suspect the perf cost is negligible

cretz (approved):

PR #1105, opened 2023-05-18
Add option to ignore unknown fields when protojson unmarshalling

mattpercy-anz:
## What was changed
We would like to use the temporal data converters for encoding/decoding payloads, but the proto JSON converter doesn't expose the unmarshalling option to ignore unknown fields. This adds that option, but maintains the current behaviour for anyone wanting it that way

## Why?
We don't want unknown fields in proto JSON to cause decode errors

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):
LGTM. Will let @Quinn-With-Two-Ns take a peek.

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
LGTM, thank you for the contribution

PR #1106, opened 2023-05-18
Set Update ID client side

Quinn-With-Two-Ns:
Match what we will do in the Java SDK and set the update ID client side if the user does not set one.

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -59,8 +59,10 @@ import (
 )
 
 // Assert that structs do indeed implement the interfaces
-var _ Client = (*WorkflowClient)(nil)
-var _ NamespaceClient = (*namespaceClient)(nil)
+var (
+	_ Client          = (*WorkflowClient)(nil)
+	_ NamespaceClient = (*namespaceClient)(nil)
+)
 
 const (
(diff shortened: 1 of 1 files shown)


mmcshane (approved):

internal/internal_workflow_client.go
```diff
@@ -1021,6 +1023,10 @@ func (wc *WorkflowClient) UpdateWorkflowWithOptions(
 	if err := wc.ensureInitialized(); err != nil {
 		return nil, err
 	}
+	// Default update ID
+	if req.UpdateID == "" {
+		req.UpdateID = uuid.New()
```
cretz, line 1028:
```suggestion
		reqVar := *req
		req = &reqVar
		req.UpdateID = uuid.New()
```

Shallow copy avoids the mutation of user value (if they are updating in a loop for example and pre-made the request, there'd be a problem). I'm sure we don't do this everywhere, but this is why many Go functions do not accept pointer-based options.

I should have brought this up in #989 and guess I didn't. This matches query but not start workflow. I don't have a strong opinion I guess on accepting pointer or non in that case, but we should not mutate the param.

cretz (approved):

PR #1107, opened 2023-05-18
Send ProtocolMessageCommands where appropriate

mmcshane:
## What was changed

Adds ProtocolMessageCommands to outgoing commands for a WFT response for update accepted and completed messages but not rejections as rejections are not written as events. Entries in the message outbox now contain the message itself as well as a predicate to test history events that may have been created in response to this message. In this way the messages and the event expectations are fully defined in internal_update.go and the event handling and replay code can remain ignorant of the mapping between message types and the resulting event types.

## Why?

Allows for update messages to be sequence with respect to commands.

## Checklist

1. Closes 

2. How was this tested:

New and existing unit tests. Features tests all still pass.

3. Any docs updates needed?

No


internal/internal_event_handlers.go
```diff
@@ -302,17 +316,40 @@ func (s *scheduledSignal) handle(result *commonpb.Payloads, err error) {
 	s.callback(result, err)
 }
 
-func (wc *workflowEnvironmentImpl) drainOutbox(sink *[]*protocolpb.Message) {
-	*sink = append(*sink, wc.outbox...)
-	wc.outbox = nil
+func (o *outbox) drain(sink *[]*protocolpb.Message) {
```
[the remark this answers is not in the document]
mmcshane: I had intended to have the api work like io.Reader to give tthe caller control over the number of messages. But it's sort of pointless. I'll make it more normal.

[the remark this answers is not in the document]
mmcshane: done

internal/internal_task_handlers.go
```diff
@@ -278,7 +278,9 @@ func isCommandEvent(eventType enumspb.EventType) bool {
 		enumspb.EVENT_TYPE_REQUEST_CANCEL_EXTERNAL_WORKFLOW_EXECUTION_INITIATED,
 		enumspb.EVENT_TYPE_SIGNAL_EXTERNAL_WORKFLOW_EXECUTION_INITIATED,
 		enumspb.EVENT_TYPE_UPSERT_WORKFLOW_SEARCH_ATTRIBUTES,
-		enumspb.EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED:
+		enumspb.EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED,
+		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED,
+		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_COMPLETED:
```
Quinn-With-Two-Ns, line 283: Do we want to include rejected for future proofing?

internal/internal_task_handlers.go
```diff
@@ -1265,6 +1270,10 @@ func skipDeterministicCheckForEvent(e *historypb.HistoryEvent) bool {
 		return true
 	case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_TIMED_OUT:
 		return true
+	case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED,
```
Quinn-With-Two-Ns, line 1273: Can we check `EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED` ? On replay we skip running the validation and skip adding the accepted message to the outbox no?

test/replaytests/replay_test.go
```diff
@@ -311,8 +311,17 @@ func (s *replayTestSuite) TestUnkownSDKFlag() {
...
 	replayer.RegisterWorkflow(UpdateWorkflow)
-	err := replayer.ReplayWorkflowHistoryFromJSONFile(ilog.NewDefaultLogger(), "update.json")
-	require.NoError(s.T(), err)
+
+	for _, name := range [...]string{
+		"update_before_protocol_message_command",
+		"update_with_protocol_message_command",
+	} {
+		s.Run(name, func() {
+			err := replayer.ReplayWorkflowHistoryFromJSONFile(ilog.NewDefaultLogger(), name+".json")
```
Quinn-With-Two-Ns, line 320: can we add some tests for failure cases? Another integration test case I would love to see is accepting, completing and accepting+completing an update in the last workflow task of a history and verifying the commands are present in history. Currently all those situations are broken and I was expecting `ProtocolMessageCommands` to fix them.

internal/internal_task_handlers.go
mmcshane: We don't run validation but we do send the message. The outbox gets drained into a side structure for the nondeterminism check so that we don't actually emit the message back to the server (same as commands)

test/replaytests/replay_test.go
mmcshane: I can do that. FWIW that was fixed even before ProtocoMessageCommand went in on the server. There _was_ a time when it was broken though - you didn't hallucinate that.

mmcshane: I added replay tests for complete and accept+complete in the last WFT. I'll add a test for a non-deterministic update implementation ... can you think of anything else apart from very obviously broken things like events in the wrong place that would get caught in `features`?

Quinn-With-Two-Ns: >can you think of anything else apart from very obviously broken things like events in the wrong place that would get caught in features

not sure I understand the question? could you rephrase it?

mmcshane: Basically just trying to brainstorm negative test scenarios beyond non-determinism.

Quinn-With-Two-Ns: No other then trying different events in the update handler to cause non determinism

internal/internal_event_handlers.go
```diff
@@ -1217,6 +1250,7 @@ func (weh *workflowExecutionEventHandlerImpl) handleWorkflowExecutionStarted(
 	if err != nil {
 		return err
 	}
+	weh.sdkFlags.tryUse(SDKFlagProtocolMessageCommand, !weh.isReplay)
```
Quinn-With-Two-Ns, line 1253: nit: Since you spent a lot of time debugging this so it might be worth documenting why we set this on start rather then first use

Quinn-With-Two-Ns (approved):

internal/internal_event_handlers.go
mmcshane: (added in handleWorkflowExecutionStarted)

cretz (approved):

PR #1108, opened 2023-05-19
Allow update panics to bubble up

mmcshane:
## What was changed

Remove panic recovery guards around invoking user update code. These panics will now escape the handler and bubble up, eventually being processed by the workflow's panic handling policy. 

## Why?

This makes update execution consistent with other workflow code.

## Checklist

1. Closes 

2. How was this tested:

Features tests to be landed after this PR

3. Any docs updates needed?

No

### DIFF
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -317,28 +317,14 @@ func newUpdateHandler(
 	}, nil
 }
 
-// validate invokes the update's validation function and maps panics to errors.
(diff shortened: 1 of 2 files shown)


cretz (approved):
Sorry if I led you astray in the early versions of this. What you have here is definitely the way to go. There _could_ be an argument here that validators should catch those panics, but meh. I forget what we do in Java on non-Temporal-exception of validators.

PR #1109, opened 2023-05-19
Check for child workflow already started error when setting env.testError

cdavis-joy:
## What was changed

Check if the error passed to Complete is a `ChildWorkflowExecutionAlreadyStarted` error.

## Why?

Was trying to test that a workflow was acting a certain way if it received this error when starting and waiting on a child workflow by returning `&temporal.ChildWorkflowExecutionAlreadyStarted` via `OnWorkflow` in a test case. 

Found that since we weren't checking for that error when setting `env.testError`, it was getting passed to `env.failureConverter.ErrorToFailure` which resulted in the error being passed to the parent workflow as a `*temporal.ApplicationError` which made `IsWorkflowExecutionAlreadyStartedError` return false.

## Checklist

1. Did not open an issue for this.

2. How was this tested:

It was tested by testing a workflow that calls `IsWorkflowExecutionAlreadyStartedError` on the error returned by `ExecuteChildWorkflow(...).Get(...)`, and returning `&temporal.ChildWorkflowExecutionAlreadyStarted` via `OnWorkflow().Return()`

3. Any docs updates needed?
Don't think so!


cretz (approved):
LGTM, will let @Quinn-With-Two-Ns review as well

internal/internal_workflow_testsuite.go
```diff
@@ -872,7 +872,8 @@ func (env *testWorkflowEnvironmentImpl) Complete(result *commonpb.Payloads, err
 		var timeoutErr *TimeoutError
 		var workflowPanicErr *workflowPanicError
 		var workflowExecutionAlreadyStartedErr *serviceerror.WorkflowExecutionAlreadyStarted
-		if errors.As(err, &canceledErr) || errors.As(err, &continueAsNewErr) || errors.As(err, &timeoutErr) || errors.As(err, &workflowExecutionAlreadyStartedErr) {
+		var childWorkflowExecutionAlreadyStartedErr *ChildWorkflowExecutionAlreadyStartedError
```
Quinn-With-Two-Ns, line 875: shouldn't this condition be protected by `isChildWorkflow`? If the `ChildWorkflowExecutionAlreadyStartedError` error is returned by the parent workflow I believe it would be wrapped?

cdavis-joy: That's a good question. I've created a test to demonstrate this issue on my fork here: https://github.com/cdavis-joy/sdk-go/commit/08b2c10c44cc4d8570793d0e9b14b770c5b2ac39


The `env.isChildWorkflow` is correctly wrapping `env.testError` in a `NewChildWorkflowExecutionError`, but the problem is that on this line, before that's happening, we don't see the error as any of these specific types we're checking for so `env.testError` gets set to whatever the `FailureConverter` returns. The `FailureConverter` is not aware of the `ChildWorkflowExecutionAlreadyStartedError` type, so it just returns an `ApplicationError`.

Sorry, I hope that helps clarify and not further muddies 😅

Quinn-With-Two-Ns: To clarify my point was if you run the same conceptual workflow as your `workflowFn` against a real temporal server your test will fail.

cdavis-joy: Oh really? What makes you say that? 

We've reproduced this issue with our usage of Temporal: a workflow attempted to start the same child workflow as one that was already running, and that child workflow execution failed with `ChildWorkflowExecutionAlreadyStartedError`

Quinn-With-Two-Ns: Just writing my own workflow to test I get this `Failure`

```
{
  "message": "child workflow execution error",
  "source": "GoSDK",
  "stackTrace": "",
  "encodedAttributes": null,
  "cause": {
    "message": "child workflow execution already started",
    "source": "GoSDK",
    "stackTrace": "",
    "encodedAttributes": null,
    "cause": null,
    "applicationFailureInfo": {
      "type": "ChildWorkflowExecutionAlreadyStartedError",
      "nonRetryable": false,
      "details": null
    }
  },
  "childWorkflowExecutionFailureInfo": {
    "namespace": "default",
    "workflowExecution": {
      "workflowId": "ABC-SIMPLE-CHILD-WORKFLOW-ID",
      "runId": ""
    },
    "workflowType": {
      "name": "SampleChildWorkflow"
    },
    "initiatedEventId": "14",
    "startedEventId": "0",
    "retryState": "NonRetryableFailure"
  }
}
```

cdavis-joy: You're correct that the test I wrote would fail against a Temporal server, since that child workflow isn't already running. The intent of this change is to make it so that you can mock that other workflow already running, by having the mocked child workflow execution return that error. 

As it stands now the testsuite is not capable of correctly returning a `ChildWorkflowExecutionAlreadyStartedError` from a mocked workflow. Instead the testsuite returns a `ChildWorkflowExecutionError` wrapping an `ApplicationError`.

Quinn-With-Two-Ns: My point is the sdk running against a real server does return a `ChildWorkflowExecutionError` wrapping an `applicationFailureInfo`

cdavis-joy: Sure, that's true, however If you were to check the type of that error in your sample workflow, you would see that it is of type `ChildWorkflowExecutionAlreadyStartedError`. When running that workflow in the testsuite however, that would not be the case. That's the outcome I'm trying to get here: that we can properly test a workflow that expects a `ChildWorkflowExecutionAlreadyStartedError`.

Maybe setting up a call to discuss offline might help? I understand this isn't a revolutionary change or anything, but it would be great for our use case to be able to accurately test this scenario.

cdavis-joy: To perhaps add some context, there is an existing unit test that covers that we can return a certain error from a mocked workflow: https://github.com/temporalio/sdk-go/blob/master/internal/internal_workflow_testsuite_test.go#L3494

That test was added as a part of this PR https://github.com/uber-go/cadence-client/pull/969 which is basically the same thing I am trying to accomplish. 

Really my change would have made sense to add when `ChildWorkflowExecutionAlreadyStartedError` was introduced, and it might make sense now for me to update that integration test to return that error from the mocked workflow instead of the outdated (at least in the child workflow context) error that it is testing for now.

cdavis-joy: @Quinn-With-Two-Ns any more thoughts on this change?

Quinn-With-Two-Ns: No looks good I missed the `NewChildWorkflowExecutionError` lower down so i originally thought your mocks was returning the wrong error

PR #1110, opened 2023-05-19
Propagate UseBuildIDForVersioning to RespondWorkflowTaskCompletedResp…

bergundy:
…onse

Not tested but I think it should be enough to cover this with feature tests.


dnr (approved):
cool, wasn't sure if there was a better way to do it

PR #1113, opened 2023-05-22
Switch to the official coveralls uploader

mmcshane:
## What was changed

Switch to the official coveralls uploader.

Secondarily
- parallelize integration tests
- tag coverage uploads (unit vs. integration)
- update actions/checkout and actions/setup-go dependencies

## Why?

goveralls has stopped working (gets a 500 from the server) and rather than debug someone's personal project we're switching to something official and more likely to remain supported.

## Checklist

1. Closes 

2. How was this tested:

Running ci workflow in my fork

3. Any docs updates needed?

No

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,53 +6,103 @@ on:
       - master
 
 jobs:
-  build-and-test:
+  build:
     runs-on: ubuntu-latest
     steps:
       - name: Print build information
         run: "echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}"
-      - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


cretz (approved):
Quite a bit for just the coveralls part, but looks good to me while we only support on OS anyways (gets a bit more annoying for hoping between jobs and OS's). Will want @Quinn-With-Two-Ns to review.

Quinn-With-Two-Ns (approved):

mmcshane:
Turns out this was a problem on the Coveralls side and the problem has gone away. I'm going to open a smaller PR that switches to the official uploader but doesn't do the matrix stuff as it doesn't speed the end-to-end process up at all and adds complexity.

PR #1123, opened 2023-06-01
Add retry to UpdateWorkflow call

mmcshane:
## What was changed

Add retry to UpdateWorkflow call

## Why?

Same retry as already exists for outcome polling. This was missed in previous changesets.

## Checklist

1. Closes 

2. How was this tested:

New unit tests here

3. Any docs updates needed?

No

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -1745,9 +1745,11 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
 	if err != nil {
 		return nil, err
 	}
-	header, _ := headerPropagated(ctx, w.client.contextPropagators)
-
-	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
+	header, err := headerPropagated(ctx, w.client.contextPropagators)
+	if err != nil {
+		return nil, err
+	}
+	grpcCtx, cancel := newGRPCContext(ctx, grpcTimeout(pollUpdateTimeout), grpcLongPoll(true), defaultGrpcRetryParameters(ctx))
 	defer cancel()
 	wfexec := &commonpb.WorkflowExecution{
 		WorkflowId: in.WorkflowID,


internal/internal_workflow_client.go
```diff
@@ -1769,28 +1765,42 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
...
+	var resp *workflowservice.UpdateWorkflowExecutionResponse
+	for parentCtx.Err() == nil {
+		grpcCtx, cancel := newGRPCContext(parentCtx, defaultGrpcRetryParameters(parentCtx))
+		var err error
+		resp, err = w.client.workflowService.UpdateWorkflowExecution(grpcCtx, &req)
+		cancel()
+		if err == context.DeadlineExceeded ||
+			status.Code(err) == codes.DeadlineExceeded {
+			continue
+		}
```
cretz, line 1779:
```suggestion
	ctx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
	defer cancel()
	for {
		resp, err := w.client.workflowService.UpdateWorkflowExecution(ctx, &req)
		if grpcCtx.Err() == nil && (err == context.DeadlineExceeded ||
			status.Code(err) == codes.DeadlineExceeded) {
			continue
		}
```

This would be clearer,  but I don't understand why you are looping. gRPC retry is handled by an interceptor internally. Manually looping should not be necessary. If an error is given back here, you should fail the caller. So:

```suggestion
	ctx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
	defer cancel()
	resp, err := w.client.workflowService.UpdateWorkflowExecution(ctx, &req)
```

cretz (commented):
I do not understand the purpose of this PR. gRPC retry in inherent in our calls, you don't need your own loop.

internal/internal_workflow_client.go
mmcshane: Our retry code clamps the timeout to [1s ... 10s]. If the client gives a 20s context timeout here should we return a deadline exceeded after 10s? That seems a little strange so I put in this retry loop for as long as the parent ctx is alive. Another option would be to not use the standard grpcContextBuilder since that's where the 10s cap comes from

mmcshane: Or it looks like I can override the timeout clamping by passing the grpcTimeout option

mmcshane: There is not "not yet but not failed" for this call. It either makes it to the requested stage of execution or it times out waiting for that to occur.

This timeout stuff is messy. Hard to tell if the grpcTimeout options is ignored totally or just ignored if the retry params are not set. Still looking.

cretz: Does a retry of the call w/ the same update ID reconnect to an existing one if present? Meaning, what if the update succeeded server side but server crashed before delivery of response to user? Won't a retry cause an ID-already-exists error?

I think to do this right we're gonna have to relay timeout error to the user until we have a fire-and-forget (or even better a "try to do this synchronously, switch to async after X duration). We can't really wait longer than a minute for infra/HA/LB reasons and I'm guessing retry is not truly idempotent. We may have to say "we're sorry, single-call sync update does not currently support timeouts > 1m at this time, please wait for implicit-sync-to-async feature".

mmcshane: > Does a retry of the call w/ the same update ID reconnect to an existing one if present? 
It does, yes.

I'm not too worried about the end deadline now. It looks like the default grpc retry policy will retry for 60s regardless. But the default retrier will make its _first_ call with a timeout of 200ms, and that seems seriously short for this use case. I'm trying to wade through the indirections of the retry logic to settle on something longer without being a long poll.

mmcshane: > What if the update succeeded server side but server crashed before delivery of response to user?

In this case the server returns the update outcome for that update ID

The 200ms comes from the implementation of defaultGrpcRetryParameters, so anything using that will make an initial request with a 200ms deadline (`internal.retryServiceOperationInitialInterval`). It is not at all clear how the timeout on the builder struct interacts with the retry policy. I think if there's a retry policy then that timeout is not observed.

cretz: > In this case the server returns the update outcome for that update ID

So I, as a user, may want to use update ID like I use workflow ID (in other SDKs by default, in Go if I set `WorkflowExecutionErrorWhenAlreadyStarted` as `true`) where I get an error if the ID has already been used. Can I not do that? One would expect that when duplicate IDs are disallowed, I should be able to get an error saying it was already used.

> The 200ms comes from the implementation of defaultGrpcRetryParameters,

Does that happen for all of our long-poll calls? We should behave like other long poll calls if we can.

Quinn-With-Two-Ns:
I presume we will want this in the Java SDK as well?

internal/internal_workflow_client.go
mmcshane: Turns out that was a misread on my part. the 200ms is the initial _backoff_ after a retryable error is returned (e.g. Unavailable). Absent any error being returned, the default context timeout is used.

mmcshane:
Big rework of this PR as I have come to understand how the retry policy is built and interpreted. Long story short, this configuration gives us the desired properties of (1) marking this RPC as a long-poll for the purposes of metric categorization, and (2) using the appropriate long-poll timeout (60s) and retrying with exponential backoff in response to retyrable errors.

cretz (approved):
Awesome

mmcshane:
@Quinn-With-Two-Ns regarding your question for also doing this for java - I'm not as familiar with how java has implemented retries but I think what we probably want are the two properties above: use the long-poll context timeout and mark the RPC as "allowed to be slow" from the sdk-side.

Issue #1127, opened 2023-06-05
TestWorkflowEnvironment support mocking Activity without pre-registering but doesn't support for Workflows


edmondop:
## Expected Behavior

Workflows that are mocked are automatically registered when using a `reflect.Func` as a `reflect.TypeOf(workflow)` when invoking the api `OnWorkflow`, the same way this happens using `OnActivity`

## Actual Behavior

In the `OnActivity` method, the Activity is registered here: https://github.com/temporalio/sdk-go/blob/master/internal/workflow_testsuite.go#LL338C1-L340C36

while in the OnWorkflow method there is no automatic registration https://github.com/temporalio/sdk-go/blob/master/internal/workflow_testsuite.go#L399

Also, when the activity is mocked by Name, since registering the activity requires a `reflect.Func`, the function fails if the activity hasn't been registered before. 

https://github.com/temporalio/sdk-go/blob/master/internal/workflow_testsuite.go#L345-L349

while this is not happening when mocking the workflow by name (which doesn't work, but doesn't return an error)

PR #1128, opened 2023-06-05
Aligning OnWorkflow with OnActivity and adding unit tests to confirm behavior is coherent

edmondop:
See #1127

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -388,19 +388,26 @@ func (e *TestWorkflowEnvironment) OnWorkflow(workflow interface{}, args ...inter
 	var call *mock.Call
 	switch fType.Kind() {
 	case reflect.Func:
-		fnType := reflect.TypeOf(workflow)
-		if err := validateFnFormat(fnType, true); err != nil {
+		if err := validateFnFormat(fType, true); err != nil {
 			panic(err)
 		}
 		fnName, _ := getWorkflowFunctionName(e.impl.registry, workflow)
 		if alias, ok := e.impl.registry.getWorkflowAlias(fnName); ok {
 			fnName = alias
 		}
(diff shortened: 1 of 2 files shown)


internal/workflow_testsuite.go
```diff
@@ -388,19 +388,26 @@ func (e *TestWorkflowEnvironment) OnWorkflow(workflow interface{}, args ...inter
...
 		}
+		e.impl.registry.RegisterWorkflowWithOptions(workflow, RegisterWorkflowOptions{DisableAlreadyRegisteredCheck: true})
 		call = e.mock.On(fnName, args...)
 	case reflect.String:
-		call = e.mock.On(workflow.(string), args...)
+		workflowName := workflow.(string)
+		_, ok := e.impl.registry.getWorkflowFn(workflowName)
+		if !ok {
+			registered := strings.Join(e.impl.registry.getRegisteredWorkflowTypes(), ", ")
+			panic(fmt.Sprintf("Workflow \""+workflowName+"\" is not registered with the TestWorkflowEnvironment, "+
```
cretz, line 405: My concern here is people might have a test case that sets `OnWorkflow` as a string before they register it. And now that's gonna panic because they did it in that order. I think for compatibility reasons we can't panic here. We have to leave today's behavior as it was if possible.

edmondop: 👍  agreed, didn't think about, it let me fix

cretz: Nah, I think you can just silently let it behave like today

edmondop: Fixed with https://github.com/temporalio/sdk-go/pull/1128/commits/a820a9c85268b6d660edd2f18bd6a66baf4e692a

cretz (approved):
LGTM, work for you @Quinn-With-Two-Ns?

Quinn-With-Two-Ns:
LGTM, thank you for the contribution @edmondop

PR #1129, opened 2023-06-07
Fix stopping dev server on Windows 

feedmeapples:
## What was changed

Starting killing the dev server process on Windows since sending signals (and SIGTERM or os.Interrupt) is not supported 

## Why?

Ensuring the dev server is stopped and doesn't leave hanging

## Checklist

1. Closes 

2. How was this tested:

Added test

3. Any docs updates needed?

### DIFF
--- a/testsuite/devserver.go
+++ b/testsuite/devserver.go
@@ -39,7 +39,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-	"syscall"
 	"time"
 
 	"go.temporal.io/sdk/client"
@@ -361,12 +360,7 @@ func retryFor(maxAttempts int, interval time.Duration, cond func() error) error 
 
 // Stop the running server and wait for shutdown to complete. Error is propagated from server shutdown.
 func (s *DevServer) Stop() error {
-	if runtime.GOOS == "windows" {
-		// On Windows, we can't send SIGTERM to the process, so we just kill it.
-		return s.cmd.Process.Kill()
-	}
-
-	if err := s.cmd.Process.Signal(syscall.SIGTERM); err != nil {
+	if err := sendInterrupt(s.cmd.Process); err != nil {
 		return err
 	}
 	return s.cmd.Wait()
--- a/testsuite/devserver.go
+++ b/testsuite/devserver.go
@@ -361,6 +361,11 @@ func retryFor(maxAttempts int, interval time.Duration, cond func() error) error 
 
 // Stop the running server and wait for shutdown to complete. Error is propagated from server shutdown.
 func (s *DevServer) Stop() error {
+	if runtime.GOOS == "windows" {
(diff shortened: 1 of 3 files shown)


bergundy (commented):
See how to interrupt a process in windows:

https://github.com/temporalio/omes/blob/main/cmd/util_windows.go

feedmeapples:
> See how to interrupt a process in windows:
> 
> https://github.com/temporalio/omes/blob/main/cmd/util_windows.go

added here! Also not sure if it is actually meant to be interrupt in the `omes` repo. I've replaced this with `SIGTERM` in dev server

feedmeapples:
having unrelated issue with sdk-go v1.23.0. Recreated the PR on top of 1.21.2 until fixing CLI's CI with newer versions

testsuite/devserver.go
```diff
@@ -82,7 +81,7 @@ type DevServerOptions struct {
 
 // Temporal CLI based DevServer
 type DevServer struct {
-	cmd    *exec.Cmd
+	Cmd    *exec.Cmd
```
feedmeapples, line 84: not a fan of making it public, though this seems to be the only option to test the RC after interrupting the server (especially for Windows)

feedmeapples: unless we don't test this and only rely on `/cli`'s tests

cretz:
> having unrelated issue with sdk-go v1.23.0

Make sure the `go.temporal.io/api` and `go.temporal.io/sdk` are _both_ on the exact same versions as the `go.mod` of the server version you are using. Even if it compiles, there can be other issues if you mismatch. Server version should decide the other two.

testsuite/interrupt_windows.go
```diff
@@ -0,0 +1,67 @@
+//go:build windows
```
cretz, line 1: This is not needed, it's assumed by the `_windows.go` filename

testsuite/interrupt_windows.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"os"
+	"syscall"
+
+	"golang.org/x/sys/windows"
+)
+
+// sendTerminate calls the break event on the given process for graceful shutdown.
+func sendTerminate(process *os.Process) error {
```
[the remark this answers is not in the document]
feedmeapples: > > having unrelated issue with sdk-go v1.23.0
> 
> Make sure the `go.temporal.io/api` and `go.temporal.io/sdk` are _both_ on the exact same versions as the `go.mod` of the server version you are using. Even if it compiles, there can be other issues if you mismatch. Server version should decide the other two.

changed back to the `master` branch

cretz (approved):
Unfortunately we aren't running CI in Windows here like other ones, but LGTM and I am just hoping it works

PR #1135, opened 2023-06-10
Allow ignoring function ref issues

Sushisource:
Allows `//workflowcheck:ignore` to work on workflow function registration sites

### DIFF
--- a/contrib/tools/workflowcheck/workflow/checker.go
+++ b/contrib/tools/workflowcheck/workflow/checker.go
@@ -157,9 +157,16 @@ func (c *Checker) Run(pass *analysis.Pass) error {
 		determinism.UpdateIgnoreMap(pass.Fset, file, ignoreMap)
 
 		ast.Inspect(file, func(n ast.Node) bool {
-			// Only handle calls
+			// Only handle calls with followable function pointers
+			_, isIgnored := ignoreMap[n]
+			for k := range ignoreMap {
+				asExprStmt, _ := k.(*ast.ExprStmt)
+				if asExprStmt != nil && asExprStmt.X == n {
+					isIgnored = true
+				}
+			}
 			callExpr, _ := n.(*ast.CallExpr)
-			if callExpr == nil {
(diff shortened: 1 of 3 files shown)


Quinn-With-Two-Ns (approved):

contrib/tools/workflowcheck/workflow/checker.go
```diff
@@ -176,7 +180,17 @@ func (c *Checker) Run(pass *analysis.Pass) error {
 			}
 			// Report if couldn't get type
 			if fn == nil {
-				pass.Reportf(callExpr.Args[0].Pos(), "unrecognized function reference format")
+				_, isIgnored := ignoreMap[n]
```
cretz, line 183: I think a better impl is to change `if callExpr == nil {` above to `if callExpr == nil || isIgnored {`. You are ignoring `RegisterWorkflow` call so check it early.

contrib/tools/workflowcheck/workflow/checker.go
```diff
@@ -152,10 +152,21 @@ func (c *Checker) Run(pass *analysis.Pass) error {
...
 		ast.Inspect(file, func(n ast.Node) bool {
-			// Only handle calls
+			// Only handle calls with followable function pointers
+			_, isIgnored := ignoreMap[n]
+			for k := range ignoreMap {
+				asExprStmt, _ := k.(*ast.ExprStmt)
+				if asExprStmt != nil && asExprStmt.X == n {
+					isIgnored = true
+				}
+			}
```
cretz, line 167: Is this for loop needed?

cretz, line 167: In retrospect, you could have just:

```suggestion
			if _, isIgnored := ignoreMap[n]; isIgnored {
				return false
			}
```

Sushisource: That didn't seem to work when I was trying it - I needed to get this inner node for things to work

PR #1136, opened 2023-06-13
Change message look ahead

Quinn-With-Two-Ns:
Draft solution to fix the issue seen during load testing update.

```
process event for temporal-bench [panic]:
go.temporal.io/sdk/internal.panicIllegalState(...)
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_command_state_machine.go:456
go.temporal.io/sdk/internal.(*commandsHelper).handleActivityTaskScheduled(0x14000cb7a40, {0x1400077a32a, 0x3}, 0x0?)
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_command_state_machine.go:1039 +0xec
go.temporal.io/sdk/internal.(*workflowExecutionEventHandlerImpl).ProcessEvent(0x1400102b8d8, 0x14000c9e740, 0x3?, 0x0)
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_event_handlers.go:1068 +0x210
go.temporal.io/sdk/internal.(*workflowExecutionContextImpl).ProcessWorkflowTask(0x1400102f290, 0x140011367b0)
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_task_handlers.go:967 +0xf04
go.temporal.io/sdk/internal.(*workflowTaskHandlerImpl).ProcessWorkflowTask(0x14000437450, 0x140011367b0, 0x1400127aa80)
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_task_handlers.go:788 +0x34c
go.temporal.io/sdk/internal.(*workflowTaskPoller).processWorkflowTask(0x1400047c6c0, 0x140011367b0)
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_task_pollers.go:342 +0x2bc
go.temporal.io/sdk/internal.(*workflowTaskPoller).ProcessTask(0x1400047c6c0, {0x101394c20?, 0x140011367b0})
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_task_pollers.go:315 +0x74
go.temporal.io/sdk/internal.(*baseWorker).processTask(0x140004483c0, {0x1013947a0, 0x14000ca61d0})
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_worker_base.go:440 +0x154
created by go.temporal.io/sdk/internal.(*baseWorker).runTaskDispatcher
	/Users/quinnklassen/go/pkg/mod/go.temporal.io/sdk@v1.23.0/internal/internal_worker_base.go:334 +0x98
```

Current theory is the previous implementation of look ahead was only looking one "page" of history ahead on certain replay scenarios. On multiple page histories with update accepted not on the original page this could cause non determinism as the state machine would not be aware of them. This PR fixes this by doing the look ahead the same way we look ahead for `markers`

Marking as draft because this deserves a unit test and integration test. Also some minor refactoring

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -49,6 +49,7 @@ import (
 	"go.temporal.io/api/workflowservice/v1"
 
 	"go.temporal.io/sdk/internal/common/retry"
+	"go.temporal.io/sdk/internal/protocol"
 
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/internal/common"
@@ -292,11 +293,11 @@ func isCommandEvent(eventType enumspb.EventType) bool {
 
 // NextCommandEvents returns events that there processed as new by the next command.
 // TODO(maxim): Refactor to return a struct instead of multiple parameters
-func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, binaryChecksum string, sdkFlags []sdkFlag, err error) {
+func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, binaryChecksum string, sdkFlags []sdkFlag, msgs []*protocolpb.Message, err error) {
 	if eh.next == nil {
-		eh.next, _, eh.nextFlags, err = eh.nextCommandEvents()
+		eh.next, _, eh.nextFlags, _, err = eh.nextCommandEvents()
 		if err != nil {
-			return result, markers, eh.binaryChecksum, sdkFlags, err
+			return result, markers, eh.binaryChecksum, sdkFlags, msgs, err
 		}
 	}
 
@@ -304,9 +305,9 @@ func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, marker
 	checksum := eh.binaryChecksum
 	sdkFlags = eh.nextFlags
 	if len(result) > 0 {
-		eh.next, markers, eh.nextFlags, err = eh.nextCommandEvents()
+		eh.next, markers, eh.nextFlags, msgs, err = eh.nextCommandEvents()
 	}
-	return result, markers, checksum, sdkFlags, err
+	return result, markers, checksum, sdkFlags, msgs, err
 }
 
 func (eh *history) hasMoreEvents() bool {
@@ -334,12 +335,12 @@ func (eh *history) verifyAllEventsProcessed() error {
 	return nil
 }
 
-func (eh *history) nextCommandEvents() (nextEvents []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, sdkFlags []sdkFlag, err error) {
+func (eh *history) nextCommandEvents() (nextEvents []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, sdkFlags []sdkFlag, msgs []*protocolpb.Message, err error) {
 	if eh.currentIndex == len(eh.loadedEvents) && !eh.hasMoreEvents() {
 		if err := eh.verifyAllEventsProcessed(); err != nil {
-			return nil, nil, nil, err
+			return nil, nil, nil, nil, err
 		}
-		return []*historypb.HistoryEvent{}, []*historypb.HistoryEvent{}, []sdkFlag{}, nil
+		return []*historypb.HistoryEvent{}, []*historypb.HistoryEvent{}, []sdkFlag{}, []*protocolpb.Message{}, nil
 	}
 
 	// Process events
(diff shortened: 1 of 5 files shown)


internal/internal_task_handlers.go
```diff
@@ -292,21 +293,21 @@ func isCommandEvent(eventType enumspb.EventType) bool {
 
 // NextCommandEvents returns events that there processed as new by the next command.
 // TODO(maxim): Refactor to return a struct instead of multiple parameters
-func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, binaryChecksum string, sdkFlags []sdkFlag, err error) {
+func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, binaryChecksum string, sdkFlags []sdkFlag, msgs []*protocolpb.Message, err error) {
```
Sushisource, line 296: I think it's probably time to address that TODO. These functions return a hilarious amount of things

internal/internal_task_handlers_interfaces_test.go
```diff
@@ -243,3 +244,67 @@ func (s *PollLayerInterfacesTestSuite) TestGetNextCommandsSdkFlags() {
...
+		}),
+		createTestEventWorkflowTaskScheduled(5, &historypb.WorkflowTaskScheduledEventAttributes{TaskQueue: &taskqueuepb.TaskQueue{Name: taskQueue}}),
+		createTestEventWorkflowTaskStarted(6),
+		{
+			EventId:   7,
+			EventType: enumspb.EVENT_TYPE_WORKFLOW_TASK_FAILED,
+		},
+		{
+			EventId:   8,
+			EventType: enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED,
```
Sushisource, line 267: How can WFT be failed but also have update accepted?

internal/internal_task_handlers_test.go
```diff
@@ -1223,6 +1238,122 @@ func (t *TaskHandlersTestSuite) TestWorkflowTask_PageToken() {
...
+			return &historypb.History{Events: nextEvents}, nil, nil
+		},
+	}
+	taskHandler := newWorkflowTaskHandler(params, nil, t.registry)
+	request, _, err := taskHandler.ProcessWorkflowTask(&workflowTask{
+		task: task, historyIterator: historyIterator,
+	}, nil)
+	response := request.(*workflowservice.RespondWorkflowTaskCompletedRequest)
+	t.NoError(err)
+	t.NotNil(response)
```
Sushisource, line 1354: Can this verify something more about the response?

Quinn-With-Two-Ns: No, the fact that there is no error is real the important part of this test. It means the history replayed successfully

internal/internal_task_handlers_interfaces_test.go
Quinn-With-Two-Ns: good catch, yeah i am missing an event here

internal/internal_task_handlers.go
Quinn-With-Two-Ns: I agree we need to refactor this. I left it out in this PR because this code is brittle and I'd prefer to the a refactor where we fix all the issues in this function including potential pagination issues and refactor to use a struct.

cretz (approved):

internal/internal_task_handlers.go
mmcshane: I kinda want to see how far we can go with this.

internal/internal_task_handlers.go
```diff
@@ -391,6 +392,15 @@ OrderEvents:
 		default:
 			if isPreloadMarkerEvent(event) {
 				markers = append(markers, event)
+			} else if attrs := event.GetWorkflowExecutionUpdateAcceptedEventAttributes(); attrs != nil {
+				msgs = append(msgs, &protocolpb.Message{
```
mmcshane, line 396: Would have preferred to have kept inferMessage for here (but adjusted to operate on a single event) but that can be a refactor.

internal/internal_task_handlers.go
```diff
@@ -391,6 +392,15 @@ OrderEvents:
 		default:
 			if isPreloadMarkerEvent(event) {
 				markers = append(markers, event)
+			} else if attrs := event.GetWorkflowExecutionUpdateAcceptedEventAttributes(); attrs != nil {
```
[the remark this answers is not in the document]
mmcshane: As long as getHistory returns just events then we're somewhat boxed in here. I don't think it breaks the concept, more like it acknowledges that protocols interact with events, which is true. Moving message inference to the server side could someday be workable though.

mmcshane (approved):

internal/internal_task_handlers.go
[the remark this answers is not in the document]
cretz: Yeah, I can understand we didn't want to go down the generic protocol event path. We just need to remember then if we add any fields or any protocol-to-event logic server side, the inverse logic may have to be updated SDK side.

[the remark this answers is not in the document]
paulnpdev: maybe add a well-placed comment in the server code warning future implementors of this coupling?

PR #1140, opened 2023-06-15
Isolate Ctrl+Break to dev server instead of console

feedmeapples:
## What was changed

Isolated the Ctrl + break event to dev server

## Why?

Stopping dev server with `devserver.Stop()` would result in the origin process to be interrupted instead of just the devserver. Ex. Go tests with devserver preemptively exited with RC `1`

## Checklist

1. Closes 

2. How was this tested:

tested with e2e tests in https://github.com/temporalio/cli/pull/257

3. Any docs updates needed?

### DIFF
--- a/testsuite/interrupt_windows.go
+++ b/testsuite/interrupt_windows.go
@@ -61,5 +61,17 @@ func sendInterrupt(process *os.Process) error {
 	if r1 == 0 {
 		return err
 	}
+
+	// Free the console after sending the interrupt
+	// To prevent sending the CTRL+BREAK signal to all processes sharing the console
+	f, err = dll.FindProc("FreeConsole")
+	if err != nil {
+		return err
+	}
+	r1, _, err = f.Call()
+	if r1 == 0 {
+		return err
+	}
+
 	return nil
 }
--- a/testsuite/interrupt_windows.go
+++ b/testsuite/interrupt_windows.go
@@ -24,7 +24,6 @@ package testsuite
 
 import (
 	"os"
-	"syscall"
 
 	"golang.org/x/sys/windows"
 )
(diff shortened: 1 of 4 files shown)


bergundy (approved):
Blindly approving, @cretz you might want to double check this

testsuite/interrupt_windows.go
```diff
@@ -61,5 +61,17 @@ func sendInterrupt(process *os.Process) error {
 	if r1 == 0 {
 		return err
 	}
+
+	// Free the console after sending the interrupt
```
cretz, line 65: Interesting. I did some reading. If easy enough, can you try to remove this new code and the code that does the `AttachConsole` and `SetConsoleCtrlHandler`? Basically, try to make this entire function be:

```go
func sendInterrupt(process *os.Process) error {
	dll, err := syscall.LoadDLL("kernel32.dll")
	if err != nil {
		return err
	}
	f, err := d.FindProc("GenerateConsoleCtrlEvent")
	if err != nil {
		return err
	}
	r, _, e := p.Call(uintptr(syscall.CTRL_BREAK_EVENT), uintptr(process.Pid))
	if r == 0 {
		return err
	}
	return nil
}
```

And see if that's good enough. Also note I moved to [`syscall`](https://pkg.go.dev/syscall?GOOS=windows) instead of `golang.org/x/sys/windows` dependency.

feedmeapples: this doesn't seem to work yet, debugging. syscall.LoadDLL is deprecated since go 1.8, debugging with windows.LoadDLL atm

feedmeapples: this still sends CTRL_BREAK_EVENT to all processes that share the console

feedmeapples: one alternative option would be start the dev server in a new console with on Windows with
```
cmd.SysProcAttr = &syscall.SysProcAttr{
    CreationFlags: syscall.CREATE_NEW_CONSOLE,
}
```

here https://github.com/temporalio/sdk-go/blob/1a0d5f99f30134236c60c45d89fa909600ff451a/testsuite/devserver.go#L112

feedmeapples: or both on Windows and unix

cretz (approved):
Great, thanks!

testsuite/process_windows.go
```diff
@@ -24,41 +24,35 @@ package testsuite
...
 	"syscall"
 
 	"golang.org/x/sys/windows"
 )
 
+// newCmd creates a new command with the given executable path and arguments.
+func newCmd(exePath string, args ...string) *exec.Cmd {
+	cmd := exec.Command(exePath, args...)
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		// isolate the process from the current console so interrupt signals are not sent to it
```
feedmeapples, line 37:
```suggestion
		// isolate the process and signals sent to it from the current console
```

PR #1141, opened 2023-06-15
Fix inconsistent BuildID field name & empty TQ in intent check

Sushisource:
Two minor fixes before release


bergundy (approved):
Seems like we should have caught this empty TQ intent in tests.
Can you make sure we have coverage for the various cases? (unit tests are good enough).

Quinn-With-Two-Ns (approved):

Issue #1144, opened 2023-06-16
Non-determinism while replaying mutable side-effect behind workflow versioning 


nitesh237:
## Expected Behavior
When replaying the mutable side-effect behind workflow versioning, the command should be generated in the same exact order as they were generated during the execution.

## Actual Behavior
We are seeing non-determinism error while replaying due to missing mutable-side effect marker.

```
PanicError: lookup failed for scheduledEventID to activityID: scheduleEventID: 13, activityID: 13 
process event for example-task-queue [panic]:
go.temporal.io/sdk/internal.panicIllegalState(...)
	/Users/nitesh/go/pkg/mod/go.temporal.io/sdk@v1.22.2/internal/internal_command_state_machine.go:455
go.temporal.io/sdk/internal.(*commandsHelper).handleActivityTaskScheduled(0x14001a315e0, {0x14001507798, 0x2}, 0x40?)
	/Users/nitesh/go/pkg/mod/go.temporal.io/sdk@v1.22.2/internal/internal_command_state_machine.go:1038 +0xf0
go.temporal.io/sdk/internal.(*workflowExecutionEventHandlerImpl).ProcessEvent(0x1400089ef90, 0x14001a29b00, 0x20?, 0x0)
	/Users/nitesh/go/pkg/mod/go.temporal.io/sdk@v1.22.2/internal/internal_event_handlers.go:1031 +0x248
go.temporal.io/sdk/internal.(*workflowExecutionContextImpl).ProcessWorkflowTask(0x140015fd3b0, 0x1400137c600)
	/Users/nitesh/go/pkg/mod/go.temporal.io/sdk@v1.22.2/internal/internal_task_handlers.go:962 +0xed0
go.temporal.io/sdk/internal.(*workflowTaskHandlerImpl).ProcessWorkflowTask(0x14000cd6c30, 0x1400137c600, 0x140018daa20)
	/Users/nitesh/go/pkg/mod/go.temporal.io/sdk@v1.22.2/internal/internal_task_handlers.go:780 +0x364
go.temporal.io/sdk/internal.(*workflowTaskPoller).processWorkflowTask(0x14000ccc900, 0x1400137c600)
	/Users/nitesh/go/pkg/mod/go.temporal.io/sdk@v1.22.2/internal/internal_task_pollers.go:329 +0x2d0
[...]

PR #1146, opened 2023-06-23
Use getNextID() when handling mutable side effects

Quinn-With-Two-Ns:
## What was changed
Use getNextID() when handling mutable side effects

## Why?
Using `nextCommandEventID` directly does not account for version markers leading to non determinism on replay

## Checklist

1. Closes 
close https://github.com/temporalio/sdk-go/issues/1144

2. How was this tested:
Added a test based on the users reproduction

3. Any docs updates needed?

### DIFF
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -1171,7 +1171,7 @@ func (h *commandsHelper) recordLocalActivityMarker(activityID string, details ma
 func (h *commandsHelper) recordMutableSideEffectMarker(mutableSideEffectID string, callCountHint int, data *commonpb.Payloads, dc converter.DataConverter) commandStateMachine {
 	// In order to avoid duplicate marker IDs, we must append the counter to the
 	// user-provided ID
-	mutableSideEffectID = fmt.Sprintf("%v_%v", mutableSideEffectID, h.nextCommandEventID)
+	mutableSideEffectID = fmt.Sprintf("%v_%v", mutableSideEffectID, h.getNextID())
 	markerID := fmt.Sprintf("%v_%v", mutableSideEffectMarkerName, mutableSideEffectID)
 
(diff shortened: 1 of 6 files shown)


internal/internal_command_state_machine.go
```diff
@@ -1171,7 +1171,7 @@ func (h *commandsHelper) recordLocalActivityMarker(activityID string, details ma
 func (h *commandsHelper) recordMutableSideEffectMarker(mutableSideEffectID string, callCountHint int, data *commonpb.Payloads, dc converter.DataConverter) commandStateMachine {
 	// In order to avoid duplicate marker IDs, we must append the counter to the
 	// user-provided ID
-	mutableSideEffectID = fmt.Sprintf("%v_%v", mutableSideEffectID, h.nextCommandEventID)
+	mutableSideEffectID = fmt.Sprintf("%v_%v", mutableSideEffectID, h.getNextID())
```
cretz, line 1174: Can you think of any concerns where fixing this breaks replay for those where replay succeeded before? Is there a case where a successfully replaying workflow will now get a different ID for mutable side effect? Or does it feel like this would only get a different ID in cases where replay was probably broken anyways?

(not that I care _that_ much because mutable side effects are so rarely used and were broken on replay until a few months back)

Quinn-With-Two-Ns: If it did get a different ID, replay would be broken. Depending on what exactly the workflow did after the mutable side effect call it may not have been detected. 

This change should not cause any workflows that were replaying to break.

Sushisource (approved):
Makes sense. Might be good to include a history that used mut side effect successfully before this change w/ some other command following it to verify the case Chad mentioned

Quinn-With-Two-Ns:
>Might be good to include a history that used mut side effect successfully before this change 

Yep, that is the history included, unless you mean it **replayed** successfully. In which case that is not possible because it would never successfully replay before my fix.

PR #1147, opened 2023-06-23
Fix history paging on looking for WFT Fail

Quinn-With-Two-Ns:
Make the Go SDK more robust to empty pages from the server when running `IsNextWorkflowTaskFailed`.

 I saw without the change to `IsNextWorkflowTaskFailed` the unit test `TestEmptyPages` would return different events and messages if I used the `MockHistoryIterator` to return empty pages every other request or not.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -224,14 +224,19 @@ func (eh *history) IsReplayEvent(event *historypb.HistoryEvent) bool {
 	return event.GetEventId() <= eh.workflowTask.task.GetPreviousStartedEventId() || isCommandEvent(event.GetEventType())
 }
 
+// IsNextWorkflowTaskFailed checks if the workflow task failed or completed. If it did complete returns some information
+// on the completed workflow task.
 func (eh *history) IsNextWorkflowTaskFailed() (isFailed bool, binaryChecksum string, flags []sdkFlag, err error) {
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -224,14 +224,19 @@ func (eh *history) IsReplayEvent(event *historypb.HistoryEvent) bool {
...
 }
 
+// IsNextWorkflowTaskFailed checks if the workflow task failed or completed. If it did complete returns some information
+// on the completed workflow task.
 func (eh *history) IsNextWorkflowTaskFailed() (isFailed bool, binaryChecksum string, flags []sdkFlag, err error) {
 	nextIndex := eh.currentIndex + 1
-	if nextIndex >= len(eh.loadedEvents) && eh.hasMoreEvents() { // current page ends and there is more pages
+	// Server can return an empty page so if we need the next event we must keep checking until we either get it
+	// or know we have no more pages to check
+	for nextIndex >= len(eh.loadedEvents) && eh.hasMoreEvents() { // current page ends and there is more pages
```
cretz, line 233: Ug. One might expect such a loop to be in `loadMoreEvents` and `loadMoreEvents` have docs that says "This will not return until it has actually loaded more events or there are no more events". But I guess this way the history iterator is non-blocking.

cretz (approved):

Sushisource (approved):

PR #1152, opened 2023-06-29
Add more mocks

Quinn-With-Two-Ns:
Add more client mocks for consistency


cretz (approved):

mocks/ScheduleClient.go
cretz, line 1: May want some test showing that this works

PR #1154, opened 2023-07-05
Add docs on update

Quinn-With-Two-Ns:
Add some short docs on update and link to our existing docs.

### DIFF
--- a/workflow/doc.go
+++ b/workflow/doc.go
@@ -346,11 +346,11 @@ For more information see our docs at https://docs.temporal.io/dev-guide/go/featu
 
 ## Validate Updates
 
-Note: This is a feature for advanced users for pre-persistence, read-only validation and that other more advanced validation
+Note: This is a feature for advanced users for pre-persistence, read-only validation. Other more advanced validation
 can and should be done in the handler.
 
-Update validators provide a mechanism to perform read-only validation (ie not modify workflow state or schedule any commands). If
-the update validator returns any error the update will fail and not be written into history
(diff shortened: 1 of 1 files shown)


workflow/doc.go
```diff
@@ -325,6 +325,55 @@ workflow also has the option to stop execution by blocking on a signal channel.
...
+	        Total: counter,
+	    }
+	    return result, nil
+	})
+
+For more information see our docs at https://docs.temporal.io/dev-guide/go/features#handle-update
+
+## Validate Updates
+
+Note: This is a feature for advanced users for pre-persistence, read-only validation and that other more advanced validation
```
cretz, line 349:
```suggestion
Note: This is a feature for advanced users for pre-persistence, read-only validation. Other more advanced validation
```

workflow/doc.go
```diff
@@ -325,6 +325,55 @@ workflow also has the option to stop execution by blocking on a signal channel.
...
+can and should be done in the handler.
+
+Update validators provide a mechanism to perform read-only validation (ie not modify workflow state or schedule any commands). If
```
cretz, line 352:
```suggestion
Update validators provide a mechanism to perform read-only validation (i.e. not modify workflow state or schedule any commands). If
```

workflow/doc.go
```diff
@@ -325,6 +325,55 @@ workflow also has the option to stop execution by blocking on a signal channel.
...
+the update validator returns any error the update will fail and not be written into history
```
cretz, line 353:
```suggestion
the update validator returns any error the update will fail and not be written into history.
```

cretz (approved):
Nothing blocking

Issue #1000, opened 2023-01-09
Panic with clear error message when workflow commands are produced in a side-effect/query function


bergundy:
The snippet below looks like valid code but will cause replay issues since the side effect function is not called on replay and thus not generating the StartTimer command.

```go
workflow.SideEffect(ctx, func(ctx workflow.Context) interface{} {
   return workflow.Sleep(ctx, 45 * time.Second)
})
```

cretz:
We should also do this for if commands are produced in a query handler

PR #1159, opened 2023-07-13
Prevent modifying workflow state in read only contexts

Quinn-With-Two-Ns:
Prevent modifying workflow state in read only contexts. Track the current execution state in the `dispatcher`

An alternative approach would be to check the state when the command is being recorded, I didn't like that option because workflow state can be modified without a command being generated (like reading from a channel) so it needs to be more generic.

closes https://github.com/temporalio/sdk-go/issues/1000

### DIFF
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -66,7 +66,7 @@ type (
 		Complete(success interface{}, err error)
 	}
 
-	// UpdateScheduluer allows an update state machine to spawn coroutines and
+	// UpdateScheduler allows an update state machine to spawn coroutines and
 	// yield itself as necessary.
 	UpdateScheduler interface {
 		// Spawn starts a new named coroutine, executing the given function f.
@@ -231,7 +231,7 @@ func (up *updateProtocol) checkAcceptedEvent(e *historypb.HistoryEvent) bool {
 		attrs.AcceptedRequest != nil
 }
 
-// defaultHandler receives the initial invocation of an upate during WFT
+// defaultHandler receives the initial invocation of an update during WFT
 // processing. The implementation will verify that an updateHandler exists for
 // the supplied name (rejecting the update otherwise) and use the provided spawn
 // function to create a new coroutine that will execute in the workflow context.
@@ -253,6 +253,8 @@ func defaultUpdateHandler(
 		return
 	}
 	scheduler.Spawn(ctx, name, func(ctx Context) {
+		defer getState(ctx).dispatcher.setIsReadOnly(false)
+		getState(ctx).dispatcher.setIsReadOnly(true)
 		eo := getWorkflowEnvOptions(ctx)
 
 		// If we suspect that handler registration has not occurred (e.g.
@@ -295,6 +297,7 @@ func defaultUpdateHandler(
 			}
 		}
 		callbacks.Accept()
+		getState(ctx).dispatcher.setIsReadOnly(false)
 		success, err := envInterceptor.inboundInterceptor.ExecuteUpdate(ctx, &input)
 		callbacks.Complete(success, err)
 	})
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -289,15 +289,17 @@ func defaultUpdateHandler(
 		if !IsReplaying(ctx) {
 			// we don't execute update validation during replay so that
 			// validation routines can change across versions
-			defer getState(ctx).dispatcher.setIsReadOnly(false)
-			getState(ctx).dispatcher.setIsReadOnly(true)
-			if err := envInterceptor.inboundInterceptor.ValidateUpdate(ctx, &input); err != nil {
+			err = func() error {
+				defer getState(ctx).dispatcher.setIsReadOnly(false)
+				getState(ctx).dispatcher.setIsReadOnly(true)
+				return envInterceptor.inboundInterceptor.ValidateUpdate(ctx, &input)
+			}()
+			if err != nil {
 				callbacks.Reject(err)
 				return
 			}
 		}
 		callbacks.Accept()
-		getState(ctx).dispatcher.setIsReadOnly(false)
 		success, err := envInterceptor.inboundInterceptor.ExecuteUpdate(ctx, &input)
 		callbacks.Complete(success, err)
 	})
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -253,8 +253,6 @@ func defaultUpdateHandler(
 		return
 	}
 	scheduler.Spawn(ctx, name, func(ctx Context) {
-		defer getState(ctx).dispatcher.setIsReadOnly(false)
-		getState(ctx).dispatcher.setIsReadOnly(true)
 		eo := getWorkflowEnvOptions(ctx)
 
 		// If we suspect that handler registration has not occurred (e.g.
@@ -291,6 +289,8 @@ func defaultUpdateHandler(
 		if !IsReplaying(ctx) {
 			// we don't execute update validation during replay so that
 			// validation routines can change across versions
+			defer getState(ctx).dispatcher.setIsReadOnly(false)
+			getState(ctx).dispatcher.setIsReadOnly(true)
 			if err := envInterceptor.inboundInterceptor.ValidateUpdate(ctx, &input); err != nil {
 				callbacks.Reject(err)
 				return
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -1449,7 +1449,6 @@ func newDecodeFuture(ctx Context, fn interface{}) (Future, Settable) {
 
 // setQueryHandler sets query handler for given queryType.
 func setQueryHandler(ctx Context, queryType string, handler interface{}) error {
-	assertNotInReadOnlyState(ctx)
 	qh := &queryHandler{fn: handler, queryType: queryType, dataConverter: getDataConverterFromWorkflowContext(ctx)}
 	err := validateQueryHandlerFn(qh.fn)
 	if err != nil {
@@ -1462,7 +1461,6 @@ func setQueryHandler(ctx Context, queryType string, handler interface{}) error {
 
 // setUpdateHandler sets update handler for a given update name.
 func setUpdateHandler(ctx Context, updateName string, handler interface{}, opts UpdateHandlerOptions) error {
-	assertNotInReadOnlyState(ctx)
 	uh, err := newUpdateHandler(updateName, handler, opts)
 	if err != nil {
 		return err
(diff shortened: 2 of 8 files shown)


internal/internal_workflow.go
```diff
@@ -149,6 +149,8 @@ type (
 	// Used to truncate internal stack frames from thread stack.
 	unblockFunc func(status string, stackDepth int) (keepBlocked bool)
 
+	ExecutionState int
```
cretz, line 152:
```suggestion
	executionState int
```

I think this and all its values should not be exported from this package

internal/internal_workflow.go
```diff
@@ -646,11 +661,26 @@ func getState(ctx Context) *coroutineState {
 	}
 	state := s.(*coroutineState)
 	if !state.dispatcher.IsExecuting() {
-		panic(panicIllegalAccessCoroutinueState)
+		panic(panicIllegalAccessCoroutineState)
 	}
 	return state
 }
 
+func isReadOnlyState(executionState ExecutionState) bool {
```
cretz, line 669:
```suggestion
func (e ExecutionState) isReadOnlyState() bool {
```

Just reads a bit clearer to me and doesn't pollute the package-level function set, but your choice. Or better yet just move this little logic inside of `assertNotInReadOnlyState` and not have a separate function that is one line and only used in one place.

internal/internal_workflow.go
```diff
@@ -924,6 +955,12 @@ func (s *coroutineState) yield(status string) {
 	s.keptBlocked = true
 }
 
+func (s *coroutineState) setState(state ExecutionState) {
+	// TODO(quinn) check blocked
+	s.state = state
+	s.dispatcher.setExecutionState(s.state)
```
cretz, line 961: In other places you set invoke this shared `setExecutionState` you are unsetting it in defer, but here you are just setting it without unsetting. It's a bit hard for me to follow the code to know if this is safe.

internal/internal_workflow.go
```diff
@@ -1051,6 +1093,7 @@ func (d *dispatcherImpl) ExecuteUntilAllBlocked(deadlockDetectionTimeout time.Du
 	d.mutex.Unlock()
 	defer func() {
 		d.mutex.Lock()
+		d.executionState = ExecutionStateNone
```
cretz, line 1096: Is this thread safe? In other places you are concerned about the thread safety of mutating this value.

internal/internal_workflow.go
```diff
@@ -158,6 +160,7 @@ type (
 		closed       atomic.Bool      // indicates that owning coroutine has finished execution
 		blocked      atomic.Bool
 		panicError   error // non nil if coroutine had unhandled panic
+		state        ExecutionState
```
cretz, line 163: Thinking about this generally, why does a coroutine need a state? Shouldn't the only thing that matters be what state the workflow is in at that time? I could understand if this were for general purpose use, but it's only for readonly.

Maybe, stepping back, we should just have `commandsDisallowed atomic.Bool` on `dispatcherImpl` and keep it simple.

internal/internal_workflow.go
Quinn-With-Two-Ns: Yes this is intentional, we sometimes `defer` unsetting because we are only temporally in a state while executing a side effect

internal/internal_workflow.go
Quinn-With-Two-Ns: I did consider that approach. I did the more complicated approach because it is minimal addition complexity and will be needed in the future to support things like local activities in update validators, that we have said we would like to prioritize.

cretz: I think "local activities in queries/validators" should be solved when we get there, because I think internally the local activities will have to block and not let other workflow code resume like regular local activities might. And I don't think local activities called from queries/validators will produce commands anyways. They are just a way to make external calls. 

Arguably we could just let queries/validators do whatever non-deterministic things they want so long as they don't mutate workflow state, but I suspect we'll call them "local activities" just to make the separation of the concept clear.

So I still think a `commandsDisallowed` simple approach will work all around well.

Quinn-With-Two-Ns: OK I am fine with that approach, I won't call it `commandsDisallowed` because in the Go SDK it is more then just `commandsDisallowed`, I am also blocking things like setting handlers or reading from channels.

cretz: :+1: Yeah, "read only mode" or something

internal/internal_update.go
```diff
@@ -253,6 +253,8 @@ func defaultUpdateHandler(
 		return
 	}
 	scheduler.Spawn(ctx, name, func(ctx Context) {
+		defer getState(ctx).dispatcher.setIsReadOnly(false)
+		getState(ctx).dispatcher.setIsReadOnly(true)
```
cretz, line 257: So I see yield stuff below. Shouldn't you just mark read-only and not read-only strictly around validator function and query function invocation? (sorry if I am commenting on your changes too early)

internal/internal_workflow.go
```diff
@@ -1103,6 +1115,18 @@ func (d *dispatcherImpl) IsExecuting() bool {
...
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	return d.readOnly
+}
+
+func (d *dispatcherImpl) setIsReadOnly(readOnly bool) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	d.readOnly = readOnly
+}
```
cretz, line 1128: Could use `atomic.Bool` if you prefer, it may be simpler

internal/internal_workflow.go
```diff
@@ -646,11 +648,20 @@ func getState(ctx Context) *coroutineState {
 	}
 	state := s.(*coroutineState)
 	if !state.dispatcher.IsExecuting() {
-		panic(panicIllegalAccessCoroutinueState)
+		panic(panicIllegalAccessCoroutineState)
 	}
 	return state
 }
 
+func assertNotInReadOnlyState(ctx Context) {
```
cretz, line 656: Should this be invoked just before command add? Or is there not a reasonable place to do that? (sorry if I'm commenting too early)

internal/internal_update.go
Quinn-With-Two-Ns: yes that's correct

internal/internal_workflow.go
Quinn-With-Two-Ns: Not really a reasonable place, there are also edge cases when functions that can't be called in read only context only sometimes generate commands, or functions that don't generate commands but still shouldn't be called in read only context. I think it is easier to implement, allows us the fail faster, and reads clearer ,because it is more consistent, if every function you shouldn't call has its own assert.

cretz: Right, I can understand adding additional checks for things that don't send commands.

> if every function you shouldn't call has its own assert.

If you don't think it'll be too crazy or hard to maintain, works for me. Put it as close to the outer call boundaries as we can (i.e. right as first line of exported thing, not behind interceptors or after other validation or whatever).

internal/internal_update.go
```diff
@@ -289,12 +289,15 @@ func defaultUpdateHandler(
...
 			// validation routines can change across versions
+			defer getState(ctx).dispatcher.setIsReadOnly(false)
+			getState(ctx).dispatcher.setIsReadOnly(true)
 			if err := envInterceptor.inboundInterceptor.ValidateUpdate(ctx, &input); err != nil {
 				callbacks.Reject(err)
 				return
 			}
 		}
 		callbacks.Accept()
+		getState(ctx).dispatcher.setIsReadOnly(false)
```
cretz, line 300: Is this line still needed?

internal/internal_workflow.go
```diff
@@ -1423,6 +1449,7 @@ func newDecodeFuture(ctx Context, fn interface{}) (Future, Settable) {
 
 // setQueryHandler sets query handler for given queryType.
 func setQueryHandler(ctx Context, queryType string, handler interface{}) error {
+	assertNotInReadOnlyState(ctx)
```
cretz, line 1452: I think all of these calls should be the first line of the exported calls, not after the interceptors. I think this makes it clearer that you're catching them all and safer to make sure it's not doing any validation or other code first.

internal/workflow.go
```diff
@@ -557,6 +565,7 @@ func ExecuteActivity(ctx Context, activity interface{}, args ...interface{}) Fut
 
 func (wc *workflowEnvironmentInterceptor) ExecuteActivity(ctx Context, typeName string, args ...interface{}) Future {
 	// Validate type and its arguments.
+	assertNotInReadOnlyState(ctx)
```
cretz, line 568: Is this line needed? Or is this already taken care of by the exposed call?

test/workflow_test.go
```diff
@@ -860,6 +860,56 @@ func (w *Workflows) LargeQueryResultWorkflow(ctx workflow.Context) (string, erro
 	return "hello", nil
 }
 
+func (w *Workflows) MutatingQueryWorkflow(ctx workflow.Context) (string, error) {
+	err := workflow.SetQueryHandler(ctx, "mutating_query", func() (string, error) {
```
cretz, line 864: Sorry I am just missing it, but where is the `setIsReadOnyl(true)` for query functions?

internal/internal_update.go
Quinn-With-Two-Ns: yes

internal/internal_workflow.go
Quinn-With-Two-Ns: yeah good catch

test/workflow_test.go
Quinn-With-Two-Ns: it's not set because when queries are run the dispatcher isn't running so that will fail before we even check.

internal/internal_update.go
```diff
@@ -289,12 +289,15 @@ func defaultUpdateHandler(
 		if !IsReplaying(ctx) {
 			// we don't execute update validation during replay so that
 			// validation routines can change across versions
+			defer getState(ctx).dispatcher.setIsReadOnly(false)
+			getState(ctx).dispatcher.setIsReadOnly(true)
 			if err := envInterceptor.inboundInterceptor.ValidateUpdate(ctx, &input); err != nil {
 				callbacks.Reject(err)
 				return
 			}
```
cretz, line 297:
```suggestion
			err := func() error {
				defer getState(ctx).dispatcher.setIsReadOnly(false)
				getState(ctx).dispatcher.setIsReadOnly(true)
				return envInterceptor.inboundInterceptor.ValidateUpdate(ctx, &input)
			}()
			if err != nil {
				callbacks.Reject(err)
				return
			}
```

Mentioned in Slack convo. I think this localizes the read-only a bit better

internal/internal_update.go
cretz: Ah, ok, added a comment above now that I understand the situation a bit better

cretz (approved):

PR #1162, opened 2023-07-15
Add support for update info

Quinn-With-Two-Ns:
Add support for update info to allow update handlers and validators to know their update ID.


internal/internal_event_handlers.go
```diff
@@ -153,7 +153,7 @@ type (
 		cancelHandler   func()                                                                     // A cancel handler to be invoked on a cancel notification
 		signalHandler   func(name string, input *commonpb.Payloads, header *commonpb.Header) error // A signal handler to be invoked on a signal event
 		queryHandler    func(queryType string, queryArgs *commonpb.Payloads, header *commonpb.Header) (*commonpb.Payloads, error)
-		updateHandler   func(name string, args *commonpb.Payloads, header *commonpb.Header, callbacks UpdateCallbacks)
+		updateHandler   func(name string, ID string, args *commonpb.Payloads, header *commonpb.Header, callbacks UpdateCallbacks)
```
cretz, line 156:
```suggestion
		updateHandler   func(name string, id string, args *commonpb.Payloads, header *commonpb.Header, callbacks UpdateCallbacks)
```

And elsewhere. Parameters should not be capitalized.

Issue #1160, opened 2023-07-13
TestWorkflowEnvironment.XByID does not work with primary ID after primary ID updated


cretz:
## Expected Behavior

Calling `SetStartWorkflowOptions` with an ID and then `SignalWorkflowByID` with that ID in delayed callback should work.

## Actual Behavior

We set the `runningWorkflows` map to the default ID and don't update the map on `SetStartWorkflowOptions` like we should. So it appears as not found.

PR #1163, opened 2023-07-15
Reassign runningWorkflows workflow ID in test environment

Quinn-With-Two-Ns:
Update the runningWorkflows map with the specified ID if set

closes: https://github.com/temporalio/sdk-go/issues/1160

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -2431,6 +2431,10 @@ func (env *testWorkflowEnvironmentImpl) setStartWorkflowOptions(options StartWor
 		wf.WorkflowTaskTimeout = options.WorkflowTaskTimeout
 	}
 	if len(options.ID) > 0 {
+		// Reassign the ID in running Workflows so SignalWorkflowByID can find the workflow
+		originalID := wf.WorkflowExecution.ID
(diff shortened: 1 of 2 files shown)


internal/internal_workflow_testsuite.go
```diff
@@ -2431,6 +2431,10 @@ func (env *testWorkflowEnvironmentImpl) setStartWorkflowOptions(options StartWor
 		wf.WorkflowTaskTimeout = options.WorkflowTaskTimeout
 	}
 	if len(options.ID) > 0 {
+		// Reassign the ID in running Workflows so SignalWorkflowByID can find the workflow
+		originalID := wf.WorkflowExecution.ID
+		env.runningWorkflows[options.ID] = env.runningWorkflows[wf.WorkflowExecution.ID]
+		delete(env.runningWorkflows, originalID)
```
cretz, line 2437: I was originally thinking maybe we wanted to also leave it in the map under the original ID, but if someone is referring to it by original ID in their tests even after setting w/ a new ID, I think they can be asked to fix their tests. Thoughts?

cretz (approved):

internal/internal_workflow_testsuite.go
Quinn-With-Two-Ns: yeah  I think it is a reasonable to  ask they fix their test

PR #1164, opened 2023-07-16
Add eager workflow start

Quinn-With-Two-Ns:
Add eager workflow start

Note: WIP


internal/client.go
```diff
@@ -816,7 +821,13 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
 	}
 
 	// Create outbound interceptor by wrapping backwards through chain
-	client.interceptor = &workflowClientInterceptor{client: client}
+	client.rootInterceptor = &workflowClientInterceptor{
+		client: client,
+		eagerDispatcher: &eagerWorkflowDispatcher{
+			workersByTaskQueue: make(map[string][]eagerWorker),
+		},
```
cretz, line 828: I think this makes sense on the client itself instead of the client interceptor

internal/internal_eager_workflow.go
```diff
@@ -0,0 +1,94 @@
...
+type eagerWorkflowExecutor struct {
+	handledResponse atomic.Bool
+	worker          eagerWorker
+}
+
+// handleResponse of an eager workflow task from a StartWorkflowExecution request.
+func (e *eagerWorkflowExecutor) handleResponse(response *workflowservice.PollWorkflowTaskQueueResponse) {
+	if !e.handledResponse.CompareAndSwap(false, true) {
+		panic("eagerWorkflowExecutor trying to handle multiple responses")
+	}
```
cretz, line 74: What is the situation you are trying to protect from here?

internal/internal_worker_base.go
```diff
@@ -310,6 +310,34 @@ func (bw *baseWorker) runPoller() {
...
+		return false
+	}
+	// Reserve a executor slot via a non-blocking attempt to take a poller
+	// request entry which essentially reserves a slot
+	select {
+	case <-bw.pollerRequestCh:
+	default:
+		return false
+	}
+	return true
```
cretz, line 324:
```suggestion
	select {
	case <-bw.pollerRequestCh:
		return true
	default:
		return false
	}
```

(just a nit if it reads clearer to you)

internal/internal_workflow_client.go
```diff
@@ -1516,6 +1520,14 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
 		Header:                   header,
 	}
 
+	var eagerExecutor *eagerWorkflowExecutor
+	if in.Options.EnableEagerStart && w.client.capabilities.GetEagerWorkflowStart() && w.eagerDispatcher != nil {
+		eagerExecutor = w.eagerDispatcher.applyToRequest(startRequest)
+		if eagerExecutor != nil {
+			defer eagerExecutor.release()
+		}
+	}
```
cretz, line 1529: I am not wondering if there is value in a separate dispatcher and executor concept since this is so straightforward and done on only this call. I wonder if you can just store a `eagerWorkflowWorkers sync.Map` on `WorkflowClient` and change this to:

```go
var eagerWorkerInUse eagerWorker
if in.Options.EnableEagerStart && w.client.capabilities.GetEagerWorkflowStart() {
	eagerWorkerIface, _ := w.client.eagerWorkflowWorkers.Load(request.GetTaskQueue().GetName())
	eagerWorkerInUse, _ = eagerWorkerIface.(eagerWorker)
	request.RequestEagerExecution = eagerWorkerInUse != nil && eagerWorkerInUse.tryReserveSlot()
}
```

And after response:

```go
if request.RequestEagerExecution {
	if task := response.GetEagerWorkflowTask(); task != nil {
		eagerWorkerInUse.processTaskAsync(&eagerWorkflowTask{task}, eagerWorkerInUse.releaseSlot)
	} else {
		eagerWorkerInUse.releaseSlot()
	}
}
```

internal/internal_eager_workflow.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *eagerWorkflowDispatcher) registerWorker(worker *workflowWorker) {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	e.workersByTaskQueue[worker.executionParameters.TaskQueue] = append(e.workersByTaskQueue[worker.executionParameters.TaskQueue], worker.worker)
+}
+
+// applyToRequest updates request if eager workflow dispatch is possible and returns the eagerWorkflowExecutor to use
+func (e *eagerWorkflowDispatcher) applyToRequest(request *workflowservice.StartWorkflowExecutionRequest) *eagerWorkflowExecutor {
+	e.lock.Lock()
+	defer e.lock.Unlock()
```
cretz, line 49: A `sync.Map` may be a better choice here, and even if not, you should use a `RWMutex` and only `RLock` here, and you should only lock map access, not the slot reservation IMO. (though as mentioned in another comment I think a simple `eagerWorkflowWorkers sync.Map` on the client itself is probably plenty and the dispatcher/executor is unnecessary)

internal/client.go
Quinn-With-Two-Ns: I preferred it in the interceptor because that is where it is used

internal/internal_eager_workflow.go
Quinn-With-Two-Ns: a caller accidentally reusing  eagerWorkflowExecutor

internal/internal_eager_workflow.go
```diff
@@ -0,0 +1,94 @@
...
+		// The processTaskAsync does not do this itself because our task is *eagerWorkflowTask, not *polledTask.
+		e.release()
+	})
+}
+
+// release the executor task slot this eagerWorkflowExecutor was holding.
+// If it is currently handling a responses or has already released the task slot
+// then do nothing.
+func (e *eagerWorkflowExecutor) release() {
+	if e.handledResponse.CompareAndSwap(false, true) {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: easier to always call `release`, gives it a similar interface to things like `cancel`

internal/client.go
cretz: The interceptor often references the object its intercepting (in this case a client)

internal/internal_workflow_client.go
Quinn-With-Two-Ns: I prefer the abstraction of a separate dispatcher and executor

internal/internal_eager_workflow.go
```diff
@@ -0,0 +1,94 @@
...
+	e.workersByTaskQueue[worker.executionParameters.TaskQueue] = append(e.workersByTaskQueue[worker.executionParameters.TaskQueue], worker.worker)
+}
+
+// applyToRequest updates request if eager workflow dispatch is possible and returns the eagerWorkflowExecutor to use
+func (e *eagerWorkflowDispatcher) applyToRequest(request *workflowservice.StartWorkflowExecutionRequest) *eagerWorkflowExecutor {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	// Try every worker that is assigned to the desired task queue.
+	workers := e.workersByTaskQueue[request.GetTaskQueue().Name]
+	rand.Shuffle(len(workers), func(i, j int) { workers[i], workers[j] = workers[j], workers[i] })
```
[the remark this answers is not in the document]
cretz: Ug, that's regrettable

internal/internal_eager_workflow.go
cretz: I think we should be allowed to trust callers on non-exported things, i.e. ourselves, here

internal/internal_eager_workflow.go
Quinn-With-Two-Ns: Don't believe it is safe to only lock the map access, not the slot reservation. Since the map could be modified while iterating. Originally I thought I might need some limits like eager activity dispatch has, so a mutex made more sense, but I don't think they are needed so I'm fine to use a `sync.Map`

internal/internal_eager_workflow.go
Quinn-With-Two-Ns: I don't

internal/internal_eager_workflow.go
cretz: Slot reservation should be per worker and concurrency safe from a caller POV. Only choosing of a worker I think needs to be synchronized.

internal/internal_eager_workflow.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: thats a good point, i'll change it so release is expected to be called once and panic if it is released twice.

internal/internal_eager_workflow.go
Quinn-With-Two-Ns: actually on further review I think a `sync.map` here is a bad choice because we loose any type safety and  insertion would not be safe . So i'll switch to a `RWMutex` mutex

cretz: FWIW, it's ok to lose type safety on `sync.Map` in this case. It's for internal use. Nobody would ever use `sync.Map` if they were subject to such requirements even on non-exported fields.

internal/internal_worker_base.go
```diff
@@ -310,6 +310,34 @@ func (bw *baseWorker) runPoller() {
 	}
 }
 
+func (bw *baseWorker) tryReserveSlot() bool {
+	if !bw.isWorkerStarted || bw.isStop() {
```
cretz, line 314: A bit scared of accessing this field in a non-thread-safe way, but probably ok since only mutated on start/stop

internal/internal_worker_base.go
```diff
@@ -323,15 +351,14 @@ func (bw *baseWorker) runTaskDispatcher() {
...
+			// for non-polled-task (local activity result as task or eager task), we don't need to rate limit
 			_, isPolledTask := task.(*polledTask)
 			if isPolledTask && bw.taskLimiter.Wait(bw.limiterContext) != nil {
 				if bw.isStop() {
 					return
 				}
 			}
-			bw.stopWG.Add(1)
-			go bw.processTask(task)
+			bw.processTaskAsync(task, func() {})
```
cretz, line 361:
```suggestion
			bw.processTaskAsync(task, nil)
```

And handle nil on the `processTaskAsync` side. Don't have to and definitely a probably-not-needed micro-optimization, but this is called for every task.

cretz (approved):
Works for me. Left a couple of optional comments.

internal/internal_worker_base.go
rustatian: @cretz @Quinn-With-Two-Ns Yes, I've found a race here on the `roadrunner-temporal` tests, on the one where we start and then stop temporal workers from the different goroutines, but protected by mutex from our side. Here is the stack trace of the race:
 
image

Quinn-With-Two-Ns: @rustatian can you link the `roadrunner-temporal` test failure? I'll look into it

rustatian: > @rustatian can you link the `roadrunner-temporal` test failure? I'll look into it

https://github.com/roadrunner-server/rr-e2e-tests/actions/runs/5790984816/job/15695006300#step:10:963

PR #1165, opened 2023-07-17
Remove unused info in CI workflow logs

jackdawm:
## What was changed
Removing the ref and head_ref from the workflow logs.

## Why?
Directly using this user-modifiable information without setting it as a variable first is considered an injection vulnerability. However, since it's not really useful, just remove it.

## Checklist

How was this tested:
Just did the static analysis and unit tests.


cretz (approved):
Thanks!

PR #1168, opened 2023-07-21
Unlock before panicking

jackdawm:
## What was changed
Unlock a locked dispatcher mutex before panicking.

## Why?
Since it needs to be unlocked to finish running coroutines, and then relocked long enough to have the status changed to not executing, defer didn't fit.

## Checklist

How was this tested:
make test is currently failing integration tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,6 @@ jobs:
   build-and-test:
     runs-on: ubuntu-latest
     steps:
-      - name: Print build information
-        run: "echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}"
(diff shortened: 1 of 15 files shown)


Quinn-With-Two-Ns:
What is the goal of this change? what bug/issue does this resolve?

cretz (approved):
While I think we sometimes lock mutexes we intentionally do not unlock, I don't think this is one of those cases. It's just really-old pre-Temporal missed code I think (and highly unlikely these conditions are ever satisfied at runtime).

Will let @Quinn-With-Two-Ns review/merge.

Quinn-With-Two-Ns (approved):
I think if these panics ever trigger the dispatcher should not be used again, but in principle yeah we should unlock

PR #1169, opened 2023-07-21
Hardcode noop metrics handler during replay

jlegrone:
## What was changed

During replay, the no-op metrics handler is hardcoded in the worker execution parameters.

## Why?

Previously `MetricsHandler` was always nil. This caused "No metrics handler configured for temporal worker" logs to be emitted during replay, which could get noisy and cause confusion about whether this was due to some sort of misconfiguration.

## Checklist

1. Closes

N/A

2. How was this tested:

Manually verified the log line is removed during replay.

3. Any docs updates needed?

No


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Thanks for the contribution @jlegrone

Issue #648, opened 2021-11-23
Expose local boolean from activity.GetInfo


cretz:
**Is your feature request related to a problem? Please describe.**

Java exposes whether an activity is local, Go does not.

**Describe the solution you'd like**

Expose the field on `activity.Info`

Issue #910, opened 2022-09-15
Add activity.InActivity(context.Context) bool


cretz:
**Is your feature request related to a problem? Please describe.**

Need to know whether the context is an activity context.

PR #1170, opened 2023-07-23
Add IsLocalActivity to ActivityInfo

Quinn-With-Two-Ns:
Expose `IsLocalActivity` in activity info

closes https://github.com/temporalio/sdk-go/issues/648


Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-go/issues/910

Issue #991, opened 2023-01-03
QueryWorkflowWithOptions not going through interceptor


cretz:
## Expected Behavior

The client interceptor for queries should intercept all query calls, not just the non-options form. If this means the with-options form has to check if the error is a query rejected error to convert it to a non-error response, so be it.

## Actual Behavior

Currently the with-options form is called by the interceptor, not the other way around as it should be.

PR #1171, opened 2023-07-23
Make QueryWorkflowWithOptions go through interceptor

Quinn-With-Two-Ns:
Make `QueryWorkflowWithOptions` go through the interceptor. Originally `QueryWorkflowWithOptions` was called by the interceptor which is wrong.

closes https://github.com/temporalio/sdk-go/issues/991

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -194,6 +194,11 @@ type (
 		// func which use a next token to get next page of history events
 		paginate func(nexttoken []byte) (*workflowservice.GetWorkflowExecutionHistoryResponse, error)
 	}
+
+	// queryRejectedError is a wrapper for QueryRejected
+	queryRejectedError struct {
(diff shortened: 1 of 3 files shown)


internal/internal_workflow_client.go
```diff
@@ -194,6 +194,11 @@ type (
 		// func which use a next token to get next page of history events
 		paginate func(nexttoken []byte) (*workflowservice.GetWorkflowExecutionHistoryResponse, error)
 	}
+
+	// queryRejectedError is a wrapper for QueryRejected
+	queryRejectedError struct {
```
cretz, line 199: Was considering suggesting making this public, but since it's only for sending this across interceptor-to-client boundary because we can't change the interceptor signature nor the client signature, I'm ok w/ it being hidden. If a user ever wants to intercept query rejection, we can discuss making this available.

cretz (approved):

PR #1175, opened 2023-07-26
[WIP] Clear workflowExecutionIsCancelling on new WFT

Quinn-With-Two-Ns:
Clear `workflowExecutionIsCancelling` on new WFT. 
 
 As far as I can tell `workflowExecutionIsCancelling` is only needed because on workflow cancelation we generate the cancel commands eagerly before the WFT started and when we process the WFT started event we reset the eventID the Go SDK uses for counting so this throws off everything in the SDK. The problem with the current implementation is it assumes all cancels after a workflow cancel request are a result of the cancel request and that is not true. 
 
 I believe after the cancel request callback is processed this whole `workflowExecutionIsCancelling` logic should not be needed since all the out of workflow execution callbacks should be done, but I could be wrong on this point. 
 
 I think the best solution would be to remove this `workflowExecutionIsCancelling` and generate the cancel commands after setting the eventID on WFT started, but there is a 2+ year history of cancellation issues around this so It is very possible I am missing something.
 
 Summary of the history of this flag:
https://github.com/temporalio/sdk-go/pull/726

Fix for https://github.com/temporalio/sdk-go/issues/1176

### DIFF
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -912,6 +912,8 @@ func (h *commandsHelper) setCurrentWorkflowTaskStartedEventID(workflowTaskStarte
 	// We must change the counter here so that others who mutate
 	// commandsCancelledDuringWFCancellation know it has since been reset
 	h.nextCommandEventIDResetCounter++
+	// Once we have handled the cancellation, we can reset the flag
+	h.workflowExecutionIsCancelling = false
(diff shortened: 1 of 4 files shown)


Quinn-With-Two-Ns:
Closing in favor of https://github.com/temporalio/sdk-go/pull/1181

PR #1178, opened 2023-07-27
Add option to ignore invalid parent spans

### LINKED ISSUE #1174: Allow ignoring invalid SpanContext in OpenTelemetry TracingInterceptor
We recently moved from OpenTracing to OpenTelemetry. Upon deploying, we found a long running workflow failed because it was trying to start a span from an OpenTracing parent span, so an error was returned here: https://github.com/temporalio/sdk-go/blob/master/contrib/opentelemetry/tracing_interceptor.go#L137

Failures to trace should have the option to do so silently. Adding an option to the TracingInterceptor to ignore the invalid SpanContext would allow us to keep this long running workflow alive when the new version of the worker deploys. It's acceptable for spans to not connect to their parents for a short time; it is not acceptable for a critical workflow to fail because of a failure to trace. 

The DataDog interceptor seems to handle this gracefully already. OpenTelemetry does not return any error when getting a SpanContext. 

**Additional context**
I've added a commit [on my fork](https://github.com/cdavis-joy/sdk-go/commit/9fc81c3847d4b19d126e9dfc040b5859f3970773) that attempts to address this. I feel like this is likely a common thing that people might run into; running workflows that straddle two tracing implementations.

cdavis-joy:
## What was changed

Adds an option to ignore errors when interpreting parent spans from headers.

## Why?

When migrating from one tracing library to another, the new library will not find the parent span headers it expects, and will throw an errors, thus failing any workflows/activities that were running at the time of the worker deploy. 

## Checklist

1. Closes #1174 

2. How was this tested:

Forced an error to return from t.readSpanFromHeader, and saw that error was not returned.

3. Any docs updates needed?

Comment added to option.


Quinn-With-Two-Ns (approved):
LGTM, will let @cretz take a look

cretz (approved):
@cdavis-joy - I think the better approach to supporting multiple tracer interceptors is to set the `SpanContextKey` and `HeaderKey` different for each. Then multiple tracing interceptors can be provided until you're ready to remove one.

But since this is opt-in, this addition is harmless. Will leave open to hear your thoughts on just setting different keys, then will merge.

cdavis-joy:
Missed adding this to the otel interceptors `Options` func. Updated.

cdavis-joy:
@cretz I misunderstood what you initially said. By changing the `HeaderKey` and `SpanContextKey`, the newer interceptor would not even attempt to interpret spans from the old interceptor, thus avoiding these errors. Thanks for that, I think we're gonna give that a shot.

cretz:
Let me do a re-run and see. If passes, we'll merge.

PR #1179, opened 2023-07-27
Expand WF context locking to cover WFT responses

mmcshane:
## What was changed

Lifted the workflow context lock out of WFTaskHandler.ProcessWorkflowTask up into the surrounding Poller. Having the lock managed at this higher level allows us to include the RespondWorkflowTaskCompleted call and the (potential) reset of the expected next event ID into the same atomic block.

## Why?

Failure to hold this lock while responding to a workflow task allows in-flight tasks to race past each other which has led to history corruption in the case where responses are not deduplicated. Furthermore the correctness of resetting the event level while not holding this lock is unclear at best.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_public.go
+++ b/internal/internal_public.go
@@ -78,6 +78,8 @@ type (
 
 	// WorkflowTaskHandler represents workflow task handlers.
 	WorkflowTaskHandler interface {
+		WorkflowContextManager
+
 		// Processes the workflow task
 		// The response could be:
 		// - RespondWorkflowTaskCompletedRequest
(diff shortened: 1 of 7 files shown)


mmcshane:
Still missing is the unit test to verify that a successor task does not begin processing until after the current WFT has been responded (and possibly reset). But tests are passing now so it's not a waste of time to review this.

mmcshane:
This is busted. Allowing the Context to remove itself from cache before it's unlocked means that it doesn't work as a mutex anymore. I'll rework.

PR #1180, opened 2023-07-28
Fix LA start to close timeout

Quinn-With-Two-Ns:
This PR fixes a few issues with local activities in Got

- `ScheduleToCloseTimeout` was overriding `StartToCloseTimeout` if both were set
- The docs in `LocalActivityOptions` were not accurate
- If a local activity did timeout it would always be considered a `SCHEDULE_TO_CLOSE` timeout and that is not retry able
- `ActivityInfo` in local activities was missing some fields

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -155,11 +155,13 @@ type (
 	// LocalActivityOptions stores local activity specific parameters that will be stored inside of a context.
 	LocalActivityOptions struct {
 		// ScheduleToCloseTimeout - The end to end timeout for the local activity including retries.
-		// This field is required.
+		// At least on of ScheduleToCloseTimeout or StartToCloseTimeout is required.
+		// defaults to StartToCloseTimeout if not set.
 		ScheduleToCloseTimeout time.Duration
 
 		// StartToCloseTimeout - The timeout for a single execution of the local activity.
-		// Optional: defaults to ScheduleToClose
+		// At least on of ScheduleToCloseTimeout or StartToCloseTimeout is required.
+		// defaults to ScheduleToCloseTimeout if not set.
 		StartToCloseTimeout time.Duration
 
 		// RetryPolicy specify how to retry activity if error happens.
@@ -253,25 +255,12 @@ func WithActivityTask(
 	contextPropagators []ContextPropagator,
 	interceptors []WorkerInterceptor,
 ) (context.Context, error) {
-	var deadline time.Time
 	scheduled := common.TimeValue(task.GetScheduledTime())
 	started := common.TimeValue(task.GetStartedTime())
 	scheduleToCloseTimeout := common.DurationValue(task.GetScheduleToCloseTimeout())
 	startToCloseTimeout := common.DurationValue(task.GetStartToCloseTimeout())
 	heartbeatTimeout := common.DurationValue(task.GetHeartbeatTimeout())
-
-	startToCloseDeadline := started.Add(startToCloseTimeout)
-	if scheduleToCloseTimeout > 0 {
-		scheduleToCloseDeadline := scheduled.Add(scheduleToCloseTimeout)
-		// Minimum of the two deadlines.
-		if scheduleToCloseDeadline.Before(startToCloseDeadline) {
-			deadline = scheduleToCloseDeadline
-		} else {
-			deadline = startToCloseDeadline
-		}
-	} else {
-		deadline = startToCloseDeadline
-	}
+	deadline := calculateActivityDeadline(scheduled, started, scheduleToCloseTimeout, startToCloseTimeout)
 
 	logger = log.With(logger,
 		tagActivityID, task.ActivityId,
@@ -332,6 +321,21 @@ func WithLocalActivityTask(
 		tagWorkflowID, task.params.WorkflowInfo.WorkflowExecution.ID,
 		tagRunID, task.params.WorkflowInfo.WorkflowExecution.RunID,
 	)
+	startedTime := time.Now()
+	scheduleToCloseTimeout := task.params.ScheduleToCloseTimeout
+	startToCloseTimeout := task.params.StartToCloseTimeout
+
+	if startToCloseTimeout == 0 {
+		startToCloseTimeout = scheduleToCloseTimeout
+	}
+	if scheduleToCloseTimeout == 0 {
+		scheduleToCloseTimeout = startToCloseTimeout
+	}
+	deadline := calculateActivityDeadline(task.scheduledTime, startedTime, scheduleToCloseTimeout, startToCloseTimeout)
+	if task.attempt > 1 && !task.expireTime.IsZero() && task.expireTime.Before(deadline) {
+		// this is attempt and expire time is before SCHEDULE_TO_CLOSE timeout
+		deadline = task.expireTime
+	}
 	return newActivityContext(ctx, interceptors, &activityEnvironment{
 		workflowType:      &workflowTypeLocal,
 		workflowNamespace: task.params.WorkflowInfo.Namespace,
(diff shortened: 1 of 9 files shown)


internal/activity.go
```diff
@@ -333,6 +322,21 @@ func WithLocalActivityTask(
...
+	startToCloseTimeout := task.params.StartToCloseTimeout
+
+	if startToCloseTimeout == 0 {
+		startToCloseTimeout = scheduleToCloseTimeout
+	}
+	if scheduleToCloseTimeout == 0 {
+		scheduleToCloseTimeout = startToCloseTimeout
+	}
+	deadline := calculateActivityDeadline(task.scheduledTime, startedTime, scheduleToCloseTimeout, startToCloseTimeout)
+	if task.attempt > 1 && !task.expireTime.IsZero() && task.expireTime.Before(deadline) {
```
Quinn-With-Two-Ns, line 336: This is still an open issue if we should removed this/can remove this. This `expireTime` is calculated based on the `ScheduleToCloseTimeout` and the workflow time, but is only used on the second attempt. This has been basically unchanged for 4+ years, but seems wrong.

internal/activity.go
```diff
@@ -156,11 +156,13 @@ type (
 	// LocalActivityOptions stores local activity specific parameters that will be stored inside of a context.
 	LocalActivityOptions struct {
 		// ScheduleToCloseTimeout - The end to end timeout for the local activity including retries.
-		// This field is required.
+		// At least on of ScheduleToCloseTimeout or StartToCloseTimeout is required.
```
cretz, line 159:
```suggestion
		// At least one of ScheduleToCloseTimeout or StartToCloseTimeout is required.
```

internal/activity.go
```diff
@@ -156,11 +156,13 @@ type (
...
+		// defaults to StartToCloseTimeout if not set.
 		ScheduleToCloseTimeout time.Duration
 
 		// StartToCloseTimeout - The timeout for a single execution of the local activity.
-		// Optional: defaults to ScheduleToClose
+		// At least on of ScheduleToCloseTimeout or StartToCloseTimeout is required.
```
cretz, line 164:
```suggestion
		// At least one of ScheduleToCloseTimeout or StartToCloseTimeout is required.
```

internal/activity.go
```diff
@@ -375,3 +382,18 @@ func newActivityContext(
...
+		scheduleToCloseDeadline := scheduled.Add(scheduleToCloseTimeout)
+		// Minimum of the two deadlines.
+		if scheduleToCloseDeadline.Before(startToCloseDeadline) {
+			return scheduleToCloseDeadline
+		} else {
+			return startToCloseDeadline
+		}
+	} else {
+		return startToCloseDeadline
+	}
```
cretz, line 398:
```suggestion
		if scheduleToCloseDeadline.Before(startToCloseDeadline) {
			return scheduleToCloseDeadline
		}
	}
	return startToCloseDeadline
```

Just reads clearer to me personally

internal/internal_event_handlers.go
```diff
@@ -172,19 +172,21 @@ type (
...
+		wc                          *workflowExecutionContextImpl
+		canceled                    bool
+		cancelFunc                  func()
+		attempt                     int32  // attempt starting from 1
+		attemptsThisWFT             uint32 // Number of attempts started during this workflow task
+		pastFirstWFT                bool   // Set true once this LA has lived for more than one workflow task
+		retryPolicy                 *RetryPolicy
+		expireTime                  time.Time
+		scheduledTime               time.Time // Time the activity was scheduled initially.
+		currentAttemptScheduledTime time.Time // Time this attempt of the activity was scheduled.
```
cretz, line 188: This field only ever looks set, never read. Am I missing some code or context?

internal/internal_task_pollers.go
```diff
@@ -564,18 +564,8 @@ func (lath *localActivityTaskHandler) executeLocalActivityTask(task *localActivi
...
-	timeout := task.params.ScheduleToCloseTimeout
-	if task.params.StartToCloseTimeout != 0 && task.params.StartToCloseTimeout < timeout {
-		timeout = task.params.StartToCloseTimeout
-	}
-	timeoutDuration := timeout
-	deadline := time.Now().Add(timeoutDuration)
-	if task.attempt > 1 && !task.expireTime.IsZero() && task.expireTime.Before(deadline) {
-		// this is attempt and expire time is before SCHEDULE_TO_CLOSE timeout
-		deadline = task.expireTime
-	}
```
cretz, line 576: I don't see anything obvious, but are there any ways that you can think of that this logic has changed with your new logic?

internal/internal_task_pollers.go
```diff
@@ -644,7 +635,11 @@ WaitResult:
 			metricsHandler.Counter(metrics.LocalActivityExecutionCanceledCounter).Inc(1)
 			return &localActivityResult{err: ErrCanceled, task: task}
 		} else if ctx.Err() == context.DeadlineExceeded {
-			return &localActivityResult{err: ErrDeadlineExceeded, task: task}
+			if task.params.ScheduleToCloseTimeout != 0 && time.Now().After(info.scheduledTime.Add(task.params.ScheduleToCloseTimeout)) {
```
cretz, line 638: I wish there was a better way to know which timeout was hit instead of having to do math here.

internal/internal_task_pollers.go
Quinn-With-Two-Ns: The only change here is using the scheduled time + `ScheduleToCloseTimeout` when calculating the deadline (used to be `time.Now`). It probably won't effect many activities because for the  first attempt `scheduled time` will be close to `time.Now` and subsequent attempts `task.expireTime` uses workflow time so it will always be shorter than either deadline

internal/internal_event_handlers.go
Quinn-With-Two-Ns: yeah good catch i'll remove. I had it in to implement `ScheduleToClose` timeout, but left that out of this  PR

internal/activity.go
```diff
@@ -156,11 +156,13 @@ type (
 	// LocalActivityOptions stores local activity specific parameters that will be stored inside of a context.
 	LocalActivityOptions struct {
 		// ScheduleToCloseTimeout - The end to end timeout for the local activity including retries.
-		// This field is required.
+		// At least on of ScheduleToCloseTimeout or StartToCloseTimeout is required.
+		// defaults to StartToCloseTimeout if not set.
```
Quinn-With-Two-Ns, line 160: The default for `ScheduleToCloseTimeout ` for normal activities is infinity. We could be consistent here, but that might be a breaking change.

internal/internal_task_pollers.go
cretz: So if today I have a schedule to close timeout of 5s, my second attempt would get the full 5s (even though that's improper)? But after this change, it will only get however much time since the beginning of the first attempt? If so, I may hit you up off-PR to communicate with one or two known heavy local activity users to confirm this is acceptable.

Quinn-With-Two-Ns: >So if today I have a schedule to close timeout of 5s, my second attempt 

Today you would have no second attempt because  schedule to close was not retried

cretz: Confirmed off-PR that second-attempt after first-attempt-immediate-failure is bounded today by `expireTime` regardless.

internal/activity.go
cretz: While wrong, let's leave for now

Quinn-With-Two-Ns: Yeah I agree, just wanted to call this out since  when we put Go  on Core this will be something we have  to deal with.

cretz (approved):

Issue #724, opened 2022-02-10
Activity cancellation causes history error


cretz:
## Expected Behavior

Should not have internal state errors

## Actual Behavior

Has internal state errors, specifically something like:

    BadRequestCancelActivityAttributes: invalid history builder state for action: add-activitytask-cancel-requested-event, ScheduledID: 99

## Steps to Reproduce the Problem

<details>
<summary>main.go</summary>

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os/exec"
	"runtime/debug"
	"strings"
	"time"

	"github.com/google/uuid"
	"go.temporal.io/sdk/client"
	"go.temporal.io/sdk/temporal"
	"go.temporal.io/sdk/worker"
	"go.temporal.io/sdk/workflow"
)

func main() {
	if err := run(); err != nil {
		log.Fatal(err)
	}
}

const useTemporalite = true
const taskQueue = "my-task-queue"

func run() error {
	var clientOpts client.Options

	// Create a logger that will break on the error we've been waiting for
	errSeenCh := make(chan struct{})
	clientOpts.Logger = &logger{
		excludeLevels: map[string]bool{"DEBUG": true},
		onMessageContains: map[string]func(level, msg string){
			"BadRequestCancelActivityAttributes": func(level, msg string) {
				debug.PrintStack()
				close(errSeenCh)
			},
		},
	}

	// Use temporalite if desired
	if useTemporalite {
		clientOpts.Namespace = "namespace-" + uuid.NewString()
		clientOpts.HostPort = "localhost:9233"
		cmd := exec.Command("temporalite", "start", "--ephemeral", "--namespace", clientOpts.Namespace, "--port", "9233")
		// cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr
		if err := cmd.Start(); err != nil {
			return fmt.Errorf("failed starting temporalite: %w", err)
		}
		defer cmd.Process.Kill()
		time.Sleep(200 * time.Millisecond)
	}

	// Create client
	c, err := client.NewClient(clientOpts)
	if err != nil {
		return fmt.Errorf("failed creating client: %w", err)
	}
	defer c.Close()

	// Start workers
	const workerCount = 10
	log.Printf("Starting %v workers", workerCount)
	for i := 0; i < workerCount; i++ {
		w, err := startWorker(c)
		if err != nil {
			return fmt.Errorf("failed starting worker: %w", err)
		}
		defer w.Stop()
	}
[...]

PR #1181, opened 2023-07-28
Rework cancellation counting

Quinn-With-Two-Ns:
# Summary

The Go SDK notoriously tries to count commands and this has lead to all sorts of issues in the Go SDK with the counter getting out of sync. At some point we introduced `workflowExecutionIsCancelling`  to try to correct some counting issues when the workflow execution was cancelled. The reason workflow cancellation was messing with the counting is because when the workflow was cancelled and the cancel event was processed by the WFT state machine all the state-machines that were waiting on that context were generating their respective cancellation command. All these cancel commands were being generated before the Go SDK had processed the WFT started event for this WFT. This is problematic because the WFT started event sets the counter to `workflowTaskStartedEventID +2` so all those command went uncounted. The solution was to count these cancelled command separately and add them to the total. 

The Go SDK approach to cancellation also causes issues when the workflow is cancelled in the same WFT as other events that may resolve other state machines like `TimerCompleted` since the cancellation happens before those event are processed and state machines are resolved. 

Even with the 2+ years of work on this there are still issues with how this correction is implemented https://github.com/temporalio/sdk-go/issues/1176.

I explored changing how cancellation is processed so it is delayed until the WFT state machine processes the WFT started event. In the Go SDK the only safe place to generate commands is after a WFT started event is processed. Unfortunately that is a breaking change because if the workflow is cancelled in the same WFT as a event that resolves a state machine is sent the order they are processed matter.

Since we can't change how cancellation is processed I think the next best solution is base the correction on the unsent commands generated before the WFT started event was processed, rather then keep a separate counter.

This should not be a breaking change to any workflow that worked before this change, but I welcome any new tests to prove this

There is a more history here:
https://github.com/temporalio/sdk-go/pull/726 

There was also a test in this [issue](https://github.com/temporalio/sdk-go/issues/724) (that we couldn't add due to CPU usage?) that I ran as well


internal/internal_command_state_machine.go
```diff
@@ -905,13 +874,21 @@ func (h *commandsHelper) setCurrentWorkflowTaskStartedEventID(workflowTaskStarte
...
-	h.nextCommandEventIDResetCounter++
+	// command. It is possible, notably during workflow cancellation, that commands are generated before the workflow
+	// task started event is processed. In this case we need to adjust the nextCommandEventID to account for these unsent
+	// commands.git
+	var uncountedCommands int64
+	for curr := h.orderedCommands.Front(); curr != nil; {
+		d := curr.Value.(commandStateMachine)
+		command := d.getCommand()
+		if command != nil {
+			fmt.Printf("command: %v \n", command)
```
Quinn-With-Two-Ns, line 885: Ah debug print!

Sushisource (approved):
This makes sense to me and is definitely cleaner than before

test/workflow_test.go
```diff
@@ -1571,6 +1571,59 @@ func (w *Workflows) CronWorkflow(ctx workflow.Context) (int, error) {
...
+	defer func() {
+		if !errors.Is(ctx.Err(), workflow.ErrCanceled) {
+			return
+		}
+		// When the Workflow is canceled, it has to get a new disconnected context to execute any Activities
+		newCtx, _ := workflow.NewDisconnectedContext(ctx)
+		err = workflow.ExecuteActivity(newCtx, activities.EmptyActivity).Get(newCtx, nil)
+	}()
+
+	//Create selector
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: yep i'll remove, this whole test was taken from the bug report .  Must have missed this comment

internal/internal_command_state_machine.go
```diff
@@ -905,13 +874,21 @@ func (h *commandsHelper) setCurrentWorkflowTaskStartedEventID(workflowTaskStarte
...
-	// We must change the counter here so that others who mutate
-	// commandsCancelledDuringWFCancellation know it has since been reset
-	h.nextCommandEventIDResetCounter++
+	// command. It is possible, notably during workflow cancellation, that commands are generated before the workflow
+	// task started event is processed. In this case we need to adjust the nextCommandEventID to account for these unsent
+	// commands.git
+	var uncountedCommands int64
+	for curr := h.orderedCommands.Front(); curr != nil; {
+		d := curr.Value.(commandStateMachine)
+		command := d.getCommand()
```
cretz, line 883: I wish this call were `createCommand()`. It was unclear to me what was happening here until I understood that's what this is.

cretz (commented):
If all the tests pass, I support any fixing of counters in any way. Will hold off approval until debug print is removed.

Quinn-With-Two-Ns:
@cretz done

cretz (approved):

PR #1182, opened 2023-07-29
Expand WF context locking to cover WFT responses

mmcshane:
## What was changed

Lifted the workflow context lock out of WFTaskHandler.ProcessWorkflowTask up into the surrounding Poller. Having the lock managed at this higher level allows us to include the RespondWorkflowTaskCompleted call and the (potential) reset of the expected next event ID into the same atomic block.

## Why?

Failure to hold this lock while responding to a workflow task allows in-flight tasks to race past each other which has led to history corruption in the case where responses are not deduplicated. Furthermore the correctness of resetting the event level while not holding this lock is unclear at best.

## Checklist

1. Closes 

2. How was this tested:

New unit test here plus some time in the bench-go rig

3. Any docs updates needed?

### DIFF
--- a/internal/internal_public.go
+++ b/internal/internal_public.go
@@ -78,6 +78,8 @@ type (
 
 	// WorkflowTaskHandler represents workflow task handlers.
 	WorkflowTaskHandler interface {
+		WorkflowContextManager
+
 		// Processes the workflow task
 		// The response could be:
 		// - RespondWorkflowTaskCompletedRequest
@@ -85,8 +87,21 @@ type (
 		// - RespondQueryTaskCompletedRequest
 		ProcessWorkflowTask(
 			task *workflowTask,
+			ctx *workflowExecutionContextImpl,
 			f workflowTaskHeartbeatFunc,
-		) (response interface{}, resetter EventLevelResetter, err error)
+		) (response interface{}, err error)
+	}
+
+	WorkflowContextManager interface {
+		// GetOrCreateWorkflowContext finds an existing cached context object
+		// for the provided task's run ID or creates a new object, adds it to
+		// cache, and returns it. In all non-error cases the returned context
+		// object is in a locked state (i.e.
+		// workflowExecutionContextImpl.Lock() has been called).
+		GetOrCreateWorkflowContext(
+			task *workflowservice.PollWorkflowTaskQueueResponse,
+			historyIterator HistoryIterator,
+		) (*workflowExecutionContextImpl, error)
 	}
 
 	// ActivityTaskHandler represents activity task handlers.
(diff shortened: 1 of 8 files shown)


internal/internal_task_pollers.go
```diff
@@ -265,6 +266,7 @@ func (bp *basePoller) getCapabilities() *workflowservice.GetSystemInfoResponse_C
 // newWorkflowTaskPoller creates a new workflow task poller which must have a one to one relationship to workflow worker
 func newWorkflowTaskPoller(
 	taskHandler WorkflowTaskHandler,
+	contextManager WorkflowContextManager,
```
cretz, line 269: Doesn't the `taskHandler` imply the `contextManager`? And if so, do you really need a separate interface and field here?

internal/internal_public.go
```diff
@@ -78,15 +78,30 @@ type (
...
 		// - RespondQueryTaskCompletedRequest
 		ProcessWorkflowTask(
 			task *workflowTask,
+			ctx *workflowExecutionContextImpl,
 			f workflowTaskHeartbeatFunc,
-		) (response interface{}, resetter EventLevelResetter, err error)
+		) (response interface{}, err error)
+	}
+
+	WorkflowContextManager interface {
```
cretz, line 95: Given my comment later, any benefit to this interface instead of just putting this one method on the `WorkflowTaskHandler` interface directly?

internal/internal_public.go
```diff
@@ -78,15 +78,30 @@ type (
...
+		// GetOrCreateWorkflowContext finds an existing cached context object
+		// for the provided task's run ID or creates a new object, adds it to
+		// cache, and returns it. In all non-error cases the returned context
+		// object is in a locked state (i.e.
+		// workflowExecutionContextImpl.Lock() has been called).
+		GetOrCreateWorkflowContext(
```
cretz, line 101: Any benefit to exporting this call outside of this package? I guess the general question is "why is `WorkflowTaskHandler` or most things in this package exposed?" Not that it matters that much I guess, just strange.

cretz (approved):
LGTM and makes sense and I can't think of any harm this would cause. Would like @Quinn-With-Two-Ns to look.

Quinn-With-Two-Ns (approved):

internal/internal_task_pollers.go
mmcshane: You're correct. What we ran into here is the (arguable) fact that ContextManager should probably be a concept independent from the TaskHandler - thus the two interfaces. But the reality is that context management is currently highly coupled to this one specific workflowTaskHandlerImpl - the context objects call back into that task handler to access the cache (direct field access no less) among other things.

Actual separation of the implementations was just going to be way too much churn for this PR.

PR #1187, opened 2023-08-05
Allow setting activity timeouts in testsuite

### LINKED ISSUE #1167: Allow setting activity timeouts in TestActivityEnvironment
Currently, the TestActivityEnvironment has a hard-coded 10 minute timeout. If I want to use the test suite to run an activity which takes longer than that, it fails with a timeout error.

Adding the ability to run the activity with options like timeout values in the test environment would help a lot in testing and debugging long activities.

Thanks!

ohad83:
## What was changed

Added 2 method to TestActivityEnvironment:
- SetActivityStartToCloseTimeout
- SetActivityScheduleToCloseTimeout

This will allow changing the default 10 minute timeouts in the current test environments.

## Why?

There are cases where we would like to test long running activities, or test activity timeouts.
The current test suite did not support that, as it had a hardcoded 10 minute timeout.

## Checklist

1. Closes #1167 

2. How was this tested:

Set the StartToClose timeout to a specific duration, and test that:
- An activity which takes longer than that duration returns an error.
- An activity which takes less than that duration finishes successfully.

Same test for ScheduleToClose, I did not see a difference between them in case of the testsuite.

3. Any docs updates needed?
Comment added to the new method, I don't believe further updates are needed.

### DIFF
--- a/internal/internal_workflow_testsuite_test.go
+++ b/internal/internal_workflow_testsuite_test.go
@@ -4031,3 +4031,49 @@ func (s *WorkflowTestSuiteUnitTest) Test_WorkflowGetCurrentHistoryLength() {
 	s.NoError(env.GetWorkflowResult(&result))
 	s.Equal(17, result)
 }
+
+func (s *WorkflowTestSuiteUnitTest) Test_ActivityWithStartToCloseTimeout() {
+	timeout := 100 * time.Millisecond
+
+	timeoutActivity := func(ctx context.Context) error {
+		time.Sleep(timeout * 2)
+		return nil
+	}
+
+	noTimeoutActivity := func(ctx context.Context) error {
+		time.Sleep(timeout / 2)
+		return nil
+	}
+
+	env := s.NewTestActivityEnvironment()
+	env.SetActivityStartToCloseTimeout(timeout)
+	env.RegisterActivity(timeoutActivity)
+	env.RegisterActivity(noTimeoutActivity)
+	_, err := env.ExecuteActivity(timeoutActivity)
+	s.Error(err)
+	_, err = env.ExecuteActivity(noTimeoutActivity)
+	s.NoError(err)
+}
+
+func (s *WorkflowTestSuiteUnitTest) Test_ActivityWithScheduleToCloseTimeout() {
+	timeout := 100 * time.Millisecond
+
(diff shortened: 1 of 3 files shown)


internal/workflow_testsuite.go
```diff
@@ -262,6 +262,16 @@ func (t *TestActivityEnvironment) SetOnActivityHeartbeatListener(
...
 
+// SetActivityStartToCloseTimeout sets the start to close timeouts of activities in the test environment.
+func (e *TestActivityEnvironment) SetActivityStartToCloseTimeout(timeout time.Duration) {
+	e.impl.testWorkflowEnvironmentShared.activityStartToCloseTimeout = timeout
+}
+
+// SetActivityStartToCloseTimeout sets the schedule to close timeouts of activities in the test environment
+func (e *TestActivityEnvironment) SetActivityScheduleToCloseTimeout(timeout time.Duration) {
+	e.impl.testWorkflowEnvironmentShared.activityScheduleToCloseTimeout = timeout
+}
```
cretz, line 273: Hrmm, longer than 10-minutes using this framework is a bit rough. You might consider a proper integration test with a real server instead.

Regardless, I wonder if we should document on these what the default is and document any relation to `SetTestTimeout` here and on that method.

internal/internal_workflow_testsuite_test.go
```diff
@@ -4031,3 +4031,49 @@ func (s *WorkflowTestSuiteUnitTest) Test_WorkflowGetCurrentHistoryLength() {
...
+
+func (s *WorkflowTestSuiteUnitTest) Test_ActivityWithStartToCloseTimeout() {
+	timeout := 100 * time.Millisecond
+
+	timeoutActivity := func(ctx context.Context) error {
+		time.Sleep(timeout * 2)
+		return nil
+	}
+
+	noTimeoutActivity := func(ctx context.Context) error {
```
cretz, line 4043: The low timing resolution/granularity of this test will lead to flakiness on an overloaded worker. I think you can just remove the no-timeout assertion part.

internal/workflow_testsuite.go
Quinn-With-Two-Ns: Yeah we should include a comment here on what the default values are

PR #1188, opened 2023-08-06
fixing spelling errors in comments

tsolodov:
## What was changed
Fixed spelling in code comments

## Why?

Fix spelling errors

## Checklist
N/A

1. Closes 
N/A

2. How was this tested:
N/A

3. Any docs updates needed?
N/A

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -172,7 +172,7 @@ type (
 		GetWorkflowHistory(ctx context.Context, workflowID string, runID string, isLongPoll bool, filterType enumspb.HistoryEventFilterType) HistoryEventIterator
 
 		// CompleteActivity reports activity completed.
-		// activity Execute method can return activity.activity.ErrResultPending to
+		// activity Execute method can return activity.ErrResultPending to
 		// indicate the activity is not completed when it's Execute method returns. In that case, this CompleteActivity() method
 		// should be called when that activity is completed with the actual result and error. If err is nil, activity task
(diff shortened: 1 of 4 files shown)


internal/activity.go
```diff
@@ -224,7 +224,7 @@ func GetWorkerStopChannel(ctx context.Context) <-chan struct{} {
 //
 //	TODO: we don't have a way to distinguish between the two cases when context is canceled because
 //	context doesn't support overriding value of ctx.Error.
-//	TODO: Implement automatic heartbeating with cancellation through ctx.
```
Quinn-With-Two-Ns, line 227: Don't believe this is a spelling mistake

internal/client.go
```diff
@@ -172,7 +172,7 @@ type (
 		GetWorkflowHistory(ctx context.Context, workflowID string, runID string, isLongPoll bool, filterType enumspb.HistoryEventFilterType) HistoryEventIterator
 
 		// CompleteActivity reports activity completed.
-		// activity Execute method can return acitivity.activity.ErrResultPending to
```
Quinn-With-Two-Ns, line 175: II think the real mistake here is `acitivity.activity` should just be `activity` , it should  probably be `[activity.ErrResultPending]`

tsolodov: let me double-check

tsolodov: indeed, fixed it

Quinn-With-Two-Ns: Thanks!

Quinn-With-Two-Ns (approved):

PR #1191, opened 2023-08-08
Fix race condition with eager workflow start and close

Quinn-With-Two-Ns:
Fix race condition in eager WF start reported by `roadrunner`

https://github.com/roadrunner-server/rr-e2e-tests/actions/runs/5790984816/job/15695006300#step:10:963

removed checking `isWorkerStarted` because  `pollerRequestCh` will not have any slots if the worker has not started, also moved when the worker is registered to after it is started instead of when it is created.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -951,6 +951,9 @@ func (aw *AggregatedWorker) Start() error {
 		if err := aw.workflowWorker.Start(); err != nil {
 			return err
 		}
+		if aw.client.eagerDispatcher != nil {
+			aw.client.eagerDispatcher.registerWorker(aw.workflowWorker)
+		}
 	}
 	if !util.IsInterfaceNil(aw.activityWorker) {
 		if err := aw.activityWorker.Start(); err != nil {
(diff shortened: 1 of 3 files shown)


internal/internal_worker.go
```diff
@@ -951,6 +951,9 @@ func (aw *AggregatedWorker) Start() error {
 		if err := aw.workflowWorker.Start(); err != nil {
 			return err
 		}
+		if aw.client.eagerDispatcher != nil {
```
cretz, line 954: What was the reasoning behind this change? (not that I disagree with it)

cretz (approved):

internal/internal_worker.go
Quinn-With-Two-Ns: Seems inefficient to register workers that had not started yet and force every eager workflow start attempt to try them over and over. Not a huge optimization, but this code path is also for low latency

cretz: :+1: Just making sure it wasn't fixing something I couldn't see

Issue #1189, opened 2023-08-07
Package docs on testsuite incorrectly showing license instead


cretz:
## Expected Behavior

https://pkg.go.dev/go.temporal.io/sdk lists all first sentences of package-level docs.

## Actual Behavior

It shows "The MIT License" for `testsuite` because https://pkg.go.dev/go.temporal.io/sdk/testsuite is showing the license as the package docs because https://github.com/temporalio/sdk-go/blob/master/testsuite/freeport.go does not put a blank line between license and `package` declaration.

PR #1192, opened 2023-08-08
Fix testsuite package docs

Quinn-With-Two-Ns:
Fix testsuite docs


Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-go/issues/1189

cretz (approved):

PR #1193, opened 2023-08-08
Enabled worker versioning tests

Quinn-With-Two-Ns:
Enabled worker versioning tests for CI


cretz (approved):
Would love to move to `testsuite.StartDevServer` one day.

Issue #1050, opened 2023-02-21
Data race in eager activity handling


Sushisource:
Came up while running integ tests for an unrelated change:

```
==================
WARNING: DATA RACE
Read at 0x00c000316a80 by goroutine 4489:
  runtime.raceread()
      <autogenerated>:1 +0x24
  go.temporal.io/sdk/internal.(*eagerActivityExecutor).handleResponse()
      /home/runner/work/sdk-go/sdk-go/internal/internal_eager_activity.go:143 +0x388
  go.temporal.io/sdk/internal.(*workflowTaskPoller).RespondTaskCompleted()
      /home/runner/work/sdk-go/sdk-go/internal/internal_task_pollers.go:427 +0x765
  go.temporal.io/sdk/internal.(*workflowTaskPoller).RespondTaskCompletedWithMetrics()
      /home/runner/work/sdk-go/sdk-go/internal/internal_task_pollers.go:386 +0x924
  go.temporal.io/sdk/internal.(*workflowTaskPoller).processWorkflowTask()
      /home/runner/work/sdk-go/sdk-go/internal/internal_task_pollers.go:347 +0x64b
  go.temporal.io/sdk/internal.(*workflowTaskPoller).ProcessTask()
      /home/runner/work/sdk-go/sdk-go/internal/internal_task_pollers.go:296 +0x93
  go.temporal.io/sdk/internal.(*baseWorker).processTask()
      /home/runner/work/sdk-go/sdk-go/internal/internal_worker_base.go:440 +0x255
  go.temporal.io/sdk/internal.(*baseWorker).runTaskDispatcher.func2()
      /home/runner/work/sdk-go/sdk-go/internal/internal_worker_base.go:334 +0x58

Previous write at 0x00c000316a80 by goroutine 4493:
  runtime.racewrite()
      <autogenerated>:1 +0x24
  go.temporal.io/sdk/internal.awaitWaitGroup.func1()
      /home/runner/work/sdk-go/sdk-go/internal/internal_utils.go:180 +0x39

Goroutine 4489 (running) created at:
  go.temporal.io/sdk/internal.(*baseWorker).runTaskDispatcher()
      /home/runner/work/sdk-go/sdk-go/internal/internal_worker_base.go:334 +0x10f
  go.temporal.io/sdk/internal.(*baseWorker).Start.func3()
      /home/runner/work/sdk-go/sdk-go/internal/internal_worker_base.go:275 +0x39

[...]

PR #1195, opened 2023-08-09
Fix race condition in eager dispatch

Quinn-With-Two-Ns:
Fix a race condition on the `WaitGroup` in `baseWorker` with eager dispatch (activity and workflow) on worker shutdown.

In Go `WaitGroups` have confusing concurrency rules https://pkg.go.dev/sync#WaitGroup.Add:

```
Note that calls with a positive delta that occur when the counter is zero must happen before a Wait. Calls with a negative delta, or calls with a positive delta that start when the counter is greater than zero, may happen at any time. Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for. If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned. See the WaitGroup example.
```

Here is the situation

1. Eager executor reserves slot
2. In a different go routine `.stop()`  is called on the worker. This closes the `stopch`  and triggers all the go routines the baseWorker started to stop and decrement the `stopWG`  to zero
after In another go routine `stopWG.wait()`  is called
3. Eager executor attempts to give a task to the base worker, base worker tries to increment `stopWG` because it will launch a go routine to process the task, but `stopWG` is zero and `stopWG.wait()` is called so that is a illegal.

The basic issue is we cannot increment `stopWG` from a go routine the `baseWorker` did not start.

To fix this I created a new go routine to launch all eager workflow tasks that the `baseWorker` now owns so it can interact with the wait group.

### DIFF
--- a/internal/internal_eager.go
+++ b/internal/internal_eager.go
@@ -29,7 +29,7 @@ type eagerWorker interface {
 	tryReserveSlot() bool
 	// releaseSlot release a task slot acquired by tryReserveSlot
 	releaseSlot()
-	// processTaskAsync process a new task on the worker asynchronously and
-	// call callback once complete
-	processTaskAsync(task interface{}, callback func())
+	// pushEagerTask pushes a new eager workflow task to the workers task queue.
+	// should only be called with a reserved slot.
+	pushEagerTask(task eagerTask)
 }
--- a/internal/internal_eager_activity_test.go
+++ b/internal/internal_eager_activity_test.go
@@ -107,6 +107,8 @@ func TestEagerActivityCounts(t *testing.T) {
 	activityWorker := newActivityWorker(nil,
 		workerExecutionParameters{TaskQueue: "task-queue1", ConcurrentActivityExecutionSize: 5}, nil, newRegistry(), nil)
 	activityWorker.worker.isWorkerStarted = true
+	go activityWorker.worker.runEagerTaskDispatcher()
+	//defer close(activityWorker.worker.stopCh)
 
 	exec.activityWorker = activityWorker.worker
 	// Fill up the poller request channel
--- a/internal/internal_eager_activity_test.go
+++ b/internal/internal_eager_activity_test.go
@@ -108,7 +108,6 @@ func TestEagerActivityCounts(t *testing.T) {
 		workerExecutionParameters{TaskQueue: "task-queue1", ConcurrentActivityExecutionSize: 5}, nil, newRegistry(), nil)
 	activityWorker.worker.isWorkerStarted = true
 	go activityWorker.worker.runEagerTaskDispatcher()
-	//defer close(activityWorker.worker.stopCh)
 
 	exec.activityWorker = activityWorker.worker
 	// Fill up the poller request channel
--- a/internal/internal_eager_workflow.go
+++ b/internal/internal_eager_workflow.go
@@ -81,10 +81,8 @@ func (e *eagerWorkflowExecutor) handleResponse(response *workflowservice.PollWor
 			task: &eagerWorkflowTask{
 				task: response,
 			},
-			callback: func() {
-				// The processTaskAsync does not do this itself because our task is *eagerWorkflowTask, not *polledTask.
-				e.worker.releaseSlot()
-			},
+			// The processTaskAsync does not do this itself because our task is *eagerWorkflowTask, not *polledTask.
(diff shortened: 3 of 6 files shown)


internal/internal_worker_base.go
```diff
@@ -365,6 +382,23 @@ func (bw *baseWorker) runTaskDispatcher() {
...
 }
 
+func (bw *baseWorker) runEagerTaskDispatcher() {
+	defer bw.stopWG.Done()
+	for {
+		select {
+		case <-bw.stopCh:
+			// drain eager dispatch queue
+			for len(bw.eagerTaskQueueCh) > 0 {
+				eagerTask := <-bw.eagerTaskQueueCh
```
Quinn-With-Two-Ns, line 392: Poll task dispatch does not do this, it just drops polled tasks on shutdown.

Quinn-With-Two-Ns: I'll note `tryReserveSlot`  will return false if we are closing so this channel will only shrink once we are closing

Quinn-With-Two-Ns: There is still an edge case that if the worker shuts down between `tryReserveSlot` and when `pushEagerTask` it may not get processed

internal/internal_eager_activity_test.go
```diff
@@ -107,6 +107,8 @@ func TestEagerActivityCounts(t *testing.T) {
 	activityWorker := newActivityWorker(nil,
 		workerExecutionParameters{TaskQueue: "task-queue1", ConcurrentActivityExecutionSize: 5}, nil, newRegistry(), nil)
 	activityWorker.worker.isWorkerStarted = true
+	go activityWorker.worker.runEagerTaskDispatcher()
+	//defer close(activityWorker.worker.stopCh)
```
cretz, line 111: Commented out code

internal/internal_eager_workflow.go
```diff
@@ -76,13 +76,16 @@ func (e *eagerWorkflowExecutor) handleResponse(response *workflowservice.PollWor
...
-	})
+	e.worker.pushEagerTask(
+		eagerTask{
+			task: &eagerWorkflowTask{
+				task: response,
+			},
+			callback: func() {
+				// The processTaskAsync does not do this itself because our task is *eagerWorkflowTask, not *polledTask.
+				e.worker.releaseSlot()
+			},
```
cretz, line 87:
```suggestion
			callback: e.worker.releaseSlot,
```

internal/internal_worker_base.go
cretz: I wonder if the eager task receive should just be on the regular `runTaskDispatcher`? Doesn't matter except for keeping the logic close. But no benefit to having this run separately I don't think.

> Poll task dispatch does not do this, it just drops polled tasks on shutdown.

Wonder if we should open another issue to fix this

cretz (commented):
Minor discussion points, but overall works for me (for now, would like to fix stopping altogether one day)

internal/internal_worker_base.go
Quinn-With-Two-Ns: >Wonder if we should open another issue to fix this

Yes I'll open an issue for that

>I wonder if the eager task receive should just be on the regular runTaskDispatcher? Doesn't matter except for keeping the logic close. But no benefit to having this run separately I don't think.

I was trying to keep the behavior the same with this change. While I don't think it's likely putting them together would effect performance, we don't really have the tools to test it right now.

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-go/issues/1050

cretz (approved):

internal/internal_eager.go
```diff
@@ -29,7 +29,7 @@ type eagerWorker interface {
 	tryReserveSlot() bool
 	// releaseSlot release a task slot acquired by tryReserveSlot
 	releaseSlot()
-	// processTaskAsync process a new task on the worker asynchronously and
-	// call callback once complete
-	processTaskAsync(task interface{}, callback func())
+	// pushEagerTask pushes a new eager workflow task to the workers task queue.
+	// should only be called with a reserved slot.
+	pushEagerTask(task eagerTask)
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah I don't think it's worth doing in this case. If we make the eager code more complicated it is more considering

PR #1198, opened 2023-08-10
Add note on heart beating and async activities

Quinn-With-Two-Ns:
Saw this multiple times in the past month were users are using `RecordHeartbeat` incorrectly

### DIFF
--- a/activity/activity.go
+++ b/activity/activity.go
@@ -71,6 +71,9 @@ func GetMetricsHandler(ctx context.Context) metrics.Handler {
 //
 // details - the details that you provided here can be seen in the workflow when it receives TimeoutError, you
 // can check error with TimeoutType()/Details().
+//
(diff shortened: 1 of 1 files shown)


activity/activity.go
```diff
@@ -71,6 +71,9 @@ func GetMetricsHandler(ctx context.Context) metrics.Handler {
 //
 // details - the details that you provided here can be seen in the workflow when it receives TimeoutError, you
 // can check error with TimeoutType()/Details().
+//
+// Note: If using asynchronous activity completion,
+// after returning [ErrResultPending] users should heartbeat with [client.Client.RecordActivityHeartbeat]
```
Sushisource, line 76: Curious: how are they even doing this if the activity has returned the error?

Sushisource (approved):

Issue #1194, opened 2023-08-09
Update sdk-name and sdk-version in task-level metadata if changed


cretz:
**Describe the solution you'd like**

Independently, if either SDK name or SDK version changed on the workflow from what was last seen on some task, set it to the current values (same values as client-name and client-version gRPC headers).

See https://github.com/temporalio/features/issues/321

PR #1205, opened 2023-08-22
Refactor NextCommandEvents

Quinn-With-Two-Ns:
Finally doing a long asked for refactor of `NextCommandEvents` from 5 return values to a struct. The logic of the code still isn't great, but it is a bit better now. 

Doing now because I know I'll touch this code more when I do: https://github.com/temporalio/sdk-go/issues/1194

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -188,6 +188,14 @@ type (
 	historyMismatchError struct {
 		message string
 	}
+
+	processedTask struct {
+		events         []*historypb.HistoryEvent
+		markers        []*historypb.HistoryEvent
+		flags          []sdkFlag
+		msgs           []*protocolpb.Message
+		binaryChecksum string
+	}
 )
 
 func newHistory(task *workflowTask, eventsHandler *workflowExecutionEventHandlerImpl) *history {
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -188,6 +188,14 @@ type (
 	historyMismatchError struct {
 		message string
 	}
+
+	processedTask struct {
```
Quinn-With-Two-Ns, line 192: If someone has a better name then this I welcome any suggestions

cretz: How about `preparedTask` since it's not yet processed?

Quinn-With-Two-Ns: I called it processed because the SDK had done some processing on it, like extracting SDK flags and reordering events. Words like prerpare or preprocess is probably better

internal/internal_task_handlers.go
```diff
@@ -301,23 +309,40 @@ func isCommandEvent(eventType enumspb.EventType) bool {
 	}
 }
 
-// NextCommandEvents returns events that there processed as new by the next command.
-// TODO(maxim): Refactor to return a struct instead of multiple parameters
-func (eh *history) NextCommandEvents() (result []*historypb.HistoryEvent, markers []*historypb.HistoryEvent, binaryChecksum string, sdkFlags []sdkFlag, msgs []*protocolpb.Message, err error) {
+// NextTask returns the next task to be processed.
+func (eh *history) NextTask() (*preparedTask, error) {
```
cretz, line 313:
```suggestion
func (eh *history) nextTask() (*preparedTask, error) {
```

Can we not expose this? May need to change the name if there's a clash. Just helps me confirm that we don't expose via something someone might use via some interface somewhere. Not required though.

cretz (approved):

internal/internal_task_handlers.go
Quinn-With-Two-Ns: done

PR #1206, opened 2023-08-24
Added IsActivityContext func

moredure:
## What was changed
Added  IsActivityContext func to the activity package

## Why?
Need a way to check whether or not context belongs to activity


moredure:
Find out that there is one exists, sorry.

Issue #1209, opened 2023-08-24
Race condition when evicting a workflow from cache


Quinn-With-Two-Ns:
## Expected Behavior
Evicting a workflow from cache does not cause a race condition in the SDK

## Actual Behavior
Evicting a workflow from cache does causes a race condition in the SDK

## Steps to Reproduce the Problem
To reproduce run add this workflow and a test to run it to `test` in the Go SDK
```
func (w *Workflows) ReproduceRace(ctx workflow.Context) error {
	re := func(ctx workflow.Context) {
		fmt.Printf("Goroutine running \n")
		ctx, _ = workflow.NewDisconnectedContext(ctx)
		childCtx, childCancel := workflow.WithCancel(ctx)
		childCtx = workflow.WithChildOptions(childCtx, workflow.ChildWorkflowOptions{WaitForCancellation: true})
		child := workflow.ExecuteChildWorkflow(childCtx, w.SleepForDuration, 10*time.Minute)
		if err := child.GetChildWorkflowExecution().Get(ctx, nil); err != nil {
			panic(err)
		}

		defer func() {
			fmt.Printf("Goroutine defer \n")
			time.Sleep(time.Second)
			childCancel()
		}()

		workflow.Sleep(ctx, time.Hour)
		fmt.Printf("!!!! Ran passed sleep, this is unexpected \n")
	}

	for i := 0; i < 10; i++ {
		fmt.Printf("Launching goroutine: %d \n", i)
		workflow.Go(ctx, re)
	}
	fmt.Printf("Workflow sleeping \n")
	workflow.Sleep(ctx, 5*time.Second)
	fmt.Printf("Workflow returning \n")
	return nil
}
```
and run with race detector on like
`go test --race  -run TestIntegrationSuite -testify.m TestConcurrentMapWriteWorkflow`

This will trigger a race condition in the SDK, the exact races change between runs, but here is one
```
WARNING: DATA RACE
Read at 0x00c0003fe6f0 by goroutine 65:
  runtime.mapaccess1()
      /opt/homebrew/Cellar/go/1.21.0/libexec/src/runtime/map.go:396 +0x26c
  go.temporal.io/sdk/internal.(*commandsHelper).getCommand()
      /Users/quinnklassen/Documents/Code/sdk-go/internal/internal_command_state_machine.go:919 +0x54
  go.temporal.io/sdk/internal.(*commandsHelper).requestCancelExternalWorkflowExecution()
[...]

PR #1210, opened 2023-08-25
Fix race condition during cache eviction

Quinn-With-Two-Ns:
Fix race condition during cache eviction

see https://github.com/temporalio/sdk-go/issues/1209 for more details

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -664,6 +664,24 @@ func assertNotInReadOnlyState(ctx Context) {
 	}
 }
 
+func assertNotInReadOnlyStateCancellation(ctx Context) {
+	s := ctx.Value(coroutinesContextKey)
+	if s == nil {
+		panic("assertNotInReadOnlyStateCtxCancellation: not workflow context")
+	}
+	state := s.(*coroutineState)
+	// For cancellation the dispatcher may not be running because workflow cancellation
+	// is sent outside of the dispatchers loop.
+	if state.dispatcher.IsClosed() {
+		panic(panicIllegalAccessCoroutineState)
+	}
+	// use the dispatcher state instead of the coroutine state because contexts can be
+	// shared
+	if state.dispatcher.getIsReadOnly() {
+		panic(panicIllegalAccessCoroutineState)
+	}
+}
+
 func getStateIfRunning(ctx Context) *coroutineState {
 	if ctx == nil {
 		return nil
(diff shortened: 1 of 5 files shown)


internal/workflow.go
```diff
@@ -638,6 +638,7 @@ func (wc *workflowEnvironmentInterceptor) ExecuteActivity(ctx Context, typeName
 
 	if cancellable {
 		cancellationCallback.fn = func(v interface{}, more bool) bool {
+			assertNotInReadOnlyStateCancellation(ctx)
```
cretz, line 641: I don't understand exactly. Are we using panics for control flow here? Is there any other way to not do stuff during coroutine exit?

internal/internal_workflow.go
```diff
@@ -664,6 +664,24 @@ func assertNotInReadOnlyState(ctx Context) {
 	}
 }
 
+func assertNotInReadOnlyStateCancellation(ctx Context) {
```
cretz, line 667: Is this "read only" the same definition as used for query and update validators?

Quinn-With-Two-Ns: No, as is documented in the function we can't use the normal check in cancellation because of how the Go SDK implemented workflow cancellation

internal/workflow.go
Quinn-With-Two-Ns: It's using panics for control flow as much as any of our other read only state checks or non determinism checks are.

Quinn-With-Two-Ns:
I'll note even without the race condition this PR is still needed because currently contexts can be cancelled in a read only context which may generate commands and that is not valid.

internal/internal_workflow.go
cretz: Hrmm, I wonder if there's another term for read-only to not be confused with the query/update-validator definition. Either way, just naming, no biggie.

internal/workflow.go
cretz: I assume what happens is we `recover()` in the panic handler but there is no workflow task to fail so it does nothing?

Is there any other/clearer way to not do stuff during coroutine exit?

internal/internal_workflow.go
cretz, line 667: One more question I forgot to ask before I mark approved. Can you think of any case where this can break a working workflow today? I can't, just checking.

Quinn-With-Two-Ns: I named them similar because it is basically the same validation for the same purpose.

internal/workflow.go
Quinn-With-Two-Ns: This assertion is almost the same as any of our other read only assertions we have in the Go SDK so it should be implemented as such.

cretz: Is there any other/clearer way to not do stuff during coroutine exit? Can this be moved to where `cancellationCallback.fn` is invoked instead of put in each callback?

(just want to confirm that there is no better way than panicking at the beginning of every cancel callback)

Quinn-With-Two-Ns: No, again this is the same thing as any of our other read only checks we added or non determinism. We always panic in those cases because it's the only option.

cretz (approved):

Issue #1212, opened 2023-08-27
Queryable registry


edmondop:
**Is your feature request related to a problem? Please describe.**

As discussed in this thread https://temporalio.slack.com/archives/CTDTU3J4T/p1692921277359619 there might be use cases where querying the worker for the registered activities and workflow could be useful

**Describe the solution you'd like**

**Describe alternatives you've considered**
Storing this information into an external data structure or as tsurdilo mentioned : "have seen users in code that creates worker and registers activities write small file to file system with what it registered and then in worker host pod/container expose like simple get api that returns this info"

cretz:
(reposting what I put on PR)

I don't think this is necessary. Users are the ones adding to the registry, so they can keep track of what they add (or they can implement the interface themselves with something that does keep track or put what to register in a collection that is then registered all at once or whatever).

Also, adding methods to an interface is a breaking change (though we still sometimes do it).

PR #1213, opened 2023-08-27
Exposing queryable Registry

edmondop:
## What was changed

Allow the client code to query the worker for the list of registered workflows, activities and their aliases

## Checklist

1. Closes 

See #1212 

3. How was this tested:

See unit tests

4. Any docs updates needed?

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1831,3 +1831,43 @@ func executeFunction(fn interface{}, args []interface{}) (interface{}, error) {
 	}
 	return res, err
 }
+
+func (w *AggregatedWorker) ListRegisteredWorkflowNames() []string {
+	w.registry.Lock()
+	defer w.registry.Unlock()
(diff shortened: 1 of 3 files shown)


cretz:
I don't think this is necessary. Users are the ones adding to the registry, so they can keep track of what they add (or they can implement the interface themselves with something that does keep track).

Also, adding methods to an interface is a breaking change (though we still sometimes do it).

edmondop:
That was the first idea discussed on Slack and that made a lot of sense to me as well, but I suspected there were use cases where querying the state would make things much easier.

On the interface breaking problem, can you share something more? This is an additive change to an interface in the SDK, I haven't fully understood what type of best practices need to be applied at this layer.

cretz:
> I suspected there were use cases where querying the state would make things much easier.

I think it's easier to keep a slice of things to register. But even if this is easier, we have to ask ourselves whether altering an interface just for this is worth it.

Issue #1207, opened 2023-08-24
Handle interfaces in workflowcheck


ndtretyak:
**Is your feature request related to a problem? Please describe.**
In the following example I use `time.Now()`, but `workflowcheck` does not report this problem because `RegisterWorflow` is called with `Handler.BadWorkflow` rather than `handlerImpl.BadWorkflow`.

```go
type HandlerInterface interface {
	BadWorkflow(ctx workflow.Context) (int64, error)
}

type handlerImpl struct{}

func (w handlerImpl) BadWorkflow(ctx workflow.Context) (int64, error) {
	return time.Now().Unix(), nil
}

func CreateWorker(c client.Client, h HandlerInterface) worker.Worker {
	w := worker.New(c, "example", worker.Options{})
	w.RegisterWorkflow(h.BadWorkflow)
	return w
}

func main() {
	....
	w := CreateWorker(c, handlerImpl{})
}
```

**Describe the solution you'd like**
I'd like `workflowcheck` to match the interface passed to `RegisterWorkflow` with its implementations.

cretz:
A static analyzer is not going to be able to follow implementations of interfaces because Go does not have a way to define which types are meant to implement an interface at compile time. It would be non-trivial to walk the control graph backwards from `RegisterWorkflow` to fine what was really used. If at all possible, consider registering functions on concrete types. If you must use an interface, maybe have a `Register(worker.WorkflowRegistry)` method on it that each implementation implements and you call generally.

We may be able to, for every `SomeInterface`, look for `var _ SomeInterface = (*SomeType)(nil)` to find all `SomeType`s that implement it. We may also be able to have a way for you to mark workflow entry points when they couldn't otherwise be discovered at compile time. But both options are more brittle than registering functions/methods on the type that implements them.

Quinn-With-Two-Ns:
@cretz Do we need users to mark workflow entry points? Can't we assume any function that takes a `ctx workflow.Context` as the first parameter is a workflow and analyze it?

cretz:
> Can't we assume any function that takes a ctx workflow.Context as the first parameter is a workflow and analyze it?

We can probably do this, though the existing behavior of bubbling up to top-level workflow call would be lost. But maybe that is preferred? It would also be a breaking change, but probably harmless.

PR #1215, opened 2023-08-29
Check all functions with workflow.Context as first parameter

ndtretyak:
Instead of checking only arguments of `RegisterWorkflow` consider all functions like `func (workflow.Context, ...) ...` as workflow definitions (or functions called from workflows).

Closes https://github.com/temporalio/sdk-go/issues/1207

### DIFF
--- a/contrib/tools/workflowcheck/workflow/checker.go
+++ b/contrib/tools/workflowcheck/workflow/checker.go
@@ -198,7 +198,11 @@ func isWorkflowFunc(f *ast.FuncDecl, pass *analysis.Pass) bool {
 		return false
 	}
 	obj := named.Obj()
-	return obj.Pkg() != nil && obj.Pkg().Path() == "go.temporal.io/sdk/workflow" && obj.Name() == "Context"
+	if obj.Pkg() == nil || obj.Name() != "Context" {
+		return false
+	}
+	path := obj.Pkg().Path()
+	return path == "go.temporal.io/sdk/workflow" || path == "go.temporal.io/sdk/internal"
 }
 
 type configFileFlag struct{ checker *determinism.Checker }


cretz:
Nice, thanks! Please sign CLA.

Can you show an example of the CLI output for:

```go

func MyWorkflow(workflow.Context) error {
    doSomething()
}

func doSomething() {
  time.Now()
}
```

I want to make sure that this still shows the tree, though I'm sure it does.

ndtretyak:
Well, I should have tried to run the CLI myself. It turned out that `workflow.Context` belongs to `go.temporal.io/sdk/internal` rather than `go.temporal.io/sdk/workflow`. But in unit tests this is not the case.

After adding an additional check for the `internal` package I got the following graph

```
go.temporal.io/sdk/contrib/tools/workflowcheck/workflow/testdata/src/a.MyWorkflow is non-deterministic, reason: calls non-deterministic function go.temporal.io/sdk/contrib/tools/workflowcheck/workflow/testdata/src/a.doSomething
  go.temporal.io/sdk/contrib/tools/workflowcheck/workflow/testdata/src/a.doSomething is non-deterministic, reason: calls non-deterministic function go.temporal.io/sdk/contrib/tools/workflowcheck/workflow/testdata/src/a.doSomethingElse
    go.temporal.io/sdk/contrib/tools/workflowcheck/workflow/testdata/src/a.doSomethingElse is non-deterministic, reason: calls non-deterministic function time.Now
      time.Now is non-deterministic, reason: declared non-deterministic
```

cretz (approved):
Looks great to me! Will let @Quinn-With-Two-Ns take a look.

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
LGTM makes more sense for it to work this way to me

PR #1217, opened 2023-08-29
Fix the documentation of ListWorkflow for querying open workflows

eigenvariable:
## What was changed

This fixes the documentation around using `ListWorkflow` for querying open workflows.

## Why?

The comment appeared to be incorrect, at least WRT temporal cloud.  `CloseTime = missing` doesn't appear to be a valid query, whereas `CloseTime is null` seems to do what you'd expect.

## Checklist

1. Closes 

2. How was this tested:

This is just a change to a comment. I tested the accuracy of the documentation via trial and error (also via the Temporal Cloud web UI).

3. Any docs updates needed?


Quinn-With-Two-Ns (approved):
Thanks for the docs fix

Issue #1158, opened 2023-07-10
Add slog logger implementation


cretz:
**Describe the solution you'd like**

Now that https://pkg.go.dev/log/slog is a thing, we would like to add something like:

```go
type slogLogger struct{ underlying *slog.Logger }

var _ log.Logger = (*slogLogger)(nil)
var _ log.WithLogger = (*slogLogger)(nil)

func NewStructuredLogger(underlying *slog.Logger) log.Logger   { return &slogLogger{underlying} }
func (s *slogLogger) Debug(msg string, keyvals ...interface{}) { s.underlying.Debug(msg, keyvals...) }
func (s *slogLogger) Info(msg string, keyvals ...interface{})  { s.underlying.Info(msg, keyvals...) }
func (s *slogLogger) Warn(msg string, keyvals ...interface{})  { s.underlying.Warn(msg, keyvals...) }
func (s *slogLogger) Error(msg string, keyvals ...interface{}) { s.underlying.Error(msg, keyvals...) }
func (s *slogLogger) With(keyvals ...interface{}) log.Logger {
	return &slogLogger{s.underlying.With(keyvals...)}
}
```

But, we can't do it yet because it's not in the oldest Go version we support. So once it is we can add that. In the meantime, a sample can be added or people can just refer to this issue for that simple implementation.

cretz:
Actually, what's interesting is that if you ignore `log.WithLogger`, you can just provide an `*slog.Logger` directly since our interface conforms. How lucky.

Quinn-With-Two-Ns:
>But, we can't do it yet because it's not in the oldest Go version we support

Why do we need to wait? Can't we include the logger conditionally on the go version with build tags?

cretz:
Ah, yes, build tags work. Can impl whenever. May at least want to wait until officially released (because we need to update CI to that Go version to confirm this code works).

Quinn-With-Two-Ns:
Since `slog` already satisfies are logger interface, what is the value of wrapping?

cretz:
None, can document instead. Technically explicitly implementing `With` could help in some performance cases, but :shrug:.

Quinn-With-Two-Ns:
It may still be worth adding a wrapper to support https://github.com/temporalio/sdk-go/issues/960

Nikola-Milovic:
Is this still valid?

```
cannot convert slog.Default() (value of type *slog.Logger) to type "go.temporal.io/sdk/log".WithLogger: 
[...]

Issue #960, opened 2022-11-17
Loggers use different stack depths when being called from workflow vs activity


cretz:
## Expected Behavior

The same log statements in workflow and activity should be at the same stack depth so that a shared caller-skip (e.g. in Zap or Zerolog) can work for both.

## Actual Behavior

We wrap the workflow logger in a replay safe version which properly skips logs during replay. The problem is this adds an extra stack frame for just workflows. We should either: 1) wrap activities in a similar fashion, or 2) stop using a wrapped replayer, or 3) have a `interface WithSkipCallers { AddCallerSkip(int) }` interface that loggers can implement.

Only 3 is backwards compatible. 1 or 2 could break existing user expectations.

## Replication

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"time"

	"github.com/google/uuid"
	"github.com/rs/zerolog"
	"go.temporal.io/sdk/activity"
	"go.temporal.io/sdk/client"
	sdklog "go.temporal.io/sdk/log"
	"go.temporal.io/sdk/worker"
	"go.temporal.io/sdk/workflow"
)

func main() {
	if err := run(); err != nil {
		log.Fatal(err)
	}
}

func run() error {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	log.Printf("Creating client")
	logWriter := zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: zerolog.TimeFormatUnix}
	// We add 3 more to the skip count because we are in our log
	// TODO(cretz): We need 3 for workflows but only 2 for activities because
	// workflows add an extra wrapping which increases the stack depth
	skipLogFrameCount := zerolog.CallerSkipFrameCount + 3
	c, err := client.Dial(client.Options{
		Logger: &ZeroLogLogger{
			zerolog.New(logWriter).With().CallerWithSkipFrameCount(skipLogFrameCount).Timestamp().Logger(),
		},
	})
	if err != nil {
		return fmt.Errorf("failed creating client: %w", err)
	}
	defer c.Close()

	log.Printf("Starting worker")
	taskQueue := uuid.NewString()
	w := worker.New(c, taskQueue, worker.Options{})
	w.RegisterWorkflow(MyWorkflow)
	w.RegisterActivity(MyActivity)
	if err := w.Start(); err != nil {
		return fmt.Errorf("failed starting worker client: %w", err)
	}
	defer w.Stop()

	log.Printf("Starting workflow")
[...]

Issue #685, opened 2022-01-05
Add SkipLogger interface for logger


cretz:
**Is your feature request related to a problem? Please describe.**

When someone uses a logger at the client level, sometimes we add a replay logger wrapper which changes the caller skip if someone has done a custom skip.

**Describe the solution you'd like**

Like the server does at https://github.com/temporalio/temporal/blob/v1.14.1/common/log/interface.go#L64-L68, we should also have a skip interface that we can call w/ the value of `1` on when replaying.

PR #1219, opened 2023-08-30
Add slog adapter

Quinn-With-Two-Ns:
Add an adapter for the new `slog` and add support for an optional interface `Logger` can implement to skip caller stacks.

Marking as draft because the adapter and skip feature need unit tests and I need to read how to test `slog`

closes https://github.com/temporalio/sdk-go/issues/1158

closes https://github.com/temporalio/sdk-go/issues/685

closes https://github.com/temporalio/sdk-go/issues/960

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -34,8 +34,9 @@ type (
 	}
 
 	// WithSkipCallers is an optional interface that a Logger can implement that
-	// indicate the number of stack frames to skip.
+	// may create a new child logger that skips the number of stack frames of the caller.
+	// This call must not mutate the original logger.
 	WithSkipCallers interface {
-		AddCallerSkip(int) Logger
+		WithCallerSkip(int) Logger
 	}
 )
--- a/log/slog.go
+++ b/log/slog.go
@@ -86,7 +86,7 @@ func (s *slogLogger) With(keyvals ...interface{}) Logger {
 	}
 }
 
-func (s *slogLogger) AddCallerSkip(depth int) Logger {
+func (s *slogLogger) WithCallerSkip(depth int) Logger {
 	return &slogLogger{
 		logger: s.logger,
 		depth:  s.depth + depth,
--- a/log/slog.go
+++ b/log/slog.go
@@ -1,0 +1,94 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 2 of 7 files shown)


log/logger.go
```diff
@@ -32,4 +32,10 @@ type (
 		Warn(msg string, keyvals ...interface{})
 		Error(msg string, keyvals ...interface{})
 	}
+
+	// WithSkipCallers is an optional interface that a Logger can implement that
+	// indicate the number of stack frames to skip.
+	WithSkipCallers interface {
```
Quinn-With-Two-Ns, line 38: Some users brought up the logger they pass to temporal already, unintentionally, implements this interface and this would be a breaking change for them. I may have to rename this or add an extra guard method to the interface to ensure people mean to implement it.

log/slog.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (s *slogLogger) Warn(msg string, keyvals ...interface{}) {
+	s.log(context.Background(), slog.LevelWarn, msg, keyvals...)
+}
+
+func (s *slogLogger) Error(msg string, keyvals ...interface{}) {
+	s.log(context.Background(), slog.LevelError, msg, keyvals...)
+}
+
+func (s *slogLogger) log(ctx context.Context, level slog.Level, msg string, args ...any) {
```
Quinn-With-Two-Ns, line 65: basically taken from how `slog` is implemented + adding caller skip https://cs.opensource.google/go/go/+/master:src/log/slog/logger.go;l=209

log/logger.go
cretz: I don't think it's a breaking change, it's not like you added this on the existing interface. If they do implement this interface exactly, with return type and all, great this works because the behavior should be obvious. If they make have the same method name, it's harmless because it won't match the interface. And if they _did_ want skip support but already have a method with this name, they can wrap-and-embed, only providing this function which shadows their other one for this specific use.

log/slog_test.go
```diff
@@ -0,0 +1,31 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+//go:build go1.21
+
+package log
+
+import "testing"
+
+func TestSlogAddapter(t *testing.T) {}
```
cretz, line 31: I assume coming soon? Ideally can show in-activity and in-workflow logs (via test suite is fine) show this filename in the record.

Quinn-With-Two-Ns: Yes, this was supposed to be marked as a draft,  but I guess I misclicked see:

>Marking as draft because the adapter and skip feature need unit tests and I need to read how to test slog

log/logger.go
cretz: Hrmm, I wonder how they're getting their logger from the `ReplayLogger` to make that call. Regardless I think it's a bit rough for us not to be able to add an interface because someone else had that method and was using it already.

But if this is a big enough concern, I guess we can change to `WithCallerSkip` which maybe should be the name anyways since this should probably never mutate the existing logger and we need to document that too. Granted if that is a problem too because maybe they used that method name also, then what do we do if we can never add extensibility points because of this fear?

Quinn-With-Two-Ns: >Granted if that is a problem too because maybe they used that method name also, then what do we do if we can never add extensibility points because of this fear?

If you pick method names that are used by other common loggers your asking for this kind of issue. It's an easy problem to work around, two simple solution are pick a method name users are very unlikely to have implemented themselves like `AddCallerSkipTemporal` or require the interface to have an addition method like `TemporalSkipLogger`

cretz: That's a bit rough I think, Is `WithCallerSkip` unique enough? (also, can we doc that this method should return a new logger with the skip amount but not mutate the current one?)

Quinn-With-Two-Ns: > That's a bit rough I think. Is WithCallerSkip unique/good enough? 

Since `WithCallerSkip` is used by zap I don't think it is unique.

> (also, can we doc that this method should return a new logger with the skip amount but not mutate the current one?)

We should  document that it may not mutate the existing one.

Quinn-With-Two-Ns: >AddCallerSkip is there, but not WithCallerSkip

Ah I was looking at a wrapper for Zap, not zap itself.  Still a quicks search shows other loggers define `WithCallerSkip` with the same signature as us.

cretz: But they won't match the interface then since the return type is different. The only concern I think is those that match the interface and get misused when _we_ call them internally. I'm just trying to avoid those that have written this exact method signature with the same return value and all.

Quinn-With-Two-Ns: if their `WithCallerSkip` returns a `logger` that matches our interface then they would match the interface no?

cretz: No it has to be the exact signature. Return types are not covariant when matching interfaces. E.g. https://go.dev/play/p/USVfNSHJUIk (ref https://go.dev/ref/spec#Interface_types on exact method set match). That's why I considered it mostly not a big deal above: "If they make have the same method name but different params/return, it's harmless because it won't match the interface".

cretz (approved):

Issue #329, opened 2021-01-08
Potential deadlock detected: workflow goroutine "root" didn't yield for over a second


shaunco:
## Expected Behavior
After switching to v1.3.0 of the sdk-go package, we started getting `Potential deadlock detected: workflow goroutine "root" didn't yield for over a second` panics on `ExecuteLocalActivity()` calls. Happens about 9 out of 10 times, but seems to randomly succeed that 10th time. All is good if I switch back to v1.2.0.

```
4:57PM INF Started Worker%!(EXTRA string=Namespace, string=default, string=TaskQueue, string=MY_TASK_QUEUE, string=WorkerID, string=48528@MYMACHINE@)
4:57PM ERR Workflow panic%!(EXTRA string=Namespace, string=default, string=TaskQueue, string=MY_TASK_QUEUE, string=WorkerID, string=48528@MYMACHINE@, string=WorkflowType, string=MyService.Create, string=WorkflowID, string=MyService.Create-workflow-1f309338-a827-4f34-8035-efc22fdd8fa5, string=RunID, string=bb53d3e1-9ba4-447f-825f-4f0eeb3ad2e1, string=Attempt, int32=5, string=Error, *internal.workflowPanicError=Potential deadlock detected: workflow goroutine "root" didn't yield for over a second, string=StackTrace, string=process event for MY_TASK_QUEUE [panic]:
go.temporal.io/sdk/internal.(*coroutineState).call(0xc00059b270)
	/go/pkg/mod/go.temporal.io/sdk@v1.3.0/internal/internal_workflow.go:874 +0x35e
go.temporal.io/sdk/internal.(*dispatcherImpl).ExecuteUntilAllBlocked(0xc00059b220, 0x0, 0x0)
	/go/pkg/mod/go.temporal.io/sdk@v1.3.0/internal/internal_workflow.go:959 +0x53f
go.temporal.io/sdk/internal.executeDispatcher(0x2044d60, 0xc00059cec0, 0x20490a0, 0xc00059b220)
	/go/pkg/mod/go.temporal.io/sdk@v1.3.0/internal/internal_workflow.go:566 +0x87
go.temporal.io/sdk/internal.(*syncWorkflowDefinition).OnWorkflowTaskStarted(0xc00059ce80)
	/go/pkg/mod/go.temporal.io/sdk@v1.3.0/internal/internal_workflow.go:539 +0x78
[...]

PR #1221, opened 2023-09-01
Add replay option to disable deadlock detection

jlegrone:
## What was changed

This PR adds a new field `DisableDeadlockDetection` to `WorkflowReplayerOptions`. Setting this to `true` disables deadlock detection during replay.

## Why?

The `TEMPORAL_DEBUG` environment variable is intended to serve a similar purpose, but can be [tricky to configure](https://github.com/temporalio/sdk-go/issues/329#issuecomment-1464894075) and is a global setting.

Deadlock detection can be disabled programmatically using `DeadlockDetectionTimeout` on `worker.Options`. It makes sense to support something similar in `WorkflowReplayerOptions` since attaching a debugger is such a common use case when replaying workflows.

### How was this tested:

Please see included unit test.

### Any docs updates needed?

The updated godoc is probably enough, though this would be a useful feature to mention in tutorials about workflow debugging.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1122,6 +1122,7 @@ type WorkflowReplayer struct {
 	failureConverter         converter.FailureConverter
 	contextPropagators       []ContextPropagator
 	enableLoggingInReplay    bool
+	disableDeadlockDetection bool
 	mu                       sync.Mutex
 	workflowExecutionResults map[string]*commonpb.Payloads
 }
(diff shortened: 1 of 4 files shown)


internal/internal_worker.go
```diff
@@ -1370,6 +1376,9 @@ func (aw *WorkflowReplayer) replayWorkflowHistory(logger log.Logger, service wor
 			SdkMetadata:                     true,
 		},
 	}
+	if aw.disableDeadlockDetection {
+		params.DeadlockDetectionTimeout = math.MaxInt64
```
cretz, line 1380:
```suggestion
		params.DeadlockDetectionTimeout = unlimitedDeadlockDetectionTimeout
```

Same thing, but we already have a constant for this. But not a merge blocker, just if you make another commit.

cretz (approved):
Yeah, we had traditionally avoided being able to disable deadlock detection programmatically in the Go SDK specifically because it can/will be misused. And we're supporting "debug mode" in other SDKs via parameter and env var too, so I suppose we can here.

Will wait to merge to get @Quinn-With-Two-Ns's approval

Quinn-With-Two-Ns (approved):

cretz:
@jlegrone - merging master into your branch and merging when CI passes

PR #1222, opened 2023-09-01
Update readme for slog

Quinn-With-Two-Ns:
Update readme to recommend users use slog integration.

Also added some interface checks to the loggers


cretz (approved):
Works for me, but really would love to remove altogether and rely on our other docs

PR #1223, opened 2023-09-04
chore: create `UpdateCallbacks` binding for the RoadRunner

rustatian:
## What was changed

- Create binding for the `UpdateCallbacks` in the `internalbindings` module.

## Why?

- RoadRunner temporal integration to the `WorkflowDefinition` callback.

## Checklist

1. Closes 

2. How was this tested:

RoadRunner temporal + php-sdk tests

3. Any docs updates needed?

PHP-SDK docs, new section about UpdateWorkflow support.

### DIFF
--- a/internalbindings/internalbindings.go
+++ b/internalbindings/internalbindings.go
@@ -75,6 +75,8 @@ type (
 	TimerID = internal.TimerID
 	// ContinueAsNewError used by a workflow to request continue as new
 	ContinueAsNewError = internal.ContinueAsNewError
+	// UpdateCallbacks used by a workflow to request updates
+	UpdateCallbacks = internal.UpdateCallbacks
 )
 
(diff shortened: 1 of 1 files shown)


internalbindings/internalbindings.go
```diff
@@ -75,6 +75,8 @@ type (
 	TimerID = internal.TimerID
 	// ContinueAsNewError used by a workflow to request continue as new
 	ContinueAsNewError = internal.ContinueAsNewError
+	// UpdateCallbacks used by a workflow to request updates
```
Quinn-With-Two-Ns, line 78: This comment isn't correct, the callbacks  is used to report the result of an update,

rustatian: Fixed

cretz (approved):

PR #1224, opened 2023-09-05
Handle generic maps and channels

ndtretyak:
## What was changed
`determinism` checker now unwraps generic types, so that functions iterating over a generic `map` are marked non-deterministic


contrib/tools/workflowcheck/determinism/checker.go
```diff
@@ -381,6 +366,41 @@ func (c *collector) collectFuncInfo(fn *types.Func, decl *ast.FuncDecl) {
 	})
 }
 
+func (c *collector) checkRangeType(rangeType types.Type, n ast.Node, fn *types.Func) Reason {
+L: // Unwrap named or generic type
```
Quinn-With-Two-Ns, line 370: `L:`?

ndtretyak: Should I change it to `Loop:`?

Quinn-With-Two-Ns: Originally I was confused why you added a  label at all. I see  you changed the if-else to a select so it is needed now.

Bit of a nitpick, but I'd prefer the label be `Loop`

ndtretyak: fixed

contrib/tools/workflowcheck/determinism/checker.go
```diff
@@ -381,6 +366,42 @@ func (c *collector) collectFuncInfo(fn *types.Func, decl *ast.FuncDecl) {
...
+	// Unwrap named or generic type
+	for {
+		switch rangeType.(type) {
+		case *types.Named, *types.TypeParam:
+			rangeType = rangeType.Underlying()
+		default:
+			break Loop
+		}
+	}
+	switch t := rangeType.(type) {
```
cretz, line 380:
```suggestion
	switch t := rangeType.(type) {
	case *types.Named:
		return c.checkRangeType(rangeType.Underlying(), n, fn)
	case *types.TypeParam:
		return c.checkRangeType(rangeType.Underlying(), n, fn)
```

Recursion reads a bit better to me, but not a big deal

cretz (approved):
Looks great, thanks for the contribution!

PR #1225, opened 2023-09-05
Clarify GetLastCompletionResult works with schedule

Quinn-With-Two-Ns:
Clarify GetLastCompletionResult works with schedules as well as cron


cretz (approved):
Getting some internal clarification on how this is accomplished from the scheduler workflow (I wonder if server is using some special API known only to them), but assuming this is accurate, LGTM.

Issue #1063, opened 2023-03-15
workflowcheck is failing non-deterministic checks when code is wrapped with SideEffect


mindaugasrukas:
## Expected Behavior
`workflowcheck` ignores `SideEffect` block.

## Actual Behavior
`workflowcheck` is failing `non-deterministic` check.

## Steps to Reproduce the Problem

  1. In a Workflow, add any non-deterministic code and wrap it with the SideEffect.
  1. run `workflowcheck -show-pos ./...`

## Specifications

  - Version: `go install go.temporal.io/sdk/contrib/tools/workflowcheck@latest`
  - Platform:
```
% uname -v
Darwin Kernel Version 22.3.0: Mon Jan 30 20:38:37 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T6000
```

edmondop:
Could have been fixed?  Using the latest I couldn't trigger the problem

```
package my_stuff

import (
	"go.temporal.io/sdk/workflow"
	"math/rand"
)

func MyWorkflow(context workflow.Context) {
	encodedRandom := workflow.SideEffect(context, func(ctx workflow.Context) interface{} {
		return rand.Intn(100)
	})
	var random int
	encodedRandom.Get(&random)
}

```

Quinn-With-Two-Ns:
Closing as this is resolved by https://github.com/temporalio/sdk-go/pull/1230

PR #1230, opened 2023-09-11
Code inside workflow.SideEffect is deterministic

ndtretyak:
## What was changed
Now code inside `workflow.SideEffect` is not analyzed


contrib/tools/workflowcheck/workflow/checker.go
```diff
@@ -86,16 +88,21 @@ func NewChecker(config Config) *Checker {
 	if config.DebugfFunc == nil {
 		config.DebugfFunc = log.Printf
 	}
+	// Default deterministic wrappers
+	if config.DeterministicWrappers == nil {
+		config.DeterministicWrappers = map[string][]string{"go.temporal.io/sdk/workflow": {"SideEffect"}}
```
cretz, line 93: Some notes:

1. Can we hardcode this one as always safe? (you'll also be able to get rid of the new dependency)
2. Not sure we need deterministic wrappers exposed to config
3. Include `MutableSideEffect`

cretz (commented):
Thanks! Suggestion to simplify and include `MutableSideEffect` made above.

(note to issue, this closes #1063)

contrib/tools/workflowcheck/workflow/checker.go
ndtretyak: 1. Do you mean hardcode path `go.temporal.io/sdk/workflow` inside `determinism` checker? Now this checker does not know about Temporal 🤔 
2. done
3. done

cretz: Sorry, yeah, I guess they can be in the object. Not sure it should be called "deterministic wrappers" though (we use that term for many of our async APIs in workflows).

I wonder if we can just add `//workflowcheck:ignore` to the side effect and mutable side effect functions in the SDK source. Or if that doesn't work, set them in the default ident refs. And if those don't work with anonymous functions (or function references or whatever), we should probably make them work.

ndtretyak: - `//workflowcheck:ignore` won't work since the problem is not in `SideEffect` call but in its argument which are analyzed anyway
- deterministic ident refs should not accept nondeterministic calls as arguments. I added an example for this https://github.com/temporalio/sdk-go/pull/1230/files#diff-4f801589f51fc7947c2a5830f4563fbb0c1559d26f04d49d91d896daff9a2b21R66

So, I think we need another mechanism which just skips checking arguments of a predefined function. Maybe `DeterministicWrapper` is not the right name for this mechanism though.

cretz: Hrmm, true, makes sense. Maybe `AcceptsNonDeterministicParameters` is best.

Question, with this PR, what happens here:

```go
getRandom := func() string { return uuid.NewString() }
workflow.SideEffect(ctx, getRandom)
``

I am thinking that we should not be failing until a function is referenced or invoked instead of just defined.

Comment (ndtretyak) [reply contrib/tools/workflowcheck/workflow/checker.go]: The `determinism` checker is almost stateless now. It just traverses the AST and looks for non-deterministic behaviour only in the current node. So, we don't know what `getRandom` really means inside the `SideEffect` code.

For the same reason we don't fail in the following example
```
a := time.Now
a()
```

I would suggest creating another issue for this particular problem.

cretz: Hrmm, yeah, we don't set whether a variable carries non-deterministic. My concern is not about the function reference than the local variable. I will open another issue.

cretz (approved):

contrib/tools/workflowcheck/workflow/checker.go
ndtretyak: Could you merge this please?

PR #1232, opened 2023-09-11
feat(contrib/opentelemetry): add TraceID and SpanID fields to the logger

mlcdf:
Hello :)

## What was changed

This PR adds the TraceID and the SpanID to the logger fields.

## Why?

From the [opentelemetry spec](https://opentelemetry.io/docs/specs/otel/logs/#log-correlation):

> It is a standard practice to record the execution context (trace and span ids as well as user-defined context) in the spans. OpenTelemetry extends this practice to logs where possible by including [TraceId](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-traceid) and [SpanId](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-spanid) in the LogRecords. This allows to directly correlate logs and traces that correspond to the same execution context. It also allows to correlate logs from different components of a distributed system that participated in the particular request execution.

## Checklist

2. How was this tested:

I added a test to make sure the fields are indeed present.

To test this, I needed a logger that could fill the `log.Logger` and I didn't wanted to add an extra dependency or roll out my own. So I relied on the new `slog` package (that's why I had to create new file and add the build annotation). Let me know if it's ok for you.

3. Any docs updates needed?

I'm not sure.

### DIFF
--- a/contrib/opentelemetry/tracing_interceptor.go
+++ b/contrib/opentelemetry/tracing_interceptor.go
@@ -34,6 +34,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 
 	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/log"
 )
 
 // DefaultTextMapPropagator is the default OpenTelemetry TextMapPropagator used
@@ -194,6 +195,20 @@ func (t *tracer) StartSpan(opts *interceptor.TracerStartSpanOptions) (intercepto
 	}
 
 	return &tracerSpan{Span: span}, nil
+}
+
+func (t *tracer) GetLogger(logger log.Logger, ref interceptor.TracerSpanRef) log.Logger {
+	span, ok := ref.(*tracerSpan)
+	if !ok {
+		return logger
+	}
+
+	logger = log.With(logger,
+		"TraceID", span.SpanContext().TraceID(),
+		"SpanID", span.SpanContext().SpanID(),
+	)
+
+	return logger
 }
 
 type tracerSpanRef struct{ trace.SpanContext }
--- a/contrib/opentelemetry/tracing_interceptor_logger_test.go
+++ b/contrib/opentelemetry/tracing_interceptor_logger_test.go
@@ -1,0 +1,104 @@
+// The MIT License
+//
+// Copyright (c) 2021 Temporal Technologies Inc.  All rights reserved.
+//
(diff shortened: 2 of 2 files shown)


contrib/opentelemetry/tracing_interceptor.go
```diff
@@ -196,6 +197,20 @@ func (t *tracer) StartSpan(opts *interceptor.TracerStartSpanOptions) (intercepto
...
 }
 
+func (t *tracer) GetLogger(logger log.Logger, ref interceptor.TracerSpanRef) log.Logger {
+	span, ok := ref.(*tracerSpan)
+	if !ok {
+		return logger
+	}
+
+	logger = log.With(logger,
+		"TraceID", span.SpanContext().TraceID(),
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: We could add an `IsValid()` check If this is a concern. I don't believe this function will get called if `ref` is not valid

contrib/opentelemetry/tracing_interceptor_logger_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"go.temporal.io/sdk/activity"
+	"go.temporal.io/sdk/contrib/opentelemetry"
+	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/testsuite"
+	"go.temporal.io/sdk/worker"
+	"go.temporal.io/sdk/workflow"
+)
+
+// +build1.21
```
Quinn-With-Two-Ns, line 45: this should be `//go:build go1.21` no? Go 1.20 CI is failing on this file

mlcdf: Good catch! It's fixed :)

Quinn-With-Two-Ns (approved):
I'll leave open for a couple days before merging if @cretz has any comments

cretz (approved):
LGTM.

I would originally be hesitant to add log multiple tags to all existing logs, but this is a good case for doing it and I can't see a good enough reason to provide an opt-out at this time.

mlcdf:
When may I expect a new release of the sdk?

cretz:
No exact time, but hopefully soon. In the meantime, you can depend on this commit (or `go get` + `@master`) in your `go.mod`.

PR #1239, opened 2023-09-15
Add doc links throughout our user-facing APIs

tdeebswihart:
## What was changed
I backfilled doc links throughout our user-facing APIs. I didn't touch the internal package as those aren't supposed to be for external consumption.

While here I cleaned up a few things, like adding links to referenced code samples and removing a doc comment in a spot where they're unsupported; you can't add links in the comments on interface methods or struct members.

## Why?
Linking readers to referenced APIs saves them time when trying to understand how to use temporal.

## Checklist
This was tested by running godoc locally and inspecting the rendered docs. Note that [cross-package doc links still don't work](https://github.com/golang/go/issues/56683) so cannot be verified locally.

### DIFF
--- a/activity/doc.go
+++ b/activity/doc.go
@@ -102,7 +102,7 @@ the Temporal managed service.
 	}
 
 When the Activity times out due to a missed heartbeat, the last value of the details (progress in the above sample) is
-returned from the workflow.ExecuteActivity function as the details field of TimeoutError with TimeoutType_HEARTBEAT.
+returned from the [workflow.ExecuteActivity] function as the details field of [temporal.TimeoutError] with TimeoutType_HEARTBEAT.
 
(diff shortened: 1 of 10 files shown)


cretz (approved):
Nice, thanks!

cretz:
> I didn't touch the internal package as those aren't supposed to be for external consumption.

Not necessarily true (many times that's the only way we can document struct fields). But this LGTM, feel free to merge at will.

tdeebswihart:
> > I didn't touch the internal package as those aren't supposed to be for external consumption.
> 
> Not necessarily true (many times that's the only way we can document struct fields). But this LGTM, feel free to merge at will or let us know if it won't let you.

Yeah, I'm used to those being internal only so skipped them out of habit. If missing docs still bothers me I'll take a swing at those too

Issue #1234, opened 2023-09-13
Disable creating a session worker and a versioned worker together


Quinn-With-Two-Ns:
Due to https://github.com/temporalio/sdk-go/issues/1227, users cannot combine sessions and worker versioning . Until the version API can handle the constraints of sessions we agreeded to fail when trying to combine sessions and worker version.

PR #1241, opened 2023-09-19
Disable running sessions on a versioned worker

Quinn-With-Two-Ns:
Disable running sessions and versioned workers together because they are not currently compatible

closes https://github.com/temporalio/sdk-go/issues/1234


cretz (approved):
Ug, hate that we have to panic, but it is the best way in this situation (no error and don't want to wait until start/run to fail)

PR #1245, opened 2023-09-20
Add SDK version and name to history

Quinn-With-Two-Ns:
Add SDK name and version to the SDK metadata on WFT complete. Only write if this is the first time or a change.

closes https://github.com/temporalio/sdk-go/issues/1194

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -166,6 +166,10 @@ type (
 		contextPropagators       []ContextPropagator
 		deadlockDetectionTimeout time.Duration
 		sdkFlags                 *sdkFlags
+		sdkVersionUpdated        bool
+		sdkVersion               string
+		sdkNameUpdated           bool
+		sdkName                  string
 
 		protocols *protocol.Registry
 	}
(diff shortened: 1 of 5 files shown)


internal/internal_event_handlers.go
```diff
@@ -350,6 +354,22 @@ func (wc *workflowEnvironmentImpl) Send(msg *protocolpb.Message, opts ...msgSend
 	wc.outbox = append(wc.outbox, outboxEntry{msg: msg, eventPredicate: sendCfg.pred})
 }
 
+func (wc *workflowEnvironmentImpl) getNewSdkName() string {
```
cretz, line 357:
```suggestion
func (wc *workflowEnvironmentImpl) getNewSdkNameAndReset() string {
```

A bit pedantic, but a bit rough to see mutations in getters even internal. Don't need to change though.

cretz (approved):
LGTM. Didn't dig into task processing details, but seems with the updated bool check that there is low possibility of this value flapping between workers.

internal/internal_event_handlers.go
Quinn-With-Two-Ns: Yeah that's fair. That is why I called it `getNewSdkName` rather than `getSdkName`. Adding `reset`  is fine to

Quinn-With-Two-Ns:
>LGTM. Didn't dig into task processing details, but seems with the updated bool check that there is low possibility of this value flapping between workers.

Did you mean it would flap if the two workers had different versions of the SDK?

PR #1246, opened 2023-09-21
Expose SetContinuedAsNewRunID in the TestWorkflowEnvironment

lminaudier:
## What was changed

This extends the `TestWorkflowEnvironment` to expose a new `SetContinuedAsNewRunID` API to set the workflow under test `ContinuedAsNewRunID` field in its `WorkflowInfo`.

## Why?

When building some workflows ("actors" like ones are a good example), you sometimes want to be able to distinguish when a workflow is starting and when it's being continued as new. This allows you to execute different behavior like restoring some state from an external system.

One way to check if your current workflow execution has been continued as new is to inspect the workflow info `ContinuedAsNewRunID` field. When it's set, then you now you are not starting fresh. It is nice because you don't need any special field in your workflow inputs to detect such cases.

Example:

```golang
func MyWorkflow(ctx workflow.Context, input string) error {
    if workflow.GetInfo(ctx).ContinuedAsNewRunID == "" {
        var v string
        err := workflow.ExecuteActivity(ctx, MyActivity).Get(ctx, &v)
        if err != nil {
            return err
        }
        input = v
    }

    s = selector.NewSelector()
    s.AddReceive(..., func() {
        c.Receive(ctx, &v)
        input = v
    })
    s.AddReceive(...)
    s.Select(ctx)

    return workflow.NewContinueAsNewError(ctx, MyWorkflow, input)
}
```

When testing such workflows though, you may want to test behavior only after a first `ContinueAsNew` call has been made. The easiest way to do that is to setup the test environment to fill up the workflow info `ContinuedAsNewRunID` field.

This is what this commit is doing, it extends the
`TestWorkflowEnvironment` to expose a `SetContinuedExecutionID` field to an arbitrary value.

This way in workflow unit tests, we can skip the code paths we don't care about and don't resort to mocking activities and child workflows for nothing leading to easier to implement and easier to read tests.

## Checklist

1. Closes 

N/A

2. How was this tested:

I only added a unit test (so far).

3. Any docs updates needed?

We probably want to rely on Godocs for that API.

--- 

Thoughts?

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -318,6 +318,12 @@ func (e *TestWorkflowEnvironment) SetContinueAsNewSuggested(suggest bool) {
 	e.impl.setContinueAsNewSuggested(suggest)
 }
 
+// SetContinuedExecutionID sets the value that is returned from
+// GetInfo(ctx).ContinuedExecutionRunID
+func (e *TestWorkflowEnvironment) SetContinuedAsNewRunID(rid string) {
+	e.impl.setContinuedExecutionID(rid)
+}
+
 // OnActivity setup a mock call for activity. Parameter activity must be activity function (func) or activity name (string).
 // You must call Return() with appropriate parameters on the returned *MockCallWrapper instance. The supplied parameters to
 // the Return() call should either be a function that has exact same signature as the mocked activity, or it should be
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -318,10 +318,10 @@ func (e *TestWorkflowEnvironment) SetContinueAsNewSuggested(suggest bool) {
 	e.impl.setContinueAsNewSuggested(suggest)
 }
 
-// SetContinuedExecutionID sets the value that is returned from
+// SetContinuedExecutionRunID sets the value that is returned from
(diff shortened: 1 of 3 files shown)


Quinn-With-Two-Ns (approved):
Thanks for the contribution

internal/workflow_testsuite.go
```diff
@@ -318,6 +318,12 @@ func (e *TestWorkflowEnvironment) SetContinueAsNewSuggested(suggest bool) {
 	e.impl.setContinueAsNewSuggested(suggest)
 }
 
+// SetContinuedExecutionID sets the value that is returned from
+// GetInfo(ctx).ContinuedExecutionRunID
+func (e *TestWorkflowEnvironment) SetContinuedAsNewRunID(rid string) {
+	e.impl.setContinuedExecutionID(rid)
+}
```
cretz, line 325:
```suggestion
func (e *TestWorkflowEnvironment) SetContinuedExecutionRunID(rid string) {
	e.impl.setContinuedExecutionRunID(rid)
}
```

Let's match the field name.

lminaudier: ho shoot, my bad, it was my intent but I mixed things up.

Issue #1237, opened 2023-09-13
Support build ID reachability API


Quinn-With-Two-Ns:
Build ID reachability API should be supported in the Go SDK

https://docs.temporal.io/workers#build-id-reachability

PR #920, opened 2022-09-27
Add versioning API methods to client

Sushisource:
## What was changed
Added top-level methods for the versioning API to the client.
Added worker flag to opt in to the feature

Go fmt also seemed to get a little more... aggressive? Some changes b/c of that.

## Why?
So we can use them in tctl and eventually release

## Checklist
Relates to https://github.com/temporalio/tctl/issues/250

1. Closes 

2. How was this tested:
Added some tests

3. Any docs updates needed?

PR #1247, opened 2023-09-22
Add TaskReachability API

Quinn-With-Two-Ns:
Add task reachability API `GetWorkerTaskReachability` to Go.

closes https://github.com/temporalio/sdk-go/issues/1237

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -50,9 +50,6 @@ import (
 type TaskReachability = internal.TaskReachability
 
 const (
-	// TaskReachabilityNotFetched indicates that reachability for this task
-	// was not fetched
-	TaskReachabilityNotFetched = internal.TaskReachabilityNotFetched
 	// TaskReachabilityUnspecified indicates the reachability was not specified
 	TaskReachabilityUnspecified = internal.TaskReachabilityUnspecified
 	// TaskReachabilityNewWorkflows indicates the Build Id might be used by new workflows
--- a/client/client.go
+++ b/client/client.go
@@ -44,6 +44,28 @@ import (
 	"go.temporal.io/sdk/internal/common/metrics"
 )
 
+// TaskReachability specifies which category of tasks may reach a worker on a versioned task queue.
+// Used both in a reachability query and its response.
+// WARNING: Worker versioning is currently experimental
+type TaskReachability = internal.TaskReachability
+
+const (
+	// TaskReachabilityNotFetched indicates that reachability for this task
+	// was not fetched
+	TaskReachabilityNotFetched = internal.TaskReachabilityNotFetched
+	// TaskReachabilityUnspecified indicates the reachability was not specified
+	TaskReachabilityUnspecified = internal.TaskReachabilityUnspecified
+	// TaskReachabilityNewWorkflows indicates the Build Id might be used by new workflows
+	TaskReachabilityNewWorkflows = internal.TaskReachabilityNewWorkflows
+	// TaskReachabilityExistingWorkflows indicates the Build Id might be used by open workflows
+	// and/or closed workflows.
+	TaskReachabilityExistingWorkflows = internal.TaskReachabilityExistingWorkflows
+	// TaskReachabilityOpenWorkflows indicates the Build Id might be used by open workflows.
+	TaskReachabilityOpenWorkflows = internal.TaskReachabilityOpenWorkflows
+	// TaskReachabilityClosedWorkflows indicates the Build Id might be used by closed workflows
+	TaskReachabilityClosedWorkflows = internal.TaskReachabilityClosedWorkflows
+)
+
 const (
 	// DefaultHostPort is the host:port which is used if not passed with options.
 	DefaultHostPort = internal.LocalHostPort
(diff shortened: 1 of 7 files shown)


client/client.go
```diff
@@ -518,6 +560,11 @@ type (
 		// WARNING: Worker versioning is currently experimental
 		GetWorkerBuildIdCompatibility(ctx context.Context, options *GetWorkerBuildIdCompatibilityOptions) (*WorkerBuildIDVersionSets, error)
 
+		// GetWorkerTaskReachability
+		// Returns which versions are is still in use by open or closed workflows
+		// WARNING: Worker versioning is currently experimental
+		GetWorkerTaskReachability(ctx context.Context, options *GetWorkerTaskReachabilityOptions) (*WorkerTaskReachability, error)
```
cretz, line 566: I wish we hadn't made these options as pointers, but seems it was done as part of #920

internal/worker_version_sets.go
```diff
@@ -130,6 +155,35 @@ type GetWorkerBuildIdCompatibilityOptions struct {
...
+	TaskQueues []string
+	// Reachability - The reachability this request is concerned with.
+	// Optional: defaults to all types of reachability
+	Reachability TaskReachability
+}
+
+type WorkerTaskReachability struct {
+	// Reachable - map of build IDs and their reachability information
+	// May contain an entry with [UnversionedBuildID] for an unversioned worker
+	Reachable map[string]*BuildIDReachability
```
cretz, line 173:
```suggestion
	BuildIDReachability map[string]*BuildIDReachability
```

Seems to match others

internal/worker_version_sets.go
```diff
@@ -130,6 +155,35 @@ type GetWorkerBuildIdCompatibilityOptions struct {
...
+	// Optional: defaults to all task queues
+	TaskQueues []string
+	// Reachability - The reachability this request is concerned with.
+	// Optional: defaults to all types of reachability
+	Reachability TaskReachability
+}
+
+type WorkerTaskReachability struct {
+	// Reachable - map of build IDs and their reachability information
+	// May contain an entry with [UnversionedBuildID] for an unversioned worker
```
cretz, line 172: > May contain an entry with [UnversionedBuildID] for an unversioned worker

Hrmm, I don't see this in Python or .NET. Any idea what happens on those SDKs in this situation? Are those also empty string keys?

internal/worker_version_sets.go
```diff
@@ -130,6 +155,35 @@ type GetWorkerBuildIdCompatibilityOptions struct {
...
+	Reachable map[string]*BuildIDReachability
+}
+
+type BuildIDReachability struct {
+	// TaskQueueReachable map of task queues and their reachability information
+	TaskQueueReachable map[string]*TaskQueueReachability
+}
+
+type TaskQueueReachability struct {
```
cretz, line 181: Some SDKs have a `UnretrievedTaskQueues` here. Not sure why TS doesn't.

internal/worker_version_sets.go
```diff
@@ -130,6 +155,35 @@ type GetWorkerBuildIdCompatibilityOptions struct {
...
+	// Reachability for a worker in a single task queue.
+	// If Reachability is empty, this worker is considered unreachable in this task queue.
+	Reachability []TaskReachability
```
cretz, line 184:
```suggestion
	TaskQueueReachability []TaskReachability
```

Seems to match others

cretz (commented):
Will definitely want @Sushisource's review here

internal/worker_version_sets.go
Quinn-With-Two-Ns: most likely yes

internal/worker_version_sets.go
Quinn-With-Two-Ns: hm yeah, I based the design on what TS did. I prefer what Python has with an explicate container for unfetched task queues. @Sushisource thoughts?

Sushisource: The more explicit version is preferable I think. In TS it works OK because the empty string is assigned a special type, but that's not workable here.

internal/worker_version_sets.go
```diff
@@ -47,6 +51,27 @@ const (
...
 )
 
+// TaskReachability specifies which category of tasks may reach a worker on a versioned task queue.
+// Used both in a reachability query and its response.
+type TaskReachability int
+
+const (
+	// TaskReachabilityNotFetched indicates that reachability for this task
+	// was not fetched
+	TaskReachabilityNotFetched TaskReachability = iota - 1
```
Sushisource, line 61: I'm not sure I'd put this in this enum - since it's not really a reachability condition. If you're gonna have the separate container for unfetched things, then you won't need it anyway.

client/client.go
bergundy: Still experimental and can change

internal/worker_version_sets.go
bergundy: Might be worth issuing multiple requests to populate all of the unretrieved task queues instead of having the user do it.

Sushisource: If we did that, then the API should've just done all that anyways to begin with. But we didn't, because it could be overloading sometimes. I think we should either respect that, or just not have the limitation.

Sushisource (approved):

internal/worker_version_sets.go
bergundy: The API has limits in place to prevent fanout for a single request, and we also have rate limits on visibility APIs like this one. Users may issue follow-up requests to fill the missing data and they'll stay within the system limits.

Whether or not we make multiple calls on behalf of users can be configurable but I think experience-wise we'd prefer defaulting to giving all the information and hiding this detail from users. That's of course my personal take on this, I understand you have a different one.

In any case, we can add this later as the feature matures and we get feedback.

PR #1249, opened 2023-09-26
Do not overwrite RPC headers

wxing1292:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_utils.go
+++ b/internal/internal_utils.go
@@ -77,7 +77,12 @@ func (cb *grpcContextBuilder) Build() (context.Context, context.CancelFunc) {
 		ctx = context.Background()
 	}
 	if cb.Headers != nil {
-		ctx = metadata.NewOutgoingContext(ctx, cb.Headers)
+		md, ok := metadata.FromOutgoingContext(ctx)
+		if !ok {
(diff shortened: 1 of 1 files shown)


internal/internal_utils.go
```diff
@@ -77,7 +77,12 @@ func (cb *grpcContextBuilder) Build() (context.Context, context.CancelFunc) {
 		ctx = context.Background()
 	}
 	if cb.Headers != nil {
-		ctx = metadata.NewOutgoingContext(ctx, cb.Headers)
+		md, ok := metadata.FromOutgoingContext(ctx)
+		if !ok {
+			ctx = metadata.NewOutgoingContext(ctx, cb.Headers)
+		} else {
+			ctx = metadata.NewOutgoingContext(ctx, metadata.Join(md, cb.Headers))
```
bergundy, line 84: Shouldn't we join in reverse order and allow context metadata to overwrite `cb.Headers`?

bergundy:
Thanks!

Can you provide a test that accompanies this change and a bit more details as to what this PR exactly does and why?

PR #1265, opened 2023-10-12
make compression codec more generic

nknordeen:
## What was changed
Modified the ZLib codec to accept reader/writer constructor

## Why?
Wanted to test a couple of different compression libraries, and thought the existing logic could be updated a bit to make it so people could just pass different libraries.

## Checklist

1. Closes 

2. How was this tested:
Used existing ZLib tests  to also test gzip

3. Any docs updates needed?


converter/codec.go
```diff
@@ -62,16 +62,41 @@ type ZlibCodecOptions struct {
...
+		},
+		newReader:    zlib.NewReader,
+		alwaysEncode: options.AlwaysEncode,
+		encodingType: "binary/zlib",
+	}
+}
+
+// NewCompressionCodec creates a PayloadCodec for use in NewCodecDataConverter
+// This can be used to pass compression encoders for different needs.
+func NewCompressionCodec(newWriter func(io.Writer) (io.WriteCloser, error), newReader func(io.Reader) (io.ReadCloser, error), encodingType string, alwaysEncode bool) PayloadCodec {
```
cretz, line 83: I am not sure there is value in a generic reader/writer encoder since the interface is so simple to implement. The zlib is more of a sample. See https://github.com/temporalio/samples-go/blob/main/snappycompress/data_converter.go for another sample.

nknordeen: Ya, samples-go is actually the reason I thought I'd put this together, cause I copied the snappy converter and then made a gzip one. 
That's fair though, not a big deal.  If others seem interested then I'll update to make it accept an Options param instead.

Issue #1266, opened 2023-10-12
`loggerWith.WithCallerSkip` should expand keyvals


Quinn-With-Two-Ns:
## Expected Behavior
`loggerWith.WithCallerSkip` should expand `keyvals`

## Actual Behavior
`loggerWith.WithCallerSkip` does not expand `keyvals`

```
func (l *withLogger) WithCallerSkip(depth int) Logger {
	if sl, ok := l.logger.(WithSkipCallers); ok {
		return newWithLogger(sl.WithCallerSkip(depth), l.keyvals)
	}
	return l
}
```

https://github.com/temporalio/sdk-go/pull/1219

PR #1267, opened 2023-10-12
loggerWith.WithCallerSkip expand keyvals

Quinn-With-Two-Ns:
Fix a regression introduced in https://github.com/temporalio/sdk-go/pull/1219 that caused loggers wrapped with `withLogger` to log key values as a slice 

another option is to remove `WithCallerSkip` from `withLogger`

closes https://github.com/temporalio/sdk-go/issues/1266

### DIFF
--- a/internal/log/memory_logger.go
+++ b/internal/log/memory_logger.go
@@ -31,21 +31,21 @@ import (
 	"go.temporal.io/sdk/log"
 )
 
-// MemoryLogger is Logger implementation that stores logs in memory (useful for testing). Use Lines() to get log lines.
-type MemoryLogger struct {
+// MemoryLoggerWithoutWith is Logger implementation that stores logs in memory (useful for testing). Use Lines() to get log lines.
+type MemoryLoggerWithoutWith struct {
 	lines         *[]string
 	globalKeyvals string
 }
 
-// NewMemoryLogger creates new instance of MemoryLogger.
-func NewMemoryLogger() *MemoryLogger {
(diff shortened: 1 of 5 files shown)


cretz (approved):

internal/log/memory_logger.go
```diff
@@ -55,28 +55,44 @@ func (l *MemoryLogger) println(level, msg string, keyvals []interface{}) {
...
+}
+
+// NewMemoryLogger creates new instance of MemoryLogger.
+func NewMemoryLogger() *MemoryLogger {
+	return &MemoryLogger{
+		NewMemoryLoggerWithoutWith(),
+	}
+}
+
 // With returns new logger the prepend every log entry with keyvals.
```
dandavison, line 93: (Maybe take this opportunity to fix the typos in this comment?)

internal/log/memory_logger.go
```diff
@@ -31,21 +31,21 @@ import (
 	"go.temporal.io/sdk/log"
 )
 
-// MemoryLogger is Logger implementation that stores logs in memory (useful for testing). Use Lines() to get log lines.
-type MemoryLogger struct {
+// MemoryLoggerWithoutWith is Logger implementation that stores logs in memory (useful for testing). Use Lines() to get log lines.
```
dandavison, line 34: It would be nice to have something in the docstring that explains the strange name.

```suggestion
// MemoryLoggerWithoutWith is a Logger implementation that stores logs in memory (useful for testing). Use Lines() to get log lines.
```

PR #1281, opened 2023-10-26
Prefix some errors with rule identifiers

cretz:
## What was changed

Add a couple of https://github.com/temporalio/rules references as error prefixes


Quinn-With-Two-Ns:
Just an FYI your missing a few panics for rule `TMPRL1100`, for example:
```
func (up *updateProtocol) requireState(action string, valid ...updateState) {
	for _, validState := range valid {
		if up.state == validState {
			return
		}
	}
	panicIllegalState(fmt.Sprintf("invalid action %q in update protocol %v", action, up))
}
```
should be tagged. Any `panicIllegalState` should be tagged with `TMPRL1100`

cretz:
@Quinn-With-Two-Ns et al - This is now ready for review

Quinn-With-Two-Ns (approved):

PR #1286, opened 2023-10-30
Add more marshaler options for ProtoJson converter

vbmithr:
## What was changed

I added options to be passed to marshaler code for Proto/JSON.

## Why?

Automatic OCaml deserialization of JSON-encoded Protobuf does not work if empty (nil) fields are not present.

## Checklist

1. Closes 

2. How was this tested:

I observed in my own code the expected result, i.e. JSON-encoded Protobuf now contains nil fields.

3. Any docs updates needed?


cretz (approved):
LGTM, will let @Quinn-With-Two-Ns look

Quinn-With-Two-Ns (approved):

cretz:
Thanks for the contribution!

PR #1288, opened 2023-11-07
fix typo in documentation

TheLeeeo:
Fix some typos in the example for handling errors

## What was changed
Fix some typos on the documentation about error handling

## Why?
This change was done for two reasons:
1. It was unsatisfying to see
2. If someone copies the example and uses it as a base for their own error handling the example should be working.

## Checklist
1. How was this tested:
This only includes some changes to comments so no executable code is affected. Therefor it is untested


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Thanks for the contribution, this whole error switch should be its own sample

PR #1291, opened 2023-11-08
Add context propagators to local activity executions through test environments

### LINKED ISSUE #1290: Context propagators aren't used when running local activities with the test environments
## Expected Behavior
When using the [ExecuteLocalActivity](https://github.com/temporalio/sdk-go/blob/0352634b45b1b81293c9bf3cec3bb033b82f841f/internal/workflow_testsuite.go#L196) method in `TestActivityEnvironment` struct, any context propagators configured for the test environment should be used.

## Actual Behavior
The context propagators are not used.

Write a test configured with a context propagator that invokes a local activity. Try to access the injected context values in that activity.

Example: https://gist.github.com/DeRauk/aaecab6eaca325d25e354661acae3711

## Specifications

  - Version:  v1.25.1
  - Platform: OSX

DeRauk:
## What was changed
Add context propagators when executing a local activity in a test environment.

## Why?
Fix for the linked issue

## Checklist

1. Closes #1290 

2. How was this tested:
A unit test was added

3. Any docs updates needed?
Nope

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -640,6 +640,7 @@ func (env *testWorkflowEnvironmentImpl) executeLocalActivity(
 		ActivityFn:   activityFn,
 		InputArgs:    args,
 		WorkflowInfo: env.workflowInfo,
+		Header:       env.header,
 	}
 	task := &localActivityTask{
 		activityID: "test-local-activity",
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns (commented):
Thank you for the contribution!

Quinn-With-Two-Ns (approved):

PR #1292, opened 2023-11-10
expose log tags for clients to use

nknordeen:
## What was changed
Making the log tags public

## Why?
I was copying and pasting these tags into my own projects because they correlate to very similar tags in our own logging libraries.  Exposing them might help us implement better loggers

## Checklist

Is there a different way people would prefer this to be refactored/handled?

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Quinn-With-Two-Ns:
HI @nknordeen, I don't think we would wan't to expose these tags as part of our public API we don't in any other SDK and could limit us in the future. If your using a logger supplied by the SDK then it should already have the relevant info set on it.

nknordeen:
That's fair.  Thanks for the feedback!

PR #1293, opened 2023-11-15
Multi-platform CI

### LINKED ISSUE #1098: Test CI on all platforms we support

Every platform we support should be tested in CI. See what we do in Python and .NET for an example.

cretz:
## What was changed

* Add build script with support for check, unit test, integration test (including w/ dev server), and coverage helpers
* Change CI to run on all three of ubuntu, mac, and windows
  * Coverage only on Ubuntu + Go 1.21
  * Docker compose only on Ubuntu + Go 1.21, dev server on all
* Change code here and there for multi-platform and dev server support on integration tests
* Delete Makefile

When ready to merge, will change the required build targets in GH branch protection

## Checklist

1. Closes #1098


internal/cmd/tools/copyright/licensegen.go
```diff
@@ -169,6 +169,7 @@ func isFileAutogenerated(_ string) bool {
 }
 
 func mustProcessPath(path string) bool {
+	path = filepath.ToSlash(path)
```
cretz, line 172: This was not Windows-safe before

internal/common/metrics/grpc_test.go
```diff
@@ -71,7 +71,6 @@ func TestGRPCInterceptor(t *testing.T) {
 	require.Len(t, timers, 1)
 	require.Equal(t, metrics.TemporalRequestLatency+"_my_suffix", timers[0].Name)
 	require.Equal(t, map[string]string{metrics.OperationTagName: "Check"}, timers[0].Tags)
-	require.Greater(t, timers[0].Value(), 0*time.Second)
```
cretz, line 74: On some platforms, this would actually appear to the metric collector as 0 seconds in flaky/fast cases

internal/internal_task_pollers_test.go
```diff
@@ -143,8 +130,6 @@ func TestWFTRacePrevention(t *testing.T) {
 	t.Log("Issue task1")
 	go func() { resultsChan <- poller.processWorkflowTask(&task1) }()
 
-	require.Never(t, tryWrite(completionChans[1]), 2*time.Second, 100*time.Millisecond,
-		"Should be no reader on the task1's completion channel as task0 holds the ctx lock")
```
cretz, line 147: The way that this poor `require.Never` is written, the callback can technically be invoked even after this is returned, which was causing race conditions with the closing of the channel later. Decided this part of the test isn't really needed to test the rest.

test/integration_test.go
```diff
@@ -2684,6 +2684,9 @@ func (ts *IntegrationTestSuite) TestDeterminismUpsertSearchAttributesConditional
 	options.SearchAttributes = map[string]interface{}{
 		"CustomKeywordField": "unset",
 	}
+	// TODO(cretz): There is a bug with search attribute names on standard
+	// visibility with eager workflow start
+	options.EnableEagerStart = false
```
cretz, line 2689: https://github.com/temporalio/temporal/issues/5119 (and https://github.com/temporalio/temporal/pull/5124)

test/test_utils_test.go
```diff
@@ -226,8 +227,9 @@ func (ts *ConfigAndClientSuiteBase) InitConfigAndNamespace() error {
 		return err
 	}
 	if ts.config.ShouldRegisterNamespace {
-		err = ts.registerNamespace()
-		if err != nil {
+		if err = ts.registerNamespace(); err != nil {
+			return err
+		} else if err = ts.ensureSearchAttributes(); err != nil {
```
cretz, line 232: For dev server, the `CustomKeywordField` search attribute isn't present like it is in docker compose

test/integration_test.go
```diff
@@ -2685,6 +2686,9 @@ func (ts *IntegrationTestSuite) TestDeterminismUpsertSearchAttributesConditional
 	options.SearchAttributes = map[string]interface{}{
 		"CustomKeywordField": "unset",
 	}
+	// TODO(cretz): There is a bug with search attribute names on standard
```
Quinn-With-Two-Ns, line 2689: Can we replace with the issue link

Quinn-With-Two-Ns (approved):

cretz:
I have to go mod tidy some stuff and maybe fix other things w/ remove-gogo merge before CI passes, but won't merge w/out re-review if I have to alter anything serious.

internal/cmd/build/main.go
```diff
@@ -0,0 +1,314 @@
...
+	thisDir string
+	rootDir string
+}
+
+func newBuilder() *builder {
+	var b builder
+	// Find the root directory from this directory
+	_, thisFile, _, _ := runtime.Caller(0)
+	b.thisDir = filepath.Join(thisFile, "..")
+	b.rootDir = filepath.Join(b.thisDir, "../../../")
```
tdeebswihart, line 67: You could call into `git rev-parse --show-toplevel` instead. We know it's a git repo

cretz: I think this code is much clearer than making a git call. It's really easy to just get a relative directory of a current file.

Issue #1289, opened 2023-11-07
[Feature Request] Tag workflow_task_execution_failed with error type


Quinn-With-Two-Ns:
see https://github.com/temporalio/features/issues/231

PR #1295, opened 2023-11-20
Tag workflow_task_execution_failed with error type

Quinn-With-Two-Ns:
Tag workflow_task_execution_failed with error type. Change some panics to illegal state panics so they are captured as non determinism.

closes https://github.com/temporalio/sdk-go/issues/1289


cretz:
Let's make sure we're clear in release notes here so people aren't surprised when this breaks out into two metric lines in their Prometheus metrics.

Issue #1287, opened 2023-11-03
Error "getState: illegal access from outside of workflow context" after sdk update  1.22.1 -> 1.25.1


vmalyavin:
[You can view full workflow code of reproduction in repo](https://github.com/vmalyavin/temporalissue)

We have a simple workflow cycle with select
- pereodic polling(payment status from activity every 2-3min)
- workflow-cancel signal receive
- timeout
```go
package testflow

import (
	"time"

	"go.temporal.io/sdk/workflow"

	"temporalissue/models"
	"temporalissue/workflows/testflow/signals"
)

func PollPaymentStatus20MinOrReceiveCancel(
	ctx workflow.Context, order *models.Order,
) models.PaymentStatus {
	var status models.PaymentStatus

	// cycle over timer - 20min
	itsOver := false
	overTimer := workflow.NewTimer(ctx, 20*time.Minute)
	overCallback := func(_ workflow.Future) { itsOver = true }

	for !itsOver {
		pollTimer, pollCallback := pollPaymentStatus(
			ctx, 3*time.Minute, &status, order,
		)
		selector := workflow.NewSelector(ctx).
			AddFuture(pollTimer, pollCallback).
			AddFuture(overTimer, overCallback).
			AddReceive(signals.CancellationReceive(ctx, order))
		selector.Select(ctx)

		// end cycle if paymentstatus is changed from new; or order is cancelled
		if status == models.PaymentStatusNew || order.Status == models.OrderStatusCancel {
			break
		}
	}

	return status
}
```

then after end of the cycle we 
- complete workflow

or  
- cancel it(by signal) and poll payment status for 10min after cancel:
```go
func PollPaymentStatus10Min(
	ctx workflow.Context, order *models.Order,
) models.PaymentStatus {
	var status models.PaymentStatus

	// cycle over timer - 10min
	itsOver := false
	overTimer := workflow.NewTimer(ctx, 10*time.Minute)
	overCallback := func(_ workflow.Future) { itsOver = true }

	for !itsOver {
		pollTimer, pollCallback := pollPaymentStatus(
			ctx, 2*time.Minute, &status, order,
		)
		selector := workflow.NewSelector(ctx).
			AddFuture(pollTimer, pollCallback).
			AddFuture(overTimer, overCallback)
		// no cancel receive
		selector.Select(ctx)

		// end cycle if paymentstatus is changed from new
		if status != models.PaymentStatusNew {
			break
		}
	}

	return status
}
```
[...]

PR #1296, opened 2023-11-20
Don't run workflowCancelHandler in test harness

Quinn-With-Two-Ns:
Don't run workflowCancelHandler in test harness on completed workflows. If a workflow is completed nothing else in the workflow function should run.

closes https://github.com/temporalio/sdk-go/issues/1287

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -877,16 +877,13 @@ func (env *testWorkflowEnvironmentImpl) Complete(result *commonpb.Payloads, err 
 		return
 	}
 	env.workflowDef.Close()
-	var canceledErr *CanceledError
-	if errors.As(err, &canceledErr) && env.workflowCancelHandler != nil {
-		env.workflowCancelHandler()
-	}
 
 	dc := env.GetDataConverter()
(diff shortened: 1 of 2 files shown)


internal/internal_workflow_testsuite.go
```diff
@@ -875,16 +875,13 @@ func (env *testWorkflowEnvironmentImpl) Complete(result *commonpb.Payloads, err
 		return
 	}
 	env.workflowDef.Close()
-	var canceledErr *CanceledError
-	if errors.As(err, &canceledErr) && env.workflowCancelHandler != nil {
-		env.workflowCancelHandler()
-	}
```
cretz, line 881: Can you help me understand this change? Is it just because `As` checks errors at every depth and if so, maybe we should change to just checking the exact error type instead of hierarchy? Can you confirm that there is no other internal way for a cancel to be sent to a test workflow than via `TestWorkflowEnvironment.CancelWorkflow()` (which appears to still invoke this callback)/

Quinn-With-Two-Ns: This change is because canceling a  workflow that has already returned is not valid. I the workflow function has already returned what valid effect can the cancel call have?

Issue #1248, opened 2023-09-26
Enable TCP Keep alive


Quinn-With-Two-Ns:
Enable TCP Keep alive by default

https://github.com/temporalio/features/issues/123

PR #1299, opened 2023-11-22
Enable keep alive by default

Quinn-With-Two-Ns:
Enabled keep alive by default.

https://github.com/temporalio/sdk-go/issues/1248

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -491,23 +491,26 @@ type (
 		// This value only used when TLS is nil.
 		Authority string
 
-		// Enables keep alive ping from client to the server, which can help detect abruptly closed connections faster.
-		EnableKeepAliveCheck bool
+		// Disable keep alive ping from client to the server.
+		DisableKeepAliveCheck bool
 
 		// After a duration of this time if the client doesn't see any activity it
 		// pings the server to see if the transport is still alive.
 		// If set below 10s, a minimum value of 10s will be used instead.
+		// default: 15s
 		KeepAliveTime time.Duration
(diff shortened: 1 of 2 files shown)


internal/client.go
```diff
@@ -492,23 +492,37 @@ type (
 		Authority string
 
 		// Enables keep alive ping from client to the server, which can help detect abruptly closed connections faster.
+		//
+		// Deprecated: Keep alive is now enabled by default, to disable set DisableKeepAliveCheck.
```
cretz, line 496: I'd be tempted to literally break compatibility and just remove this field. There's a difference between "deprecated" as if it is still used and "unused field". Thoughts?

cretz (commented):
Pedantic: May want to remove "TCP" from issue title (this is gRPC keep alive which is HTTP/2 keep alive using HTTP/2 pings, TCP keep alive is a separate thing).

internal/client.go
Quinn-With-Two-Ns: Yeah I thought about removing it, my only issue with removing it is I think it will only break the people who are not effected by this change. Since if you wanted it false you probably just didn't set it. I am fine removing it though if you want to

cretz: Let's bring up with team if it can wait a day or so

Quinn-With-Two-Ns: Fine with me

cretz (approved):
:+1: Reminder to add a big :boom: message to the release notes :-)

PR #1303, opened 2023-11-27
Make remoteDataConverter implement PayloadCodec

dandavison:
## What was changed
- Split `remoteDataConverter` into separate `PayloadCodec` and `DataConverter`

## Why?
- This change is used to fix bugs in and simplify the `temporal` CLI: https://github.com/temporalio/cli/pull/384

## Checklist

1. How was this tested: https://github.com/temporalio/cli/pull/384

2. Any docs updates needed?
No. The change is backwards compatible.

### DIFF
--- a/converter/codec.go
+++ b/converter/codec.go
@@ -347,27 +347,44 @@ func NewPayloadCodecHTTPHandler(e ...PayloadCodec) http.Handler {
 	return &codecHTTPHandler{codecs: e}
 }
 
-// RemoteDataConverterOptions are options for NewRemoteDataConverter.
+// RemotePayloadCodecOptions are options for RemotePayloadCodec.
 // Client is optional.
+type RemotePayloadCodecOptions struct {
+	Endpoint      string
+	ModifyRequest func(*http.Request) error
+	Client        http.Client
+}
+
+type remotePayloadCodec struct {
+	options RemotePayloadCodecOptions
+}
+
+// NewRemotePayloadCodec creates a PayloadCodec using the remote endpoint configured by RemotePayloadCodecOptions.
+func NewRemotePayloadCodec(options RemotePayloadCodecOptions) PayloadCodec {
+	return &remotePayloadCodec{options}
+}
+
+// Fields Endpoint, ModifyRequest, Client of RemotePayloadCodecOptions are also
+// exposed here in RemoteDataConverterOptions for backwards compatibility.
+
+// RemoteDataConverterOptions are options for NewRemoteDataConverter.
 type RemoteDataConverterOptions struct {
 	Endpoint      string
 	ModifyRequest func(*http.Request) error
 	Client        http.Client
 }
 
-// remoteDataConverter is a DataConverter that wraps an underlying data
-// converter and uses a remote codec to handle encoding/decoding.
 type remoteDataConverter struct {
-	parent  DataConverter
-	options RemoteDataConverterOptions
+	parent       DataConverter
+	payloadCodec PayloadCodec
 }
 
 // NewRemoteDataConverter wraps the given parent DataConverter and performs
 // encoding/decoding on the payload via the remote endpoint.
 func NewRemoteDataConverter(parent DataConverter, options RemoteDataConverterOptions) DataConverter {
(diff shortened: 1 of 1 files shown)


cretz:
I wonder if it'd be clearer to just have a `NewRemoteCodec`

converter/codec.go
```diff
@@ -364,7 +364,7 @@ type remoteDataConverter struct {
 
 // NewRemoteDataConverter wraps the given parent DataConverter and performs
 // encoding/decoding on the payload via the remote endpoint.
-func NewRemoteDataConverter(parent DataConverter, options RemoteDataConverterOptions) DataConverter {
+func NewRemoteDataConverter(parent DataConverter, options RemoteDataConverterOptions) *remoteDataConverter {
```
Quinn-With-Two-Ns, line 367: Generally you don't return an unexported type from an exported function

dandavison: OK, I'm not a Go expert -- I tentatively thought that it was reasonable in this case as a way of communicating to the caller that they can choose which interface they want to interpret the return value as implementing. But also, I was trying to make as minimal changes as possible in order to solicit initial feedback on the overall approach. I'll export the type if that's the right thing to do.

Quinn-With-Two-Ns: > I tentatively thought that it was reasonable in this case as a way of communicating to the caller that they can choose which interface they want to interpret the return value as implementing

The type does not need to be unexported for that to be the case. Unexported types are harder to work with and I believe the methods it implements won't show up on the docs so it can be hard to tell what it implements.

dandavison: Yep. I've exported it, and also changed the docstring to say "RemoteDataConverter is a DataConverter and PayloadCodec ..." 029d5aa97357d039d98c07ea6348debcf8aa260b

dandavison: What I haven't done yet is look into refactoring it cleanly as a standalone `PayloadCodec` used by a `DataConverter`, because I wanted to get an OK for this approach without making too many code changes. Not sure if you feel that's necessary or a future nice-to-have.

dandavison: @cretz if I do that do you think you would use it in the next version of the CLI? Do you envisage using a gRPC interceptor for payload encoding/decoding alone, and not for ser/de? (I'm thinking yes, but wanted to check).

converter/codec.go
```diff
@@ -342,27 +342,91 @@ func NewPayloadCodecHTTPHandler(e ...PayloadCodec) http.Handler {
...
 	Client        http.Client
 }
 
-// remoteDataConverter is a DataConverter that wraps an underlying data
-// converter and uses a remote codec to handle encoding/decoding.
+type remotePayloadCodec struct {
+	options RemotePayloadCodecOptions
+}
+
+func NewRemotePayloadCodec(options RemotePayloadCodecOptions) PayloadCodec {
```
cretz, line 357: All public Go API needs Godoc

converter/codec.go
```diff
@@ -342,27 +342,91 @@ func NewPayloadCodecHTTPHandler(e ...PayloadCodec) http.Handler {
...
+		return resultPayloads.Payloads, nil
+	}
+
+	message, _ := io.ReadAll(response.Body)
+	return payloads, fmt.Errorf("%s: %s", http.StatusText(response.StatusCode), message)
+}
+
+// RemoteDataConverterOptions are options for NewRemoteDataConverter.
+type RemoteDataConverterOptions struct {
+	RemotePayloadCodecOptions
```
cretz, line 413: This is a backwards incompatible change. I know it may seem strange (especially w/ code reuse in other languages), but I think the fields need to be copied in here same as they are in `RemotePayloadCodecOptions`. In addition to this being a backwards incompatible change, a user shouldn't have to know that we are reusing remote codec code underneath.

dandavison: Good catch @cretz! Still getting used to Go's embedded fields etc. Fixed: the fields are now present in both structs.

converter/codec.go
dandavison: Fixed, here and on `Encode`/`Decode`

Quinn-With-Two-Ns (approved):

Issue #1297, opened 2023-11-21
Update handle needs to be invoked immediately when registered with pending updates


Quinn-With-Two-Ns:
Update handle needs to be invoked immediately when registered with pending updates. If we do not it is possible a workflow function may return before the update handle is called, preventing the workflow from properly handling the update.

For example:
```Go
inFlightUpdates := 0
SetUpdateHandlerWithOptions(
	ctx,
	updateName,
	func() (T, error) {
                 inFlightUpdates = inFlightUpdates + 1
		// << change Workflow state >>

		return <result>, nil
	},
)

Await(inFlightUpdates == 0) // until Workflow is done
return
```
here the update handle won't be invoked until after the return so the `Await` won't work as intended

Issue #1235, opened 2023-09-13
Defers in `workflow.Go` should be run deterministically and in sequence on cache eviction.


Quinn-With-Two-Ns:
## Expected Behavior
Defers in `workflow.Go` should be run deterministically and in sequence on cache eviction.

## Actual Behavior

Defers in `workflow.Go` should be run in parallel.

## Steps to Reproduce the Problem
See https://github.com/temporalio/sdk-go/issues/1209

while https://github.com/temporalio/sdk-go/pull/1210 fixed the main consequence of the defers running in parallel there are still some edge cases where running the defers in parallel could cause problems for users.

When we fix this we need to make sure we think about how this will effect trying to evict a stuck workflow and make sure we don't block the cache eviction on this case

PR #1306, opened 2023-11-28
Handle updates immediately when registered

Quinn-With-Two-Ns:
Handle updates immediately when registered and give priority to update request over the main workflow coroutine if a handler is registered. Now when no update handlers have been registered instead of yielding the update coroutine we queue the update request and only create the coroutine when the handler is registered or there is no more coroutines to run. All update coroutines are now created at the front of the scheduler so they run before the root coroutine.

This change is a major change to how the Go SDK handles updates, it should be backwards compatible in the sense old workflows should continue to run, but new workflows will behave observably different to users.

closes https://github.com/temporalio/sdk-go/issues/1297


test/workflow_test.go
```diff
@@ -2128,6 +2128,63 @@ func (w *Workflows) PanicOnSignal(ctx workflow.Context) error {
...
 }
 
+func (w *Workflows) WaitOnUpdate(ctx workflow.Context) (int, error) {
+	inflightUpdates := 0
+	updatesRan := 0
+	sleepHandle := func(ctx workflow.Context) error {
+		inflightUpdates++
+		updatesRan++
+		err := workflow.Sleep(ctx, time.Second)
+		inflightUpdates--
```
Quinn-With-Two-Ns, line 2138: avoiding using defer here because https://github.com/temporalio/sdk-go/issues/1235

internal/internal_event_handlers.go
```diff
@@ -881,6 +886,41 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() (*commonpb.Payloads, erro
...
+func (wc *workflowEnvironmentImpl) DrainUnhandledUpdates() bool {
+	rerun := false
+	// Check if any blocked updates remain when we have no more coroutines to run and let them run so they are rejected.
+	// Generally iterating a map in workflow code is bad because it is non deterministic
+	// this case is fine since all these update handles will be rejected and not recorded in history.
+	for name, us := range wc.blockedUpdates {
+		for _, u := range us {
+			u()
+		}
+		rerun = true
```
dandavison, line 918: Is it definitely true that `us` was non-empty?

internal/internal_event_handlers.go
```diff
@@ -881,6 +886,41 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() (*commonpb.Payloads, erro
...
+		for _, update := range blockedUpdates {
+			update()
+		}
+		delete(wc.blockedUpdates, name)
+	}
+	return updatesHandled
+}
+
+func (wc *workflowEnvironmentImpl) DrainUnhandledUpdates() bool {
+	rerun := false
```
dandavison, line 910: nit: I think this variable name is reflecting the way it is used by a caller, rather than its role here. Suggest `anyExecuted` or something like that.

internal/internal_workflow.go
```diff
@@ -1095,8 +1104,10 @@ func (d *dispatcherImpl) ExecuteUntilAllBlocked(deadlockDetectionTimeout time.Du
 		d.mutex.Unlock()
 	}()
 	allBlocked := false
+	d.coroutines = append(d.newEagerCoroutines, d.coroutines...)
+	d.newEagerCoroutines = nil
 	// Keep executing until at least one goroutine made some progress
-	for !allBlocked {
+	for !allBlocked || d.allBlockedCallback() {
```
dandavison, line 1110: nit: this reads a bit oddly ("If not allblocked or allblocked"). I'm trying to think of a better name for the callback. Maybe something like `tryAdvance()`?

internal/internal_event_handlers.go
```diff
@@ -881,6 +886,41 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() (*commonpb.Payloads, erro
...
+}
+
+func (wc *workflowEnvironmentImpl) QueueUpdate(name string, f func()) {
+	wc.blockedUpdates[name] = append(wc.blockedUpdates[name], f)
+}
+
+func (wc *workflowEnvironmentImpl) HandleUpdates(name string) bool {
+	updatesHandled := false
+	if blockedUpdates, ok := wc.blockedUpdates[name]; wc.sdkFlags.tryUse(SDKPriorityUpdateHandling, !wc.isReplay) && ok {
+		updatesHandled = true
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: No reason

internal/internal_event_handlers.go
Quinn-With-Two-Ns: Yes, but i'll admit that isn't  obvious. We could move the  line into the range block to cover that case.

internal/internal_update_test.go
```diff
@@ -169,7 +171,6 @@ func TestUpdateValidatorFnValidation(t *testing.T) {
 }
 
 func TestDefaultUpdateHandler(t *testing.T) {
-	t.Parallel()
```
Sushisource, line 172: Why can't this be parallel any more?

internal/internal_event_handlers.go
```diff
@@ -881,6 +886,41 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() (*commonpb.Payloads, erro
...
+		delete(wc.blockedUpdates, name)
+	}
+	return updatesHandled
+}
+
+func (wc *workflowEnvironmentImpl) DrainUnhandledUpdates() bool {
+	rerun := false
+	// Check if any blocked updates remain when we have no more coroutines to run and let them run so they are rejected.
+	// Generally iterating a map in workflow code is bad because it is non deterministic
+	// this case is fine since all these update handles will be rejected and not recorded in history.
```
Sushisource, line 913: It wasn't immediately clear to me that rejection would happen because without the context of what the functions in `blockedUpdates` really are (wrappers around the update that will also reject if no one registered, not just the user handler) it seems like this might be executing handlers directly.

I'm not sure there's really any great way to deal with that, though.

Sushisource (commented):
This makes sense to me overall. Just a few comments but nothing major.

internal/internal_update_test.go
Quinn-With-Two-Ns: The workflow environment  this test shares isn't meant to be multithreaded

internal/internal_workflow.go
```diff
@@ -1123,8 +1134,10 @@ func (d *dispatcherImpl) ExecuteUntilAllBlocked(deadlockDetectionTimeout time.Du
 			}
 		}
 		// Set allBlocked to false if new coroutines where created
-		allBlocked = allBlocked && lastSequence == d.sequence
-		if len(d.coroutines) == 0 {
+		allBlocked = allBlocked && lastSequence == d.sequence && len(d.newEagerCoroutines) == 0
+		d.coroutines = append(d.newEagerCoroutines, d.coroutines...)
+		d.newEagerCoroutines = nil
```
dandavison, line 1139: This looks like it will be copying the contents of the `d.coroutines` array every time, even when `d.newEagerCoroutines` is empty. How about

```suggestion
		if len(d.newEagerCoroutines) > 0 {
			d.coroutines = append(d.newEagerCoroutines, d.coroutines...)
			d.newEagerCoroutines = nil
			allBlocked = false
		} else {
			allBlocked = allBlocked && lastSequence == d.sequence
		}
```

internal/internal_event_handlers.go
Quinn-With-Two-Ns: Yeah I can document more  clearly `blockedUpdates` doesn't contain the handlers directly, but spawns a coroutine  to handle them.

>It's also not immediately obvious that it's guaranteed that all updates will be rejected. I guess this is the case because, if the updates were registered, then they would have had their handlers run before all the coroutines block.

Yeah it is up to the handler to reject, I'd rather keep  the rejection logic all in the handler rather then sprinkle it around.

Quinn-With-Two-Ns: >That being the case - perhaps it would make more sense for this function to be called RejectUnhandledUpdates

I didn't name it that because I thought if I added a dynamic update handle then they would be handled here, we don't support dynamic anything in Go yet so maybe that isn't a good reason.

internal/internal_workflow.go
```diff
@@ -1123,8 +1134,10 @@ func (d *dispatcherImpl) ExecuteUntilAllBlocked(deadlockDetectionTimeout time.Du
 			}
 		}
```
dandavison, line 1135: In Typescript, the Update handler preempts _immediately_ the code that was executing the  `setHandler` call, whereas here we're going to let all outstanding coroutines run until they're blocked/exit, and only then turn to the queued eager Update handler. I think that's going to result in (subtly) different semantics, e.g. related to whether or not an Update handler will see changes made by other concurrent routines. I'm not necessarily suggesting a change; that may fall well within expected inter-SDK variability. But, it would be technically possible I think to execute the pending eager coroutines inside the inner for loop, rather than waiting for that loop to exit (and I suspect it would bring Typescript and Go semantics closer together, but I haven't done the work to come up with an example.)

Quinn-With-Two-Ns: Yes no coroutine in the Go SDK executes immediately when created. The same is true if users create  there own coroutine or resolve a future. The runtimes schedule work differently. I agree it is probably technically possible to make the updates more  "eager", but it isn't required to get the behavior WRT updates running before the main workflow function and allowing updates to run after being registered before the main function is finished.

dandavison: > It isn't required to get the behavior WRT updates running before the main workflow function and allowing updates to run after being registered before the main function is finished.

Hm, I think that's not true. Our requirements for signal and update hold that pending signals and updates must get an opportunity to influence the workflow return value, and the params passed to CAN. But the workflow return value may be computed by one of the pending coroutines in the inner loop. Similarly, one of the pending coroutines in the inner loop could be a signal or update handler that CANs.

I'm thinking that we can meet that requirement by ensuring that, when we yield back to the executor loop in `setUpdateHandler`, the new handler coroutine is the very next coroutine to be handled.

Here's a failing test that demonstrates what I'm saying. I think that because your test workflows are using `workflow.Await` before the end, they are passing, but that we actually have a stronger requirement.

```go
func (w *Workflows) UpdateSetHandlerOnly(ctx workflow.Context) (int, error) {
	updatesRan := 0
	updateHandle := func(ctx workflow.Context) error {
		updatesRan++
		return nil
	}
	workflow.SetUpdateHandler(ctx, "update", updateHandle)
	return updatesRan, nil
}
func (ts *IntegrationTestSuite) TestUpdateAlwaysHandled() {
	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()
	options := ts.startWorkflowOptions("test-update-always-handled")
	options.StartDelay = time.Hour
	run, err := ts.client.ExecuteWorkflow(ctx, options, ts.workflows.UpdateSetHandlerOnly)
	ts.NoError(err)
	// Send an update before the first workflow task
	_, err = ts.client.UpdateWorkflow(ctx, run.GetID(), run.GetRunID(), "update")
	ts.NoError(err)
	// ts.NoError(updateHandle.Get(ctx, nil))
	var result int
	ts.NoError(run.Get(ctx, &result))
	ts.Equal(0, result)
}
```

The equivalent test in Typescript passes: https://github.com/dandavison/temporalio-sdk-typescript/blob/125241bbd3a1b3cf29e59d47eaa0e979ad3d7e69/packages/test/src/test-integration-update.ts#L217-L243

dandavison (changes requested):
This is great and it's been teaching about sdk-go. I'm currently thinking that we do need to implement maximally eager scheduling in order to uphold the requirements we've agreed upon (explained in comment).

internal/internal_workflow.go
Quinn-With-Two-Ns: This test passes with my changes

test/integration_test.go
```diff
@@ -2423,6 +2424,143 @@ func (ts *IntegrationTestSuite) TestMaxConcurrentSessionExecutionSizeWithRecreat
...
+	// Get the result
+	var result int
+	ts.NoError(run.Get(ctx, &result))
+	ts.Equal(0, result)
+}
+
+func (ts *IntegrationTestSuite) TestWaitOnUpdate() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	// We want to start a single long-running activity in a session
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Oh yeah this is not relevant

internal/internal_workflow.go
Quinn-With-Two-Ns: No worries, I don't see how CAN is different then returning any normal value from the workflow function

internal/internal_flags.go
```diff
@@ -44,7 +44,10 @@ const (
 	// a workflow task response's command set to order messages with respect to
 	// commands.
 	SDKFlagProtocolMessageCommand = 3
-	SDKFlagUnknown                = math.MaxUint32
+	// SDKPriorityUpdateHandling will cause update request to be handled before the main workflow method.
+	// It will also cause the SDK to immediately handle updates when a handler is registered.
+	SDKPriorityUpdateHandling = 4
```
cretz, line 49: Updates are experimental, do we have to put this on everyone's history forever now to preserve compatibility for experimental users?

cretz (approved):

internal/internal_flags.go
Quinn-With-Two-Ns: We can deprecate these at some point, like when update goes GA or after a few releases

cretz: Can we just release saying in some rare cases updates are incompatible with last release since they are experimental?

cretz: I think we should be ok breaking workflows using experimental updates instead of adding a flag for everyone. With this change, you can never remove this flag enumerate. So we made something experimental so we could make changes, but we're not allowing ourselves to make changes, so the code is forever burdened.

cretz: Yeah, that's an acceptable compromise. I understand the number 4 is now reserved forever, but ok. I personally prefer to break now, but your choice.

dandavison:
Great, that second modification to the loop was pretty clean. So my only remaining question here is about backwards compatibility: suppose that we decide that we are interested in guaranteeing that updates execute in the order supplied by sdk-core, rather than the order of handler registration. If we were to release this now, would that make such work more difficult than it would be without having released this change?

dandavison (approved):
LGTM!

internal/internal_flags.go
cretz: Sorry to unresolve this...can we make a new issue or something so we don't forget to remove this flag some day instead of setting for everyone forever?

cretz (approved):

internal/internal_flags.go
Quinn-With-Two-Ns: https://github.com/temporalio/sdk-go/issues/1315

PR #1308, opened 2023-11-28
Bump api-go version to fix Payload marshaling

tdeebswihart:
## What was changed
I upgraded our api-go dependency

## Why?
Marshaling for binary/null payloads was completely broken, and while there Chad and I decided that shorthand marshaling should be opt-in rather than opt-out.


tdeebswihart:
You can see from https://github.com/temporalio/features/pull/375 that the feature tests pass when that repo has been updated

cretz (approved):
Unless needed immediately, can also wait until API repo is tagged

tdeebswihart:
API repo is tagged, I'll set this to merge once tests pass

PR #1311, opened 2023-11-30
Run cloud tests in series to avoid wfID conflict

Quinn-With-Two-Ns:
Run cloud tests in series to avoid workflow ID conflict. When running parallel integration tests against the same cloud namespace it is possible if the same test runs at the same time the workflow IDs will conflict causing test failure.

example:
https://github.com/temporalio/sdk-go/actions/runs/7049993619/job/19189705118

Note: this isn't a perfect, the only way to properly fix is to run in separate namespaces or modify all the tests to add a UUID to every workflow and child workflow ID.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,8 +88,8 @@ jobs:
     strategy:
       matrix:
         go-version: ["1.20", "1.21"]
-        # Try to avoid running tests in parallel to avoid workflow ID conflict
-        max-parallel: 1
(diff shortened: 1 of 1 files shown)


cretz (approved):
> or modify all the tests to add a UUID

This only runs the `TestBasic` test, so you can probably just modify that one test

Quinn-With-Two-Ns:
>This only runs the TestBasic test, so you can probably just modify that one test
If we see this failure any more I'll do that

PR #1319, opened 2023-12-08
Allow client to suggest delay until next retry

ast2023:
## What was changed
A new error type was introduced, which allows requesting next retry interval from an activity.

## Why?
Allow developers to adjust retry interval on specific errors.  

## Checklist
1. Closes 1858

2. How was this tested:
Created a workflow/activity. Activity would return the new error with the retry interval. Verified that the value was received by the server.

3. Any docs updates needed?
The new error should be documented in the developer's guide.


temporal/error.go
```diff
@@ -174,6 +175,15 @@ func NewApplicationError(message, errType string, details ...interface{}) error
 	return internal.NewApplicationError(message, errType, false, nil, details...)
 }
 
+// NewApplicationErrorWithRetryDelay creates new instance of retryable *ApplicationErrorWithRetryRetryDelay with message,
+// type, retry delay and optional details.
+// Use ApplicationErrorWithRetryDelay to indicate that the activity needs to be restarted after retryDelay interval.
+// errType can be used to allow control from Workflow if error is retryable or not. Add the same type in
+// to RetryPolicy.NonRetryableErrorTypes to avoid retrying of particular error types.
+func NewApplicationErrorWithRetryDelay(message, errType string, retryDelay time.Duration, details ...interface{}) error {
+	return internal.NewApplicationErrorWithRetryDelay(message, errType, retryDelay, nil, details...)
```
ast2023, line 184: This logic is similar to on in NewApplicationError. Should not we receive the error from the user on NewApplicationErrorWithRetryDelay call?

internal/error.go
```diff
@@ -129,6 +129,16 @@ type (
 		details      converter.EncodedValues
 	}
 
+	ApplicationErrorWithRetryDelay struct {
```
bergundy, line 132: I wouldn't put the retry on an error class if we're not serializing it as part of ApplicationFailure.
I still think we want to add the field in ApplicationFailureInfo even if other disagree.

If we must go with the alternative approach, you would probably want to export a `SetNextRetryDelayOnFailure` method on the `activity` package or maybe something more general that could set other attributes as well (although, even though other seem to think so, I disagree that we need to support setting other activity attributes **on failure**).

Quinn-With-Two-Ns: +1 to `SetNextRetryDelayOnFailure` method on the activity package

ast2023: @bergundy thank you for the recommendation.

> I wouldn't put the retry on an error class if we're not serializing it as part of ApplicationFailure.
> I still think we want to add the field in ApplicationFailureInfo even if other disagree.

ApplicationErrorWithRetryDelay is not an error class. You cannot restore one from a failure. ApplicationErrorWithRetryDelay is an entity which implements Error interface and Delay interface.
You see what I did, I labeled a tail as a leg and now claim that a dog has 5 legs 😄 
But seriously, please help me better understand the reason you think we should not put it on the error.
Do you mean that activity to SDK protocol should define SDK to the server protocol?

> If we must go with the alternative approach, you would probably want to export a SetNextRetryDelayOnFailure method on the activity package or maybe something more general that could set other attributes as well (although, even though other seem to think so, I disagree that we need to support setting other activity attributes on failure).

I let user return 2 things which are related in one call: ApplicationError and RetryDelay. I agree there is some issue that what the call produces is of type error, but I did not want to change the type of what Executor expects for backward computability. Regarding other attributes, they may be set by the user in the way which SDK team finds appropriate. They do not have to follow this if the attribute does not relate to the failure.

cretz: > But seriously, please help me better understand the reason you think we should not put it on the error.

IMO the same reason as was suggested in the discussions on this topic that it should not be on the error in the protos - it should not be a product of the error but a settable property anywhere in the activity for when an error occurs.

+1 to SetNextRetryDelayOnFailure method on the activity package

Sushisource: As mentioned when we met - I think it can actually go both places and that makes good sense. I think it makes sense to have an API to attach a next delay to any error you're returning, and also the general `SetNextRetryDelayOnFailure` API. Both can be quite useful depending on the structure of your code, and both can end up in the same place in the actual API response.

Quinn-With-Two-Ns: If we do want to attach it to the error I think we should have something more general then forcing users to use `ApplicationErrorWithRetryDelay`. I would suggest we check an interface like `retryDelay()` and if the error implements that interface we use it. If we allow both I am not sure which takes precedent.

Sushisource: +1 - in my mind you would use an interface just like you've said, which you could augment any error with something that wraps the original and implements that interface. As for precedence, I imagine the per-error one would take precedence over the general "next delay" call, since the former is more specific/granular than the latter.

cretz: It was a mistake by the original developers to make `temporal.NewApplicationError` return `error` instead of `*ApplicationError` and thus we can't really attach things easily to it. Luckily our other SDKs don't suffer from this. I am not a fan of having two ways personally, but I don't have strong objection so long as the easier, simpler, clearer way everyone agrees (`activity.SetNextRetryDelayOnFailure`) exists.

I don't suspect there will be many users writing errors conforming to that interface. Looking at the failure converter, using the interface approach, how can I set error type, some details, and next retry delay in my error? Can I see a snippet of code of how that might look?

Quinn-With-Two-Ns: >Looking at the failure converter, using the interface approach

The failure converter is not relevant. If you look at the current `extractActivityRequestsFrom` it is called on the error returned from the activity. I am suggesting that function should check for an interface rather then a specific type. I don't want to give our types any special behaviour that users types can't have.

cretz: Can you give an example of returning an error from an activity with this new API with error type, some details, and next retry delay set on the application error? That way I can at least visualize what is meant with the interface approach.

internal/error.go
```diff
@@ -127,6 +135,17 @@ type (
 		nonRetryable bool
 		cause        error
 		details      converter.EncodedValues
+		extra        ExtraInformation
+	}
+
+	ApplicationErrorWithRetryDelay struct {
```
cretz, line 141: Unused type

internal/error.go
```diff
@@ -303,6 +322,28 @@ func NewApplicationError(msg string, errType string, nonRetryable bool, cause er
 	return applicationErr
 }
 
+func NewApplicationErrorWithExtraInformation(errorDescriptors ErrorAttributes, extraInformation ExtraInformation) error {
```
cretz, line 325: To match other parts of the Go SDK, can this be:

```go
type ApplicationErrorOptions struct {
    Details []any
    Cause error
    NonRetryable bool
    NextAttemptRetryDelay time.Duration
}

func NewApplicationErrorWithOptions(message, errType string, options ApplicationErrorOptions) error
```

internal/failure_converter.go
```diff
@@ -92,10 +93,17 @@ func (dfc *DefaultFailureConverter) ErrorToFailure(err error) *failurepb.Failure
...
 			Details:      convertErrDetailsToPayloads(err.details, dfc.dataConverter),
 		}
 		failure.FailureInfo = &failurepb.Failure_ApplicationFailureInfo{ApplicationFailureInfo: failureInfo}
+	//case *ApplicationErrorWithRetryDelay:
+	//	failureInfo := &failurepb.ApplicationFailureInfo{
+	//		Type:         err.errType,
+	//		NonRetryable: false,
+	//		Details:      convertErrDetailsToPayloads(err.details, dfc.dataConverter),
+	//	}
+	//	failure.FailureInfo = &failurepb.Failure_ApplicationFailureInfo{ApplicationFailureInfo: failureInfo}
```
cretz, line 106: Commented out code

temporal/error.go
```diff
@@ -166,6 +168,88 @@ var (
 	ErrSkipScheduleUpdate = internal.ErrSkipScheduleUpdate
 )
 
+// Use instance of the interface to set all the desired attributes of a new ApplicationError
+// To get a new instance use ErrorAttributes function
+type ApplicationErrorAttributes interface {
```
cretz, line 173: I don't think there's any value in this being an interface

activity/activity.go
```diff
@@ -41,6 +41,8 @@ type (
 
 	// RegisterOptions consists of options for registering an activity
 	RegisterOptions = internal.RegisterActivityOptions
+
+	ActivityExtra = internal.ActivityExtra
```
cretz, line 45: While we don't have to, should probably wait on the name of the proto from https://github.com/temporalio/api/pull/334 and name this similarly. I'm thinking:

```suggestion
	ExtraAttributes = internal.ActivityExtraAttributes
```

But we'll see what that message gets named in that PR. I have an outstanding comment at https://github.com/temporalio/api/pull/334/files#r1422999174. I don't think it's clear to call it `ActivityRequests` there, and no reason not to use similar naming there and here since this will be applied across SDKs as a new Temporal term.

activity/activity.go
```diff
@@ -109,3 +111,7 @@ func GetWorkerStopChannel(ctx context.Context) <-chan struct{} {
 func IsActivity(ctx context.Context) bool {
 	return internal.IsActivity(ctx)
 }
+
+func ExtraInformation() ActivityExtra {
+	return internal.NewActivityExtra()
+}
```
cretz, line 117: Recommend:

```go
func GetExtraAttributes(ctx context.Context) ExtraAttributes {
    return internal.GetActivityExtraAttributes(ctx)
}

func SetExtraAttributes(ctx context.Context, attrs ExtraAttributes) error {
    return internal.SetActivityExtraAttributes(ctx, attrs)
}
```

There are some alternative designs, but I feel this is simplest. Also, please add Godoc to all public API. Also, please mark this as experimental for now until we are confident in the server-side implementation.

internal/extra_information.go
cretz, line 1: Not deserving of a new file IMO

temporal/error.go
Sushisource: +1

bergundy: I think we want Chad's way of constructing application error, don't see a reason for including anything in this file other than exporting the internal constructor.

internal/error.go
bergundy: I would also return `*ApplicationError` here instead of the `error` interface.

activity/activity.go
```diff
@@ -41,6 +41,10 @@ type (
 
 	// RegisterOptions consists of options for registering an activity
 	RegisterOptions = internal.RegisterActivityOptions
+
+	// ExtraInformation is an additional data which should be sent to the server to influence its decision
+	// about the next retry
+	ExtraInformation = internal.ActivityExtraInformation
```
cretz, line 47: "extra information" implies that it's something like the existing "GetInfo" (also, it's confusing to call it "information" here when we use "info" elsewhere). I think "extra attributes" is probably best. But we can wait until we see what happens in https://github.com/temporalio/api/pull/334/files#r1422999174. There are a lot of different words like "extra", "information", "requests", etc being used.

internal/activity.go
```diff
@@ -60,6 +60,11 @@ type (
 		IsLocalActivity   bool          // true if it is a local activity
 	}
 
+	// ActivityExtraInformation contains additional data to be sent to the server on activity failure
+	ActivityExtraInformation struct {
+		retryDelay *time.Duration
```
cretz, line 65:
```suggestion
		RetryDelay time.Duration
```

Can we just make this a public field like we do with activity info? Also, would prefer if we could avoid pointers to primitives like this and instead just assume unset/0 means unset.

activity/activity.go
```diff
@@ -109,3 +113,9 @@ func GetWorkerStopChannel(ctx context.Context) <-chan struct{} {
 func IsActivity(ctx context.Context) bool {
 	return internal.IsActivity(ctx)
 }
+
+// NewExtraInformation initialize structure to provide ExtraInformation to the server from the activity
+// use WithXXX methods available for ExtraInformation instance to add your data
+func NewExtraInformation() ExtraInformation {
```
cretz, line 119: I don't think this call is necessary

internal/activity.go
```diff
@@ -353,6 +358,16 @@ func WithLocalActivityTask(
 	})
 }
 
+func NewExtraInformation() ActivityExtraInformation {
```
cretz, line 361: I don't think this call is necessary

internal/activity.go
```diff
@@ -353,6 +358,16 @@ func WithLocalActivityTask(
...
+	info := ActivityExtraInformation{}
+	return info
+}
+
+func (b ActivityExtraInformation) WithRetryDelay(delay time.Duration) ActivityExtraInformation {
```
cretz, line 366: Would recommend just letting users set this as a field

internal/error.go
```diff
@@ -127,6 +135,17 @@ type (
 		nonRetryable bool
 		cause        error
 		details      converter.EncodedValues
+		extra        ExtraInformation
```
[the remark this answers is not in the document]
cretz: I don't think you need "extra information" broken out, just put the next retry field on here explicitly for now (though of course I disagree with it being on error at all).

temporal/error.go
```diff
@@ -166,6 +168,80 @@ var (
...
+}
+
+func (b errorAttributes) Details() []interface{} {
+	return b.details[:]
+}
+
+// NewApplicationErrorWithExtraInformation creates new instance of *ApplicationError type, all the attributes of the
+// newly created error could be controlled through instance of ApplicationErrorAttributes see: ErrorAttributes
+// This function also receives some extra information. See activity.ActivityExtraInformation for details.
+func NewApplicationErrorWithExtraInformation(attributes ApplicationErrorAttributes, extras ExtraInformation) error {
```
cretz, line 241: I think this should look like https://github.com/temporalio/sdk-go/pull/1319#discussion_r1425330300:

```
type ApplicationErrorOptions struct {
    Details []any
    Cause error
    NonRetryable bool
    NextAttemptRetryDelay time.Duration
}

func NewApplicationErrorWithOptions(message, errType string, options ApplicationErrorOptions) error
```

You don't need all the extra indirection.

internal/activity.go
bergundy: Yes please, we don't use the builder pattern in this SDK AFAIK.

temporal/error.go
bergundy: Yes please. This is way more inline with the other SDK APIs.

temporal/error.go
```diff
@@ -166,6 +168,17 @@ var (
...
+// To get a new instance use ErrorAttributes function
+type ApplicationErrorAttributes = internal.ApplicationErrorAttributes
+
+// NewApplicationErrorWithExtraRequests creates new instance of *ApplicationError type, all the attributes of the
+// newly created error could be controlled through instance of ApplicationErrorAttributes.
+// This function also receives some extra requests. See activity.ActivityExtraRequests for details.
+func NewApplicationErrorWithExtraRequests(msg, errType string, attributes ApplicationErrorAttributes, extras ExtraRequests) error {
+	return internal.NewApplicationErrorWithExtraRequests(msg, errType, attributes, extras)
+}
+
```
cretz, line 181: I thought per https://github.com/temporalio/sdk-go/pull/1319#discussion_r1425865743 we were going to do something like this:

```go
type ApplicationErrorOptions struct {
    Details []any
    Cause error
    NonRetryable bool
    NextAttemptRetryDelay time.Duration
}

func NewApplicationErrorWithOptions(message, errType string, options ApplicationErrorOptions) error
```

I strongly suggest we take this approach. (will save review of the other parts until after this fundamental part is decided)

ast2023: @cretz This is what I thought we discussed in December. I captured the screenshot for my notes as outcome. The comment you mentioned is regarding moving message and error type out of the error attribvutes structure which makes them required at the compile time. Please let me know if you would like to discuss further. 
Screenshot 2024-01-11 at 2 31 26 PM

cretz: I do think the single `WithOptions` and a single options set is simpler. Especially since now this will only be an error option, this is no different than any other optional field on application error I think.

bergundy: I think it's okay to go with the first option (single Options struct) and document that _currently_ the retry delay is only applicable in activities.

temporal/error.go
```diff
@@ -166,6 +168,17 @@ var (
...
 )
 
+// ApplicationErrorAttributes should be used to set all the desired attributes of a new ApplicationError
+// To get a new instance use ErrorAttributes function
+type ApplicationErrorAttributes = internal.ApplicationErrorAttributes
+
+// NewApplicationErrorWithExtraRequests creates new instance of *ApplicationError type, all the attributes of the
+// newly created error could be controlled through instance of ApplicationErrorAttributes.
+// This function also receives some extra requests. See activity.ActivityExtraRequests for details.
+func NewApplicationErrorWithExtraRequests(msg, errType string, attributes ApplicationErrorAttributes, extras ExtraRequests) error {
```
Quinn-With-Two-Ns, line 178: This name should be `NewApplicationErrorWithOptions`

internal/error.go
```diff
@@ -119,14 +119,22 @@ Workflow consumers will get an instance of *WorkflowExecutionError. This error w
 */
 
 type (
+	ApplicationErrorAttributes struct {
```
Quinn-With-Two-Ns, line 122: Per @cretz comments I think this should be `ApplicationErrorOptions` no?

internal/extra_requests.go
```diff
@@ -0,0 +1,28 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package internal
+
+type ExtraRequests interface {
```
Quinn-With-Two-Ns, line 27: What is this interface for?

internal/error.go
```diff
@@ -119,14 +119,22 @@ Workflow consumers will get an instance of *WorkflowExecutionError. This error w
 */
 
 type (
+	ApplicationErrorOptions struct {
```
Quinn-With-Two-Ns, line 122: This struct and fields should be documented.

temporal/error.go
```diff
@@ -166,6 +166,17 @@ var (
 	ErrSkipScheduleUpdate = internal.ErrSkipScheduleUpdate
 )
 
+// ApplicationErrorOptions should be used to set all the desired attributes of a new ApplicationError
+// To get a new instance use ErrorAttributes function
+type ApplicationErrorOptions = internal.ApplicationErrorOptions
+
+// NewApplicationErrorWithOptions creates new instance of *ApplicationError type, all the attributes of the
+// newly created error could be controlled through instance of ApplicationErrorOptions.
+// This function also receives some extra requests. See activity.ApplicationErrorOptions for details.
```
Quinn-With-Two-Ns, line 175: > This function also receives some extra requests

Is this still relevant?

temporal/error.go
```diff
@@ -166,6 +166,17 @@ var (
...
+func NewApplicationErrorWithOptions(msg, errType string, attributes ApplicationErrorOptions) error {
```
cretz, line 176:
```suggestion
func NewApplicationErrorWithOptions(msg, errType string, options ApplicationErrorOptions) error {
```

internal/error.go
```diff
@@ -303,6 +311,27 @@ func NewApplicationError(msg string, errType string, nonRetryable bool, cause er
 	return applicationErr
 }
 
+func NewApplicationErrorWithOptions(msg string, errType string, attributes ApplicationErrorOptions) error {
```
cretz, line 314: Might consider making `NewApplicationError` call this overload internally

internal/internal_task_pollers.go
cretz, line 1: I don't think any changes need to be made to this file

internal/error.go
```diff
@@ -119,14 +119,22 @@ Workflow consumers will get an instance of *WorkflowExecutionError. This error w
...
-		errType      string
-		nonRetryable bool
-		cause        error
-		details      converter.EncodedValues
+		msg            string
+		errType        string
+		nonRetryable   bool
+		cause          error
+		details        converter.EncodedValues
+		nextRetryDelay time.Duration
```
cretz, line 137: I think we should add a `func (*ApplicationError) NextRetryDelay() time.Duration` akin to the other getters on this struct

cretz (commented):
Looking good, I love the simplicity. Some minor things.

internal/error.go
```diff
@@ -303,6 +317,27 @@ func NewApplicationError(msg string, errType string, nonRetryable bool, cause er
 	return applicationErr
 }
 
+func NewApplicationErrorWithOptions(msg string, errType string, attributes ApplicationErrorOptions) error {
```
bergundy, line 320: Can this return `*ApplicationError` and follow the recommended Go practice?

ast2023: Could you kindly share the practices you have in mind?

internal/error.go
ast2023: @cretz these getter functions are only used in tests. When we will need access to the ApplicationError fields outside of the package we can allways add the function.

internal/error.go
ast2023: @Quinn-With-Two-Ns done

internal/error.go
cretz: It is often better Go practice to return the concrete structure (w/ pointer as needed, such as here) instead of the interface (`error`). But we haven't followed that in `NewApplicationError` or `NewApplicationErrorWithCause` or any other `NewXError` call in the `temporal` package, so I can understand both arguments. I do not have an opinion and could totally understand leaving it as is for consistency.

internal/error.go
Quinn-With-Two-Ns: Users may also use those getters

internal/error.go
```diff
@@ -119,14 +119,28 @@ Workflow consumers will get an instance of *WorkflowExecutionError. This error w
 */
 
 type (
+	// ApplicationErrorOptions represents a combination of error attributes and additional requests.
+	// All fields are optional, providing flexibility in error customization.
+	// NextRetryInterval is a request from server to override retry interval calculated by the
+	// server according to the RetryPolicy set by the Workflow.
+	// IMPORTANT: NextRetryInterval is meaningful only within the context of errors originating from Activity.
```
Quinn-With-Two-Ns, line 126: Nit:  Generally in the Go SDK we would provide documentation for members on the member variables themselves

internal/error.go
```diff
@@ -119,14 +119,28 @@ Workflow consumers will get an instance of *WorkflowExecutionError. This error w
...
+	// It is impossible to specify immediate retry as it is indistinguishable from the default value.
```
Quinn-With-Two-Ns, line 127: If users do want immediate retries is the "work around" to just pass a very short timeout?

internal/error.go
cretz: @ast2023 - they are used by users too. Just because they are internal, if they are on the struct they are available.

internal/error.go
```diff
@@ -119,14 +119,28 @@ Workflow consumers will get an instance of *WorkflowExecutionError. This error w
...
+	ApplicationErrorOptions struct {
```
Quinn-With-Two-Ns, line 128: Should also note this does not apply to local activities,

internal/error.go
ast2023: @Quinn-With-Two-Ns you are correct. It is so trivial solution that I did not think it is required to write it down. I can add it.

Quinn-With-Two-Ns: You are correct it is trivial, but sometimes it is nice to point it out anyway for users so they know it is intended. Not a big deal if you don't want to include it.

Quinn-With-Two-Ns (approved):
LGTM, I'll wait for @cretz or @bergundy to review as well before merging

internal/error.go
```diff
@@ -523,6 +548,8 @@ func (e *ApplicationError) Unwrap() error {
 	return e.cause
 }
 
+func (e *ApplicationError) NextRetryDelay() time.Duration { return e.nextRetryDelay }
```
cretz, line 551: Should have godoc, but not that big of a deal.

internal/failure_converter.go
```diff
@@ -91,9 +95,10 @@ func (dfc *DefaultFailureConverter) ErrorToFailure(err error) *failurepb.Failure
 	switch err := err.(type) {
 	case *ApplicationError:
 		failureInfo := &failurepb.ApplicationFailureInfo{
-			Type:         err.errType,
-			NonRetryable: err.nonRetryable,
-			Details:      convertErrDetailsToPayloads(err.details, dfc.dataConverter),
+			Type:           err.errType,
+			NonRetryable:   err.NonRetryable(),
+			Details:        convertErrDetailsToPayloads(err.details, dfc.dataConverter),
+			NextRetryDelay: durationpb.New(err.nextRetryDelay),
```
cretz, line 101: Sorry I missed this before. We should leave this as `nil` if `nextRetryDelay` is `0`.

cretz (commented):
Sorry, only one minor thing about proto nil and I can mark approved

internal/error.go
ast2023: I am failing to find documentation which will explain NextRetryDelay better than the name of the method.

cretz: Godoc rules often stipulate Godoc for all exported items regardless of obviousness (yes that means usually if redundant). Granted we don't have such a strict requirement, but we usually follow it.

PR #1320, opened 2023-12-14
Cleanup dead code and incorrect use of go routine in context pkg

### LINKED ISSUE #1270: Clean up internal/context
`internal/context.go` has some tech debt that should be cleaned up including

* Commented out code that should either be deleted or enabled https://github.com/temporalio/sdk-go/blob/269b55b8d184849209c6973fe913a427fafc07e3/internal/context.go#L348
* Incorrect use of `go` in a workflow context (this code appears to be dead, but should be removed regardless) https://github.com/temporalio/sdk-go/blob/269b55b8d184849209c6973fe913a427fafc07e3/internal/context.go#L239

tharun208:
## What was changed
Cleaned up dead code in context pkg and replaced incorrect use of go routine in `propogateCancel` with a panic msg

## Why? 
Clean tech-debt from context pkg

## Checklist

1. Closes #1270 

2. How was this tested:
    local test-suites

4. Any docs updates needed?
    No

### DIFF
--- a/internal/context.go
+++ b/internal/context.go
@@ -198,13 +198,14 @@ func WithCancel(parent Context) (ctx Context, cancel CancelFunc) {
 
 // NewDisconnectedContext returns a new context that won't propagate parent's cancellation to the new child context.
 // One common use case is to do cleanup work after workflow is canceled.
-//  err := workflow.ExecuteActivity(ctx, ActivityFoo).Get(ctx, &activityFooResult)
-//  if err != nil && temporal.IsCanceledError(ctx.Err()) {
-//    // activity failed, and workflow context is canceled
-//    disconnectedCtx, _ := workflow.NewDisconnectedContext(ctx);
-//    workflow.ExecuteActivity(disconnectedCtx, handleCancellationActivity).Get(disconnectedCtx, nil)
-//    return err // workflow return CanceledError
-//  }
+//
+//	err := workflow.ExecuteActivity(ctx, ActivityFoo).Get(ctx, &activityFooResult)
+//	if err != nil && temporal.IsCanceledError(ctx.Err()) {
+//	  // activity failed, and workflow context is canceled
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
Thanks @tharun208 

Now another approach instead of deleting the code would be to make it work. That would be a non trivial effort since it isn't working right now and is untested. Given that no users has asked for these features, and to implement properly I think they would require generating timer commands I am fine to remove it.

There are still some references to the removed code such as:
* https://github.com/temporalio/sdk-go/pull/1320/files#diff-6d7161d5c4503fb1cb21cd22eeef02384739d27c6622b8ad43e2179a267e64a1R253
* https://github.com/temporalio/sdk-go/blob/7a9640d531f5eba5c3fa365e770e34bfb5676337/internal/context.go#L52

That we should remove

tharun208:
@Quinn-With-Two-Ns done removed the mentioned snippet

internal/context.go
```diff
@@ -236,14 +237,7 @@ func propagateCancel(parent Context, child canceler) {
...
 	} else {
-		go func() {
-			s := NewSelector(parent)
-			s.AddReceive(parent.Done(), func(c ReceiveChannel, more bool) {
-				child.cancel(false, parent.Err())
-			})
-			s.AddReceive(child.Done(), func(c ReceiveChannel, more bool) {})
-			s.Select(parent)
-		}()
+		panic("parent context not found")
```
Quinn-With-Two-Ns, line 240: I think this panic statement is wrong and misleading. The issue is not that the parent context is not found, the issue is the parent context is cancellable, but does not have a `*cancelCtx` in it's hierarchy. We should reflect that in the panic message

tharun208: updated the message

Quinn-With-Two-Ns (approved):
Thanks for the contribution!

PR #1327, opened 2023-12-19
Include update state machine fields in panic

Quinn-With-Two-Ns:
Include update state machine fields in panic. This will help debug update issues by making it easier to interpret the panics message.

### DIFF
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -130,7 +130,7 @@ func (up *updateProtocol) requireState(action string, valid ...updateState) {
 			return
 		}
 	}
-	panicIllegalState(fmt.Sprintf("invalid action %q in update protocol %v", action, up))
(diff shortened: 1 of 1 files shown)


Sushisource (approved):

internal/internal_update.go
```diff
@@ -130,7 +130,7 @@ func (up *updateProtocol) requireState(action string, valid ...updateState) {
 			return
 		}
 	}
-	panicIllegalState(fmt.Sprintf("invalid action %q in update protocol %v", action, up))
+	panicIllegalState(fmt.Sprintf("invalid action %q in update protocol %+v", action, up))
```
cretz, line 133: Does this occur for user mistakes or just mistakes we've made internal to the SDK/server? If the former, I would prefer more structure to the message

Quinn-With-Two-Ns: Mistakes we have made

cretz (approved):

PR #1333, opened 2024-01-03
Bump go-api version to fix protojson vuln

tdeebswihart:
## What was changed
I bumped the version of api-go

## Why?

To fix the protojson DOS vulns recently patched in the upstream golang/protobuf. See https://github.com/temporalio/api-go/pull/143 for details


go.mod
```diff
@@ -10,12 +10,12 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.8.4
-	go.temporal.io/api v1.26.1
+	go.temporal.io/api v1.26.1-0.20240103185939-608bdd111e4b
```
cretz, line 13: How can we make sure we don't release the SDK with an untagged API version?

tdeebswihart: I'm happy to cut a release of api-go if we're ok with releasing all of https://github.com/temporalio/api-go/compare/v1.26.0...master

tdeebswihart: @cretz we can't cut a release of api-go because we haven't released the gogoproto changes into the wild.

cretz: Right, I just want to make sure we remember to tag this before next SDK release

cretz:
Would rather have @Quinn-With-Two-Ns's approval here over my own. He may prefer not to merge this at this time.

Quinn-With-Two-Ns (approved):

PR #1335, opened 2024-01-04
Add last BuildID to workflow info

Sushisource:
## What was changed
Add the BuildID of the last completed task to info. Importantly this *does not* use the worker's current build id if it isn't in history, as that would be nondeterministic, unlike Binary Checksum which is guaranteed to be set as soon as the first WFT is completed which is not the case here.

## Why?
Useful

## Checklist

1. Part of https://github.com/temporalio/features/issues/253
2. How was this tested:
Added IT

3. Any docs updates needed?


internal/internal_task_handlers.go
```diff
@@ -267,10 +270,16 @@ func (eh *history) isNextWorkflowTaskFailed() (task finishedTask, err error) {
...
 		var flags []sdkFlag
 		if nextEventType == enumspb.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
-			binaryChecksum = nextEvent.GetWorkflowTaskCompletedEventAttributes().BinaryChecksum
-			for _, flag := range nextEvent.GetWorkflowTaskCompletedEventAttributes().GetSdkMetadata().GetLangUsedFlags() {
+			completedAttrs := nextEvent.GetWorkflowTaskCompletedEventAttributes()
+			binaryChecksum = completedAttrs.BinaryChecksum
+			workerVersion := completedAttrs.GetWorkerVersion()
+			if workerVersion != nil {
+				buildID = workerVersion.GetBuildId()
+			}
```
bergundy, line 281:
```suggestion
			buildID = completedAttrs.GetWorkerVersion().GetBuildId()
```

Sushisource:
Realized I did this a bit of a silly way at first. Draft while I fix it.

test/worker_versioning_test.go
```diff
@@ -273,3 +274,63 @@ func (ts *WorkerVersioningTestSuite) TestReachabilityVersions() {
...
+	ts.NoError(ts.client.SignalWorkflow(ctx, wfHandle.GetID(), wfHandle.GetRunID(), "fakesig", ""))
+	var lastBuildID string
+	enval, err := ts.client.QueryWorkflow(ctx, wfHandle.GetID(), wfHandle.GetRunID(), "get-last-build-id", nil)
+	ts.NoError(err)
+	ts.NoError(enval.Get(&lastBuildID))
+	worker1.Stop()
+	ts.Equal("1.0", lastBuildID)
+
+	// Add new compat ver
+	err = ts.client.UpdateWorkerBuildIdCompatibility(ctx, &client.UpdateWorkerBuildIdCompatibilityOptions{
```
bergundy, line 306: nit: this doesn't require versioning to test, you can use build ID with unversioned workers.

bergundy: Could make sense to test with and without versioning just to be safe.

bergundy (approved):

internal/internal_task_handlers.go
Sushisource: Already different :)

test/worker_versioning_test.go
Sushisource: The bit you reviewed is subtly wrong in terms of querying on a cached worker after the worker has completed a task with its' ID. New test also makes sure the first task doesn't have an ID, which covers this path.

internal/workflow.go
```diff
@@ -1016,14 +1020,22 @@ type UpdateInfo struct {
...
 	if wInfo.BinaryChecksum == "" {
 		return getBinaryChecksum()
 	}
 	return wInfo.BinaryChecksum
 }
 
+// GetCurrentBuildID returns the Build ID of the worker that processed this task, which may be
+// empty. During replay this id may not equal the id of the replaying worker. If not replaying and
+// this worker has a defined Build ID, it will equal that ID. It is safe to use for branching.
+func (wInfo *WorkflowInfo) GetCurrentBuildID() string {
```
Quinn-With-Two-Ns, line 1035: May want to add a note for queries

internal/internal_task_handlers.go
```diff
@@ -1016,6 +1026,11 @@ ProcessEvents:
 		} else {
 			w.workflowInfo.BinaryChecksum = binaryChecksum
 		}
+		if isReplay {
+			w.workflowInfo.currentTaskBuildID = currentBuildID
+		} else if !isReplay && !isQueryOnlyTask {
```
Quinn-With-Two-Ns, line 1031: can we add a comment here explaining why a query only task is treated special.

Quinn-With-Two-Ns (approved):

Issue #1313, opened 2023-12-01
Support OpenTelemetry Metrics


cretz:
**Describe the solution you'd like**

Now that it is stable (ref #113 and https://opentelemetry.io/blog/2023/otel-go-metrics-sdk-stable/), we should add `NewMetricsHandler` to `go.temporal.io/sdk/contrib/opentelemetry`.

PR #1336, opened 2024-01-04
Add otel metrics support

Quinn-With-Two-Ns:
Add open telemetry metric support

closes https://github.com/temporalio/sdk-go/issues/1313


contrib/opentelemetry/handler.go
```diff
@@ -0,0 +1,115 @@
...
+			return nil
+		}))
+	return metrics.GaugeFunc(func(f float64) {
+		config.Store(f)
+	})
+}
+
+func (m MetricsHandler) Timer(name string) client.MetricsTimer {
+	// TODO handle error?
+	h, _ := m.Meter.Float64Histogram(name, metric.WithUnit("s"))
```
Quinn-With-Two-Ns, line 111: We could also add an option to provide a default histogram hint here, I left it off for now since that feature was experimental

contrib/opentelemetry/handler.go
```diff
@@ -0,0 +1,115 @@
...
+	return metrics.CounterFunc(func(d int64) {
+		c.Add(context.Background(), d, metric.WithAttributeSet(m.Attributes))
+	})
+}
+
+func (m MetricsHandler) Gauge(name string) client.MetricsGauge {
+	// TODO(https://github.com/open-telemetry/opentelemetry-go/issues/3984) replace with sync gauge once supported
+	var config atomic.Value
+	config.Store(0.0)
+	m.Meter.Float64ObservableGauge(name,
```
Quinn-With-Two-Ns, line 99: Currently sync gauges are not stable/implemented in the OTEL Go SDK

contrib/opentelemetry/handler.go
```diff
@@ -0,0 +1,115 @@
...
+	return metrics.GaugeFunc(func(f float64) {
+		config.Store(f)
+	})
+}
+
+func (m MetricsHandler) Timer(name string) client.MetricsTimer {
+	// TODO handle error?
+	h, _ := m.Meter.Float64Histogram(name, metric.WithUnit("s"))
+	return metrics.TimerFunc(func(t time.Duration) {
+		h.Record(context.Background(), t.Seconds(), metric.WithAttributeSet(m.Attributes))
```
Quinn-With-Two-Ns, line 113: Instead of using a `context.Background()` we could add an API to attach a context to the handler. That will be useful once `Examplars` are supported in the SDK to tie metrics and traces

contrib/opentelemetry/handler.go
```diff
@@ -0,0 +1,115 @@
...
+
+type MetricsHandler struct {
+	// Meter to create metrics
+	Meter metric.Meter
+	// Attributes currently set on the handler
+	Attributes attribute.Set
+}
+
+// NewMetricsHandler returns a client.MetricsHandler that is backed by the given Meter
+func NewMetricsHandler(meter metric.Meter) client.MetricsHandler {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I wasn't worried about  future proofing since this package is not 1.0, but I fine making this change now.

contrib/opentelemetry/handler.go
```diff
@@ -0,0 +1,115 @@
...
+	if meter == nil {
+		meter = otel.GetMeterProvider().Meter("temporal-sdk-go")
+	}
+	return MetricsHandler{
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: By "hiding this struct and providing ways to extract specific pieces" do you mean hiding the members and adding getters?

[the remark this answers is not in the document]
cretz: Yes, or hiding the struct altogether and making an interface w/ getters. But I'm kinda ok with what you have today too (though I think I may prefer pointer receivers, but honestly not even that sure there either). I don't have strong opinions with this comment.

contrib/opentelemetry/go.mod
```diff
@@ -23,6 +25,8 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/robfig/cron v1.2.0 // indirect
 	github.com/stretchr/objx v0.5.0 // indirect
+	go.opentelemetry.io/otel/metric v1.21.0
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yes can confirm

contrib/opentelemetry/handler.go
```diff
@@ -0,0 +1,151 @@
...
+		meter:      m.meter,
+		attributes: attribute.NewSet(attributes...),
+		onError:    m.onError,
+	}
+}
+
+func (m MetricsHandler) Counter(name string) client.MetricsCounter {
+	c, err := m.meter.Int64UpDownCounter(name)
+	if err != nil {
+		m.onError(err)
```
cretz, line 122:
```suggestion
		m.onError(err)
		return client.MetricsNopHandler.Counter(name)
```

(same for timer)

contrib/opentelemetry/handler.go
```diff
@@ -0,0 +1,151 @@
...
+	return metrics.CounterFunc(func(d int64) {
+		c.Add(context.Background(), d, metric.WithAttributeSet(m.attributes))
+	})
+}
+
+func (m MetricsHandler) Gauge(name string) client.MetricsGauge {
+	// TODO(https://github.com/open-telemetry/opentelemetry-go/issues/3984) replace with sync gauge once supported
+	var config atomic.Value
+	config.Store(0.0)
+	m.meter.Float64ObservableGauge(name,
```
cretz, line 133: Handle error

cretz (approved):
I'm still kinda thinking we should make `*MetricsHandler` the impl instead of `MetricsHandler`, but it's not a big deal

PR #1337, opened 2024-01-04
Remove override for propagateCancel

Quinn-With-Two-Ns:
Remove override for `propagateCancel` from the determinism since it no longer starts a go routine.


Quinn-With-Two-Ns:
I realized this change may break some people who don't keep their Go SDK up to date, but do keep their workflow checker. I think I will leave the checker as is for now to avoid breaking users

PR #1338, opened 2024-01-05
Fix flaky build id issue

Sushisource:
This should fix the only case where I could imagine the id not being set. Could not repro locally.


internal/internal_task_handlers.go
```diff
@@ -1028,10 +1028,16 @@ ProcessEvents:
 		}
 		if isReplay {
 			w.workflowInfo.currentTaskBuildID = currentBuildID
-		} else if !isReplay && !isQueryOnlyTask {
+		} else if !isQueryOnlyTask {
 			// Query only tasks should use the build ID from the workflow task, not the worker's
```
Quinn-With-Two-Ns, line 1032: This comment seem irrelevant now?

internal/internal_task_handlers.go
```diff
@@ -1028,10 +1028,16 @@ ProcessEvents:
...
 			// build id, since the user cares about what affected workflow state.
 			w.workflowInfo.currentTaskBuildID = w.wth.workerBuildID
+		} else {
+			// It is a query only task, and we're not replaying somehow - still use the task's build
```
Quinn-With-Two-Ns, line 1036: I   don't  understand this comment, didn't we discuss before that queries are always treated as not replaying?

Sushisource: Yeah, I can just get rid of the not replaying part

internal/internal_task_handlers.go
Sushisource: I made them better

internal/internal_task_handlers.go
```diff
@@ -1028,10 +1028,15 @@ ProcessEvents:
...
-		} else if !isReplay && !isQueryOnlyTask {
-			// Query only tasks should use the build ID from the workflow task, not the worker's
-			// build id, since the user cares about what affected workflow state.
+		} else if !isQueryOnlyTask {
+			// Real new (non-query, non-replay) tasks use the worker's build ID
 			w.workflowInfo.currentTaskBuildID = w.wth.workerBuildID
+		} else {
+			// It *is* a query only task, use the task's build ID (queries are considered non-replay
+			// in go sdk), but only if one was actually discovered.
+			if currentBuildID != "" {
```
Quinn-With-Two-Ns, line 1037: When is `currentBuildID` empty?

Sushisource: It can be empty here if the lookahead didn't find anything, but we wouldn't want to overwrite some previous task which did find something.

Quinn-With-Two-Ns: Yeah I am wondering what scenario look ahead can't find anything?

Sushisource: The problem is sort of specific to queries on the first task I think. No WFT completed event exists yet to be looked-ahead-to (well, it does actually in the query, but we don't read it since we're already up-to-date), but the `nextTask` code still runs, which then returns this as empty.

Quinn-With-Two-Ns: Ah I didn't think of the first WFT thanks for clarifying

Quinn-With-Two-Ns (approved):

Issue #359, opened 2021-02-11
Add deterministic range over map helper


mfateev:
**Is your feature request related to a problem? Please describe.**
Go built in range over map is not deterministic. The workaround has to be implemented by each user.

**Describe the solution you'd like**
Some helper to support deterministic iteration over map.

Quinn-With-Two-Ns:
Was deferring this until [range over func](https://github.com/golang/go/wiki/RangefuncExperiment) support was added in Go 1.22, but looks like that didn't make that release.

PR #1340, opened 2024-01-08
Add deterministic range over map helper

Quinn-With-Two-Ns:
Add deterministic range over map helpers for ordered and unordered maps.

Ideally this would be implemented with [range over func](https://github.com/golang/go/wiki/RangefuncExperiment), but it is only experimental in golang 1.22 so we would have to wait over a year to use it assuming it is stabilized in 1.23.

Closes https://github.com/temporalio/sdk-go/issues/359

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -31,6 +31,9 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/slices"
+
 	"google.golang.org/protobuf/types/known/durationpb"
 
 	commonpb "go.temporal.io/api/common/v1"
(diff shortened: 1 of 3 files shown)


Sushisource (approved):

workflow/workflow.go
```diff
@@ -635,3 +636,17 @@ func IsContinueAsNewError(err error) bool {
 func DataConverterWithoutDeadlockDetection(c converter.DataConverter) converter.DataConverter {
 	return internal.DataConverterWithoutDeadlockDetection(c)
 }
+
+// DeterministicKeys returns the keys of a map in deterministic (sorted) order. To be used in for
```
cretz, line 640: Pick whether to call this `DeterministicKeys` or `SortedKeys`. I don't have a preference

internal/workflow.go
```diff
@@ -2043,3 +2046,27 @@ func convertFromPBRetryPolicy(retryPolicy *commonpb.RetryPolicy) *RetryPolicy {
...
+	}
+	slices.Sort(r)
+	return r
+}
+
+// DeterministicKeysFunc returns the keys of a map in a deterministic (sorted) order.
+// cmp(a, b) should return a negative number when a < b, a positive number when
+// a > b and zero when a == b. Keys are sorted by cmp.
+// To be used in for loops in workflows for deterministic iteration.
+func DeterministicKeysFunc[K comparable, V any](m map[K]V, cmp func(a K, b K) int) []K {
```
cretz, line 2065:
```suggestion
func DeterministicKeysFunc[K, V any](m map[K]V, cmp func(a K, b K) int) []K {
```

I know the sorting doesn't require the comparable constraint, do maps? I guess that would make sense (and ignore me then)

Quinn-With-Two-Ns: yes maps do

workflow/workflow.go
Quinn-With-Two-Ns: Oh yeah I meant to update everything to `DeterministicKeys`

cretz (approved):

cretz (approved):

PR #1344, opened 2024-01-10
Replace go.uber.org/atomic with sync/atomic

serprex:
## What was changed
go 1.19 introduced atomic types
https://tip.golang.org/doc/go1.19#atomic_types

## Why?
sync/atomic is already being used here. I intend to update dependencies so that `go.uber.org/atomic` can also be removed as an indirect dependency

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.8.4
 	go.temporal.io/api v1.26.1-0.20240103185939-608bdd111e4b
-	go.uber.org/atomic v1.9.0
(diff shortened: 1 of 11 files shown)


serprex:
I always struggle with how to manage go.mod/go.sum changes

cretz (approved):
> I always struggle with how to manage go.mod/go.sum changes

Yeah, our repo makes this a bit hard. Basically you need to run `go mod tidy` in every directory that has a `go.mod` in this repo.

This LGTM, will wait for @Quinn-With-Two-Ns to take a look

Quinn-With-Two-Ns (approved):
LGTM, I have been meaning to do this anyway

cretz:
Merged, thanks!

PR #1345, opened 2024-01-12
Replace some Sprintfs with string concatenation

dnr:
## What was changed
Replace Sprintf with concatenation in a few places.

## Why?
This was visible in a profile of some worker pods doing heavy replays so I thought I'd try to improve it.

I created a benchmark out of the [schedules replay test](https://github.com/temporalio/temporal/blob/main/service/worker/scheduler/replay_test.go):

```
before:
     272          21243887 ns/op         4882488 B/op      91354 allocs/op
after:
     278          20966780 ns/op         4862933 B/op      90165 allocs/op
```
(best of 10 runs for each, after had less variance also)

So a 1-2% improvement in cpu and a tiny bit less memory.

## Checklist

2. How was this tested:
existing unit tests, plus simple benchmark


internal/internal_workflow.go
```diff
@@ -751,7 +759,7 @@ func (c *channelImpl) Receive(ctx Context, valuePtr interface{}) (more bool) {
 				}
 				break // Corrupt signal. Drop and reset process.
 			}
-			state.yield(fmt.Sprintf("blocked on %s.Receive", c.name))
+			state.yield(printfString{fmt: "blocked on %s.Receive", arg1: c.name})
```
cretz, line 762:
```suggestion
			state.yield("blocked on "+c.name+".Receive")
```

Can you microbenchmark this approach?

cretz (commented):
Not sure this is worth it, but made a comment where maybe simple string concatenation can be good enough for you

internal/internal_workflow.go
dnr: Ah, yeah, that provides a basically identical improvement, and is simpler. I'll switch to that. I might have hoped the compiler could optimize this but I guess Go doesn't go for stuff like that

Quinn-With-Two-Ns: Are you running with PGO during your benchmark?

dnr: No, I don't think I've ever built a server with PGO. That's an interesting angle to try. I couldn't find any indication that the compiler would partially-evaluate a fmt.Sprintf even with PGO though.

(In general we don't care much about cpu usage since the server is generally io-bound, but in this situation it was replaying a ton of schedule workflows and going over its cpu reservation, so I looked for some quick improvements there.)

Quinn-With-Two-Ns (approved):

cretz (approved):

PR #1350, opened 2024-01-13
Disable eager activities if tq rate limits is set

Quinn-With-Two-Ns:
Currently server does not properly rate limit eager activities. Disabling for now until the Server can address this.


Sushisource (approved):
Makes sense. Is there an item for this in Core? Or a more general one I can link to if I make one for Core?

Quinn-With-Two-Ns:
@Sushisource I opened this https://github.com/temporalio/features/issues/405

cretz (approved):

bergundy:
> Makes sense. Is there an item for this in Core? Or a more general one I can link to if I make one for Core?

There isn't an issue for this in Core but I opened a related issue a few days ago for worker-local rate limits with eager activities: https://github.com/temporalio/sdk-core/issues/669.

PR #1353, opened 2024-01-15
regenerate mocks under client to allow dynamic returns based on arguments

yinsidij:
## What was changed

mocker for interfaces under `client` folder

## Why?

Looks the mocks are old and only supports returning value rather than the function.
With updated mocker by latest mockery, (example https://github.com/temporalio/sdk-go/compare/master...yinsidij:yjiao/regenerate_mocks?expand=1#diff-baf21238e036d6550f43915acceadaf7448d1ff493453a0da0d41a232e20fa6fR217)
It allows dynamic return based on argument. 

For example, with the newly generated mock, we could do following that provides more flexibility.
```
	mockClient.On("ExecuteWorkflow", mock.Anything, mock.AnythingOfType("internal.StartWorkflowOptions"), mock.AnythingOfType("string"), mock.Anything).Return(
		func(ctx context.Context, options client.StartWorkflowOptions, workflow string, newCustomerMetadata *datamodel.NewCustomerMetadata) (client.WorkflowRun, error) {
			if newCustomerMetadata.CustomerID == "bad" {
				return nil, fmt.Errorf("error")
			}
			return nil, nil
		},
	).Maybe()
```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz:
Please approve CLA

yinsidij:
> Please approve CLA and fix copyright headers of files

signed CLA and iterated the PR.

yinsidij:
one CI job failed. somehow I cannot find a way to re trigger the job.

I locally ran this test and passed. Looks like rerun should be also resolve it.
image

Could you please help rerun the failed CI jobs?
@Quinn-With-Two-Ns @cretz

mocks/Client.go
```diff
@@ -162,23 +228,30 @@ func (_m *Client) DescribeWorkflowExecution(ctx context.Context, workflowID stri
 }
 
 // ExecuteWorkflow provides a mock function with given fields: ctx, options, workflow, args
-func (_m *Client) ExecuteWorkflow(ctx context.Context, options client.StartWorkflowOptions, workflow interface{}, args ...interface{}) (client.WorkflowRun, error) {
```
Quinn-With-Two-Ns, line 165: All these `internal.*`  need  to be replaced by  the appropriate exposed type.

yinsidij: @Quinn-With-Two-Ns 
thanks for review.
Actually `internal.WorkflowRun` is equivalent to `client.WorkflowRun`. FYI.
https://github.com/temporalio/sdk-go/blob/41d2288301cc205de56e1d7f5841e56edcb5f40d/client/client.go#L100

I will still make the change

Quinn-With-Two-Ns: It requires manual approval

Quinn-With-Two-Ns (approved):
Thanks for the contribution

yinsidij:
> Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

Thanks. Looks like I don't have permission to merge. What is the process for merging the PR ?

PR #1356, opened 2024-01-15
Dump coroutine stack on deadlock timeout

### LINKED ISSUE #969: Ability to dump stack of deadlocking coroutine

Right now when we hit deadlock detector in Go, we panic w/ the stack where we timed out but we don't show the stack of blocked coroutines in the workflow

Go does not have a "show stack of specific goroutine". In fact, Go doesn't even allow you to technically identify/address goroutines. But internally they do maintain a Go ID. I have seen projects take it via ugly stack parsing and via ugly internal reflection. So we could track these IDs when `workflow.Go` is called (we don't have to worry about users using real goroutines), then parse `runtime.Stack()` to filter out only the stack trace for known running coroutines of that workflow. If this is expensive to get that ID, we can make it opt-in (we don't care about how expensive it is as deadlock time though).

So, here's the goal:

* Allow users to see a stack of all running coroutines of a workflow

Here's the allowed compromise:

* If it's too expensive to create the identifier of the coroutine for tracking when deadlock detection hits, we could make it opt in.

See if there are any known projects that have done these things we could take inspiration from

TheJokr:
## What was changed
I added a new function `getCoroStackTrace` to retrieve the current stack trace of a `coroutineState`'s goroutine via the address of the `coroutineState` (non-cooperatively). As discussed in #969, Go does not let you dump a specific goroutine, so the only alternative is to dump all goroutines and search for the one you want. The linked issue suggests determining a unique ID per goroutine, but this already exists for free in the form of `*coroutineState`.

Each `coroutineState` occupies a unique location in memory, so by searching for its address in the stack trace, we can find its goroutine. I also included the name of the coroutine's initial function in the search to avoid false positives. A `runtime.KeepAlive` call in the initial function ensures the pointer value will always be present in the stack trace, otherwise its lifetime ends after the initial yield. Once found, the stack trace is then formatted similar to `getStackTrace`.

`getCoroStackTrace` is used to generate the stack trace for a `workflowPanicError` when a deadlock timeout occurs. This mimics the previous (panicing) behavior, but reports the stack trace of the deadlocked coroutine instead of the deadlock detector. To ensure ample space for the all-goroutines stack trace, I increased the size of `stackBuf` to 1MB.

## Why?
To simplify debugging workflow deadlocks or deadlock timeouts, especially spurious ones that are hard to reproduce.
Previously, a stack trace of the deadlock detector goroutine was emitted, which tells you very little about where the deadlock occured.

## Checklist

1. Closes #969 

2. How was this tested:
By running the existing unit and integration tests, as well as adding 1 new unit test for the stack trace specifically.

4. Any docs updates needed?
No, from the user's perspective a deadlock still leads to a panic error, just with a different stack trace.


internal/internal_workflow.go
```diff
@@ -1046,23 +1084,26 @@ func (s *coroutineState) stackTrace() string {
 	return <-stackCh
 }
 
+func (s *coroutineState) run(ctx Context, f func(ctx Context)) {
+	defer runtime.KeepAlive(&s) // keep receiver argument alive for getCoroStackTrace
```
cretz, line 1088: Can you help me understand this a bit? `s` is passed to `getCoroStackTrace` so doesn't that imply liveness? (also, not sure it's worth breaking this call out to a separate method makes since unless you are saying that the pointer is unstable)

cretz (commented):
Overall the idea looks right, just one question about detail. Would also like @Quinn-With-Two-Ns to weign in.

internal/internal_workflow.go
```diff
@@ -972,20 +974,53 @@ func getStackTrace(coroutineName, status string, stackDepth int) string {
...
 	// If the start is after the end, the depth was invalid originally so return
 	// the entire raw stack
 	if omitTop > omitEnd {
-		return rawStack
+		return stack
+	}
+	return strings.Join(lines[omitTop:omitEnd], "\n")
+}
+
+func getCoroStackTrace(crt *coroutineState, status string, stackDepth int) string {
```
Quinn-With-Two-Ns, line 997: Not a huge fan of this approach of using the pointer to `coroutineState` as the identifying ID since Go Lang is free to move the pointer afaik. I would prefer using a generated UUID approach personally.

internal/internal_workflow.go
```diff
@@ -972,20 +974,53 @@ func getStackTrace(coroutineName, status string, stackDepth int) string {
...
+
+	var coroStack []byte
+	needle := []byte(fmt.Sprintf("/internal.(*coroutineState).run(%p,", crt))
+	for _, c := range coros {
+		if bytes.Contains(c, needle) {
+			coroStack = c
+			break
+		}
+	}
+	if coroStack == nil {
```
Quinn-With-Two-Ns, line 1013: nit: would prefer these two situations be errors returned and handled by the caller

internal/workflow_deadlock_test.go
```diff
@@ -73,15 +71,17 @@ func TestDataConverterWithoutDeadlockDetection(t *testing.T) {
 		env.RegisterActivity(activityFn)
 		env.ExecuteWorkflow(workflowFn)
 		require.True(t, env.IsWorkflowCompleted())
+		return env.GetWorkflowError()
 	}
 
-	// Run with a slow converter and confirm it panics
+	// Run with a slow converter and confirm a deadlock is detected
+	conv := converter.GetDefaultDataConverter()
```
Quinn-With-Two-Ns, line 78: Why was this change made, was there a change in behaviour?

internal/internal_workflow.go
TheJokr: Sure! The liveness problem doesn't concern the `coroutineState` itself, but the pointer *value* `s` as an argument to this function. Without the `KeepAlive` call, the last use of `s` is at `s.initialYield(...)` (deferred calls evaluate their arguments [at the point of the defer statement](https://go.dev/ref/spec#Defer_statements)). Hence, when `f` is called, Go considers `s` dead **and doesn't print its value anymore in the goroutine's stack trace**. The search then can't find the stack we want.

`runtime.KeepAlive` is a special function in the sense that it's a no-op, but marks its argument as "used" for Go's liveness analysis. So replacing it with some regular Go statement runs the risk of the compiler optimizing the statement out.

To make sure `s` is alive throughout the entire function (including deferred calls!), `runtime.KeepAlive(s)` needs to be the very last statement. This means it must be deferred first, but since deferred calls evaluate their arguments eagerly, the value that is being kept alive is a copy of `s`. By instead keeping `&s` alive, `s` must also be kept alive since `&s` points to it.

TheJokr: N.B. the same concerns apply to any function argument we might want to use to identify the coroutine. This is not specific to `s` being a pointer itself.

internal/internal_workflow.go
TheJokr: I will change that. How would you like me to handle that error inside the deadlock detector? Format the error into the stack trace (similar to how the the return value looks now)?

Ignoring it seems like a bad idea, and I can't log it since `*coroutineState` doesn't have access to the environment's logger.

internal/workflow_deadlock_test.go
TheJokr: The dispatcher doesn't panic anymore on a deadlock timeout, it just returns a `workflowPanicError` similar to what a panic inside the coroutine does (if you meant the comment).

Wrt. passing conv as a parameter instead of by capture, I just felt like it is cleaner that way. This wasn't possible before since `require.Panics` mandated a specific function signature. If you'd prefer, I can revert that change.

internal/internal_workflow.go
Quinn-With-Two-Ns: >How would you like me to handle that error inside the deadlock detector? Format the error into the stack trace (similar to how the the return value looks now)?

Yes, formatting it into the stack trace is fine IMO.

internal/internal_workflow.go
TheJokr: If Go moved the `coroutineState`, it would have to update all pointers to it. That is, both `crt` here and the value in the stack trace would change. However, I do see the (small) possibility of the pointer changing between the stack trace dump and the formatting of `needle`.

A different identifier needs to be stored both inside `coroutineState` and passed as an argument to `s.run(...)`. The stack trace truncates UUIDs passed as function arguments, so it would have to be something smaller. Right now, the stack trace code reads 8 bytes of each argument, so we could use a monotonic, wrapping `uint64` counter.

I personally like the pointer approach because its completely free until a stack trace is taken, but I'm also open to implementing a different approach (like the one above). Let me know what you prefer.

internal/workflow_deadlock_test.go
Quinn-With-Two-Ns: Yes that's what I meant, that makes sense.

internal/internal_workflow.go
Quinn-With-Two-Ns: I'm fine leaving it using the pointer for now. If we get lots of reports of "<coroutine stack not found>" then we can investigate changing the implementation.

TheJokr: I added a comment to the relevant error-return in case the race condition does start to occur. Go's GC is currently non-moving, so it shouldn't for the time being: https://tip.golang.org/doc/gc-guide#Tracing_Garbage_Collection

TheJokr:
Updated to latest `master` and refactored `getCoroStackTrace`: The function now returns proper errors instead of embedding them in the stack trace string. Additionally, I got rid of `bytes.Split(stacks, ...)` to avoid the large slice allocation if there are many active goroutines.

Quinn-With-Two-Ns (approved):
LGTM, thank you for the  contribution.  Will wait to merge to see if  @cretz has any thoughts.

cretz (approved):
LGTM, no extra thoughts @Quinn-With-Two-Ns - if you're good with it, I am.

Issue #1357, opened 2024-01-16
SDK should evict workflow execution from cache on failure to respond to a workflow task


Quinn-With-Two-Ns:
## Expected Behavior

SDK should evict workflow execution from cache on failure to respond to a workflow task

## Actual Behavior

SDK may not evict workflow execution from cache on failure to respond to a workflow task

PR #1358, opened 2024-01-16
Evict workflow from cache on RespondTaskCompleted failure

Quinn-With-Two-Ns:
Evict workflow from cache on RespondTaskCompleted failure

closes https://github.com/temporalio/sdk-go/issues/1357

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -379,6 +379,8 @@ func (wtp *workflowTaskPoller) processWorkflowTask(task *workflowTask) error {
 		}
 		response, err := wtp.RespondTaskCompletedWithMetrics(completedRequest, taskErr, task.task, startTime)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
Wow, crazy that was broken this whole time. Did this work with something like sending a complete where an activity is missing any specified timeouts? That's a common case where completing the task "fails" but I can't remember if that counts as the call failing or succeeds but with an error inside. Might be worth adding an integ test for that too since it's easy.

Quinn-With-Two-Ns:
>Did this work with something like sending a complete where an activity is missing any specified timeouts?

That wouldn't cause any issue before because when receiving the new workflow task the SDK would detect the cache as stale and evict at that point. This only causes issues in very specific  scenarios that manage to bypass the SDKs cache checks and those situations are very hard to trigger with an integration test.

PR #1361, opened 2024-01-17
regenerate mocks with latest mockery

yinsidij:
## What was changed

regenerate mocker for interfaces for existing mockers under `mocks`
with https://github.com/temporalio/sdk-go/pull/1353, this PR updates all existing mockers (i.e. ones under `mockers` folder)

## Why?

It allows mocker to support returning function beside value.

For example, with the newly generated mock, we could do following that provides more flexibility.

	mockClient.On("ExecuteWorkflow", mock.Anything, mock.AnythingOfType("internal.StartWorkflowOptions"), mock.AnythingOfType("string"), mock.Anything).Return(
		func(ctx context.Context, options client.StartWorkflowOptions, workflow interface{}, args ...interface{}) (client.WorkflowRun, error) {
                 // logics based on different arguments and return
		},
	).Maybe()

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


mocks/HistoryEventIterator.go
```diff
@@ -22,12 +22,16 @@
...
 
-// Code generated by mockery v1.0.0. DO NOT EDIT.
+// Code generated by mockery v1.0.0.
+// Modified manually for type alias to work correctly.
+// https://github.com/vektra/mockery/issues/236
+
 package mocks
 
 import (
+	"go.temporal.io/api/history/v1"
```
Quinn-With-Two-Ns, line 32: It  is our convention to name proto modules *pb like `historypb`. We should stick to that convention here

yinsidij: @Quinn-With-Two-Ns thanks. Iterated.

Quinn-With-Two-Ns (approved):

PR #1362, opened 2024-01-17
Exit coroutine in order

Quinn-With-Two-Ns:
Run coroutine exit in sequence to avoid race detector errors on defers in coroutines.

closes https://github.com/temporalio/sdk-go/issues/1235

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -48,7 +48,8 @@ import (
 )
 
 const (
-	defaultSignalChannelSize = 100000 // really large buffering size(100K)
+	defaultSignalChannelSize    = 100000 // really large buffering size(100K)
+	defaultCoroutineExitTimeout = 100 * time.Millisecond
 
 	panicIllegalAccessCoroutineState = "getState: illegal access from outside of workflow context"
 )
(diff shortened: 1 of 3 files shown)


internal/internal_workflow.go
```diff
@@ -1182,16 +1194,14 @@ func (d *dispatcherImpl) Close() {
...
 	d.mutex.Unlock()
-	// This loop breaks our expectation that only one workflow
-	// coroutine is running at any time because it triggers all workflow goroutines
-	// to call their defers at once. Adding synchronization seemed more problematic because
-	// it could block eviction if there was a deadlock.
-	for i := 0; i < len(d.coroutines); i++ {
-		c := d.coroutines[i]
-		if !c.closed.Load() {
-			c.exit()
+	// We need to exit the coroutines in a separate goroutine because:
```
cretz, line 1197: So to confirm, `exit` has always been non-blocking? Or was the send `c.unblock` blocking (even though it obviously didn't wait for code to complete)? Can you confirm that returning from dispatcher `Close()` before the `exit`/`unblock`s have even been sent is safe?

Quinn-With-Two-Ns: >So to confirm, exit has always been non-blocking

Yes

> Can you confirm that returning from dispatcher Close() before the exit/unblocks have even been sent is safe?

It should be as safe as returning after they are sent, but not processed (the old behaviour)

cretz (approved):
LGTM. A little scary, but definitely a good thing for predictable exit order. I can't see any obvious dangers that weren't always there.

PR #1366, opened 2024-01-26
Exclude codes.DeadlineExceeded from the list of retryable gRPC codes

alexshtin:
## What was changed

Exclude `codes.DeadlineExceeded` from the list of retryable gRPC codes.

## Why?

gRPC library doesn't retry "context errors" (`codes.DeadlineExceeded` and `codes.Canceled`), so remove them from the list of retryable codes to eliminate the confusion.

`isRetriable`: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/retry/retry.go#L269

`isContextError`: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/retry/retry.go#L283


Quinn-With-Two-Ns:
This list of errors is common across SDKs, do we need to update them as well?

alexshtin:
For that you need to check gRPC library implementation in those langs. I hope it is consistent there.

cretz:
> gRPC library doesn't retry "context errors" (codes.DeadlineExceeded and codes.Canceled)

`DeadlineExceeded` can be a client context error or a server error. I am a bit worried to assume they are only "context errors".

alexshtin:
I think this should be checked/fixed on server. Server should never explicitly return CDE if user context is not timed out yet. I checked persistence layer, and CDE is always replaced with internal error (`persistence.TimeoutError`) which then becomes `codes.Unknown` on gRPC layer. There are other places when CDE for some internal context might be generated and passed up as user context deadline exceeded.

cretz:
To confirm this PR, since the interceptor skips retrying this code no matter what, this is just to remove the unnecessary item from our list? There should be no runtime effect, correct?

cretz (commented):
LGTM, would like @Quinn-With-Two-Ns approval before merge.

alexshtin:
Yes, it doesn't change behaviour. CDE was never retried (I've just discovered this today). So I decided to align our codes with gRPC lib implementation.

Quinn-With-Two-Ns (approved):

PR #1368, opened 2024-01-29
Typed Search Attributes

### LINKED ISSUE #1218: [Feature Request] Typed search attributes
Describe the solution you'd like:

Implementation of [typed search attributes](https://github.com/temporalio/proposals/blob/master/sdk-typed-search-attributes.md).

cretz:
## What was changed

Implementation of [typed search attributes](https://github.com/temporalio/proposals/blob/master/sdk-typed-search-attributes.md). Specifically:

* Added `TypedSearchAttributes` to `client.StartWorkflowOptions` and deprecated `SearchAttributes`
* Added several new things to `temporal` package including:
  * `SearchAttributes` collection with methods and constructor
  * `SearchAttributeKey` and `SearchAttributeUpdate` interfaces
  * `SearchAttributeKeyX` types and constructors for typed keys
* Added `TypedSearchAttributes` to `client.ScheduleWorkflowAction` and 💥removed `SearchAttributes` and added `UntypedSearchAttributes` (needed until https://github.com/temporalio/temporal/issues/4787 is fixed)
* Added `TypedSearchAttributes` to `client.ScheduleOptions` and deprecated `SearchAttributes`
* Added `GetTypedSearchAttributes` to `workflow` and deprecated `WorkflowInfo.SearchAttributes`
* Added `UpsertTypedSearchAttributes` to `workflow` and deprecated `UpsertSearchAttributes`
* Tests

Much of this work was done by @Quinn-With-Two-Ns

### 💥 BREAKING CHANGE

Users that used `SearchAttributes` on `client.ScheduleWorkflowAction` will now have a compile-time failure (not to be confused with search attributes on the schedule itself). There was no good way to have both untyped and typed coexist for create, describe, and update of a workflow action on a schedule. The release notes should document this break intentionally. This is the same decision we reached [in Python](https://github.com/temporalio/sdk-python/pull/366).

## Checklist

1. Closes #1218

*NOTE:* Currently there are two dev-server bugs (lack of `TemporalScheduledX` search attrs in workflow and a case where a SA name is appearing as `Text01`) that I'm hoping have since been fixed and are awaiting CLI release. That's why CI is failing, but can still review.

### DIFF
--- a/internal/internal_search_attributes.go
+++ b/internal/internal_search_attributes.go
@@ -297,7 +297,7 @@ func NewSearchAttributes(attributes ...SearchAttributeUpdate) SearchAttributes {
 }
 
 // GetString gets a value for the given key and whether it was present.
-func (sa *SearchAttributes) GetString(key SearchAttributeKeyString) (string, bool) {
+func (sa SearchAttributes) GetString(key SearchAttributeKeyString) (string, bool) {
 	value, ok := sa.untypedValue[key]
 	if !ok {
 		return "", false
@@ -306,7 +306,7 @@ func (sa *SearchAttributes) GetString(key SearchAttributeKeyString) (string, boo
 }
 
 // GetKeyword gets a value for the given key and whether it was present.
-func (sa *SearchAttributes) GetKeyword(key SearchAttributeKeyKeyword) (string, bool) {
+func (sa SearchAttributes) GetKeyword(key SearchAttributeKeyKeyword) (string, bool) {
 	value, ok := sa.untypedValue[key]
 	if !ok {
 		return "", false
@@ -315,7 +315,7 @@ func (sa *SearchAttributes) GetKeyword(key SearchAttributeKeyKeyword) (string, b
 }
 
 // GetBool gets a value for the given key and whether it was present.
-func (sa *SearchAttributes) GetBool(key SearchAttributeKeyBool) (bool, bool) {
+func (sa SearchAttributes) GetBool(key SearchAttributeKeyBool) (bool, bool) {
 	value, ok := sa.untypedValue[key]
 	if !ok {
 		return false, false
@@ -324,7 +324,7 @@ func (sa *SearchAttributes) GetBool(key SearchAttributeKeyBool) (bool, bool) {
 }
 
 // GetInt64 gets a value for the given key and whether it was present.
-func (sa *SearchAttributes) GetInt64(key SearchAttributeKeyInt64) (int64, bool) {
+func (sa SearchAttributes) GetInt64(key SearchAttributeKeyInt64) (int64, bool) {
 	value, ok := sa.untypedValue[key]
 	if !ok {
 		return 0, false
@@ -333,7 +333,7 @@ func (sa *SearchAttributes) GetInt64(key SearchAttributeKeyInt64) (int64, bool) 
 }
 
 // GetFloat64 gets a value for the given key and whether it was present.
-func (sa *SearchAttributes) GetFloat64(key SearchAttributeKeyFloat64) (float64, bool) {
+func (sa SearchAttributes) GetFloat64(key SearchAttributeKeyFloat64) (float64, bool) {
 	value, ok := sa.untypedValue[key]
 	if !ok {
 		return 0.0, false
@@ -342,7 +342,7 @@ func (sa *SearchAttributes) GetFloat64(key SearchAttributeKeyFloat64) (float64, 
 }
 
 // GetTime gets a value for the given key and whether it was present.
-func (sa *SearchAttributes) GetTime(key SearchAttributeKeyTime) (time.Time, bool) {
+func (sa SearchAttributes) GetTime(key SearchAttributeKeyTime) (time.Time, bool) {
 	value, ok := sa.untypedValue[key]
 	if !ok {
 		return time.Time{}, false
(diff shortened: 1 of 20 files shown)


workflow/workflow.go
```diff
@@ -571,11 +577,37 @@ func GetLastError(ctx Context) error {
...
+// Deprecated: use [UpsertTypedSearchAttributes] instead.
+//
 // [Visibility]: https://docs.temporal.io/visibility
 func UpsertSearchAttributes(ctx Context, attributes map[string]interface{}) error {
 	return internal.UpsertSearchAttributes(ctx, attributes)
 }
 
+// UpsertTypedSearchAttributes is used to add, update, or remove workflow search attributes. The search attributes can
+// be used in query of List/Scan/Count workflow APIs. The key and value type must be registered on temporal server side;
+// The value has to deterministic when replay; The value has to be Json serializable.
```
Quinn-With-Two-Ns, line 589: `The value has to deterministic when replay`

It does?

cretz: I just took this doc from `UpsertSearchAttributes`. But I just confirmed that we only check event type for upsert. So I will remove this statement from both Godocs on next commit (still going to wait for CLI publish to see if some CLI errors are solved).

Quinn-With-Two-Ns:
What happens if I mix and match upserting typed and untyped search attributes and what do the typed and untyped getters return ?

internal/internal_search_attributes.go
```diff
@@ -0,0 +1,535 @@
...
+	// SearchAttributeKeyTime represents a search attribute key for a date time attribute type.
+	SearchAttributeKeyTime struct {
+		baseSearchAttributeKey
+	}
+
+	// SearchAttributeKeyKeywordList represents a search attribute key for a list of keyword attribute type.
+	SearchAttributeKeyKeywordList struct {
+		baseSearchAttributeKey
+	}
+)
```
Sushisource, line 97: Can we do a `SearchAttributeKeyRawPayload`? As I found while doing Omes stuff, being able to pass though payloads is very convenient in some cases. We can add a docstring to the effect of "this is for advanced use cases only"

cretz: We can, though I'm not as big of a fan of us adding tools for ourselves only to save a `switch` or similar. @Quinn-With-Two-Ns - thoughts? (also worth discussing whether it's this PR or another)

cretz:
> What happens if I mix and match upserting typed and untyped search attributes

Can you clarify the scenario a bit better? Do you mean for something like workflow start? Or workflow use?

> and what do the typed and untyped getters return ?

Untyped getter returns what it always had, no behavior change there. Typed returns only type-present search attributes which is supposed to be all of them.

internal/internal_search_attributes.go
Sushisource: I don't think it's only useful for us - it's useful when doing something like copying data from one workflow to another, or enabling things like the use case we were talking about where someone might scrape a workflow's history and apply things in it to the current workflow.

cretz: That is a general problem with the inability to work with raw values in Go. The goal here was to not be able to provide the incorrect type even if you wanted to. But I don't mind the "raw" key if @Quinn-With-Two-Ns also agrees. Also open to other ideas that doesn't put it on par with all other properly-typed keys.

Quinn-With-Two-Ns: Hm to copy from one workflow to another can't a user do `SearchAttributes.Copy()` ? If the only use is out Omes DSL I wouldn't add "raw" keys. If we did want it it should be in all languages not just Go, so maybe we should move the discussion out of this PR?

Quinn-With-Two-Ns:
>Can you clarify the scenario a bit better? Do you mean for something like workflow start? Or workflow use?

In a workflow someone calling a mix of the typed and untyped upsert

>Untyped getter returns what it always had, no behavior change there. Typed returns only type-present search attributes which is supposed to be all of them.

So if a user uses the typed API to upsert will the result not be present in the untyped getter?

cretz:
> In a workflow someone calling a mix of the typed and untyped upsert

Each would result in independent commands. Untyped behaves as it always has, typed behaves as expected. I'm not sure we need to solve for advanced cases of mixing these in the same task.

test/integration_test.go
```diff
@@ -1221,6 +1221,15 @@ func (ts *IntegrationTestSuite) TestWorkflowWithParallelMutableSideEffects() {
 	ts.NoError(ts.executeWorkflow("test-wf-parallel-mutable-side-effects", ts.workflows.WorkflowWithParallelMutableSideEffects, nil))
 }
 
+func (ts *IntegrationTestSuite) TestWorkflowTypedSearchAttributes() {
```
Quinn-With-Two-Ns, line 1224: Can we have a test that confirms `TemporalChangeVersion` is set properly and accessible with typed search attributes?

cretz: :+1: Will add

internal/internal_search_attributes.go
Sushisource: Well they may not necessarily be other SAs is the problem. It could be input to the workflow, from a signal, etc, which you're accepting as a Payload

workflow/workflow.go
cretz: Updated

internal/internal_search_attributes.go
Quinn-With-Two-Ns: @Sushisource can we open a features issue to discuss this? I think it's not included in Chads original proposal for TS attributes https://github.com/temporalio/proposals/pull/75 and if we did do it I'd want it in all langs that make sense not just Go.

Quinn-With-Two-Ns (approved):

Issue #887, opened 2022-08-22
TestSuite: Workflow names seemingly collide with activity names in mock expectations


krousey:
I have a set of workflows and activities implemented as methods on a struct. I happened to name an activity method with the same name as the workflow method. Everything actually works when running the workflow, but the test framework errors out complaining about different amount of return values. The workflow method only returns an error while the activity method returns a value and an error.

If I change the name of the activity method in all places, the tests start passing.

## Expected Behavior

Passing tests.

## Actual Behavior

```console
go test
--- FAIL: Test_Workflow (0.00s)
    workflows_test.go:20: 
        	Error Trace:	/tmp/example/workflows_test.go:20
        	Error:      	Received unexpected error:
        	            	workflow execution error (type: GenericName, workflowID: default-test-workflow-id, runID: default-test-run-id): mock of GenericName has incorrect number of returns, expected 1, but actual is 2
        	Test:       	Test_Workflow
FAIL
exit status 1
FAIL	example	0.003s
```

## Steps to Reproduce the Problem

workflows.go:
```go
package example

import (
	"context"
	"time"

	"go.temporal.io/sdk/workflow"
)

type Workflows struct{}

func (w *Workflows) GenericName(ctx workflow.Context, input string) error {
	return workflow.ExecuteActivity(
		workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
			StartToCloseTimeout: time.Minute,
		}),
		Activities{}.GenericName,
		input).Get(ctx, nil)

}

type Activities struct{}

func (a Activities) GenericName(ctx context.Context, input string) (string, error) {
	return input, nil
}
```

workflows_test.go:
```go
package example

import (
	"testing"

	"github.com/stretchr/testify/mock"
	"github.com/stretchr/testify/require"
	"go.temporal.io/sdk/testsuite"
)

func Test_Workflow(t *testing.T) {
	testSuite := testsuite.WorkflowTestSuite{}
	env := testSuite.NewTestWorkflowEnvironment()

	env.OnActivity(new(Activities).GenericName, mock.Anything, "input").Return("output", nil)

	env.ExecuteWorkflow(new(Workflows).GenericName, "input")

	require.True(t, env.IsWorkflowCompleted())
	require.NoError(t, env.GetWorkflowError())
}
```
  
## Specifications

 go.mod:
```go
module example

go 1.18

require (
	github.com/stretchr/testify v1.8.0
	go.temporal.io/sdk v1.16.0
)

require (
[...]

PR #1371, opened 2024-01-30
Fix testing suite name conflict for workflow and activity (#887)

itechdima:
## What was changed

Splitted activity and workflow mocks so you're able to use same names to mock workflows and activities, but the logic of all exported functions kept the same

## Why?

To be able to use same workflow and activity function names when they are struct methods (avoid tautology, etc.)

## Checklist

1. Closes 
#887 

2. How was this tested:

Added `Test_SameWorkflowAndActivityNames` function that uses same workflow and activity name

3. Any docs updates needed?

No

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -29,6 +29,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/stretchr/testify/mock"
@@ -62,8 +63,9 @@ type (
 
 	// TestWorkflowEnvironment is the environment that you use to test workflow
 	TestWorkflowEnvironment struct {
-		mock mock.Mock
-		impl *testWorkflowEnvironmentImpl
+		workflowMock mock.Mock
+		activityMock mock.Mock
+		impl         *testWorkflowEnvironmentImpl
 	}
 
 	// TestActivityEnvironment is the environment that you use to test activity
@@ -269,7 +271,7 @@ func (e *TestWorkflowEnvironment) RegisterWorkflow(w interface{}) {
 
 // RegisterWorkflowWithOptions registers workflow implementation with the TestWorkflowEnvironment
 func (e *TestWorkflowEnvironment) RegisterWorkflowWithOptions(w interface{}, options RegisterWorkflowOptions) {
-	if len(e.mock.ExpectedCalls) > 0 {
+	if len(e.workflowMock.ExpectedCalls) > 0 {
 		panic("RegisterWorkflow calls cannot follow mock related ones like OnWorkflow or similar")
 	}
 	e.impl.RegisterWorkflowWithOptions(w, options)
@@ -282,7 +284,7 @@ func (e *TestWorkflowEnvironment) RegisterActivity(a interface{}) {
 
 // RegisterActivityWithOptions registers activity implementation with TestWorkflowEnvironment
 func (e *TestWorkflowEnvironment) RegisterActivityWithOptions(a interface{}, options RegisterActivityOptions) {
-	if len(e.mock.ExpectedCalls) > 0 {
+	if len(e.activityMock.ExpectedCalls) > 0 {
 		panic("RegisterActivity calls cannot follow mock related ones like OnActivity or similar")
 	}
 	e.impl.RegisterActivityWithOptions(a, options)
@@ -361,7 +363,7 @@ func (e *TestWorkflowEnvironment) OnActivity(activity interface{}, args ...inter
 		}
 		fnName := getActivityFunctionName(e.impl.registry, activity)
 		e.impl.registry.RegisterActivityWithOptions(activity, RegisterActivityOptions{DisableAlreadyRegisteredCheck: true})
-		call = e.mock.On(fnName, args...)
+		call = e.activityMock.On(fnName, args...)
 
 	case reflect.String:
 		name := activity.(string)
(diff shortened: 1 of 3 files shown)


internal/workflow_testsuite.go
```diff
@@ -942,16 +951,28 @@ func (e *TestWorkflowEnvironment) AssertExpectations(t mock.TestingT) bool {
...
+	}
+	return true
 }
 
 // AssertNotCalled asserts that the method was not called with the given arguments.
 // See AssertCalled for more info.
 func (e *TestWorkflowEnvironment) AssertNotCalled(t mock.TestingT, methodName string, arguments ...interface{}) bool {
-	return e.mock.AssertNotCalled(t, methodName, arguments...)
+	dummyT := &testing.T{}
+	if !(e.workflowMock.AssertNotCalled(dummyT, methodName, arguments...) || e.activityMock.AssertNotCalled(dummyT, methodName, arguments...)) {
```
Quinn-With-Two-Ns, line 965: Did you mean to check `activityMock` here?

internal/workflow_testsuite.go
```diff
@@ -942,16 +951,28 @@ func (e *TestWorkflowEnvironment) AssertExpectations(t mock.TestingT) bool {
...
+		return e.workflowMock.AssertNotCalled(t, methodName, arguments...) && e.activityMock.AssertNotCalled(t, methodName, arguments...)
+	}
+	return true
 }
 
 // AssertNumberOfCalls asserts that a method was called expectedCalls times.
 func (e *TestWorkflowEnvironment) AssertNumberOfCalls(t mock.TestingT, methodName string, expectedCalls int) bool {
-	return e.mock.AssertNumberOfCalls(t, methodName, expectedCalls)
+	dummyT := &testing.T{}
+	if !(e.workflowMock.AssertNumberOfCalls(dummyT, methodName, expectedCalls) || e.activityMock.AssertNumberOfCalls(dummyT, methodName, expectedCalls)) {
```
Quinn-With-Two-Ns, line 974: Did you mean to check `activityMock` here?

internal/workflow_testsuite.go
```diff
@@ -942,16 +951,28 @@ func (e *TestWorkflowEnvironment) AssertExpectations(t mock.TestingT) bool {
...
+	dummyT := &testing.T{}
+	if !(e.workflowMock.AssertCalled(dummyT, methodName, arguments...) || e.activityMock.AssertCalled(dummyT, methodName, arguments...)) {
+		return e.workflowMock.AssertCalled(t, methodName, arguments...) && e.activityMock.AssertCalled(t, methodName, arguments...)
+	}
+	return true
 }
 
 // AssertNotCalled asserts that the method was not called with the given arguments.
 // See AssertCalled for more info.
 func (e *TestWorkflowEnvironment) AssertNotCalled(t mock.TestingT, methodName string, arguments ...interface{}) bool {
```
Quinn-With-Two-Ns, line 963: I think this fix is acceptable. I don't believe it is backwards compatible technically but if fixing this breaks their test the test was probably wrong to begin with.

Do you think we need workflow and activity versions of this and `AssertNumberOfCalls`  that differentiate between activity and workflow mocks?

Quinn-With-Two-Ns: also  `AssertNumberOfCalls`

Quinn-With-Two-Ns: @cretz in the issue you mention backwards compatibility concerns. Did you have other concerns then these 3 functions?

itechdima: Do you mean `AssertActivityCalled`/`AssertWorkflowCalled` and others? 
We might add them as well with keeping general AssertCalled (and others)

cretz: > Did you have other concerns then these 3 functions?

That was my only concern I can recall, our high-level assertion calls. I just want to make sure we don't break anyone's existing testing code.

> Do you mean AssertActivityCalled/AssertWorkflowCalled and others?
> We might add them as well with keeping general AssertCalled (and others)

Hrmm, I don't have a strong opinion here. I am guessing we don't want to give `mock.Mock` to the user, so I guess we might as well expose.

itechdima: > I just want to make sure we don't break anyone's existing testing 

I don't know the best approach here, but I can agree with @Quinn-With-Two-Ns that if these changes break anyones test, the test itself was logically wrong

Another point is that temporal allows you to use same names for activity and workflows, but SDK doesn't allow you to test it :)

Quinn-With-Two-Ns: @cretz to be clear I think this change could cause some tests that passed before to fail, but those tests depended on workflow and activity mocks overriding each other which is a bug. I am also not sure how popular these functions even are. I think it is worth the risk of breaking some bad tests to fix this weird footgun for users.

cretz: Yeah, I definitely don't mind if someone was relying on the name ambiguity and they break. But other than that, if we feel proper tests won't break, I'm good.

Quinn-With-Two-Ns:
@itechdima I am seeing very consistent CI failures do you know what causes them?

itechdima:
@Quinn-With-Two-Ns 
Hope that I fixed them :)

itechdima:
I'm not sure why another test has failed..
could it be flaky test?

cretz:
Rerunning to see

itechdima:
@cretz 
Thanks :) 
Looks like there were some flakiness

Quinn-With-Two-Ns (approved):

PR #1373, opened 2024-02-01
Handle nil valuePtr case in update handler Get

Quinn-With-Two-Ns:
Avoid passing a `nil` to `EncocedValue.Get`. This matches the behaviour of other `Get` calls in the SDK


dandavison (approved):
LGTM!

cretz (approved):

PR #1374, opened 2024-02-01
Release Go SDK version 1.26.0-RC.1

Quinn-With-Two-Ns:
Release Go SDK 1.26.0-rc.1

Doing to do a release candidate since there are multiple breaking changes in this release.


internal/version.go
```diff
@@ -30,7 +30,7 @@ package internal
 const (
 	// SDKVersion is a semver (https://semver.org/) that represents the version of this Temporal GoSDK.
 	// Server validates if SDKVersion fits its supported range and rejects request if it doesn't.
-	SDKVersion = "1.25.1"
+	SDKVersion = "1.26.0"
```
Quinn-With-Two-Ns, line 33: Note I did the full release here, possible we should put `-rc.1` that implies we trust what parses this to follow semvar

cretz (approved):

internal/version.go
cretz: I think what you have here now is ok too.

Issue #1375, opened 2024-02-02
Data race due to missing lock/unlock


nghiant3223:
## Expected Behavior
No data race in writing/reading stick cache size.

## Actual Behavior
Data race sometimes happens. This behavior is flaky.

```
==================
WARNING: DATA RACE
Read at 0x0001043c4998 by goroutine 8:
  go.temporal.io/sdk/internal.NewWorkerCache()
      /Users/nghianguyen/go/pkg/mod/go.temporal.io/sdk@v1.20.0/internal/internal_worker_cache.go:86 +0x3d8
  go.temporal.io/sdk/internal.NewAggregatedWorker()
      /Users/nghianguyen/go/pkg/mod/go.temporal.io/sdk@v1.20.0/internal/internal_worker.go:1426 +0x3b0
  go.temporal.io/sdk/internal.NewWorker()
      /Users/nghianguyen/go/pkg/mod/go.temporal.io/sdk@v1.20.0/internal/worker.go:275 +0x80
  go.temporal.io/sdk/worker.New()
      /Users/nghianguyen/go/pkg/mod/go.temporal.io/sdk@v1.20.0/worker/worker.go:245 +0x9c
  github.com/temporal/test.TestTemporal.func2()
      /Users/nghianguyen/Desktop/tsworkspace/temporal/test/temporal_test.go:32 +0x3c
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1576 +0x188
  testing.(*T).Run.func1()
      /usr/local/go/src/testing/testing.go:1629 +0x40

Previous write at 0x0001043c4998 by goroutine 7:
  go.temporal.io/sdk/internal.SetStickyWorkflowCacheSize()
      /Users/nghianguyen/go/pkg/mod/go.temporal.io/sdk@v1.20.0/internal/internal_worker_cache.go:69 +0x6c
  go.temporal.io/sdk/worker.SetStickyWorkflowCacheSize()
      /Users/nghianguyen/go/pkg/mod/go.temporal.io/sdk@v1.20.0/worker/worker.go:276 +0x40
  github.com/temporal/test.TestTemporal.func1()
      /Users/nghianguyen/Desktop/tsworkspace/temporal/test/temporal_test.go:15 +0x38
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1576 +0x188
  testing.(*T).Run.func1()
      /usr/local/go/src/testing/testing.go:1629 +0x40

[...]

PR #1377, opened 2024-02-03
Fix data race on desiredWorkflowCacheSize

nghiant3223:
## What was changed
Add lock/unlock to `desiredWorkflowCacheSize` in `func NewWorkerCache() *WorkerCache`.

## Why?

Prevent data race.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/1375

2. How was this tested: Run the test in ticket description to verify.


cretz (approved):
LGTM, will let @Quinn-With-Two-Ns weigh in.

As mentioned on the other PR, if you are hitting this race, it means whatever you're doing is trying to update a cache variable that is already being used in its one spot, so you have bigger problems. Cache size should only be set _before_ everything.

Quinn-With-Two-Ns:
+1 to everything @cretz said

Quinn-With-Two-Ns (approved):

PR #1383, opened 2024-02-08
Introduce a new constructor to override retry policy in ContinueAsNewError

### LINKED ISSUE #1384: Add an option to override retry policy
For a long running workflows that is restarted periodically using `NewContinueAsNewError`, the existing retry policy is propagated to the new workflow. This behavior is fine until we need to adjust the retry policy. Without such an option, the workflow needs to be re-submitted using `StartWorkflowOptions`.

Note that this option is already available in `sdk-java`: https://github.com/temporalio/sdk-java/blob/72ebff1dcb2c14b3aecc7de14a2b9aa21ef9cc4a/temporal-sdk/src/main/java/io/temporal/workflow/ContinueAsNewOptions.java#L95

Similar to `WithWorkflowRunTimeout`, provide an option to override the retry policy.

**Additional context**

jiezhang:
## What was changed
Introduce a new constructor, `NewContinueAsNewErrorWithOptions `, to override retry policy in ContinueAsNewError.

## Why?
For a long running workflow that is restarted periodically using `NewContinueAsNewError`, this new option could be used to override the existing `RetryPolicy`. Without this feature, we would need re-submit the workflow using `StartWorkflowOptions`.

Note that this option is already available in `sdk-java`: https://github.com/temporalio/sdk-java/blob/72ebff1dcb2c14b3aecc7de14a2b9aa21ef9cc4a/temporal-sdk/src/main/java/io/temporal/workflow/ContinueAsNewOptions.java#L95

## Checklist

1. Closes #1384 

2. How was this tested:
```
go run . unit-test -run Test_ContinueAsNewError
=== RUN   Test_ContinueAsNewError
--- PASS: Test_ContinueAsNewError (0.00s)
=== RUN   Test_ContinueAsNewErrorWithOptions
--- PASS: Test_ContinueAsNewErrorWithOptions (0.00s)
PASS

go run . integration-test -run TestIntegrationSuite/TestContinueAsNew
--- PASS: TestIntegrationSuite (19.30s)
    --- PASS: TestIntegrationSuite/TestContinueAsNew (4.08s)
    --- PASS: TestIntegrationSuite/TestContinueAsNewCarryOver (4.03s)
    --- PASS: TestIntegrationSuite/TestContinueAsNewWithRetryPolicy (4.05s)
    --- PASS: TestIntegrationSuite/TestContinueAsNewWithWithChildWF (6.09s)
PASS

go run . integration-test -run TestIntegrationSuite/TestChildWFWithRetryPolicy
--- PASS: TestIntegrationSuite (11.31s)
    --- PASS: TestIntegrationSuite/TestChildWFWithRetryPolicy_LongRunning (4.15s)
    --- PASS: TestIntegrationSuite/TestChildWFWithRetryPolicy_LongRunningWithCustomRetry (6.07s)
    --- PASS: TestIntegrationSuite/TestChildWFWithRetryPolicy_ShortLived (0.04s)
PASS
```

3. Any docs updates needed?
N/A

### DIFF
--- a/internal/error.go
+++ b/internal/error.go
@@ -206,7 +206,7 @@ type (
 		// 2. Unlike other options that can be overridden using WithWorkflowTaskQueue, WithWorkflowRunTimeout, etc.
 		//    we can't introduce an option, say WithWorkflowRetryPolicy, for backward compatibility.
 		//    See #676 or IntegrationTestSuite::TestContinueAsNewWithWithChildWF for more details.
-		RetryPolicy *commonpb.RetryPolicy
+		RetryPolicy *RetryPolicy
 	}
 
 	// ContinueAsNewErrorOptions specifies optional attributes to be carried over to the next run.
@@ -482,7 +482,7 @@ func NewContinueAsNewErrorWithOptions(ctx Context, options ContinueAsNewErrorOpt
 	err := NewContinueAsNewError(ctx, wfn, args...)
 	contErr := err.(*ContinueAsNewError)
 	if options.RetryPolicy != nil {
-		contErr.RetryPolicy = convertToPBRetryPolicy(options.RetryPolicy)
+		contErr.RetryPolicy = options.RetryPolicy
 	}
 
 	return err
--- a/internal/error.go
+++ b/internal/error.go
@@ -201,9 +201,11 @@ type (
 		// It specifies the retry policy which gets carried over to the next run.
 		// If not set, the current workflow's retry policy will be carried over automatically.
 		//
-		// NOTE: Unlike other options that can be overridden using WithWorkflowTaskQueue, WithWorkflowRunTimeout, etc.
-		// we can't introduce an option, say WithWorkflowRetryPolicy, for backward compatibility.
-		// See #676 or IntegrationTestSuite::TestContinueAsNewWithWithChildWF for more details.
+		// NOTES:
+		// 1. This is always nil when returned from a client as a workflow response.
+		// 2. Unlike other options that can be overridden using WithWorkflowTaskQueue, WithWorkflowRunTimeout, etc.
+		//    we can't introduce an option, say WithWorkflowRetryPolicy, for backward compatibility.
+		//    See #676 or IntegrationTestSuite::TestContinueAsNewWithWithChildWF for more details.
 		RetryPolicy *commonpb.RetryPolicy
 	}
 
@@ -476,7 +478,7 @@ func NewContinueAsNewError(ctx Context, wfn interface{}, args ...interface{}) er
 }
 
 // NewContinueAsNewErrorWithOptions creates ContinueAsNewError instance with additional options.
-func NewContinueAsNewErrorWithOptions(ctx Context, wfn interface{}, options ContinueAsNewErrorOptions, args ...interface{}) error {
+func NewContinueAsNewErrorWithOptions(ctx Context, options ContinueAsNewErrorOptions, wfn interface{}, args ...interface{}) error {
 	err := NewContinueAsNewError(ctx, wfn, args...)
 	contErr := err.(*ContinueAsNewError)
 	if options.RetryPolicy != nil {
--- a/internal/error.go
+++ b/internal/error.go
@@ -196,6 +196,11 @@ type (
 		// VersioningIntent specifies whether the continued workflow should run on a worker with a
 		// compatible build ID or not. See VersioningIntent.
 		VersioningIntent VersioningIntent
+
+		// Propagates the retry policy to the next run of the workflow.
+		// This is usually set by StartWorkflowOptions.RetryPolicy,
+		// but it could be overridden using workflow.WithWorkflowRetryPolicy.
+		RetryPolicy *commonpb.RetryPolicy
 	}
 
 	// UnknownExternalWorkflowExecutionError can be returned when external workflow doesn't exist
@@ -447,9 +452,10 @@ func IsCanceledError(err error) bool {
 //
 //	 ctx - use context to override any options for the new workflow like run timeout, task timeout, task queue.
 //		  if not mentioned it would use the defaults that the current workflow is using.
-//	       ctx := WithWorkflowRunTimeout(ctx, 30 * time.Minute)
-//	       ctx := WithWorkflowTaskTimeout(ctx, 5 * time.Second)
-//		  ctx := WithWorkflowTaskQueue(ctx, "example-group")
+//	       ctx = WithWorkflowRunTimeout(ctx, 30 * time.Minute)
+//	       ctx = WithWorkflowTaskTimeout(ctx, 5 * time.Second)
+//	       ctx = WithWorkflowTaskQueue(ctx, "example-group")
+//	       ctx = WithWorkflowRetryPolicy(ctx, retryPolicy)
 //	 wfn - workflow function. for new execution it can be different from the currently running.
 //	 args - arguments for the new workflow.
 func NewContinueAsNewError(ctx Context, wfn interface{}, args ...interface{}) error {
@@ -489,6 +495,7 @@ func (wc *workflowEnvironmentInterceptor) NewContinueAsNewError(
 		WorkflowRunTimeout:       options.WorkflowRunTimeout,
 		WorkflowTaskTimeout:      options.WorkflowTaskTimeout,
 		VersioningIntent:         options.VersioningIntent,
+		RetryPolicy:              options.RetryPolicy,
 	}
 }
 
--- a/internal/error.go
+++ b/internal/error.go
@@ -197,10 +197,17 @@ type (
 		// compatible build ID or not. See VersioningIntent.
 		VersioningIntent VersioningIntent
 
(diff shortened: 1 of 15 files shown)


Quinn-With-Two-Ns:
Thanks for the contribution, Can you please add integration tests to verify this change? Just a simple test showing when you continue-as-new the `RetryPolicy` in `WorkflowInfo` is updated.

internal/error.go
```diff
@@ -489,6 +495,7 @@ func (wc *workflowEnvironmentInterceptor) NewContinueAsNewError(
 		WorkflowRunTimeout:       options.WorkflowRunTimeout,
 		WorkflowTaskTimeout:      options.WorkflowTaskTimeout,
 		VersioningIntent:         options.VersioningIntent,
+		RetryPolicy:              options.RetryPolicy,
```
Quinn-With-Two-Ns, line 498: This appears to be a breaking change since `options.RetryPolicy` can also be set by `workflow.WithChildWorkflowOptions`. Workflows may be setting a retry policy for a child workflow and now we may also override the parents. This a problem with other options as well, but now we are spreading it to `RetryPolicy` as well. I will need to discuss with the team if this is an acceptable change.

Quinn-With-Two-Ns: Yes, I fully understand the justification. We just take backwards compatibility very seriously and don't want to break any previously working workflows.

Quinn-With-Two-Ns: Your change will now cause that `RetryPolicy` in `WithChildWorkflowOptions ` to be used as the new `RetryPolicy` on continue as new I believe.

cretz: I think the issue is if you had these steps:

* Set `WithChildWorkflowOptions`
* Execute child workflow
* Return continue-as-new error with no retry policy

The behavior before your change was to have continue-as-new not set a retry policy. But now after this change, the retry policy will come from `WithChildWorkflowOptions` which is an incompatible change from what it did before which was no retry policy set from context options.

cretz: > I think the scenarios you mentioned are covered by

The scenario I mention combines setting retry policy for child workflow and also returning continue as new.

Take this workflow (typed in GH so I may have typos):

```go
func MyWorkflow(ctx context.Context) error {
	ctx = workflow.WithChildOptions(workflow.ChildWorkflowOptions{
		RetryPolicy: &temporal.RetryPolicy{MaximumAttempts: 5},
	})
	err := workflow.ExecuteChildWorkflow(ctx, MyChildWorkflow).Get(ctx, nil)
	if err != nil {
		return err
	}
	// This never used to use the retry policy before, but now it does
	return workflow.NewContinueAsNewError(ctx, MyWorkflow)
}
```

Before your change (i.e. how it works today), you would not be setting a retry policy on continue as new (which means it won't retry for workflows by default). But after your change, now continue as new has a retry policy of max-attempts 5. Unless I'm misunderstanding somewhere, this is the backwards incompatible change being referenced.

cretz: > Fortunately, this issue can be easily worked around by using a different context, as illustrated in TestContinueAsNewWithWithChildWF.

We have a standing rule not to break behavior compatibility of existing code on things like this. We can't just ask people to change code that works for them today and silently start using retry policy set for something else on continue as new.

We could make a `NewContinueAsNewErrorWithOptions` that accepts some options, but they might as well instantiate `&workflow.ContinueAsNewError{...}` themselves.

internal/error.go
```diff
@@ -196,6 +196,18 @@ type (
 		// VersioningIntent specifies whether the continued workflow should run on a worker with a
 		// compatible build ID or not. See VersioningIntent.
 		VersioningIntent VersioningIntent
+
+		// This is optional and may be used to override the retry policy which gets carried over to the next run.
+		// If not set, the existing retry policy will be carried over automatically.
+		// For backward compatibility, we can't introduce an option, say WithWorkflowRetryPolicy,
+		// to override the retry policy attached to the workflow context.
+		// See #676 for more details.
+		RetryPolicy *commonpb.RetryPolicy
```
cretz, line 205:
```suggestion
		RetryPolicy *RetryPolicy
```

Can we make this the user-friendly version?

internal/error.go
```diff
@@ -459,6 +471,17 @@ func NewContinueAsNewError(ctx Context, wfn interface{}, args ...interface{}) er
 	return i.NewContinueAsNewError(ctx, wfn, args...)
 }
 
+// NewContinueAsNewErrorWithOptions creates ContinueAsNewError instance with additional options.
+func NewContinueAsNewErrorWithOptions(ctx Context, wfn interface{}, options ContinueAsNewErrorOptions, args ...interface{}) error {
```
cretz, line 475:
```suggestion
func NewContinueAsNewErrorWithOptions(ctx Context, options ContinueAsNewErrorOptions, wfn interface{}, args ...interface{}) *ContinueAsNewError {
```

To match things that accept options and varargs (e.g. `Client.ExecuteWorkflow`) let's put the options before the workflow. Also, I wonder if we would like to return the actual continue as new error so they can alter things if they'd like. @Quinn-With-Two-Ns - preference?

internal/error.go
```diff
@@ -196,6 +196,18 @@ type (
...
+		// If not set, the existing retry policy will be carried over automatically.
+		// For backward compatibility, we can't introduce an option, say WithWorkflowRetryPolicy,
+		// to override the retry policy attached to the workflow context.
+		// See #676 for more details.
+		RetryPolicy *commonpb.RetryPolicy
+	}
+
+	// ContinueAsNewErrorOptions specifies optional attributes to be carried over to the next run.
+	ContinueAsNewErrorOptions struct {
+		RetryPolicy *RetryPolicy
```
cretz, line 210:
```suggestion
		// If nil, the current workflow's retry policy will be used.
		RetryPolicy *RetryPolicy
```

internal/error.go
cretz, line 205:
```suggestion
		// Note, this is always nil when returned from a client as a workflow response.
		RetryPolicy *commonpb.RetryPolicy
```

We don't seem to set this on event so it's not able to be properly set on `workflowRunImpl.GetWithOptions` like a user might expect.

cretz: > WorkflowOptions

Which workflow options? `client.StartWorkflowOptions` accepts a non-proto `RetryPolicy` as does `workflow.ChildWorkflowOptions`, but it's possible we have left some proto ones around.

> ExecuteActivityOptions

Which execute activity options? `workflow.ActivityOptions` accepts a non-proto `RetryPolicy`, but it's possible we have left some proto ones around.

internal/error.go
cretz: > the options as the last parameter:

But it isn't last, it's before `args`. The concern was "accept options _and varargs_". Would definitely support options last if it could be last, but in this case, I think it's worth the workflow and args being next to each other instead of split. Would like @Quinn-With-Two-Ns's opinion here though.

> In addition, none of the constructor within this file returns the concrete type (such as *ContinueAsNewError, *ApplicationError.

Yeah, I don't have a strong opinion here. I am ok continuing to return just `error`. Can get others' opinions too.

internal/error.go
cretz: It is always `nil` _when returned from a client_ because that is powered by workflow events, and we don't have the retry policy no that event. See `workflowRunImpl.GetWithOptions` to see the logic of how we build this error for clients (not to be confused with how this error is used within workflows).

cretz: Not all of those are user facing and it's acknowledged that we put raw input/headers. But I think most of our user-facing options sets that accept retry policy (start workflow options, child workflow options, activity options, etc) accept the non-proto form.

cretz:
Sure, done! (and I'll try to keep on top of that for each commit you make henceforth after quick scan of commit)

internal/error.go
cretz: > If it can be used interchangeably, why do we need commonpb.RetryPolicy in the first place?

Explained in Slack, but that's the proto our API recognizes but is a bit annoying for users to use so we have a higher level form.

internal/error.go
```diff
@@ -459,6 +477,20 @@ func NewContinueAsNewError(ctx Context, wfn interface{}, args ...interface{}) er
 	return i.NewContinueAsNewError(ctx, wfn, args...)
 }
 
+// NewContinueAsNewErrorWithOptions creates ContinueAsNewError instance with additional options.
+func NewContinueAsNewErrorWithOptions(ctx Context, options ContinueAsNewErrorOptions, wfn interface{}, args ...interface{}) error {
```
Quinn-With-Two-Ns, line 481: Awkward this doesn't  go through the `NewContinueAsNewError`  interceptor, but doing it properly would require changing the interceptors type signature to take  `ContinueAsNewErrorOptions`. @cretz thoughts?

cretz: It does go through the existing `NewContinueAsNewError` interceptor for the initial creation (which can set this value too), just not with the options, correct. I think that's ok for now.

cretz (approved):
This LGTM assuming @Quinn-With-Two-Ns is also ok with this

PR #1389, opened 2024-02-19
Update CI to Go lang 1.22

Quinn-With-Two-Ns:
Update CI to Go lang 1.22


cretz (approved):
I know there was something that made us concerned with 1.20 to 1.21, but if you wanted to get rid of both versions and just check the latest, I'd be ok with that too. But having both is the safest option of course.

PR #1391, opened 2024-02-19
Clarify OnUpsertTypedSearchAttributes

Quinn-With-Two-Ns:
Clarify `OnUpsertTypedSearchAttributes` does not expect the actual args to `workflow.UpsertTypedSearchAttributes` , but the result `temporal.SearchAttributes`. We can't take the actual args since we can't have mock asserts on functions.

### DIFF
--- a/internal/internal_search_attributes.go
+++ b/internal/internal_search_attributes.go
@@ -370,7 +370,7 @@ func (sa SearchAttributes) Size() int {
 func (sa SearchAttributes) GetUntypedValues() map[SearchAttributeKey]interface{} {
 	untypedValueCopy := make(map[SearchAttributeKey]interface{}, len(sa.untypedValue))
 	for key, value := range sa.untypedValue {
-		// Filer out nil values
+		// Filter out nil values
 		if value == nil {
 			continue
 		}
--- a/internal/internal_search_attributes.go
+++ b/internal/internal_search_attributes.go
@@ -287,19 +287,13 @@ func NewSearchAttributes(attributes ...SearchAttributeUpdate) SearchAttributes {
 	for _, attr := range attributes {
 		attr(&sa)
 	}
-	// remove nil values
-	for key, value := range sa.untypedValue {
-		if value == nil {
-			delete(sa.untypedValue, key)
-		}
-	}
 	return sa
 }
 
 // GetString gets a value for the given key and whether it was present.
 func (sa SearchAttributes) GetString(key SearchAttributeKeyString) (string, bool) {
 	value, ok := sa.untypedValue[key]
-	if !ok {
+	if !ok || value == nil {
 		return "", false
 	}
 	return value.(string), true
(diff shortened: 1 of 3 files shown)


internal/internal_workflow_testsuite_test.go
```diff
@@ -1812,6 +1814,7 @@ func (s *WorkflowTestSuiteUnitTest) Test_MockUpsertTypedSearchAttributes() {
 	env = s.NewTestWorkflowEnvironment()
 	env.OnUpsertTypedSearchAttributes(NewSearchAttributes()).Return(errors.New("empty")).Once()
 	env.OnUpsertTypedSearchAttributes(NewSearchAttributes(CustomIntKey.ValueSet(1))).Return(nil).Once()
+	env.OnUpsertTypedSearchAttributes(NewSearchAttributes(CustomIntKey.ValueUnset())).Return(nil).Once()
```
Quinn-With-Two-Ns, line 1817: For the mocking to work on `OnUpsertTypedSearchAttributes ` we need to keep nil values in the map

internal/workflow_testsuite.go
```diff
@@ -516,8 +516,9 @@ func (e *TestWorkflowEnvironment) OnUpsertSearchAttributes(attributes interface{
 // OnUpsertTypedSearchAttributes setup a mock for workflow.UpsertTypedSearchAttributes call.
 // If mock is not setup, the UpsertTypedSearchAttributes call will only validate input attributes.
 // If mock is setup, all UpsertTypedSearchAttributes calls in workflow have to be mocked.
-func (e *TestWorkflowEnvironment) OnUpsertTypedSearchAttributes(attributes ...interface{}) *MockCallWrapper {
-	call := e.workflowMock.On(mockMethodForUpsertTypedSearchAttributes, attributes...)
+// Note: The mock is called with a temporal.SearchAttributes constructed from the inputs to workflow.UpsertTypedSearchAttributes.
+func (e *TestWorkflowEnvironment) OnUpsertTypedSearchAttributes(attributes interface{}) *MockCallWrapper {
+	call := e.workflowMock.On(mockMethodForUpsertTypedSearchAttributes, attributes)
```
Quinn-With-Two-Ns, line 521: Alternatively we could mock on `SearchAttributesUpdates`, but I think they are harder to use since you can't be very flexible on the input assertion.

internal/internal_search_attributes.go
```diff
@@ -287,19 +287,13 @@ func NewSearchAttributes(attributes ...SearchAttributeUpdate) SearchAttributes {
 	for _, attr := range attributes {
 		attr(&sa)
 	}
-	// remove nil values
-	for key, value := range sa.untypedValue {
```
Quinn-With-Two-Ns, line 291: This check also didn't totally exclude nulls from ending up in `untypedValue ` anyway.

PR #1392, opened 2024-02-21
Make it possible to cancel workflows by ID when testing

recht:
## What was changed
Added a function in `TestWorkflowEnvironment` to allow canceling arbitrary workflows, in line with `SignalWorkflowByID` and `QueryWorkflowByID`

## Why?
This makes it possible to cancel child executions and test the behavior of doing so.


Quinn-With-Two-Ns:
Thanks for the contribution, can you please add some unit tests to verify this change? I would like to see a test like this https://github.com/temporalio/sdk-go/blob/5621b7fb1748e043c9d1a07f3b40940c402d1870/internal/internal_workflow_testsuite_test.go#L570 for the new API and a test that cancels a child workflow using the new API.

Quinn-With-Two-Ns (approved):

PR #1393, opened 2024-02-22
allow client to request keep scheduled workflow id as is.

ast2023:
## What was changed
Add option to keep scheduled workflow id as is.

## Why?
User request.

## Checklist
1. How was this tested:
Not yet tested.

### DIFF
--- a/internal/schedule_client.go
+++ b/internal/schedule_client.go
@@ -363,6 +363,10 @@ type (
 		//
 		// [Visibility]: https://docs.temporal.io/visibility
 		TypedSearchAttributes SearchAttributes
+
+		// KeepOriginalWorkflowID - If true, and the action would start a workflow, a timestamp will not be
+		// appended to the scheduled workflow id.
+		KeepOriginalWorkflowID bool
 	}
 
 	// ScheduleWorkflowExecution contains details on a workflows execution stared by a schedule.
--- a/internal/internal_schedule_client.go
+++ b/internal/internal_schedule_client.go
@@ -39,6 +39,7 @@ import (
 	taskqueuepb "go.temporal.io/api/taskqueue/v1"
 	workflowpb "go.temporal.io/api/workflow/v1"
 	"go.temporal.io/api/workflowservice/v1"
+
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/log"
 )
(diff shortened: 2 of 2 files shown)


internal/schedule_client.go
```diff
@@ -363,6 +363,10 @@ type (
 		//
 		// [Visibility]: https://docs.temporal.io/visibility
 		TypedSearchAttributes SearchAttributes
+
+		// KeepOriginalWorkflowID - If true, and the action would start a workflow, a timestamp will not be
+		// appended to the scheduled workflow id.
+		KeepOriginalWorkflowID bool
```
Quinn-With-Two-Ns, line 369: What would happen if users try to set this to true and call `CreateSchedule` on an older server version that doesn't support this field? I suspect it will succeed, but ignore the call fields correct?

internal/schedule_client.go
```diff
@@ -363,6 +363,10 @@ type (
 		//
 		// [Visibility]: https://docs.temporal.io/visibility
 		TypedSearchAttributes SearchAttributes
+
+		// KeepOriginalWorkflowID - If true, and the action would start a workflow, a timestamp will not be
```
Quinn-With-Two-Ns, line 367: Since this only applies if the action is `ScheduleWorkflowAction` is there a reason you didn't put the field on the action?

internal/schedule_client.go
ast2023: I think you are right. Protobuf allows us to ignore fields which we do not recognize. So server will just ignore the new field.

internal/schedule_client.go
ast2023: Initially it was in ScheduleAction. Then @dnr pointed out that Policy is the more logical place for this field. There is no Policy in the SDK ScheduleOptions. So I left the field at the options top level. 
How can users set Policy from the workflow code?

internal/schedule_client.go
Quinn-With-Two-Ns: We should probably document this option may not apply depending on the server version, ideally say what server version this option will be respected

internal/schedule_client.go
Quinn-With-Two-Ns: I am not sure why policy is more logical if this only applies to one type of action. @dnr do you expect this option to apply to other type of actions?

dnr: We don't really have a distinct type for the start workflow action. Maybe we should have. I think "policies" make sense as general configuration that can affect schedule execution.

For other actions: suppose we had an action to start a top-level activity. I think top-level activities would still have names, so it could apply there too. Same for starting a nexus operation, though I'm not sure about naming there.

Quinn-With-Two-Ns: hmm if this option is meant to apply to other future actions like top level activity of nexus operation then I think we should change the name to not include `Workflow` since that is specific to one type of action.

PR #1394, opened 2024-02-22
Add the ability to soft launch SDK flags

Quinn-With-Two-Ns:
Add the ability to soft launch SDK flags. This allows us to add an SDK flag in one minor release of the SDK, but not use it. This will allow us to maintain backwards compatibility if the flags is set in a later release.

### DIFF
--- a/internal/internal_flags.go
+++ b/internal/internal_flags.go
@@ -46,9 +46,16 @@ const (
 	SDKFlagProtocolMessageCommand = 3
 	// SDKPriorityUpdateHandling will cause update request to be handled before the main workflow method.
 	// It will also cause the SDK to immediately handle updates when a handler is registered.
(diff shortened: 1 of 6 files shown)


internal/internal_flags.go
```diff
@@ -46,10 +46,17 @@ const (
 	SDKFlagProtocolMessageCommand = 3
 	// SDKPriorityUpdateHandling will cause update request to be handled before the main workflow method.
 	// It will also cause the SDK to immediately handle updates when a handler is registered.
+	// This flag is currently implemented, but not set to preserve rollback safety between Go SDK 1.26.x and 1.25.x.
 	SDKPriorityUpdateHandling = 4
```
Quinn-With-Two-Ns, line 50: This flag mostly applies to update processing , so another option is I rework the flag to only apply to update processing which I think I can do. That would mean the can only not be rolled back if the User was using update so should be safe for the Server team to use. Thoughts @yiminc @bergundy ?

PR #1396, opened 2024-02-22
Make errors support verbose formatting to print wrapped errors

### LINKED ISSUE #1395: Make TestWorkflowEnvironment return errors with stack traces
If a test fails, for example because a context is used wrongly, there's almost no information available to debug, and you have to go through error unwrapping and checking in order to get to a usable stacktrace.

The usual solution is to implement a `Format` function on errors that prints more information when formatted using `%+v`. Functions like `require.NoError` also use this, so if there is a verbose version then that will be printed.

recht:
This makes it possible to print the entire hierarchy when using `%+v` - in particular, when running tests this will show causes and stack traces, not just the error message, which can help debugging tests.

Closes #1395

### DIFF
--- a/internal/error.go
+++ b/internal/error.go
@@ -27,6 +27,7 @@ package internal
 import (
 	"errors"
 	"fmt"
+	"io"
 	"reflect"
 	"strings"
 	"time"
@@ -551,6 +552,10 @@ func (e *ApplicationError) Message() string {
 	return e.msg
 }
 
+func (e *ApplicationError) Format(s fmt.State, verb rune) {
+	formatError(e, s, verb, e.Unwrap())
+}
+
 // Type returns error type represented as string.
 // This type can be passed explicitly to ApplicationError constructor.
 // Also any other Go error is converted to ApplicationError and type is set automatically using reflection.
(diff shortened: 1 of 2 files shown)


internal/error_test.go
```diff
@@ -1245,3 +1245,44 @@ func Test_convertFailureToError_SaveFailure(t *testing.T) {
...
+func Test_verbose_error_formatting(t *testing.T) {
+	tests := []struct {
+		name     string
+		err      error
+		expected string
+	}{
+		{
+			name:     "WorkflowExecutionError",
+			err:      NewWorkflowExecutionError("wid", "rid", "workflowType", newWorkflowPanicError("test message", "stack trace")),
+			expected: "stack trace\ntest message\nworkflow execution error (type: workflowType, workflowID: wid, runID: rid): test message",
```
cretz, line 1258: It seems odd to show inner error information before outer error information

internal/error.go
```diff
@@ -820,6 +845,24 @@ func (e *WorkflowExecutionError) Unwrap() error {
 	return e.cause
 }
 
+func formatError(e error, s fmt.State, verb rune, verboseValue any) {
```
cretz, line 848: I fear the future proof-ness of this. If Go adds new print flags that'd work for errors, we'd silently not support them.

I wonder if it'd be better to change where you're formatting these errors to handle more advanced types instead of adding a formatting feature to a few of the errors here. I think that gives you the most flexibility instead of assuming exactly what people want when `%+v` is put (you want stack trace, but another might want struct fields). We tend to put accessors on the errors for users to extract and display the way they prefer.

(having said that, I'm not _completely_ against the change, I think it just makes too many assumptions of what users may want to see instead of encouraging them to extract the information themselves when they're displaying)

internal/error_test.go
recht: I basically copied the behavior from github.com/pkg/errors - I did also think the same but for consistency it seemed like the best approach

PR #1403, opened 2024-02-29
Update workflow check with new function added in the SDK

Quinn-With-Two-Ns:
Update workflow check with new function added in the SDK

### DIFF
--- a/contrib/tools/workflowcheck/workflow/checker.go
+++ b/contrib/tools/workflowcheck/workflow/checker.go
@@ -45,6 +45,15 @@ var DefaultIdentRefs = determinism.DefaultIdentRefs.Clone().SetAll(determinism.I
 	// Reported as non-deterministic because it iterates over a map, so mark
 	// deterministic explicitly
 	"(*go.temporal.io/sdk/internal.cancelCtx).cancel": false,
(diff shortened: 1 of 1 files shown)


cretz (approved):
LGTM. May be worth an update to the test cases showing use of one or more of these, but doesn't matter much to me.

aaomidi:
This PR breaks workflow check.

```
workflowcheck ./...
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x104880200]

goroutine 1015 [running]:
go/types.(*Checker).handleBailout(0x140022fa200, 0x14002407b98)
        /nix/store/2022s0jnrn2iyxjaikfy51w5fvifp38b-go-1.22.0/share/go/src/go/types/check.go:367 +0x9c
panic({0x104a82be0?, 0x104c9fde0?})
        /nix/store/2022s0jnrn2iyxjaikfy51w5fvifp38b-go-1.22.0/share/go/src/runtime/panic.go:770 +0x124
 ```

PR #1404, opened 2024-02-29
Recognize SDKPriorityUpdateHandling flag

Quinn-With-Two-Ns:
Recognize, but do not handle SDKPriorityUpdateHandling flag to avoid unknown flag error. This will allow for safe downgrade  for workers using SDK version between  `1.25.1` and `1.26.0-RC.2` that set this flag unconditionally.


Quinn-With-Two-Ns:
`features-test` aren't expected to pass since this is pre `gogo/proto`

cretz (approved):

PR #1407, opened 2024-03-01
Update api-go to allow invalid UTF-8 in proto strings

tdeebswihart:
## What was changed
I've upgraded the version of our Go API in order to allow invalid UTF-8 data in our proto strings.

Note that the tests for this behavior are in a separate pr (#1408) as they require unreleased changes to the server.

## Why?
The version of gogo/protobuf we used allowed invalid UTF-8 in strings which is a violation of the proto3 spec. We were unaware of this until our new google/protobuf based structs began failing to deserialize data we'd used previously

## Checklist
- [x] Add tests
- [x] Release api-go and use that here instead of the commit sha


internal/cmd/build/main.go
```diff
@@ -88,7 +88,7 @@ func (b *builder) run() error {
 
 func (b *builder) check() error {
 	// Run go vet
-	if err := b.runCmd(b.cmdFromRoot("go", "vet", "./...")); err != nil {
+	if err := b.runCmd(b.cmdFromRoot("go", "vet", "-tags", "protolegacy", "./...")); err != nil {
```
dnr, line 91: should we add a comment to `CONTRIBUTING.md` in the root about this tag being effectively required now?

test/integration_test.go
```diff
@@ -277,6 +277,26 @@ func (ts *IntegrationTestSuite) TestBasic() {
 	ts.assertMetricCountAtLeast("temporal_long_request", 3, "operation", "PollWorkflowTaskQueue")
 }
 
+func (ts *IntegrationTestSuite) TestBasic_InvalidUTF8() {
+	var expected []string
+	err := ts.executeWorkflow("test-basic", ts.workflows.Basic, &expected)
+	ts.NoError(err)
+	ts.EqualValues(expected, ts.activities.invoked())
+	// See https://grokbase.com/p/gg/golang-nuts/153jjj8dgg/go-nuts-fm-suffix-in-function-name-what-does-it-mean
```
dnr, line 285: there is no `fm` suffix here?

test/integration_test.go
```diff
@@ -277,6 +277,26 @@ func (ts *IntegrationTestSuite) TestBasic() {
 	ts.assertMetricCountAtLeast("temporal_long_request", 3, "operation", "PollWorkflowTaskQueue")
 }
 
+func (ts *IntegrationTestSuite) TestBasic_InvalidUTF8() {
```
dnr, line 280: where is the invalid utf8 in this test?

internal/cmd/build/main.go
Quinn-With-Two-Ns: Yes we should add a note, what do you mean by "required" @dnr ?

dnr: Well, the new integration test fails without it... (maybe not the one in this pr but the one in the next pr)

I meant if they were running tests or things without using this helper. But I guess that's not supported, so maybe not worth explaining

test/integration_test.go
tdeebswihart: I'd thought I'd removed the test from this PR, let me take another look now that I've eaten lunch

internal/cmd/build/main.go
tdeebswihart: I've added it to our README

README.md
```diff
@@ -53,5 +53,19 @@ See [contrib/tools/workflowcheck](contrib/tools/workflowcheck) for a tool to det
 ## Contributing
 We'd love your help in making the Temporal Go SDK great. Please review our [contribution guidelines](CONTRIBUTING.md).
 
+## Go build and run tags
+
+Prior to SDK version v1.26.0 our protobuf code generator allowed invalid UTF-8 data to be stored as proto strings. This isn't actually allowed by the proto3 spec, so if you're building our SDK manually and think you may store arbitrary binary data in our strings you should set `-tags protolegacy` when building against our SDK.
```
Quinn-With-Two-Ns, line 58: "building our SDK manually" can we say  "using our SDK" with Go everything is always distributed as source right so the users are always building our SDK when they include it as a depdency?

README.md
```diff
@@ -53,5 +53,19 @@ See [contrib/tools/workflowcheck](contrib/tools/workflowcheck) for a tool to det
...
+
+Example:
+
+``` shell
+$ go build -tags protolegacy myworker/main.go
+```
+
+If you see don't but see an error like `grpc: error unmarshalling request: string field contains invalid UTF-8` then you will need to enable this when building your code.
```
Quinn-With-Two-Ns, line 66: "If you see don't but see an error like" -> "If you see an error like"

tdeebswihart: 🤦 I'll fix that, thanks Quinn.

README.md
tdeebswihart: You're right, that's poor phrasing on my part. That's fixed in the latest commit

Quinn-With-Two-Ns (approved):
Approving for the RC, would like tests merged before I release this change as part of an official release

PR #1409, opened 2024-03-04
API key support via credentials abstraction

### LINKED ISSUE #1401: API key client option

See https://github.com/temporalio/features/issues/426

cretz:
## What was changed

* Added `client.Credential` interface (cannot be implemented by users for now)
* Added `client.NewAPIKeyCredentials` and `client.NewAPIKeyCallbackCredentials` calls to create credentials
  * These don't accept "options" structures for future proofing like some other calls, that ok? We can add as many of these calls as needed later.
  * We chose to overwrite any existing authorization headers so there is no ambiguity (e.g. it is confusing to say "creds overwrites header provider, but context overwrites creds" IMO, they should just not set the creds if they need to overwrite). This ok?
* Added `client.NewMTLSCredentials` to simplify TLS config for client key/cert only

Yes, all of this is mostly trivial for users to do themselves (or just call `ctx = metadata.AppendToOutgoingContext(ctx, "Authorization", "Bearer " + myKey)` before each RPC call) and we still have the ways to do themselves, but one of the primary goals of this PR is to make cloud access simpler for users (though this benefits self-hosted too).

## Checklist

1. Closes #1401

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -757,7 +757,7 @@ func HistoryFromJSON(r io.Reader, options HistoryJSONOptions) (*historypb.Histor
 	return internal.HistoryFromJSON(r, options.LastEventID)
 }
 
-// NewAPIKeyCredentials creates credentials that can be provided to
+// NewAPIKeyStaticCredentials creates credentials that can be provided to
 // ClientOptions to use a fixed API key.
 //
 // This is the equivalent of providing a headers provider that sets the
@@ -767,12 +767,12 @@ func HistoryFromJSON(r io.Reader, options HistoryJSONOptions) (*historypb.Histor
 //
 // Note, this uses a fixed header value for authentication. Many users that want
 // to rotate this value without reconnecting should use
-// [NewAPIKeyCallbackCredentials].
-func NewAPIKeyCredentials(apiKey string) Credentials {
-	return internal.NewAPIKeyCredentials(apiKey)
+// [NewAPIKeyDynamicCredentials].
+func NewAPIKeyStaticCredentials(apiKey string) Credentials {
+	return internal.NewAPIKeyStaticCredentials(apiKey)
 }
 
-// NewAPIKeyCallbackCredentials creates credentials powered by a callback that
+// NewAPIKeyDynamicCredentials creates credentials powered by a callback that
 // is invoked on each request. The callback accepts the context that is given by
 // the calling user and can return a key or an error. When error is non-nil, the
 // client call is failed with that error. When string is non-empty, it is used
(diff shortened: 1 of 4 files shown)


bergundy:
Should we force TLS if API key creds are provided?

client/client.go
```diff
@@ -752,3 +756,41 @@ type HistoryJSONOptions struct {
...
+//
+// This is the equivalent of providing a headers provider that sets the
+// "Authorization" header with "Bearer " + the given key. This will overwrite
+// any "Authorization" header that may be on the context or from existing header
+// provider.
+//
+// Note, this uses a fixed header value for authentication. Many users that want
+// to rotate this value without reconnecting should use
+// [NewAPIKeyCallbackCredentials].
+func NewAPIKeyCredentials(apiKey string) Credentials {
```
bergundy, line 771: Wonder if it's worth calling this `NewAPIKeyStaticCredentials` and the callback based one `NewAPIKeyDynamicCredentials`. Not critical, just thought we may want to consider alternative names.

cretz: I don't mind alternative names, these work for me. Will leave this discussion open for others before changing.

bergundy:
> > Should we force TLS if API key creds are provided?
> 
> No, I believe header config should be completely separate from TLS config. We obviously should discourage users from accepting these headers in their self-hosted setups without TLS, but that is up to them. We expect servers requiring TLS to fail on connection regardless of headers set.

I'm just worried that it'd be too easy to leak credentials this way and that the default is TLS off, which would force users to remember to turn it on in addition to using credentials.

client/client.go
```diff
@@ -90,6 +91,9 @@ type (
 	// ConnectionOptions are optional parameters that can be specified in ClientOptions
 	ConnectionOptions = internal.ConnectionOptions
 
+	// Credentials are optional credentials that can be specified in ClientOptions.
+	Credentials = internal.Credentials
```
Quinn-With-Two-Ns, line 95: This is considered stable from the SDK perspective right? On Temporal cloud it is not yet AFAIK.

cretz:
> I'm just worried that it'd be too easy to leak credentials this way and that the default is TLS off, which would force users to remember to turn it on in addition to using credentials.

This already exists today with all headers and proto bodies and everything else. It's a handshake failure if server requires TLS and you don't use it, nothing leaked there. So your only leaking concern is about post-connection, which already exists today. I expect many people won't use this and will use per-call headers anyways. I think your concern is about TLS disabled by default and general leaking, not just leaking this one header after successful connection establishment.

client/client.go
cretz: I think so unless we're concerned about the API. Self-hosted users could leverage this today instead of the header-providing mechanisms they currently use.

bergundy:
> It's a handshake failure if server requires TLS and you don't use it, nothing leaked there.

Yeah, I think you're right, the header is only sent post TLS handshake IIUC. Not concerned about that.

client/client.go
Sushisource: +1 to calling this static. Callback one I think can stay as is but I see Roey's name as good too. Either way.

internal/client.go
```diff
@@ -421,6 +422,9 @@ type (
 		// default: default
 		Namespace string
 
+		// Optional: Set the credentials for this client.
+		Credentials Credentials
```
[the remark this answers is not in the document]
cretz: Linking to the type would be `[Credentials]`, but the type is already linked where this doc is

client/client.go
cretz: Updated names

Quinn-With-Two-Ns (approved):

cretz:
I'm not going to merge this quite yet. I want to have a team discussion on how this looks in other SDKs a bit first.

cretz:
Ok, after discussion, we are ok with this pattern

Issue #1411, opened 2024-03-06
Replace use of *updatepb.WaitPolicy with enum defined in the Go SDK


Quinn-With-Two-Ns:
Using `updatepb.WaitPolicy` in `UpdateWorkflowWithOptionsRequest` is not ideal since it require declaring a whole struct and a protobuf enum just to set the wait policy. We can replace this with an "enum" defined in the Go SDK.

https://github.com/temporalio/sdk-go/blob/efabf4622182554045df2bb9ab9f78d159ffd928/internal/internal_workflow_client.go#L785

PR #1412, opened 2024-03-06
Add UpdateLifeCycleStage enum

Quinn-With-Two-Ns:
Add UpdateLifeCycleStage enum instead of using the protobuf enum

closes https://github.com/temporalio/sdk-go/issues/1411


client/client.go
```diff
@@ -63,6 +63,25 @@ const (
...
+// NOTE: Experimental
+type UpdateLifeCycleStage = internal.UpdateLifeCycleStage
+
+const (
+	// UpdateLifeCycleStageUnspecified indicates the wait stage was not specified
+	// NOTE: Experimental
+	UpdateLifeCycleStageUnspecified = internal.UpdateLifeCycleStageUnspecified
+	// UpdateLifeCycleStageAdmitted indicates the update is admitted
+	// NOTE: Experimental
+	UpdateLifeCycleStageAdmitted = internal.UpdateLifeCycleStageAdmitted
```
cretz, line 76: Wonder if we should leave "admitted" unexposed for now

internal/interceptor.go
```diff
@@ -351,7 +351,7 @@ type ClientUpdateWorkflowInput struct {
 	Args                []interface{}
 	RunID               string
 	FirstExecutionRunID string
-	WaitPolicy          *updatepb.WaitPolicy
+	LifeCycleWaitStage  UpdateLifeCycleStage
```
cretz, line 354: Not sure we need to add `LifeCycle` in the field name. Just `WaitStage` or maybe `WaitForStage` makes sense to me

client/client.go
```diff
@@ -63,6 +63,25 @@ const (
 	TaskReachabilityClosedWorkflows = internal.TaskReachabilityClosedWorkflows
 )
 
+// UpdateLifeCycleStage indicates the stage of an update request.
+// NOTE: Experimental
+type UpdateLifeCycleStage = internal.UpdateLifeCycleStage
```
cretz, line 68: Can we call this `WorkflowUpdateLifeCycleStage` just to disambiguate any future uses of the general term "update"?

client/client.go
Quinn-With-Two-Ns: Let me check if the server will ever return admitted from a nonblocking poll

Quinn-With-Two-Ns: Testing with a non blocking poll, the server does expose when the update is admitted

cretz (approved):

client/client.go
Quinn-With-Two-Ns: Yes we should, I'll do that as part of this PR

PR #1415, opened 2024-03-07
Expose DevServer methods that allow sending SIGSTOP/CONT/KILL.

chronos-tachyon:
## What was changed
* Add `Pause(<-chan struct{}) error` to send `SIGSTOP` and `SIGCONT` to a dev server
* Add `Kill() error` to send `SIGKILL` to a dev server

## Why?
This will enable writing [features repo](https://github.com/temporalio/features) tests that verify the client's retry behavior.  [A comment on features#326](https://github.com/temporalio/features/issues/326#issuecomment-1732071582) says that @rross reports that the Java SDK doesn't retry correctly on the first connection.  By starting a dev server and freezing it before opening the first connection, we can test that this works correctly in all SDKs.

### DIFF
--- a/testsuite/devserver.go
+++ b/testsuite/devserver.go
@@ -358,6 +358,36 @@ func retryFor(maxAttempts int, interval time.Duration, cond func() error) error 
 	return lastErr
 }
 
+// Pause temporarily pauses the running server until the provided channel is
+// closed in another goroutine.  This can be used to simulate what would happen
+// if the host running the server experienced a network partition or a major
+// load spike that prevented it from servicing client requests.
+//
+// This operation is not available on Windows.
+func (s *DevServer) Pause(ch <-chan struct{}) error {
+	if err := sendStop(s.cmd.Process); err != nil {
+		return err
+	}
+	<-ch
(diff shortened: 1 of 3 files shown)


testsuite/devserver.go
```diff
@@ -358,6 +358,36 @@ func retryFor(maxAttempts int, interval time.Duration, cond func() error) error
...
+	if err := sendStop(s.cmd.Process); err != nil {
+		return err
+	}
+	<-ch
+	return sendContinue(s.cmd.Process)
+}
+
+// Kill immediately terminates the running server.  This can be used to
+// simulate what would happen if the host running the server experienced a
+// spontaneous reboot or sudden loss of power.
```
josh-berry, line 377: Since the dev-server does not persist its storage, `KILL` is equivalent to setting the server on fire with no backups. You may want to just use Pause() here instead—`KILL` will close connections cleanly because the kernel will do it for you. (A sudden loss of power etc. will look more similar to Pause() than Kill().)

Alternatively, if you want to simulate a spontaneous (but orderly) shutdown, you could keep Kill(), and have the server persist its state somewhere.

PR #1416, opened 2024-03-07
Make client Dial context controlable by callers

lminaudier:
## What was changed

This PR exposes new `DialContext` and `NewClientFromExistingWithContext` APIs that take a context as input and propagates it down the stack until all synchronous server calls so timeout and cancellation can be controlled by the callers.

The PR also replaces `grpc.Dial` call by `grpc.DialContext` and propagate user defined context down to the grpc dial call properly. Doing so means when users use `grpc.WithBlock()` dial option, they will be in control of the dial timeout

## Why?

When clients use `Dial` or `NewClientFromExisting` APIs to start, they connect to the server and try to load capabilities synchronously. However, clients have no control over the timeouts and cancellation of such requests. They can't extend timeouts to accommodate slow networks or appropriately cancel those calls.

## Checklist

1. Closes 

2. How was this tested:

I expect no behavior change. Timeouts are kept the same unless you use the new APIs and define proper context deadlines.

3. Any docs updates needed?

Godoc updates might be enough?

--- 

I noticed that the gRPC client uses the `Dial` API

https://github.com/temporalio/sdk-go/blob/efabf4622182554045df2bb9ab9f78d159ffd928/internal/grpc_dialer.go#L144

and not the `DialContext` one. However, it seems that users have the possibility to set arbitrary `grpc.DialOption` via the `ConnectionOptions`

https://github.com/temporalio/sdk-go/blob/efabf4622182554045df2bb9ab9f78d159ffd928/internal/client.go#L517-L525

which includes the `grpc.WithBlock()` one. When that's set it's probably better to use `DialContext` with a context having a proper timeout value set. I am not sure the direction maintainers would like to go. Propagating a context and switch over to `DialContext` or stay with `Dial` and not promote usage of `grpc.WithBlock()` option like [recommended upstream](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md)

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -220,7 +220,7 @@ type (
 // subjected to change in the future.
 // NOTE: the context.Context should have a fairly large timeout, since workflow execution may take a while to be finished
 func (wc *WorkflowClient) ExecuteWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (WorkflowRun, error) {
-	if err := wc.ensureInitialized(); err != nil {
+	if err := wc.ensureInitialized(ctx); err != nil {
 		return nil, err
 	}
 
@@ -296,7 +296,7 @@ func (wc *WorkflowClient) GetWorkflow(ctx context.Context, workflowID string, ru
 
 // SignalWorkflow signals a workflow in execution.
 func (wc *WorkflowClient) SignalWorkflow(ctx context.Context, workflowID string, runID string, signalName string, arg interface{}) error {
-	if err := wc.ensureInitialized(); err != nil {
+	if err := wc.ensureInitialized(ctx); err != nil {
 		return err
 	}
 
@@ -316,7 +316,7 @@ func (wc *WorkflowClient) SignalWorkflow(ctx context.Context, workflowID string,
 func (wc *WorkflowClient) SignalWithStartWorkflow(ctx context.Context, workflowID string, signalName string, signalArg interface{},
 	options StartWorkflowOptions, workflowFunc interface{}, workflowArgs ...interface{},
 ) (WorkflowRun, error) {
-	if err := wc.ensureInitialized(); err != nil {
+	if err := wc.ensureInitialized(ctx); err != nil {
 		return nil, err
 	}
 
@@ -358,7 +358,7 @@ func (wc *WorkflowClient) SignalWithStartWorkflow(ctx context.Context, workflowI
 // workflowID is required, other parameters are optional.
 // If runID is omit, it will terminate currently running workflow (if there is one) based on the workflowID.
 func (wc *WorkflowClient) CancelWorkflow(ctx context.Context, workflowID string, runID string) error {
-	if err := wc.ensureInitialized(); err != nil {
+	if err := wc.ensureInitialized(ctx); err != nil {
 		return err
 	}
 
@@ -369,7 +369,7 @@ func (wc *WorkflowClient) CancelWorkflow(ctx context.Context, workflowID string,
 // workflowID is required, other parameters are optional.
 // If runID is omit, it will terminate currently running workflow (if there is one) based on the workflowID.
 func (wc *WorkflowClient) TerminateWorkflow(ctx context.Context, workflowID string, runID string, reason string, details ...interface{}) error {
-	if err := wc.ensureInitialized(); err != nil {
+	if err := wc.ensureInitialized(ctx); err != nil {
 		return err
 	}
 
@@ -439,7 +439,7 @@ func (wc *WorkflowClient) getWorkflowHistory(
 func (wc *WorkflowClient) getWorkflowExecutionHistory(ctx context.Context, rpcMetricsHandler metrics.Handler, isLongPoll bool,
 	request *workflowservice.GetWorkflowExecutionHistoryRequest, filterType enumspb.HistoryEventFilterType,
 ) (*workflowservice.GetWorkflowExecutionHistoryResponse, error) {
-	if err := wc.ensureInitialized(); err != nil {
+	if err := wc.ensureInitialized(ctx); err != nil {
 		return nil, err
 	}
 
@@ -471,7 +471,7 @@ func (wc *WorkflowClient) getWorkflowExecutionHistory(ctx context.Context, rpcMe
 // completed event will be reported; if err is CanceledError, activity task canceled event will be reported; otherwise,
 // activity task failed event will be reported.
 func (wc *WorkflowClient) CompleteActivity(ctx context.Context, taskToken []byte, result interface{}, err error) error {
-	if err := wc.ensureInitialized(); err != nil {
+	if err := wc.ensureInitialized(ctx); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


internal/internal_workflow_client.go
```diff
@@ -1162,8 +1162,15 @@ func (wc *WorkflowClient) loadCapabilities() (*workflowservice.GetSystemInfoResp
 	}
 
 	// Fetch the capabilities
-	ctx, cancel := context.WithTimeout(context.Background(), getSystemInfoTimeout)
-	defer cancel()
+
+	// We set a default timeout if none is set in the context to stay
+	// backward compatible with the old behavior where the timeout was
+	// hardcoded to 5s.
```
cretz, line 1168: Hrmm, I am a bit concerned to remove 5s timeout just because context has a deadline. This call should really complete quick. Just because I have an outer context governing lots of things I may be doing and have set a timeout for 10m lets say, doesn't mean that I should have to wait 10m for this specific call to fail. I think we should add the timeout on top of the given context no matter what (i.e. a `min` of user deadline or 5s, still giving user cancelability).

IIRC the use case bringing about the PR was for human interaction in a gRPC interceptor, but I am not sure it's a fair default that we should allow this call to start taking as long as the outer context's deadline. Is there any way you can solve your use case without removing this timeout for everyone that just happens to have a deadline on their context?

cretz (commented):
@Quinn-With-Two-Ns - besides the comment I made, I think this is a good thing to do. Thoughts?

internal/internal_workflow_client.go
Quinn-With-Two-Ns: Yeah this is not a backwards compatible change

Quinn-With-Two-Ns: Yeah this is not a backwards compatible change in its current form

lminaudier: > IIRC the use case bringing about the PR was for human interaction in a gRPC interceptor, but I am not sure it's a fair default that we should allow this call to start taking as long as the outer context's deadline. Is there any way you can solve your use case without removing this timeout for everyone that just happens to have a deadline on their context?

Not anymore. Human interaction in interceptor is worked around by performing the interaction outside of the scope of the gRPC request. What I am trying to address here is slow/spotty networks. 

Some people operate very far from Temporal servers (basically the other side of the globe), some are working with ISPs with slow DNS resolution times, working from bad wifi connections from airports, hotels, ... working over slow VPNs, working in transit from trains over a spotty network, relying on slow 3G network, ... usually those are people on call trying to respond to incidents and willing to run some automations in Temporal for instance. Hitting a 5s timeout when you combine a couple of those situations and a bit of network retries is possible.

Couple of years ago, there was wifi network congestion in some offices for weeks also impacting latencies and people used to extend their `tctl` `--context_timeout` flag, I believe to control the overall RPC timeout before [loading server capabilities became a thing](https://github.com/temporalio/sdk-go/pull/706). That kind of flexibility was nice but we kinda lost it.

Sadly, we don't instrument developers laptops, I don't have metrics giving us the P99 dial latency for all our clients, nor traces to pinpoint where time is spent. This is based on couple of user reports that despite retries still hit the timeout.

5s timeout work 99% of the time but maybe 30s is a more appropriate default timeout in our situation. As you don't control environments clients run into having something customizable makes sense to me. Otherwise, your only option when facing timeouts are to work around the problem by ssh'ing into a node closer to server, fidgeting with auth, ... or try to find a better network.

If we take that stance, wether we can the overall dial timeout or every single RPC timeout triggered by the client dial call is a tradeoff. Expose simple imperfect configuration option or complex low level ones that compounds each time you add a new RPC call in the middle of the dial call.

cretz: Sounds like you may need the ability to control the get-system-info timeout more directly instead of changing everyone's default of 5s to their outer context. Many use large contexts with large deadlines but definitely don't have an initial connection/bootstrapping timeout expectation of that.

In general if any call like this takes longer than 5s you're going to have a subpar Temporal experience. We have task timeouts and such on pollers and Temporal in general is not built with the expectation that very simply calls to the server can take many seconds.

Quinn-With-Two-Ns: Yeah it seems like what we need is the option for users to explicitly set the timeout of `get-system-info`.

lminaudier: > Maybe we can have a client option for this.

Sure thing, I can revert back to that!

> In general if any call like this takes longer than 5s you're going to have a subpar Temporal experience. We have task timeouts and such on pollers and Temporal in general is not built with the expectation that very simple calls to the server can take many seconds. You're going to hit task timeouts and delays everywhere with an over-5s latency.

For sure! The issue is purely for CLI tools triggering workflows asynchronously from engineers laptops, not for workers processing tasks. Workers are in the cloud doing their job in a timely manner.

internal/internal_workflow_client.go
```diff
@@ -1147,8 +1147,31 @@ func (wc *WorkflowClient) OperatorService() operatorservice.OperatorServiceClien
...
+	return func(config *loadCapabilitiesConfig) {
+		if timeout > 0 {
+			config.getSystemInfoTimeout = timeout
+		}
+	}
+}
+
 // Get capabilities, lazily fetching from server if not already obtained.
-func (wc *WorkflowClient) loadCapabilities() (*workflowservice.GetSystemInfoResponse_Capabilities, error) {
+func (wc *WorkflowClient) loadCapabilities(ctx context.Context, opts ...loadCapabilitiesOption) (*workflowservice.GetSystemInfoResponse_Capabilities, error) {
```
cretz, line 1169: I think you're overcomplicating a non-exported function with this new option concept. Just pass in the timeout as a param and change the global val name to `defaultGetSystemInfoTimeout` and use the default if the passed in one is 0.

cretz (approved):
This looks good to me. Will let @Quinn-With-Two-Ns weigh in.

lminaudier:
Not 100% sure why CI complains. Locally for the check job I get

```
𝝺 pwd
.../internal/cmd/build

𝝺 go run . check
go: downloading honnef.co/go/tools v0.4.6
go: downloading golang.org/x/tools v0.16.0
2024/03/11 15:27:48 Running /opt/homebrew/bin/go in /Users/loic.minaudier/go/src/github.com/lminaudier/temporal-go-sdk with args [vet ./...]
2024/03/11 15:27:49 Installing github.com/kisielk/errcheck
2024/03/11 15:27:49 Running /Users/loic.minaudier/go/bin/errcheck in /Users/loic.minaudier/go/src/github.com/lminaudier/temporal-go-sdk with args [./...]
2024/03/11 15:27:50 Installing honnef.co/go/tools/cmd/staticcheck
2024/03/11 15:27:51 Running /Users/loic.minaudier/go/bin/staticcheck in /Users/loic.minaudier/go/src/github.com/lminaudier/temporal-go-sdk with args [./...]
2024/03/11 15:27:57 Running /opt/homebrew/bin/go in /Users/loic.minaudier/go/src/github.com/lminaudier/temporal-go-sdk with args [run ./internal/cmd/tools/copyright/licensegen.go --verifyOnly]
internal/cmd/build on ⛵ sdm1 (atlas-dev) (dial-context)
```

cretz:
Can you merge to latest master and run the `check` again? Another `DialClient`-based test was added to that file.

lminaudier:
Very interesting CI setup (I mean merging latest main before running checks).

Just rebased and fixed the test file.

Quinn-With-Two-Ns (approved):

cretz:
Merged, thanks!

PR #1421, opened 2024-03-12
Switch github runner to macos-14 (m1)

Quinn-With-Two-Ns:
Switch github runner to macos-14 (m1). Currently `macos-latest` runs on `macos-12` on intel. Reading some docs at some point github will change `macos-latest` to point at `macos-14`, but we can update our CI sooner.

https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
   build-and-test:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-12, macos-14, windows-latest]
+        os: [ubuntu-latest, macos-intel, macos-arm, windows-latest]
         go-version: ["1.21", "1.22"]
         include:
           - os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
We could also keep `macos-12` as a runner if we want mac+intel coverage

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   build-and-test:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-14, windows-latest]
```
cretz, line 12: I like keeping both (and maybe adding ubuntu-arm). Go doesn't really need it because we can trust the runtime across CPU arch's, but I wonder if it's a good practice to have all CIs of Temporal-distributed software test against ubuntu x64/arm, max x64/arm, and windows x64. Your call.

cretz:
> We could also keep macos-12 as a runner if we want mac+intel coverage

Question: Will macos x64 continue to be supported by GH in the future, or must we fixate on an old version? Reading around, macos 14 supports intel processors, does GH just not offer it?

Quinn-With-Two-Ns:
>Will macos x64 continue to be supported by GH in the future, or must we fixate on an old version?Will macos x64 continue to be supported by GH in the future, or must we fixate on an old version?

Reading some  announcement post they have no plans to add new x64 macos runners

cretz (approved):

PR #1425, opened 2024-03-21
Add `HistogramBucketBoundaries` option to contrib/opentelemetry

boxofrad:
## What was changed
Adds an option to the OpenTelemetry metrics handler to customise the histogram bucket boundaries for timing metrics.

## Why?
The default boundaries (`[0, 5, 10, 25, ... 10000]`) work well for relatively slow operations, such as the end-to-end workflow latency, but are less suitable for operations that should only take milliseconds.

### DIFF
--- a/contrib/opentelemetry/handler.go
+++ b/contrib/opentelemetry/handler.go
@@ -39,9 +39,10 @@ var _ client.MetricsHandler = MetricsHandler{}
 // MetricsHandler is an implementation of client.MetricsHandler
 // for open telemetry.
 type MetricsHandler struct {
-	meter      metric.Meter
-	attributes attribute.Set
-	onError    func(error)
+	meter                     metric.Meter
+	attributes                attribute.Set
+	onError                   func(error)
+	histogramBucketBoundaries func(string) []float64
 }
 
 // MetricsHandlerOptions are options provided to NewMetricsHandler.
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns:
`HistogramBucketBoundaries` is just a hint 

https://pkg.go.dev/go.opentelemetry.io/otel/metric#WithExplicitBucketBoundaries

```
This option is considered "advisory", and may be ignored by API implementations.
```

The whole advice API is experimental as well

https://github.com/open-telemetry/opentelemetry-specification/blob/4ac43da9d746fea42523187685f341321a1d6ee1/specification/metrics/api.md#instrument-advice

The proper way to configure the bucket size is with a `View`

boxofrad:
Hey @Quinn-With-Two-Ns 👋🏻 

Thanks so much for explaining this! As you can probably tell, I'm new to OpenTelemetry, so wasn't aware this was something you configure at the view level.

I'll close this PR, but for anyone who stumbles across it, here's my view definition:

```Go
var MetricViews = []metric.View{
	metric.NewView(
		metric.Instrument{
			Name:  "*_latency",
			Unit:  "s",
			Scope: instrumentation.Scope{Name: "temporal-sdk-go"},
		},
		metric.Stream{
			Aggregation: metric.AggregationExplicitBucketHistogram{
				Boundaries: []float64{
					0,

					// Custom lower boundaries (for very quick operations).
					0.01, 0.05, 0.1, 0.5, 1, 2,

					// Default boundaries.
					5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000,
				},
			},
		},
	),
}
```

Quinn-With-Two-Ns:
No worries @boxofrad when we should show how to configure the bucket size when we add metrics to our open telemetry samples

Issue #1172, opened 2023-07-24
Temporarily expose gRPC client connection for cloud API use


cretz:
**Describe the solution you'd like**

The in-development cloud API wants to reuse the gRPC connection from the Go client to instantiate the gRPC API. This API is not yet stable and may move to a different location at some point, so this is just a temporary need at this time.

I think a `GetGRPCConnection(internal.Client) *grpc.ClientConn` should be added to https://pkg.go.dev/go.temporal.io/sdk/internalbindings with a warning that it is experimental and may be removed in the future.

cretz:
This is not the approach we're going to take, see #1426 for a more full-fledged cloud client.

PR #1426, opened 2024-03-22
Experimental cloud operations client

cretz:
## What was changed

* Added `client.CloudOperationsClient`
  * Only one method for now, `CloudService()`, but has room for high-level wrappers if we ever want them
* Added `client.DialCloudOperationsClient()` and `client.CloudOperationsClientOptions`
* Added integration test as just a smoke test to confirm cloud client works

(was blocked by https://github.com/temporalio/api/pull/378, but now updated)


go.mod
```diff
@@ -27,7 +27,9 @@ require (
...
 	golang.org/x/net v0.22.0 // indirect
 	golang.org/x/text v0.14.0 // indirect
-	google.golang.org/genproto/googleapis/api v0.0.0-20240304212257-790db918fca8 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20240304212257-790db918fca8 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20240318140521-94a12d6c2237 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20240318140521-94a12d6c2237 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace go.temporal.io/api => ../temporal-api-go
```
cretz, line 35: Obviously won't pass CI until https://github.com/temporalio/api/pull/378 done

Quinn-With-Two-Ns:
This PR superseded this issue https://github.com/temporalio/sdk-go/issues/1172 correct?

cretz:
Yes, closing that one now.

Quinn-With-Two-Ns (approved):
No issue on the code side, assuming we as an org want to do this.

go.mod
cretz: Fixed

internal/internal_task_handlers_test.go
```diff
@@ -2153,7 +2153,7 @@ func Test_NonDeterministicCheck(t *testing.T) {
 		delete(commandTypes, cmd)
 	}
 
-	require.Equal(t, 15, len(commandTypes), "If you see this error, you are adding new command type. "+
+	require.Equal(t, 17, len(commandTypes), "If you see this error, you are adding new command type. "+
```
cretz, line 2156: Note, we expect this will be built by @bergundy and team as they build out Nexus support, this is just needed to pass CI

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.9.0
-	go.temporal.io/api v1.32.0
+	go.temporal.io/api v1.32.1-0.20240423190756-d1ff87aac800
```
cretz, line 13: We are not planning on merging this to `master` until a tagged version of API is done (we are letting other in-flight API projects drive the need for tagging)

cretz: This has been updated

cretz:
Updated API version. @anekkanti - can you review?

Quinn-With-Two-Ns:
I think we also should make samples for each SDK showing how to use the cloud API

cretz:
CI can't do the cloud test from forks because it requires secrets. I am going to close this and push to a local branch so the cloud test will run.

PR #1429, opened 2024-03-29
Support Update and Get for the new versioning rules

antlai-temporal:
## What was changed
This is the first installment of supporting the new versioning rules in the go sdk. It provides
support for updating and listing both assignment and redirect rules.

It does not include reachability info yet.

## Checklist

1. Closes 
Nothing to close yet

2. How was this tested:

Added unit and system tests. To pass system tests use a versioning-2 server branch as target.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -259,7 +259,7 @@ type (
 	// WARNING: Worker versioning is currently experimental
 	TaskQueueReachability = internal.TaskQueueReachability
 
-	// UpdateWorkerVersioningRulesOptions is the input to Client.UpdateWorkerVersioningRules.
+	// UpdateWorkerVersioningRulesOptions is the input to [Client.UpdateWorkerVersioningRules].
 	// WARNING: Worker versioning-2 is currently experimental
 	UpdateWorkerVersioningRulesOptions = internal.UpdateWorkerVersioningRulesOptions
 
@@ -271,7 +271,7 @@ type (
 	VersioningConflictToken = internal.VersioningConflictToken
 
 	// VersioningRampByPercentage is a VersionRamp that sends a proportion of the traffic
-	// to the target Build ID
+	// to the target Build ID.
 	// WARNING: Worker versioning-2 is currently experimental
 	VersioningRampByPercentage = internal.VersioningRampByPercentage
 
@@ -280,10 +280,20 @@ type (
 	// WARNING: Worker versioning-2 is currently experimental
 	VersioningAssignmentRule = internal.VersioningAssignmentRule
 
+	// VersioningAssignmentRuleWithTimestamp contains an assignment rule annotated
+	// by the server with its creation time.
+	// WARNING: Worker versioning-2 is currently experimental
+	VersioningAssignmentRuleWithTimestamp = internal.VersioningAssignmentRuleWithTimestamp
+
 	// VersioningAssignmentRule is a BuildID redirect rule for a task queue.
 	// It changes the behavior of currently running workflows and new ones.
 	// WARNING: Worker versioning-2 is currently experimental
 	VersioningRedirectRule = internal.VersioningRedirectRule
+
+	// VersioningRedirectRuleWithTimestamp contains a redirect rule annotated
+	// by the server with its creation time.
+	// WARNING: Worker versioning-2 is currently experimental
+	VersioningRedirectRuleWithTimestamp = internal.VersioningRedirectRuleWithTimestamp
 
 	// VersioningOpInsertAssignmentRule is an operation for UpdateWorkerVersioningRulesOptions
 	// that inserts the rule to the list of assignment rules for this Task Queue.
@@ -327,7 +337,7 @@ type (
 	// WARNING: Worker versioning-2 is currently experimental
 	VersioningOpDeleteRedirectRule = internal.VersioningOpDeleteRedirectRule
 
-	// VersioningOpCommitBuildId is an operation for UpdateWorkerVersioningRulesOptions
+	// VersioningOpCommitBuildID is an operation for UpdateWorkerVersioningRulesOptions
 	// that completes  the rollout of a BuildID and cleanup unnecessary rules possibly
 	// created during a gradual rollout. Specifically, this command will make the following changes
 	// atomically:
@@ -341,13 +351,13 @@ type (
 	// pollers have been seen recently for this Build ID. Use the `force`
 	// option to disable this validation.
 	// WARNING: Worker versioning-2 is currently experimental
-	VersioningOpCommitBuildId = internal.VersioningOpCommitBuildId
+	VersioningOpCommitBuildID = internal.VersioningOpCommitBuildID
 
-	// GetWorkerVersioningOptions is the input to Client.GetWorkerVersioningRules.
+	// GetWorkerVersioningOptions is the input to [Client.GetWorkerVersioningRules].
 	// WARNING: Worker versioning-2 is currently experimental
 	GetWorkerVersioningOptions = internal.GetWorkerVersioningOptions
 
-	// WorkerVersioningRules is the response for Client.GetWorkerVersioningRules.
+	// WorkerVersioningRules is the response for [Client.GetWorkerVersioningRules].
 	// WARNING: Worker versioning-2 is currently experimental
 	WorkerVersioningRules = internal.WorkerVersioningRules
 
--- a/client/client.go
+++ b/client/client.go
@@ -259,6 +259,98 @@ type (
 	// WARNING: Worker versioning is currently experimental
 	TaskQueueReachability = internal.TaskQueueReachability
 
+	// UpdateWorkerVersioningRulesOptions is the input to Client.UpdateWorkerVersioningRules.
+	// WARNING: Worker versioning-2 is currently experimental
+	UpdateWorkerVersioningRulesOptions = internal.UpdateWorkerVersioningRulesOptions
+
+	// VersioningConflictToken is a conflict token to serialize calls to Client.UpdateWorkerVersioningRules.
+	// An update with an old token fails with `serviceerror.FailedPrecondition`.
+	// The current token can be obtained with [GetWorkerVersioningRules],
+	// or returned by a successful [UpdateWorkerVersioningRules].
+	// WARNING: Worker versioning-2 is currently experimental
+	VersioningConflictToken = internal.VersioningConflictToken
+
+	// VersioningRampByPercentage is a VersionRamp that sends a proportion of the traffic
+	// to the target Build ID
+	// WARNING: Worker versioning-2 is currently experimental
+	VersioningRampByPercentage = internal.VersioningRampByPercentage
+
+	// VersioningAssignmentRule is a BuildID  assigment rule for a task queue.
+	// Assignment rules only affect new workflows.
+	// WARNING: Worker versioning-2 is currently experimental
+	VersioningAssignmentRule = internal.VersioningAssignmentRule
+
+	// VersioningAssignmentRule is a BuildID redirect rule for a task queue.
+	// It changes the behavior of currently running workflows and new ones.
+	// WARNING: Worker versioning-2 is currently experimental
+	VersioningRedirectRule = internal.VersioningRedirectRule
+
+	// VersioningOpInsertAssignmentRule is an operation for UpdateWorkerVersioningRulesOptions
+	// that inserts the rule to the list of assignment rules for this Task Queue.
+	// The rules are evaluated in order, starting from index 0. The first
+	// applicable rule will be applied and the rest will be ignored.
+	// By default, the new rule is inserted at the beginning of the list
+	// (index 0). If the given index is too larger the rule will be
+	// inserted at the end of the list.
+	// WARNING: Worker versioning-2 is currently experimental
+	VersioningOpInsertAssignmentRule = internal.VersioningOpInsertAssignmentRule
+
+	// VersioningOpReplaceAssignmentRule is an operation for UpdateWorkerVersioningRulesOptions
+	// that replaces the assignment rule at a given index. By default presence of one
+	// unconditional rule, i.e., no hint filter or ramp, is enforced, otherwise
+	// the delete operation will be rejected. Set `force` to true to
+	// bypass this validation.
+	// WARNING: Worker versioning-2 is currently experimental
+	VersioningOpReplaceAssignmentRule = internal.VersioningOpReplaceAssignmentRule
+
+	// VersioningOpDeleteAssignmentRule is an operation for UpdateWorkerVersioningRulesOptions
+	// that deletes the assignment rule at a given index. By default presence of one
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -550,18 +642,34 @@ type (
...
+		// Deprecated: Use [DescribeTaskQueue] with the versioning-2 api.
 		GetWorkerTaskReachability(ctx context.Context, options *GetWorkerTaskReachabilityOptions) (*WorkerTaskReachability, error)
 
+		// UpdateWorkerVersioningRules
+		// Allows updating the worker-build-id based assignment and redirect rules for a given task queue. This is used in
+		// conjunction with workers who specify their build id and thus opt into the feature.
+		// The errors it can return:
+		//  - serviceerror.FailedPrecondition when the conflict token is invalid
+		// WARNING: Worker versioning-2 is currently experimental, and requires server 1.XX+
+		UpdateWorkerVersioningRules(ctx context.Context, options *UpdateWorkerVersioningRulesOptions) (VersioningConflictToken, error)
```
cretz, line 666: It's very unfortunate that the update call doesn't return the same thing as the get call. I think we should consider changing the API so it does. But if we can't, this should change to returning `UpdateWorkerVersioningRulesResponse` that may contain the conflict token instead of assuming the conflict token is the only thing we ever will return here.

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+	VersioningRampByPercentage struct {
+		// Percentage of traffic with a value in [0,100)
+		Percentage float32
+	}
+
+	// VersioningAssignmentRule is a BuildID assigment rule for a task queue.
+	// Assignment rules only affect new workflows.
+	VersioningAssignmentRule struct {
+		// The BuildID of new workflows affected by this rule.
+		TargetBuildId string
```
cretz, line 50:
```suggestion
		TargetBuildID string
```

Capitalize both letters in ID in Go. This applies to several things in this review.

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+		// A strategy for gradual workflow deployment.
+		Ramp VersioningRamp
+		// An optional registration time set by the server.
+		Timestamp *time.Time
```
cretz, line 54: No need to use `nil` time, rather can state that "zero" time means unset.

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+		TargetBuildId string
+		// An optional registration time set by the server.
+		Timestamp *time.Time
+	}
+
+	//VersioningConflictToken is a conflict token to serialize updates.
+	// An update with an old token fails with `serviceerror.FailedPrecondition`.
+	// The current token can be obtained with [GetWorkerVersioningRules], or returned by a successful [UpdateWorkerVersioningRules].
+	VersioningConflictToken struct {
+		token []byte
```
cretz, line 70: Users cannot read this value and therefore cannot pass it around other systems if they need to. We should just give them a byte array directly, we don't need to wrap it in a struct or assume it needs to be hidden from them. This is similar to task tokens we use in other situations.

client/client.go
cretz, line 666: We did this callback-based for schedules and just looped while we got conflict errors. But that's not really required here, just wanted to point it out.

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+		// A conflict token to serialize updates.
+		ConflictToken VersioningConflictToken
+		Operation     UpdateVersioningOp
+	}
+
+	// UpdateVersioningOp is an interface for the different operations that can be
+	// performed when updating the worker versioning rules for a task queue.
+	//
+	// Possible operations are:
+	//   - VersioningOpInsertAssignmentRule
```
cretz, line 86:
```suggestion
	//   - [VersioningOpInsertAssignmentRule]
```

Putting brackets around these makes them link in Godoc

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+	//   - VersioningOpReplaceAssignmentRule
+	//   - VersioningOpDeleteAssignmentRule
+	//   - VersioningOpInsertRedirectRule
+	//   - VersioningOpReplaceRedirectRule
+	//   - VersioningOpDeleteRedirectRule
+	//   - VersioningOpCommitBuildId
+	UpdateVersioningOp interface {
```
cretz, line 93: Seems like this should just be `VersioningOp`, though consider calling it `VersioningOperation` (no reason to shorten the last word I don't think)

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+)
+
+func (uw *UpdateWorkerVersioningRulesOptions) validateAndConvertToProto(namespace string) (*workflowservice.UpdateWorkerVersioningRulesRequest, error) {
+	if namespace == "" {
+		return nil, errors.New("missing namespace argument")
+	}
+	if uw.TaskQueue == "" {
+		return nil, errors.New("missing TaskQueue field")
+	}
+	if !uw.Operation.isValidUpdateVersioningOp() {
```
cretz, line 131:
```suggestion
	if err := uw.Operation.validate(); err != nil {
```

This is often a better approach to letting individual implementations validate themselves and return a suitable error. Consider using this over booleans for user error clarity.

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+			},
+		}
+	case *VersioningOpCommitBuildId:
+		req.Operation = &workflowservice.UpdateWorkerVersioningRulesRequest_CommitBuildId_{
+			CommitBuildId: &workflowservice.UpdateWorkerVersioningRulesRequest_CommitBuildId{
+				TargetBuildId: v.TargetBuildId,
+				Force:         v.Force,
+			},
+		}
+	}
```
cretz, line 188: IMO it's best to add a `default:` clause that returns an error for unrecognized type

client/client.go
```diff
@@ -259,6 +259,98 @@ type (
 	// WARNING: Worker versioning is currently experimental
 	TaskQueueReachability = internal.TaskQueueReachability
 
+	// UpdateWorkerVersioningRulesOptions is the input to Client.UpdateWorkerVersioningRules.
+	// WARNING: Worker versioning-2 is currently experimental
+	UpdateWorkerVersioningRulesOptions = internal.UpdateWorkerVersioningRulesOptions
```
cretz, line 264: This is gonna sound annoying, but you'll want to copy this Godoc on each `internal` equivalent. Usually for things that just "wrap" internal you don't need to, but since this is an alias, people can arrive at the internal version and not here and deserve to get the same docs.

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+	ConflictToken   VersioningConflictToken
+}
+
+func versioningAssignmentRuleToProto(rule *VersioningAssignmentRule) *taskqueuepb.BuildIdAssignmentRule {
+	// Assumed `rule` already validated
+	result := &taskqueuepb.BuildIdAssignmentRule{
+		TargetBuildId: rule.TargetBuildId,
+	}
+
+	switch r := rule.Ramp.(type) {
```
cretz, line 226: Need a default case that fails here

internal/worker_versioning_rules.go
```diff
@@ -0,0 +1,338 @@
...
+	case *VersioningRampByPercentage:
+		result.Ramp = &taskqueuepb.BuildIdAssignmentRule_PercentageRamp{
+			PercentageRamp: &taskqueuepb.RampByPercentage{
+				RampPercentage: r.Percentage,
+			},
+		}
+	}
+	// Ignore `rule.Timestamp`
```
cretz, line 234: Can you clarify why timestamp is ignored? What's its purpose if unused? Should we just remove the field?

cretz (commented):
Left initial comments, but may need versioning to move a bit further along server side

client/client.go
antlai-temporal: Yes, returning the same as get call will save a few roundtrips, let's wait for the server change

internal/worker_versioning_rules.go
antlai-temporal: All the new protos are using Id not ID, should we change this everywhere?

internal/worker_versioning_rules.go
antlai-temporal: I wanted to avoid that they can change the byte array outside the package, and we could ended up adding other things, like a timestamp to expire the token

internal/worker_versioning_rules.go
antlai-temporal: Makes sense, thanks

internal/worker_versioning_rules.go
antlai-temporal: Thanks, good point.

internal/worker_versioning_rules.go
antlai-temporal: Yes, let's be consistent with the others.

internal/worker_versioning_rules.go
antlai-temporal: Yes, the error could add more context on why it didn't validate...

internal/worker_versioning_rules.go
cretz: No, protobuf is unrelated to Go and we don't want Go-specific rules to apply to proto, so leave the generated `Id` there. This is only about the Go part.

internal/worker_versioning_rules.go
cretz: I don't think we need to avoid that. No need to hide this (and it may have value for a user to store the byte array if they want). This same thing applies to "task token".

internal/worker_versioning_rules.go
antlai-temporal: Yes, I keep forgetting this is not Rust...

internal/worker_versioning_rules.go
antlai-temporal: The protos sometimes return a `TimestampedBuildIdAssignmentRule` or `*RedirectRule` that just add the timestamp to `BuildIdAssignmentRule`. Rather than expose both types I merged then making the Timestamp optional, but then I have to convert to the actual inner proto ignoring the timestamp. 
I should improve that comment...

internal/worker_versioning_rules.go
antlai-temporal: Makes sense thanks

client/client.go
antlai-temporal: I see, I was wondering why there was so much cut/paste of docs in the codebase, but this explains...

internal/worker_versioning_rules.go
cretz: We have found in the past, especially with more complicated model sets, that staying close to the API structure is safer for future changes. I think our objects in Go should very closely (exactly?) align with those in the API if we can.

Quinn-With-Two-Ns: +1 staying closer to the API structure makes future changes potentially much simpler

client/client.go
```diff
@@ -550,18 +642,34 @@ type (
 		// Allows you to update the worker-build-id based version sets for a particular task queue. This is used in
 		// conjunction with workers who specify their build id and thus opt into the feature.
 		// WARNING: Worker versioning is currently experimental
```
Quinn-With-Two-Ns, line 644: nit: I think we should remove the experimental warning since I don't think something should be both experimental and deprecated.

internal/worker_versioning_rules.go
antlai-temporal: I prefer to wrap it, maybe my paranoid years of working on security, it also hides the implementation of the conflict token, and makes it easier to extend it.

client/client.go
antlai-temporal: I agree that in general we should not deprecate experimental features and just remove them instead, but in this case we decided to keep an experimental feature in deprecated mode for one release...

internal/worker_versioning_rules.go
antlai-temporal: OK, I'll make visible types TimestampedBuildIdAssignmentRule and TimestampedCompatibleBuildIdRedirectRule in the SDK, but frankly, this looks like a complete overkill to me...

internal/worker_versioning_rules.go
cretz: Please let users have access to this token to do what they want with it

internal/worker_versioning_rules.go
cretz: > this looks like a complete overkill to me

That's ok, the Go implementations often do

client/client.go
Quinn-With-Two-Ns: I am not asking to remove the API, just remove the experimental tag on it.

client/client.go
```diff
@@ -550,18 +652,34 @@ type (
...
+		// Deprecated: Use [UpdateWorkerVersioningRules] with the versioning-2 api.
 		UpdateWorkerBuildIdCompatibility(ctx context.Context, options *UpdateWorkerBuildIdCompatibilityOptions) error
 
 		// GetWorkerBuildIdCompatibility
 		// Returns the worker-build-id based version sets for a particular task queue.
 		// WARNING: Worker versioning is currently experimental
+		// Deprecated: Use [GetWorkerVersioningRules] with the versioning-2 api.
```
Quinn-With-Two-Ns, line 661: We mention `versioning-2` a lot in this PR, is there a definition somewhere we can point users too?

client/client.go
antlai-temporal: I understand, I think that API is both experimental and deprecated, so having both tags seems more accurate. Is there a tooling issue of having both tags?

client/client.go
antlai-temporal: Good point. This is currently a placeholder, similar to 1.XX+ that you see in other places. When we had a new "formal" name and description of the new versioning API I will change that.

client/client.go
cretz: It can't be both experimental and deprecated IMO, but I don't have a strong opinion on leaving both. It's just confusing.

antlai-temporal:
Addressing the change in UpdateWorkerVersioningRulesResponse in a separate PR

cretz:
@antlai-temporal - Saw this was merged with no approvals. But no big deal since we'll be able to review again when the `versioning-2` to `master` PR is opened.

antlai-temporal:
> @antlai-temporal - Saw this was merged with no approvals. But no big deal since we'll be able to review again when the `versioning-2` to `master` PR is opened.

@cretz Yes, I'm assuming that this is a first level filter, I don't close any discussions that are not fully resolved, and the real review will happen in the final merge, but hopefully it will be easier after all the previous discussions. 
My main drive right now is to  keep catching up with the api changes, e.g., the change in Update response, and provide a working SDK for the server folks...

PR #1431, opened 2024-04-03
Change UpdateWorkerVersioningRules response to match GetWorkerVersioningRules

antlai-temporal:
## What was changed

Change the response of UpdateWorkerVersioningRules to match GetWorkerVersioningRules as suggested by @cretz 

## Why?

Eliminates unnecessary round-trips to the server, simpler API...


cretz (approved):
Approved (pending final review on final PR to master)

PR #1432, opened 2024-04-04
Rename insert_compatible_redirect_rule to add_compatible_redirect_rule in api

antlai-temporal:
## What was changed

The api is renaming insert_compatible_redirect_rule to add_compatible_redirect_rule to reflect that there can only be one rule
active per source. 

This trivial PR just renames types to be consistent with the new api, and uses the new go proto stubs.


antlai-temporal:
This is just a trivial PR to keep up with api changes, no need to review right now, there will be a final PR review for api and sdk before the merge in main...

antlai-temporal:
This was merged without comments because it is just a rename to follow the last api change.

PR #1433, opened 2024-04-05
Add `SetFailureConverter` to `WorkflowTestSuite`

boxofrad:
## What was changed
Adds a `SetFailureConverter` method to the `WorkflowTestSuite`.

## Why?
We use a custom failure converter to marshal/unmarshal structured errors in our application. These errors contain a "code" (similar to a gRPC status) which gets lost without the failure converter, so there are currently certain branches of our code that we're unable to test.

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -57,6 +57,7 @@ type (
 		logger                      log.Logger
 		metricsHandler              metrics.Handler
 		contextPropagators          []ContextPropagator
+		failureConverter            converter.FailureConverter
 		header                      *commonpb.Header
 		disableRegistrationAliasing bool
 	}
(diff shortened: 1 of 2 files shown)


internal/workflow_testsuite.go
```diff
@@ -159,6 +160,12 @@ func (s *WorkflowTestSuite) SetContextPropagators(ctxProps []ContextPropagator)
 	s.contextPropagators = ctxProps
 }
 
+// SetFailureConverter sets the failure converter for this WorkflowTestSuite. If you don't set the failure converter,
+// the default failure converter will be used.
+func (s *WorkflowTestSuite) SetFailureConverter(failureConverter converter.FailureConverter) {
```
cretz, line 165: I wonder if this should be on the activity/workflow environment instead (akin to data converter). @Quinn-With-Two-Ns - thoughts?

Quinn-With-Two-Ns: Yeah I think we should be consistent with how we set the data converter.

boxofrad: Makes sense! Added here: https://github.com/temporalio/sdk-go/pull/1484 (and I'll close this PR).

PR #1436, opened 2024-04-08
Update reapply

Quinn-With-Two-Ns:
Add support for `Update Admitted` events. This will allow the Go SDK to support reapplied updates from, for example, workflow reset. This is not currently supported on any released server versions so the tests are all through unit tests or histories generated with unreleased server versions.

### DIFF
--- a/converter/grpc_interceptor_test.go
+++ b/converter/grpc_interceptor_test.go
@@ -74,6 +74,7 @@ func TestPayloadCodecGRPCClientInterceptor(t *testing.T) {
 	)
 	require.NoError(err)
 
+	//lint:ignore SA1019 test
 	c, err := grpc.Dial(
 		server.addr,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
@@ -115,6 +116,7 @@ func TestFailureGRPCClientInterceptor(t *testing.T) {
 	)
 	require.NoError(err)
 
+	//lint:ignore SA1019 test
 	c, err := grpc.Dial(
 		server.addr,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
@@ -181,6 +183,7 @@ func (t *testGRPCServer) waitUntilServing() error {
 	// Try 20 times, waiting 100ms between
 	var lastErr error
 	for i := 0; i < 20; i++ {
+		//lint:ignore SA1019 test
 		conn, err := grpc.Dial(t.addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 		if err != nil {
 			lastErr = err
--- a/converter/grpc_interceptor_test.go
+++ b/converter/grpc_interceptor_test.go
@@ -74,7 +74,6 @@ func TestPayloadCodecGRPCClientInterceptor(t *testing.T) {
 	)
 	require.NoError(err)
 
-	//lint:ignore SA1019 test
 	c, err := grpc.Dial(
 		server.addr,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
@@ -116,7 +115,6 @@ func TestFailureGRPCClientInterceptor(t *testing.T) {
 	)
 	require.NoError(err)
 
-	//lint:ignore SA1019 test
 	c, err := grpc.Dial(
 		server.addr,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
@@ -183,7 +181,6 @@ func (t *testGRPCServer) waitUntilServing() error {
 	// Try 20 times, waiting 100ms between
 	var lastErr error
 	for i := 0; i < 20; i++ {
-		//lint:ignore SA1019 test
 		conn, err := grpc.Dial(t.addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 		if err != nil {
 			lastErr = err
--- a/internal/common/metrics/grpc_test.go
+++ b/internal/common/metrics/grpc_test.go
@@ -50,6 +50,7 @@ func TestGRPCInterceptor(t *testing.T) {
 
 	// Create client with interceptor
 	handler := metrics.NewCapturingHandler()
+	//lint:ignore SA1019 test
 	cc, err := grpc.Dial(l.Addr().String(),
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithUnaryInterceptor(metrics.NewGRPCInterceptor(handler, "_my_suffix")))
--- a/internal/common/metrics/grpc_test.go
+++ b/internal/common/metrics/grpc_test.go
@@ -50,7 +50,6 @@ func TestGRPCInterceptor(t *testing.T) {
 
 	// Create client with interceptor
 	handler := metrics.NewCapturingHandler()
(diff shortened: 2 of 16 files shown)


go.mod
```diff
@@ -10,24 +10,23 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.9.0
-	go.temporal.io/api v1.29.1
-	golang.org/x/sys v0.18.0
+	go.temporal.io/api v1.31.1-0.20240408135840-a6cecb28ed20
```
Quinn-With-Two-Ns, line 13: Note: currently the api changes for update admitted event are not in a tagged api release.

internal/common/metrics/grpc_test.go
```diff
@@ -50,6 +50,7 @@ func TestGRPCInterceptor(t *testing.T) {
 
 	// Create client with interceptor
 	handler := metrics.NewCapturingHandler()
+	//lint:ignore SA1019 test
```
Quinn-With-Two-Ns, line 53: I need to open a separate issue to address this

converter/grpc_interceptor_test.go
```diff
@@ -74,6 +74,7 @@ func TestPayloadCodecGRPCClientInterceptor(t *testing.T) {
 	)
 	require.NoError(err)
 
+	//lint:ignore SA1019 test
```
cretz, line 77: Is this to ignore deprecation? Which thing here is deprecated?

internal/internal_task_handlers.go
```diff
@@ -1005,19 +1022,33 @@ ProcessEvents:
...
+		historyMessages := nextTask.acceptedMsgs
 		flags := nextTask.flags
 		binaryChecksum := nextTask.binaryChecksum
 		nextTaskBuildId := nextTask.buildID
+		admittedUpdates := nextTask.admittedMsgs
 		// Check if we are replaying so we know if we should use the messages in the WFT or the history
 		isReplay := len(reorderedEvents) > 0 && reorderedHistory.IsReplayEvent(reorderedEvents[len(reorderedEvents)-1])
 		var msgs *eventMsgIndex
 		if isReplay {
+			requestUpdateBodies := make(map[string]*anypb.Any)
```
cretz, line 1034:
```suggestion
			requestUpdateBodies := make(map[string]*anypb.Any, len(admittedUpdates))
```

(pedantic, but a good habit)

internal/internal_task_handlers.go
```diff
@@ -1005,19 +1022,33 @@ ProcessEvents:
...
+			for _, updateRequest := range admittedUpdates {
+				requestUpdateBodies[updateRequest.GetProtocolInstanceId()] = updateRequest.GetBody()
+			}
+			// Check if we need to replace the body of the update request with the body of the update admitted event
+			for _, msg := range historyMessages {
+				if body, ok := requestUpdateBodies[msg.GetProtocolInstanceId()]; ok {
+					msg.Body = body
```
cretz, line 1041: To confirm, you want to replace this even if body is `nil` on the admitted request? Or I guess we can assume body is never nil in an update request?

cretz (approved):
> This is not currently supported on any released server versions so the tests are all through unit tests or histories generated with unreleased server versions.

I wonder if we should hold off merging until it's implemented in the server. Clearly nobody needs this until it even works. IME until something is implemented in the server, its API/protos very likely may change. If you do merge now, let's make an issue to add an integration test at some point.

internal/internal_task_handlers.go
Quinn-With-Two-Ns: Yeah I am assuming the latter, @dandavison the admitted body would never be `nil` correct?

converter/grpc_interceptor_test.go
cretz: Looks like they did in 1.63.0 and just changed their mind in 1.63.1, putting it off until 1.64.0: https://github.com/grpc/grpc-go/pull/7103.

Quinn-With-Two-Ns:
@dandavison Looking at your PR I think the only difference , other then the event name, is I am just replacing the body. I don't think there is any difference between the two approaches but for consistency It would be good to standardize on one. The latest proposal said to synthesize the update request message from the event so I am fine doing that.

converter/grpc_interceptor_test.go
Quinn-With-Two-Ns: Yeah but they still plan to deprecate it

Quinn-With-Two-Ns:
@dandavison Can you review this PR as well when you get a chance?

test/replaytests/replay_test.go
```diff
@@ -431,6 +431,20 @@ func (s *replayTestSuite) TestMultipleUpdates() {
...
+	replayer.RegisterWorkflow(CounterWorkflow)
+	err := replayer.ReplayWorkflowHistoryFromJSONFile(ilog.NewDefaultLogger(), "update-reset-accepted.json")
+	s.NoError(err)
+}
+
+func (s *replayTestSuite) TestResetWithUpdateRejected() {
+	replayer := worker.NewWorkflowReplayer()
+	replayer.RegisterWorkflow(CounterWorkflow)
+	err := replayer.ReplayWorkflowHistoryFromJSONFile(ilog.NewDefaultLogger(), "update-reset-rejected.json")
+	s.NoError(err)
```
dandavison, line 445: So these tests just check that replay does not error, but do not make any assertions about the state attained by replay, is that right? Should we make some assertions (e.g. assertions about log output? Is that the point of the logging in the workflow and validator?)? Otherwise what is the point of the `CounterWorkflow` body implementation? (Sorry, I'm new to these tests).

test/replaytests/update-reset-rejected.json
```diff
@@ -0,0 +1,226 @@
...
+          "input": {
+            "header": {},
+            "name": "fetch_and_add",
+            "args": {
+              "payloads": [
+                {
+                  "metadata": {
+                    "encoding": "anNvbi9wbGFpbg=="
+                  },
+                  "data": "MA=="
```
dandavison, line 92: This seems to be `0`, which is non-negative, so why is the update rejected?

dandavison (commented):
Great, that seems nice and clean. I have one question about the replay tests that I want to get to the bottom of. It's probably me mis-understanding the tests.

test/replaytests/replay_test.go
Quinn-With-Two-Ns: > but do not make any assertions about the state attained by replay, is that right?

They assert the replay succeeded which means the workflow took the same steps it took on initial execution which is dependent on the state of the workflow. 

We could remove the logs, they don't serve any purpose on replay.

test/replaytests/update-reset-rejected.json
Quinn-With-Two-Ns: because there is not update accepted event

dandavison: So, if we were making everything perfectly consistent, we'd make this value in history negative, right?

Issue #1435, opened 2024-04-06
It is not possible to unit test a child workflow that calls continue as new.


mfateev:
## Expected Behavior
A child workflow continues as new, and the parent waits for the child to complete.

## Actual Behavior
The parent receives ChildWorkflowExecutionError with ContinueAsNewError wrapped error.

## Steps to Reproduce the Problem

  1. Add the following test to [child-workflow-continue-as-new](https://github.com/temporalio/samples-go/tree/main/child-workflow-continue-as-new) sample.

```
package child_workflow_continue_as_new

import (
	"testing"

	"github.com/stretchr/testify/suite"
	"go.temporal.io/sdk/testsuite"
)

type UnitTestSuite struct {
	suite.Suite
	testsuite.WorkflowTestSuite
}

func TestUnitTestSuite(t *testing.T) {
	suite.Run(t, new(UnitTestSuite))
}

func (s *UnitTestSuite) Test_ExclusiveChoiceWorkflow() {
	env := s.NewTestWorkflowEnvironment()

	env.RegisterWorkflow(SampleChildWorkflow)
	env.ExecuteWorkflow(SampleParentWorkflow)

	s.True(env.IsWorkflowCompleted())
	s.NoError(env.GetWorkflowError())
}
```
Test fails with 
```
2024/04/05 17:52:02 ERROR Parent execution received child execution failure. Error child workflow execution error (type: SampleChildWorkflow, workflowID: child_workflow:default-test-run-id, runID: child_workflow:default-test-run-id_RunID, initiatedEventID: 0, startedEventID: 0): continue as new
    workflow_test.go:26: 
        	Error Trace:	/Users/maxim/temporal/samples-go/child-workflow-continue-as-new/workflow_test.go:26
        	Error:      	Received unexpected error:
        	            	workflow execution error (type: SampleParentWorkflow, workflowID: default-test-workflow-id, runID: default-test-run-id): child workflow execution error (type: SampleChildWorkflow, workflowID: child_workflow:default-test-run-id, runID: child_workflow:default-test-run-id_RunID, initiatedEventID: 0, startedEventID: 0): continue as new
        	Test:       	TestUnitTestSuite/Test_ExclusiveChoiceWorkflow
--- FAIL: TestUnitTestSuite (0.00s)
    --- FAIL: TestUnitTestSuite/Test_ExclusiveChoiceWorkflow (0.00s)

```

PR #1437, opened 2024-04-09
Add support for unit testing child workflows that call continue-as-new

mfateev:
## What was changed
Added support for unit testing child workflows that call continue-as-new. 

The change is safe as it wasn't possible to test such scenarios.

## Why?
The unit testing framework doesn't support tests that include child workflows that call continue-as-new.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/1435

2. How was this tested:
Added Test_ChildWorkflow_ContinueAsNew unit test.


cretz (approved):

cretz:
Will let @Quinn-With-Two-Ns review to see if he spots anything

PR #1438, opened 2024-04-09
datadog: Add tracer option to check error eligibility in the span

vaihtovirta:
## What was changed

Added an option to datadog tracer that allows to determine whether an error is eligible to be added to the DD span.

Similar functionality in gorm dd instrumentation: [WithErrorCheck](https://github.com/DataDog/dd-trace-go/blob/50ffac63fe850477042ce4c4bad8de9efe15142d/contrib/gorm.io/gorm.v1/option.go#L70-L77)

## Why?

Not all errors need to be reported to Datadog (intermediate retry error, validation error, etc).
Users can filter out such errors via specifying a custom function via the option.

## Checklist

1. How was this tested:
I've added simple unit tests.

2. Any docs updates needed?
I didn't find docs related to temporal datadog tracer in the repository, but potentially this page can mention the new option: https://docs.temporal.io/dev-guide/go/observability#useful-resources

### DIFF
--- a/contrib/datadog/tracing/interceptor.go
+++ b/contrib/datadog/tracing/interceptor.go
@@ -47,8 +47,10 @@ type TracerOptions struct {
 	// DisableQueryTracing can be set to disable query tracing.
 	DisableQueryTracing bool
 
-	// CheckError can be set to a custom function to determine if an error should be traced.
-	CheckError func(err error) bool
+	// OnFinish sets finish options.
+	// If unset, this will use [tracer.WithError]
+	// in case [interceptor.TracerFinishSpanOptions.Error] is non-nil and not [workflow.IsContinueAsNewError].
+	OnFinish func(options *interceptor.TracerFinishSpanOptions) []tracer.FinishOption
 }
 
 // NewTracingInterceptor convenience method that wraps a NeTracer() with a tracing interceptor
@@ -60,11 +62,23 @@ func NewTracingInterceptor(opts TracerOptions) interceptor.Interceptor {
 // NewTracer creates an interceptor for setting on client options
 // that implements Datadog tracing for workflows.
 func NewTracer(opts TracerOptions) interceptor.Tracer {
+	if opts.OnFinish == nil {
+		opts.OnFinish = func(options *interceptor.TracerFinishSpanOptions) []tracer.FinishOption {
+			var finishOpts []tracer.FinishOption
+
+			if err := options.Error; err != nil && !workflow.IsContinueAsNewError(err) {
+				finishOpts = append(finishOpts, tracer.WithError(err))
+			}
+
+			return finishOpts
+		}
+	}
+
 	return &tracerImpl{
 		opts: TracerOptions{
 			DisableSignalTracing: opts.DisableSignalTracing,
 			DisableQueryTracing:  opts.DisableQueryTracing,
-			CheckError:           opts.CheckError,
+			OnFinish:             opts.OnFinish,
 		},
 	}
 }
@@ -118,7 +132,7 @@ func (t *tracerImpl) SpanFromContext(ctx context.Context) interceptor.TracerSpan
 	if !ok {
 		return nil
 	}
-	return &tracerSpan{CheckError: t.opts.CheckError, Span: span}
+	return &tracerSpan{OnFinish: t.opts.OnFinish, Span: span}
 }
 
 func (t *tracerImpl) ContextWithSpan(ctx context.Context, span interceptor.TracerSpan) context.Context {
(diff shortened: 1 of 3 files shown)


contrib/datadog/tracing/interceptor.go
```diff
@@ -46,6 +46,9 @@ type TracerOptions struct {
 
 	// DisableQueryTracing can be set to disable query tracing.
 	DisableQueryTracing bool
+
+	// ErrCheckFn can be set to a custom function to determine if an error should be traced.
+	ErrCheckFn func(err error) bool
```
cretz, line 51: `Fn` not an abbreviation we usually use. How about:

```suggestion
	// CheckError can be set to a custom function to determine if an error should be traced.
	CheckError func(error) bool
```

Or maybe even better and more flexible:

```suggestion
	// OnFinish sets finish options. If unset, this will use [tracer.WithError] if 
	// [interceptor.TracerFinishSpanOptions.Error] is non-nil and not
	// [workflow.IsContinueAsNewError].
	OnFinish func(*interceptor.TracerFinishSpanOptions) []tracer.FinishOption
```

@Quinn-With-Two-Ns, preference?

contrib/datadog/tracing/interceptor.go
```diff
@@ -242,8 +247,21 @@ func (t *tracerSpan) ForeachBaggageItem(handler func(k string, v string) bool) {
...
+	isErrEligible := err != nil && !workflow.IsContinueAsNewError(err) && t.shouldErrBeTraced(err)
+
+	if isErrEligible {
 		opts = append(opts, tracer.WithError(err))
 	}
+
 	t.Span.Finish(opts...)
 }
+
+func (t *tracerSpan) shouldErrBeTraced(err error) bool {
```
cretz, line 261: Can just inline this, no need for a separate method (but no biggie)

cretz (commented):
Looks mostly good to me, will want @Quinn-With-Two-Ns's thoughts.

contrib/datadog/tracing/interceptor.go
vaihtovirta: Addressed

contrib/datadog/tracing/interceptor.go
vaihtovirta: Updated the naming, will be waiting for the final decision regarding the more flexible option.

contrib/datadog/tracing/interceptor.go
```diff
@@ -242,8 +247,15 @@ func (t *tracerSpan) ForeachBaggageItem(handler func(k string, v string) bool) {
 
 func (t *tracerSpan) Finish(options *interceptor.TracerFinishSpanOptions) {
 	var opts []tracer.FinishOption
-	if err := options.Error; err != nil && !workflow.IsContinueAsNewError(err) {
+
+	err := options.Error
+	isErrEligible := err != nil &&
+		!workflow.IsContinueAsNewError(err) &&
+		t.CheckError != nil && t.CheckError(err)
```
cretz, line 254:
```suggestion
		(t.CheckError == nil || t.CheckError(err))
```

Otherwise no error will ever be tracked unless they set the option. Alternatively you could just default check error to be a default impl that checks for continue-as-new error and tell users that so if they override, they can choose whether continue-as-new is counted.

contrib/datadog/tracing/interceptor.go
Quinn-With-Two-Ns: +1 to using `TracerFinishSpanOptions` to allow us to add more fields in the future.

vaihtovirta: Implemented 👌

contrib/datadog/tracing/interceptor.go
```diff
@@ -57,10 +62,23 @@ func NewTracingInterceptor(opts TracerOptions) interceptor.Interceptor {
...
+		opts.OnFinish = func(options *interceptor.TracerFinishSpanOptions) []tracer.FinishOption {
+			var finishOpts []tracer.FinishOption
+
+			if err := options.Error; err != nil && !workflow.IsContinueAsNewError(err) {
+				finishOpts = append(finishOpts, tracer.WithError(err))
+			}
+
+			return finishOpts
+		}
+	}
```
vaihtovirta, line 75: Instead of having if-else statement in the Finish function, I've decided to set a default function in the constructor.

cretz (approved):
LGTM, @Quinn-With-Two-Ns?

Quinn-With-Two-Ns (approved):
LGTM, I think you  just need to rebase and that would fix the `feature` tests

PR #1444, opened 2024-04-16
Ensure consistent labels for rpc metrics

dnr:
## What was changed
Ensure that all rpc metrics have a `namespace` label, by setting it to `_unknown_` if it can't be derived from the request.

This was mentioned here: https://github.com/temporalio/sdk-go/issues/861#issuecomment-1191524477 and the resolution was to derive the `namespace` label from the request, which is great, but it should be consistent.

## Why?
Many observability systems don't like the same metric to be emitted with different label sets.

## Checklist

1. Closes 

2. How was this tested:
updated unit test

3. Any docs updates needed?

### DIFF
--- a/internal/common/metrics/grpc.go
+++ b/internal/common/metrics/grpc.go
@@ -27,11 +27,11 @@ import (
 	"strings"
 	"time"
 
-	"google.golang.org/grpc/status"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // HandlerContextKey is the context key for a MetricHandler value.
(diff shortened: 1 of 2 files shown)


internal/common/metrics/grpc.go
```diff
@@ -63,16 +63,17 @@ func NewGRPCInterceptor(defaultHandler Handler, suffix string) grpc.UnaryClientI
 
 		// Only take method name after the last slash
 		operation := method[strings.LastIndex(method, "/")+1:]
-		tags := map[string]string{OperationTagName: operation}
 
 		// Since this interceptor can be used for clients of different name, we
 		// attempt to extract the namespace out of the request. All namespace-based
 		// requests have been confirmed to have a top-level namespace field.
+		namespace := "_unknown_"
```
cretz, line 70: I wonder if an empty string is better. Is there a common industry approach for label placeholder values?

Quinn-With-Two-Ns (approved):
We should probably ensure whatever we do here we do across all SDKs

internal/common/metrics/grpc.go
dnr: I don't know if there is a standard. I might have picked a different value myself, but this is what the server uses for the same situation so it seemed like a good idea to match

PR #1453, opened 2024-05-01
Implement DescribeTaskQueueEnhanced

antlai-temporal:
## What was changed

This PR implements `DescribeTaskQueueEnhanced`,  the new API that provides  buildID reachability information for workers, using the new rules-based versioning scheme.


internal/worker_versioning_info.go
```diff
@@ -0,0 +1,310 @@
...
+	TaskQueueTypeUnspecified = iota
+	// TaskQueueTypeWorkflow indicates the task queue has workflow type.
+	TaskQueueTypeWorkflow
+	// TaskQueueTypeActivity indicates the task queue has activity type.
+	TaskQueueTypeActivity
+	// TaskQueueTypeNexus indicates the task queue is used for dispatching Nexus requests.
+	TaskQueueTypeNexus
+)
+
+//	BuildIDTaskReachability specifies which category of tasks may reach a versioned worker of a certain Build ID.
```
cretz, line 50:
```suggestion
// BuildIDTaskReachability specifies which category of tasks may reach a versioned worker of a certain Build ID.
```

internal/worker_versioning_info.go
```diff
@@ -0,0 +1,310 @@
...
+	// it has been used by any existing execution within the retention period.
+	BuildIDTaskReachabilityUnreachable
+)
+
+type (
+	// TaskQueueVersionSelection is a task queue filter based on versioning.
+	// It is an optional component of [DescribeTaskQueueEnhancedOptions].
+	TaskQueueVersionSelection struct {
+		// Include specific Build IDs.
+		BuildIds []string
```
cretz, line 79:
```suggestion
		BuildIDs []string
```

internal/worker_versioning_info.go
```diff
@@ -0,0 +1,310 @@
...
+
+	enumspb "go.temporal.io/api/enums/v1"
+)
+
+// TaskQueueType specifies which category of tasks are associated with a queue.
+type TaskQueueType int
+
+const (
+	// TaskQueueTypeUnspecified indicates the task queue type was not specified.
+	TaskQueueTypeUnspecified = iota
```
cretz, line 41: What is the use of unspecified? I know protos have it, but do SDKs need it? Is there ever a case where users need it in either request or response direction?

internal/worker_versioning_info.go
```diff
@@ -0,0 +1,310 @@
...
+		ReportTaskReachability: o.ReportTaskReachability,
+	}
+
+	return opt, nil
+}
+
+func workerVersionCapabilitiesFromResponse(response *common.WorkerVersionCapabilities) *WorkerVersionCapabilities {
+	if response == nil {
+		return nil
+	}
```
cretz, line 183:
```suggestion
```

Not needed, those get calls below are nil-safe

internal/worker_versioning_info.go
```diff
@@ -0,0 +1,310 @@
...
+		Pollers: pollers,
+	}
+}
+
+func taskQueueVersionInfoFromResponse(response *taskqueuepb.TaskQueueVersionInfo) TaskQueueVersionInfo {
+	if response == nil {
+		return TaskQueueVersionInfo{}
+	}
+
+	typesInfo := make(map[TaskQueueType]TaskQueueTypeInfo)
```
cretz, line 229:
```suggestion
	typesInfo := make(map[TaskQueueType]TaskQueueTypeInfo, len(response.GetTypesInfo()))
```

Pedantic good practice on map creation (here and elsewhere), not required

client/client.go
```diff
@@ -641,8 +714,16 @@ type (
 		//  - serviceerror.Internal
 		//  - serviceerror.Unavailable
 		//  - serviceerror.NotFound
+		// Deprecated: Use [DescribeTaskQueueEnhanced]
 		DescribeTaskQueue(ctx context.Context, taskqueue string, taskqueueType enumspb.TaskQueueType) (*workflowservice.DescribeTaskQueueResponse, error)
 
+		// DescribeTaskQueueEnhanced  returns information about the target task queue, broken down by Build Id:
```
Quinn-With-Two-Ns, line 720: What is the behaviour if `DescribeTaskQueueEnhanced` is called on an old server version that doesn't support the enhanced flag?

client/client.go
```diff
@@ -64,6 +64,45 @@ const (
 	TaskReachabilityClosedWorkflows = internal.TaskReachabilityClosedWorkflows
 )
 
+// TaskQueueType specifies which category of tasks are associated with a queue.
+// WARNING: Worker versioning-2 is currently experimental
```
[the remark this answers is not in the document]
antlai-temporal: Yes, this is just a placeholder until we decide how to call the new versioning. Maybe something like "Worker versioning rules".

client/client.go
```diff
@@ -641,8 +714,16 @@ type (
...
 		//  - serviceerror.NotFound
+		// Deprecated: Use [DescribeTaskQueueEnhanced]
 		DescribeTaskQueue(ctx context.Context, taskqueue string, taskqueueType enumspb.TaskQueueType) (*workflowservice.DescribeTaskQueueResponse, error)
 
+		// DescribeTaskQueueEnhanced  returns information about the target task queue, broken down by Build Id:
+		//   - List of pollers
+		//   - Workflow Reachability status
+		//   - Backlog info for Workflow and/or Activity tasks
+		// WARNING: Worker versioning-2 is currently experimental, and requires server 1.XX+
+		DescribeTaskQueueEnhanced(ctx context.Context, options *DescribeTaskQueueEnhancedOptions) (*TaskQueueInfo, error)
```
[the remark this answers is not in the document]
antlai-temporal: Makes sense, replacing...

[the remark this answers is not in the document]
antlai-temporal: I agree, we should not use "enhanced" if there is no "vanilla" method...

internal/worker_versioning_info.go
antlai-temporal: Thanks

internal/worker_versioning_info.go
antlai-temporal: Done

internal/worker_versioning_info.go
antlai-temporal: I'm just following the protos here. Maybe for documentation make it more explicit in `DescribeTaskQueueEnhancedOptions` that we want all types as opposed to just omit `TaskQueueTypes`...

internal/worker_versioning_info.go
antlai-temporal: Yes, but I think we want to return a nil pointer as opposed to an empty `&WorkerVersionCapabilities{}`

client/client.go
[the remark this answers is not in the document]
antlai-temporal: Yes, we don't need to return a pointer, the struct just contains a map. I'll leave options, some methods use options other request, most versioning APIs so far used options...

client/client.go
antlai-temporal: Good question. It will return an empty `TaskQueueDescription{}`, I will add a comment...

internal/worker_versioning_info.go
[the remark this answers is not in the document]
antlai-temporal: I don't have clear what is best here. The "info" is not all versioning specific (e.g. pollers info), but most is indexed by buildID, sometimes just the "default one" if not set. At the same time "describe task queue" feels like a top level command not a versioning one. So I didn't bundle it with the other file `worker_versioning_rules.go` that is just versioning, but I kept the "versioning" in the file name...

client/client.go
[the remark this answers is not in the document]
cretz: That does make me wonder, maybe we should just call this `DescribeTaskQueueWithOptions` and one of the options is the enhanced options which, if set, means they want enhanced.

internal/worker_versioning_info.go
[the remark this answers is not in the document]
cretz: Will let @Quinn-With-Two-Ns weigh in, but `internal_X_client.go` for client things does make sense.

internal/worker_versioning_info.go
cretz: No need to follow protos exactly I don't think. But if there was a need in knowing an unset value for this, it'd be worth it, but I don't think there is. I don't mind this here, I just don't think it'll ever be used.

internal/worker_versioning_info.go
cretz: Ah, true! My mistake.

cretz (approved):
LGTM. Still think we might want a better file organization/name, but some of that can be discussed in the PR that moves versioning-2 to master.

internal/worker_versioning_info.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah I think grouping under `internal_X_client.go` makes sense

As for having a separated versioning client like schedules I am not convinced it actually provides a better experience for the user. With schedules I see the the fact that it uses a different client causing some friction with little actual benefit.

client/client.go
Quinn-With-Two-Ns: hm generally if you call an API that is not supported Go APIs and our APIs return an error.

Quinn-With-Two-Ns (approved):

client/client.go
antlai-temporal: Let me think about this a bit more. I think we could infer that the server was "old" from the answer and return an error. The return values are not guaranteed to be correct, e.g., the request could filter by "nexus queue type" and that defaults to "workflow queue type" in old servers, so I think returning an error or nothing is the best path.

Also, I think we should not deprecate the old DescribeTaskQueue just yet. There are few users for that, mostly for getting pollers info, but we may want to give them a way to get that info in old servers without deprecation for a while.

Jumping ahead, for the cli, we probably should have a "legacy" flag to DescribeTaskQueue that defaults to the old DescribeTaskQueue

@cretz @Quinn-With-Two-Ns What do you think?

cretz: If `DescribeTaskQueueResponse.versions_info` is empty/unset, is that enough to know that it's not supported server side? What situations would that value not be set when calling this?

> Jumping ahead, for the cli, we probably should have a "legacy" flag to DescribeTaskQueue that defaults to the old DescribeTaskQueue

Can discuss over there, but that sounds about right. We're going to break compatibility, but that may not be a concern at this stage. Worth discussion on that side.

antlai-temporal: "If DescribeTaskQueueResponse.versions_info is empty/unset, is that enough to know that it's not supported server side? What situations would that value not be set when calling this?"

I think we also need to check that the server is returning info in the legacy proto fields "pollers" and "task-queue_status", and the map in "versions_info" is empty

cretz: What situations would versions_info come back empty when we send a request for enhanced _except_ if it was an old server? Is it accepted that an enhanced info request would ever return this set as empty?

antlai-temporal: I think that if select out the "Unversioned" queue in the queue selector, and there are no BuildDs used, that may give us an empty map (TBC). But the old server will populate the legacy fields with the "default" unversioned queue, so we could figure that out...

antlai-temporal: Adding a `detectEnhancedNotSupported()` method, not deprecating vanilla `DescribeTaskQueue` just yet...

internal/worker_versioning_info.go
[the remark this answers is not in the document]
antlai-temporal: Renaming `worker_versioning_info*.go` to `internal_versioning_client*.go`

TBD if we want further consolidation with `worker_versioning_rules*.go`.  My first impression is these do quite different things (manipulate rules) than the others (get polling stats and reachability info), and may be more "worker" than "client" in that sense, so may be a bit confusing to bundle everything together...

internal/worker_versioning_info.go
```diff
@@ -237,6 +237,15 @@ func taskQueueVersionInfoFromResponse(response *taskqueuepb.TaskQueueVersionInfo
 	}
 }
 
+func detectEnhancedNotSupported(response *workflowservice.DescribeTaskQueueResponse) error {
+	// A server before 1.24 returns a non-enhanced proto, which only fills `pollers` and `taskQueueStatus` fields
```
cretz: Can probably just check task queue status. Pollers seems unnecessary and may be empty in both cases.

internal/worker_versioning_info.go
```diff
@@ -237,6 +237,15 @@ func taskQueueVersionInfoFromResponse(response *taskqueuepb.TaskQueueVersionInfo
 	}
 }
 
+func detectEnhancedNotSupported(response *workflowservice.DescribeTaskQueueResponse) error {
```
cretz: Not sure this needs to be a separate top-level function in a separate file, but at least qualify it with "taskQueue" or something in the name since it shares the function space with the rest of the package.

internal/worker_versioning_info.go
antlai-temporal: I don't know exactly what the old implementation did, I'm just being overly cautious here...

internal/worker_versioning_info.go
antlai-temporal: Makes sense, renaming...

internal/worker_versioning_info.go
cretz: I think you can ignore poller count, that can be empty in both cases

PR #1456, opened 2024-05-08
Add support for NextRetryDelay for local activities

Quinn-With-Two-Ns:
Add support for NextRetryDelay for local activities and the test suite

closes https://github.com/temporalio/sdk-go/issues/1454

### DIFF
--- a/internal/error.go
+++ b/internal/error.go
@@ -328,10 +328,11 @@ func NewApplicationError(msg string, errType string, nonRetryable bool, cause er
 
 func NewApplicationErrorWithOptions(msg string, errType string, options ApplicationErrorOptions) error {
 	applicationErr := &ApplicationError{
-		msg:          msg,
-		errType:      errType,
-		cause:        options.Cause,
-		nonRetryable: options.NonRetryable,
+		msg:            msg,
+		errType:        errType,
+		cause:          options.Cause,
+		nonRetryable:   options.NonRetryable,
+		nextRetryDelay: options.NextRetryDelay,
 	}
 	// When return error to user, use EncodedValues as details and data is ready to be decoded by calling Get
 	details := options.Details
@@ -582,6 +583,8 @@ func (e *ApplicationError) Unwrap() error {
 	return e.cause
 }
 
+// NextRetryDelay returns the delay to wait before retrying the activity.
+// a zero value means to use the activities retry policy.
 func (e *ApplicationError) NextRetryDelay() time.Duration { return e.nextRetryDelay }
 
 // Error from error interface
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1279,7 +1279,7 @@ func (w *workflowExecutionContextImpl) retryLocalActivity(lar *localActivityResu
 	return false
 }
 
-func getRetryBackoff(lar *localActivityResult, now time.Time, dataConverter converter.DataConverter) time.Duration {
+func getRetryBackoff(lar *localActivityResult, now time.Time, _ converter.DataConverter) time.Duration {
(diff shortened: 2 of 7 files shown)


internal/internal_task_handlers.go
```diff
@@ -1279,7 +1279,7 @@ func (w *workflowExecutionContextImpl) retryLocalActivity(lar *localActivityResu
 	return false
 }
 
-func getRetryBackoff(lar *localActivityResult, now time.Time, dataConverter converter.DataConverter) time.Duration {
+func getRetryBackoff(lar *localActivityResult, now time.Time, _ converter.DataConverter) time.Duration {
```
cretz, line 1282: Could just remove the param here and from the caller, but meh

internal/error.go
```diff
@@ -328,10 +328,11 @@ func NewApplicationError(msg string, errType string, nonRetryable bool, cause er
...
 	applicationErr := &ApplicationError{
-		msg:          msg,
-		errType:      errType,
-		cause:        options.Cause,
-		nonRetryable: options.NonRetryable,
+		msg:            msg,
+		errType:        errType,
+		cause:          options.Cause,
+		nonRetryable:   options.NonRetryable,
+		nextRetryDelay: options.NextRetryDelay,
```
cretz, line 335: Oops, looks like this was missed in #1319, heh

internal/internal_task_handlers.go
```diff
@@ -1291,22 +1291,31 @@ func getRetryBackoffWithNowTime(p *RetryPolicy, attempt int32, err error, now, e
...
+	if applicationErr, ok := err.(*ApplicationError); ok {
+		backoffInterval = applicationErr.nextRetryDelay
+	}
+	// Calculate next backoff interval if the error did not contain the next backoff interval.
 	// attempt starts from 1
-	backoffInterval := time.Duration(float64(p.InitialInterval) * math.Pow(p.BackoffCoefficient, float64(attempt-1)))
-	if backoffInterval <= 0 {
-		// math.Pow() could overflow
-		if p.MaximumInterval > 0 {
+	if backoffInterval == time.Duration(0) {
```
cretz, line 1303:
```suggestion
	if backoffInterval == 0 {
```

Pedantic (literals are untyped)

internal/internal_task_handlers.go
```diff
@@ -1291,22 +1291,31 @@ func getRetryBackoffWithNowTime(p *RetryPolicy, attempt int32, err error, now, e
...
+		backoffInterval = time.Duration(float64(p.InitialInterval) * math.Pow(p.BackoffCoefficient, float64(attempt-1)))
+		if backoffInterval <= 0 {
+			// math.Pow() could overflow
+			if p.MaximumInterval > 0 {
+				backoffInterval = p.MaximumInterval
+			} else {
+				return noRetryBackoff
+			}
+		}
```
cretz, line 1312:
```suggestion
		if backoffInterval <= 0 {
			// math.Pow() could overflow
			if p.MaximumInterval > 0 {
				backoffInterval = p.MaximumInterval
			}
		}
```

And add after all this code (so it applies w/ next retry):

```go
if backoffInterval <= 0 {
	return noRetryBackoff
}
```

Quinn-With-Two-Ns: hm, but do we want a negative `next retry` to disable retries?

cretz: Well, non-positive, but that seems right in my head. What would happen today if this negative value was returned?

Quinn-With-Two-Ns: hm good point,  the code will treat it as `noRetryBackoff` anyway if it is negative.

internal/error.go
Quinn-With-Two-Ns: Yeah 😬

cretz (approved):

Issue #1455, opened 2024-05-06
Update handler should require `workflow.Context` as first parameter


dandavison:
Consider the following code:

```go
	workflowFn := func(ctx workflow.Context) error {
		workflow.SetUpdateHandler(ctx, "my-update-handler", func(arg string) error {
			err := workflow.Sleep(ctx, time.Minute)
			return err
		})
        ...
	}
```

The relevant documentation is:

> SetUpdateHandler forwards to SetUpdateHandlerWithOptions with an zero-initialized UpdateHandlerOptions struct. See SetUpdateHandlerWithOptions for more details

> SetUpdateHandlerWithOptions binds an update handler function to the specified name such that update invocations specifying that name will invoke the handler. The handler function can take as input any number of parameters so long as they can be serialized/deserialized by the system. The handler can take a [workflow.Context] as its first parameter but this is not required. The update handler must return either a single error or a single serializable object along with a single error. The update handler function is invoked in the context of the workflow and thus is subject to the same restrictions as workflow code, namely, the update handler must be deterministic. As with other workflow code, update code is free to invoke and wait on the results of activities. Update handler code is free to mutate workflow state.

The code type-checks, but is incorrect, since the update handler is attempting to use the `workflow.Context` instance from the main workflow body coroutine.

## Expected Behavior
`SetUpdateHandler` should be documented as taking a handler function that must take a `workflow.Context` as its first parameter.

If the first argument is found not to be of the correct type at run time, the `SetUpdateHandler` call should fail with an error message explaining to the user that they should alter the signature of their handler.

## Actual Behavior
It fails at run time with an error in SDK coroutine scheduler internals:
```
Error trying to block on coroutine which is already blocked,
most likely a wrong Context is used to do blocking call (like Future.Get() or Channel.Receive()
```

PR #1457, opened 2024-05-08
Require update handler to have a context

Quinn-With-Two-Ns:
Require update handler to have a context as the first parameter.

Note this is a breaking change

closes: https://github.com/temporalio/sdk-go/issues/1455


cretz (approved):
Please make sure to make this very clear in release notes. This is going to cause some current users some issues for sure and they won't know it until their workflow breaks.

Issue #1460, opened 2024-05-09
QueryRejectCondition not used in QueryWorkflowWithOptions


tsurdilo:
query reject condition not being passed - https://github.com/temporalio/sdk-go/blob/386a6d3df3ae426a1354d42ce1c8b39c36cc798c/internal/internal_workflow_client.go#L899

PR #1461, opened 2024-05-09
Fix QueryRejectCondition parameter in QueryWorkflowWithOptions

Quinn-With-Two-Ns:
Actually pass through QueryRejectCondition on QueryWorkflowWithOptions

closes https://github.com/temporalio/sdk-go/issues/1460


zjrosen:
I'd potentially add one more test that handles a canceled state versus the pure failed state, otherwise this looks good to me!

cretz (approved):

PR #1462, opened 2024-05-10
Experimental cloud operations client

### LINKED ISSUE #1459: Cloud Operations API Client

See https://github.com/temporalio/features/issues/440 and https://github.com/temporalio/sdk-go/pull/1426. We need a separate cloud client using the cloud protos (they are separate today). The client should be marked experimental.

cretz:
## What was changed

* Added `client.CloudOperationsClient`
  * Only one method for now, `CloudService()`, but has room for high-level wrappers if we ever want them
* Added `client.DialCloudOperationsClient()` and `client.CloudOperationsClientOptions`
* Added integration test as just a smoke test to confirm cloud client works

Users must deal w/ a required version header and an API key that can only live for a limited time during the current beta cloud API. We decided we'll no longer block this issue on those experience concerns.

(this was started in #1426, but moved to local branch to run cloud tests)

## Checklist

1. Closes #1459


cretz:
Updated to latest, and this is ready to go, pending the `TEMPORAL_CLIENT_CLOUD_API_KEY` secret addition

test/cloud_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/stretchr/testify/suite"
+	"go.temporal.io/api/cloud/cloudservice/v1"
+	"go.temporal.io/sdk/client"
+)
+
+var cloudTestsFlag bool
+
+func init() { flag.BoolVar(&cloudTestsFlag, "cloud-tests", false, "Enable cloud tests") }
+
+func TestCloudSuite(t *testing.T) {
```
Quinn-With-Two-Ns, line 43: Nit: Can we rename this to something like `TestCloudOpsSuit` or similar to clarify these tests are just for the clouds Ops API.

cretz: I can change this. I was actually thinking that this could encompass all cloud-only tests and not just operations ones (granted it's only operations tests today that are cloud only).

cretz: Changed

Quinn-With-Two-Ns: What kind of tests would be cloud-only that aren't the cloud ops tests?

We have other tests that already run against cloud https://github.com/temporalio/sdk-go/actions/runs/9784335348/job/27014986110, but I guess they are not cloud only.

Quinn-With-Two-Ns (approved):

test/cloud_test.go
cretz: I am not sure, but we can refactor when we get there (maybe we'll never there)

PR #1464, opened 2024-05-10
Remove DisableStickyExecution

Quinn-With-Two-Ns:
Remove DisableStickyExecution, this flag has done nothing for 4 years and is time to remove since it causes confusion.

### DIFF
--- a/internal/worker.go
+++ b/internal/worker.go
@@ -102,21 +102,16 @@ type (
 		// default: false
 		EnableLoggingInReplay bool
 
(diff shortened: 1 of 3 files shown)


dandavison (approved):
LGTM! I was just confused by this for a while today. I saw the deprecation notice, but it didn't occur to me that it actually might do nothing.

cretz (approved):

mfateev (approved):

PR #1466, opened 2024-05-10
Nexus

bergundy:
## What was changed

**EDIT**: Merged #1473 and #1475 into this PR, it now included the entire Nexus implementation for the SDK.

Added the `temporalnexus` package and implemented the handler side for Nexus, including registering and dispatching Nexus Operations.

Tests only pass with server `main`, so this PR should not be merged until the server is released.
A future PR will complete the nexus work allowing invoking Nexus Operations from a workflow.

See the [proposal](https://github.com/temporalio/proposals/blob/6db78a083ec8fe397527a3f0d7146c77c8204e96/nexus/sdk-go.md) for more information.

Also now memoizing `worker.Start()` to return consistent errors to callers and avoid rerunning the function unnecessarily.

Merge Checklist:

- [ ] Release Server
- [ ] Depend on tagged Nexus SDK

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,15 @@
 module go.temporal.io/sdk
 
-go 1.20
+go 1.21
+
+toolchain go1.21.1
 
 require (
 	github.com/facebookgo/clock v0.0.0-20150410010913-600d898af40a
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/mock v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
+	github.com/nexus-rpc/sdk-go v0.0.8-0.20240617225139-cd9d6c50e99d
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.9.0
@@ -16,6 +19,8 @@ require (
 	google.golang.org/grpc v1.64.0
 	google.golang.org/protobuf v1.34.1
 )
+
+require github.com/gorilla/mux v1.8.0 // indirect
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
--- a/internal/common/metrics/constants.go
+++ b/internal/common/metrics/constants.go
@@ -43,16 +43,11 @@ const (
 	WorkflowTaskNoCompletionCounter     = TemporalMetricsPrefix + "workflow_task_no_completion"
 
 	ActivityPollNoTaskCounter             = TemporalMetricsPrefix + "activity_poll_no_task"
-	NexusPollNoTaskCounter                = TemporalMetricsPrefix + "nexus_poll_no_task"
 	ActivityScheduleToStartLatency        = TemporalMetricsPrefix + "activity_schedule_to_start_latency"
 	ActivityExecutionFailedCounter        = TemporalMetricsPrefix + "activity_execution_failed"
-	NexusTaskScheduleToStartLatency       = TemporalMetricsPrefix + "nexus_task_schedule_to_start_latency"
-	NexusTaskExecutionFailedCounter       = TemporalMetricsPrefix + "nexus_task_execution_failed"
 	UnregisteredActivityInvocationCounter = TemporalMetricsPrefix + "unregistered_activity_invocation"
 	ActivityExecutionLatency              = TemporalMetricsPrefix + "activity_execution_latency"
-	NexusTaskExecutionLatency             = TemporalMetricsPrefix + "nexus_task_execution_latency"
 	ActivitySucceedEndToEndLatency        = TemporalMetricsPrefix + "activity_succeed_endtoend_latency"
-	NexusTaskEndToEndLatency              = TemporalMetricsPrefix + "nexus_task_endtoend_latency"
 	ActivityTaskErrorCounter              = TemporalMetricsPrefix + "activity_task_error"
 
 	LocalActivityTotalCounter             = TemporalMetricsPrefix + "local_activity_total"
@@ -86,6 +81,12 @@ const (
 	StickyCacheSize                = TemporalMetricsPrefix + "sticky_cache_size"
 
 	WorkflowActiveThreadCount = TemporalMetricsPrefix + "workflow_active_thread_count"
+
+	NexusPollNoTaskCounter          = TemporalMetricsPrefix + "nexus_poll_no_task"
+	NexusTaskScheduleToStartLatency = TemporalMetricsPrefix + "nexus_task_schedule_to_start_latency"
+	NexusTaskExecutionFailedCounter = TemporalMetricsPrefix + "nexus_task_execution_failed"
+	NexusTaskExecutionLatency       = TemporalMetricsPrefix + "nexus_task_execution_latency"
+	NexusTaskEndToEndLatency        = TemporalMetricsPrefix + "nexus_task_endtoend_latency"
 )
 
 // Metric tag keys
--- a/internal/common/metrics/constants.go
+++ b/internal/common/metrics/constants.go
@@ -43,11 +43,16 @@ const (
 	WorkflowTaskNoCompletionCounter     = TemporalMetricsPrefix + "workflow_task_no_completion"
 
 	ActivityPollNoTaskCounter             = TemporalMetricsPrefix + "activity_poll_no_task"
+	NexusPollNoTaskCounter                = TemporalMetricsPrefix + "nexus_poll_no_task"
 	ActivityScheduleToStartLatency        = TemporalMetricsPrefix + "activity_schedule_to_start_latency"
 	ActivityExecutionFailedCounter        = TemporalMetricsPrefix + "activity_execution_failed"
+	NexusTaskScheduleToStartLatency       = TemporalMetricsPrefix + "nexus_task_schedule_to_start_latency"
+	NexusTaskExecutionFailedCounter       = TemporalMetricsPrefix + "nexus_task_execution_failed"
 	UnregisteredActivityInvocationCounter = TemporalMetricsPrefix + "unregistered_activity_invocation"
 	ActivityExecutionLatency              = TemporalMetricsPrefix + "activity_execution_latency"
+	NexusTaskExecutionLatency             = TemporalMetricsPrefix + "nexus_task_execution_latency"
 	ActivitySucceedEndToEndLatency        = TemporalMetricsPrefix + "activity_succeed_endtoend_latency"
+	NexusTaskEndToEndLatency              = TemporalMetricsPrefix + "nexus_task_endtoend_latency"
 	ActivityTaskErrorCounter              = TemporalMetricsPrefix + "activity_task_error"
 
 	LocalActivityTotalCounter             = TemporalMetricsPrefix + "local_activity_total"
@@ -91,6 +96,8 @@ const (
 	WorkerTypeTagName         = "worker_type"
 	WorkflowTypeNameTagName   = "workflow_type"
 	ActivityTypeNameTagName   = "activity_type"
+	NexusServiceTagName       = "nexus_service"
+	NexusOperationTagName     = "nexus_operation"
 	TaskQueueTagName          = "task_queue"
 	OperationTagName          = "operation"
 	CauseTagName              = "cause"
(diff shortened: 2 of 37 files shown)


go.mod
```diff
@@ -1,32 +1,37 @@
 module go.temporal.io/sdk
 
-go 1.20
+go 1.21
+
+toolchain go1.21.1
```
cretz, line 5: Curious, is this just an artifact of your tooling or was this change required?

go.mod
```diff
@@ -1,32 +1,37 @@
...
 
 require (
 	github.com/facebookgo/clock v0.0.0-20150410010913-600d898af40a
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/mock v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
+	github.com/nexus-rpc/sdk-go v0.0.8-0.20240502185337-2b47041a2cc2
```
cretz, line 12: I know we're not merging this PR until server tagged, but would also recommend this Nexus project is tagged before any sdk-go release with this dependency

internal/common/metrics/constants.go
```diff
@@ -43,11 +43,16 @@ const (
 	WorkflowTaskNoCompletionCounter     = TemporalMetricsPrefix + "workflow_task_no_completion"
 
 	ActivityPollNoTaskCounter             = TemporalMetricsPrefix + "activity_poll_no_task"
+	NexusPollNoTaskCounter                = TemporalMetricsPrefix + "nexus_poll_no_task"
```
cretz, line 46: I think putting all `Nexus` constants together is a bit easier to read than interleaved with their activity equivalents

internal/internal_nexus_task_handler.go
```diff
@@ -0,0 +1,353 @@
...
+				var ok bool
+				err, ok = recovered.(error)
+				if !ok {
+					err = fmt.Errorf("panic: %v", recovered)
+				}
+
+				nctx.Log.Error("Panic captured while handling nexus task", tagStackTrace, string(debug.Stack()), tagError, err)
+			}
+		}()
+		opres, err = h.nexusHandler.StartOperation(ctx, req.GetService(), req.GetOperation(), input, startOptions)
```
cretz, line 159: I assume (inbound) interceptors are coming later?

internal/internal_nexus_task_handler.go
```diff
@@ -0,0 +1,353 @@
...
+func (h *nexusTaskHandler) internalError(err error) *nexuspb.HandlerError {
+	h.logger.Error("error processing nexus task", "error", err)
+	return nexusHandlerError(nexus.HandlerErrorTypeInternal, "internal error")
+}
+
+func (h *nexusTaskHandler) goContextForTask(nctx *NexusOperationContext, header nexus.Header) (context.Context, context.CancelFunc, *nexuspb.HandlerError) {
+	// Associate the NexusOperationContext with the context.Context used to invoke operations.
+	ctx := context.WithValue(context.Background(), nexusOperationContextKey, nctx)
+
+	timeoutStr := header.Get(nexus.HeaderRequestTimeout)
```
cretz, line 269: Hrmm, I might have expected server to handle timeout and send cancellation. If a handler chooses not to respect timeout, what happens? If it is also handled server side, I think it's best to not also do it here except maybe with some considerable leeway to ensure server's cancellation logic is the one always processed.

internal/internal_nexus_worker.go
```diff
@@ -0,0 +1,86 @@
...
+			opts.client,
+			opts.executionParameters.DataConverter,
+			&workflowservice.RespondNexusTaskCompletedRequest{
+				Namespace: opts.executionParameters.Namespace,
+				Identity:  opts.executionParameters.Identity,
+			},
+			&workflowservice.RespondNexusTaskFailedRequest{
+				Namespace: opts.executionParameters.Namespace,
+				Identity:  opts.executionParameters.Identity,
+			},
```
cretz, line 40: Can probably just pass the two fields in to build these requests instead of pre-building a request and cloning just for these two fields

internal/internal_nexus_task_handler.go
```diff
@@ -0,0 +1,353 @@
...
+
+func (p *payloadSerializer) Deserialize(_ *nexus.Content, v any) error {
+	return p.converter.FromPayload(p.payload, v)
+}
+
+func (p *payloadSerializer) Serialize(v any) (*nexus.Content, error) {
+	panic("unimplemented") // not used - operation outputs are directly serialized to payload.
+}
+
+var nopReadCloser = io.NopCloser(bytes.NewReader([]byte{}))
```
cretz, line 353: This is not a `nopReadCloser`, this `emptyReaderNopCloser`. This is going to return an EOF immediately (which is maybe what you want, though can change empty byte slice to `nil`)

internal/internal_nexus_task_poller.go
```diff
@@ -0,0 +1,167 @@
...
+	res, failure, err := ntp.taskHandler.Execute(response)
+
+	// Execution latency (in-SDK processing time).
+	metricsHandler.Timer(metrics.NexusTaskExecutionLatency).Record(time.Since(executionStartTime))
+	if err != nil || failure != nil {
+		metricsHandler.Counter(metrics.NexusTaskExecutionFailedCounter).Inc(1)
+	}
+
+	// Drop the task, nothing to report back.
+	// This is only expected due to context deadline errors.
```
cretz, line 130: If this is the case, confirm it. I am a bit concerned other errors will get swallowed.

internal/internal_nexus_worker.go
```diff
@@ -0,0 +1,86 @@
...
+		executionParameters: opts.executionParameters,
+		workflowService:     opts.workflowService,
+		worker:              baseWorker,
+		stopC:               workerStopChannel,
+	}, nil
+}
+
+// Start the worker.
+func (w *nexusWorker) Start() error {
+	err := verifyNamespaceExist(w.workflowService, w.executionParameters.MetricsHandler, w.executionParameters.Namespace, w.worker.logger)
```
cretz, line 74: Unrelated to this PR specifically, but sure wouldn't mind if this moved up to the aggregate worker instead of in each

internal/internal_worker.go
```diff
@@ -953,6 +994,14 @@ func (aw *AggregatedWorker) Start() error {
 	}
 	proto.Merge(aw.capabilities, capabilities)
 
+	return aw.memoizedStart()
```
cretz, line 997: I think this whole method should go inside memoized start. No need to repeat stuff above for each call.

internal/nexus_operations.go
```diff
@@ -0,0 +1,23 @@
...
+	Client    Client
+	TaskQueue string
+	Log       log.Logger
+}
+
+// nexusOperationContextKey is a key for associating a [NexusOperationContext] with a [context.Context].
+var nexusOperationContextKey = struct{}{}
+
+// NexusOperationContextFromGoContext gets the [NexusOperationContext] associated with the given [context.Context].
+func NexusOperationContextFromGoContext(ctx context.Context) (nctx *NexusOperationContext, ok bool) {
```
cretz, line 20: Would call it `GetNexusOperationContext` similar to how the `activity` package has `GetX` calls that accept the context. Also, I would suggest not returning a boolean and just letting the result be nil if not in an operation context. Also, I would suggest making a new struct with fields from what's in context each time to discourage mutation.

Alternatively, consider `GetNexusOperationTemporalClient`, `GetNexusOperationInfo` (with just `TaskQueue` in it for now), and `GetNexusOperationLogger` to more consistently match what `activity` package does.

temporalnexus/operation.go
cretz, line 1: I assume calling-from-workflow stuff is coming later? Will it be in the `workflow` package? Will that code have to reference this package? I suggest considering adding it in this PR even in stub form to get circular imports correct.

cretz (commented):
Looks great. Mostly minimal stuff.

go.mod
bergundy: I just ran `go get ...` and `go mod tidy` but it may have been required due to the nexus sdk using `slog`.

go.mod
bergundy: Sure, I can add this to a checklist.

bergundy: Done.

internal/internal_nexus_task_handler.go
bergundy: Yeah, I didn't include those for now.
I had them in a local branch but decided to start slim.

I'll make sure its tracked.

internal/internal_nexus_task_handler.go
bergundy: We don't have sticky execution so there's not a way for the server to send cancelation. The server propagates this from the client request and also has its own context deadline.

I think it's good to cancel work that we know can't complete in time and have the server propagate this timeout.
As for whether the context deadline in the SDK should be shorter/same/longer than the one tracked on the server, fair point, but maybe shorter here is better so the SDK doesn't get a false sense of completion and the metrics we emit can be more accurate.

internal/internal_nexus_task_handler.go
bergundy: Sure, I can rename. It can't be nil because the Nexus SDK tries to consume it.

internal/internal_nexus_task_poller.go
bergundy: I'll just return the error just in case, thanks!

internal/internal_worker.go
bergundy: Works for me, this was something that @Quinn-With-Two-Ns requested, so just confirming he's also okay with that.

internal/nexus_operations.go
bergundy: This is all internal so I wouldn't worry about it too much at the moment, we can expose some of this later as we see fit.

temporalnexus/operation.go
bergundy: Yeah, it'll be in the `workflow` package.
I started working on it and so far there aren't any circular deps.

internal/internal_nexus_worker.go
bergundy: Yeah, I'll do that.

internal/common/metrics/constants.go
bergundy: Should I group the activity and workflow ones too?

cretz: Nah, I'd just add your stuff separate

internal/internal_nexus_task_handler.go
cretz: But what happens in racy cases where SDK and server hit at the same time? If server-side happens first does that look the exact same as if the client-side one hit first and reported this failure back? It's important to have one system be the arbiter of true timeout errors. If you want a just-in-case for the other system, no prob, but having two separate systems that race each other to report timeout failure can result in racy inconsistencies.

internal/internal_nexus_task_handler.go
cretz: I mean you can use `nil` instead of `[]byte{}`

internal/nexus_operations.go
cretz: Oh, my mistake. I think we will want to provide it to users one day, but if not exposed today, no prob

internal/internal_worker.go
Quinn-With-Two-Ns: Users may retry starting their worker if it failed that is why I requested we don't memorize it, I think the likely hood is low, but it's very little effort to not memorize it so why not just avoid the breaking change?

bergundy: I changed it to what @cretz suggested, I'm open to changing back.
I don't have a strong opinion here but slightly prefer memoizing the entire thing because it's easier to reason about.

internal/internal_nexus_task_handler.go
bergundy: Keep in mind that clients also set deadlines on the request context.
I think giving the most up-to-date and accurate timeout to all of the processes involved in handling this request is preferable. That's how gRPC does it and this is essentially the SDK handling RPCs.

Also note that on context deadline errors we don't respond to the server, we just drop the task so I'm not as concerned with the racy inconsistency you're talking about.

Quinn-With-Two-Ns (approved):

cretz:
Probably obvious, but let's not merge this until there's a server that works with it

go.mod
Quinn-With-Two-Ns: Just FYI we should remove this before merging to main, it seems to mess up out CI trying to test multiple versions of Go

bergundy:
Rebased and merged into the `nexus` branch.
I'll issue a separate PR to merge `nexus` into `main` once a server supporting Nexus is released.

Issue #1458, opened 2024-05-09
Set Temporal-Namespace header on every namespace-specific gRPC request


cretz:
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/475. This can be done with an interceptor that type asserts to `interface { GetNamespace() string }` (we've confirmed this works and will continue to work because that field will be present).

cretz:
Note, once this is released, any samples/docs that reference explicit namespace setting we can probably alter

GSmithApps:
Hey @cretz and @Quinn-With-Two-Ns 👋  This change removes the need to have the `DialOptions` shown on line 407 of [this commit](https://github.com/temporalio/documentation/commit/b87eab2c9914e62ed8868fb3e111e4e223c904aa#diff-a5e4c731af1568fb94988abc664627a9d778fe28bcfab17c0e6b482705234102L411), correct? Meaning in the interval [1.26.0, 1.33.0), it needed to be there, and in [1.33.0,  TBD), it doesn’t need to be there, correct?

Quinn-With-Two-Ns:
Yep

PR #1467, opened 2024-05-11
Set Temporal-Namespace header on every namespace-specific gRPC request

Quinn-With-Two-Ns:
Set Temporal-Namespace header on every namespace-specific gRPC request.

closes https://github.com/temporalio/sdk-go/issues/1458

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -39,7 +39,6 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	healthpb "google.golang.org/grpc/health/grpc_health_v1"
-	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/durationpb"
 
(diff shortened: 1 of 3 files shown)


cretz (approved):

internal/internal_workflow_client.go
```diff
@@ -1167,6 +1168,8 @@ func (wc *WorkflowClient) loadCapabilities(ctx context.Context, getSystemInfoTim
 	}
 	ctx, cancel := context.WithTimeout(ctx, getSystemInfoTimeout)
 	defer cancel()
+	// Manually append the namespace to the header since the request does not have it.
+	ctx = metadata.AppendToOutgoingContext(ctx, temporalNamespaceHeaderKey, wc.namespace)
```
Quinn-With-Two-Ns, line 1172: @cretz I am manually setting the namespace hear here because it appears to be needed

cretz (approved):
Still approved. We should consider, as a separate issue, tests in every SDK that use data plane in cloud with API key to confirm this works.

PR #1470, opened 2024-05-14
Add new VersioningIntent options

ShahabT:
## What was changed

Added `InheritBuildID` and `UseAssignmentRules` to `VersioningIntent` enum, replacing (now deprecated) `Default` and `Compatible` options, respectively.

## Why?

New options follow new versioning terminology and are more to-the-point. The old option can be confusing when used with new versioning rule concepts:
- `Compatible` could be seen as to have to do with using/not-using `CompatibleRedirectRule`s, while this option is not about redirect rules and solely means keep the same build ID as the parent/previous wf instead of selecting an independent build ID.
- `Default` is not clear because there is no more a "default version set" concept that this could refer to, instead the build ID has to be chosen based on the assignment rules. Moreover, the existence of the concept of "default build ID" (as use in enhanced DescribeTaskQueue API) which is a different (but related) concept will cause additional confusion if we were to keep this option name.

## Checklist

1. Closes 

2. How was this tested:

Unit test added.

3. Any docs updates needed?

Not yet.

### DIFF
--- a/temporal/build_id_versioning.go
+++ b/temporal/build_id_versioning.go
@@ -49,9 +49,9 @@ const (
 	// VersioningIntentInheritBuildID indicates the command should inherit the current Build ID of the
 	// Workflow triggering it, and not use Assignment Rules. (Redirect Rules are still applicable)
 	// This is the default behavior for commands running on the same Task Queue as the current worker.
-	VersioningIntentInheritBuildID= internal.VersioningIntentInheritBuildID
+	VersioningIntentInheritBuildID = internal.VersioningIntentInheritBuildID
 	// VersioningIntentUseAssignmentRules indicates the command should use the latest Assignment Rules
 	// to select a Build ID independently of the workflow triggering it.
 	// This is the default behavior for commands not running on the same Task Queue as the current worker.
-	VersioningIntentUseAssignmentRules= internal.VersioningIntentUseAssignmentRules
+	VersioningIntentUseAssignmentRules = internal.VersioningIntentUseAssignmentRules
 )
--- a/temporal/build_id_versioning.go
+++ b/temporal/build_id_versioning.go
@@ -39,17 +39,17 @@ const (
 	// version if possible. It may not be possible if the target task queue does not also have
 	// knowledge of the current worker's build ID.
 	// WARNING: Worker versioning is currently experimental
-	// Deprecated. This has the same effect as VersioningIntentInheritBuildId, use that instead.
+	// Deprecated. This has the same effect as VersioningIntentInheritBuildID, use that instead.
 	VersioningIntentCompatible = internal.VersioningIntentCompatible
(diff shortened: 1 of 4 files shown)


antlai-temporal (approved):

antlai-temporal:
One unrelated integration test (debugged by David) fails. It is one line fix, to pass the test suite with the new server... If you could add it... Thanks!

diff --git a/test/integration_test.go b/test/integration_test.go
index a0a6d55..3376473 100644
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -4270,7 +4270,7 @@ func (ts *IntegrationTestSuite) TestScheduleBackfillCreate() {
        defer func() {
                ts.NoError(handle.Delete(ctx))
        }()
-       time.Sleep(5 * time.Second)
+       time.Sleep(10 * time.Second)
        description, err := handle.Describe(ctx)
        ts.NoError(err)
        ts.EqualValues(60, description.Info.NumActions)

temporal/build_id_versioning.go
```diff
@@ -39,9 +39,19 @@ const (
...
 	VersioningIntentCompatible = internal.VersioningIntentCompatible
 	// VersioningIntentDefault indicates that the command should run on the target task queue's
 	// current overall-default build ID.
 	// WARNING: Worker versioning is currently experimental
+	// Deprecated. This has the same effect as VersioningIntentUseAssignmentRules, use that instead.
 	VersioningIntentDefault = internal.VersioningIntentDefault
+	// VersioningIntentInheritBuildID indicates the command should inherit the current Build ID of the
+	// Workflow triggering it, and not use Assignment Rules. (Redirect Rules are still applicable)
+	// This is the default behavior for commands running on the same Task Queue as the current worker.
+	VersioningIntentInheritBuildID= internal.VersioningIntentInheritBuildID
```
cretz, line 52:
```suggestion
	VersioningIntentInheritBuildID = internal.VersioningIntentInheritBuildID
```

Usually `go fmt` would catch this

temporal/build_id_versioning.go
```diff
@@ -39,9 +39,19 @@ const (
...
+	// VersioningIntentUseAssignmentRules indicates the command should use the latest Assignment Rules
+	// to select a Build ID independently of the workflow triggering it.
+	// This is the default behavior for commands not running on the same Task Queue as the current worker.
+	VersioningIntentUseAssignmentRules= internal.VersioningIntentUseAssignmentRules
```
cretz, line 56:
```suggestion
	VersioningIntentUseAssignmentRules = internal.VersioningIntentUseAssignmentRules
```

Sushisource (approved):

dnr (commented):
Why didn't we just rename the constants to the new names and then add aliases from the old names to the new ones to maintain source compatibility?

cretz:
I think from a user POV that's basically the same thing, but sure.

PR #1473, opened 2024-05-15
Execute nexus operation from a workflow

bergundy:
## What was changed

Added the ability to execute nexus operations from a workflow.
This PR is stacked on top of https://github.com/temporalio/sdk-go/pull/1466.

## Checklist

- [x] [Implement test framework logic](https://github.com/temporalio/sdk-go/pull/1475)

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1697,14 +1697,18 @@ func isCommandMatchEvent(d *commandpb.Command, e *historypb.HistoryEvent, obes [
 		eventAttributes := e.GetNexusOperationScheduledEventAttributes()
 		commandAttributes := d.GetScheduleNexusOperationCommandAttributes()
 
-		if eventAttributes.GetService() != commandAttributes.GetService() || eventAttributes.GetOperation() != commandAttributes.GetOperation() {
+		return eventAttributes.GetService() == commandAttributes.GetService() &&
+			eventAttributes.GetOperation() == commandAttributes.GetOperation()
+
+	case enumspb.COMMAND_TYPE_REQUEST_CANCEL_NEXUS_OPERATION:
+		if e.GetEventType() != enumspb.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUESTED {
 			return false
 		}
 
-		return true
+		eventAttributes := e.GetNexusOperationCancelRequestedEventAttributes()
+		commandAttributes := d.GetRequestCancelNexusOperationCommandAttributes()
 
-	case enumspb.COMMAND_TYPE_REQUEST_CANCEL_NEXUS_OPERATION:
-		return e.GetEventType() == enumspb.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUESTED
+		return eventAttributes.GetScheduledEventId() == commandAttributes.GetScheduledEventId()
 	}
 
 	return false
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -336,7 +336,9 @@ func isCommandEvent(eventType enumspb.EventType) bool {
 		enumspb.EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED,
 		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED,
 		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_COMPLETED,
-		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_REJECTED:
+		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_REJECTED,
+		enumspb.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED,
+		enumspb.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUESTED:
 		return true
 	default:
 		return false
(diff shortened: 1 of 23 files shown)


internal/internal_task_handlers.go
```diff
@@ -1687,6 +1689,22 @@ func isCommandMatchEvent(d *commandpb.Command, e *historypb.HistoryEvent, obes [
...
 		return true
+
+	case enumspb.COMMAND_TYPE_SCHEDULE_NEXUS_OPERATION:
+		if e.GetEventType() != enumspb.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED {
+			return false
+		}
+		eventAttributes := e.GetNexusOperationScheduledEventAttributes()
+		commandAttributes := d.GetScheduleNexusOperationCommandAttributes()
+
+		if eventAttributes.GetService() != commandAttributes.GetService() || eventAttributes.GetOperation() != commandAttributes.GetOperation() {
```
cretz, line 1700: To confirm, we allow endpoint to be changed without failing determinism but not service or operation? One might expect service and endpoint to be the same kind of thing, or at least it may be confusing to explain to users that half of the Nexus client creation can be changed and the other half cannot. I think it should be both endpoint and service or neither. An argument for both is that endpoints can be changed server side.

internal/workflow.go
```diff
@@ -2119,3 +2119,160 @@ func DeterministicKeysFunc[K comparable, V any](m map[K]V, cmp func(a K, b K) in
...
+
+func (wc *workflowEnvironmentInterceptor) ExecuteNexusOperation(ctx Context, client NexusClient, operation any, input any, options NexusOperationOptions) NexusOperationFuture {
+	mainFuture, mainSettable := newDecodeFuture(ctx, nil /* this param is never used */)
+	executionFuture, executionSettable := NewFuture(ctx)
+	result := &nexusOperationFutureImpl{
+		decodeFutureImpl: mainFuture.(*decodeFutureImpl),
+		executionFuture:  executionFuture.(*futureImpl),
+	}
+
+	// Immediately return if the context has an error without spawning the child workflow
```
cretz, line 2225:
```suggestion
	// Immediately return if the context has an error without spawning the operation
```

internal/workflow.go
```diff
@@ -2119,3 +2119,160 @@ func DeterministicKeysFunc[K comparable, V any](m map[K]V, cmp func(a K, b K) in
...
+	var ok bool
+	var operationName string
+	if operationName, ok = operation.(string); ok {
+	} else if regOp, ok := operation.(interface{ Name() string }); ok {
+		operationName = regOp.Name()
+	} else {
+		return executeNexusOperationParams{}, fmt.Errorf("invalid 'operation' parameter, must be an OperationReference or a string")
+	}
+	// TODO(bergundy): Validate operation types against input once there's a good way to extract the generic types from
+	// OperationReference in the Nexus Go SDK.
```
cretz, line 2202: I believe this can be done reflectively and is a valuable thing to do (we do it for other external calls where we know the outbound parameters)

internal/workflow.go
```diff
@@ -2119,3 +2119,160 @@ func DeterministicKeysFunc[K comparable, V any](m map[K]V, cmp func(a K, b K) in
...
+	params, err := wc.prepareNexusOperationParams(ctx, client, operation, input, options)
+	if err != nil {
+		executionSettable.Set(nil, err)
+		mainSettable.Set(nil, err)
+		return result
+	}
+
+	var operationID string
+	seq := wc.env.ExecuteNexusOperation(params, func(r *commonpb.Payload, e error) {
+		mainSettable.Set(&commonpb.Payloads{Payloads: []*commonpb.Payload{r}}, e)
```
cretz, line 2243: Is it possible for the returned payload to be null (instead of be present with a null body)?

cretz (commented):
Looks good, not much to add here. I didn't look into the guts of the state machine stuff, seems to be copied from child workflow. I hope we have enough tests to cover most angles there especially with regards to cancel before/during/after start/finish.

internal/internal_task_handlers.go
```diff
@@ -1687,6 +1689,22 @@ func isCommandMatchEvent(d *commandpb.Command, e *historypb.HistoryEvent, obes [
...
+			return false
+		}
+
+		return true
+
+	case enumspb.COMMAND_TYPE_REQUEST_CANCEL_NEXUS_OPERATION:
```
Quinn-With-Two-Ns, line 1706: Shouldn't we also check the ID here like other cancel operations?

internal/internal_workflow_testsuite.go
```diff
@@ -2279,6 +2279,14 @@ func (env *testWorkflowEnvironmentImpl) executeChildWorkflowWithDelay(delayStart
 	go childEnv.executeWorkflowInternal(delayStart, params.WorkflowType.Name, params.Input)
 }
 
+func (wc *testWorkflowEnvironmentImpl) ExecuteNexusOperation(params executeNexusOperationParams, callback func(*commonpb.Payload, error), startedHandler func(opID string, e error)) int64 {
+	panic("TODO")
```
Quinn-With-Two-Ns, line 2283: Do we have an issue for this?

bergundy: I just put it here in the PR checklist.
I started working on. I think I'll be able to submit it today though, it's not too bad.
I had to make a test env `Client` with support for some basic methods because the operations all require one.

bergundy: https://github.com/temporalio/sdk-go/pull/1475

bergundy (commented):
Addressing comments shortly.

internal/workflow.go
bergundy: https://github.com/temporalio/sdk-go/issues/1476

internal/internal_task_handlers.go
cretz: Unresolving to continue conversation. I think you should either allow the two things that create a client to be changed or disallow the two things that create a client from being changed. Having one of those two changeable is confusing. If you're looking for a task queue analogy, I think that's endpoint + service and if we had an "activity client", creating it would probably have the task queue like the nexus client has endpoint + service.

bergundy:
@cretz @Quinn-With-Two-Ns anything preventing your approval?

cretz (approved):
I was hoping to continue the conversation about having only half of the params for making a new nexus client be subject to determinism checks. But maybe another time.

Quinn-With-Two-Ns (approved):
I was waiting for test suite support in this PR, but doing it in a separate PR works to.

PR #1478, opened 2024-05-17
get-free-port: prevent duplicate ports on Linux

mjameswh:
## What changed

- When assigning ephemeral ports for use by the Temporal Server, on all platforms except MacOS and Windows, immediately open a connection to that port, and force the connection into `TIME_WAIT` state. This avoids various race conditions that could happen due to the port being reallocated between the moment the port is obtained and the moment the server actually binds to that port. See temporalio/cli#564 for details.

### DIFF
--- a/testsuite/freeport.go
+++ b/testsuite/freeport.go
@@ -1,3 +1,25 @@
+// The MIT License
+//
+// Copyright (c) 2022 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


testsuite/freeport.go
```diff
@@ -1,81 +1,75 @@
...
 import (
 	"fmt"
 	"net"
+	"runtime"
 )
 
-// Modified from Temporalite which itself modified from
-// https://github.com/phayes/freeport/blob/95f893ade6f232a5f1511d61735d89b1ae2df543/freeport.go
+// Copied and adapted from
+// https://github.com/mjameswh/temporalio-cli/blob/6b5a708e51da89dbdfd4fda7b7219ae12bcd8ffb/temporalcli/devserver/freeport.go
```
cretz, line 10: Can point to PR instead of personal repo

cretz (approved):
LGTM but let @Quinn-With-Two-Ns  confirm

testsuite/freeport.go
mjameswh: Indeed. Fixed.

testsuite/freeport.go
```diff
@@ -25,57 +25,73 @@ package testsuite
...
+// setting the `SO_REUSEADDR` socket option (Go already does that by default,
+// but other languages may not); otherwise, the OS may fail with a message such
+// as "address already in use". Windows default behavior is already appropriate
+// in this regard; on that platform, `SO_REUSEADDR` has a different meaning and
+// should not be set (setting it may have unpredictable consequences).
+func getFreePort(host string) (string, int, error) {
+	l, err := net.Listen("tcp", host+":0")
 	if err != nil {
-		return "", 0, err
+		return "", 0, fmt.Errorf("failed to assign a free port: %v", err)
```
Quinn-With-Two-Ns, line 51:
```suggestion
		return "", 0, fmt.Errorf("failed to assign a free port: %w", err)
```
nit

Quinn-With-Two-Ns (approved):

PR #1480, opened 2024-05-18
Send original update request back in response

Quinn-With-Two-Ns:
Server will use this value to reconstruct the update if it is lost in certain cases

Undoing this PR more or less https://github.com/temporalio/sdk-go/pull/1139


cretz (approved):
Is there an equivalent Java and Core issue?

EDIT: I have opened https://github.com/temporalio/sdk-java/issues/2073 and https://github.com/temporalio/sdk-core/issues/743.

internal/internal_update.go
```diff
@@ -157,8 +157,11 @@ func (up *updateProtocol) Accept() {
 		Body: protocol.MustMarshalAny(&updatepb.Acceptance{
 			AcceptedRequestMessageId:         up.requestMsgID,
 			AcceptedRequestSequencingEventId: up.requestSeqID,
+			AcceptedRequest:                  proto.Clone(&up.initialRequest).(*updatepb.Request),
```
alexshtin, line 160: Why clone?

alexshtin (approved):

Issue #1477, opened 2024-05-17
temporal_worker_task_slots_available only reports workflow slots when using opentelemetry metrics handler


antmendoza:
## Expected Behavior
SDK should report the [slots available](https://docs.temporal.io/references/sdk-metrics#worker_task_slots_available) for workflow, activity and local activity worker 

it works as expected when the metrics handler is tally https://github.com/temporalio/samples-go/blob/main/metrics/worker/main.go#L18-L24 

## Actual Behavior
Using the opentelemetry metrics handler 

```
	exp, err := otlpmetrichttp.New(ctx, otlpmetrichttp.WithEndpointURL("http://localhost:4318"))
	if err != nil {
		panic(err)
	}
	meterProvider := metric.NewMeterProvider(metric.WithReader(
		metric.NewPeriodicReader(exp, metric.WithInterval(1*time.Second)),
	))
	handler := opentelemetry.NewMetricsHandler(
		opentelemetry.MetricsHandlerOptions{
			Meter: meterProvider.Meter("temporal-sdk-go"),
		},
	)

```
the SDK only reports slots available for `worker_type="WorkflowWorker"`

![Screenshot 2024-05-17 at 17 22 44](https://github.com/temporalio/sdk-go/assets/514932/e636bea2-5478-45b1-8c8f-48fc67887537)

Other metrics that have different values for workflow_type seem ok. 

![Screenshot 2024-05-17 at 17 22 27](https://github.com/temporalio/sdk-go/assets/514932/9e0825d1-ad8b-4f85-a9c8-7970c9875783)

You can find a repro here 

https://github.com/antmendoza/my-temporal-pocs/tree/main/go/_5888

```
go.temporal.io/api v1.33.0
go.temporal.io/sdk v1.26.1
go.temporal.io/sdk/contrib/opentelemetry v0.5.0

```

Quinn-With-Two-Ns:
This appears to be a regression or at least change in behaviour of the [OTEL Go SDK](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric) testing with version `1.21.0` then `temporal_worker_task_slots_available` is reported correctly, but testing with `1.26.0` it is not. I have a work around for our integration and will need to open up and issue with the OTEL Go SDK if this is intentional or not.

PR #1482, opened 2024-05-19
Fix gauge metrics with latest otel SDK

Quinn-With-Two-Ns:
As reported [here](https://github.com/temporalio/sdk-go/issues/1477) there appears to be a change in behaviour in Go OTEL SDK library when creating multiple gauges with the same name. Now only the first callbacks registered callback is used and all others are ignored. This causes some metrics to be missing. To work around this until `SyncGauge` is released in the Go SDK I switched to using `RegisterCallback` as these callbacks are still called.

While working in this area I also made it so gauges don't emit zero if no value has been reported.


contrib/opentelemetry/go.mod
```diff
@@ -4,14 +4,14 @@ go 1.20
 
 require (
 	github.com/stretchr/testify v1.9.0
-	go.opentelemetry.io/otel v1.21.0
-	go.opentelemetry.io/otel/sdk v1.21.0
-	go.opentelemetry.io/otel/trace v1.21.0
+	go.opentelemetry.io/otel v1.23.0
+	go.opentelemetry.io/otel/sdk v1.23.0
+	go.opentelemetry.io/otel/trace v1.23.0
```
cretz, line 9: I assume 1.23 instead of 1.26 just because we want lowest-working version?

contrib/opentelemetry/handler.go
```diff
@@ -130,12 +130,20 @@ func (m MetricsHandler) Counter(name string) client.MetricsCounter {
 func (m MetricsHandler) Gauge(name string) client.MetricsGauge {
 	// TODO(https://github.com/open-telemetry/opentelemetry-go/issues/3984) replace with sync gauge once supported
 	var config atomic.Value
-	config.Store(0.0)
-	_, err := m.meter.Float64ObservableGauge(name,
-		metric.WithFloat64Callback(func(ctx context.Context, o metric.Float64Observer) error {
```
cretz, line 135: Hrmm, were you able to dig up what changed on their side for why this callback doesn't work for us anymore? I am just curious, if not no worries.

cretz (approved):

contrib/opentelemetry/go.mod
Quinn-With-Two-Ns: Yeah, this was the lowest version that our old implementation broke on

contrib/opentelemetry/handler.go
Quinn-With-Two-Ns: I wasn't able to find what exact commit broke it unfortunately

PR #1484, opened 2024-05-20
Add `SetFailureConverter` to the activity and workflow test environments

boxofrad:
## What was changed

Adds a `SetFailureConverter` method to `TestWorkflowEnvironment` and `TestActivityEnvironment`.

Supersedes #1433.

## Why?
We use a custom failure converter to marshal/unmarshal structured errors in our application. These errors contain a "code" (similar to a gRPC status) which gets lost without the failure converter, so there are currently certain branches of our code that we're unable to test.


Quinn-With-Two-Ns (approved):

boxofrad:
Thanks for the review @Quinn-With-Two-Ns and @cretz 🙌

I've merged the base branch, so it looks like CI might need another approval.

PR #1485, opened 2024-05-21
Switch to using Float64Gauge in otel integration

Quinn-With-Two-Ns:
Otel Go SDK just released `Float64Gauge` in `v1.27.0` so we can use that instead of our hacks around async gauges.


contrib/opentelemetry/go.mod
```diff
@@ -1,12 +1,14 @@
 module go.temporal.io/sdk/contrib/opentelemetry
 
-go 1.20
+go 1.21
+
+toolchain go1.22.2
```
cretz, line 5: Reading https://go.dev/ref/mod#go-mod-file-toolchain, I am trusting this has no effect on users using 1.21?

cretz (approved):

contrib/opentelemetry/go.mod
Quinn-With-Two-Ns: It has no effect on users, that  being said it messes with our CI because it causes the unit tests to always be run with  GO version `1.22.2` so I will remove.

PR #1487, opened 2024-05-22
Bump api version to 1.34.0

gow:
## What was changed
Bumped API version to 1.34.0

## Why?
There is a new release.

## Checklist

1. Closes 

2. How was this tested:
manually verified.

3. Any docs updates needed?
No


Quinn-With-Two-Ns:
Looks like `https://github.com/grpc/grpc-go` deprecates `Dail`
```
converter/grpc_interceptor_test.go:77:12: grpc.Dial is deprecated: use NewClient instead.  Will be supported throughout 1.x.  (SA1019)
```

We can probably just add comments to ignore the deprecation for now, we could also replace with `NewClient` but it will behave slightly differently and that could effect users

```
One subtle difference between NewClient and Dial and DialContext is that the former uses "dns" as the default name resolver, while the latter use "passthrough" for backward compatibility. This distinction should not matter to most users, but could matter to legacy users that specify a custom dialer and expect it to receive the target string directly.
``

Quinn-With-Two-Ns (approved):

cretz (approved):

Quinn-With-Two-Ns:
Superseded by https://github.com/temporalio/sdk-go/commit/4c8b9e16567616c47a8485a4f6f6d8a022baa1c2

PR #1488, opened 2024-05-23
Replace grpc.Dial with grpc.NewClient

gow:
## What was changed
calls to `grpc.Dial()` have been replaced with `grpc.NewClient()`

## Why?
`grpc.Dial()` is deprecated

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/converter/grpc_interceptor_test.go
+++ b/converter/grpc_interceptor_test.go
@@ -74,7 +74,7 @@ func TestPayloadCodecGRPCClientInterceptor(t *testing.T) {
 	)
 	require.NoError(err)
 
-	c, err := grpc.Dial(
+	c, err := grpc.NewClient(
 		server.addr,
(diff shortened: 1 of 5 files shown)


Quinn-With-Two-Ns:
`Dial` and `NewClient` do behave slightly differently https://github.com/grpc/grpc-go/blob/v1.64.0/clientconn.go#L216, we can probably just call this out in the release notes WDYT @cretz

cretz:
@Quinn-With-Two-Ns - yeah, which resolver is used by default won't matter to most, and if it does it'll break upon connection. We even document on `ClientOptions.HostPort` that while `dns` resolver is registered by default, they have to use `dns:///` to opt-in to using it, but now this change will use it by default, so can probably clean up that Godoc. Let's just make it real clear in release notes.

Issue #1449, opened 2024-04-25
Clarify waiting semantics for UpdateWorkflow


drewhoskins-temporal:
**Is your feature request related to a problem? Please describe.**
Per https://github.com/temporalio/sdk-go/blob/836106795bd31710bc509f8a306a876e4be70429/client/client.go#L575-L586
it's unclear from the docstrings how to start a workflow and wait for acceptance/rejection.
I see https://github.com/temporalio/sdk-go/blob/master/internal/internal_workflow_client.go#L756 and, per the parent task, want to make sure we require `WaitPolicy` and don't default to anything that could be dangerous.

**Describe the solution you'd like**
Make sure availability tradeoff is documented clearly and user is opting in to a wait policy.

**Describe alternatives you've considered**

**Additional context**
Parent: https://github.com/temporalio/features/issues/469

PR #1489, opened 2024-05-24
Workflow update client API refactor

Quinn-With-Two-Ns:
This PR makes multiple changes to the client experience around update:
* Merge `UpdateWorkflowWithOptions` and `UpdateWorkflow` to a single api 
* Makes `WaitForStage` a required parameter in update
* Create a Go SDK enum for update lifecyle
*  Updated start update polling to poll until stage reached or ACCEPTED. If more polling needed, switches to result polling.

closes https://github.com/temporalio/sdk-go/issues/1414
closes https://github.com/temporalio/sdk-go/issues/1449
closes https://github.com/temporalio/sdk-go/issues/1411

since this has multiple breaking changes feature tests are expected to fail until I make a separate branch

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -215,10 +215,10 @@ type (
 	// ScheduleBackfillOptions configure the parameters for backfilling a schedule.
 	ScheduleBackfillOptions = internal.ScheduleBackfillOptions
 
-	// UpdateWorkflowOptionsRequest encapsulates the parameters for
+	// UpdateWorkflowOptions encapsulates the parameters for
 	// sending an update to a workflow execution.
 	// NOTE: Experimental
-	UpdateWorkflowOptionsRequest = internal.UpdateWorkflowOptionsRequest
+	UpdateWorkflowOptions = internal.UpdateWorkflowOptions
 
 	// WorkflowUpdateHandle represents a running or completed workflow
 	// execution update and gives the holder access to the outcome of the same.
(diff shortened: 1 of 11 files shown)


client/client.go
```diff
@@ -566,24 +585,16 @@ type (
...
 		// is running in in parallel with the calling thread. Errors returned
 		// from the server will be exposed through the return value of
 		// WorkflowUpdateHandle.Get(). Errors that occur before the
 		// update is requested (e.g. if the required workflow ID field is
-		// missing from the UpdateWorkflowWithOptionsRequest) are returned
+		// missing from the UpdateWorkflowOptionsRequest) are returned
 		// directly from this function call.
 		// NOTE: Experimental
-		UpdateWorkflowWithOptions(ctx context.Context, request *UpdateWorkflowWithOptionsRequest) (WorkflowUpdateHandle, error)
+		UpdateWorkflow(ctx context.Context, request *UpdateWorkflowOptionsRequest) (WorkflowUpdateHandle, error)
```
cretz, line 597:
```suggestion
		UpdateWorkflow(ctx context.Context, options UpdateWorkflowOptions) (WorkflowUpdateHandle, error)
```

Wonder if it is a "request" or "options", and since it will always be required, probably not value in accepting a pointer (I know we are inconsistent here). Even if it is a "request", then can probably remove the word "Options" from within. I don't have strong opinions here.

internal/internal_workflow_client.go
```diff
@@ -1792,31 +1779,69 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
...
-				Name:   in.UpdateName,
-				Args:   argPayloads,
-			},
-		},
-	})
-	if err != nil {
-		return nil, err
+	desiredLifecycleStage := updateLifeCycleStageToProto(in.WaitForStage)
+	var resp *workflowservice.UpdateWorkflowExecutionResponse
+	for ctx.Err() == nil {
```
cretz, line 1784:
```suggestion
	for {
```

Can probably just let the gRPC call fail as it would with other calls when a context is closed

client/client.go
cretz: :+1: I can see arguments for both `Request` and `Options` for the value of `X` in `UpdateWorkflowX`, but probably don't need both. (I have no strong opinion on which, I kinda like "options" but it's a very weak preference, your choice)

cretz (approved):

PR #1490, opened 2024-05-24
Fix bugs when using a custom `FailureConverter` in tests

boxofrad:
## What was changed

Follow up to #1484, fixing a couple of issues with using a custom failure converter:

1. We weren't correctly threading the failure converter down to activities and child workflows, so they were still using the default converter.
2. The `NonRetryable` flag on the protobuf failure type wasn't being honored for custom errors.

## Checklist

1. How was this tested:

I tested this change against our app's test suite, but it could probably use a case in the SDK's test suite itself. I'd be happy to add that if you're happy with this approach generally.


Quinn-With-Two-Ns:
Thanks for another contribution, could we also include some test to verify the fixes you made and make sure we don't regress in the future?

boxofrad:
Yep, sure thing! It's a long weekend here in the UK, but I'll pick it up next week 🙇🏻

boxofrad:
Hey @Quinn-With-Two-Ns 👋🏻  I've added a test case for this now, let me know what you think!

Issue #1420, opened 2024-03-12
Support the new versioning API


antlai-temporal:
**Is your feature request related to a problem? Please describe.**
The versioning API is changing with new abstractions that need to be supported by the Go SDK.

cretz:
@antlai-temporal - does this apply to all SDKs? Maybe an overarching features issue that links to the other SDK issues could be helpful. Or if you're just doing Go first, having that issue so that you can link to the other issues when you create them later would be helpful.

PR #1494, opened 2024-05-29
Support the new versioning API

antlai-temporal:
## What was changed

This is a PR to support the new versioning API in the SDK-go that combines other (partially) reviewed PRs in the versioning-2 
branch. In particular:

#1453 
#1431 
#1432 
#1429 
#1470 

This PR will remain in draft mode until we can automate integration tests with CI, which requires a new docker image of the future server 1.24 (or a release candidate) for ubuntu, and a new CLI for mac/windows. Currently, all the integration tests pass using a local, up-to-date server (ubuntu).

However, I would appreciate early feedback to minimize delay on this important pre-release feature.  Thanks for your help!

## Checklist

1. Closes 
#1420


internal/client.go
```diff
@@ -352,6 +352,25 @@ type (
 		// GetWorkerTaskReachability returns which versions are is still in use by open or closed workflows.
 		GetWorkerTaskReachability(ctx context.Context, options *GetWorkerTaskReachabilityOptions) (*WorkerTaskReachability, error)
 
+		// DescribeTaskQueueEnhanced returns information about the target task queue, broken down by Build Id:
+		//   - List of pollers
+		//   - Workflow Reachability status
+		//   - Backlog info for Workflow and/or Activity tasks
+		// When not supported by the server, it returns an empty [TaskQueueDescription]
```
Quinn-With-Two-Ns, line 359: Doesn't the implementation return an error?

internal/internal_versioning_client.go
```diff
@@ -0,0 +1,319 @@
...
+		Identity string
+		// Polling rate. A value of zero means it was not set.
+		RatePerSecond float64
+		// Optional poller versioning capabilities. Available when a worker has opted into the worker versioning feature.
+		WorkerVersionCapabilities *WorkerVersionCapabilities
+	}
+
+	// TaskQueueTypeInfo specifies task queue information per task type and Build ID.
+	// It is included in [TaskQueueVersionInfo].
+	TaskQueueTypeInfo struct {
```
Quinn-With-Two-Ns, line 129: Is there a reason we did not include `TaskQueueStats`?

internal/worker_version_sets.go
```diff
@@ -45,10 +45,20 @@ const (
 	// VersioningIntentCompatible indicates that the command should run on a worker with compatible
 	// version if possible. It may not be possible if the target task queue does not also have
 	// knowledge of the current worker's build ID.
+	// Deprecated. This has the same effect as VersioningIntentInheritBuildID, use that instead.
```
Quinn-With-Two-Ns, line 48: I think it has to be `Deprecated:` not `Deprecated.` and it may have to be in its own paragraph

https://go.dev/wiki/Deprecated

internal/internal_versioning_client.go
antlai-temporal: They are coming, see this PR  https://github.com/temporalio/api/pull/414, but not waiting for this ...

internal/worker_version_sets.go
antlai-temporal: Thanks, I'll make sure they are all properly formatted.

internal/internal_versioning_client.go
Quinn-With-Two-Ns: Yep  all good, for some reason I thought `TaskQueueStats` was going to be done the same time versioning was.

client/client.go
```diff
@@ -231,53 +270,189 @@ type (
...
 	WorkerTaskReachability = internal.WorkerTaskReachability
 
 	// BuildIDReachability describes the reachability of a buildID
-	// WARNING: Worker versioning is currently experimental
+	// WARNING: Legacy worker versioning is currently experimental
 	BuildIDReachability = internal.BuildIDReachability
 
 	// TaskQueueReachability Describes how the Build ID may be reachable from the task queue.
-	// WARNING: Worker versioning is currently experimental
+	// WARNING: Legacy worker versioning is currently experimental
```
Quinn-With-Two-Ns, line 317: Is "Legacy worker"  versioning deprecated?

internal/client.go
antlai-temporal: The server does not return an error, at the wire level we use the old `describe` api and all the `enhanced fields` in the protobuf are just ignored. The result returned could be non-empty but it is likely incorrect because some of the ignored fields change the query. What we do is just ignore these values when deserializing the protobuf and return an empty TaskQueueDescription.

client/client.go
antlai-temporal: Yes, the docs team are going to use "Legacy worker versioning" or "Legacy versioning" for the deprecated one, and plain "worker versioning"  for the current one.

antlai-temporal: Do you want me to mark all the constants as `Deprecated:` and remove the experimental?

Quinn-With-Two-Ns: Yes I think we should notify users these APIs are deprecated, do we plan to remove these APIs at some point as well? if so we should also note that in the deprecation  description.

internal/client.go
Quinn-With-Two-Ns: Sorry maybe I am missing something,  but doesn't the [implementation ](https://github.com/temporalio/sdk-go/pull/1494/files#diff-e86803d9c6ede32f27826e8eff45f92375372cb213b6b46e1c0513633e2957ccR1062)of `DescribeTaskQueueEnhanced` call  `detectTaskQueueEnhancedNotSupported` and that will  return an error if it is not supported?

internal/internal_versioning_client.go
```diff
@@ -0,0 +1,319 @@
...
+
+	taskQueueTypes := make([]enumspb.TaskQueueType, len(o.TaskQueueTypes))
+	for i, t := range o.TaskQueueTypes {
+		taskQueueTypes[i] = taskQueueTypeToProto(t)
+	}
+
+	opt := &workflowservice.DescribeTaskQueueRequest{
+		Namespace: namespace,
+		TaskQueue: &taskqueuepb.TaskQueue{
+			// Sticky queues not supported
```
Quinn-With-Two-Ns, line 167: nit: Think we should note this in `DescribeTaskQueueEnhanced`?

internal/internal_versioning_client.go
```diff
@@ -0,0 +1,319 @@
...
+
+// TaskQueueType specifies which category of tasks are associated with a queue.
+type TaskQueueType int
+
+const (
+	// TaskQueueTypeUnspecified indicates the task queue type was not specified.
+	TaskQueueTypeUnspecified = iota
+	// TaskQueueTypeWorkflow indicates the task queue has workflow type.
+	TaskQueueTypeWorkflow
+	// TaskQueueTypeActivity indicates the task queue has activity type.
```
Quinn-With-Two-Ns, line 44: nit: I prefer the wording we have for `TaskQueueTypeNexus ` vs the wording for `TaskQueueTypeWorkflow`

internal/internal_versioning_client.go
```diff
@@ -0,0 +1,319 @@
...
+		// Include task reachability for the requested versions and all task types
+		// (task reachability is not reported per task type).
+		ReportTaskReachability bool
+	}
+
+	// WorkerVersionCapabilities includes a worker's build identifier
+	// and whether it is choosing to use the versioning feature.
+	// It is an optional component of [PollerInfo].
+	WorkerVersionCapabilities struct {
+		BuildID       string
```
Quinn-With-Two-Ns, line 110: nit: I know it is annoying but we should add GoDocs for this

internal/client.go
antlai-temporal: I need to be more explicit in the docs. `detectTaskQueueEnhancedNotSupported`  is a best effort function. If there are no pollers and no taskqueue status for the requested queue the response is empty, so I cannot tell... Otherwise, yes, I get an error.

Quinn-With-Two-Ns: Yeah may be better to be consistent then if we can't always return an error if it is not supported

client/client.go
```diff
@@ -231,53 +270,189 @@ type (
...
+
+	// VersioningOpInsertAssignmentRule is an operation for UpdateWorkerVersioningRulesOptions
+	// that inserts the rule to the list of assignment rules for this Task Queue.
+	// The rules are evaluated in order, starting from index 0. The first
+	// applicable rule will be applied and the rest will be ignored.
+	// By default, the new rule is inserted at the beginning of the list
+	// (index 0). If the given index is too larger the rule will be
+	// inserted at the end of the list.
+	// WARNING: Worker versioning is currently experimental
+	VersioningOpInsertAssignmentRule = internal.VersioningOpInsertAssignmentRule
```
Quinn-With-Two-Ns, line 398: Generally initialisms in Go prefer constant case so either `op` or `OP`. I would prefer `Operation` here since `op` could be `operation` or `option` and without reading the docs I can't tell. @cretz any thoughts?
 
https://google.github.io/styleguide/go/decisions#initialisms

internal/client.go
```diff
@@ -352,6 +352,25 @@ type (
...
+		// WARNING: Worker versioning is currently experimental, and requires server 1.24+
+		DescribeTaskQueueEnhanced(ctx context.Context, options *DescribeTaskQueueEnhancedOptions) (TaskQueueDescription, error)
```
Quinn-With-Two-Ns, line 361: I know the old new version took pointers for options, I think @cretz prefers it without pointers. The SDK is **not** consistent here, but we should try to be consistent for new APIs. I think without a pointer is probably nicer since we don't have to deal with`nil`.

internal/client.go
antlai-temporal: Thinking about this a bit more. The only case that we cannot detect is the empty return case, but we get correct responses regardless if we return empty.
 All the other cases are handled ok by `detectTaskQueueEnhancedNotSupported` and we get an error, which gives as more useful info than just returning empty.
I'm inclined to keep it "inconsistent" but make it more clear in the docs. In fact, it is more "incomplete" than "inconsistent" because we always get correct answers...

antlai-temporal: something like this:
// When not supported by the server, it returns an empty [TaskQueueDescription] if there is no information
// about the task queue, or an error when the response identifies an unsupported server.

client/client.go
antlai-temporal: There is no clear plans when or how will get removed, but I will make it explicit that they are being replaced by the new versioning apis

internal/internal_versioning_client.go
antlai-temporal: It is in the (internal) `DescribeTaskQueueEnhancedOptions` docs, but yes, it is worth making this more clear at the top level.

internal/internal_versioning_client.go
antlai-temporal: me too

internal/internal_versioning_client.go
antlai-temporal: Thanks, I missed that one.

client/client.go
antlai-temporal: I was mostly following what we had with the old versioning, e.g., `BuildIDOpAddNewIDInNewDefaultSet`. I guess that strictly speaking Op is not an initialism, it is just an abbreviation, but we could make it OP. My only concern is that in the middle of a word sometimes it is harder to "parse", i.e.,  VersioningOPAddRedirectRule vs VersioningOpAddRedirectRule , is "OPA" or "PA" (you need look-ahead :-) ).

About the confusion with Options, we always have "Options" with no shortcuts, so that is less confusing with "OP" for operations. I worry if we extend operations it is just too much typing, e.g., VersioningOperationDeleteRedirectRule...

Anyway, happy to change things, @cretz comments?

internal/client.go
antlai-temporal: Happy to change that, I saw 50/50 of APIs using pointers, and the old versioning always using it. 
But if we do change it, we should also change the new `UpdateWorkerVersioningRules` and `GetWorkerVersioningRules` to not have pointers, so that all are consistent...
@cretz what do you think?

internal/client.go
Quinn-With-Two-Ns: works for me

client/client.go
```diff
@@ -231,53 +270,189 @@ type (
...
+
+	// TaskQueueTypeInfo specifies task queue information per task type and Build ID.
+	// It is included in [Client.TaskQueueVersionInfo].
+	// WARNING: Worker versioning is currently experimental.
+	TaskQueueTypeInfo = internal.TaskQueueTypeInfo
+
+	// PollerInfo provides information about a worker/client polling a task queue.
+	// It is used by [Client.TaskQueueTypeInfo].
+	// WARNING: Worker versioning is currently experimental.
+	PollerInfo = internal.PollerInfo
```
cretz, line 346: Can this be called `TaskQueuePollerInfo` or `WorkerPollerInfo` or something? I think `PollerInfo` is a bit too generic for this big package.

client/client.go
cretz: I have no strong opinion. `Operation` works for me. I do appreciate how we're prefixing all of these w/ the same prefix though.

internal/client.go
cretz: :+1: Confirmed on Slack, without pointers is best IMO

internal/internal_workflow_client.go
```diff
@@ -1036,6 +1036,81 @@ func (wc *WorkflowClient) GetWorkerTaskReachability(ctx context.Context, options
 	return converted, nil
 }
 
+// DescribeTaskQueueEnhanced returns information about the target task queue, broken down by Build Id:
+//   - List of pollers
+//   - Workflow Reachability status
+//   - Backlog info for Workflow and/or Activity tasks
+//
+// WARNING: Worker versioning is currently experimental, and requires server 1.24+
+func (wc *WorkflowClient) DescribeTaskQueueEnhanced(ctx context.Context, options *DescribeTaskQueueEnhancedOptions) (TaskQueueDescription, error) {
```
cretz, line 1045: It is a bit confusing to have half of versioning client code in this file and half in another, but I understand why. There is an argument to be made for a `VersioningClient()` similar to our `ScheduleClient()`, but I'm ok w/ this since it's only 3 methods. No need to change anything, just thinking out loud.

internal/worker_versioning_rules.go
cretz, line 1: I think all of this should go in `internal_versioning_client.go` instead of a new file, but not a super-strong opinion

cretz (commented):
LGTM, only a couple of minor comments

client/client.go
antlai-temporal: TaskQueuePollerInfo sounds great

cretz (approved):

Quinn-With-Two-Ns (approved):

PR #1495, opened 2024-05-29
Minor logger improovments

jetexe:
## What was changed

1. Move `WithLogger` interface close to others (same package, different file)
2. Add skip 1 for default `WithLogger` implementation

## Why?

1. Moving  `WithLogger` interface allow user to see all available interfaces in one place. It increase user experience
2. Adding `Skip` function to the `withLogger` constructor functions allows to skip `with_logger.go` from the stack

## Checklist

1. How was this tested: All existing tests passed, local using replaced package
1. Any docs updates needed? nope

### DIFF
--- a/log/with_logger.go
+++ b/log/with_logger.go
@@ -24,11 +24,6 @@
 
 package log
 
-// WithLogger is an interface that prepend every log entry with keyvals.
-type WithLogger interface {
-	With(keyvals ...interface{}) Logger
-}
-
 // With creates a child Logger that includes the supplied key-value pairs in each log entry. It does this by
 // using the supplied logger if it implements WithLogger; otherwise, it does so by intercepting every log call.
 func With(logger Logger, keyvals ...interface{}) Logger {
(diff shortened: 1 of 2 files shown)


log/with_logger.go
```diff
@@ -57,7 +52,7 @@ type withLogger struct {
 }
 
 func newWithLogger(logger Logger, keyvals ...interface{}) *withLogger {
-	return &withLogger{logger: logger, keyvals: keyvals}
+	return &withLogger{logger: Skip(logger, 1), keyvals: keyvals}
```
Quinn-With-Two-Ns, line 55: Just to confirm, this code path would only matter if the logger did not implement `WithLogger` but did implement `WithSkipCallers`?

jetexe: Yes, you right.
`Skip` function will ignore logger in case it doesn't implement `WithSkipCallers` interface

Quinn-With-Two-Ns: hm I wonder how common this actually is, but in principle this change makes sense I think

Quinn-With-Two-Ns (approved):

PR #1498, opened 2024-05-31
testsuite.StartDevServer: Respect timeout during dial

abhinav:
## What was changed

**Background:**
testsuite.StartDevServer accepts a context.Context
meant to indicate how long the caller is willing to wait for the result.
This context is used when downloading the Temporal CLI,
but is not considered when trying to dial to the server in
`waitServerReady`.

**Change:**
This changes `waitServerReady` to respect the timeout configured in the
context--checking in on it after each attempt, and returning early
if the context has expired.
Similarly, the `Dial` attempt now uses `DialContext` so that if the context
expires, the dial operation also returns early if it can.

## Why?

Without this, if an attempt to start the server failed
(e.g. because of an invalid `--log-format` argument),
`waitServerReady` will block the test for 1 minute
before giving up and returning an error.
This is a pretty long time to wait for a test to fail,
when the caller is expecting, say, 5 seconds at most.

## Checklist

1. How was this tested:
   A unit test was added for the new functionality,
   and run with `go test -run WaitServerReady -count 1000`
   to verify that it's not flaky.

3. Any docs updates needed?
   I don't think so.

### DIFF
--- a/testsuite/devserver.go
+++ b/testsuite/devserver.go
@@ -339,9 +339,9 @@ func extractZip(r io.Reader, toExtract string, w io.Writer) error {
 // Returns a connected client created using the provided options.
 func waitServerReady(ctx context.Context, options client.Options) (client.Client, error) {
 	var returnedClient client.Client
-	lastErr := retryFor(600, 100*time.Millisecond, func() error {
+	lastErr := retryFor(ctx, 600, 100*time.Millisecond, func() error {
 		var err error
-		returnedClient, err = client.Dial(options)
+		returnedClient, err = client.DialContext(ctx, options)
 		return err
 	})
 	if lastErr != nil {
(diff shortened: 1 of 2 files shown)


testsuite/devserver.go
```diff
@@ -363,7 +363,12 @@ func retryFor(maxAttempts int, interval time.Duration, cond func() error) error
 		} else {
 			lastErr = curE
 		}
-		time.Sleep(interval)
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(interval):
```
cretz, line 369: Would prefer a ticker

cretz (approved):
LGTM, minor suggestion

testsuite/devserver.go
abhinav: Updated

cretz: :+1: Not a problem (ticker adjusts itself to slow receivers)

cretz:
LGTM, will let @Quinn-With-Two-Ns review

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
CI failing looks like an issue with the server `v1.24.0` docker images. Will follow up with the server team.

PR #1507, opened 2024-06-10
Add integration test for reset with update events

Quinn-With-Two-Ns:
Add integration test for reset with update events.

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -1806,6 +1806,63 @@ func (ts *IntegrationTestSuite) TestResetWorkflowExecution() {
 	ts.Equal(originalResult, newResult)
 }
 
+func (ts *IntegrationTestSuite) TestResetWorkflowExecutionWithUpdate() {
+	ctx := context.Background()
+	wfId := "reset-workflow-execution-with-update"
+	run, err := ts.client.ExecuteWorkflow(ctx,
+		ts.startWorkflowOptions(wfId), ts.workflows.UpdateBasicWorkflow)
+	ts.NoError(err)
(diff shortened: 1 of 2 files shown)


test/integration_test.go
```diff
@@ -1801,6 +1801,63 @@ func (ts *IntegrationTestSuite) TestResetWorkflowExecution() {
...
+	// Reset the workflow
+	resp, err := ts.client.ResetWorkflowExecution(ctx, &workflowservice.ResetWorkflowExecutionRequest{
+		Namespace: ts.config.Namespace,
+		WorkflowExecution: &commonpb.WorkflowExecution{
+			WorkflowId: run.GetID(),
+			RunId:      run.GetRunID(),
+		},
+		Reason:                    "integration test",
+		WorkflowTaskFinishEventId: 4,
+		ResetReapplyType:          enumspb.RESET_REAPPLY_TYPE_ALL_ELIGIBLE,
```
Quinn-With-Two-Ns, line 1836: @dandavison The `api` docs say the default value for `ResetReapplyType` is `RESET_REAPPLY_TYPE_ALL_ELIGIBLE`, but that does not appear to be the case as if I don't specify it manually the test fails. Is this a bug?

Sushisource: I think that's because you also have specified exclude. You can't specify both. One or the other.

Sushisource (approved):

test/integration_test.go
Sushisource: I throw an error on that in Core actually

PR #1509, opened 2024-06-11
Allow SDK to handle speculative WFT with command events

Quinn-With-Two-Ns:
Allow SDK to handle speculative WFT with command events. Originally I thought this change was going to be quite bad, but I realized I didn't need to touch the cacheing logic and could keep it all in the history iteration.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -592,16 +592,15 @@ func (w *workflowExecutionContextImpl) Unlock(err error) {
 	defer w.mutex.Unlock()
 	if err != nil || w.err != nil || w.isWorkflowCompleted ||
 		(w.wth.cache.MaxWorkflowCacheSize() <= 0 && !w.hasPendingLocalActivityWork()) {
-		// TODO: in case of closed, it asumes the close command always succeed. need server side change to return
+		// TODO: in case of closed, it assumes the close command always succeed. need server side change to return
 		// error to indicate the close failure case. This should be rare case. For now, always remove the cache, and
 		// if the close command failed, the next command will have to rebuild the state.
 		if w.wth.cache.getWorkflowCache().Exist(w.workflowInfo.WorkflowExecution.RunID) {
 			w.wth.cache.removeWorkflowContext(w.workflowInfo.WorkflowExecution.RunID)
 			w.cached = false
-		} else {
-			// sticky is disabled, manually clear the workflow state.
-			w.clearState()
 		}
+		// Clear the state so other tasks waiting on the context know it should be discarded.
+		w.clearState()
 	} else if !w.cached {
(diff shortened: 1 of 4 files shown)


internal/internal_task_handlers.go
```diff
@@ -115,8 +115,10 @@ type (
 		isWorkflowCompleted bool
 		result              *commonpb.Payloads
 		err                 error
-
+		// previousStartedEventID is the event ID of the workflow task started of the previous workflow task.
```
Sushisource, line 118:
```suggestion
		// previousStartedEventID is the ID of the workflow task started event of the previous workflow task.
```

Or also keep "event" before ID if you want

internal/internal_task_handlers.go
```diff
@@ -115,8 +115,10 @@ type (
...
 		previousStartedEventID int64
+		// lastHandledEventID is the event ID of the last command that the workflow state machine processed.
+		lastHandledEventID int64
```
Sushisource, line 121:
```suggestion
		// lastHandledEventID is the event ID of the last command-event that the workflow state machine processed.
		lastHandledEventID int64
```

Sushisource: Is this actually true though? It looks like this is set for _all_ processed events, not just command-events

Sushisource (approved):
Makes sense to me, just minor comment changes

internal/internal_task_handlers.go
Quinn-With-Two-Ns: Yeah I didn't mean command event here, but saying event would be more clear

cretz (commented):
I admit not understanding this very much, but is it possible to have an integration test that would break without this change but works now?

Quinn-With-Two-Ns:
>I admit not understanding this very much, but is it possible to have an integration test that would break without this change but works now?

No, the server won't trigger this code path yet.

cretz:
> No, the server won't trigger this code path yet.

Arguably this dead code shouldn't make it into the repo until it can be exercised (I fear adding code that we can't confirm works in end-to-end in CI). But if we're confident there are no dangers in adding this and it'll just remain dead code for now, no prob.

Is there a server PR that, when released, will trigger this? Would like us to track the server work to make sure we can add the integration test when we can.

Quinn-With-Two-Ns:
>Arguably this dead code shouldn't make it into the repo until it can be exercised 

This code is exercised by unit tests and is called in integration tests as well. There is a clear contract between the SDK and Server we are verifying the SDK side satisfies it's side of the contract.

>Would like us to track the server work to make sure we can add the integration test when we can.

Agree, the server work is tracked in Jira, but I'll open up a feature issue for testing this scenario across all SDKs.

cretz (approved):

PR #1514, opened 2024-06-14
Update some docs around grpc Dial options

Quinn-With-Two-Ns:
Update some docs around grpc Dial options because `grpc-go` deprecated `grpc.Dial` so we switched to `grpc.NewClient` that doesn't support some of these dial options because they are anti patterns not recommended by the gRPC team.

https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md 

Another thing we could do is add a flag to use the old deprecated client options to help who did rely on these options.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -538,7 +538,8 @@ type (
 		MaxPayloadSize int
 
 		// Advanced dial options for gRPC connections. These are applied after the internal default dial options are
-		// applied. Therefore any dial options here may override internal ones.
+		// applied. Therefore any dial options here may override internal ones. Dial options WithBlock, WithTimeout,
+		// WithReturnConnectionError, and FailOnNonTempDialError are ignored.
 		//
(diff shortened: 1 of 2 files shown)


Sushisource (approved):

internal/client.go
```diff
@@ -538,7 +538,8 @@ type (
 		MaxPayloadSize int
 
 		// Advanced dial options for gRPC connections. These are applied after the internal default dial options are
-		// applied. Therefore any dial options here may override internal ones.
+		// applied. Therefore any dial options here may override internal ones. Dial options WithBlock, WithTimeout,
+		// WithReturnConnectionError, and FailOnNonTempDialError are ignored.
```
cretz, line 542:
```suggestion
		// WithReturnConnectionError, and FailOnNonTempDialError are ignored since [grpc.NewClient] is used
```

Might as well tell them why, but no biggie

cretz (approved):

PR #1515, opened 2024-06-15
expose mock Calls for testing activity order

yinsidij:
## Why?

I would like to test the order of activity execution in the workflow unit tests. Looks like it’s lack of the feature.
Exposing the Calls of mock would be one solution.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Quinn-With-Two-Ns:
To test the order of mocks you can use `OnActivity(..).NotBefore()` This is a feature in `testify` https://github.com/stretchr/testify/issues/741

yinsidij:
> stretchr/testify#741

Thank you for the quick response @Quinn-With-Two-Ns . it works.

Issue #1450, opened 2024-04-25
Metric `workflow_task_execution_failed` is used with different sets of labels causing errors with Prometheus


wsny:
## Expected Behavior
Should not receive errors about metrics being registered incorrectly.

## Actual Behavior
Receiving a Prometheus error about the workflow_task_execution_failed metric being registered with different labels.

```
a previously registered descriptor with the same fully-qualified name as Desc{fqName: "temporal_workflow_task_execution_failed_total", help: "temporal_workflow_task_execution_failed_total counter", constLabels: {}, variableLabels: {app,client_name,namespace,environment,workflow_type,failure_reason,task_queue,worker_type,deployment,activity_type,env}} has different label names or a different help string
```

## Steps to Reproduce the Problem

I am still working on STR's (don't know how to break all the below with test code yet), but I can see that this metric is being used with different tags in different places.

#1295 added a tag here
https://github.com/temporalio/sdk-go/blob/69bc6c3a19cdb21d5252c3ad4a490109b9a2f39c/internal/internal_task_pollers.go#L421

But it is not being added in these uses of the metric.
https://github.com/temporalio/sdk-go/blob/69bc6c3a19cdb21d5252c3ad4a490109b9a2f39c/internal/internal_event_handlers.go#L1108
https://github.com/temporalio/sdk-go/blob/69bc6c3a19cdb21d5252c3ad4a490109b9a2f39c/internal/internal_event_handlers.go#L1298

And that will cause issues with Prometheus, as shown in this example.

```go
package main

import (
	"fmt"
	"time"

	"github.com/prometheus/client_golang/prometheus"
	"github.com/uber-go/tally/v4"
	tallyprom "github.com/uber-go/tally/v4/prometheus"

	"go.temporal.io/sdk/client"
	sdktally "go.temporal.io/sdk/contrib/tally"
)

func main() {
	metricsHandler := setupPrometheus()
	// These lines in any order will cause an error where the metrics have different labels
	metricsHandler.Counter("testing").Inc(1)
	metricsHandler.WithTags(map[string]string{"someName": "someValue"}).Counter("testing").Inc(1)
}

// Just setup stuff for Prometheus below here
var PrometheusSanitizeOptions = tally.SanitizeOptions{
	NameCharacters:       tally.ValidCharacters{Ranges: tally.AlphanumericRange, Characters: []rune{'_'}},
[...]

PR #1516, opened 2024-06-18
Add the missing failure_reason tag.

wuhao5:
## What was changed
When omitting the metric "WorkflowTaskExecutionFailureCounter", there are two other occasions that don't specify
the additional "failure_reason" tag. See [here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_pollers.go#L421) that omits the same metric with additional tag.

## Why?
When using [prometheus](https://prometheus.io/) as a client backend, it enforces the same metric be consistent with the
same tags. If not, it will generate an error, and no metric is reported.

## Checklist

1. Closes: simple change, no issue is created.

2. How was this tested:
Implementing a new `metrics.Handler` that is backed by Prometheus, executing an intentional failing activity or workflow, no metrics for the errors are reported.

3. Any docs updates needed?
N/A.

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -54,6 +54,7 @@ import (
 const (
 	queryResultSizeLimit             = 2000000 // 2MB
 	changeVersionSearchAttrSizeLimit = 2048
+	failureReasonPanic               = "Panic"
 )
 
 // Assert that structs do indeed implement the interfaces
@@ -1105,7 +1106,7 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	}
 	defer func() {
 		if p := recover(); p != nil {
-			weh.metricsHandler.Counter(metrics.WorkflowTaskExecutionFailureCounter).Inc(1)
+			weh.metricsHandler.WithTags(metrics.WorkflowTaskFailedTags(failureReasonPanic)).Counter(metrics.WorkflowTaskExecutionFailureCounter).Inc(1)
 			topLine := fmt.Sprintf("process event for %s [panic]:", weh.workflowInfo.TaskQueueName)
 			st := getStackTraceRaw(topLine, 7, 0)
 			weh.Complete(nil, newWorkflowPanicError(p, st))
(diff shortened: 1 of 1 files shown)


internal/internal_event_handlers.go
```diff
@@ -54,6 +54,7 @@ import (
 const (
 	queryResultSizeLimit             = 2000000 // 2MB
 	changeVersionSearchAttrSizeLimit = 2048
+	failureReasonPanic               = "Panic"
```
cretz, line 57: I think, at least so far, we want only `WorkflowError` and `NonDeterminismError` to exist

internal/internal_event_handlers.go
```diff
@@ -1105,7 +1106,7 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	}
 	defer func() {
 		if p := recover(); p != nil {
-			weh.metricsHandler.Counter(metrics.WorkflowTaskExecutionFailureCounter).Inc(1)
```
cretz, line 1108: I am a bit surprised this is/was here. @Quinn-With-Two-Ns - are we double-counting this metric for task failures (I see we also increment it at task completion time)?

internal/internal_event_handlers.go
wuhao5: Maybe the right way is to rename the metric name?

internal/internal_event_handlers.go
Quinn-With-Two-Ns: Yes that is called out here https://github.com/temporalio/sdk-go/issues/1450 the fix is to only record this metric in one place. We need to verify in what conditions the current two locations emit a metrics, if there is any duplication or scenarios that are not covered, what other SDKs do, and align Go SDK with other SDKs and only emit the metric in one place.

internal/internal_event_handlers.go
Quinn-With-Two-Ns: The only two allowed names are
```
		failureReason := "WorkflowError"
		if failWorkflowTask.Cause == enumspb.WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR {
			failureReason = "NonDeterminismError"
		}
``

But, the fix is not to add a tag here but consolidate these metrics into a single location

internal/internal_event_handlers.go
wuhao5: Feel free to toss this PR. Or let me know what you would like me to proceed.

(I am not familiar with other SDKs)

Quinn-With-Two-Ns: Yeah it may be difficult to proceed without knowledge of other SDKs and how the Go SDK processes workflow tasks

wuhao5: SG, do you want me to close this PR then?

Quinn-With-Two-Ns: yes please

Issue #1479, opened 2024-05-18
Wrap GRPC::CANCELED and DEADLINE_EXCEEDED in an SDK Timeout exception for Update


drewhoskins-temporal:
So that users only have to catch one predictable exception

PR #1524, opened 2024-06-20
Wrap GRPC::CANCELED and DEADLINE_EXCEEDED in an SDK Timeout exception…

Quinn-With-Two-Ns:
Based on https://github.com/temporalio/features/issues/483#issuecomment-2150832036

Closes https://github.com/temporalio/sdk-go/issues/1479

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -453,6 +453,12 @@ type (
 	// WARNING: Worker versioning is currently experimental.
 	WorkerVersioningRules = internal.WorkerVersioningRules
 
+	// WorkflowUpdateRPCTimeoutOrCancelledError is an error that occurs when an update RPC call times out or is cancelled.
+	//
+	// Note, this is not related to any general concept of timing out or cancelling a running update, this is only related to the client call itself.
+	// NOTE: Experimental
+	WorkflowUpdateRPCTimeoutOrCancelledError = internal.WorkflowUpdateRPCTimeoutOrCancelledError
+
 	// Client is the client for starting and getting information about a workflow executions as well as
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -453,6 +453,12 @@ type (
 	// WARNING: Worker versioning is currently experimental.
 	WorkerVersioningRules = internal.WorkerVersioningRules
 
+	// WorkflowUpdateRPCTimeoutOrCancelledError is an error that occurs when an update RPC call times out or is cancelled.
+	//
+	// Note, this is not related to any general concept of timing out or cancelling a running update, this is only related to the client call itself.
+	// NOTE: Experimental
+	WorkflowUpdateRPCTimeoutOrCancelledError = internal.WorkflowUpdateRPCTimeoutOrCancelledError
```
cretz, line 460: I used `RPC` because it's `RPCError` in Python to differentiate between RPC and non-RPC errors (e.g. if we have an actual client-side update timeout error which is different than this one). But Go uses the `serviceerror` package to differentiate RPC errors. So if you want, you could change this to `WorkflowUpdateServiceTimeoutOrCancelledError` or something like that, unsure. But I'm fine with this too.

cretz (approved):

client/client.go
cretz, line 460: If you would like, you can only use one `l`, e.g. `WorkflowUpdateRPCTimeoutOrCanceledError`, since that's what we do at https://pkg.go.dev/go.temporal.io/api/serviceerror#Canceled. Different languages do different things there.

PR #1530, opened 2024-06-25
Add workflow safe Mutex and Semaphore

Quinn-With-Two-Ns:
Add workflow safe Mutex and Semaphore. Based on Go standard [Mutex](https://pkg.go.dev/sync) and [Semaphore](https://pkg.go.dev/golang.org/x/sync/semaphore). Notable difference for the `Mutex` is our mutex supports cancellation of a lock request. I did this to match `Semaphore` and most other Go SDK primitives.

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -90,7 +90,8 @@ type (
 
 	// Implements Semaphore interface
 	semaphoreImpl struct {
-		count int64
+		size int64
+		cur  int64
 	}
 
 	// Dispatcher is a container of a set of coroutines.
@@ -1788,26 +1789,27 @@ func (m *mutexImpl) IsLocked() bool {
 
 func (s *semaphoreImpl) Acquire(ctx Context, n int64) error {
 	err := Await(ctx, func() bool {
-		return s.count >= n
+		return s.size-s.cur >= n
 	})
 	if err != nil {
 		return err
 	}
-	s.count -= n
+	s.cur += n
 	return nil
 }
 
 func (s *semaphoreImpl) TryAcquire(ctx Context, n int64) bool {
-	if s.count < n {
-		return false
+	assertNotInReadOnlyState(ctx)
+	success := s.size-s.cur >= n
+	if success {
+		s.cur += n
 	}
-	s.count -= n
-	return true
+	return success
 }
 
 func (s *semaphoreImpl) Release(n int64) {
-	s.count += n
-	if s.count > 0 {
+	s.cur -= n
+	if s.cur < 0 {
 		panic("Mutex.Release() released more than held")
 	}
 }
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -83,6 +83,16 @@ type (
 		settable Settable // used to unblock the future when all coroutines have completed
 	}
 
+	// Implements Mutex interface
+	mutexImpl struct {
+		locked bool
+	}
+
+	// Implements Semaphore interface
+	semaphoreImpl struct {
+		count int64
+	}
+
 	// Dispatcher is a container of a set of coroutines.
 	dispatcher interface {
 		// ExecuteUntilAllBlocked executes coroutines one by one in deterministic order
@@ -278,6 +288,8 @@ var _ Channel = (*channelImpl)(nil)
 var _ Selector = (*selectorImpl)(nil)
 var _ WaitGroup = (*waitGroupImpl)(nil)
 var _ dispatcher = (*dispatcherImpl)(nil)
+var _ Mutex = (*mutexImpl)(nil)
+var _ Semaphore = (*semaphoreImpl)(nil)
 
 // 1MB buffer to fit combined stack trace of all active goroutines
 var stackBuf [1024 * 1024]byte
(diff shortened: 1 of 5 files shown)


internal/internal_workflow.go
```diff
@@ -1742,3 +1754,56 @@ func (us updateSchedulerImpl) Spawn(ctx Context, name string, highPriority bool,
...
+func (s *semaphoreImpl) TryAcquire(ctx Context, n int64) bool {
+	if s.count < n {
+		return false
+	}
+	s.count -= n
+	return true
+}
+
+func (s *semaphoreImpl) Release(n int64) {
+	s.count += n
```
cretz, line 1808: Arguably could have a panic here if count exceeds the initial number given at construction

internal/internal_workflow.go
```diff
@@ -1742,3 +1754,56 @@ func (us updateSchedulerImpl) Spawn(ctx Context, name string, highPriority bool,
...
+}
+
+func (m *mutexImpl) Unlock() {
+	if !m.locked {
+		panic("Mutex.Unlock() was called on an unlocked mutex")
+	}
+	m.locked = false
+}
+
+func (m *mutexImpl) IsLocked() bool {
```
cretz, line 1784: Would be ok if you didn't expose this, but up to you

internal/internal_workflow.go
```diff
@@ -1742,3 +1754,56 @@ func (us updateSchedulerImpl) Spawn(ctx Context, name string, highPriority bool,
...
+		return !m.locked
+	})
+	if err != nil {
+		return err
+	}
+	m.locked = true
+	return nil
+}
+
+func (m *mutexImpl) TryLock(ctx Context) bool {
```
cretz, line 1769: Arguably don't need a context here, but I don't mind it, so long as it's understood that calling `Lock` with a cancelled context is different than calling `TryLock` with a cancelled context.

cretz (approved):
LGTM. You are lucky that only one awaiter is ever awoken at a time, heh.

internal/internal_workflow.go
Quinn-With-Two-Ns: Let me double check what Go's semaphore does, I don't think it cares

internal/internal_workflow.go
Quinn-With-Two-Ns: I exposed it because I wanted a condition a user could check in an `Await` or update validator

internal/internal_workflow.go
Quinn-With-Two-Ns: Ah I added a context here so I could add a check this wasn't being called in an read only context.

internal/internal_workflow.go
Quinn-With-Two-Ns: AH yes it should panic

Quinn-With-Two-Ns: Added a panic here

internal/internal_workflow.go
```diff
@@ -1742,3 +1755,61 @@ func (us updateSchedulerImpl) Spawn(ctx Context, name string, highPriority bool,
...
+	if m.locked {
+		return false
+	}
+	m.locked = true
+	return true
+}
+
+func (m *mutexImpl) Unlock() {
+	if !m.locked {
+		panic("Mutex.Unlock() was called on an unlocked mutex")
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yes, this is programming error, same as accessing a null value or access an array out of bounds which all panic and is consistent with how Go's native mutext behaves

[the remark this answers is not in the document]
Sushisource: Sure, of course that's true, I'm just thinking maybe we want to afford some different control for our alternatives given that our environment is different from a normal programming environment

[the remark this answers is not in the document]
Quinn-With-Two-Ns: I think in our environment it is even more important this is a panic vs an error that users could unintentionally handle  incorrectly.

internal/internal_workflow.go
```diff
@@ -1742,3 +1755,61 @@ func (us updateSchedulerImpl) Spawn(ctx Context, name string, highPriority bool,
...
+	if success {
+		s.cur += n
+	}
+	return success
+}
+
+func (s *semaphoreImpl) Release(n int64) {
+	s.cur -= n
+	if s.cur < 0 {
+		panic("Mutex.Release() released more than held")
```
dandavison, line 1813:
```suggestion
		panic("Semaphore.Release() released more than held")
```

internal/internal_workflow.go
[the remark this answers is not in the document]
Sushisource: Sure, that's probably true. Works for me then. I think it's worth at least considering the options here but I agree it sounds like there's not a great case for making them return errors instead.

Sushisource (approved):

test/integration_test.go
```diff
@@ -1562,6 +1562,122 @@ func (ts *IntegrationTestSuite) TestUpdateWorkflowCancelled() {
...
+		WorkflowID:   run.GetID(),
+		RunID:        run.GetRunID(),
+		UpdateName:   "update",
+		WaitForStage: client.WorkflowUpdateStageAccepted,
+		Args:         []interface{}{false},
+	})
+	ts.NoError(err)
+	// Cancel the workflow, this should cancel any update blocking on the mutex
+	ts.NoError(ts.client.CancelWorkflow(ctx, run.GetID(), run.GetRunID()))
+	ts.Error(updateToCancel.Get(ctx, nil))
```
dandavison, line 1622: Is this assertion checking that the blocking wait operation on a mutex is cancellable, or just checking that canceling a workflow cancels pending updates?

(I'm finding the tests a bit hard to follow; I think they would be easier to follow if there were a few smaller tests each testing something narrower.)

Quinn-With-Two-Ns: The former, the later is not a feature. I would leave the test as is since is is trying to test multiple updates at once competing for the  lock.

PR #1533, opened 2024-07-02
Warn on unhandled updates

Quinn-With-Two-Ns:
Warn on unhandled updates and add a function to wait for all update handlers to finish.

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -52,6 +52,12 @@ const (
 	defaultCoroutineExitTimeout = 100 * time.Millisecond
 
 	panicIllegalAccessCoroutineState = "getState: illegal access from outside of workflow context"
+	unhandledUpdateWarningMessage    = "Workflow finished while update handlers are still running. This may have interrupted work that the" +
+		"update handler was doing, and the client that sent the update will receive a 'workflow execution" +
+		"already completed' RPCError instead of the update result. You can wait for all update" +
+		"handlers to complete by using `await workflow.Await(ctx, () -> workflow.AllHandlersFinished())`. Alternatively, if both you and the clients sending the update" +
+		"are okay with interrupting running handlers when the workflow finishes, and causing clients to" +
+		"receive errors, then you can disable this warning via UnfinishedPolicy in UpdateHandlerOptions."
 )
 
 type (
@@ -216,7 +222,9 @@ type (
 		signalChannels           map[string]Channel
 		queryHandlers            map[string]*queryHandler
 		updateHandlers           map[string]*updateHandler
-		VersioningIntent         VersioningIntent
+		// runningUpdatesHandles is a map of update handlers that are currently running.
+		runningUpdatesHandles map[string]UpdateInfo
+		VersioningIntent      VersioningIntent
 	}
 
 	// ExecuteWorkflowParams parameters of the workflow invocation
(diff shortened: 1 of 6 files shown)


internal/internal_workflow.go
```diff
@@ -52,6 +52,12 @@ const (
 	defaultCoroutineExitTimeout = 100 * time.Millisecond
 
 	panicIllegalAccessCoroutineState = "getState: illegal access from outside of workflow context"
+	unhandledUpdateWarningMessage    = "Workflow finished while update handlers are still running. This may have interrupted work that the" +
```
cretz, line 55: Make sure that there is a space at the beginning/end of each concatenated string here or you'll get `theupdate`

internal/internal_workflow.go
```diff
@@ -661,9 +667,27 @@ func executeDispatcher(ctx Context, dispatcher dispatcher, timeout time.Duration
 		return
 	}
 
-	us := getWorkflowEnvOptions(ctx).getUnhandledSignalNames()
+	weo := getWorkflowEnvOptions(ctx)
+	us := weo.getUnhandledSignalNames()
 	if len(us) > 0 {
-		env.GetLogger().Info("Workflow has unhandled signals", "SignalNames", us)
+		env.GetLogger().Warn("Workflow has unhandled signals", "SignalNames", us)
```
cretz, line 673: Hrmm. We are warning in other SDKs on signals _with handlers_ whose handlers don't complete, but we are not warning on just any signal coming in that has no handler. Arguably to be more effective here we'd differentiate signals that users don't care about from ones they do, i.e. keeping track of all signal channels requested, and only warn on the latter. But I don't mind this change too, but it'll warn on signals the workflow doesn't care about unlike other SDKs. Maybe that's ok.

workflow/workflow.go
```diff
@@ -692,3 +707,12 @@ func DeterministicKeys[K constraints.Ordered, V any](m map[K]V) []K {
 func DeterministicKeysFunc[K comparable, V any](m map[K]V, cmp func(K, K) int) []K {
 	return internal.DeterministicKeysFunc(m, cmp)
 }
+
+// AllHandlersFinished returns true if all update handlers have finished execution.
+// Consider waiting on this condition before workflow return or continue-as-new, to prevent
+// interruption of in-progress handlers by workflow exit:
+//
+//	workflow.Await(ctx, () -> workflow.AllHandlersFinished(ctx))
```
cretz, line 715:
```suggestion
//	workflow.Await(ctx, func() bool { return workflow.AllHandlersFinished(ctx) })
```

A bit more accurate, but doesn't matter much. May also want to update the warning message that has `await` statements and such which aren't a thing in Go.

cretz (commented):
LGTM, only minor things. May want @dandavison to look as well.

internal/internal_workflow.go
Quinn-With-Two-Ns: I think it is hard to assume a users intent here.  Just  because they didn't request a signal channel doesn't mean they didn't  care about, they may have just have not accessed the channel by the time the workflow completed. We were already logging on a all signal channels so I left it as is.

cretz: :+1: It is a bit different than other SDKs, but that's probably fine. I wonder if we should consider in other SDKs to _also_ warn on signals that the workflow has no handlers.

dandavison (approved):

internal/workflow.go
```diff
@@ -386,6 +399,9 @@ type (
 		// performing side-effects. A panic from this function will be treated
 		// as equivalent to returning an error.
 		Validator interface{}
+		// UnfinishedPolicy is the policy to apply when a workflow exits while
+		// the update handler is still running.
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: this is the update option struct so this does not apply to signals

cretz (approved):

PR #1534, opened 2024-07-03
Add otel Carrier implementation for Temporal headers.

mcbadger88:
## What was changed
Adds a Opentelemetry 'Carrier' implementation for Temporal Headers. This can be used by third party libraries to propagate Opentelemetry data via the Temporal headers. For example to propagate trace ID from the starter to workers.

Similar to implementations for other protocols here https://github.com/open-telemetry/opentelemetry-go/blob/0485de287ec48767a58e9fc2eda30b3dc1836668/propagation/propagation.go#L104 and here  https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/metadata_supplier.go#L21.

## Why?
I have made these changes in order for external libraries to be able to add Opentelemetry trace info to the Temporal Headers. Specifically my use case is a custom temporal interceptor that we have written.

Example usage:

```
import  (
    "go.opentelemetry.io/otel/propagation"
    "go.temporal.io/sdk/interceptor"
)

carrier := temporal_propagator.TemporalHeaderCarrier(interceptor.WorkflowHeader(ctx)

// In order to copy otel TextMap data from context into Temporal Headers
propagation.TextMapPropagator.Inject(ctx, carrier)

// In order to copy the otel TextMap data from Temporal Headers into context
propagation.TextMapPropagator.Extract(ctx, carrier)
```

## Checklist

1. Closes 

2. How was this tested:
Unit test
I was unable to figure out where in the temporal interceptor code the trace ID is copied into the temporal headers, but perhaps that could be updated to use this and then it will be covered by an integration test? Appreciate any pointers on that.

3. Any docs updates needed?
not sure?

### DIFF
--- a/contrib/opentelemetry/temporal_header_carrier.go
+++ b/contrib/opentelemetry/temporal_header_carrier.go
@@ -1,0 +1,46 @@
+package opentelemetry
+
+import (
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/sdk/converter"
+)
+
+// TemporalHeaderCarrier adapts temporal header to satisfy the Carrier interface.
+type TemporalHeaderCarrier map[string]*commonpb.Payload
+
+// Get returns the value associated with the passed key.
+// nolint:fabric_structfunctionordering,fabric_typesordering // https://github.com/anzx/fabric-codestyle/issues/114, https://github.com/anzx/fabric-codestyle/issues/115
+func (h TemporalHeaderCarrier) Get(key string) string {
+	value := ""
+	payload := h[key]
+
+	if err := converter.GetDefaultDataConverter().FromPayload(payload, &value); err != nil {
+		//log.Warn(context.Background(), "error during temporal data conversion from payload", log.Str("error", err.Error()))
+		return ""
+	}
+
+	return value
+}
+
+// Keys lists the keys stored in this carrier.
+// nolint:fabric_structfunctionordering,fabric_typesordering // https://github.com/anzx/fabric-codestyle/issues/114, https://github.com/anzx/fabric-codestyle/issues/115
+func (h TemporalHeaderCarrier) Keys() []string {
+	keys := make([]string, 0, len(h))
+	for k := range h {
+		keys = append(keys, k)
+	}
+
+	return keys
+}
+
+// Set stores the key-value pair.
+// nolint:fabric_structfunctionordering // https://github.com/anzx/fabric-codestyle/issues/114
+func (h TemporalHeaderCarrier) Set(key string, value string) {
+	// Convert value to payload
+	payload, err := converter.GetDefaultDataConverter().ToPayload(value)
+	if err != nil {
+		//log.Warn(context.Background(), "error during temporal data conversion to payload", log.Str("error", err.Error()))
+		return
+	}
+	h[key] = payload
+}
--- a/contrib/opentelemetry/temporal_header_carrier.go
+++ b/contrib/opentelemetry/temporal_header_carrier.go
@@ -9,7 +9,6 @@ import (
 type TemporalHeaderCarrier map[string]*commonpb.Payload
 
 // Get returns the value associated with the passed key.
-// nolint:fabric_structfunctionordering,fabric_typesordering // https://github.com/anzx/fabric-codestyle/issues/114, https://github.com/anzx/fabric-codestyle/issues/115
 func (h TemporalHeaderCarrier) Get(key string) string {
 	value := ""
 	payload := h[key]
@@ -23,7 +22,6 @@ func (h TemporalHeaderCarrier) Get(key string) string {
 }
 
 // Keys lists the keys stored in this carrier.
-// nolint:fabric_structfunctionordering,fabric_typesordering // https://github.com/anzx/fabric-codestyle/issues/114, https://github.com/anzx/fabric-codestyle/issues/115
 func (h TemporalHeaderCarrier) Keys() []string {
 	keys := make([]string, 0, len(h))
 	for k := range h {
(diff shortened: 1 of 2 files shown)


contrib/opentelemetry/temporal_header_carrier.go
```diff
@@ -0,0 +1,43 @@
+package opentelemetry
+
+import (
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/sdk/converter"
+)
+
+// TemporalHeaderCarrier adapts temporal header to satisfy the Carrier interface.
+type TemporalHeaderCarrier map[string]*commonpb.Payload
```
Quinn-With-Two-Ns, line 9: Is this PR just to create this type alias and associate methods?

mcbadger88: apologies I was just about to create an issue for this, yes the PR is to add a Opentelemetry "carrier" implementation, it would be used like this:

```
import  (
    "go.opentelemetry.io/otel/propagation"
    "go.temporal.io/sdk/interceptor"
)

carrier := temporal_propagator.TemporalHeaderCarrier(interceptor.WorkflowHeader(ctx)

// In order to copy otel TextMap data from context into Temporal Headers
propagation.TextMapPropagator.Inject(ctx, carrier)

// In order to copy the otel TextMap data from Temporal Headers into context
propagation.TextMapPropagator.Extract(ctx, carrier)
```

mcbadger88: Carrier definition https://opentelemetry.io/docs/specs/otel/context/api-propagators/#carrier

mcbadger88: It can be added to the SDK but I was unable to find the place where the Trace ID gets copied to the headers, the reason it would be beneficial for the temporal interceptor to use this is that it makes the interceptors compatible with other opentelemetry instrumentations

i.e. if you have temporal interceptors installed, and from within a workflow you make an Otel instrumented grpc request, the trace ID will get propagated all the way through

mcbadger88: would you be able to point me to the interceptor code that deals with copying trace ID into the header? I can try to update it and get the tests working

Quinn-With-Two-Ns: My suggestion was if you think others would find this useful, to open source it as its own library.

mcbadger88: the thing is, I am suggesting for the temporal interceptor to use this carrier, so that the temporal interceptors are compatible with other opentelemetry library instrumentations. Would I be better to raise an issue for this?

It would need to be used here (and some refactoring of the parent function) https://github.com/temporalio/sdk-go/blob/master/contrib/opentelemetry/tracing_interceptor.go#L146. Looks like @jvonfricken and @cretz worked on the original

Quinn-With-Two-Ns: `TextMapPropagator` is already an option you can specify, What additional customization point do you need?

mcbadger88: so this request is about the carrier (not the propagator), the suggestion I have is to replace `[textMapCarrier](https://github.com/temporalio/sdk-go/blob/master/contrib/opentelemetry/tracing_interceptor.go#L146)` as the carrier here is actually the temporal header. We would like to be able to use a common function with the temporal.io otel interceptor, so that we can be guaranteed that any custom interceptors we write will use the same header keys as the temporal.io otel interceptor. 

It also allows for extensions if more data types are required to be sent in the future, they can easily be added to the temporal headers whilst aligning with Open telemetry patterns.

mcbadger88: fwiw I raised an issue here, in case it explains it better https://github.com/temporalio/sdk-go/issues/1535

Quinn-With-Two-Ns: >the suggestion I have is to replace textMapCarrier [link](https://github.com/temporalio/sdk-go/blob/master/contrib/opentelemetry/tracing_interceptor.go#L146) as the carrier here is actually the temporal header

The carrier is just a payload in the header, not the whole header.

>so that we can be guaranteed that any custom interceptors we write will use the same header keys as the temporal.io otel interceptor.

I am not sure how replacing the carrier accomplishes your goal. The `HeaderKey` and `SpanContextKey` are options in `TracerOptions`

mcbadger88: For example if the temporal starter is not owned by us, we cannot specify the Header key that the interceptor uses. If the workers then have our own custom interceptor installed, we cannot read the trace data from that Header key as it is not exposed.

cretz: > For example if the temporal starter is not owned by us

How can you ensure that it is using OTel interceptor at all? If you can ensure the starter will set the client options to use a certain interceptor, surely you can ensure the starter can set certain interceptor options?

emmaCullen: Yes whilst it is not guaranteed that the starter is using the otel interceptor, it does maximises compatibility if all interceptors are able to use the same carrier. If another upstream team (or company!) installs otel they are likely to use your interceptor and not likely to set their own HeaderKey as they dont know what the downstream will use. 

This is about maximising compatibility, if you wanted to expose the HeaderKey that you use that could also be a solution.

cretz: > if you wanted to expose the HeaderKey that you use that could also be a solution.

We intentionally don't expose this because it is so easy to set and we don't want people to build a reliance upon it. This is why we make the option available.

emmaCullen: I understand what you are saying, unfortunately it doesnt help in our case where we do not control the starter service.

Quinn-With-Two-Ns:
Marking as closed as since this does not appear to present compelling value to the Go SDK otel integration

PR #1538, opened 2024-07-05
add Name method to ReceiveChannel and SendChannel interface

### LINKED ISSUE #1304: Add GetName() func to Signal Chanel to return human readable name
While dynamically creating signal channels and assigning callback functions to it, it is not possible to identify the channel name in the callback funciton

adding a GetName() string func to the Chanel interface

bentveljanzx:
## What was changed
- Added `Name` method to `SendChannel` and `ReceiveChannel` interface
- Added tests to assert `Name`returns back the same given name.

## Why?

- We have custom wrappers around signalling receive and send calls so that we can attach OpenTelemetry to our signal payload. Since we want our spans to contain the signal name and happen before calling into the underlying `channel.Send` method, ideally we could simplify our API to not require a redundant sending of the channelName again.
  - ie. `telemetrytemporal.ChannelSend[T](ctx workflow.Context, channel workflow.SendChannel, channelName string, payload T)`

## Checklist

Closes #1304

2. How was this tested:
- Added "TestChannelName" in `internal/internal_coroutines_test.go` which creates named channels and asserts that the `Name` method returns back the same name you gave it.

3. Any docs updates needed?
- N/A

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -69,6 +69,10 @@ type (
 	// commonChannel contains methods common to both SendChannel and ReceiveChannel
 	commonChannel interface {
 		// Name returns the name of the channel.
+		// If the channel was retrieved from a GetSignalChannel call, Name returns the signal name.
+		//
+		// A Channel created without an explicit name will use a generated name by the SDK and
+		// is not deterministic.
 		Name() string
 	}
 
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -68,8 +68,8 @@ var (
 type (
 	// commonChannel contains methods common to both SendChannel and ReceiveChannel
 	commonChannel interface {
-		// Name returns the name of the channel.
-		// If the channel was retrieved from a GetSignalChannel call, Name returns the signal name.
+		// Name returns the name of the Channel.
+		// If the Channel was retrieved from a GetSignalChannel call, Name returns the signal name.
 		//
 		// A Channel created without an explicit name will use a generated name by the SDK and
 		// is not deterministic.
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -66,8 +66,16 @@ var (
 )
 
 type (
+	// commonChannel contains methods common to both SendChannel and ReceiveChannel
+	commonChannel interface {
+		// Name returns the name of the channel.
+		Name() string
+	}
+
 	// SendChannel is a write only view of the Channel
 	SendChannel interface {
+		commonChannel
+
 		// Send blocks until the data is sent.
 		Send(ctx Context, v interface{})
 
@@ -80,6 +88,8 @@ type (
 
 	// ReceiveChannel is a read only view of the Channel
 	ReceiveChannel interface {
+		commonChannel
+
 		// Receive blocks until it receives a value, and then assigns the received value to the provided pointer.
 		// Returns false when Channel is closed.
 		// Parameter valuePtr is a pointer to the expected data structure to be received. For example:
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -66,19 +66,14 @@ var (
 )
 
 type (
-	// commonChannel contains methods common to both SendChannel and ReceiveChannel
-	commonChannel interface {
+	// SendChannel is a write only view of the Channel
+	SendChannel interface {
 		// Name returns the name of the Channel.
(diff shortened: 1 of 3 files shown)


internal/workflow.go
```diff
@@ -66,8 +66,16 @@ var (
 )
 
 type (
+	// commonChannel contains methods common to both SendChannel and ReceiveChannel
+	commonChannel interface {
+		// Name returns the name of the channel.
```
Quinn-With-Two-Ns, line 71: Can we add a few things to the docs: 
1. Specify that for a `SignalChannel` , `Name` is the signal name. Some places in Temporal refer to signal "Name" as "Type" so it may not obvious.
2. Specify that channels created without an explicit name have a name generated by the SDK and they should not assume it is deterministic. Normally all workflow APIs are deterministic, but in this case the channel name would not be.

Quinn-With-Two-Ns:
Overall LGTM, just a couple of documentation points.

internal/workflow.go
bentveljanzx: Added clarity to the documentation as requested

internal/workflow.go
```diff
@@ -66,8 +66,20 @@ var (
...
+		// If the Channel was retrieved from a GetSignalChannel call, Name returns the signal name.
+		//
+		// A Channel created without an explicit name will use a generated name by the SDK and
+		// is not deterministic.
+		Name() string
+	}
+
 	// SendChannel is a write only view of the Channel
 	SendChannel interface {
+		commonChannel
```
cretz, line 81: I wonder if this should be on the receive side only (send channel only used if user explicitly chooses to break it up, and they can't get things like `Len()` either even though _technically_ we could provide it).

cretz, line 81: I wonder if, even though we use aliases which make this harder, inlining the `Name()` method with the docs is clearer than a shared interface. It's just two places.

internal/workflow.go
```diff
@@ -66,8 +66,20 @@ var (
 )
 
 type (
+	// commonChannel contains methods common to both SendChannel and ReceiveChannel
+	commonChannel interface {
+		// Name returns the name of the Channel.
+		// If the Channel was retrieved from a GetSignalChannel call, Name returns the signal name.
+		//
+		// A Channel created without an explicit name will use a generated name by the SDK and
+		// is not deterministic.
```
cretz, line 75:
```suggestion
		// A Channel created without an explicit name will use a generated name by the SDK and
		// is deterministic.
```

It is deterministic correct? (built from channel sequence)

Quinn-With-Two-Ns: It is not given our normal definition for what is deterministic, and I would **not** want suers to have any expectation of the SDK generated name

cretz: Our normal definition is "same during replay" I thought. Definitely no expectation of generated name, but it's still deterministic. Maybe we should just say "generated name by the SDK" without any additional info about determinism.

internal/workflow.go
cretz: Just wondering aloud. Can ask the same of `Len`. Go allows you to get `len` and `cap` on both recv-only and send-only channels but we decided `Len()` should only be on receive. I just wonder if we need to apply whatever that thought process is here. I don't mind if not (though I think it'd be worth copying the definition twice vs making a new interface).

bentveljanzx: For our use-case we only use `Name` for the `SendChannel` case, so if anything I'd only want it on the `SendChannel`.

Quinn-With-Two-Ns (approved):

internal/workflow.go
bentveljanzx: Will make the change if there's agreement between @Quinn-With-Two-Ns and yourself :)

One readability benefit of sharing the interface is it'll ensure documentation remains synced over time if it gets modified

Quinn-With-Two-Ns: Yes I think inlining is better for readability

bentveljanzx: Updated and inlined. Also made tests explicitly test `ReceiveChannel` and `SendChannel` interfaces with Name

bentveljanzx:
If there's anything else outstanding / blocking merge, please let me know :)

Quinn-With-Two-Ns:
Thanks, merged!

Issue #1474, opened 2024-05-16
Tracing interceptor support for Nexus Operations


bergundy:
We need to add support for propagating headers from a Workflow to a Nexus Operation handler to the StartOperation request.

In order to do that we'll need to first add an `NexusOperationInboundInterceptor`.

Quinn-With-Two-Ns:
I think we just need to add support for the [TracingInterceptor](https://github.com/temporalio/sdk-go/blob/master/interceptor/tracing_interceptor.go)

PR #1544, opened 2024-07-11
Add utility to get metrics handler for a Nexus operation

pdoerner:
## What was changed

Added a metrics handler to the NexusContext and a utility for extracting this handler from a `context.Context`

## Why?

To be able to record metrics about operations

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalnexus/operation.go
+++ b/temporalnexus/operation.go
@@ -22,10 +22,21 @@ import (
 
 	"github.com/nexus-rpc/sdk-go/nexus"
 	"go.temporal.io/api/common/v1"
+
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/internal"
+	"go.temporal.io/sdk/internal/common/metrics"
 	"go.temporal.io/sdk/workflow"
 )
+
+// GetMetricsHandler returns a metrics handler to be used in a Nexus operation's context.
+func GetMetricsHandler(ctx context.Context) metrics.Handler {
+	nctx, ok := internal.NexusOperationContextFromGoContext(ctx)
+	if !ok {
+		panic("temporalnexus GetMetricsHandler: Not a valid Nexus context")
+	}
+	return nctx.MetricsHandler
+}
 
 type syncOperation[I, O any] struct {
 	nexus.UnimplementedOperation[I, O]
--- a/internal/internal_nexus_task_handler.go
+++ b/internal/internal_nexus_task_handler.go
@@ -14,6 +14,7 @@ import (
 	"go.temporal.io/api/common/v1"
 	nexuspb "go.temporal.io/api/nexus/v1"
 	"go.temporal.io/api/workflowservice/v1"
+
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/internal/common/metrics"
 	"go.temporal.io/sdk/log"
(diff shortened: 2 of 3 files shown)


bergundy (approved):
LGTM but I'd appreciate a look from @Quinn-With-Two-Ns.

Quinn-With-Two-Ns (approved):
Yeah looks fine, I would always encourage adding tests that this is properly wired up

internal/cmd/build/go.mod
```diff
@@ -1,6 +1,8 @@
 module go.temporal.io/sdk/internal/cmd/build
 
-go 1.20
+go 1.21
+
+toolchain go1.22.4
```
Quinn-With-Two-Ns, line 5: We don't want to specify a toolchain since it will override our CI that trys to test the SDK on all the versions of Go we support.

temporalnexus/operation.go
```diff
@@ -22,11 +22,32 @@ import (
...
+func GetMetricsHandler(ctx context.Context) metrics.Handler {
+	nctx, ok := internal.NexusOperationContextFromGoContext(ctx)
+	if !ok {
+		panic("temporalnexus GetMetricsHandler: Not a valid Nexus context")
+	}
+	return nctx.MetricsHandler
+}
+
+// GetLogger returns a logger to be used in a Nexus operation's context.
+func GetLogger(ctx context.Context) log.Logger {
```
Quinn-With-Two-Ns, line 43: @bergundy I forgot, Does go SDK nexus have interceptors yet? The activity equivalents for these calls are interceptable. If it doesn't  support interceptors yet then no worries.

pdoerner: I didn't see any interceptor equivalent for Nexus yet

internal/cmd/build/go.mod
pdoerner: Yeah didn't realize GoLand added this.

temporalnexus/operation.go
```diff
@@ -22,11 +22,32 @@ import (
...
+	"go.temporal.io/sdk/internal/common/metrics"
+	"go.temporal.io/sdk/log"
 	"go.temporal.io/sdk/workflow"
 )
 
+// GetMetricsHandler returns a metrics handler to be used in a Nexus operation's context.
+func GetMetricsHandler(ctx context.Context) metrics.Handler {
+	nctx, ok := internal.NexusOperationContextFromGoContext(ctx)
+	if !ok {
+		panic("temporalnexus GetMetricsHandler: Not a valid Nexus context")
```
Quinn-With-Two-Ns, line 37: I wonder if it would be easier if `NexusOperationContextFromGoContext` did the check for a valid nexus context? That is how the corresponding activity call works. I don't have a very strong opinion here.

pdoerner: I also don't have a strong opinion but I went this route because the other usages of `NexusOperationContextFromGoContext` return an error rather than panic and I didn't want `GetMetricsHandler` or `GetLogger` to return an error to keep them parallel to the activity and workflow versions.

Quinn-With-Two-Ns (approved):

temporalnexus/operation.go
Quinn-With-Two-Ns: Yeah at some point will probably need to add a `IsNexusContext` or something like we have for Activities

temporalnexus/operation.go
bergundy: We have interceptors from the workflow not for the handler yet.

https://github.com/temporalio/sdk-go/issues/1474

test/nexus_test.go
```diff
@@ -312,6 +313,11 @@ func TestSyncOperationFromWorkflow(t *testing.T) {
 	tc := newTestContext(t, ctx)
 
 	op := temporalnexus.NewSyncOperation("op", func(ctx context.Context, c client.Client, outcome string, o nexus.StartOperationOptions) (string, error) {
+		require.NotPanicsf(t, func() {
```
bergundy, line 316: Would rather check the metrics labels and logger tags but this is fine for now. Can open an issue for later.

bergundy (approved):
Could you just post in a comment what logger tags and metrics labels you're seeing and open an issue to test it?

PR #1546, opened 2024-07-12
WorkerTuner & Resource based autotuning

Sushisource:
## What was changed
Refactor slot reservation behind interface, interface exposed experimentally
Added resource based autotuner

## Why?
Parity with ongoing work in all other SDKs

## Checklist

1. Closes 

2. How was this tested:
Added lots of tests, will be stressing in omes.

3. Any docs updates needed?
Public docs to come in public preview


internal/internal_worker.go
```diff
@@ -1827,8 +1839,18 @@ func getReadOnlyChannel(c chan struct{}) <-chan struct{} {
 }
 
 func setWorkerOptionsDefaults(options *WorkerOptions) {
+	if options.Tuner != nil {
+		if options.MaxConcurrentWorkflowTaskExecutionSize != 0 ||
+			options.MaxConcurrentActivityExecutionSize != 0 ||
+			options.MaxConcurrentLocalActivityExecutionSize != 0 {
+			panic("cannot set MaxConcurrentWorkflowTaskExecutionSize, MaxConcurrentActivityExecutionSize, or MaxConcurrentLocalActivityExecutionSize with Tuner")
```
cretz, line 1846: Ug, I hate that worker creation doesn't have normal errors, but understood

internal/internal_worker.go
```diff
@@ -1827,8 +1839,18 @@ func getReadOnlyChannel(c chan struct{}) <-chan struct{} {
...
+		}
+	}
+	maxConcurrentWFT := options.MaxConcurrentWorkflowTaskExecutionSize
+	maxConcurrentAct := options.MaxConcurrentActivityExecutionSize
+	maxConcurrentLA := options.MaxConcurrentLocalActivityExecutionSize
```
cretz, line 1851: Can probably just mutate the options still at this point since I don't expect these three fields are ever even referenced anymore after this function returns, but no big deal of course

internal/internal_worker_base.go
```diff
@@ -332,61 +347,100 @@ func (bw *baseWorker) runPoller() {
...
 	for {
+		bw.stopWG.Add(1)
+		go func() {
+			defer bw.stopWG.Done()
+			s, err := bw.slotSupplier.ReserveSlot(ctx, &bw.options.slotReservationData)
+			if err != nil {
+				if !errors.Is(err, context.Canceled) {
+					bw.logger.Error(fmt.Sprintf("Error while trying to reserve slot: %v", err))
+				} else {
+					close(reserveChan)
```
cretz, line 362: It's rare to ever need to close channels in Go except as a broadcast mechanism. It can lead to confusing behavior. In this case the next send attempt in this loop will panic and every receive from here on out will immediately succeed with zero value

Did you maybe mean to create the channel _inside_ the loop? Even if, would recommend sending `nil` to it instead of erroring.

internal/internal_worker_base.go
```diff
@@ -332,61 +347,100 @@ func (bw *baseWorker) runPoller() {
...
 		select {
 		case <-bw.stopCh:
+			cancelfn()
 			return
-		case <-bw.pollerRequestCh:
+		case permit := <-reserveChan:
+			if permit == nil { // There was an error reserving a slot
+				// Avoid spamming reserve hard in the event it's constantly failing
+				time.Sleep(time.Second)
+				continue
```
cretz, line 381: IIUC the channel can never be used again at this point correct, it is closed. Which means this select call will _always_ resolve immediately with a nil value. So why `continue`?

internal/internal_worker_base.go
```diff
@@ -332,61 +347,100 @@ func (bw *baseWorker) runPoller() {
...
+			select {
+			case reserveChan <- s:
+			case <-ctx.Done():
+				bw.releaseSlot(s, "worker shutting down")
+			}
+		}()
+
 		select {
 		case <-bw.stopCh:
+			cancelfn()
```
cretz, line 375: It is a tad safer and clearer to put this as `defer cancelfn()` right below where it is created (obviously doesn't matter that much in here, but is safer for future code edits that may come through)

internal/internal_worker_base.go
```diff
@@ -332,61 +347,100 @@ func (bw *baseWorker) runPoller() {
...
+		permit := eagerOrPolled.getPermit()
+
+		if !task.isEmpty() {
+			bw.slotSupplier.MarkSlotUsed(permit)
+		}
+
+		defer func() {
+			bw.releaseSlot(permit, "task processed")
+
+			if p := recover(); p != nil {
```
cretz, line 422: Hrmm. Did you see a situation where a panic happened, or is this a result of some other refactoring?

cretz (commented):
Admittedly didn't do a deep dive comparing against existing/changed semaphore/permit logic because I found it a bit hard to follow and I haven't been in this code a while. Regardless, it does take a solid amount of bravery to work in these depths, would welcome running as many possible tests as we can muster.

internal/internal_worker_base.go
Sushisource: It's from inlining the `processTask` function that was only ever called from here

internal/tuning.go
```diff
@@ -0,0 +1,266 @@
...
+	GetActivityTaskSlotSupplier() SlotSupplier
+	GetLocalActivitySlotSupplier() SlotSupplier
+}
+
+// SlotPermit is a permit to use a slot.
+//
+// WARNING: Custom implementations of SlotSupplier are currently experimental.
+type SlotPermit struct {
+	//lint:ignore U1000 pointless to guarantee uniqueness for now
+	int
```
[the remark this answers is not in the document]
Sushisource: Otherwise pointers to SlotPermits are not unique

contrib/resourcetuner/resourcetuner.go
```diff
@@ -0,0 +1,345 @@
...
+	return &ResourceBasedSlotSupplier{controller: controller, options: options}
+}
+
+func (r *ResourceBasedSlotSupplier) ReserveSlot(ctx context.Context, reserveCtx worker.SlotReserveContext) (*worker.SlotPermit, error) {
+	for {
+		if reserveCtx.NumIssuedSlots() < r.options.MinSlots {
+			return &worker.SlotPermit{}, nil
+		}
+		r.lastIssuedMu.Lock()
+		mustWaitFor := r.options.RampThrottle - time.Since(r.lastSlotIssuedAt)
```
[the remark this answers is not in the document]
Sushisource: Yeah, it's caught by that (as well as other weird things, some of which seem windows specific). Will add a comment.

contrib/resourcetuner/resourcetuner.go
```diff
@@ -0,0 +1,345 @@
...
+				return nil, ctx.Err()
+			}
+		}
+
+		maybePermit := r.TryReserveSlot(reserveCtx)
+		if maybePermit != nil {
+			return maybePermit, nil
+		}
+		select {
+		case <-time.After(10 * time.Millisecond):
```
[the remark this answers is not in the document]
Sushisource: Good point, I'll just make it a sleep.

contrib/resourcetuner/resourcetuner.go
```diff
@@ -0,0 +1,345 @@
...
+	options ResourceBasedSlotSupplierOptions,
+) *ResourceBasedSlotSupplier {
+	return &ResourceBasedSlotSupplier{controller: controller, options: options}
+}
+
+func (r *ResourceBasedSlotSupplier) ReserveSlot(ctx context.Context, reserveCtx worker.SlotReserveContext) (*worker.SlotPermit, error) {
+	for {
+		if reserveCtx.NumIssuedSlots() < r.options.MinSlots {
+			return &worker.SlotPermit{}, nil
+		}
```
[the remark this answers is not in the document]
Sushisource: Yes, good catch, it'd make sense to put a lock around this bit to avoid bursting from 0 too fast.

contrib/resourcetuner/resourcetuner.go
```diff
@@ -0,0 +1,345 @@
...
+// NewResourceBasedSlotSupplier creates a ResourceBasedSlotSupplier given the provided
+// ResourceController and ResourceBasedSlotSupplierOptions. All ResourceBasedSlotSupplier instances
+// must use the same ResourceController.
+//
+// WARNING: Resource based tuning is currently experimental.
+func NewResourceBasedSlotSupplier(
+	controller *ResourceController,
+	options ResourceBasedSlotSupplierOptions,
+) *ResourceBasedSlotSupplier {
+	return &ResourceBasedSlotSupplier{controller: controller, options: options}
```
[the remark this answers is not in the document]
Sushisource: I'll add some checks for <= 0 too

worker/tuning.go
```diff
@@ -0,0 +1,61 @@
...
+// Currently, you cannot implement your own slot supplier. You can use the provided
+// FixedSizeSlotSupplier and ResourceBasedSlotSupplier slot suppliers.
+type SlotSupplier = internal.SlotSupplier
+
+// SlotReserveContext contains information that SlotSupplier instances can use during
+// reservation calls.
+type SlotReserveContext = internal.SlotReserveContext
+
+// CreateFixedSizeTuner creates a WorkerTuner that uses fixed size slot suppliers.
+func CreateFixedSizeTuner(numWorkflowSlots, numActivitySlots, numLocalActivitySlots int) WorkerTuner {
```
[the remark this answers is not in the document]
Sushisource: I don't expect this one would ever change, but, may as well use them everywhere sure.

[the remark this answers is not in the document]
Sushisource: I had a note to change them to New and spaced it. Will do. I'll add the errors too.

internal/tuning.go
```diff
@@ -0,0 +1,266 @@
...
+// FixedSizeSlotSupplier and ResourceBasedSlotSupplier slot suppliers.
+//
+// WARNING: Custom implementations of SlotSupplier are currently experimental.
+type SlotSupplier interface {
+	// ReserveSlot is called before polling for new tasks. The implementation should block until
+	// a slot is available, then return a permit to use that slot. Implementations must be
+	// thread-safe.
+	//
+	// Any returned error besides context.Canceled will be logged and the function will be retried.
+	ReserveSlot(ctx context.Context, reserveCtx SlotReserveContext) (*SlotPermit, error)
```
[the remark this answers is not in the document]
Sushisource: Yeah, my concern was implying that they could be cancelled. I figure if they want to pass information, they can just do it via whatever struct is implementing this interface.

internal/internal_task_pollers.go
```diff
@@ -346,7 +350,13 @@ func (wtp *workflowTaskPoller) processWorkflowTask(task *workflowTask) error {
 		return err
 	}
 	var taskErr error
-	defer func() { wfctx.Unlock(taskErr) }()
+	defer func() {
+		// TODO: I had to add this check to avoid a panic in unit tests where this is nil.
+		//    Seems harmless to keep this, but, wanted to get another opinion.
```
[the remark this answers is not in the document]
Sushisource: It intentionally returns nil. I tried to fix up the fake test stuff to not do that, but it ends up requiring you to fake out a whole boatload of other stuff.

internal/internal_worker_base.go
Sushisource: I'll just send an explicit nil. In any case this is one of the things I needed to add a test for still so will do that for sure.

Sushisource: Ah, so, there was a good reason for this. The explicit send of nil will block forever since, very likely, the other side waiting on it has already exited b/c of the stop channel. In fact, nothing needs to be done at all in this branch. In the branch with the logging, sending nil works.

internal/tuning.go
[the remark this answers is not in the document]
cretz: Ah, that's right. Can you make this a non-exported field then instead of embedded so this implementation detail isn't visible to users?

internal/tuning.go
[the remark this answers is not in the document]
cretz: The issue is contextual state per reservation. I don't think people see Go context as meaning async. In fact, since reserve context is rarely used, would totally support just a Go context and a method to retrieve the reserve context from it.

Also, arguably `SlotPermit` should be an (empty) interface so that people can pass back their own objects that are sent back on mark used and release. People might want to attach a state to a slot (even if you don't do that, should at least provide the slot permit to mark used and release so they can keep a map).

internal/internal_task_pollers.go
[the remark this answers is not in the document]
cretz: But to confirm, this is only tests right? This never returns nil at runtime? No problem of course, just want to confirm that code-writer expectations of this existing still hold.

internal/internal_worker_base.go
cretz: Do sending in a select that is bounded by context close as well

internal/internal_eager.go
```diff
@@ -26,9 +26,10 @@ package internal
 type eagerWorker interface {
 	// tryReserveSlot tries to reserver a task slot on the worker without blocking
 	// caller is expected to release the slot with releaseSlot
-	tryReserveSlot() bool
+	tryReserveSlot() *SlotPermit
 	// releaseSlot release a task slot acquired by tryReserveSlot
-	releaseSlot()
+	// TODO: Make reason typed
```
Quinn-With-Two-Ns, line 31: Is this TODO mean to be resolved in this PR?

worker/tuning.go
```diff
@@ -0,0 +1,61 @@
...
+	return internal.CreateFixedSizeTuner(numWorkflowSlots, numActivitySlots, numLocalActivitySlots)
+}
+
+// CreateCompositeTuner creates a WorkerTuner that uses a combination of slot suppliers.
+func CreateCompositeTuner(workflowSlotSupplier, activitySlotSupplier, localActivitySlotSupplier SlotSupplier) WorkerTuner {
```
Quinn-With-Two-Ns, line 54: I think all of these constructors should take options.

internal/internal_nexus_worker.go
```diff
@@ -61,19 +62,24 @@ func newNexusWorker(opts nexusWorkerOptions) (*nexusWorker, error) {
 		params,
 	)
 
+	workerType := "NexusWorker"
+	metricsHandler := params.MetricsHandler.WithTags(metrics.WorkerTags(workerType))
+	tss := newTrackingSlotSupplier(NewFixedSizeSlotSupplier(opts.executionParameters.ConcurrentNexusTaskExecutionSize), metricsHandler)
```
Quinn-With-Two-Ns, line 67: Is there a reason we didn't support Nexus.

internal/tuning.go
```diff
@@ -0,0 +1,266 @@
...
+		issuedSlots: int(t.issuedSlotsAtomic.Load()),
+	})
+	if permit != nil {
+		t.issuedSlotsAtomic.Add(1)
+		t.publishMetrics(false)
+	}
+	return permit
+}
+func (t *trackingSlotSupplier) MarkSlotUsed(permit *SlotPermit) {
+	if permit == nil {
```
Quinn-With-Two-Ns, line 235: Why is a `nil` permit ever being used to reserver a slot?

internal/tuning.go
```diff
@@ -0,0 +1,266 @@
...
+	// thread-safe.
+	//
+	// Any returned error besides context.Canceled will be logged and the function will be retried.
+	ReserveSlot(ctx context.Context, reserveCtx SlotReserveContext) (*SlotPermit, error)
+	// TryReserveSlot is called when attempting to reserve slots for eager workflows and activities.
+	// It should return a permit if a slot is available, and nil otherwise. Implementations must be
+	// thread-safe.
+	TryReserveSlot(reserveCtx SlotReserveContext) *SlotPermit
+
+	MarkSlotUsed()
```
Quinn-With-Two-Ns, line 79: How come these APIs don't take a slot context?

internal/internal_nexus_worker.go
Sushisource: Yeah, I wanted to have at least some discussion about whether or not we want to add it to the tuning interface. Probably the answer is yes but it's worth discussing. If we decide yes it's easy enough to add subsequently.

internal/internal_eager.go
Sushisource: It will be. Will include it as part of adding the remaining contexts per your other comment.

internal/tuning.go
Sushisource: I'll add them. I was hoping I wouldn't have to fully pub the API like I did, but now that it is I'll add them, even if they're gonna be mostly empty.

internal/tuning.go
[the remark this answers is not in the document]
Sushisource: Retrieving it from the Go context just feels icky to me, how it introduces a number of possible error paths.

I'd be fine with sticking the Go context inside this context though, and since all the other methods will get their own context then they'll have that.

Attaching state to the permit was done in Java by just having a field inside `SlotPermit`, which I also prefer since it allows us to be clear about what the type of a permit is instead of it just being `any`. I just didn't add it since it's unclear if anyone even wants that at the moment, but, it's trivial to add to this too.

internal/internal_task_pollers.go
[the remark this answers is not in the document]
Sushisource: Yeah, it never returns nil under non-test circumstances

internal/tuning.go
Sushisource: Also changed to panic.

internal/internal_worker.go
Sushisource: That actually causes problems since sometimes the tests will end up passing these around and this function gets run more than once and then the checks that you haven't set both maxConcurrentXXX and Tuner fail.

internal/internal_eager_activity.go
```diff
@@ -34,9 +33,7 @@ import (
 type eagerActivityExecutor struct {
 	eagerActivityExecutorOptions
 
-	activityWorker eagerWorker
-	heldSlotCount  int
-	countLock      sync.Mutex
+	activityWorker *baseWorker
```
Quinn-With-Two-Ns, line 36: Can we avoid giving `eagerActivityExecutor` the full `activityWorker `? I am not sure why it is needed here, but not in the workflow case.

internal/internal_eager_activity.go
```diff
@@ -76,75 +74,62 @@ func (e *eagerActivityExecutor) applyToRequest(
...
 
-func (e *eagerActivityExecutor) reserveOnePendingSlot() bool {
-	// Lock during count checks. Nothing in here blocks including the channel
-	// receive to serve a slot.
-	e.countLock.Lock()
-	defer e.countLock.Unlock()
-	// Confirm that, if we have a max, held count isn't already there
-	if e.maxConcurrent > 0 && e.heldSlotCount >= e.maxConcurrent {
+func (e *eagerActivityExecutor) reserveOnePendingSlot() *SlotPermit {
+	// Confirm that, if we have a max, issued count isn't already there
```
Quinn-With-Two-Ns, line 92: Why was the previous logic not sufficient? It already handled the case where `maxConcurrent` could be unset.

internal/internal_worker.go
```diff
@@ -265,6 +256,15 @@ func ensureRequiredParams(params *workerExecutionParameters) {
 	if params.FailureConverter == nil {
 		params.FailureConverter = GetDefaultFailureConverter()
 	}
+	if params.Tuner == nil {
+		// Err cannot happen since these slot numbers are guaranteed valid
+		params.Tuner, _ = NewFixedSizeTuner(
+			FixedSizeTunerOptions{
```
Quinn-With-Two-Ns, line 262: Shouldn't nexus slots be here?

worker/tuning.go
```diff
@@ -0,0 +1,75 @@
...
+type SlotReleaseContext = internal.SlotReleaseContext
+
+// FixedSizeTunerOptions are the options used by NewFixedSizeTuner.
+type FixedSizeTunerOptions = internal.FixedSizeTunerOptions
+
+// CompositeTunerOptions are the options used by NewCompositeTuner.
+type CompositeTunerOptions = internal.CompositeTunerOptions
+
+// NewFixedSizeTuner creates a WorkerTuner that uses fixed size slot suppliers.
+func NewFixedSizeTuner(opts FixedSizeTunerOptions) (WorkerTuner, error) {
```
Quinn-With-Two-Ns, line 63: Nit:
```suggestion
func NewFixedSizeTuner(options FixedSizeTunerOptions) (WorkerTuner, error) {
```
Go SDK uses `options` more often then not.

worker/tuning.go
```diff
@@ -0,0 +1,75 @@
...
+// THE SOFTWARE.
+
+package worker
+
+import (
+	"go.temporal.io/sdk/internal"
+)
+
+// WorkerTuner allows for the dynamic customization of some aspects of worker behavior.
+type WorkerTuner = internal.WorkerTuner
```
Quinn-With-Two-Ns, line 32: These should all be marked experimental right?

internal/internal_task_pollers.go
```diff
@@ -346,7 +350,13 @@ func (wtp *workflowTaskPoller) processWorkflowTask(task *workflowTask) error {
 		return err
 	}
 	var taskErr error
-	defer func() { wfctx.Unlock(taskErr) }()
+	defer func() {
+		// This check exists to avoid a panic in unit tests where this is nil. In non-mocked
```
Quinn-With-Two-Ns, line 354: Can we just make sure it is not nil in unit tests as well or is it really hard? I don't like adding checks like this in production code just to make tests a bit simpler

internal/internal_worker.go
```diff
@@ -457,22 +470,28 @@ func newActivityTaskWorker(taskHandler ActivityTaskHandler, service workflowserv
...
 			fatalErrCb:        workerParams.WorkerFatalErrorCallback,
 			userContextCancel: workerParams.UserContextCancel,
+			metricsHandler:    metricsHandler,
+			slotReservationData: slotReservationData{
+				taskQueue: workerParams.TaskQueue,
+				logger:    workerParams.Logger,
+			},
 		},
 		workerParams.Logger,
-		workerParams.MetricsHandler,
```
Quinn-With-Two-Ns, line 475: Why move just the `MetricsHandler` into the options? TBH I am not sure why `newBaseWorker` takes an options struct and parameters?

worker/tuning.go
```diff
@@ -0,0 +1,75 @@
...
+
+// SlotPermit is a permit to use a slot.
+type SlotPermit = internal.SlotPermit
+
+// SlotSupplier controls how slots are handed out for workflow and activity tasks as well as
+// local activities when used in conjunction with a WorkerTuner.
+//
+// Currently, you cannot implement your own slot supplier. You can use the provided
```
Quinn-With-Two-Ns, line 40: Is this something we are enforcing?  Sorry maybe I just missed the check?

Quinn-With-Two-Ns: Also why are we enforcing this? I don't think we did in Java?

internal/tuning.go
```diff
@@ -0,0 +1,395 @@
...
+// WARNING: Custom implementations of SlotSupplier are currently experimental.
+type SlotPermit struct {
+	//lint:ignore U1000 pointless to guarantee pointers to SlotPermits are unique
+	_uniqueInt int
+}
+
+// SlotReserveContext contains information that SlotSupplier instances can use during
+// reservation calls. It embeds a standard Context.
+type SlotReserveContext interface {
+	context.Context
```
Quinn-With-Two-Ns, line 58: We should not create a custom context type. That is a accepted Go idiom https://go.dev/wiki/CodeReviewComments#contexts

worker/tuning.go
Sushisource: Oh, no this comment is just left over from when I thought I would be able to keep the interface private

internal/internal_eager_activity.go
Sushisource: This is the same logic - `heldSlotCount` is gone now

internal/internal_eager_activity.go
Sushisource: Because this needs to access the slot supplier to get the replacement for `heldSlotCount`. I could pass that through as a method on eagerWorker, or I guess have a reference to the atomic issuedCount in here, if that's better

internal/internal_worker.go
Sushisource: Yes, fixed thanks

internal/internal_worker.go
Sushisource: I dunno, why it takes both - just already was that way.. I moved the metric handler in because it needs to be in the base worker options because it needed to be tagged before calling `newBaseWorker` now, and in the process I observed what you did and figured there was no reason for it to be a separate option.

I could also just move the other parameters in too if we want. I only didn't since I didn't touch them.

internal/tuning.go
Sushisource: I was just going with Chad's suggestion here: https://github.com/temporalio/sdk-go/pull/1546#discussion_r1693589674

I don't care much either way. I would be happy to do my original thing where the std context is separate.

internal/internal_task_pollers.go
Sushisource: Well, that worked out. Turns out it no longer happens anymore anyway, which must've been a result of refactoring along the way.

worker/tuning.go
```diff
@@ -0,0 +1,86 @@
...
+	"go.temporal.io/sdk/internal"
+)
+
+// WorkerTuner allows for the dynamic customization of some aspects of worker behavior.
+//
+// WARNING: Custom implementations of SlotSupplier are currently experimental.
+type WorkerTuner = internal.WorkerTuner
+
+// SlotPermit is a permit to use a slot.
+type SlotPermit = internal.SlotPermit
```
Quinn-With-Two-Ns, line 37: Is there a reason `SlotPermit ` is not experimental?

internal/internal_eager_activity.go
Quinn-With-Two-Ns: Yeah I would extend `eagerWorker` interface to provide the functionality you need. Eager activities used to be implemented like this and having different parts of the system reach into implementation details like this caused a number of bugs.

internal/tuning.go
Quinn-With-Two-Ns: Yeah I would not embedded the context, I would keep them separate. I am not sure about this `Slot*Context ` , it has an established meaning in Go, maybe `Slot*Request`?

internal/tuning.go
```diff
@@ -0,0 +1,392 @@
...
+		workflowSlotSupplier:      options.WorkflowSlotSupplier,
+		activitySlotSupplier:      options.ActivitySlotSupplier,
+		localActivitySlotSupplier: options.LocalActivitySlotSupplier,
+		nexusSlotSupplier:         options.NexusSlotSupplier,
+	}, nil
+}
+
+// FixedSizeTunerOptions are the options used by NewFixedSizeTuner.
+type FixedSizeTunerOptions struct {
+	NumWorkflowSlots      int
```
Quinn-With-Two-Ns, line 167: Can we add docs for all these user exposed options?

worker/tuning.go
Sushisource: Since it has to be used by the fixed size supplier which is ostensibly not. But we could mark this as experimental too if we like.

internal/internal_worker.go
Quinn-With-Two-Ns: Yeah it's just now a bit silly to pass `workerType` in at all because your tagging the metric handler outside `newBaseWorker`, but tagging the logger with the same tag in `newBaseWorker`

Quinn-With-Two-Ns: Why does it need to be tagged before calling `newBaseWorker`?  Why can't we just construct `newTrackingSlotSupplier ` inside `newBaseWorker`?

internal/internal_worker.go
```diff
@@ -410,7 +425,6 @@ func newSessionWorker(service workflowservice.WorkflowServiceClient, params work
 	params.TaskQueue = creationTaskqueue
 	// Although we have session token bucket to limit session size across creation
 	// and recreation, we also limit it here for creation only
-	params.ConcurrentActivityExecutionSize = maxConcurrentSessionExecutionSize
```
Quinn-With-Two-Ns, line 413: Why is this line being removed?

internal/internal_worker.go
```diff
@@ -317,19 +318,26 @@ func newWorkflowTaskWorkerInternal(
...
+		maxTaskPerSecond: defaultWorkerTaskExecutionRate,
+		taskWorker:       poller,
+		identity:         params.Identity,
+		workerType:       workerType,
+		stopTimeout:      params.WorkerStopTimeout,
+		fatalErrCb:       params.WorkerFatalErrorCallback,
+		metricsHandler:   metricsHandler,
+		slotReservationData: slotReservationData{
+			taskQueue: params.TaskQueue,
+			logger:    params.Logger,
```
Quinn-With-Two-Ns, line 337: Wouldn't you want the logger here tagged? This is also identical in the 4 calls to `newBaseWorker` so I think we can just move it into `newBaseWorker`

internal/internal_worker.go
Sushisource: It makes it easier for tests that call `newBaseWorker` directly to use a specific slot supplier.

But, yeah I can just get rid of the other params to newBaseWorker and move workerType out of the options.

internal/internal_worker.go
Sushisource: Because the field doesn't exist any more, the session bucket ought to be responsible for any limiting specific to sessions

internal/internal_worker.go
Sushisource: Moved

internal/internal_eager_activity.go
Sushisource: Brought this back per discussion

internal/internal_worker.go
Quinn-With-Two-Ns: Hm  need to think about this a bit, but regardless isn't the comment above wrong now?

worker/tuning.go
Quinn-With-Two-Ns: I would mark this all experimental personally

internal/internal_worker.go
Quinn-With-Two-Ns: We are repeating the same basic code 4 times for each worker, i would prefer to optimize for the actual SDK code and not tests

Quinn-With-Two-Ns: In the SDK `newBaseWorker` is only used `newTrackingSlotSupplier` so i think this common code should just be in `newBaseWorker`

internal/internal_worker.go
Sushisource: Removed

internal/internal_worker.go
Sushisource: Sure, changed.

internal/tuning.go
```diff
@@ -0,0 +1,455 @@
...
+	// implementations.
+	UserData any
+	// Specifically eager activities need to keep track of their own concurrent max separately and
+	// this helps them do that. It can be used for other specific use cases in the future.
+	extraReleaseCallback func()
+}
+
+// SlotReservationInfo contains information that SlotSupplier instances can use during
+// reservation calls. It embeds a standard Context.
+type SlotReservationInfo interface {
```
Quinn-With-Two-Ns, line 65: Should it also include worker type?

internal/tuning.go
```diff
@@ -0,0 +1,455 @@
...
+	// activities, this is the same as the workflow's task queue.
+	TaskQueue() string
+	// NumIssuedSlots returns the number of slots that have already been issued by the supplier.
+	NumIssuedSlots() int
+}
+
+// SlotMarkUsedInfo contains information that SlotSupplier instances can use during
+// SlotSupplier.MarkSlotUsed calls.
+type SlotMarkUsedInfo interface {
+	allFnsCommonInfo
```
Quinn-With-Two-Ns, line 77: Nit: for readability in go docs it is better to not embedded the methods

internal/internal_worker.go
Quinn-With-Two-Ns: Looking at this change, it looks like you always giving the session `creationWorker` the activity workers slot supplier? That doesn't seem right. Shouldn't it have a fixed slot supplier that is configured by `maxConcurrentSessionExecutionSize `?

internal/tuning.go
Sushisource: Well that seems like an awful reason. Let's not use a language feature because the docs renderer is bad. :laughing: 

I get your concern, though, I'll change it, it just seems really stupid that the docs render isn't good at handling that.

internal/internal_worker.go
Quinn-With-Two-Ns: It still looks like we are constructing the logger and metric handler outside `newBaseWorker`

internal/tuning.go
Sushisource: Eh, Java doesn't have it either. The problem is there's no typed way to do that right now since they're just strings. Could make a brand new enum for it I suppose. I will add the other things Java has.

Quinn-With-Two-Ns: I would be fine passing the string through I don't think it needs to be typed. Fine leaving stuff off here that we can add later. I just wasn't sure if it was intentional or not

internal/tuning.go
```diff
@@ -0,0 +1,455 @@
...
+		localActivitySlotSupplier: laSS,
+		nexusSlotSupplier:         nexusSS,
+	}, nil
+}
+
+// FixedSizeSlotSupplier is a slot supplier that will only ever issue at most a fixed number of
+// slots.
+type FixedSizeSlotSupplier struct {
+	// The maximum number of slots that this supplier will ever issue.
+	NumSlots int
```
Quinn-With-Two-Ns, line 242: Is there a reason this is exposed when we have `MaxSlots`? A user may think they could modify this number to change the number of slots

internal/internal_worker.go
Sushisource: Yes, that makes sense. This seems dumb on my part but I didn't even realize two workers were being made here.

internal/internal_worker.go
Sushisource: Sorry, I misread you. I read it as you not wanting to pass _in_ workerType, hence them needing to be outside, but yeah I agree it'd be better the other way around.

internal/tuning.go
Sushisource: Good point, hidden.

internal/tuning.go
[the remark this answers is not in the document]
cretz: Concur, don't need to retrieve from Go context (can keep separate or, how I like, embed the Go context in the slot reserve context). My only remaining concern is using Go contexts in non-cancellable/non-long-lived situations like "try". I think we should as Go context is for more than just long-running functions.

internal/tuning.go
cretz: Coming back here, the separation of the two contexts and not calling the latter a context to avoid confusion makes sense to me

cretz (approved):
None of my existing concerns remain or are blockers. Always a bit scary changing internal polling machinations, so goes without saying that tests will be very important. Part of me wonders if we should make a larger effort to make this all new code instead of changing existing code just so we could have some oh-no flag to flip back, but that would seem hard without a lot of just copied functions.

internal/internal_worker_base.go
Quinn-With-Two-Ns: @Sushisource I agree this looks like a bug

internal/internal_worker.go
Quinn-With-Two-Ns: Sounds good, can you please add the comment back then

internal/internal_worker_base.go
Sushisource: This channel doesn't get closed any more

internal/internal_worker_base.go
```diff
@@ -332,61 +356,101 @@ func (bw *baseWorker) runPoller() {
...
+
 	for {
+		bw.stopWG.Add(1)
+		go func() {
+			defer bw.stopWG.Done()
+			s, err := bw.slotSupplier.ReserveSlot(ctx, &bw.options.slotReservationData)
+			if err != nil {
+				if !errors.Is(err, context.Canceled) {
+					bw.logger.Error(fmt.Sprintf("Error while trying to reserve slot: %v", err))
+					reserveChan <- nil
```
Quinn-With-Two-Ns, line 371: This can deadlock if the worker never pulls from `reserveChan`, because it is stopping. you need to select with ctx.Done(). I think it would be better to just retry the `ReserveSlot ` call in this goroutine rather then coordinating with the outside goroutine, but your call

internal/internal_worker_base.go
```diff
@@ -332,61 +356,101 @@ func (bw *baseWorker) runPoller() {
...
+	reserveChan := make(chan *SlotPermit)
+
 	for {
+		bw.stopWG.Add(1)
+		go func() {
+			defer bw.stopWG.Done()
+			s, err := bw.slotSupplier.ReserveSlot(ctx, &bw.options.slotReservationData)
+			if err != nil {
+				if !errors.Is(err, context.Canceled) {
+					bw.logger.Error(fmt.Sprintf("Error while trying to reserve slot: %v", err))
```
Quinn-With-Two-Ns, line 370: We should not use `fmt.Sprintf` , we should use structured logging like the rest of the SDK `logger.Error("Error while trying to reserve slot", "error", err)`

internal/internal_worker_base.go
Sushisource: I tried it and turned out to not be much simpler. You still need two loops and was giving me some nonobvious problems.

Quinn-With-Two-Ns (approved):

PR #1548, opened 2024-07-16
Fix bug in updateWorkflowByID

Quinn-With-Two-Ns:
`updateWorkflowByID` was not targeting the correct workflow handle. This caused updating a child workflow to fail.


cretz (approved):

dandavison (approved):
LGTM

PR #1553, opened 2024-07-19
Add TaskQueueStats to DescribeTaskQueueEnhanced

ShahabT:
## What was changed

Add TaskQueueStats to DescribeTaskQueueEnhanced to report backlog info and task add/dispatch rate.

## Why?

Useful for worker scaling.

## Checklist

1. Closes 

2. How was this tested:

Functional test added.

3. Any docs updates needed?

Yes, will do in a separate PR.


client/client.go
```diff
@@ -345,24 +345,25 @@ type (
 	TaskQueueVersionSelection = internal.TaskQueueVersionSelection
 
 	// TaskQueueDescription is the response to [Client.DescribeTaskQueueEnhanced].
-	// WARNING: Worker versioning is currently experimental.
```
cretz, line 348: I assume this is being removed because this stuff is now considered GA? Isn't this all part of worker versioning technically the stats are per build ID, or is this RPC and the way it comes back considered GA?

internal/internal_versioning_client.go
```diff
@@ -174,6 +216,7 @@ func (o *DescribeTaskQueueEnhancedOptions) validateAndConvertToProto(namespace s
 		TaskQueueTypes:         taskQueueTypes,
 		ReportPollers:          o.ReportPollers,
 		ReportTaskReachability: o.ReportTaskReachability,
+		ReportStats: o.ReportStats,
```
cretz, line 219: Did this file get properly `go fmt`d?

cretz, line 219: Is this available in the current dev server? We won't want to make this available until users can actually use it and we have tested it the same way they might.

cretz (commented):
Mostly LGTM, just want to wait until it is available to users server side so we can test it like users will

client/client.go
ShahabT: That's something I wanted to discuss actually. The versioning stuff are not GA for sure. I'd like to be able to say the Enhanced DecribeTaskQueue API parts that don't have to do with versioning is GA. It's a little tricky because of the `versions_info` map.

What do you guys think of a change like below to the API? It removes the empty string ugliness, also it allows us to keep the build_id related parts experimental and unversioned part GA.
image
cc @cretz @Quinn-With-Two-Ns

ShahabT: considering to rename `TaskQueueVersionInfo` to just `TaskQueueInfo` as well. That would fit better with unversioned.

cretz: That API change works for me, but the empty string does too. I have no strong opinion, though I could understand not wanting to go through a cycle of API changes just to be able to split up the "unstable" comment.

internal/internal_versioning_client.go
```diff
@@ -126,11 +128,51 @@ type (
...
 
+	// TaskQueueStats contains statistics about task queue backlog and activity.
+	//
+	// For workflow task queue type, this result is partial because tasks sent to sticky queues are not included. Read
+	// comments above each metric to understand the impact of sticky queue exclusion on that metric accuracy.
+	TaskQueueStats struct {
+		// The approximate number of tasks backlogged in this task queue. May count expired tasks but eventually
+		// converges to the right value. Can be relied upon for scaling decisions.
+		//
+		// Special note for workflow task queue type: this metric does not count sticky queue tasks. However, because
```
Quinn-With-Two-Ns, line 139: Does the sticky task queue here get its own `TaskQueueStats`?

ShahabT: There is no plan for supporting user-facing API for TaskQueueStats for sticky queues. There will be a tdbg DescribeTaskQueuePartition which will support sticky queues. Also, the backlog hint will be still sent in poll response.

@Quinn-With-Two-Ns Please lmk if there is objections/concerns.

Quinn-With-Two-Ns: Talked offline, there may be some cases where not knowing stats for sticky task queue may make more difficult, but this is a know/accepted limitation of the API

internal/internal_worker.go
```diff
@@ -1608,7 +1608,14 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 		tagNamespace, client.namespace,
 		tagTaskQueue, taskQueue,
 		tagWorkerID, workerParams.Identity,
+		tagBuildID, workerParams.getBuildID(),
```
Quinn-With-Two-Ns, line 1611: Are you adding `tagBuildID ` twice here?

ShahabT: fixed.

ShahabT:
@cretz @Quinn-With-Two-Ns it seems we are unblocked, dependency-wise. Can we get this PR merged?

test/worker_versioning_test.go
```diff
@@ -800,6 +801,75 @@ func (ts *WorkerVersioningTestSuite) TestReachabilityVersionsWithRules() {
...
+		ts.validateTaskQueueStats(expectedWorkflowStats, taskQueueInfo.VersionsInfo[""].TypesInfo[client.TaskQueueTypeWorkflow].Stats)
+		ts.validateTaskQueueStats(expectedActivityStats, taskQueueInfo.VersionsInfo[""].TypesInfo[client.TaskQueueTypeActivity].Stats)
+	}
+
+	// Basic workflow runs two activities
+	handle, err := ts.client.ExecuteWorkflow(ctx, ts.startWorkflowOptions("basic-wf"), ts.workflows.Basic)
+	ts.NoError(err)
+
+	// Give time for the task to be sent to the TQ
+	time.Sleep(500 * time.Millisecond)
```
Quinn-With-Two-Ns, line 828: We should try to avoid sleep in tests, it is very flaky, we should wait for the condition to become true

ShahabT: replaced it with eventually.

Quinn-With-Two-Ns (approved):
LGTM, good to merge once CI is passing. If you need any help there please reach out

internal/internal_versioning_client.go
```diff
@@ -220,6 +278,20 @@ func taskQueueTypeInfoFromResponse(response *taskqueuepb.TaskQueueTypeInfo) Task
...
+func statsFromResponse(stats *taskqueuepb.TaskQueueStats) *TaskQueueStats {
+	if stats == nil {
+		return nil
+	}
+
+	return &TaskQueueStats{
+		ApproximateBacklogCount: stats.GetApproximateBacklogCount(),
+		ApproximateBacklogAge:   stats.GetApproximateBacklogAge().AsDuration(),
+		TasksAddRate:            stats.TasksAddRate,
+		TasksDispatchRate:       stats.TasksDispatchRate,
```
ShahabT, line 294: @cretz @Quinn-With-Two-Ns, realizing that eager activities are enabled by default, I'm now inclined to remove separate add/dispatch rate fields and replace both with a `BacklogIncreaseRate` field. Reason being that the individual rates can be very off an easily mislead people because they'd not include eager activities and sticky workflow tasks.
wdyt?

(The API can remain the same, or I can change that in the next release as well by deprecating the separate rates and adding a new `backlog_increase_rate` filed.)

```suggestion
	return &TaskQueueStats{
		ApproximateBacklogCount: stats.GetApproximateBacklogCount(),
		ApproximateBacklogAge:   stats.GetApproximateBacklogAge().AsDuration(),
		ApproximateBacklogIncreaseRate:            stats.TasksAddRate - stats.TasksDispatchRate,
```

cretz: IMO we should give the same thing that the underlying API gives. If there are concerns about user confusion, they can be documented. If we must _also_ provide helpers or derived fields we can (and we should document how they are derived). The SDK should not deviate so far from the API in this point (many users will use our API directly).

ShahabT: OK. I kept the separate fields with enough documentation and add a derivative field as well.

cretz (approved):

PR #1555, opened 2024-07-19
Nexus

bergundy:
## What was changed

- Added the `temporalnexus` package and implemented the handler side for Nexus, including registering and dispatching Nexus Operations.
- Added the ability to execute Nexus Operations from a workflow.
- Added basic support for running Nexus Operations in the test environment.
- Added memoizing to `worker.Start()` to return consistent errors to callers and avoid rerunning the function unnecessarily.
- Updated the integration test's dev server to run CLI `0.14.0-nexus.0` which includes server `1.25.0-rc.0`.

See the [proposal](https://github.com/temporalio/proposals/blob/b72c49b0c2278e916265b00a49638006f8fce469/nexus/sdk-go.md) for more information.

Most of this code has been reviewed already in #1466, #1473, and #1475, which are all squashed in the first commit.


go.mod
```diff
@@ -1,32 +1,37 @@
 module go.temporal.io/sdk
 
-go 1.20
+go 1.21
```
cretz, line 3: Hrmm, do we want to bump minimum Go version? Not that I mind, just want to make sure it's a conscious decision and not buried in this PR

go.mod
```diff
@@ -1,32 +1,37 @@
...
+
+toolchain go1.21.1
 
 require (
 	github.com/facebookgo/clock v0.0.0-20150410010913-600d898af40a
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/mock v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
+	github.com/nexus-rpc/sdk-go v0.0.8-0.20240617225139-cd9d6c50e99d
```
cretz, line 12: Can you tag this?

go.mod
bergundy: I'm not sure why this happened, maybe because the nexus SDK depends on `slog`?

go.mod
bergundy: It's tagged, let me fix the file.

bergundy: Fixed.

testsuite/devserver.go
```diff
@@ -199,6 +199,7 @@ func downloadIfNeeded(ctx context.Context, options *DevServerOptions, logger log
 		return "", fmt.Errorf("unsupported architecture %v", arch)
 	}
 	infoURL := fmt.Sprintf("https://temporal.download/cli/%v?platform=%v&arch=%v&sdk-name=sdk-go&sdk-version=%v", url.QueryEscape(version), platform, arch, internal.SDKVersion)
+	fmt.Println(infoURL)
```
bergundy, line 202: I'm removing this.

go.mod
cretz: Yeah, or more importantly, Nexus Go SDK itself requires 1.21+ which infects all users of it. Technically that's the oldest non-EOL, @Quinn-With-Two-Ns - thoughts?

Also, another concern I now have is about transitive dependencies. Nexus SDK is now forcing `github.com/gorilla/mux` to at least `v1.8.0` for every user of the Temporal Go SDK. I am not sure we want to force this dependency or our needed version of it on every user.

bergundy: I'm sure we can do without this dependency. I know Go's built-in router has advanced in recent versions.
Can we live with this for the immediate future and investigate next week?

Quinn-With-Two-Ns: Yeah `github.com/gorilla/mux` is quite a rough dependency to start requiring for non nexus users.

Quinn-With-Two-Ns: Upgrading to `go 1.21` is fine if it is needed

cretz (approved):
Would like CI to pass though, and need to make sure that we make clear in release notes that we now require Go 1.21+.

cretz (approved):

Issue #1539, opened 2024-07-08
Support "query" when listing schedules


cretz:
**Describe the solution you'd like**

Need to support the query option when listing schedules same as we do when listing workflows.

PR #1556, opened 2024-07-19
Add support for query parameter when listing schedules

justinp-tt:
## What was changed

Add support for query parameter when listing schedules

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/1539

2. How was this tested:

Added integration test cases

### DIFF
--- a/internal/internal_schedule_client.go
+++ b/internal/internal_schedule_client.go
@@ -199,6 +199,7 @@ func (sc *scheduleClient) List(ctx context.Context, options ScheduleListOptions)
 			Namespace:       sc.workflowClient.namespace,
 			MaximumPageSize: int32(options.PageSize),
 			NextPageToken:   nextToken,
+			Query:           options.Query,
 		}
 
 		return sc.workflowClient.workflowService.ListSchedules(grpcCtx, request)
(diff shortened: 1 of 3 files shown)


cretz (approved):
LGTM (though `Eventually` has a bit of a poor impl on the `testify` side, but probably fine here). Will leave open a bit to confirm @Quinn-With-Two-Ns has nothing to add.

Also, can you confirm you have privileges to "merge" this? If not, we can merge for you.

internal/internal_schedule_client.go
```diff
@@ -199,6 +199,7 @@ func (sc *scheduleClient) List(ctx context.Context, options ScheduleListOptions)
 			Namespace:       sc.workflowClient.namespace,
 			MaximumPageSize: int32(options.PageSize),
 			NextPageToken:   nextToken,
+			Query:           options.Query,
```
Quinn-With-Two-Ns, line 202: Is there any issue on the server side when passing both a `Query` string and a `NextPageToken`? I don't think we do this anywhere else in the Go SDK and want to make sure the server does the correct thing.

dnr: ListWorkflowExecutions? The SDK doesn't call it but it exposes it to be used that way. IOW the token does not encapsulate the query, you still need the query.

Quinn-With-Two-Ns: yeah the SDK doesn't wrap `ListWorkflowExecutions ` so I wasn't sure on the contract.

Quinn-With-Two-Ns (approved):

Issue #723, opened 2022-02-10
Unable to reset workflow with completed childworkflow whose child workflowID is generated by SDK


yycptt:
## Expected Behavior
When reseting a workflow with no pending child execution. (Reset with pending child currently is not supported.)
After reset, parent workflow should continue execution without any error

## Actual Behavior
After reset, parent workflow will encounter a non-deterministic error during replay. The error happens when processing the child workflow init event and can't find the corresponding child workflow command/state machine.

The root cause is we use child workflow ID to find the corresponding command. However if child workflowID is not specified in child option, SDK will automatically generate one based on the workflow's runID. After reset, the workflow's runID changes but in workflow history the child workflow ID is still based on the original runID. Hence the not found and non-deterministic error.

## Steps to Reproduce the Problem

  1. Run a workflow with a child workflow. Do NOT specify the childworkflow ID in child option.
  2. Reset the workflow to an event_id after the child workflow completed event. Any workflow task close event (completed/failed/timeout) will work.
  3. Check the new workflow's history, the workflow task will fail due to non-deterministic error.

## Specifications

  - Version:
  - Platform:

yycptt:
Server stores a workflow's original runID which won't change during reset. So one potential solution is generate child workflow ID based on this original runID. 

But this solution may conflict with some future work related to reset. 
@yiminc Would you mind provide some insights here? Thanks.

askreet:
I hit this in #1385. I'm curious if there's any risk in using the OriginalRunId in the meantime.

Quinn-With-Two-Ns:
https://github.com/temporalio/sdk-go/pull/1803

PR #1557, opened 2024-07-22
Fix non-deterministic error when resetting a parent workflow with a child workflow

arithran:
## What was changed

Fix non-deterministic error when resetting a parent workflow with a child workflow

## Why?

Includes the following change from cadence: https://github.com/uber-go/cadence-client/pull/1118

## Checklist

1. Closes 
2. https://github.com/temporalio/sdk-go/issues/723

3. How was this tested:

unit tests manual tests

4. Any docs updates needed?


Quinn-With-Two-Ns:
Closing since https://github.com/temporalio/sdk-go/pull/1803 already resolved this issue https://github.com/temporalio/sdk-go/pull/1803

PR #1558, opened 2024-07-22
Update SDK version

bergundy:
Forgot to bump the version in `version.go` before cutting the `1.28.0` release.

### DIFF
--- a/internal/version.go
+++ b/internal/version.go
@@ -30,7 +30,7 @@ package internal
 const (
 	// SDKVersion is a semver (https://semver.org/) that represents the version of this Temporal GoSDK.
 	// Server validates if SDKVersion fits its supported range and rejects request if it doesn't.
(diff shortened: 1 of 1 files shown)


internal/version.go
```diff
@@ -30,7 +30,7 @@ package internal
 const (
 	// SDKVersion is a semver (https://semver.org/) that represents the version of this Temporal GoSDK.
 	// Server validates if SDKVersion fits its supported range and rejects request if it doesn't.
-	SDKVersion = "1.27.0"
+	SDKVersion = "1.28.0"
```
bergundy, line 33:
```suggestion
	SDKVersion = "1.28.1"
```

bergundy: Too late to fix `1.28.0`, will bump straight to `.1`.

Quinn-With-Two-Ns (approved):

PR #1560, opened 2024-07-23
Refactor internal UpdateWorkflow

stephanos:
## What was changed

Splits out `createUpdateWorkflowInput`, `createUpdateWorkflowRequest` and `updateHandleFromResponse`.

## Why?

For re-use by MultiOperation feature.

## Checklist

1. Closes 

2. How was this tested:

Existing tests. No functional change is introduced.

3. Any docs updates needed?

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -52,6 +52,7 @@ import (
 	taskqueuepb "go.temporal.io/api/taskqueue/v1"
 	updatepb "go.temporal.io/api/update/v1"
 	"go.temporal.io/api/workflowservice/v1"
+
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/internal/common/metrics"
 	"go.temporal.io/sdk/internal/common/retry"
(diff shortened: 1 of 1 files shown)


internal/internal_workflow_client.go
```diff
@@ -1866,42 +1881,18 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
...
 	for {
 		var err error
 		resp, err = func() (*workflowservice.UpdateWorkflowExecutionResponse, error) {
 			grpcCtx, cancel := newGRPCContext(ctx, grpcTimeout(pollUpdateTimeout), grpcLongPoll(true), defaultGrpcRetryParameters(ctx))
 			defer cancel()
-			wfexec := &commonpb.WorkflowExecution{
-				WorkflowId: in.WorkflowID,
-				RunId:      in.RunID,
-			}
-			return w.client.workflowService.UpdateWorkflowExecution(grpcCtx, &workflowservice.UpdateWorkflowExecutionRequest{
```
stephanos, line 1888: Note that `workflowservice.UpdateWorkflowExecutionRequest` is only created once, outside of the for loop, now.

cretz:
I think refactoring for some future code should wait until that future code exists

dandavison:
> I think refactoring for some future code should be in the same PR with that future code. There is no need to do it ahead of time if it's not leveraged.

I disagree:

- PRs that mix refactoring with functional changes are hard to review, hard to revert, and hard to understand the squashed commit
- Requiring refactoring to be done in the same PR or "at the same time" as functional changes introduces unnecessary friction into the development process
- Allowing collaborators to sequence PRs as they see fit is an essential part of maintaining a healthy multi-collaborator project that proceeds at good velocity
- Many of our codebases are written in a style with too many long functions, and thus in many cases refactoring such as this is desirable for the sake of refactoring alone.

cretz:
If it's that difficult to marry the code that needs refactoring with the refactoring itself (which I don't agree it is difficult), at the least they should be done at the same time to justify the refactoring. The real friction is guessing at premature abstractions that end up not being accurate when it comes time to apply them.

dandavison:
> guessing at premature abstractions that end up not being accurate when it comes time to apply them.

True, I only listed bullets in favor of a position; that is definitely a bullet in the other direction. Although, if it is just breaking out helper functions, that is much more innocuous and has a higher probability of being a (mild) improvement, than creating new stateful (e.g. struct/class) abstractions that might not be appropriate or required.

Quinn-With-Two-Ns:
I would much rather this refactor come in the PR that needs it.

stephanos:
Okay, fair enough 👍  I just thought we could lighten the review load.

Issue #1561, opened 2024-07-23
Support Search Attributes in Schedules via Query Update


justinp-tt:
Support Search Attributes in Schedules via Query Update

cretz:
Can the title be changed here to reflect this is for schedules and is unrelated to query? Also, this seems to be very Go specific. We usually use the https://github.com/temporalio/features repository to make cross-SDK issues like this one so we track it for each SDK.

PR #1562, opened 2024-07-23
Add support for managing schedule search attributes when when updating a schedule

justinp-tt:
## What was changed

Add support for managing schedule search attributes when when updating a schedule

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/1561 

2. How was this tested:
Added new integration test

### DIFF
--- a/internal/schedule_client.go
+++ b/internal/schedule_client.go
@@ -416,12 +416,31 @@ type (
 		// Memo - Non-indexed user supplied information.
 		Memo *commonpb.Memo
 
-		// SearchAttributes - Indexed info that can be used in query of List schedules APIs. The key and value type must be registered on Temporal server side.
-		// Use GetSearchAttributes API to get valid key and corresponding value type.
+		// SearchAttributes - Specifies Search Attributes that will be attached to the schedule. Search Attributes
+		// are additional indexed information attributed to the schedule and used for search and visibility. The key and
+		// its value type must be registered on Temporal server side.
 		// For supported operations on different server versions see [Visibility].
+		//
+		// nil: leave any pre-existing assigned search attributes intact
+		// empty: remove any and all pre-existing assigned search attributes
+		// attributes present: replace any and all pre-existing assigned search attributes with the defined search
+		//                     attributes, i.e. upsert
 		//
 		// [Visibility]: https://docs.temporal.io/visibility
 		SearchAttributes *commonpb.SearchAttributes
+
+		// TypedSearchAttributes - Specifies Search Attributes that will be attached to the schedule. Search Attributes
+		// are additional indexed information attributed to the schedule and used for search and visibility. The key and
+		// its value type must be registered on Temporal server side.
+		// For supported operations on different server versions see [Visibility].
+		//
+		// nil: leave any pre-existing assigned search attributes intact
+		// empty: remove any and all pre-existing assigned search attributes
+		// attributes present: replace any and all pre-existing assigned search attributes with the defined search
+		//                     attributes, i.e. upsert
+		//
+		// [Visibility]: https://docs.temporal.io/visibility
+		TypedSearchAttributes SearchAttributes
 	}
 
 	// SchedulePolicies describes the current polcies of a schedule.
@@ -477,9 +496,9 @@ type (
 		// Schedule - New schedule to replace the existing schedule with
 		Schedule *Schedule
 
-		// SearchAttributes - Optional indexed info that can be used for querying via the List schedules APIs.
+		// TypedSearchAttributes - Optional indexed info that can be used for querying via the List schedules APIs.
 		// The key and value type must be registered on Temporal server side.
-		SearchAttributes *SearchAttributes
+		TypedSearchAttributes *SearchAttributes
 	}
 
 	// ScheduleUpdateInput describes the current state of the schedule to be updated.
--- a/internal/schedule_client.go
+++ b/internal/schedule_client.go
@@ -429,11 +429,6 @@ type (
 		// its value type must be registered on Temporal server side.
 		// For supported operations on different server versions see [Visibility].
 		//
-		// nil: leave any pre-existing assigned search attributes intact
-		// empty: remove any and all pre-existing assigned search attributes
-		// attributes present: replace any and all pre-existing assigned search attributes with the defined search
-		//                     attributes, i.e. upsert
-		//
 		// [Visibility]: https://docs.temporal.io/visibility
 		TypedSearchAttributes SearchAttributes
 	}
@@ -493,6 +488,11 @@ type (
 
 		// TypedSearchAttributes - Optional indexed info that can be used for querying via the List schedules APIs.
 		// The key and value type must be registered on Temporal server side.
+		//
+		// nil: leave any pre-existing assigned search attributes intact
+		// empty: remove any and all pre-existing assigned search attributes
+		// attributes present: replace any and all pre-existing assigned search attributes with the defined search
+		//                     attributes, i.e. upsert
 		TypedSearchAttributes *SearchAttributes
 	}
 
--- a/internal/schedule_client.go
+++ b/internal/schedule_client.go
@@ -421,11 +421,6 @@ type (
 		// its value type must be registered on Temporal server side.
 		// For supported operations on different server versions see [Visibility].
 		//
-		// nil: leave any pre-existing assigned search attributes intact
-		// empty: remove any and all pre-existing assigned search attributes
-		// attributes present: replace any and all pre-existing assigned search attributes with the defined search
-		//                     attributes, i.e. upsert
-		//
 		// [Visibility]: https://docs.temporal.io/visibility
 		SearchAttributes *commonpb.SearchAttributes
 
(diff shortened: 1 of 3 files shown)


internal/schedule_client.go
```diff
@@ -476,6 +476,10 @@ type (
 	ScheduleUpdate struct {
 		// Schedule - New schedule to replace the existing schedule with
 		Schedule *Schedule
+
+		// SearchAttributes - Optional indexed info that can be used for querying via the List schedules APIs.
+		// The key and value type must be registered on Temporal server side.
+		SearchAttributes *SearchAttributes
```
Quinn-With-Two-Ns, line 482: Don't think we need to take a pointer here. Also what is the behaviour if `SearchAttributes ` is null? I assume it leaves the `SearchAttributes` unchanged?

justinp-tt: Maybe I misunderstood, but I seem to recall the pointer being a suggestion from @cretz . @cretz -- am I off base here?

Yes, null means unchanged.

test/integration_test.go
```diff
@@ -4980,9 +4980,14 @@ func (ts *IntegrationTestSuite) TestScheduleUpdate() {
...
 	}()
+
+	stringKey := temporal.NewSearchAttributeKeyString("CustomStringField")
+	sa := temporal.NewSearchAttributes(stringKey.ValueSet("CustomStringFieldValue"))
+
 	updateFunc := func(input client.ScheduleUpdateInput) (*client.ScheduleUpdate, error) {
 		return &client.ScheduleUpdate{
-			Schedule: &input.Description.Schedule,
+			Schedule:         &input.Description.Schedule,
+			SearchAttributes: &sa,
```
dnr, line 4990: is there also a test (somewhere else in the file?) that an update without SearchAttributes set doesn't modify the search attributes?

internal/schedule_client.go
justinp-tt: Pointers default to nil, which can be useful to indicate that `SearchAttributes` has not been set, as opposed to having an empty but allocated struct. The behavior I would expect:

nil: leave current SAs untouched
empty: remove current SAs

I don't know if there is precedence for this pattern, though.

test/integration_test.go
justinp-tt: I agree the test should be added. I'd like to better understand the testing responsibilities for SDK/CLI and OSS. Do we generally expect all contract-related test cases to reside in both places? Or maybe edge case contract tests in OSS and happy path in CLI/SDK?

internal/schedule_client.go
justinp-tt: A part of me would expect this to operate as an idempotent HTTP PUT REST(ish) resource, which effectively replaces the resource and has a symmetrical contract with GET (i.e. describe).

test/integration_test.go
dnr: I think there's a test in the server for a missing SearchAttributes (at the grpc level) to update the schedule with no change to SAs. What I'd like to test here is that leaving out SearchAttributes in ScheduleUpdate actually sends a grpc call with a missing SearchAttributes. I'd be happy with a unit test that does that, but it might be easier to do an integration test since it's already set up here.

internal/schedule_client.go
dnr: Sadly, it is the way it is because SearchAttributes was a later addition to ScheduleUpdate (and Memo will be another addition). The Schedule field works as you think it should, but we couldn't make SearchAttributes work that way too since existing clients aren't sending it.

And yes, there should be some way (at the Go level) to unset all custom search attributes. And empty (but non-nil) map seems like the most straightforward way, but there could be others.

cretz: Hrmm, I don't recall making this suggestion, but it's possible I did. And looking at it, it makes sense. Unfortunately for this call, we need to differentiate between unset (i.e. don't change) and empty (i.e. actually empty).

test/integration_test.go
```diff
@@ -4980,9 +4980,14 @@ func (ts *IntegrationTestSuite) TestScheduleUpdate() {
 		err = handle.Delete(ctx)
 		ts.NoError(err)
 	}()
+
+	stringKey := temporal.NewSearchAttributeKeyString("CustomStringField")
+	sa := temporal.NewSearchAttributes(stringKey.ValueSet("CustomStringFieldValue"))
+
 	updateFunc := func(input client.ScheduleUpdateInput) (*client.ScheduleUpdate, error) {
```
cretz, line 4987: Thinking about use cases ,I think the `ScheduleDescription` should get a `TypedSearchAttributes` field (in addition to the raw one already present) with the high-level typed search attributes container/collection, and that field could just be passed here as a pointer or slightly mutated then passed here as a pointer. Think about the use case where a user just wants to alter one search attribute but leave all the rest alone. How would that look?

internal/schedule_client.go
justinp-tt: Thanks. I'll go ahead and add tests to ensure:
1. A nil SearchAttributes leaves SAs untouched.
2. An empty SearchAttributes removes any pre-existing SAs.

test/integration_test.go
Quinn-With-Two-Ns: In general the SDK APIs should have tests for all of their documented behaviour. What the server tests at the grpc level doesn't validate the API at the SDK level.

test/integration_test.go
justinp-tt: > Think about the use case where a user just wants to alter one search attribute but leave all the rest alone. How would that look?

This is a replacement, so they would need to pass in all SAs they want applied to the schedule, not just new ones.

internal/schedule_client.go
justinp-tt: Added tests

test/integration_test.go
justinp-tt: Added tests

test/integration_test.go
justinp-tt: Added TypedSearchAttributes

internal/schedule_client.go
```diff
@@ -416,12 +416,31 @@ type (
...
+		//
 		// [Visibility]: https://docs.temporal.io/visibility
 		SearchAttributes *commonpb.SearchAttributes
+
+		// TypedSearchAttributes - Specifies Search Attributes that will be attached to the schedule. Search Attributes
+		// are additional indexed information attributed to the schedule and used for search and visibility. The key and
+		// its value type must be registered on Temporal server side.
+		// For supported operations on different server versions see [Visibility].
+		//
+		// nil: leave any pre-existing assigned search attributes intact
```
cretz, line 437: Nil is not possible for this field

internal/schedule_client.go
```diff
@@ -416,12 +416,31 @@ type (
...
 		//
+		// nil: leave any pre-existing assigned search attributes intact
+		// empty: remove any and all pre-existing assigned search attributes
+		// attributes present: replace any and all pre-existing assigned search attributes with the defined search
+		//                     attributes, i.e. upsert
+		//
 		// [Visibility]: https://docs.temporal.io/visibility
 		SearchAttributes *commonpb.SearchAttributes
+
+		// TypedSearchAttributes - Specifies Search Attributes that will be attached to the schedule. Search Attributes
```
cretz, line 432: This doc discusses `ScheduleDescription` as if it's used for creating/updating schedules, but it's read only information.

internal/schedule_client.go
```diff
@@ -476,6 +495,10 @@ type (
 	ScheduleUpdate struct {
 		// Schedule - New schedule to replace the existing schedule with
 		Schedule *Schedule
+
+		// TypedSearchAttributes - Optional indexed info that can be used for querying via the List schedules APIs.
+		// The key and value type must be registered on Temporal server side.
+		TypedSearchAttributes *SearchAttributes
```
cretz, line 501: Should clarify what `nil` vs empty means here IMO (the stuff on description should probably be here)

test/integration_test.go
cretz, line 4987: May also be worth a test that just mutates a single search attribute on the existing collection and leaves the rest alone, but no big deal

justinp-tt: Test added. To be clear, the server behavior today is upsert. If we add a single search attribute on the existing collection, the entire collection is replaced with the single search attribute. This is inline with the CLI proposal as well:

For each `--schedule-search-attribute flag`, do a replace of the whole collection, including removing ones that were present before and not given.

test/integration_test.go
```diff
@@ -4993,9 +5002,91 @@ func (ts *IntegrationTestSuite) TestScheduleUpdate() {
...
+	}, time.Second, 100*time.Millisecond)
+
+	// updating a single search attribute on an existing collection acts as an upsert on the entire collection
+	newSa := temporal.NewSearchAttributes(stringKey.ValueSet("Changed"))
+	updateFunc = func(input client.ScheduleUpdateInput) (*client.ScheduleUpdate, error) {
+		return &client.ScheduleUpdate{
+			Schedule:              &input.Description.Schedule,
+			TypedSearchAttributes: &newSa,
+		}, nil
+	}
```
cretz, line 5051: I think it's worth demonstrating/testing what updating an attribute without affecting the others looks like. I think it'd be something like:

```go
	updateFunc = func(input client.ScheduleUpdateInput) (*client.ScheduleUpdate, error) {
		newSa := temporal.NewSearchAttributes(
			input.TypedSearchAttributes.Copy(),
			stringKey.ValueSet("Changed"),
		)
		return &client.ScheduleUpdate{
			Schedule:              &input.Description.Schedule,
			TypedSearchAttributes: &newSa,
		}, nil
	}
```

Then check that it only updated that one.

cretz (approved):
LGTM. Just one thing about tests, but not considering it blocking

Issue #1447, opened 2024-04-23
Support for WorkflowIdConflictPolicy


stephanos:
See https://github.com/temporalio/features/issues/437

PR #1563, opened 2024-07-24
Support for WorkflowIdConflictPolicy

stephanos:
## What was changed

Added `WorkflowIdConflictPolicy` to ExecuteWorkflow and SignalWithStart.

_NOTE: I haven't seen any `WorkflowIdReusePolicy` validation inside of `testsuite.WorkflowTestSuite` (except for child workflows), so I didn't add any for the `WorkflowIdConflictPolicy` either._

## Checklist

Closes https://github.com/temporalio/sdk-go/issues/1447

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -633,10 +633,15 @@ type (
 		// Optional: defaulted to 10 secs.
 		WorkflowTaskTimeout time.Duration
 
-		// WorkflowIDReusePolicy - Whether server allow reuse of workflow ID, can be useful
-		// for dedupe logic if set to RejectDuplicate.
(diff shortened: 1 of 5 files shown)


cretz:
@stephanos - can you also make sure this is added for `ScheduleWorkflowAction`?

stephanos:
@cretz I might be missing something, but when I looked at the [schedule code](https://github.com/temporalio/sdk-go/blob/master/internal/internal_schedule_client.go) I saw it's not using `StartWorkflowOptions` anywhere and therefore doesn't have either `WorkflowIdReusePolicy` or `WorkflowIdConflictPolicy` fields?

cretz:
My mistake, this is not needed, I was confusing with languages that reused common sets of start options.

cretz (approved):

Issue #1567, opened 2024-07-29
Worker can panic if it receives a legacy query task with no history and the SDK has a destroyed execution in cache


Quinn-With-Two-Ns:
## Expected Behavior

SDK detects this is a legacy query task and shouldn't have any history and resets the cached exeuciton

## Actual Behavior

SDK panics

```
2024-07-25T10:50:16.216-0700    ERROR   internal/internal_worker_base.go:519    Unhandled panic.        {"Namespace": "default", "TaskQueue": "fuzzer:zvsrg654", "WorkerID": "785534
@ShtinUbuntu2@", "WorkerType": "WorkflowWorker", "PanicError": "runtime error: index out of range [0] with length 0", "PanicStack": "base worker for WorkflowWorker [panic]:\ngo.tem
poral.io/sdk/internal.(*workflowTaskHandlerImpl).GetOrCreateWorkflowContext(0xc0002be410, 0xc0006dfa00, {0xf775a0, 0xc00029bb20})\n\t/home/shtin/go/pkg/mod/go.temporal.io/sdk@v1.28
.1/internal/internal_task_handlers.go:772 +0xafe\ngo.temporal.io/sdk/internal.(*workflowTaskPoller).processWorkflowTask(0xc000228b40, 0xc000cd4a50)\n\t/home/shtin/go/pkg/mod/go.tem
poral.io/sdk@v1.28.1/internal/internal_task_pollers.go:344 +0x19b\ngo.temporal.io/sdk/internal.(*workflowTaskPoller).ProcessTask(0xc000228b40, {0xc496e0, 0xc000cd4a50})\n\t/home/sh
tin/go/pkg/mod/go.temporal.io/sdk@v1.28.1/internal/internal_task_pollers.go:321 +0x78\ngo.temporal.io/sdk/internal.(*baseWorker).processTask(0xc00034c420, {0xc4a320, 0xc001335330})
\n\t/home/shtin/go/pkg/mod/go.temporal.io/sdk@v1.28.1/internal/internal_worker_base.go:528 +0x153\ngo.temporal.io/sdk/internal.(*baseWorker).processTaskAsync.func1()\n\t/home/shtin
/go/pkg/mod/go.temporal.io/sdk@v1.28.1/internal/internal_worker_base.go:379 +0x45\ncreated by go.temporal.io/sdk/internal.(*baseWorker).processTaskAsync in goroutine 26\n\t/home/sh
[...]

PR #1568, opened 2024-07-29
Protect against legacy queries when state is destroyed

Quinn-With-Two-Ns:
Protect against legacy queries when workflow execution is destroyed. Previously the SDK would panic since legacy queries against a cached execution have no history. 

closes https://github.com/temporalio/sdk-go/issues/1567


cretz (approved):
Nice. I can see how a 0-cache integration test didn't catch this. I wonder if we should have some kind of test that still has cache/sticky enabled but clears cache after each task completion to exercise all the sticky history refetch behavior.

Quinn-With-Two-Ns:
Clearing cache actually wouldn't catch this, it requires a destroyed cached execution. Maybe the long term solution is to refactor the SDK to remove this concept of a `destroyed cached execution.`

PR #1572, opened 2024-07-29
Add rule code to warning

dandavison:


internal/internal_workflow.go
```diff
@@ -52,7 +52,7 @@ const (
 	defaultCoroutineExitTimeout = 100 * time.Millisecond
 
 	panicIllegalAccessCoroutineState = "getState: illegal access from outside of workflow context"
-	unhandledUpdateWarningMessage    = "Workflow finished while update handlers are still running. This may have interrupted work that the" +
+	unhandledUpdateWarningMessage    = "TMPRL1102 Workflow finished while update handlers are still running. This may have interrupted work that the" +
```
Quinn-With-Two-Ns, line 55:
```suggestion
	unhandledUpdateWarningMessage    = "[TMPRL1102] Workflow finished while update handlers are still running. This may have interrupted work that the" +
```

dandavison: Right good catch I'll change that in the others

Quinn-With-Two-Ns (approved):

PR #1579, opened 2024-08-05
Update-with-Start operation 

stephanos:
## What was changed

Adds support for Update-with-Start, using the MultiOperation API (https://github.com/temporalio/api/pull/367).

## Why?

To allow Update-with-Start feature.

## Checklist

1. Closes 

2. How was this tested:

Added tests.

3. Any docs updates needed?

Yes! I will update the docs once the PR is merged?


test/integration_test.go
```diff
@@ -38,14 +38,13 @@ import (
...
 
-	"go.opentelemetry.io/otel/baggage"
-
 	"github.com/opentracing/opentracing-go"
 	"github.com/pborman/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"github.com/uber-go/tally/v4"
+	"go.opentelemetry.io/otel/baggage"
```
stephanos, line 47: This was moved by my IDE and I was about to move it back, but it seems better now.

internal/client.go
```diff
@@ -643,9 +644,18 @@ type (
 		// Optional: defaulted to Fail.
 		WorkflowIDConflictPolicy enumspb.WorkflowIdConflictPolicy
 
+		// WorkflowOperation - Operation to execute at Workflow Start. For example, see PrepareUpdateWorkflowOperation
+		// to perform Update-with-Start. Note that if the workflow is already running and WorkflowIDConflictPolicy is
+		// set to UseExisting, the start is skipped and only the operation is executed.
+		//
+		// Optional: defaults to nil.
+		WorkflowOperation StartWorkflowOperation
```
stephanos, line 652: Initially I wanted to make this a list of operations, but I think having a single operation is just simpler and will work for most use cases. If we wanted to support multiple operations, we can introduce a "composite operation" type which can take a list of operations.

internal/internal_workflow_client.go
```diff
@@ -1617,27 +1605,35 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+		}
+
+		// Allow already-started error
+		if e, ok := err.(*serviceerror.WorkflowExecutionAlreadyStarted); ok && !in.Options.WorkflowExecutionErrorWhenAlreadyStarted {
+			runID = e.RunId
+		} else if err != nil {
+			return nil, err
+		} else {
+			runID = response.RunId
+		}
```
stephanos, line 1631: `WorkflowExecutionAlreadyStarted` doesn't work for Update-with-Start as it is a client-side setting the Server doesn't know about, so the request will fail regardless of this setting. I've considered some alternatives to ignoring it (e.g. enforcing it to be in-sync with the `WorkflowIdConflictPolicy`), but this seems to be the best approach.

There is a case to be made to deprecate `WorkflowExecutionErrorWhenAlreadyStarted` once the `WorkflowIdConflictPolicy` is officially released.

internal/client.go
```diff
@@ -1003,6 +1027,32 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+	if run == nil {
+		return nil, errors.New("invalid workflow run: is nil")
+	}
+	if runImpl, ok := run.(*workflowRunImpl); ok {
+		if runImpl.startRequest != op.startRequest {
+			return nil, errors.New("invalid workflow run: does not belong to update operation")
+		}
+	}
+	return op.interceptor.updateHandleFromResponse(
+		context.Background(), enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_UNSPECIFIED, op.resp)
```
stephanos, line 1051: By using wait stage unspecified, it returns a handle (either completed or lazy) immediately without blocking.

internal/client.go
```diff
@@ -1003,6 +1027,32 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+// Get returns a WorkflowUpdateHandle, or an error if the workflow run is invalid.
+func (op *UpdateWorkflowOperation) Get(run WorkflowRun) (WorkflowUpdateHandle, error) {
+	if run == nil {
+		return nil, errors.New("invalid workflow run: is nil")
+	}
+	if runImpl, ok := run.(*workflowRunImpl); ok {
+		if runImpl.startRequest != op.startRequest {
+			return nil, errors.New("invalid workflow run: does not belong to update operation")
+		}
+	}
```
stephanos, line 1049: Open to other ideas; this was the best way I could think of to ensure the operation and workflow run match.

internal/internal_workflow_client.go
```diff
@@ -2027,6 +2138,52 @@ func (w *workflowClientInterceptor) PollWorkflowUpdate(
...
+	case *updatepb.Outcome_Success:
+		return &completedUpdateHandle{
+			value:            newEncodedValue(v.Success, w.client.dataConverter),
+			baseUpdateHandle: baseUpdateHandle{ref: resp.GetUpdateRef()},
+		}, nil
+	default:
+		return nil, fmt.Errorf("unsupported outcome type %T", resp.GetOutcome().GetValue())
+	}
+}
+
```
stephanos, line 2186: Same as before.

internal/internal_workflow_client.go
```diff
@@ -1658,9 +1654,118 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+			return nil, fmt.Errorf("%w: %w", errInvalidWorkflowOperation, err)
+		}
+
+		if updateReq.WorkflowExecution.WorkflowId == "" {
+			updateReq.WorkflowExecution.WorkflowId = startRequest.WorkflowId
+		}
+		if t.startRequest == nil {
+			t.startRequest = startRequest
+		} else if t.startRequest != startRequest && t.resp != nil {
+			return nil, fmt.Errorf("%w: update operation was already executed", errInvalidWorkflowOperation)
```
stephanos, line 1686: Since the `StartWorkflowOperation` is stateful, I think re-using of a successfully executed one should be discouraged. The usage is clearer if there is a 1:1 relationship between the operation and the ExecuteWorkflow call.

internal/client.go
```diff
@@ -1003,6 +1027,32 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+func PrepareUpdateWorkflowOperation(options UpdateWorkflowOptions) (*UpdateWorkflowOperation, error) {
+	input, err := createUpdateWorkflowInput(options)
+	if err != nil {
+		return nil, err
+	}
+	return &UpdateWorkflowOperation{input: input}, nil
+}
+
+// Get returns a WorkflowUpdateHandle, or an error if the workflow run is invalid.
+func (op *UpdateWorkflowOperation) Get(run WorkflowRun) (WorkflowUpdateHandle, error) {
```
stephanos, line 1041: I'm going back and forth on this API. Options I'm considering:

**(1) `Get(run WorkflowRun) (WorkflowUpdateHandle, error)`** (as written here)

The safest but most verbose option. The `run` parameter is used to ensure the function is invoked _after_ `ExecuteWorkflow` and that the operation matches the workflow run. But it requires an extra error check that is _only_ there to prevent a mistake by the user, not to surface an actual runtime error.

**(2) `Get(run WorkflowRun) WorkflowUpdateHandle`**

Removes the error check. If the user passes in `nil`, it would panic (akin to a NPE). If the user passes in the wrong workflow run, it would panic (not ideal).

**(3) `Get(ctx context.Context, valuePtr interface{}) error`**

If the operation embeds `WorkflowUpdateHandle`, the extra step to obtain the handle is removed entirely. While this is the most convenient option, it also means that there is a risk the user invokes the handle's functions before executing the operation (if `Get` is invoked too early, it would return the appropriate error).

**(4) `GetResult(ctx context.Context, run WorkflowRun, valuePtr interface{}) error`**

Combined with (1), adding an additional shortcut to get the update outcome directly compresses two calls into one. This is based on the assumption that in most cases, users care about the result more than anything else from the `WorkflowUpdateHandle` (but they could still obtain it, if needed).

Essentially it comes down to safety vs convenience. Since all users would benefit from less verbose code, I think the question is how likely the user errors are the safest options would prevent/make clearer.

internal/client.go
```diff
@@ -713,6 +723,20 @@ type (
 		callbacks []*commonpb.Callback
 	}
 
+	// StartWorkflowOperation is a type of operation that can be executed as part of a workflow start.
+	StartWorkflowOperation interface {
+		isStartWorkflowOperation()
+	}
```
stephanos, line 729: One gripe I have with this name is that it makes it seem like it's the "StartWorkflow" operation (ie an operation to start a workflow).

test/integration_test.go
```diff
@@ -3556,6 +3555,137 @@ func (ts *IntegrationTestSuite) TestUpdateSettingHandlerInHandler() {
...
+	ts.Equal(run1.GetRunID(), run4.GetRunID())
+
+	updateHandle, err = updateOp.Get(run4)
+	ts.NoError(err)
+	ts.NoError(updateHandle.Get(ctx, &updateResult))
+	ts.Equal(2, updateResult)
+
+	ts.NoError(run1.Get(ctx, &workflowResult))
+	ts.Equal(2, workflowResult)
+}
```
stephanos, line 3687: NOTE: I wanted to break out the individual test cases into nested tests via `ts.Run` but none of the other tests are doing that. I think it would help with scoping (it's easy to accidentally use the wrong variable) and readability. Let me know what you think.

internal/internal_workflow_client.go
```diff
@@ -1658,9 +1654,117 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+		}
+
+		if updateReq.WorkflowExecution.WorkflowId == "" {
+			updateReq.WorkflowExecution.WorkflowId = startRequest.WorkflowId
+		}
+		if t.startRequest == nil {
+			t.startRequest = startRequest
+		} else if t.startRequest != startRequest && t.resp != nil &&
+			t.resp.Stage != enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED {
+			return nil, fmt.Errorf("%w: update operation was already executed", errInvalidWorkflowOperation)
```
stephanos, line 1687: Since the StartWorkflowOperation is stateful, I think re-using a successfully executed one needs to be disallowed. The usage is clearer if there is a 1:1 relationship between the operation and the ExecuteWorkflow call.

internal/internal_workflow_client.go
```diff
@@ -1658,9 +1654,111 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+	switch t := operation.(type) {
+	case *UpdateWorkflowOperation:
+		updateReq, err := w.createUpdateWorkflowRequest(ctx, t.input)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %w", errInvalidWorkflowOperation, err)
+		}
+
+		if updateReq.WorkflowExecution.WorkflowId == "" {
+			updateReq.WorkflowExecution.WorkflowId = startRequest.WorkflowId
+		}
```
stephanos, line 1682: If the user set the Workflow ID manually and it doesn't match with the Start Workflow's, the server will return a validation error.

internal/client.go
```diff
@@ -1003,6 +1024,26 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+	return &UpdateWorkflowOperation{input: input}, nil
+}
+
+// Get blocks on the outcome of the update.
+func (op *UpdateWorkflowOperation) Get(ctx context.Context, valuePtr interface{}) error {
+	if op.WorkflowUpdateHandle == nil {
+		return fmt.Errorf("update operation has not been executed")
+	}
+	return op.WorkflowUpdateHandle.Get(ctx, valuePtr)
+}
```
stephanos, line 1043: I got here after some trial and error. Options I've considered:

**(1) `Get(ctx context.Context, valuePtr interface{}) error`** (as written here)

When the operation embeds `WorkflowUpdateHandle`, an extra step to obtain the handle is removed entirely. While this is the most convenient option, it also means that there is a risk the user invokes the handle's functions _before_ executing the operation (however, if `Get` is invoked too early, it returns the appropriate error).

```go
var result any
if err := upd.Get(context.Background(), &result); err != nil {
    return nil, err 
}
``` 

**(2) `Get(run WorkflowRun) (WorkflowUpdateHandle, error)`**

The safest but most verbose option. The `run` parameter is used to ensure the function is invoked _after_ `ExecuteWorkflow`. But it requires an extra error check that is _only_ there to prevent a mistake by the user, not to surface an actual runtime error.

```go
updHandle, err := upd.Get(context.Background(), run, &result)
if err != nil {
    return nil, err
}

var result any
if err := updHandle.GetResult(context.Background() &result); err != nil {
    return nil, err 
}
``` 

**(2.1) `GetResult(ctx context.Context, run WorkflowRun, valuePtr interface{}) error`**

Combined with (2), adding an additional shortcut to get the update outcome directly compresses two calls into one. This is based on the assumption that in most cases, users care about the result more than anything else from the `WorkflowUpdateHandle` (but they could still obtain it, if needed).

```go
var result any
if err := upd.GetResult(context.Background(), run, &result); err != nil {
    return nil, err 
}
``` 

**(3) `Handle(run WorkflowRun) WorkflowUpdateHandle`**

Removes the error check from (2). If the user passes in `nil`, it would panic (akin to a NPE).

```go
var result any
if err := upd.Handle(run).Get(context.Background(), &result); err != nil {
    return nil, err 
}
``` 

---

Essentially it comes down to safety vs convenience. Since all users would benefit from less verbose code, I think the question is how likely the user would make the mistake to request the result too early.

internal/client.go
stephanos: Also, the name was chosen to allow more than just Update operations in the future. For instance a `SignalWorkflowOperation`.

internal/internal_workflow_client.go
```diff
@@ -1617,27 +1605,35 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+		} else if err != nil {
+			return nil, err
+		} else {
+			runID = response.RunId
+		}
 	} else {
+		response, err := w.executeWorkflowWithOperation(grpcCtx, startRequest, in.Options.WorkflowOperation)
+		if err != nil {
+			return nil, err
+		}
```
stephanos, line 1636: With regards to user-initiated retries, there is a catch here. Since the `startRequest` is re-created for each `ExecuteWorkflow` call - unless the user sets their own RequestID - the generated RequestID will be different each time. So in order to make the retry work, the user would need to use the UseExisting workflow id conflict policy. This will need to be documented.

internal/internal_workflow_client.go
```diff
@@ -1658,9 +1654,119 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+			}
+
+			resp := multiResp.Responses[i].Response
+			if opResp, ok := resp.(*workflowservice.ExecuteMultiOperationResponse_Response_UpdateWorkflow); ok {
+				updOperation.WorkflowUpdateHandle, err = w.updateHandleFromResponse(
+					context.Background(),
+					enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_UNSPECIFIED,
+					opResp.UpdateWorkflow)
+				if err != nil {
+					return nil, fmt.Errorf("%w: %w", startErr, err)
```
stephanos, line 1760: This will only happen if the server returns an unknown outcome type.

internal/client.go
```diff
@@ -1003,6 +1024,26 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+	return &UpdateWorkflowOperation{input: input}, nil
+}
+
+// Get blocks on the outcome of the update *after* it was executed via Client.ExecuteWorkflow.
+func (op *UpdateWorkflowOperation) Get(ctx context.Context, valuePtr interface{}) error {
+	if op.WorkflowUpdateHandle == nil {
+		return fmt.Errorf("update operation has not been executed")
+	}
+	return op.WorkflowUpdateHandle.Get(ctx, valuePtr)
+}
```
stephanos, line 1043: I got here after some trial and error. Options I've considered:

**(1) `Get(run WorkflowRun) (WorkflowUpdateHandle, error)`**

The safest but most verbose option. The `run` parameter can be used to ensure the function is invoked _after_ `ExecuteWorkflow`. It could also verify that the update handle's WorkflowId and RunId match the run's (which doesn't 100% guarantee that they were used together, but it's probably good enough). But all that requires an extra error check that is _only_ there to prevent a mistake by the user, not to surface an actual client/server error. So users should be able to catch this in testing already.

```go
updHandle, err := upd.Get(run)
if err != nil {
    return nil, err
}

var result any
if err := updHandle.GetResult(context.Background() &result); err != nil {
    return nil, err 
}
``` 

**(1.1) `GetResult(ctx context.Context, run WorkflowRun, valuePtr interface{}) error`**

Combined with (1), adding an additional shortcut to get the update outcome directly compresses two calls into one. This is based on the assumption that in most cases, users care about the result more than anything else from the `WorkflowUpdateHandle` (but they could still obtain it, if needed).

```go
var result any
if err := upd.GetResult(context.Background(), run, &result); err != nil {
    return nil, err 
}
``` 

**(2) `Handle(run WorkflowRun) WorkflowUpdateHandle`**

Removes the error check from (1). If the user passes in `nil`, it would panic (akin to a NPE since this is a user error). If the WorkflowId/RunId don't match, it could panic/ignore that (far from ideal). Overall my least favorite option.

```go
var result any
if err := upd.Handle(run).Get(context.Background(), &result); err != nil {
    return nil, err 
}
``` 

**(3) `Get(ctx context.Context, valuePtr interface{}) error`** (as written here)

When the operation embeds `WorkflowUpdateHandle`, the extra step to obtain the handle is removed entirely. While this is the most convenient option, it also means that there is a risk the user invokes the handle's functions _before_ executing the operation (however, if `Get` is invoked too early, it returns the appropriate error - the same cannot be said for WorkflowID and RunID, which would be empty).

```go
var result any
if err := upd.Get(context.Background(), &result); err != nil {
    return nil, err 
}
``` 

---


(1) is what we discussed in our design sessions. After implementing it, it felt unnecessarily complex/verbose to me compared to (4). But I do see the risk that introduces, too. Tell me what you think.

internal/internal_workflow_client.go
```diff
@@ -1658,9 +1654,119 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+	} else {
+		return nil, err
+	}
+
+	// extract errors or responses per operation
+	var startErr error
+	var abortedErr *serviceerror.MultiOperationAborted
+	var startResp *workflowservice.StartWorkflowExecutionResponse
+	for i, opReq := range multiRequest.Operations {
+		switch opReq.Operation.(type) {
```
cretz, line 1726: You add a default in your other type switches (pedantic though, I don't mind if you don't here)

internal/internal_workflow_client.go
cretz: IMO we should make it clear on the operations field that use of the field changes the default already-started behavior of the SDK (today the default behavior is to not fail if already started, but setting this field means it will fail, correct?)

client/client.go
```diff
@@ -920,6 +928,11 @@ type MetricsTimer = metrics.Timer
 // MetricsNopHandler is a noop handler that does nothing with the metrics.
 var MetricsNopHandler = metrics.NopHandler
 
+// PrepareUpdateWorkflowOperation returns an UpdateWorkflowOperation that can be used to perform Update-with-Start,
+// or an error in case the update operation is invalid. After executing Client.ExecuteWorkflow with the
+// UpdateWorkflowOperation in the start options, the update result can be obtained.
+var PrepareUpdateWorkflowOperation = internal.PrepareUpdateWorkflowOperation
```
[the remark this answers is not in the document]
stephanos: I had this feedback from Drew and it made sense to me:

> Also I slightly prefer Prepare as was originally sketched, as it implies that another step must follow it and gets developers to look twice at the comment above the method.  More clearly differentiates itself from a regular operation whereas New didn't stand out to me.

internal/internal_workflow_client.go
```diff
@@ -179,6 +182,7 @@ type (
 		dataConverter    converter.DataConverter
 		failureConverter converter.FailureConverter
 		registry         *registry
+		startRequest     *workflowservice.StartWorkflowExecutionRequest
```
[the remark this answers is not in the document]
stephanos: Oh, I forgot to remove that again. That was used in an earlier draft to match the operation to the correct workflow run (which doesn't work this way).

internal/internal_workflow_client.go
stephanos: Right; that's because this switch operates on the previously assembled operations (not the responses). So it would fall into the default if there's a bug in the function earlier. But I can still see it being useful.

internal/internal_workflow_client.go
```diff
@@ -1658,9 +1654,119 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+			resp := multiResp.Responses[i].Response
+			if opResp, ok := resp.(*workflowservice.ExecuteMultiOperationResponse_Response_StartWorkflow); ok {
+				startResp = opResp.StartWorkflow
+			} else {
+				panic(fmt.Sprintf("StartOperation response has the wrong type: %T", resp))
+			}
+		case *workflowservice.ExecuteMultiOperationRequest_Operation_UpdateWorkflow:
+			updOperation := operation.(*UpdateWorkflowOperation)
+
+			if multiErr != nil {
```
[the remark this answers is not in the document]
stephanos: Yeah, I had considered using two loops: one for errors and one for responses instead of mixing them together. Maybe I'll do that instead.

internal/internal_workflow_client.go
stephanos: Yes, setting the `WorkflowExecutionErrorWhenAlreadyStarted` to true will make it fail. I added a comment to that field, but I agree, adding it to the new field also is helpful 👍

internal/internal_workflow_client.go
```diff
@@ -1658,9 +1654,119 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+	case *UpdateWorkflowOperation:
+		if t.WorkflowUpdateHandle != nil {
+			return nil, fmt.Errorf("%w: was already executed", errInvalidWorkflowOperation)
+		}
+
+		updateReq, err := w.createUpdateWorkflowRequest(ctx, t.input)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %w", errInvalidWorkflowOperation, err)
+		}
+		if updateReq.WorkflowExecution.WorkflowId == "" {
```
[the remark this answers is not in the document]
stephanos: Right; the Server will reject the request if the Workflow IDs don't match. I can add client-side validation, too? The parameters to add client-side validation as well isn't always clear to me.

internal/internal_workflow_client.go
```diff
@@ -1929,44 +2011,74 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
...
+		return nil, errors.New("WaitForStage WorkflowUpdateStageAdmitted is not supported")
+	}
+
+	return &ClientUpdateWorkflowInput{
+		UpdateID:            updateID,
+		WorkflowID:          options.WorkflowID,
+		UpdateName:          options.UpdateName,
+		Args:                options.Args,
+		RunID:               options.RunID,
+		FirstExecutionRunID: options.FirstExecutionRunID,
```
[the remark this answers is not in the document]
stephanos: That's true. I can add additional client-side validation for that.

internal/client.go
stephanos: > And I think we should consider doing signal operation right now and deprecating signal with start.

You mean as part of this PR?

internal/internal_workflow_client.go
stephanos: Just saying it out loud: I assume there's no reason to go all in with the default and set the `WorkflowIdConflictPolicy` to `UseExisting` for Update-with-Start to match today's default of ExecuteWorkflow? (in that world, we'd need to error on certain invalid combinations of `WorkflowExecutionErrorWhenAlreadyStarted` and `WorkflowIdConflictPolicy`)

client/client.go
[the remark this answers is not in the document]
cretz: I think in Go, "New" makes more sense, even if it may not in other languages, but I do not have a strong opinion, I too am fine with either

internal/internal_workflow_client.go
```diff
@@ -1658,9 +1654,119 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+	startOp := &workflowservice.ExecuteMultiOperationRequest_Operation{
+		Operation: &workflowservice.ExecuteMultiOperationRequest_Operation_StartWorkflow{
+			StartWorkflow: startRequest,
+		},
+	}
+
+	var withStartOp *workflowservice.ExecuteMultiOperationRequest_Operation
+	switch t := operation.(type) {
+	case *UpdateWorkflowOperation:
+		if t.WorkflowUpdateHandle != nil {
```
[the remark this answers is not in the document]
stephanos: 🤦 Right, I see that now.

internal/internal_workflow_client.go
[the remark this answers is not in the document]
cretz: Server side rejection works for me. There is not a clear rule on client-side vs server side. While you may think we want to catch it early, server validation logic may change. _But_ we should at least document the behavior on client side if surprising (not sure this is surprising though). Users do need to be made aware they don't need to provide a workflow execution for this update (in fact, they _shouldn't_).

internal/client.go
cretz: > You mean as part of this PR?

Nah, but soon IMO.

> What's your reasoning?

Because in general the idea of having additional operations on a workflow makes more sense than a single additional operation. Users of our API (now and HTTP in the future) are lucky enough to have this representation, we should too. The same reasons it is `repeated` in proto are the reasons for making it a slice here. We could have done composite there too, why did we not?

internal/internal_workflow_client.go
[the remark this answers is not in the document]
stephanos: Split it into two loops, I think it's clearer now.

internal/client.go
stephanos: Okay, for some reason I was approached this with the single-threaded "inside a workflow" mindset. The "wait for handle" use case makes perfect sense. I made it concurrency-safe now.

internal/client.go
```diff
@@ -713,6 +725,22 @@ type (
...
+
+	// UpdateWorkflowOperation is used to perform Update-with-Start.
+	// See PrepareUpdateWorkflowOperation for details.
+	UpdateWorkflowOperation struct {
+		input     *ClientUpdateWorkflowInput
+		executing atomic.Bool
+		done      atomic.Bool
+		doneCh    chan struct{}
+		handle    WorkflowUpdateHandle
+		err       error
```
cretz, line 741: May want a comment making it clear these fields should not be accessed until after `doneCh` is received from

internal/client.go
```diff
@@ -1003,6 +1031,54 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+		return nil, errors.New("FirstExecutionRunID is not allowed to be used on a prepared update operation")
+	}
+	return &UpdateWorkflowOperation{input: input, doneCh: make(chan struct{})}, nil
+}
+
+// Get blocks until a server response has been received; or the context deadline is exceeded.
+func (op *UpdateWorkflowOperation) Get(ctx context.Context) (WorkflowUpdateHandle, error) {
+	if op.done.Load() {
+		return op.handle, op.err
+	}
```
cretz, line 1054:
```suggestion
```

Not needed IMO (neither is the atomic bool)

internal/client.go
```diff
@@ -1003,6 +1031,54 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+	select {
+	case <-op.doneCh:
+		return op.handle, op.err
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
+}
+
+func (op *UpdateWorkflowOperation) execute() error {
```
cretz, line 1063:
```suggestion
func (op *UpdateWorkflowOperation) markExecuted() error {
```

Pedantic I know, can ignore. Really, probably clearer if the field name is something like "used" or something

internal/client.go
```diff
@@ -1003,6 +1031,54 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+	if op.executing.Swap(true) {
+		return fmt.Errorf("was already executed")
+	}
+	return nil
+}
+
+func (op *UpdateWorkflowOperation) set(handle WorkflowUpdateHandle, err error) {
+	if op.done.Load() {
+		panic("handle has already been set")
+	}
```
cretz, line 1073: Not needed in this private method IMO where we know we only set after we mark `execute()`. If you must have this check, consider simply checking whether executing is true by this time (but I don't think it's needed, we can trust ourselves to set it once.

cretz (commented):
Mostly LGTM. I personally like the "additional options" (plural) approach, but everything else seems right to me

internal/internal_workflow_client.go
cretz: To clarify, if default behavior _without_ setting `WorkflowOperation` is one thing by default on already-exists and default behavior _with_ setting `WorkflowOperation` is another thing by default, we need to document on the field the difference in default behavior.

internal/client.go
Sushisource: +1, makes sense for this to be a list I think.

internal/client.go
Sushisource: Additional works for me, or `WithStartWorkflowOperation` which mimics the existing signal with start names.

internal/internal_workflow_client.go
```diff
@@ -75,7 +75,10 @@ const (
 	pollUpdateTimeout = 60 * time.Second
 )
 
-var maxListArchivedWorkflowTimeout = time.Minute * 3
+var (
+	maxListArchivedWorkflowTimeout = time.Minute * 3
+	errInvalidWorkflowOperation    = fmt.Errorf("invalid WorkflowOperation")
+)
```
Sushisource, line 81: These should probably be consts

internal/internal_workflow_client.go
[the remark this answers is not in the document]
Sushisource: +1 to catching server side.

internal/internal_workflow_client.go
```diff
@@ -1661,6 +1656,120 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+			withStartOp,
+		},
+	}
+	multiResp, err := w.client.workflowService.ExecuteMultiOperation(ctx, &multiRequest)
+
+	var multiErr *serviceerror.MultiOperationExecution
+	if errors.As(err, &multiErr) {
+		if len(multiErr.OperationErrors()) != len(multiRequest.Operations) {
+			panic(fmt.Sprintf("MultiOperation error response from server is incomplete: %v instead of %v",
+				len(multiErr.OperationErrors()), len(multiRequest.Operations)))
```
Sushisource, line 1707: Not sure this should be a panic. I get that it's an invariant the server is supposed to uphold, but from the client's perspective it's just bad data returned from the server. Maybe return an error here (with a bit more explanation that server is misbehaving). Same with the following ones.

internal/client.go
drewhoskins-temporal: I'm with Stephan on adding a composite operation later in the relatively small chance we build that.
* That's "gracefully revealed complexity" which makes "easy things easy and hard things possible." 
* I don't want to mislead developers into thinking it's supported and waste their time integrating only to learn via runtime exceptions via testing that it doesn't work.
* There's no real product interest in supporting a list of operations for the time being, and it's not in danger of being added to our roadmap even in 2025.

internal/internal_workflow_client.go
stephanos: Right; I can make `maxListArchivedWorkflowTimeout` a const, but calling `fmt.Errorf` in Go is not allowed for a const.

internal/client.go
Sushisource: On further thought I agree. There's more types of future invalid combinations we can prevent by encoding composition into the type system (more so in not-Golang, but, still).

However I agree the mismatch with the API is weird. Ideally we'd change the API.

internal/client.go
drewhoskins-temporal: `WithStartWorkflowOperation` and `StartWorkflowAdditionalOperation` both fine with me.

I like "With Start" as it mirrors how we talk about it in the docs and conceptual guides.

internal/client.go
drewhoskins-temporal: Re: supporting signal-with-start, it might be worth doing upfront just to validate that our design is properly able to handle both cases.  But haven't thought it through and up to you.

internal/client.go
stephanos: Ah, right; when the channel is closed, it will unblock immediately. I removed `done` now.

internal/internal_workflow_client.go
Quinn-With-Two-Ns: +1 to being an error. I think the SDK should handle bad data from the server as gracefully as it can

client/client.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I favor new since prepare ,to me, implies something on the server is being prepared and may lead users to believe they need to "unprepared" if they don't use an operation

internal/client.go
stephanos: Hm, as I was removing `done`, I also extended the integration test exercising the concurrency and it's failing now.

internal/client.go
cretz: :+1: I personally like the collection (in all APIs), but don't have a strong opinion, this is fine too. I do think we should consider our HTTP/gRPC API user facing and we should have it as similar as the SDK APIs, but meh.

I do still also prefer the term "additional" operations. But even if that is also not what y'all want, I would recommend at least removing the word `Workflow` from the field name since it's in workflow options, but don't have a strong opinion there either.

internal/client.go
```diff
@@ -643,9 +644,20 @@ type (
...
 		WorkflowIDConflictPolicy enumspb.WorkflowIdConflictPolicy
 
+		// WithStartWorkflowOperation - Operation to execute with Workflow Start.
+		// For example, see NewUpdateWorkflowOperation to perform Update-with-Start. Note that if the workflow is
+		// already running and WorkflowIDConflictPolicy is set to UseExisting, the start is skipped and only the
+		// operation is executed. If instead the policy is set to Fail (the default), nothing is executed and
+		// an error will be returned (i.e. the option WorkflowExecutionErrorWhenAlreadyStarted is ignored).
+		//
+		// Optional: defaults to nil.
+		WithStartWorkflowOperation WithStartWorkflowOperation
```
Quinn-With-Two-Ns, line 654: What happens if this is set when a user calls `SignalWithStartWorkflow`?

stephanos: It will be ignored. Would a comment on `SignalWithStartWorkflow` or `WithStartOperation` (or both) be best here?

Quinn-With-Two-Ns: I think we should document it in `WithStartOperation` and error if it is set when calling `SignalWithStartWorkflow`. You can also document it in `SignalWithStartWorkflow` but I don't think that is required

internal/internal_workflow_client.go
```diff
@@ -1661,6 +1658,120 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+		case *workflowservice.ExecuteMultiOperationRequest_Operation_StartWorkflow:
+			if opResp, ok := resp.(*workflowservice.ExecuteMultiOperationResponse_Response_StartWorkflow); ok {
+				startResp = opResp.StartWorkflow
+			} else {
+				return nil, fmt.Errorf("%w: StartWorkflow response has the wrong type %T", errInvalidServerResponse, resp)
+			}
+		case *workflowservice.ExecuteMultiOperationRequest_Operation_UpdateWorkflow:
+			if opResp, ok := resp.(*workflowservice.ExecuteMultiOperationResponse_Response_UpdateWorkflow); ok {
+				handle, err := w.updateHandleFromResponse(
+					context.Background(),
```
Quinn-With-Two-Ns, line 1757: You need to use the proper context here

stephanos: Ah, good catch 😬

internal/internal_workflow_client.go
```diff
@@ -1661,6 +1661,120 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
+	}
+
+	var withStartOp *workflowservice.ExecuteMultiOperationRequest_Operation
+	switch t := operation.(type) {
+	case *UpdateWorkflowOperation:
+		if err := t.markExecuted(); err != nil {
+			return nil, fmt.Errorf("%w: %w", errInvalidWorkflowOperation, err)
+		}
+
+		updateReq, err := w.createUpdateWorkflowRequest(ctx, t.input)
```
Quinn-With-Two-Ns, line 1682: What header does the update request end up getting when using the tracing interceptor? I wonder how this multi op will display in traces? @cretz I think Python modled signal with start a similar way how do the traces appear there?

internal/client.go
stephanos: > And I think we should consider doing signal operation right now and deprecating signal with start.

Until the server's MultiOperation API supports Signal, we could do this right now:

```
func (wc *WorkflowClient) ExecuteWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (WorkflowRun, error) {
	// TODO: use MultiOperation API once there is server-side support
	if op, ok := options.WithStartOperation.(*SignalWorkflowOperation); ok {
		return wc.SignalWithStartWorkflow(ctx, options.ID, op.signalName, op.arg, options, workflow, args...)
	}
```

The two downsides I'm seeing are (1) anyone using interceptors might be surprised and (2) error messages might change once we move to the MultiOperation API (behavior should be identical AFAICT).

cretz: Yes, my only concern was consistency from user POV. A user should be able to signal with start, update with start, signal + update with start, etc the same way. We don't have to get there in this PR, just a general thought about proper consistent APIs.

internal/internal_workflow_client.go
cretz: Yeah signal with start is a single call with a single header that I'm guessing becomes separate headers on the separate start/signal events worker side. Here I would expect we'd populate both the start and update headers with the inbound headers and they would remain/appear separate in the separate events worker side.

internal/client.go
stephanos: FWIW, I prototyped the Signal work and it could look like this:

```go
        sig, _ := client.NewSignalWorkflowOperation("SIGNAL", 1)
	wfHandle, err := c.ExecuteWorkflow(
		context.Background(),
		client.StartWorkflowOptions{
			ID:                       wfID,
			TaskQueue:                taskQueue,
			WithStartOperation:       sig,
			WorkflowIDConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
		},
		Workflow,
	)
```

cretz: Not the biggest fan of "WithStart" over "Additional" and ideally the top-level `NewSignalWorkflowOperation` would be named to make it clear it's only for this use case and not general purpose.

internal/internal_workflow_client.go
stephanos: I see there are two tracing methodologies: OpenTelemetry and OpenTracing. Would this be needed for both or just one (given OpenTracing's standing compared to OTEL)?

stephanos: Also (I could be wrong), I don't see a tracing integration for Update itself anywhere.

cretz: There is a common tracing interceptor they all use, it's just a different tracer they provide (they all ultimately call https://pkg.go.dev/go.temporal.io/sdk/interceptor#NewTracingInterceptor).

Upon review, you are correct, we didn't properly wire `ClientOutboundInterceptor.UpdateWorkflow` in the tracing interceptor to populate header or `WorkflowInboundInterceptor.ExecuteUpdate` to rehydrate span from header. We need to fix that, it does work in other SDKs. This can be fixed separately (I have opened https://github.com/temporalio/sdk-go/issues/1588). It does look like we did context propagation though (a prereq).

I do still think we need to make sure that context propagation works (i.e. that `headerPropagated` is being called outbound) which seems to be the case so we're probably good here until that other issue is tackled. If you'd like, you can add an integration test like `TestContextPropagator` for multi-op.

internal/client.go
cretz: The issue is the name of the update operation is unrelated to this use, it's named generically as if it's used in many places

client/client.go
```diff
@@ -920,6 +928,13 @@ type MetricsTimer = metrics.Timer
 // MetricsNopHandler is a noop handler that does nothing with the metrics.
 var MetricsNopHandler = metrics.NopHandler
 
+// NewUpdateWorkflowOperation returns an UpdateWorkflowOperation to perform Update-with-Start,
+// or an error in case the update operation is invalid. After executing Client.ExecuteWorkflow with the
+// UpdateWorkflowOperation in the start options, the update result can be obtained.
+func NewUpdateWorkflowOperation(options UpdateWorkflowOptions) (*UpdateWorkflowOperation, error) {
```
cretz, line 934: Want to close on this concern. This is a generically named thing and I think it'd be more reasonable to be named similar to the field/type how/where it is used. This name makes one think this is the common update workflow operation function.

Quinn-With-Two-Ns: If we did do multiop in the future would we reuse `UpdateWorkflowOperation`? I don't see any reason why we wouldn't so I have no issue with the current name.

internal/internal_workflow_client.go
```diff
@@ -769,11 +775,12 @@ type UpdateWorkflowOptions struct {
 	UpdateID string
 
 	// WorkflowID is a required field indicating the workflow which should be
-	// updated.
+	// updated. However, it is optional when using UpdateWorkflowOperation.
 	WorkflowID string
 
 	// RunID is an optional field used to identify a specific run of the target
 	// workflow.  If RunID is not provided the latest run will be used.
+	// Note that it is incompatible with UpdateWorkflowOperation.
```
Quinn-With-Two-Ns, line 783: If you have to say a lot of options don't actually work it may be worth creating a new options struct

client/client.go
cretz: I don't think we should assume that would be the case, but even if it was, this is too generically named even for that IMO.

internal/internal_workflow_client_test.go
```diff
@@ -976,6 +976,161 @@ func (s *workflowRunSuite) TestGetWorkflowNoExtantWorkflowAndNoRunId() {
...
+	s.Nil(err)
+
+	_, err = s.workflowClient.ExecuteWorkflow(
+		context.Background(),
+		StartWorkflowOptions{
+			ID:                 workflowID,
+			TaskQueue:          taskqueue,
+			WithStartOperation: updOp,
+		}, workflowType,
+	)
```
stephanos, line 1020: @cretz @Quinn-With-Two-Ns 

With the naming discussion in mind, what do y'all think about this direction?

```go
op := client.UpdateWithStartWorkflowOperation{
	UpdateName:   "UPDATE",
	Args:         []interface{}{5},
	WaitForStage: client.WorkflowUpdateStageCompleted,
}

wfHandle, err := c.ExecuteWorkflow(
	context.Background(),
	client.StartWorkflowOptions{
		...
		WithStartOperation:       op,
	},
	Workflow,
)
if err != nil {
	panic(err)
}

var updateResult int
updHandle, err := op.Get(context.Background())
if err != nil {
	panic(err)
}
```

Benefits:

- removes need for constructor
- easier to find from option `WithStartOperation`'s type: just look for implementations in IDE
- since the operation is just a struct, there's no risk assuming it performs an action of any kind
- one less error handling check
- `UpdateWithStartWorkflowOperation` can be a copy of `UpdateWorkflowOptions` (and omit invalid fields)

Downsides:

- extra type to maintain / no re-use (to keep in sync with `UpdateWorkflowOptions`) [could also embed instead]
- validation error performed at execution stage, ie later

cretz: This looks good to me. Assuming you are maintaining private state on that struct though, make sure to mention that 1) fields are not expected to be mutated after it is in use (and either have no effect or would would be racy), and 2) that the struct is only usable once.

drewhoskins-temporal: If we go with this approach and naming, I recommend renaming `Args` as `UpdateArgs` to make it clear _which_ args it is.

internal/client.go
```diff
@@ -1003,6 +1034,47 @@ func DialCloudOperationsClient(ctx context.Context, options CloudOperationsClien
...
+	input, err := createUpdateWorkflowInput(options)
+	if err != nil {
+		return nil, err
+	}
+	if options.RunID != "" {
+		return nil, errors.New("RunID is not allowed to be used on an UpdateWorkflowOperation")
+	}
+	if options.FirstExecutionRunID != "" {
+		return nil, errors.New("FirstExecutionRunID is not allowed to be used on an UpdateWorkflowOperation")
+	}
```
drewhoskins-temporal, line 1049: These error messages could be more helpful.  The issue here seems to be that it doesn't make sense to pass in a run ID for a workflow update for which the workflow does not yet exist, in which case, say that rather than making a very literal code-oriented error message.
Also, if my reasoning is correct, these validations seem more appropriate in the start workflow call, where we actually know that's how this operation will be used.

internal/internal_workflow_client_test.go
Quinn-With-Two-Ns: Yeah there is a another downside as Chad pointed out that this lets the user modify struct field after call which is not valid and it is not consistent with the rest of the Go SDK. I prefer the current approach maybe just renaming `NewUpdateWorkflowOperation` to `NewUpdateWorkflowWithStartOperation` and take a `UpdateWorkflowWithStartOperationOptions` as a parameter.

stephanos: Notes/input from a meeting just now:

- it's advantageous to run the Update validation from `ExecuteWorkflow` since then the interceptors are in the loop
- using a function instead of a struct here is more idiomatic Go
- UpdateWithStart works without "workflow" suffix as it's a proper name (will be used through the product/docs)
- using the existing `UpdateWorkflowOptions` makes it easier for users to migrate from Update to Update-with-Start, but risks that users use invalid parameters only to find out at runtime - whereas using a dedicated struct makes migrating harder but is safer

cretz: > UpdateWithStart works without "workflow" suffix as it's a proper name (will be used through the product/docs)

For the same reasons we call it `ExecuteWorkflow` and not just `Execute`, and `UpdateWorkflow` and not `Update`, I think it's important to use the word workflow when on a generic thing or top-level and talking about something specific to a workflow. We do it everywhere else and I don't think it's worth saving a word to become inconsistent in this special case.

cretz (approved):

Quinn-With-Two-Ns (approved):

Issue #1580, opened 2024-08-06
Do not warn on unfinished handlers if the workflow has failed


Quinn-With-Two-Ns:
It was decided we would not warn on unfinished handlers if the workflow has failed.

PR #1581, opened 2024-08-06
Do not warn on unfinished handlers if workflow failed

Quinn-With-Two-Ns:
Do not warn on unfinished handlers if workflow failed

closes https://github.com/temporalio/sdk-go/issues/1580

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -682,7 +682,7 @@ func executeDispatcher(ctx Context, dispatcher dispatcher, timeout time.Duration
 	if len(us) > 0 {
 		env.GetLogger().Warn("Workflow has unhandled signals", "SignalNames", us)
 	}
-	//
+	// Warn if there are any update handlers still running
 	type warnUpdate struct {
 		Name string `json:"name"`
 		ID   string `json:"id"`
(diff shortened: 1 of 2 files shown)


cretz (approved):

internal/internal_workflow.go
```diff
@@ -696,7 +696,11 @@ func executeDispatcher(ctx Context, dispatcher dispatcher, timeout time.Duration
 			})
 		}
 	}
-	if len(updatesToWarn) > 0 {
+
+	// Verify that the workflow did not fail. If it did we will not warn about unhandled updates.
+	var canceledErr *CanceledError
+	var contErr *ContinueAsNewError
+	if len(updatesToWarn) > 0 && (rp.error == nil || errors.As(rp.error, &canceledErr) || errors.As(rp.error, &contErr)) {
 		env.GetLogger().Warn(unhandledUpdateWarningMessage, "Updates", updatesToWarn)
```
Sushisource, line 704: You know we never discussed this but perhaps it makes sense to log them at debug level for workflow failing case?

Not a big deal either way.

Sushisource (approved):

internal/internal_workflow.go
Quinn-With-Two-Ns: Hmm I think I would punt on this for no. I suspect we will be doing more work on warning on dangling "things" (like activities) in workflow and can resist then

PR #1583, opened 2024-08-08
Switch omes to post-merge

Sushisource:


.github/workflows/omes.yml
```diff
@@ -0,0 +1,17 @@
+name: Omes Testing
+on:
+  push:
+    branches:
+      - master
```
dandavison, line 5: The GitHub jobs suggest [something more complicated](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges). But perhaps `on push master` is a superset of `close and merge PR`. (Also the GitHub docs thing looks like it would trigger on a merge to non-master/main).

Sushisource: This is what we use virtually everywhere for workflows like this

dandavison (approved):
LGTM!

Quinn-With-Two-Ns (approved):

Issue #1588, opened 2024-08-13
Support updates in tracing interceptor


cretz:
**Describe the solution you'd like**

Need to support `ClientOutboundInterceptor.UpdateWorkflow` in the tracing interceptor to populate header and `WorkflowInboundInterceptor.ExecuteUpdate` (and validate too maybe) to put span on context.

PR #1595, opened 2024-08-16
Support updates in tracing interceptor

Quinn-With-Two-Ns:
Support updates in tracing interceptor

closes https://github.com/temporalio/sdk-go/issues/1588

### DIFF
--- a/interceptor/tracing_interceptor.go
+++ b/interceptor/tracing_interceptor.go
@@ -551,7 +551,7 @@ func (t *tracingWorkflowInboundInterceptor) ValidateUpdate(
 	in *UpdateInput,
 ) error {
 	// Only add tracing if enabled and not replaying
-	if t.root.options.DisableUpdateTracing || workflow.IsReplaying(ctx) {
+	if t.root.options.DisableUpdateTracing {
 		return t.Next.ValidateUpdate(ctx, in)
 	}
 	// Start span reading from header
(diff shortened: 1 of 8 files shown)


interceptor/tracing_interceptor.go
```diff
@@ -515,6 +546,77 @@ func (t *tracingWorkflowInboundInterceptor) HandleQuery(
...
+) (interface{}, error) {
+	// Only add tracing if enabled and not replaying
+	if t.root.options.DisableUpdateTracing || workflow.IsReplaying(ctx) {
+		return t.Next.ExecuteUpdate(ctx, in)
+	}
+	// Start span reading from header
+	info := workflow.GetInfo(ctx)
+	currentUpdateInfo := workflow.GetCurrentUpdateInfo(ctx)
+	span, ctx, err := t.root.startSpanFromWorkflowContext(ctx, &TracerStartSpanOptions{
+		Operation: "ExecuteUpdate",
```
cretz, line 598: Hrmm, I see we didn't call this method `HandleUpdate` like we have `HandleSignal` and `HandleQuery` (and as we may have in other SDKs), but I wonder if we still want the span name to match the other SDKs or match the interceptor method.

Quinn-With-Two-Ns: I'm fine matching other SDK's, I don't think the span name needs to match the interceptor name

cretz (approved):

PR #1597, opened 2024-08-16
Experimental user metadata and workflow metadata query support

### LINKED ISSUE #1565: User metadata and workflow metadata query

See https://github.com/temporalio/features/issues/486. This can only be worked on in Go at this time because CI is testing against unstable server currently.

cretz:
## What was changed

* Added `StartWorkflowOptions.StaticSummary` and `StartWorkflowOptions.StaticDetails` fields that are populated on workflow start event and execution config
* Added `workflow.SetCurrentDetails` and `workflow.GetCurrentDetails` functions that set current details available on workflow metadata query
* Added `workflow.NewTimerWithOptions` that accepts a `TimerOptions` which has a `Summary` to be put on the timer event
  * Note, this required a new outbound interceptor method and may have broken compile for PHP SDK internalbindings (by intention)
* Added `workflow.SetQueryHandlerWithOptions` that accepts a `QueryHandlerOptions` which has a `Description`
* Added `workflow.GetSignalChannelWithOptions` that accepts a `SignalChannelOptions` which has a `Description`
* Added support for `__temporal_workflow_metadata` query that returns query, signal, and update definitions and the current details

Note, some things are not supported:

* Child workflow user metadata - https://github.com/temporalio/temporal/issues/6412
* Scheduled workflow user metadata - https://github.com/temporalio/temporal/issues/6413
* Continue as new user metadata - known and to be discussed
* Set summaries for internal timers - #1602

## Checklist

1. Closes #1565


internal/internal_schedule_client.go
```diff
@@ -637,6 +637,11 @@ func convertToPBScheduleAction(
 			return nil, err
 		}
 
+		userMetadata, err := buildUserMetadata(action.Summary, action.Details, dataConverter)
```
cretz, line 640: TODO: Need a test for this

internal/interceptor.go
```diff
@@ -231,6 +231,9 @@ type WorkflowOutboundInterceptor interface {
 	// NewTimer intercepts workflow.NewTimer.
 	NewTimer(ctx Context, d time.Duration) Future
 
+	// NewTimer intercepts workflow.NewTimerWithOptions.
+	NewTimerWithOptions(ctx Context, d time.Duration, options TimerOptions) Future
```
cretz, line 235: Based on available options, a new outbound method seemed like the best approach here. Should be safe since we require the base to be embedded.

internal/internal_command_state_machine.go
```diff
@@ -1556,8 +1565,19 @@ func (h *commandsHelper) getSignalID(initiatedEventID int64) string {
...
+func (h *commandsHelper) startTimer(
+	attributes *commandpb.StartTimerCommandAttributes,
+	options TimerOptions,
+	dc converter.DataConverter,
+) commandStateMachine {
+	var summary *commonpb.Payload
+	if options.Summary != "" {
+		var err error
+		if summary, err = dc.ToPayload(options.Summary); err != nil {
+			panic(err)
```
cretz, line 1577: This should be rare (it's just serializing string), but there is precedent with other `commandsHelper` calls panicking if they can't convert.

internal/internal_workflow.go
```diff
@@ -1567,6 +1582,79 @@ func GetUnhandledSignalNames(ctx Context) []string {
...
+	for k := range eo.queryHandlers {
+		ret.Definition.QueryDefinitions = append(ret.Definition.QueryDefinitions, &sdk.WorkflowInteractionDefinition{
+			Name: k,
+			// TODO(cretz): Allow query descriptions?
+			// Description: ,
+		})
+	}
+	// Signals
+	// TODO(cretz): This is all signal channels asked for in workflow _and_ all ever sent, is that what we want? Or do
+	// we only want all ever asked for?
```
cretz, line 1632: Remove this TODO before merging if all agree what's here is acceptable

internal/internal_time.go
```diff
@@ -45,7 +45,7 @@ type (
 
 		// NewTimer - Creates a new timer that will fire callback after d(resolution is in seconds).
 		// The callback indicates the error(TimerCanceledError) if the timer is canceled.
-		NewTimer(d time.Duration, callback ResultHandler) *TimerID
+		NewTimer(d time.Duration, options TimerOptions, callback ResultHandler) *TimerID
```
cretz, line 48: This may break PHP SDK since it's exposed via `internalbindings` but unsure if they use it. This seemed like the best way and should cause a clear compile break.

internal/interceptor.go
```diff
@@ -231,6 +231,9 @@ type WorkflowOutboundInterceptor interface {
 	// NewTimer intercepts workflow.NewTimer.
 	NewTimer(ctx Context, d time.Duration) Future
 
+	// NewTimer intercepts workflow.NewTimerWithOptions.
```
Quinn-With-Two-Ns, line 234: Can other primitives support metadata like activities? Do we plan to add support for them?

internal/interceptor.go
```diff
@@ -231,6 +231,9 @@ type WorkflowOutboundInterceptor interface {
...
+	NewTimerWithOptions(ctx Context, d time.Duration, options TimerOptions) Future
+
```
Quinn-With-Two-Ns, line 236: Often the SDK will create internal timers, it would be nice if we can leverage metadata to give those timers appropriate names. Not saying you need to do that in this PR, but if we agree we should open an issue

internal/interceptor.go
cretz: Yes and yes! We intentionally decided that workflow summary/details and timer summary was the first use case. But the user metadata is on the command/event and not the attributes for this reason. Any UI/CLI-visible metadata you could ever need on an event we can look into adding.

internal/interceptor.go
cretz: Agreed. Will set reminder to open issue there.

internal/client.go
```diff
@@ -707,6 +710,23 @@ type (
...
+		Summary string
+
+		// Details - General details for this workflow execution that will appear in UI/CLI. This can be in Temporal
+		// markdown format and can span multiple lines. This is a fixed value on the workflow that cannot be updated.
+		// For details that can be updated, use SetCurrentDetails within the workflow.
+		//
+		// Optional: defaults to none/empty.
+		//
+		// NOTE: Experimental
+		Details string
```
Quinn-With-Two-Ns, line 728: > This is a fixed value on the workflow that cannot be updated. For details that can be updated, use SetCurrentDetails within the workflow.

It's kind of confusing `Details` works differently then search attributes and memo here. Is there a reason we decided to be incosistent here?

internal/internal_command_state_machine.go
```diff
@@ -385,11 +387,15 @@ func (h *commandsHelper) newRequestCancelNexusOperationStateMachine(attributes *
 	}
 }
 
-func (h *commandsHelper) newTimerCommandStateMachine(attributes *commandpb.StartTimerCommandAttributes) *timerCommandStateMachine {
+func (h *commandsHelper) newTimerCommandStateMachine(
+	attributes *commandpb.StartTimerCommandAttributes,
+	summary *commonpb.Payload,
```
Sushisource, line 392: imo this should just be `*UserMetadata`  directly. The state machine doesn't need to know or care about summary and later on if/when more is added this won't need to change.

internal/internal_time.go
Sushisource: I think they do but it should be fine

internal/internal_workflow.go
```diff
@@ -226,6 +228,8 @@ type (
 		// runningUpdatesHandles is a map of update handlers that are currently running.
 		runningUpdatesHandles map[string]UpdateInfo
 		VersioningIntent      VersioningIntent
+		// currentDetails is the user-set string returned on metadata query.
```
Sushisource, line 231:
```suggestion
		// currentDetails is the user-set string returned on metadata query as WorkflowMetadata.current_details
```
Nice and obvious

internal/internal_workflow.go
```diff
@@ -1567,6 +1582,79 @@ func GetUnhandledSignalNames(ctx Context) []string {
...
+			},
+		},
+		CurrentDetails: eo.currentDetails,
+	}
+	// Queries
+	for k := range eo.queryHandlers {
+		ret.Definition.QueryDefinitions = append(ret.Definition.QueryDefinitions, &sdk.WorkflowInteractionDefinition{
+			Name: k,
+			// TODO(cretz): Allow query descriptions?
+			// Description: ,
```
Sushisource, line 1627: Why not? (Applies to all)

internal/internal_workflow.go
Sushisource: I think it should be only asked for. It's entirely possible to miss-send and now it looks like you support something you don't.

internal/internal_workflow.go
```diff
@@ -1567,6 +1582,79 @@ func GetUnhandledSignalNames(ctx Context) []string {
...
+	}
+	// Updates
+	for k, v := range eo.updateHandlers {
+		ret.Definition.UpdateDefinitions = append(ret.Definition.UpdateDefinitions, &sdk.WorkflowInteractionDefinition{
+			Name:        k,
+			Description: v.description,
+		})
+	}
+	// Sort interaction definitions
+	sortWorkflowInteractionDefinitions(ret.Definition.QueryDefinitions)
```
Sushisource, line 1648: The API definition doesn't make any claims about sorting (though, perhaps it should, since it's useful for this to be stable).

internal/schedule_client.go
```diff
@@ -281,6 +281,23 @@ type (
 		//
 		// Deprecated - This is only for update of older search attributes. This may be removed in a future version.
 		UntypedSearchAttributes map[string]*commonpb.Payload
+
+		// Summary - Single-line summary for this workflow execution that will appear in UI/CLI. This can be in
+		// single-line Temporal markdown format.
```
Sushisource, line 286: None of these strings that reference this markdown format have info about what it is. Probably because it doesn't exist yet, but, if it does somewhere, link would be good.

Sushisource (commented):
I thought we had the ability to add this to child workflows straight away too?

internal/internal_command_state_machine.go
cretz: :+1: Agreed can change

internal/internal_workflow.go
cretz: This adds a bit more information to store, but it may be worth it here and I may just go ahead and add a `GetSignalChannelWithOptions` to capture that description too then.

internal/client.go
cretz: Can you clarify which difference? The fact that it's not arbitrary key-value or the fact that it's a string and not a payload.

It's not arbitrary key value because it was decided in the user metadata project not to support arbitrary key value for any user metadata but make explicit fields as needed (and we can add more on the proto as needed). It's a string because the UI has an expectation on this being a string for display.

internal/internal_workflow.go
cretz: Yeah, might as well add a `SetQueryHandlerWithOptions` and a `GetSignalChannelWithOptions` right now (was a bit concerned doing too much at once)

internal/internal_workflow.go
cretz: Agreed, we don't want any other sorting expectations to accidentally creep in here. I will make a note to update the API definition.

internal/schedule_client.go
cretz: Yeah, as this concept is stabilized (i.e. non-experimental) and we have something at docs.temporal.io, would definitely want to link it here.

cretz:
> I thought we had the ability to add this to child workflows straight away too?

Yes, will add

internal/client.go
Quinn-With-Two-Ns: The fact that it can't be updated

cretz: Ah, this is by intention based on the design discussions. We want mutable information to be available via query, the metadata is meant to be immutable.

cretz:
Actually I am not seeing child workflow as working. I put in a bunch of work to wire it up for child workflows then my integration tests show it didn't propagate. I will get with authors of https://github.com/temporalio/temporal/pull/5857, but in the meantime I am gutting the support I added for child workflows before committing the next round of updates.

internal/internal_workflow.go
cretz: Added

internal/interceptor.go
cretz: Opened #1602

internal/internal_command_state_machine.go
cretz: Changed

internal/internal_workflow.go
cretz: Done

internal/internal_workflow.go
cretz: Done

internal/internal_workflow.go
cretz: Done: https://github.com/temporalio/api/pull/445

internal/client.go
Quinn-With-Two-Ns: Forgive my ignorance, but do these two `Details` appear as separate in the UI?

cretz: I think they should, yes. We should discuss this actually.

So today we have fixed summary/details you set on a workflow. Then we have a wholly-unrelated concept of "current details" (was unable to think of a better name) that is queryable. But there has been discussion on making the query occur by default when you arrive  on the UI. There is no overlap or any relation between "fixed workflow start details" and "queryable workflow current details" except in the English term "details" we happened to have reused here.

It would be a much larger task to try and update the workflow one from the query. But one thing we can consider doing now that I am hearing we may issue the query upon page arrival, is we can default the current details to the workflow details one. The reason I am hesitant is because I don't want people to wonder why on "describe workflow" they don't get the current details.

Quinn-With-Two-Ns: Yeah i think I would make a stronger distinction in the SDK API that there are two details:
* One set on start that is fixed and does not require a worker to be running
* One that is set in the workflow that DOES require the workflow to be running

I think they deserve separate names in the SDK, maybe we can call one `static workflow details` and one `dynamic workflow details`?

cretz: Are you saying I should change the start workflow options to be `StaticSummary` and `StaticDetails`, and change the method in workflow to be `workflow.SetDynamicDetails`? One of the concerns is this now deviates from the API naming that was settled on a while back and is now implemented. I wouldn't want to have static summary not have the same prefix as static details. Maybe `StaticSummary` and `StaticDetails` and keep `workflow.SetCurrentDetails`? Or maybe `FixedSummary` and `FixedDetails`? Or maybe this can simply be documented away?

Sushisource (approved):

internal/client.go
Quinn-With-Two-Ns: I think we should use different names for different connects, I am not too picky on the exact names.

>. Maybe StaticSummary and StaticDetails and keep workflow.SetCurrentDetails

The works for me.

cretz: Ok, let me bring up with the group and get group-wide consensus.

cretz:
Updated to change to `StaticSummary`/`StaticDetails` for workflows

internal/client.go
cretz: Changed to `StaticSummary`/`StaticDetails`.

internal/internal_schedule_client.go
```diff
@@ -637,6 +637,11 @@ func convertToPBScheduleAction(
 			return nil, err
 		}
 
+		userMetadata, err := buildUserMetadata(action.staticSummary, action.staticDetails, dataConverter)
+		if err != nil {
+			return nil, err
+		}
+
```
gow, line 644: We need a similar change when the schedule is updated. The `UpdateSchedule` API on the server is structured to completely replace the `schedule.Action` with the value provided in the update API. So if the client wants to update a single field, it should echo back the whole schedule with the updated field. The client already seems to be doing that.
But in my testing I noticed that the usermetadata is not being copied over between these two calls - https://github.com/temporalio/sdk-go/blob/master/internal/internal_schedule_client.go#L276-L280. Looks like `ScheduleWorkflowAction` struct needs some updates?
cc @dnr

cretz: This same code is used for update. The way update works in SDKs is it's a get-then-update which calls these same helpers to go from user model to proto model.

Quinn-With-Two-Ns (approved):

Issue #1598, opened 2024-08-17
Typed nil versioning ramp panics


antlai-temporal:
## Expected Behavior
The interface `Ramp` is an optional field for `VersioningAssignmentRule`, we should be able to set
it to `nil` regardless of the type of the `nil` object 

## Actual Behavior
Currently we can set it to  untyped `nil`, but with a typed `nil` it panics with a 
bad memory reference.

## Steps to Reproduce the Problem
```
func Test_WorkerVersioningRules_typed_nil(t *testing.T) {
	ramp := &VersioningRampByPercentage{
		Percentage: 45.0,
	}
	ramp = nil

	u := UpdateWorkerVersioningRulesOptions{
		TaskQueue: "myQueue",
		Operation: &VersioningOperationInsertAssignmentRule{
			RuleIndex: 0,
			Rule: VersioningAssignmentRule{
				TargetBuildID: "2.0",
				Ramp:          ramp,
			},
		},
	}
      /// This panics....!!
	_, err := u.validateAndConvertToProto("my_namespace")
}
```

antlai-temporal:
Assignment rules have been deprecated and soon to be removed. Closing...

PR #1599, opened 2024-08-17
Enable typed nil in versioning ramp

antlai-temporal:
## What was changed

This fixes a versioning bug in the latest implementation.

See #1598 for details

1. Closes 
#1598 
2. How was this tested:

Unit test

### DIFF
--- a/internal/worker_versioning_rules.go
+++ b/internal/worker_versioning_rules.go
@@ -308,10 +308,13 @@ func versioningAssignmentRuleToProto(rule *VersioningAssignmentRule) *taskqueuep
 
 	switch r := rule.Ramp.(type) {
 	case *VersioningRampByPercentage:
-		result.Ramp = &taskqueuepb.BuildIdAssignmentRule_PercentageRamp{
-			PercentageRamp: &taskqueuepb.RampByPercentage{
-				RampPercentage: r.Percentage,
-			},
+		// Ramp is optional, checking for typed `nil`
+		if r != nil {
+			result.Ramp = &taskqueuepb.BuildIdAssignmentRule_PercentageRamp{
+				PercentageRamp: &taskqueuepb.RampByPercentage{
+					RampPercentage: r.Percentage,
+				},
+			}
 		}
 	}
 
@@ -419,8 +422,10 @@ func (r *VersioningAssignmentRule) validateRule() error {
 	}
 	switch ramp := r.Ramp.(type) {
 	case *VersioningRampByPercentage:
-		if err := ramp.validateRamp(); err != nil {
-			return err
+		if ramp != nil {
+			if err := ramp.validateRamp(); err != nil {
+				return err
+			}
 		}
 		// Ramp is optional, defaults to "nothing to validate"
 	}
--- a/internal/worker_versioning_rules_test.go
+++ b/internal/worker_versioning_rules_test.go
@@ -139,3 +139,23 @@ func Test_VersioningIntent(t *testing.T) {
 		})
 	}
 }
+
+func Test_WorkerVersioningRules_typed_nil(t *testing.T) {
+	ramp := &VersioningRampByPercentage{
+		Percentage: 45.0,
+	}
+	ramp = nil
+
+	u := UpdateWorkerVersioningRulesOptions{
+		TaskQueue: "myQueue",
+		Operation: &VersioningOperationInsertAssignmentRule{
+			RuleIndex: 0,
+			Rule: VersioningAssignmentRule{
+				TargetBuildID: "2.0",
+				Ramp:          ramp,
+			},
+		},
+	}
+	_, err := u.validateAndConvertToProto("my_namespace")
+	assert.NoError(t, err)
+}


internal/worker_versioning_rules.go
```diff
@@ -419,8 +422,10 @@ func (r *VersioningAssignmentRule) validateRule() error {
 	}
 	switch ramp := r.Ramp.(type) {
 	case *VersioningRampByPercentage:
-		if err := ramp.validateRamp(); err != nil {
-			return err
+		if ramp != nil {
```
Quinn-With-Two-Ns, line 425: Can me move the `nil` check out of the switch so when we add more `Ramp` types we don't need to duplicate this check?

internal/worker_versioning_rules_test.go
```diff
@@ -139,3 +139,23 @@ func Test_VersioningIntent(t *testing.T) {
 		})
 	}
 }
+
+func Test_WorkerVersioningRules_typed_nil(t *testing.T) {
+	ramp := &VersioningRampByPercentage{
```
Quinn-With-Two-Ns, line 144:
```suggestion
        var ramp *VersioningRampByPercentage
```
I think this should be equivalent and show what your trying to test a bit clearer

antlai-temporal: Yes, done, I was doing cut/paste from another example...

internal/worker_versioning_rules.go
antlai-temporal: `Ramp` is an interface and to know that a non-nil interface contains a nil pointer I need to type cast. Can't think of a way to do that without repeating the type cast for each concrete type, kind of defeating the purpose of lifting the check...

internal/worker_versioning_rules.go
```diff
@@ -308,10 +308,13 @@ func versioningAssignmentRuleToProto(rule *VersioningAssignmentRule) *taskqueuep
 
 	switch r := rule.Ramp.(type) {
 	case *VersioningRampByPercentage:
-		result.Ramp = &taskqueuepb.BuildIdAssignmentRule_PercentageRamp{
-			PercentageRamp: &taskqueuepb.RampByPercentage{
-				RampPercentage: r.Percentage,
-			},
+		// Ramp is optional, checking for typed `nil`
+		if r != nil {
```
cretz, line 312: If the type switch passes but this is nil, I think this the `Ramp` value is incorrectly being set. The `Ramp` value should be nil or a non-nil instance, it should not be a non-nil interface with nil value. Code is going to struggle to use this if `rule.Ramp == nil` does not work.

antlai-temporal: This was reported by a customer that wrote some innocent looking code that was failing:
```
r := &VersioningRampByPercentage{
...
}
if (some condition) {
     r = nil
}
```
I think this should be valid code, otherwise we are really going to confuse people...
But yes, coming from other languages it took me by surprise that `nil` has type...

cretz: > that wrote some innocent looking code that was failing:

It's not that code exactly. This only happens if you are taking a pointer to a nil variable. See https://go.dev/play/p/F_uAKDduIau.

> `nil` has type...

It does not usually have a type unless it is not being set properly on the interface. What's happening is the interface is non-nil but its implementation is nil. This is a common Go confusion. We don't usually explicitly handle this.

What will happen today if you provide a pointer to a nil `*ScheduleWorkflowAction` in the `Schedule.Action` parameter? It will panic, like this code does, and like all of our other places where we accept an interface and people improperly set it to a non-nil value w/ a nil impl.

antlai-temporal: This is the example shared by the customer https://play.golang.com/p/hhUikD1wDvs , yes, you need to do a bit more than that...

cretz: Right, this user is taking a pointer of a nil variable and assigning it to an interface. This is a known Go situation and in all the places we use interfaces, we don't test for nil implementations, just actual nil values. This same issue would happen if they used our other places that accept interfaces.

antlai-temporal: Do we document this somewhere, or this is just standard " good Go API behavior"? The nil struct pointer passed to a function that takes an interface looks like a foot gun...

cretz: I think this is a standard Go confusion/problem and a standard Go footgun. It is definitely not specific to Temporal and surprising to many users that encounter it the first time. If we want, and @Quinn-With-Two-Ns is ok with, handling this specific interface nilness in this case we can, but in general we never have and most libraries do not and most libraries do not document this Go behavior because it is known (even if it is a footfun).

Drahflow: @cretz Said customer from above here.
> It's not that code exactly.
It is exactly that code which lead to the `nil` access prompting my original report. The `ramp := &VersioningRampByPercentage` creates the variable with `*VersioningRampByPercentage` type (which becomes a typed `nil` once passed within `VersioningAssignmentRule` if it was `nil`).

In the original report ( https://temporalio.slack.com/archives/CTRCR8RBP/p1723805024798129?thread_ts=1714989745.540629&cid=CTRCR8RBP ) I also mentioned "And I cannot declare `ramp` of type internal.VersioningRamp to avoid that, because it's not re-exported by client afaics."

PR #1600, opened 2024-08-18
Add TraceFlags to otel tracing interceptor

ohad83:
According to the otel log datamodel, there needs to be a `TraceFlags` byte field, which currently defines only the `SAMPLED` flag:
https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-traceflags

## What was changed
Added the `TraceFlags` field to otel tracing interceptor

## Why?
To conform to otel's log data model, the `TraceFlags` field has to appear and define the `SAMPLED` flag.

## Checklist

1. How was this tested:

Added a TraceFlags assert to the current logger test

### DIFF
--- a/contrib/opentelemetry/tracing_interceptor.go
+++ b/contrib/opentelemetry/tracing_interceptor.go
@@ -26,6 +26,7 @@ package opentelemetry
 import (
 	"context"
 	"fmt"
+
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/baggage"
@@ -237,6 +238,7 @@ func (t *tracer) GetLogger(logger log.Logger, ref interceptor.TracerSpanRef) log
 	logger = log.With(logger,
 		"TraceID", span.SpanContext().TraceID(),
 		"SpanID", span.SpanContext().SpanID(),
+		"TraceFlags", span.SpanContext().TraceFlags(),
 	)
 
 	return logger
--- a/contrib/opentelemetry/tracing_interceptor_logger_test.go
+++ b/contrib/opentelemetry/tracing_interceptor_logger_test.go
@@ -81,9 +81,7 @@ func TestLogFields(t *testing.T) {
 	span := rec.Ended()[0]
 	assert.Contains(t, buf.String(), "TraceID="+span.Parent().TraceID().String())
 	assert.Contains(t, buf.String(), "SpanID="+span.Parent().SpanID().String())
-
-	assert.Contains(t, buf.String(), "TraceID="+span.Parent().TraceID().String())
-	assert.Contains(t, buf.String(), "SpanID="+span.Parent().SpanID().String())
+	assert.Contains(t, buf.String(), "TraceFlags="+span.Parent().TraceFlags().String())
 }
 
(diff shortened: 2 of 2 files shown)


cretz:
Hrmm, `GetLogger` I am not sure is meant to be just for OTel logging or the OTel log model. Notice there's a lot of fields in that spec we don't add. I am not sure we want every OTel tracing user to have to have this new attribute in their logger that is always the same value anyways (the majority I am sure are not using OTel logging). The identifier attributes make a bit more sense, but not sure about many of the others. Having said that, you can provide your own interceptor after/inside this one that implements `GetLogger` to add all sorts of attributes including this one.

ohad83:
I'm sorry @cretz  but I'm very confused about your answer. The `GetLogger` I changed is in the **opentelemetry** tracing interceptor, if someone doesn't want to use open telemetry why would they use this interceptor? This just makes it conform to the protocol, and it's used in monitoring products which support open telemetry. 
Specifically in my case it's used by GCP monitoring to connect logs to the trace, and if the log doesn't inform the system that the trace is sampled it won't try to connect them, because that's what the protocol says.

cretz:
> if someone doesn't want to use open telemetry why would they use this interceptor?

They do want to use stable OpenTelemetry tracing. This is the OpenTelemetry tracing interceptor, not just general purpose OpenTelemetry interceptor. OpenTelemetry metrics and OpenTelemetry logging are separate things, and we offer a metrics handler and logger abstraction for those.

I think we can consider this as maybe an opt-in option. But I don't think we want to infect everyone's logs that may not use OTel logging with this always-the-same-value attribute.

cretz:
@Quinn-With-Two-Ns - any thoughts here?

Quinn-With-Two-Ns:
>your own interceptor after/inside this one that implements GetLogger to add all sorts of attributes including this one

I agree

>They do want to use stable OpenTelemetry tracing, just maybe not logging. This is the OpenTelemetry tracing interceptor, not just general purpose OpenTelemetry interceptor. 

I agree this is an otel tracing interceptor not a logging interceptor. Supporting otel logging could be a useful addition, but should be separate from tracing support. I would need to spend more time reviewing the otel logging spec to consider what integrations is possible. For now I would r

ohad83:
If to support OTel logging I need to write my own interceptor, you might as well remove the whole logging changes in this interceptor. It infects the logs of anyone not using OTel logging and anyone who does use it needs to implement it anyway and know what parts you do implement and what you don't.

By the way, regarding the "always the same value", [sampling](https://opentelemetry.io/docs/concepts/sampling) is a real, much-used part of OTel tracing in real world production scenarios. Ignoring it is, in my opinion, a wrong choice in a module that's meant to implement OTel for your product.

Quinn-With-Two-Ns:
>anyone not using OTel logging and anyone who does use it needs to implement it anyway and know what parts you do implement and what you don't

Yes maybe the logging changes should have been separated out to a separate interceptor. This would be something we would look at when adding support for otel logging across our SDKs.

ohad83:
OK, I'm closing this PR as it seems like you prefer not changing it. Opened #1603  to add an option to disable the logging so I can implement an actual OTel logging interceptor without fearing side effects from this interceptor.

PR #1601, opened 2024-08-19
store signal received when defaultFunc is executing

coder336:
## What was changed
We are storing the signal value in the channel when a signal is received in select statement.

## Why?
When a signal is sent to the channel while the AddDefault method (long running) is executing, the signal is lost without being received. This was because if AddDefault is specified, the readyFunc is never executed. By storing the value in channel's recValue beforehand, we are making it available for any future receives on that channel without dropping the signal.

## Checklist

1. Closes 
n/a

2. How was this tested:
Manual testing
https://github.com/altafDevRev/temporal-samples/pull/1

3. Any docs updates needed?
No.

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -1382,8 +1382,8 @@ func (s *selectorImpl) Select(ctx Context) {
 					if readyBranch != nil {
 						return false
 					}
+					c.recValue = &v
 					readyBranch = func() {
-						c.recValue = &v
 						f(c, more)
 					}
 					return true
--- a/internal/internal_workflow_testsuite_test.go
+++ b/internal/internal_workflow_testsuite_test.go
@@ -1382,6 +1382,36 @@ func (s *WorkflowTestSuiteUnitTest) Test_MockActivityWait() {
 	s.NoError(env.GetWorkflowError())
 }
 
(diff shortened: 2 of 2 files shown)


Quinn-With-Two-Ns:
> Any backwards compatibility concerns w/ workflows who may have accidentally formed an expectation of lost signal?

Yes there is significant backwards compatibility concerns with a change like this. If there is a issue here any change would need to be backwards compatibility with old workflow executions that did use this behaviour.

coder336:
Added the test.
I believe the signal loss is a more significant issue. The `Select` method should offer the same functionality as the standard Go `select`, allowing you to listen on multiple channels with a default without losing a signal. If other clients are relying on this behaviour, they're likely accounting for it too.

Quinn-With-Two-Ns:
@coder336 I agree that data loss is an important issue, but backwards compatibility in workflows is the **highest priority**. These changes must be done in a backwards compatible way and will take multiple releases of the Go SDK to be enabled. Given the scope of these changes and the timeline the fix would be over I would ask that you file a bug report and someone from the Temporal team can drive this to resolution.

Quinn-With-Two-Ns:
Thanks again for the bug report and reproduction. I filed an issue to track this here: 

https://github.com/temporalio/sdk-go/issues/1624

PR #1603, opened 2024-08-20
Add DisableLogging option to otel tracing interceptor

ohad83:
The otel tracing interceptor currently adds a partial implementation of otel logging. This may be undesired as it pollutes the logs while not fully implementing otel logging.

## What was changed
Added a `DisableLogging` option to the OTel tracing interceptor.

## Why?
The current OTel tracing interceptor adds the `SpanID` and `TraceID` to the logs. For users who don't use OTel logging, this pollutes their logs without a way to disable it. For users who do use OTel logging, this is not enough as it doesn't fully conform to the OTel protocol. This option adds a way to disable it so it either doesn't add unneeded log attributes when you don't need it, and allows OTel users to implement their own logger without having side effects from this interceptor.

### DIFF
--- a/contrib/opentelemetry/tracing_interceptor.go
+++ b/contrib/opentelemetry/tracing_interceptor.go
@@ -26,6 +26,7 @@ package opentelemetry
 import (
 	"context"
 	"fmt"
+
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/baggage"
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
To be clear we would consider this option when de design otel logging support. We need to consider these things holistically at the moment we are not looking to complicate the otel tracing options.

cretz:
Hrmm, I think it's reasonable for people to be able to disable adding these tags to every log statement from the tracing interceptor.

PR #1605, opened 2024-08-20
Handle Nexus links

rodrigozhou:
## What was changed

Handle Nexus links

## Why?

Support Nexus bi-directional links.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


go.mod
```diff
@@ -13,9 +13,9 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.9.0
-	go.temporal.io/api v1.36.0
-	golang.org/x/sync v0.7.0
-	golang.org/x/sys v0.22.0
+	go.temporal.io/api v1.37.1-0.20240820161735-b85544bde489
```
bergundy, line 16: Let's merge the API PR before this PR.

internal/internal_nexus_task_handler.go
```diff
@@ -162,11 +169,23 @@ func (h *nexusTaskHandler) handleStartOperation(ctx context.Context, nctx *Nexus
 	if callbackHeader == nil {
 		callbackHeader = make(map[string]string)
 	}
+	var nexusLinks []nexus.Link
+	for _, link := range req.GetLinks() {
+		linkData, err := proto.Marshal(link)
```
bergundy, line 174: I think the API can just send a link type that's `type + data` and avoid unmarshalling in the server and then marshaling in the SDK.

internal/internal_nexus_task_handler.go
```diff
@@ -217,11 +236,27 @@ func (h *nexusTaskHandler) handleStartOperation(ctx context.Context, nctx *Nexus
 	}
 	switch t := opres.(type) {
 	case *nexus.HandlerStartOperationResultAsync:
+		var links []*common.Link
+		for _, nexusLink := range t.Links {
+			switch nexusLink.Type {
```
bergundy, line 241: Same, here, just accept a generic `Link`

temporalnexus/operation.go
```diff
@@ -223,7 +229,29 @@ func (o *workflowRunOperation[I, O]) Start(ctx context.Context, input I, options
...
-	return &nexus.HandlerStartOperationResultAsync{OperationID: handle.ID()}, nil
+
+	// Create the link information about the new workflow and return to the caller.
+	link := &common.Link{
+		Variant: &common.Link_WorkflowEvent_{
+			WorkflowEvent: &common.Link_WorkflowEvent{
+				Namespace:  nctx.Namespace,
+				WorkflowId: handle.ID(),
+				RunId:      handle.RunID(),
+			},
```
rodrigozhou, line 240: This is the link from the handler side, and I'm not setting the event_id or anything else at this moment.

internal/internal_nexus_task_handler.go
```diff
@@ -162,11 +169,27 @@ func (h *nexusTaskHandler) handleStartOperation(ctx context.Context, nctx *Nexus
...
 	}
+	var nexusLinks []nexus.Link
+	for _, link := range req.GetLinks() {
+		if link == nil {
+			continue
+		}
+		linkURL, err := url.Parse(link.GetUrl())
+		if err != nil {
+			nctx.Log.Error("failed to parse link url: %s", link.GetUrl(), tagError, err)
+			return nil, h.internalError(fmt.Errorf("failed to parse link url: %w", err)), nil
```
bergundy, line 180: This would be a bad request

internal/internal_nexus_task_handler.go
```diff
@@ -162,11 +169,27 @@ func (h *nexusTaskHandler) handleStartOperation(ctx context.Context, nctx *Nexus
 	if callbackHeader == nil {
 		callbackHeader = make(map[string]string)
 	}
+	var nexusLinks []nexus.Link
```
bergundy, line 172: You can allocate the size to be equal to `len(req.GetLinks())`

converter/link_converter_test.go
```diff
@@ -0,0 +1,287 @@
...
+					Path:     "/namespaces//workflows/wf-id//run-id/history",
+					RawPath:  "/namespaces//workflows/wf-id%2F/run-id/history",
+					RawQuery: "referenceType=EventReference&eventID=1&eventType=WorkflowExecution",
+				},
+				Type: "temporal.api.common.v1.Link.WorkflowEvent",
+			},
+			errMsg: "failed to parse link to Link_WorkflowEvent",
+		},
+		{
+			name: "invalid event id missing",
```
bergundy, line 260: This should be valid and we should make sure we don't put 0 as the event ID in the workflow run operation response link.

converter/link_converter.go
```diff
@@ -0,0 +1,156 @@
...
+		// TODO(rodrigozhou): do we want to return an error here?
+	}
+
+	return we, nil
+}
+
+func convertLinkWorkflowEventEventReferenceToURLQuery(eventRef *commonpb.Link_WorkflowEvent_EventReference) string {
+	values := url.Values{
+		linkWorkflowEventReferenceTypeKey: []string{string(eventRef.ProtoReflect().Descriptor().Name())},
+		linkEventReferenceEventIDKey:      []string{strconv.FormatInt(eventRef.GetEventId(), 10)},
```
bergundy, line 137: Avoid setting this to 0.

converter/link_converter.go
```diff
@@ -0,0 +1,156 @@
...
+	linkEventReferenceEventIDKey      = "eventID"
+	linkEventReferenceEventTypeKey    = "eventType"
+)
+
+var (
+	rePatternNamespace  = fmt.Sprintf(`(?P<%s>[a-z0-9-.]+)`, urlPathNamespaceKey)
+	rePatternWorkflowID = fmt.Sprintf(`(?P<%s>[^/]+)`, urlPathWorkflowIDKey)
+	rePatternRunID      = fmt.Sprintf(`(?P<%s>[a-zA-Z0-9-]+)`, urlPathRunIDKey)
+	urlPathRE           = regexp.MustCompile(fmt.Sprintf(
+		`^/+namespaces/+%s/+workflows/+%s/+%s/+history/*$`,
```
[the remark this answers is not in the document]
rodrigozhou: What about a trailing slash?

converter/link_converter.go
```diff
@@ -0,0 +1,156 @@
...
+
+	linkWorkflowEventReferenceTypeKey = "referenceType"
+	linkEventReferenceEventIDKey      = "eventID"
+	linkEventReferenceEventTypeKey    = "eventType"
+)
+
+var (
+	rePatternNamespace  = fmt.Sprintf(`(?P<%s>[a-z0-9-.]+)`, urlPathNamespaceKey)
+	rePatternWorkflowID = fmt.Sprintf(`(?P<%s>[^/]+)`, urlPathWorkflowIDKey)
+	rePatternRunID      = fmt.Sprintf(`(?P<%s>[a-zA-Z0-9-]+)`, urlPathRunIDKey)
```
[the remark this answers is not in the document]
rodrigozhou: I didn't add anything on the server side. I did here in SDK only.

converter/link_converter.go
```diff
@@ -0,0 +1,156 @@
...
+			link.Type,
+			we.ProtoReflect().Descriptor().FullName(),
+		)
+	}
+
+	if link.URL.Scheme != urlSchemeTemporalKey {
+		return nil, fmt.Errorf("failed to parse link to Link_WorkflowEvent")
+	}
+
+	matches := urlPathRE.FindStringSubmatch(link.URL.EscapedPath())
```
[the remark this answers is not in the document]
rodrigozhou: I just thought that validating the individual parts of the split a bit ugly. Regex looks cleaner.

converter/link_converter.go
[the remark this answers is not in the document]
bergundy: Ah good point. There's a want to reuse this code in the server codebase to avoid duplicating this logic.

bergundy:
We should expose the links in the workflow API on the `NexusOperationExecution` https://pkg.go.dev/go.temporal.io/sdk@v1.28.1/workflow#NexusOperationExecution and in workflow info when a workflow is started with a link.

We can do that in a follow up PR AFAIC.

converter/link_converter.go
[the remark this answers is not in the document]
cretz: :+1: Can move it and leave it public and document it as user facing since it will be

converter/link_converter_test.go
rodrigozhou: What value should we put if event ID is missing then?

converter/link_converter.go
```diff
@@ -0,0 +1,156 @@
...
+	rePatternRunID      = fmt.Sprintf(`(?P<%s>[a-zA-Z0-9-]+)`, urlPathRunIDKey)
+	urlPathRE           = regexp.MustCompile(fmt.Sprintf(
+		`^/+namespaces/+%s/+workflows/+%s/+%s/+history/*$`,
+		rePatternNamespace,
+		rePatternWorkflowID,
+		rePatternRunID,
+	))
+)
+
+func ConvertLinkWorkflowEventToNexusLink(we *commonpb.Link_WorkflowEvent) nexus.Link {
```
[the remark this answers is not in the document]
rodrigozhou: Yes, we need these functions in the server.

converter/link_converter.go
```diff
@@ -0,0 +1,156 @@
...
+	enumspb "go.temporal.io/api/enums/v1"
+)
+
+const (
+	urlSchemeTemporalKey = "temporal"
+	urlPathNamespaceKey  = "namespace"
+	urlPathWorkflowIDKey = "workflowID"
+	urlPathRunIDKey      = "runID"
+	urlPathTemplate      = "/namespaces/%s/workflows/%s/%s/history"
+	urlTemplate          = "temporal://" + urlPathTemplate
```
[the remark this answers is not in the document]
rodrigozhou: Moved the file to `temporalnexus`.

temporalnexus/link_converter_test.go
```diff
@@ -0,0 +1,320 @@
...
+			output: nexus.Link{
+				URL: &url.URL{
+					Scheme:   "temporal",
+					Path:     "/namespaces/ns/workflows/wf-id/run-id/history",
+					RawPath:  "/namespaces/ns/workflows/wf-id/run-id/history",
+					RawQuery: "eventID=1&eventType=WorkflowExecutionStarted&referenceType=EventReference",
+				},
+				Type: "temporal.api.common.v1.Link.WorkflowEvent",
+			},
+			outputURL: "temporal:///namespaces/ns/workflows/wf-id/run-id/history?eventID=1&eventType=WorkflowExecutionStarted&referenceType=EventReference",
```
cretz, line 70: I like the idea of the run ID being a query param too, but meh not important

cretz (approved):

internal/internal_nexus_task_handler.go
```diff
@@ -217,11 +240,21 @@ func (h *nexusTaskHandler) handleStartOperation(ctx context.Context, nctx *Nexus
 	}
 	switch t := opres.(type) {
 	case *nexus.HandlerStartOperationResultAsync:
+		var links []*nexuspb.Link
+		for _, nexusLink := range t.Links {
+			links = append(links, &nexuspb.Link{
+				Url:  nexusLink.URL.String(),
```
bergundy, line 246: Hmm... I wonder if it would be worth escaping the raw query. It doesn't seem like `String()` does that for you. But this is probably fine, the server would just ignore if the URL can't be parsed.

temporalnexus/link_converter_test.go
bergundy: In the UI it's in the path.

bergundy (approved):

temporalnexus/link_converter_test.go
cretz: How much of this is meant to match UI path vs HTTP API paths?

PR #1613, opened 2024-08-29
Add Nexus links tests

rodrigozhou:
## What was changed

Add Nexus links tests

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/test/nexus_test.go
+++ b/test/nexus_test.go
@@ -651,11 +651,14 @@ func TestAsyncOperationFromWorkflow(t *testing.T) {
 			false,
 			enums.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT,
 		)
+		var nexusOperationScheduleEventID int64
 		var targetEvent *historypb.HistoryEvent
 		for iter.HasNext() {
 			event, err := iter.Next()
 			require.NoError(t, err)
-			if event.GetEventType() == enums.EVENT_TYPE_NEXUS_OPERATION_STARTED {
+			if event.GetEventType() == enums.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED {
+				nexusOperationScheduleEventID = event.GetEventId()
+			} else if event.GetEventType() == enums.EVENT_TYPE_NEXUS_OPERATION_STARTED {
 				targetEvent = event
 				break
 			}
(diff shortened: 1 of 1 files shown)


test/nexus_test.go
```diff
@@ -517,6 +555,72 @@ func TestAsyncOperationFromWorkflow(t *testing.T) {
...
+		require.Equal(t, tc.testConfig.Namespace, link.GetWorkflowEvent().GetNamespace())
+		require.Equal(t, handlerWfID, link.GetWorkflowEvent().GetWorkflowId())
+		require.NotEmpty(t, link.GetWorkflowEvent().GetRunId())
+		requireProtoEqual(
+			t,
+			&common.Link_WorkflowEvent_EventReference{
+				EventType: enums.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED,
+			},
+			link.GetWorkflowEvent().GetEventRef())
+		handlerRunID := link.GetWorkflowEvent().GetRunId()
```
bergundy, line 588: nit: you could have just used `requireProtoEqual` on the entire struct.

test/nexus_test.go
```diff
@@ -290,10 +298,22 @@ func TestNexusWorkflowRunOperation(t *testing.T) {
 
 	nc := tc.newNexusClient(t, service.Name)
 
+	link := &common.Link_WorkflowEvent{
```
bergundy, line 301: This is redundant IMHO since we're already testing adding links from a workflow.

bergundy (approved):
I missed this in the last PR but please mark these functions as experimental in a different PR since this one has to remain a draft until we have a CLI with server 1.26.
https://github.com/temporalio/sdk-go/blob/master/temporalnexus/link_converter.go#L65-L89

test/nexus_test.go
rodrigozhou: I cannot because I don't have the handler run id.

test/nexus_test.go
```diff
@@ -31,6 +31,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
```
cretz, line 34: Is it worth a new test dependency just to show failures in this one test that should not happen anyways? I think https://pkg.go.dev/google.golang.org/protobuf/proto#Equal should be good enough for this case right?

rodrigozhou: Removed

cretz (approved):

Issue #1612, opened 2024-08-28
Request to make queryRejectedError public to enhance Interceptor error handling


deepika-awasthi:
**Is your feature request related to a problem? Please describe.**

When using QueryWorkflowWithOptions with QUERY_REJECT_CONDITION_NOT_OPEN, query rejection errors cannot be caught in the interceptor because the queryRejectedError is private in the Go SDK. Making this error public would allow better handling in error interceptors

**Describe the solution you'd like**

Making queryRejectedError public so that it can be explicitly handled in interceptors, similar to other common errors.

**Describe alternatives you've considered**

- Using QueryWorkflowWithOptionsResponse.QueryRejected directly is not an ideal solution because the rejection status is returned as part of the workflow's response, meaning it bypasses the typical error handling flow used in interceptors. 
- String matching the error message is also an option, but it is not a reliable solution.

**Additional context**

This issue affects error handling consistency and prevents query rejection errors from being managed properly in interceptors.

PR #1616, opened 2024-08-31
Expose QueryRejectedError

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-go/issues/1612


cretz (approved):

interceptor/interceptor.go
```diff
@@ -220,6 +220,10 @@ type ScheduleClientCreateInput = internal.ScheduleClientCreateInput
 // ClientOutoundInterceptor.UpdateWorkflow.
 type ClientUpdateWorkflowInput = internal.ClientUpdateWorkflowInput
 
+// QueryRejectedError is a possible error that can be returned by
+// ClientOutboundInterceptor.QueryWorkflow to indicate that the query was rejected by the server.
+type QueryRejectedError = internal.QueryRejectedError
```
cretz, line 225: Arguably may be better to expose in `temporal` package.

PR #1618, opened 2024-09-03
Update internal timers to use NewTimerWithOptions

### LINKED ISSUE #1602: Update internal timers to use NewTimerWithOptions and a summary

Since #1597, there is now a way to provide a summary for a timer, so we should do so for internal timers we create inside the SDK

yuandrew:
## What was changed

- Change internal uses of `NewTimer` to `NewTimerWithOptions`
- Created new `AwaitWithTimeoutAndOption` to allow specifying options for the underlying `timer` that's created

## Why?

Because there's now a way to give a summary for internal timers, change internal code to use the new timer with summary command.

## Checklist

1. Closes #1602 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -213,10 +213,10 @@ type WorkflowOutboundInterceptor interface {
 	// AwaitWithTimeout intercepts workflow.AwaitWithTimeout.
 	AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool) (bool, error)
 
-	// AwaitWithTimeoutWithOptions intercepts workflow.AwaitWithTimeoutWithOptions.
+	// AwaitWithOptions intercepts workflow.AwaitWithOptions.
 	//
 	// NOTE: Experimental
-	AwaitWithTimeoutAndOptions(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (bool, error)
+	AwaitWithOptions(ctx Context, options AwaitOptions, condition func() bool) (bool, error)
 
 	// ExecuteActivity intercepts workflow.ExecuteActivity.
 	// interceptor.WorkflowHeader will return a non-nil map for this context.
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -214,6 +214,8 @@ type WorkflowOutboundInterceptor interface {
 	AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool) (bool, error)
 
 	// AwaitWithTimeoutWithOptions intercepts workflow.AwaitWithTimeoutWithOptions.
+	//
+	// NOTE: Experimental
 	AwaitWithTimeoutAndOptions(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (bool, error)
 
 	// ExecuteActivity intercepts workflow.ExecuteActivity.
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -211,7 +211,7 @@ type WorkflowOutboundInterceptor interface {
 	Await(ctx Context, condition func() bool) error
 
 	// AwaitWithTimeout intercepts workflow.AwaitWithTimeout.
-	AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool) (bool, error)
+	AwaitWithTimeout(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (bool, error)
 
 	// ExecuteActivity intercepts workflow.ExecuteActivity.
 	// interceptor.WorkflowHeader will return a non-nil map for this context.
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -211,7 +211,10 @@ type WorkflowOutboundInterceptor interface {
 	Await(ctx Context, condition func() bool) error
 
 	// AwaitWithTimeout intercepts workflow.AwaitWithTimeout.
-	AwaitWithTimeout(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (bool, error)
+	AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool) (bool, error)
+
+	// AwaitWithTimeoutWithOptions intercepts workflow.AwaitWithTimeoutWithOptions.
+	AwaitWithTimeoutAndOptions(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (bool, error)
 
 	// ExecuteActivity intercepts workflow.ExecuteActivity.
 	// interceptor.WorkflowHeader will return a non-nil map for this context.
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -252,6 +252,8 @@ type WorkflowOutboundInterceptor interface {
 	NewTimer(ctx Context, d time.Duration) Future
 
 	// NewTimer intercepts workflow.NewTimerWithOptions.
+	//
+	// NOTE: Experimental
 	NewTimerWithOptions(ctx Context, d time.Duration, options TimerOptions) Future
 
 	// Sleep intercepts workflow.Sleep.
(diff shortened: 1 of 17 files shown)


workflow/deterministic_wrappers.go
```diff
@@ -102,11 +102,11 @@ func Await(ctx Context, condition func() bool) error {
 // The following code will block until the captured count
 // variable is set to 5, or one hour passes.
 //
-//	workflow.AwaitWithTimeout(ctx, time.Hour, func() bool {
+//	workflow.AwaitWithTimeout(ctx, time.Hour, options, func() bool {
 //	  return count == 5
 //	})
-func AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool) (ok bool, err error) {
-	return internal.AwaitWithTimeout(ctx, timeout, condition)
+func AwaitWithTimeout(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (ok bool, err error) {
```
yuandrew, line 108: Does changing this function signature break anything/anyone external?

internal/interceptor.go
```diff
@@ -211,7 +211,7 @@ type WorkflowOutboundInterceptor interface {
 	Await(ctx Context, condition func() bool) error
 
 	// AwaitWithTimeout intercepts workflow.AwaitWithTimeout.
-	AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool) (bool, error)
+	AwaitWithTimeout(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (bool, error)
```
Quinn-With-Two-Ns, line 214: We cannot change the API like this unfortunately

Quinn-With-Two-Ns: We will need to add a new API to pass options

workflow/deterministic_wrappers.go
Quinn-With-Two-Ns: Yes it would break all callers when they update their SDK. That is not okay. Go lang doesn't give us a lot of tools to work with here so we need to just add a new API like `AwaitWithOptions` or `AwaitWithTimeoutAndOptions`

yuandrew: had a hunch that'd be the case

internal/workflow.go
```diff
@@ -380,9 +380,18 @@ type (
 		// WARNING: Worker versioning is currently experimental
 		VersioningIntent VersioningIntent
 
-		// TODO(cretz): Expose once https://github.com/temporalio/temporal/issues/6412 is fixed
```
cretz, line 383: I don't think we need to change anything related to static summary/details or child workflows in this timer-specific issue. Also, while that issue is technically fixed, it would need to be in the dev server and we need integration tests.

internal/workflow.go
```diff
@@ -496,7 +505,7 @@ func AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool)
 func (wc *workflowEnvironmentInterceptor) AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool) (ok bool, err error) {
 	state := getState(ctx)
 	defer state.unblocked()
-	timer := NewTimer(ctx, timeout)
+	timer := NewTimerWithOptions(ctx, timeout, TimerOptions{"AwaitWithTimeout"})
```
cretz, line 508:
```suggestion
	timer := NewTimerWithOptions(ctx, timeout, TimerOptions{Summary: "AwaitWithTimeout"})
```

This is clearer to readers and a bit safer

internal/workflow.go
yuandrew: sounds good, I'll take these changes out of this PR

internal/workflow.go
```diff
@@ -513,6 +513,37 @@ func (wc *workflowEnvironmentInterceptor) AwaitWithTimeout(ctx Context, timeout
...
+func AwaitWithTimeoutAndOptions(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (ok bool, err error) {
+	assertNotInReadOnlyState(ctx)
+	state := getState(ctx)
+	return state.dispatcher.interceptor.AwaitWithTimeoutAndOptions(ctx, timeout, options, condition)
+}
+
+func (wc *workflowEnvironmentInterceptor) AwaitWithTimeoutAndOptions(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (ok bool, err error) {
+	state := getState(ctx)
+	defer state.unblocked()
+	if options.Summary == "" {
```
yuandrew, line 527: not sure why someone might call `AwaitWithTimeoutAndOptions` without specifying a summary, but defaulting to "AwaitWithTimeoutAndOptions"

cretz: I'm 50/50 here. We do not default a summary in `NewTimerWithOptions`, not sure we need to here. But then again it makes some sense since we are in the option-less equivalent calls. No strong opinion here.

yuandrew: After some thought, i think I'm leaning the other way now, if someone chooses to use `AwaitWithTimeoutAndOptions`, they should probably be responsible for setting their `TimerOptions` properly

internal/interceptor.go
```diff
@@ -213,6 +213,11 @@ type WorkflowOutboundInterceptor interface {
 	// AwaitWithTimeout intercepts workflow.AwaitWithTimeout.
 	AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool) (bool, error)
 
+	// AwaitWithTimeoutWithOptions intercepts workflow.AwaitWithTimeoutWithOptions.
+	//
+	// NOTE: Experimental
+	AwaitWithTimeoutAndOptions(ctx Context, timeout time.Duration, options TimerOptions, condition func() bool) (bool, error)
```
Quinn-With-Two-Ns, line 219: I wonder if we should just call this `AwaitWithOptions` and take timeout as an optional option?

internal/workflow.go
Quinn-With-Two-Ns: Yeah I'd be consistent with `NewTimerWithOptions` . We should set some sensible default if the user is calling `AwaitWithTimeout`

internal/interceptor.go
yuandrew: As in move `timeout` into the `TimerOptions` struct? If that's what you mean, do we want to change the `NewTimerWithOptions` to match?

Quinn-With-Two-Ns: More like create `AwaitOptions` and put `timeout` and `TimerOptions` under it. This is so if we need to add more fields we don't have to add more overrides

workflow/deterministic_wrappers.go
```diff
@@ -111,6 +111,21 @@ func AwaitWithTimeout(ctx Context, timeout time.Duration, condition func() bool)
...
+// Returns ok=false if timed out, and err CanceledError if the ctx is canceled.
+// The following code will block until the captured count
+// variable is set to 5, or one hour passes.
+//
+//	workflow.AwaitWithOptions(ctx, options, func() bool {
+//	  return count == 5
+//	})
+//
+// NOTE: Experimental
+func AwaitWithOptions(ctx Context, options internal.AwaitOptions, condition func() bool) (ok bool, err error) {
```
cretz, line 125: You'll need a type alias for `AwaitOptions` in the public package here to reference the internal, otherwise users can't even call it. This may surface if/when you write integration tests.

yuandrew: Is it worth adding a unit test for `AwaitWithOptions` that does the same thing as `AwaitWithTimeout` just to exercise the new function codepath in our tests?

cretz: Doesn't necessarily do the same thing, it also allows you to add a summary. Might as well test it, but I don't have a strong opinion. I think it'd have caught that certain needed things weren't exposed here though.

test/integration_test.go
```diff
@@ -6266,6 +6266,38 @@ func (ts *IntegrationTestSuite) TestUserMetadata() {
 	ts.Equal("my-timer", str)
 }
 
+func (ts *IntegrationTestSuite) TestAwaitWithOptionsTimeout() {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	var str string
+
+	// Start workflow
+	fmt.Println("start workflow")
```
cretz, line 6275: Can we remove this print?

cretz (approved):
LGTM, may want to wait until @Quinn-With-Two-Ns's approval though

Quinn-With-Two-Ns (approved):

cretz:
@yuandrew - Looks like you can merge if/when CI passes (I kicked it again, looks like flaky test)

yuandrew:
@cretz  looks like there's a regression with the `TestWorkerVersioningTestSuite/TestCommitRules` test, looking to repro this locally

PR #1623, opened 2024-09-06
mock: in order mock calls

ReyOrtiz:
## What was changed

Adding an `InOrderMockCalls(calls ...*MockCallWrapper)` helper method to setup mock calls in order. It just loops through the `calls` and sets the order by calling `.NotBefore()`

## Why?

A more intuitive and cleaner way to define mock calls in order. For example:

```
env.InOrderMockCalls(
	env.OnActivity(namedActivity, mock.Anything, "call1").Return("result1", nil),
	env.OnActivity(namedActivity, mock.Anything, "call2").Return("result2", nil),
	env.OnActivity(namedActivity, mock.Anything, "call3").Return("result3", nil),
)

```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -385,6 +385,12 @@ func (env *testWorkflowEnvironmentImpl) setContinuedExecutionRunID(rid string) {
 	env.workflowInfo.ContinuedExecutionRunID = rid
 }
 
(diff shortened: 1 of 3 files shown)


Quinn-With-Two-Ns:
This isn't a helper method the SDK needs to implement , if a user needed this convince method they are free to define it in their own code. In General we try to offer similar helpers to https://github.com/stretchr/testify.

PR #1626, opened 2024-09-09
Nexus handler error translation

bergundy:
## What was changed

- Handle `serviceerror`s, `ApplicationError`s, and other expected client errors returned from Nexus handlers.
- Also now exposing internal errors by default to match activity and child workflow behavior. We'll have to revisit hiding details from the caller when we start targeting cross org use cases.

## Why?

Better out of the box developer experience.

2. How was this tested:

Added integration tests.

### DIFF
--- a/internal/internal_nexus_task_handler.go
+++ b/internal/internal_nexus_task_handler.go
@@ -233,8 +233,7 @@ func (h *nexusTaskHandler) handleStartOperation(
 				},
 			}, nil, nil
 		}
-		// Default to expose details for now. We may make this configurable eventually.
-		err = convertServiceError(convertApplicationError(err), true)
+		err = convertKnownErrors(err)
 		var handlerErr *nexus.HandlerError
 		if errors.As(err, &handlerErr) {
 			return nil, nexusHandlerErrorToProto(handlerErr), nil
@@ -306,8 +305,7 @@ func (h *nexusTaskHandler) handleCancelOperation(ctx context.Context, nctx *Nexu
 		return nil, nil, ctx.Err()
 	}
 	if err != nil {
-		// Default to expose details for now. We may make this configurable eventually.
-		err = convertServiceError(convertApplicationError(err), true)
+		err = convertKnownErrors(err)
 		var handlerErr *nexus.HandlerError
 		if errors.As(err, &handlerErr) {
 			return nil, nexusHandlerErrorToProto(handlerErr), nil
@@ -325,7 +323,7 @@ func (h *nexusTaskHandler) handleCancelOperation(ctx context.Context, nctx *Nexu
 
 func (h *nexusTaskHandler) internalError(err error) *nexuspb.HandlerError {
 	h.logger.Error("error processing nexus task", "error", err)
-	return nexusHandlerError(nexus.HandlerErrorTypeInternal, "internal error")
+	return nexusHandlerError(nexus.HandlerErrorTypeInternal, err.Error())
 }
 
 func (h *nexusTaskHandler) goContextForTask(nctx *NexusOperationContext, header nexus.Header) (context.Context, context.CancelFunc, *nexuspb.HandlerError) {
@@ -423,9 +421,24 @@ func (p *payloadSerializer) Serialize(v any) (*nexus.Content, error) {
 
 var emptyReaderNopCloser = io.NopCloser(bytes.NewReader([]byte{}))
 
-// statusGetter represents Temporal serviceerrors which have a Status() method.
-type statusGetter interface {
-	Status() *status.Status
+// convertKnownErrors converts known errors to corresponding Nexus HandlerError.
+func convertKnownErrors(err error) error {
+	// Handle common errors returned from various client methods.
+	if workflowErr, ok := err.(*WorkflowExecutionError); ok {
+		return nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, workflowErr.Error())
+	}
+	if queryRejectedErr, ok := err.(*QueryRejectedError); ok {
+		return nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, queryRejectedErr.Error())
+	}
+
+	// Not using errors.As to be consistent ApplicationError checking with the rest of the SDK.
+	if appErr, ok := err.(*ApplicationError); ok {
+		if appErr.NonRetryable() {
+			return nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, appErr.Error())
+		}
+		return nexus.HandlerErrorf(nexus.HandlerErrorTypeInternal, appErr.Error())
+	}
+	return convertServiceError(err)
 }
 
 // convertServiceError converts a serviceerror into a Nexus HandlerError if possible.
@@ -434,10 +447,12 @@ type statusGetter interface {
 // Roughly taken from https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
 // and
 // https://github.com/grpc-ecosystem/grpc-gateway/blob/a7cf811e6ffabeaddcfb4ff65602c12671ff326e/runtime/errors.go#L56.
-func convertServiceError(err error, exposeDetails bool) error {
+func convertServiceError(err error) error {
 	var st *status.Status
-	var stGetter statusGetter
-	if !errors.As(err, &stGetter) {
+
+	// Temporal serviceerrors have a Status() method.
+	stGetter, ok := err.(interface{ Status() *status.Status })
+	if !ok {
 		// Not a serviceerror, passthrough.
 		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/internal_nexus_task_handler.go
```diff
@@ -416,3 +422,89 @@ func (p *payloadSerializer) Serialize(v any) (*nexus.Content, error) {
...
+// a default message with minimal information is attached to the returned error.
+// Roughly taken from https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+// and
+// https://github.com/grpc-ecosystem/grpc-gateway/blob/a7cf811e6ffabeaddcfb4ff65602c12671ff326e/runtime/errors.go#L56.
+func convertServiceError(err error, exposeDetails bool) error {
+	var st *status.Status
+	// Temporal serviceerrors have a Status() method.
+	stGetter, ok := err.(interface{ Status() *status.Status })
+	if !ok {
+		// Not a serviceerror, passthrough.
```
Quinn-With-Two-Ns, line 437: What if the error is wrapping a service error?

cretz: I personally like only checking the outer one. I don't think we should treat errors the same just because they may have the error somewhere in their chain.

internal/internal_nexus_task_handler.go
```diff
@@ -416,3 +422,89 @@ func (p *payloadSerializer) Serialize(v any) (*nexus.Content, error) {
...
 
 var emptyReaderNopCloser = io.NopCloser(bytes.NewReader([]byte{}))
+
+// convertServiceError converts a serviceerror into a Nexus HandlerError if possible.
+// If exposeDetails is true, the error message from the given error is exposed in the converted HandlerError, otherwise,
+// a default message with minimal information is attached to the returned error.
+// Roughly taken from https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+// and
+// https://github.com/grpc-ecosystem/grpc-gateway/blob/a7cf811e6ffabeaddcfb4ff65602c12671ff326e/runtime/errors.go#L56.
+func convertServiceError(err error, exposeDetails bool) error {
```
cretz, line 432: Doesn't seem like there's much value in splitting `convertServiceError` and `convertApplicationError`, might as well just have `convertError`. Also doesn't seem like `exposeDetails` parameter has any value since it is never set to false on this private function. Can wait for the future need to hide details before adding that functionality, it's dead currently.

internal/internal_nexus_task_handler.go
bergundy: Good call, switched to `errors.As` instead.

Quinn-With-Two-Ns: Can we just use https://pkg.go.dev/go.temporal.io/api@v1.39.0/serviceerror#ToStatus

bergundy: I was debating what to do here but I think most users will probably wrap just to provide more context but will expect to maintain the semantics.

internal/internal_nexus_task_handler.go
```diff
@@ -416,3 +422,93 @@ func (p *payloadSerializer) Serialize(v any) (*nexus.Content, error) {
...
+		return nexus.HandlerErrorf(nexus.HandlerErrorTypeInternal, "internal error")
+	}
+	return err
+}
+
+// convertApplicationError converts a Temporal ApplicationError to a Nexus HandlerError, respecting the non_retryable
+// flag.
+func convertApplicationError(err error) error {
+	var appErr *ApplicationError
+	if errors.As(err, &appErr) {
```
Quinn-With-Two-Ns, line 507: Normally the Go SDK explicitly does does not check the error chain for `ApplicationError`

Quinn-With-Two-Ns: Same with the Server

internal/internal_nexus_task_handler.go
bergundy: Agree there's not much value, for the split. Initially there was more logic in the application error conversion since I thought we may want to handle other TemporalErrors but I can merge.

I already know that we'll want to allow hiding details and had this logic in the server codebase (where I ported this status error conversion logic from) so I figured I'd keep it in here.

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: The problem is `serviceerrors ` is thrown low down even the SDK may wrap it at times

internal/internal_nexus_task_handler.go
bergundy: Restructured slightly.

internal/internal_nexus_task_handler.go
bergundy: Changed this.

internal/internal_nexus_task_handler.go
cretz: > I think most users will probably wrap just to provide more context but will expect to maintain the semantics.

I don't necessarily, if they wrap, the expect the behavior of what error they wrap as, not what was wrapped. I shouldn't have to be concerned that somewhere down my chain is a Temporal client error that is now going to affect my Nexus error type.

internal/internal_nexus_task_handler.go
```diff
@@ -416,3 +420,92 @@ func (p *payloadSerializer) Serialize(v any) (*nexus.Content, error) {
...
+	case codes.DeadlineExceeded:
+		if !exposeDetails {
+			errMessage = "request timeout"
+		}
+		return nexus.HandlerErrorf(nexus.HandlerErrorTypeDownstreamTimeout, errMessage)
+	}
+
+	// Default to internal error. This should only happen for codes.OK, which is unexpected for serviceerrors.
+	if !exposeDetails {
+		return nexus.HandlerErrorf(nexus.HandlerErrorTypeInternal, "internal error")
```
cretz, line 508: Dead code. Can you wait until you need it?

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: What is the motivation for handling service errors automatically? I don't see it called out in the prerelease docs and it is really only a Go SDK thing AFAIK

internal/internal_nexus_task_handler.go
bergundy: I'd rather keep this because I know I will need it later when we add the ability to control the level of details exposed from handlers (and callers).

internal/internal_nexus_task_handler.go
bergundy: It's not just a Go SDK thing, it's something all SDKs will need to handle.
If you signal a workflow that's not found from a handler for example, you'll want that operation to fail by default instead of the operation being retried indefinitely.

Quinn-With-Two-Ns: OK, but you don't do that by looking at the `serviceerror`. For example what if a handle tries to get a result from the workflow and the workflow has failed, do you want the handler to be retried indefinitely?

prasek: shouldn't the Nexus handler be catching `serviceerror` and returning a Nexus error? similar to [this](https://github.com/temporalio/temporal/blob/5bb4d5f5c56694921fe0023859e8ca2c51e5974c/components/nexusoperations/frontend/handler.go#L123)?

but for normal SDK users, who may be writing a `NewSyncOperation` may try to do something like this
```
			err = c.SignalWorkflow(ctx, workflowIDWaitForSignal(input.BusinessID), "", "done", DoneSignal{Done: true})
			if err != nil {
				//TODO: handle &serviceerror.NotFound{Message:"workflow execution already completed"}
				return nil, err
			}
```

but if they don't handle the `serviceerror` explicitly, and just return the error, then it will be treated as a 500 and be retried indefinitely, which probably isn't the behavior a Temporal SDK user would be expecting, and most definitely would not find delightful.

Automatically mapping the `serviceerror` reduces boilerplate in the Temporal SDK user's Nexus handlers, and also doesn't prevent an SDK user from writing their own error handling, it just helps reduce boilerplate and unexpected 500 infinite retries.

Quinn-With-Two-Ns: I am not disagreeing with reducing boilerplate. My issue is the current approach is incomplete and could mislead users. Not all errors returned from the client are `serviceerror` or have `serviceerror` in the error chain. From a users perspective it will look like seemingly random errors are retried and others aren't

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: I have no issue with the dead code, but It should be tested

internal/internal_nexus_task_handler.go
bergundy: Hmm... I was looking at random client methods and saw that servicerrors give good coverage.
I can scan the client implementation for other notable errors.

bergundy: Is there a specific error that you have in mind?

I saw `QueryRejectedError` and `WorkflowUpdateServiceTimeoutOrCanceledError`, both of which can be internal errors AFAICT.

Quinn-With-Two-Ns: I think `servicerror` misses some common cases like getting the result of a workflow that has failed would return an error that is not a `servicerror`

If we can clarify the exact desired user experience then we can work backwards to an implementation that gives us that. I think that is really important because other SDKs throw different client errors (we are very inconsistent here across SDKs)

bergundy: > getting the result of a workflow that has failed would return an error

What would this translate to? I think it's still an internal error (the default behavior).

Quinn-With-Two-Ns: If an update is rejected or fails that would also be an error that is not a service error, I would have to double check what error that is.

Quinn-With-Two-Ns: but internal is retryable right? Feels like it should be an `UnsuccessfulOperationError`?

bergundy: > If an update is rejected or fails that would also be an error that is not a service error, I would have to double check what error that is.

Doesn't look like that's returned as an error:

```
		case *updatepb.Outcome_Failure:
			return &ClientPollWorkflowUpdateOutput{
				Error: w.client.failureConverter.FailureToError(v.Failure),
			}, nil
```

bergundy: > but internal is retryable right? Feels like it should be an `UnsuccessfulOperationError`?

Yes, internal is retryable. What do you mean by "it"?

bergundy: Decided to make it non retryable but I couldn't use `UnsuccessfulOperationError` because that's only usable in `Start` and `GetResult`. It's going to be a bad request for lack of better outcome.

bergundy: Also decided to not use `errors.As` and only consider errors directly returned by client methods.
If users wrap the errors, they should decide what error type they want.
Let's get user feedback on this and verify it matches their expectation before we remove the experimental label.

internal/internal_nexus_task_handler.go
```diff
@@ -416,3 +420,65 @@ func (p *payloadSerializer) Serialize(v any) (*nexus.Content, error) {
 }
 
 var emptyReaderNopCloser = io.NopCloser(bytes.NewReader([]byte{}))
+
+// convertKnownErrors converts known errors to corresponding Nexus HandlerError.
+func convertKnownErrors(err error) error {
+	// Handle common errors returned from various client methods.
```
Quinn-With-Two-Ns, line 426: So will each SDK need to decided what is a common error and how to map it ?

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: Yeah I would really like to get feedback from users here if this is helpful and behaves how they expect.

Quinn-With-Two-Ns (approved):

PR #1629, opened 2024-09-10
Fix Version test failure due to 1.25.0 server release

yuandrew:
## What was changed
Changing a test assert based off of changed behavior from 1.25.0 server release. Test now checks for either no ramp or percentage is 100%

## Why?
1.25.0 server release caused this test to start failing. Server was changed so we can specify a 100% ramp as equivalent to no ramp.

## Checklist

1. Closes 

2. How was this tested:

Test passes locally

3. Any docs updates needed?

### DIFF
--- a/test/worker_versioning_test.go
+++ b/test/worker_versioning_test.go
@@ -334,7 +334,7 @@ func (ts *WorkerVersioningTestSuite) TestCommitRules() {
 	ts.Equal(1, len(res.AssignmentRules))
 	ts.Equal("2.0", res.AssignmentRules[0].Rule.TargetBuildID)
 	_, ok := res.AssignmentRules[0].Rule.Ramp.(*client.VersioningRampByPercentage)
-	ts.Falsef(ok, "Still has a percentage ramp")
+	ts.Truef(ok, "Ramp at 100 but type isn't updated")
 }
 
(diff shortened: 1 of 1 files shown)


test/worker_versioning_test.go
```diff
@@ -334,7 +334,7 @@ func (ts *WorkerVersioningTestSuite) TestCommitRules() {
 	ts.Equal(1, len(res.AssignmentRules))
 	ts.Equal("2.0", res.AssignmentRules[0].Rule.TargetBuildID)
 	_, ok := res.AssignmentRules[0].Rule.Ramp.(*client.VersioningRampByPercentage)
-	ts.Falsef(ok, "Still has a percentage ramp")
+	ts.Truef(ok, "Ramp at 100 but type isn't updated")
```
yuandrew, line 337: I am not sure exactly what changed in server to cause this change in behavior, so this could be an insufficient bandaid fix. 

Printing the value of *Ramp, it prints `100`. If I understand the Go flow here, `validateRamp()` should return an Error from this line? `return errors.New("invalid percentage in `Ramp`, not in [0,100)")`, which should cause `res.AssignmentRules[0].Rule.Ramp` to not be type `*client.VersioningRampByPercentage` anymore?

cretz (approved):

Quinn-With-Two-Ns (approved):

PR #1632, opened 2024-09-11
Attempt to read cgroups for resource tuner limits

Sushisource:
Respect container-imposed limits from the resource-based tuner.

Tested this with omes while running from within a container. Adding integ tests that start up docker didn't exactly feel practical.


Sushisource:
Supersedes https://github.com/temporalio/sdk-go/pull/1630

contrib/resourcetuner/cgroups.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/containerd/cgroups/v3/cgroup2"
```
cretz, line 38: Naive general cgroup question from someone that doesn't know much about cgroups (made as PR thread on this line because it's easier than top-level PR comment), does querying system limits inside a cgroup not represent the cgroup limit? What is the difference in a container between the observable system stats and the cgroup stats?

Sushisource: It doesn't, no, you explicitly have to go ask for your cgroup limits. Which makes some sense since you might still care about overall host capacity, and then the way you ask that would magically change inside a cgroup.

The stats part can depend on what's being read, too. You can read the same normal overall system usage, or you can read usage within a cgroup.

cretz: Makes sense. I would have expected it to be a security issue to see things outside of your cgroup, but I see that's only with namespacing. If at all possible would like to fashion a test here. I wonder if we need to step back and think of a test harness that can test resource tuning in a CI way. (but approving anyways since this is still experimental)

PR #1634, opened 2024-09-12
Dump junit xml

### LINKED ISSUE #1633: Modify test suite to dump junit XML
As in https://github.com/temporalio/sdk-python/pull/617 and https://github.com/temporalio/cli/pull/650, I propose to modify the test suite to output junit XML. The motivation for this is that, given appropriate tools, it will make it easier for us to study patterns of flakiness, test timings etc, across matrix builds within a run, and across runs. Producing the files now will allow us to test tools for doing this.

dandavison:
Closes https://github.com/temporalio/sdk-go/issues/1633

> As in https://github.com/temporalio/sdk-python/pull/617 and https://github.com/temporalio/cli/pull/650, I propose to modify the test suite to output junit XML. The motivation for this is that, given appropriate tools, it will make it easier for us to study patterns of flakiness, test timings etc, across matrix builds within a run, and across runs. Producing the files now will allow us to test tools for doing this.


.github/workflows/ci.yml
```diff
@@ -70,6 +70,14 @@ jobs:
           file: coverage.out
           format: golang
 
+      - name: 'Upload junit-xml artifacts'
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: junit-xml--${{github.run_id}}--${{github.run_attempt}}--${{matrix.os}}-${{matrix.go-version}}
+          path: .build/junit-xml
```
dandavison, line 78: This coupling (the output dir is hard-coded in both) between the Go build script and the CI yaml isn't ideal, but I'm thinking we can live with it for now. If we want to eliminate it we could have `-junit-output-dir` and `-junit-file-stem` CLI arguments.

internal/cmd/build/go.sum
```diff
@@ -3,18 +3,27 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/BurntSushi/toml v1.4.1-0.20240526193622-a339e1f7089c h1:pxW6RcqyfI9/kWtOwnv/G+AzdKuy2ZrqINhenH4HyNs=
```
dandavison, line 3: All changes to `go.mod` and `go.sum` are the result of running `go get gotest.tools/gotestsum` (looks like it was already a transitive dependency IIUC).

.github/workflows/ci.yml
yuandrew: Why is this not ideal? Is this because we need to make sure we keep these 2 paths in sync?

.github/workflows/ci.yml
```diff
@@ -70,6 +70,14 @@ jobs:
           file: coverage.out
           format: golang
 
+      - name: 'Upload junit-xml artifacts'
```
yuandrew, line 73: nit: None of the other step names in this file wrap the strings with `'`

yuandrew (approved):

.github/workflows/ci.yml
dandavison: Right, the `ci.yml` is invoking a Go build script which also hard-codes this path, and it would be ideal for it to be hard-coded in one place only. But it's not unusual to have some duplication like that in GitHub Actions job definitions, so I'd say probably OK for now.

yuandrew: Should we make a comment here that this should match `junitDir` in the build script, just for added safety?

.github/workflows/ci.yml
dandavison: Thanks! Removed.

.github/workflows/ci.yml
dandavison: Sure! Done.

dandavison:
@Quinn-With-Two-Ns To output junit XML I had to switch to using `gotestsum`. One question here is are you OK with the new test output format from `gotestsum`? From a quick look it appears that one big difference is that output from successful tests is hidden by default. I was optimistic that we'd be OK with this switch since https://github.com/temporalio/temporal uses it, and it looks like it's quite configurable, and fairly widely used by high profile Go projects: https://github.com/gotestyourself/gotestsum

Quinn-With-Two-Ns:
No concern about the test format. Question about the general PR though, from the issue it says

>The motivation for this is that, given appropriate tools, it will make it easier for us to study patterns of flakiness

Does that appropriate tool exist yet today?

dandavison:
> Does that appropriate tool exist yet today?

Good question. I couldn't find something that would work well, e.g. correlating tests across our matrix builds, so I've [started](https://github.com/dandavison/tringa) creating our own from scratch. I'm not quite ready to have others test it (but absolutely welcome any help developing it).

yuandrew:
Closing this older effort, there's no current workstream to use junit xml, but not to say this can never happen

PR #1636, opened 2024-09-16
Fix Nexus test env to respect ScheduleToCloseTimeout

rodrigozhou:
## What was changed

Fix Nexus test env to respect `ScheduleToCloseTimeout` and return operation time out error.

## Why?

Users can test the `ScheduleToCloseTimeout` when writing their tests.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -2369,6 +2369,7 @@ func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(
 
 	var opID string
 	if params.options.ScheduleToCloseTimeout > 0 {
+		// Timer to fail the nexus operation due to schedule to close timeout.
 		env.NewTimer(
 			params.options.ScheduleToCloseTimeout,
 			TimerOptions{},
@@ -2386,7 +2387,11 @@ func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(
 					},
 				))
 				env.postCallback(func() {
-					handle.startedCallback(opID, timeoutErr)
+					// For async operation, there are two scenarios:
+					// 1. operation already started: the callback has already been called with the operation id,
+					//    and calling again is no-op;
+					// 2. operation didn't start yet: there's no operation id to set.
+					handle.startedCallback("", timeoutErr)
 					handle.completedCallback(nil, timeoutErr)
 				}, true)
 			},
@@ -2424,8 +2429,8 @@ func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(
 				handle.completedCallback(v.SyncSuccess.GetPayload(), nil)
 			}, true)
 		case *nexuspb.StartOperationResponse_AsyncSuccess:
-			opID = v.AsyncSuccess.GetOperationId()
 			env.postCallback(func() {
+				opID = v.AsyncSuccess.GetOperationId()
 				handle.startedCallback(v.AsyncSuccess.GetOperationId(), nil)
 				if handle.cancelRequested {
 					handle.cancel()
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -2351,7 +2351,11 @@ func (env *testWorkflowEnvironmentImpl) newTestNexusTaskHandler() *nexusTaskHand
 	)
 }
 
-func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(params executeNexusOperationParams, callback func(*commonpb.Payload, error), startedHandler func(opID string, e error)) int64 {
+func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(
+	params executeNexusOperationParams,
+	callback func(*commonpb.Payload, error),
+	startedHandler func(opID string, e error),
+) int64 {
 	seq := env.nextID()
 	taskHandler := env.newTestNexusTaskHandler()
 	handle := &testNexusOperationHandle{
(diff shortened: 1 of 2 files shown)


bergundy (changes requested):
The task was to actually enforce the timeout, e.g. if the schedule-to-close-timeout is 1 hour, there should be a timer set to time out the operation when that timer fires.

internal/internal_workflow_testsuite.go
```diff
@@ -2363,6 +2367,32 @@ func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(params executeNexu
...
+						Message: "operation timed out",
+						FailureInfo: &failurepb.Failure_TimeoutFailureInfo{
+							TimeoutFailureInfo: &failurepb.TimeoutFailureInfo{
+								TimeoutType: enumspb.TIMEOUT_TYPE_SCHEDULE_TO_CLOSE,
+							},
+						},
+					},
+				))
+				env.postCallback(func() {
+					handle.startedCallback(opID, timeoutErr)
```
bergundy, line 2389: There's no need to send the operation ID if the operation timed out, the startedCallback should have already been called with the operation ID at this point.

test/nexus_test.go
```diff
@@ -897,6 +897,83 @@ func TestWorkflowTestSuite_WorkflowRunOperation(t *testing.T) {
...
+	sleepDuration := 500 * time.Millisecond
+	handlerWF := func(ctx workflow.Context, _ nexus.NoValue) (nexus.NoValue, error) {
+		return nil, workflow.Sleep(ctx, sleepDuration)
+	}
+
+	op := temporalnexus.NewWorkflowRunOperation(
+		"op",
+		handlerWF,
+		func(ctx context.Context, _ nexus.NoValue, opts nexus.StartOperationOptions) (client.StartWorkflowOptions, error) {
+			return client.StartWorkflowOptions{ID: opts.RequestID}, nil
```
bergundy, line 910: Can you put a 100ms sleep here (don't have to commit it) just to verify that time is not skipped while this is running?

bergundy (approved):
Had a couple of small comments but otherwise LGTM.

test/nexus_test.go
rodrigozhou: That's a good call actually. I added another test to timeout before the operation starts.

internal/internal_workflow_testsuite.go
```diff
@@ -2363,6 +2367,37 @@ func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(params executeNexu
 	}
 	env.runningNexusOperations[seq] = handle
 
+	var opID string
+	if params.options.ScheduleToCloseTimeout > 0 {
+		// Timer to fail the nexus operation due to schedule to close timeout.
```
cretz, line 2372: Hrmm, do we usually mimic timeouts like these in test suites? How does activity execution work in the test suite with regards to schedule to close timeout? Everything LGTM and will mark approved, just want to make sure this is normal for our test suite.

bergundy: Looks like some timeouts are broken in the test env (according to @rodrigozhou who looked into it). AFAIC, this is part of the behavior for nexus operations. It's enforced in the real server and the Java test server, don't see a good reason not to enforce it here.

cretz: I think we need to understand/confirm how timeouts are expected to work in the Go test suite for external calls. I admit being unfamiliar. Do activities in the test suite respect their timeouts? (they might, I just want to check)

internal/internal_workflow_testsuite.go
```diff
@@ -2887,6 +2923,10 @@ func (h *testNexusOperationHandle) completedCallback(result *commonpb.Payload, e
 // startedCallback is a callback registered to handle operation start.
 // Must be called in a postCallback block.
 func (h *testNexusOperationHandle) startedCallback(opID string, e error) {
+	if h.started {
```
Quinn-With-Two-Ns, line 2926: How can we get duplicate starts?

test/nexus_test.go
```diff
@@ -960,6 +1047,77 @@ func TestWorkflowTestSuite_WorkflowRunOperation_WithCancel(t *testing.T) {
 	}
 }
 
+func TestWorkflowTestSuite_NexusSyncOperation_ScheduleToCloseTimeout(t *testing.T) {
+	sleepDuration := 500 * time.Millisecond
+	op := temporalnexus.NewSyncOperation(
```
Quinn-With-Two-Ns, line 1052: A sync operation should also have a task timeout of 10s right? Are we enforcing that?

internal/internal_workflow_testsuite.go
Quinn-With-Two-Ns: When an activity is run by the workflow test environment (not mocked) we do propagate the timeout in the activities environment and rely on the activity to return to fail the operation which isn't how the real server works of course.

cretz: :+1: Works for me

cretz (approved):

test/nexus_test.go
bergundy: No. Good call, we should set the `Request-Timeout` header here as well.
I'll submit a PR for that.

internal/internal_workflow_testsuite.go
bergundy: If the timer and start request race.

test/nexus_test.go
bergundy: https://github.com/temporalio/sdk-go/pull/1653

PR #1637, opened 2024-09-16
Use stable CLI for tests

bergundy:
Addressing a TODO I put in the code when working on Nexus while it wasn't available in a stable server / CLI release.

Also:
- Use dashes instead of underscores in the test endpoint names since that logic changed since the server RC was cut.
- [Enable 1.25.0 tests with docker-compose](https://github.com/temporalio/sdk-go/pull/1637/commits/719c86a665adcee0f7cb3f1acae74b2c4bcd4453)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,13 +88,6 @@ jobs:
       - name: Docker compose - integration tests
         if: ${{ matrix.testDockerCompose }}
         run: go run . integration-test
-        env:
-          # TODO(bergundy): Remove this flag once server 1.25.0 is out.
-          DISABLE_NEXUS_TESTS: "1"
-          # TODO(bergundy): Remove this flag too once server 1.25.0 is out. Thanks Roey! :)
-          DISABLE_BACKLOG_STATS_TESTS: "1"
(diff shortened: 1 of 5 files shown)


Quinn-With-Two-Ns:
What about all the `DISABLE_SERVER_1_25_TESTS` code?

bergundy:
> What about all the `DISABLE_SERVER_1_25_TESTS` code? That was also added as part of the nexus PR and should be cleaned up no?
> 
> [50e993e](https://github.com/temporalio/sdk-go/commit/50e993ef971b840afb9f85bb6853fe4891dabebb)


They're mostly related to worker versioning apart from https://github.com/temporalio/temporal/issues/6319. I'll defer to the versioning crew for enabling those.

@ShahabT, @carlydf, @antlai-temporal

Quinn-With-Two-Ns:
And `DISABLE_NEXUS_TESTS` there is also a TODO under your name https://github.com/temporalio/sdk-go/blob/592ae12141a56d9c52a1cb40eb28fa06d5550254/.github/workflows/ci.yml#L93 hehe

cretz (approved):

PR #1645, opened 2024-09-20
Call the ShutdownWorker API as part of workflow worker cleanup

lina-temporal:
## What was changed
- The [`ShutdownWorker API`](https://github.com/temporalio/api/pull/443) is now called from the Go SDK as part of workflow worker shutdown.

## Why?
- `ShutdownWorker` is primarily used to signal to the Temporal Matching service that a given sticky task queue will no longer be polled. This lets Matching immediately direct tasks to the regular task queue when a workflow worker shuts down, rather than waiting for the sticky queue's timeout to elapse.

## Checklist
2. How was this tested:

- Updated tests/mocks
- `go test ./...`
- `cd internal/cmd/build && go run . check`

3. Any docs updates needed?
- No


internal/internal_worker.go
```diff
@@ -393,6 +393,12 @@ func (ww *workflowWorker) Start() error {
 
 // Stop the worker.
 func (ww *workflowWorker) Stop() {
+	// Best-effort cleanup, used to signal to the server that the sticky queue will
+	// no longer be polled. Errors are ignored.
+	defer func() {
+		_ = ww.poller.Cleanup()
```
Quinn-With-Two-Ns, line 399: `Stop` should not complete until this call is finished. The contract for stopping a worker is once a worker is stopped all resources associated with the worker should be cleaned up and the worker will no longer use the connection.

lina-temporal: `defer` is synchronous, `Stop` won't complete until `ww.poller.Cleanup()` does

Quinn-With-Two-Ns: You'll probably want to call this after all the pollers have stopped as well, otherwise couldn't a sticky poll request come in after you call Shutdown?

lina-temporal: That's why this call is in a `defer`; is the implication that `workflowTaskPoller` is still running after `workflowWorker`? It shouldn't be, since:

1. `baseWorker.Stop()` is synchronous, and responsible for new calls to `PollTask`, and
2. `close(bw.stopCh)` occurs within `baseWorker.Stop()`,
3. All polling requests are `select`'d against their results, as well as `bw.stopCh`, to immediately cancel the gRPC. Additionally, before a task is actually processed, `isStopping()` is checked, to account for the interleaving between polling and processing.

There's a slight interleaving where the gRPC call completes service side, right as a gRPC context `cancel` occurs client-side, but I'm not worried about that as an edge case, given that this is a best-effort optimization anyways.

Quinn-With-Two-Ns: > is the implication that workflowTaskPoller is still running after workflowWorker?

Sorry, I should have clarified a task worker could still respond to a sticky poll request, since your running this regardless of if the worker stop actually succeeded. It is probably pointless to make this call if the worker didn't stop right?

Quinn-With-Two-Ns: on that point if `WorkerStopTimeout` was exceeded while shutting down the worker we should probably just terminate and not make another RPC? I understand we expect this call to be quick, but my experience is in customer setups "quick" calls on Temporals end can take a while.

lina-temporal: > Sorry, I should have clarified a task worker could still respond to a sticky poll request, since your running this regardless of if the worker stop actually succeeded. It is probably pointless to make this call if the worker didn't stop right?

No, what we're trying to prevent here is the history service from pushing new tasks to a shutdown (or "doomed", if you will) poller. The worker can be in a few states, with regard to this cleanup call:

1. Not anywhere within the polling/processing loop (cleanup works because the poller won't run again after `Stop()` is called)
2. `select` blocking on the stop channel and gRPC result (cleanup works by virtue of the `stopC` being part of `select`)
3. Worker about to start polling, or start processing (cleanup works because the `workflowTaskPoller` beings polling and processing with checking if it `isStopping()`)
4. Worker is in the middle of processing a task outside of `select`'ing off of the `stopCh` - good the task will complete successfully, and it won't ever poll again; correct behavior.

lina-temporal: > on that point if WorkerStopTimeout was exceeded while shutting down the worker we should probably just terminate and not make another RPC? I understand we expect this call to be quick, but my experience is in customer setups "quick" calls on Temporals end can take a while.

We don't care. This is best effort.

Quinn-With-Two-Ns: Being best effort doesn't mean we can violate or ignore `WorkerStopTimeout` IMO.

l1na-forever: > Being best effort doesn't mean we can violate or ignore WorkerStopTimeout IMO.

Can you clarify? What exactly do you want to see changed?

lina-temporal: > Being best effort doesn't mean we can violate or ignore WorkerStopTimeout IMO.

Can you clarify? What exactly do you want to see changed?

Quinn-With-Two-Ns: >Worker is in the middle of processing a task outside of select'ing off of the stopCh - good the task will complete successfully, and it won't ever poll again; correct behavior.

So to be clear if a worker responds to task after this `Cleanup ` call the server won't schedule it on the sticky task queue?

lina-temporal: > So to be clear if a worker responds to task after this Cleanup call the server won't schedule it on the sticky task queue?

Correct, the worker's scheduling is based solely on last polled time (versus last task completion time).

cretz:
If at all possible, would like this to wait on https://github.com/temporalio/temporal/pull/6511 to be merged just to confirm we don't need any backwards-incompatible API changes.

internal/internal_worker.go
Quinn-With-Two-Ns: >Can you clarify? What exactly do you want to see changed?

My thought was to move this cleanup [here](https://github.com/temporalio/sdk-go/blob/d10e87118a07b44fd09bf88d39a628f0e6e70c34/internal/internal_worker_base.go#L583) and only run the shutdown function if the `stopTimeout` was not exceeded. I think this is how our other SDKs would also abort shutdown after timeout @cretz ?

cretz: > I think this is how our other SDKs would also abort shutdown after timeout @cretz ?

Didn't read thread, but yes should call worker shutdown _before_ graceful timeout starts IMO. Basically worker shutdown is:

1. Stop accepting new work
2. Wait graceful timeout period (default 0)
3. Cancel all outstanding activities and wait for all outstanding workflow tasks to complete
4. Wait for all outstanding activities to complete

So I agree, this new worker-shutdown API is probably best before 2 not after it.

Quinn-With-Two-Ns: ah `graceful timeout` is not what I am referring to here , I am talking about `stopTimeout ` that we define as 

```
WorkerStopTimeout is the time delay before hard terminate worker
```

internal/internal_task_pollers.go
```diff
@@ -303,6 +307,36 @@ func newWorkflowTaskPoller(
...
+// Best-effort attempt to indicate to Matching service that this workflow task
+// poller's sticky queue will no longer be polled. Should be called when the
+// poller is stopping. Failure to call ShutdownWorker is logged, but otherwise
+// ignored.
+func (wtp *workflowTaskPoller) Cleanup() error {
+	ctx := context.Background()
+	grpcCtx, cancel := newGRPCContext(ctx, grpcMetricsHandler(wtp.metricsHandler))
+	defer cancel()
+
+	_, err := wtp.service.ShutdownWorker(grpcCtx, &workflowservice.ShutdownWorkerRequest{
```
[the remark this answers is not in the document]
lina-temporal: > Is there any way for us to test in our integration tests that this works and has effect? Even if not now, can we open an issue for when it is available server side to confirm it has effect (e.g. that sticky was reset across two workers)? 

Not exactly. The call signals Matching to route tasks away from a shutdown worker's sticky queue (where the total timeout is 10 seconds). However, the caller that would actually see the effect from the rerouting isn't the end client, it's History service, which [handles the rerouting](https://github.com/temporalio/temporal/blob/99f15c4dabf9a22763aaa13653c83a832bd0fffa/service/history/transfer_queue_active_task_executor.go#L272-L287). The end result is that any future tasks previously destined for the sticky queue will have a lower schedule-to-start latency.

The important takeaway is that this isn't very observable from a testing point of view this far away in the stack - the signals between success and failure are fuzzy, it's just latency. If the call fails, the worst case scenario is that schedule-to-start is more latent (by up to 10s). If the call succeeds, we know schedule-to-start is hopefully less latent, but that isn't guaranteed by any means. `ShutdownWorker` could work perfectly, and the task's `ScheduleToStart` may be slow from an entirely unrelated reason in History or Matching services. 

> My fear is that we may do something wrong on the request building or something and this gRPC call always errors and we would just never know that it never worked.

Because the effect of this call takes place entirely within the service, asynchronously, IMO the testing needs to verify two independent facts:

- Does the worker call `ShutdownWorker` and receive positive acknowledgement from the service? The call doesn't have an effect directly visible to the worker, so here we instead need to verify that the server will validate and accept the request.
- Does the service process `ShutdownWorker` properly/does the call have an effect? Because we're testing deeper within the service, we get a strong discrete signal for if the call had an effect; I can directly call `MatchingService` and assert that it returns `StickyWorkerUnavailable` for a shut down queue. This is a much better signal for test success than merely trying to guess based on downstream latency.

internal/internal_worker.go
lina-temporal: > But we need to make sure on server side the "worker shutdown" does not mess with any tasks in flight now or in the future, just prevents new work.

It doesn't

> Didn't read thread, but yes should call worker shutdown before graceful timeout IMO. 

This will render `ShutdownWorker` useless if there's a chance pollers are still active. It *must* be called after the pollers have stopped - else their poll call will page the task queue partition back in. 

> WorkerStopTimeout is the time delay before hard terminate worker

You're still free to hard terminate the worker before/during/after `ShutdownWorker`. What we really want to avoid is calling `ShutdownWorker` before any outstanding polling calls are processed, since that'll cause the server to unload-reload the task queue.

Quinn-With-Two-Ns: Yeah so I would move the move the `ShutdownWorker` call [here](https://github.com/temporalio/sdk-go/blob/d10e87118a07b44fd09bf88d39a628f0e6e70c34/internal/internal_worker_base.go#L583) and only run the shutdown function if the stopTimeout was not exceeded since there is no easy way in the Go SDK to wait for just the pollers to finish.

internal/internal_task_pollers.go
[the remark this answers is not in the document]
cretz: :+1: I wonder if we can demonstrate a user-visible effect, e.g. such as having two workers, and the worker that got a sticky workflow is able to be shutdown and have the next workflow task immediately go to the next worker without penalty.

Regardless, I understand we cannot assert this behavior in integration tests yet anyways.

internal/internal_worker.go
lina-temporal: Move it to `baseWorker`? Then we lose the point of the `Cleanup` interface on pollers, no? I can't even access the `poller` variable that `workflowWorker` privately defines, then.

Quinn-With-Two-Ns: Don't see any issue with it being in the base worker , that was what I did in my [POC](https://github.com/Quinn-With-Two-Ns/sdk-go/commit/5251873b285de576771731089215f256e0f93044#diff-d91d342aa5b1a0bfdd2759edeace6058f9e3f782aa027019bc440c3d1922ec63R583) unless I am missing something? 

Note: My branch is incorrect because it is not waiting for the pollers to shutdown

lina-temporal: Oh, I didn't realize you could access the task worker from `bw.options`. Sure, I'll update.

lina-temporal: One thing I noticed after running the tests is that the default `WorkerStopTimeout` is `0`; that means that a `timer.After` on it will return immediately, preventing our cleanup goroutine from getting a chance to run. Should we set a default `WorkerStopTimeout` (or even enforce a tiny minimum, like 1-3s) for users?

cretz: I was considering `WorkerStopTimeout` as Go's version of graceful (at least that's what https://pkg.go.dev/go.temporal.io/sdk/internal#WorkerOptions.WorkerStopTimeout mentions), even if the last few steps of waiting are not the same. But yes, we should probably make the shutdown worker call before this timer starts (where we've canceled https://pkg.go.dev/go.temporal.io/sdk/activity#GetWorkerStopChannel but not yet activity context) since this shutdown worker API is meant to prevent new work which is the first step of worker shutdown.

Quinn-With-Two-Ns: >Should we set a default WorkerStopTimeout (or even enforce a tiny minimum, like 1-3s) for users?

Hm we can't start enforcing a minimum now, that would break uses who aren't setting one. We could add a default let me double check what other SDKs do.

Quinn-With-Two-Ns: Discussed offline I think your current approach is fine , apologize for wasting time on another path

internal/internal_task_pollers.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: @lina-temporal did you add any test to verify `ShutdownWoker` is only treated as best effort? I was expecting to see a unit test that mocked the service stub returned an error for `ShutdownWoker` and verify that from the users perspective the shutdown still succeeded. If I missed it please just point me at the test.

cretz (approved):
Marking approved because I made a comment or two on this PR and don't want there to be confusion that any of it is blocking. Please wait for @Quinn-With-Two-Ns's approval before merging.

internal/internal_task_pollers.go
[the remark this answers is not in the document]
lina-temporal: Sure, added a test to assert that `Cleanup` is best effort (`TestCleanupIsBestEffort`).

lina-temporal:
FYI, server implementation of the API has been merged.

Quinn-With-Two-Ns (approved):

PR #1652, opened 2024-09-27
MultiOperation retry non-durable Update

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/internal_workflow_client.go
```diff
@@ -1715,75 +1715,99 @@ func (w *workflowClientInterceptor) executeWorkflowWithOperation(
...
-	if errors.As(err, &multiErr) {
-		if len(multiErr.OperationErrors()) != len(multiRequest.Operations) {
-			return nil, fmt.Errorf("%w: %v instead of %v operation errors",
-				errInvalidServerResponse, len(multiErr.OperationErrors()), len(multiRequest.Operations))
+	var startResp *workflowservice.StartWorkflowExecutionResponse
+	var updateResp *workflowservice.UpdateWorkflowExecutionResponse
+	for {
+		multiResp, err := func() (*workflowservice.ExecuteMultiOperationResponse, error) {
+			grpcCtx, cancel := newGRPCContext(ctx, grpcTimeout(pollUpdateTimeout), grpcLongPoll(true), defaultGrpcRetryParameters(ctx))
+			defer cancel()
```
stephanos, line 1724: I'm not sure how to test this without waiting for 60s

internal/internal_workflow_client.go
```diff
@@ -1715,75 +1715,99 @@ func (w *workflowClientInterceptor) executeWorkflowWithOperation(
...
-	} else if err != nil {
-		return nil, err
-	}
 
-	if len(multiResp.Responses) != len(multiRequest.Operations) {
-		return nil, fmt.Errorf("%w: %v instead of %v operation results",
-			errInvalidServerResponse, len(multiResp.Responses), len(multiRequest.Operations))
+		if w.updateIsDurable(updateResp) {
+			break
+		}
```
stephanos, line 1802: This with the `for` loop addresses the retries.

internal/internal_workflow_client.go
```diff
@@ -1715,75 +1715,99 @@ func (w *workflowClientInterceptor) executeWorkflowWithOperation(
...
+			multiResp, err := w.client.workflowService.ExecuteMultiOperation(grpcCtx, &multiRequest)
+			if err != nil {
+				if ctx.Err() != nil {
+					return nil, NewWorkflowUpdateServiceTimeoutOrCanceledError(err)
+				}
+				if status := serviceerror.ToStatus(err); status.Code() == codes.Canceled || status.Code() == codes.DeadlineExceeded {
+					return nil, NewWorkflowUpdateServiceTimeoutOrCanceledError(err)
+				}
+				return nil, err
+			}
```
stephanos, line 1735: This is for the timeout.

internal/internal_workflow_client.go
```diff
@@ -1715,75 +1715,99 @@ func (w *workflowClientInterceptor) executeWorkflowWithOperation(
...
+				case *workflowservice.ExecuteMultiOperationRequest_Operation_UpdateWorkflow:
+					if !errors.As(opErr, &abortedErr) {
+						startErr = fmt.Errorf("%w: %w", errInvalidWorkflowOperation, opErr)
+					}
+				default:
+					// this would only happen if a case statement for a newly added operation is missing above
+					return nil, fmt.Errorf("%w: %T", errUnsupportedOperation, t)
+				}
+			}
+			return nil, startErr
```
cretz, line 1768: Can `startErr` be `nil` here? Even if server-side contract guarantees not, we might as well set a default somehow.

stephanos: Right, if the server misbehaves, it could be `nil`. I'll set a default 👍

Quinn-With-Two-Ns (approved):

PR #1653, opened 2024-10-01
Enforce nexus request timeout in workflow test suite

bergundy:
Also fixes issue where headers injected by workflow outbound interceptors in the test env were not normalized to lower case as done by the real server.


internal/internal_workflow_testsuite.go
```diff
@@ -2354,6 +2354,17 @@ func (env *testWorkflowEnvironmentImpl) newTestNexusTaskHandler() *nexusTaskHand
...
+	// Use lower case header values to simulate how the Nexus SDK (used internally by the "real" server) would transmit
+	// these headers over the wire.
+	nexusHeader := make(map[string]string, len(params.nexusHeader))
+	for k, v := range params.nexusHeader {
+		nexusHeader[strings.ToLower(k)] = v
+	}
+	params.nexusHeader = nexusHeader
+	// The real server allows requests to take up to 10 seconds, mimic that behavior here.
+	// Note that if a user sets the Request-Timeout header, it gets overridden.
+	params.nexusHeader[strings.ToLower(nexus.HeaderRequestTimeout)] = "10s"
```
rodrigozhou, line 2366: We also set the request timeout for the cancel operation.

bergundy: Yes, this will set that timeout for all "requests".

Quinn-With-Two-Ns:
Seems like there is an issue with this PR only on Windows

bergundy:
> Seems like there is an issue with this PR only on Windows

Changed the test slightly, CI should pass now AFAICT.

bergundy:
Looks like this is good to go @Quinn-With-Two-Ns.

internal/internal_workflow_testsuite.go
```diff
@@ -2358,6 +2358,17 @@ func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(
...
```
Quinn-With-Two-Ns, line 2370: Isn't the real server behaviour the minimum of 10s or the remaining schedule to close?

bergundy: Good point. Not at the moment, but we may enforce that eventually.

Quinn-With-Two-Ns (approved):

PR #1655, opened 2024-10-02
Error for unused Update operation

stephanos:
## What was changed

When the request for the update operation's handle times out, it now returns a more specific error if the operation was never actually executed as part of a MultiOperation.

## Why?

It's easy to forget to specify `WithStartOperation` (happened to me when I worked on a Go sample). This signals more clearly to the user that they forgot to run it. However, IMO the error cannot be more specific than that, since there _could_ be other reasons why the operation was never executed.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -1083,7 +1083,7 @@ func (op *UpdateWithStartWorkflowOperation) Get(ctx context.Context) (WorkflowUp
 		return op.handle, op.err
 	case <-ctx.Done():
 		if !op.executed.Load() {
-			return nil, fmt.Errorf("%v: %v", ctx.Err(), fmt.Errorf("operation was not executed"))
+			return nil, fmt.Errorf("%w: %w", ctx.Err(), fmt.Errorf("operation was not executed"))
 		}
 		return nil, ctx.Err()
 	}
--- a/internal/client.go
+++ b/internal/client.go
@@ -746,7 +746,7 @@ type (
 		// workflow completion callback. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
 		callbacks []*commonpb.Callback
 		// links. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
-		links     []*commonpb.Link
+		links []*commonpb.Link
 	}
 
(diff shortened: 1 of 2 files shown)


internal/client.go
```diff
@@ -746,7 +746,7 @@ type (
 		// workflow completion callback. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
 		callbacks []*commonpb.Callback
 		// links. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
-		links     []*commonpb.Link
+		links []*commonpb.Link
```
stephanos, line 749: Huh? Seems correct now.

internal/internal_workflow_client_test.go
```diff
@@ -1027,6 +1027,35 @@ func (s *workflowRunSuite) TestExecuteWorkflowWithUpdate_Retry() {
...
+		StartWorkflowOptions{
+			ID:        workflowID,
+			TaskQueue: taskqueue,
+			// WithStartOperation is not specified!
+		}, workflowType,
+	)
+	require.NoError(s.T(), err)
+
+	_, err = updOp.Get(ctxWithTimeout)
+	require.EqualError(s.T(), err, "context deadline exceeded: operation was not executed")
```
stephanos, line 1056: I'm assuming the cause of the deadline should be on the right.

internal/client.go
```diff
@@ -1082,6 +1082,9 @@ func (op *UpdateWithStartWorkflowOperation) Get(ctx context.Context) (WorkflowUp
 	case <-op.doneCh:
 		return op.handle, op.err
 	case <-ctx.Done():
+		if !op.executed.Load() {
+			return nil, fmt.Errorf("%v: %v", ctx.Err(), fmt.Errorf("operation was not executed"))
```
Quinn-With-Two-Ns, line 1086: for error wrapping shouldn't one of these be a `%w`?

stephanos: Indeed! I just learned that Go 1.20 supports multiple `%w`, changing it to that now.

Quinn-With-Two-Ns (approved):

PR #1656, opened 2024-10-02
Remove history truncation

Quinn-With-Two-Ns:
Remove history truncation. As far back in the Go SDK as I can find we have this code to try to truncate any events past `atWorkflowTaskCompletedEventID`. It is not clear what the point of this code is and it has a number of bugs. Some of the issues are:
* It can panic if `h.Events[0].GetEventId() > atWorkflowTaskCompletedEventID`
* It is trying to check if the last event in history after truncation is a `EVENT_TYPE_WORKFLOW_TASK_COMPLETED`, but `atWorkflowTaskCompletedEventID` is from the workflow tasks `GetStartedEventId` so it should be a `EVENT_TYPE_WORKFLOW_TASK_STARTED`.
* If history does go beyond `GetStartedEventId` then that indicated this workflow task is stale and will be rejected by the server anyway

Note: The cadence team also pointed out the strangeness of this code https://github.com/uber-go/cadence-client/blob/4d4c09f5053b53caea43100e170c327058731f5c/internal/internal_task_pollers.go#L961

### DIFF
--- a/internal/internal_task_handlers_test.go
+++ b/internal/internal_task_handlers_test.go
@@ -2725,7 +2725,6 @@ func TestHistoryIterator(t *testing.T) {
 				WorkflowId: "test-workflow-id",
 				RunId:      "test-run-id",
 			},
-			3,
 			metrics.NopHandler,
 			"test-task-queue",
 		),
--- a/internal/internal_task_handlers_test.go
+++ b/internal/internal_task_handlers_test.go
@@ -2688,3 +2688,52 @@ func TestResetIfDestroyedTaskPrep(t *testing.T) {
 		requireContainsMsgWithID(t, task.Messages, wftNewMsgID)
 	})
 }
+
+func TestHistoryIterator(t *testing.T) {
+	testEvents := []*historypb.HistoryEvent{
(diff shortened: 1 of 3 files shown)


cretz:
> If history does go beyond `GetStartedEventId` then that indicated this workflow task is stale and will be rejected by the server anyway

What if there is a signal sent between when we first got partial history from `PollWorkflowTaskQueueResponse` and started iterating to get the rest? Wouldn't then the signal be past the last task completed we expected? I wonder if this can be somehow tested by having a history too large to fit in `PollWorkflowTaskQueueResponse.history` and have first history page iteration be held up, then send a signal, then let history iteration continue.

Quinn-With-Two-Ns:
>Wouldn't then the signal be past the last task completed we expected

When we are processing a new workflow task we never expect any event past the last workflow task started event because the server **will** buffer the signal

cretz:
> When we are processing a new workflow task we never expect any event past the last workflow task started event because the server will buffer the signal

:+1: So to confirm, server guarantees that if a worker is processing a task, and that task is still valid, fetching history will never contain any events after the task being processed?

cretz (approved):

PR #1657, opened 2024-10-03
Passthrough if a Nexus sync handler returns a Payload

bergundy:
~~Matches behavior of workflows and activities.~~

EDIT: I noticed we only support returning payloads from activities, we should consider allowing returning payloads from workflows too for consistency.


bergundy:
Decided not to go with this and come up with a better story for passthrough payloads.

PR #1658, opened 2024-10-03
Ensure WorkflowTaskExecutionFailureCounter is called with a tag

### LINKED ISSUE #1450: Metric `workflow_task_execution_failed` is used with different sets of labels causing errors with Prometheus

yuandrew:
## What was changed
Ensure the WorkflowTaskExecutionFailureCounter metric always has a tag set included.

## Why?

Prometheus returns an error when you try incrementing the same counter if one instance has a tag set and the other didn't.

I validated that Prometheus is okay with different tag sets (in this case if we happen to call one instance with `WorkflowError` and the other with `NonDeterminismError` tag sets). I'm not sure there's a scenario where we increment this counter from different sources, but if there is, this shouldn't be an issue with Prometheus.

## Checklist

1. Closes #1450

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -1170,7 +1170,7 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	}
 	defer func() {
 		if p := recover(); p != nil {
-			weh.metricsHandler.Counter(metrics.WorkflowTaskExecutionFailureCounter).Inc(1)
+			incrementWorkflowTaskFailureCounter(weh.metricsHandler, "NonDeterminismError")
 			topLine := fmt.Sprintf("process event for %s [panic]:", weh.workflowInfo.TaskQueueName)
 			st := getStackTraceRaw(topLine, 7, 0)
 			weh.Complete(nil, newWorkflowPanicError(p, st))
@@ -1373,7 +1373,7 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessMessage(
 ) error {
 	defer func() {
 		if p := recover(); p != nil {
-			weh.metricsHandler.Counter(metrics.WorkflowTaskExecutionFailureCounter).Inc(1)
+			incrementWorkflowTaskFailureCounter(weh.metricsHandler, "NonDeterminismError")
 			topLine := fmt.Sprintf("process message for %s [panic]:", weh.workflowInfo.TaskQueueName)
 			st := getStackTraceRaw(topLine, 7, 0)
 			weh.Complete(nil, newWorkflowPanicError(p, st))
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -471,7 +471,7 @@ func (wtp *workflowTaskPoller) RespondTaskCompletedWithMetrics(
 		if failWorkflowTask.Cause == enumspb.WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR {
 			failureReason = "NonDeterminismError"
 		}
-		metricsHandler.WithTags(metrics.WorkflowTaskFailedTags(failureReason)).Counter(metrics.WorkflowTaskExecutionFailureCounter).Inc(1)
+		incrementWorkflowTaskFailureCounter(metricsHandler, failureReason)
 		completedRequest = failWorkflowTask
 	}
 
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -1960,3 +1960,7 @@ func (s *semaphoreImpl) Release(n int64) {
 		panic("Semaphore.Release() released more than held")
 	}
 }
+
+func incrementWorkflowTaskFailureCounter(metricsHandler metrics.Handler, failureReason string) {
+	metricsHandler.WithTags(metrics.WorkflowTaskFailedTags(failureReason)).Counter(metrics.WorkflowTaskExecutionFailureCounter).Inc(1)
+}


cretz (approved):

internal/internal_event_handlers.go
```diff
@@ -1170,7 +1170,7 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	}
 	defer func() {
 		if p := recover(); p != nil {
-			weh.metricsHandler.Counter(metrics.WorkflowTaskExecutionFailureCounter).Inc(1)
+			incrementWorkflowTaskFailureCounter(weh.metricsHandler, "NonDeterminismError")
```
Quinn-With-Two-Ns, line 1173: I don't think the panics here are always caused just by non determinism, what is users workflow code panics because of a code bug?

Quinn-With-Two-Ns: Also I am pretty sure we can still end up emitting emitting this metric twice , that should be addressed to close https://github.com/temporalio/sdk-go/issues/1450

yuandrew: > I don't think the panics here are always caused just by non determinism, what is users workflow code panics because of a code bug?

From my testing, when a user's workflow code panics, we trigger the call in `internal_task_pollers.go`. The only way I saw us recover a panic from the defer call is when we `panicIllegalState` from the state machine. But to be safe, I can add a check to check the panic message for `TMPRL1100` to use `NonDeterminismError` and default to `WorkflowError` for all other scenarios.

> Also I am pretty sure we can still end up emitting emitting this metric twice , that should be addressed to close https://github.com/temporalio/sdk-go/issues/1450

Do you have any idea what scenario could cause this double logging? I've tried to hit this scenario the last few days and haven't been able to come up with anything. 

The issue itself is on Prometheus erroring out due to the metric being emitted twice, once due to having a tag and the other without, which is now resolved with this PR (calling with different logs doesn't cause Prometheus to throw an error). But I'm happy to address this double logging, I'm just not able to find a scenario where we double log.

Quinn-With-Two-Ns: Sure let me try to come up with a counter example

yuandrew: I got a scenario that hits both `ProcessEvent` and `RespondTaskCompletedWithMetrics`. It looks like when both of these functions emit this metric, only the `RespondTaskCompletedWithMetrics` inc gets counted. 

A few scenarios I ran:
1. When `RespondTaskCompletedWithMetrics` is the only one emitting the metric, the count is 1
2. When `ProcessEvent` is the only one emitting the metric, the count is 0
3. When neither emit the metric, the metric doesn't get emitted to http://localhost:9090/metrics altogether.

Is there some kind of scoping or something that's causing the earlier `ProcessEvent` metric to get reset? I'm using the metrics sample to emit metrics to localhost://9090/metrics https://github.com/yuandrew/playground/tree/duplicate_logging

PR #1659, opened 2024-10-04
Nexus: Fix link not being attached to a workflow started via a Handler function

bergundy:
Not testing this since server 1.25.0 didn't get support for links.

### DIFF
--- a/temporalnexus/operation.go
+++ b/temporalnexus/operation.go
@@ -223,7 +223,7 @@ func (o *workflowRunOperation[I, O]) Start(
 		}
 		return &nexus.HandlerStartOperationResultAsync{
 			OperationID: handle.ID(),
-			Links:       []nexus.Link{handle.Link()},
+			Links:       []nexus.Link{handle.link()},
 		}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


temporalnexus/operation.go
```diff
@@ -262,11 +252,14 @@ type WorkflowHandle[T any] interface {
 	ID() string
 	// ID is the workflow's run ID.
 	RunID() string
+	// Link to the WorkflowExecutionStarted event of the workflow represented by this handle.
+	Link() nexus.Link
```
Quinn-With-Two-Ns, line 256: hm why does this need to be on the interface? Are users expected to use this method?

Quinn-With-Two-Ns:
I understand if we can't write an end to end test for this, but what is stopping us from unit testing this change?

temporalnexus/operation.go
bergundy: It doesn't but I don't think it hurts.

Another approach would be to put this on the struct implementation and have an unexported method on the interface to ensure users don't try to implement it. This is important since the handler's return value is the interface.

bergundy:
> I understand if we can't write an end to end test for this, but what is stopping us from unit testing this change?

We'll add integration tests soon. server `v1.25.1` already supports links so it's just a matter of upgrading the server version in the CLI.

temporalnexus/operation.go
bergundy: I'll change to that alternative.

bergundy: Done.

Quinn-With-Two-Ns (approved):

cretz (approved):

PR #1660, opened 2024-10-08
Include updateID and updateName in update logger

Quinn-With-Two-Ns:
Include updateID and updateName in update logger.

closes https://github.com/temporalio/sdk-go/issues/1648

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -278,7 +278,11 @@ func GetCurrentUpdateInfo(ctx Context) *UpdateInfo {
 	return internal.GetCurrentUpdateInfo(ctx)
 }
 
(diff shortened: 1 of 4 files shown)


cretz (approved):

workflow/workflow.go
```diff
@@ -278,7 +278,11 @@ func GetCurrentUpdateInfo(ctx Context) *UpdateInfo {
 	return internal.GetCurrentUpdateInfo(ctx)
 }
 
-// GetLogger returns a logger to be used in workflow's context
+// GetLogger returns a logger to be used in workflow's context.
+// This logger does not record logs during replay.
```
yuandrew, line 282: 😊

PR #1661, opened 2024-10-08
Panic if endpoint or service is empty in NewNexusClient

bergundy:


internal/workflow.go
```diff
@@ -2401,6 +2401,12 @@ type nexusClient struct {
 //
 // NOTE: Experimental
 func NewNexusClient(endpoint, service string) NexusClient {
+	if endpoint == "" {
```
Quinn-With-Two-Ns, line 2404: Just as a sanity check the server does not allow an empty endpoint name to be registered or an empty service right?

cretz (approved):

internal/workflow.go
bergundy: Yes, we verify that.

Quinn-With-Two-Ns (approved):

internal/workflow.go
```diff
@@ -2401,6 +2401,12 @@ type nexusClient struct {
...
+		panic("endpoint must not be empty")
+	}
+	if service == "" {
+		panic("service must not be empty")
+	}
 	return nexusClient{endpoint, service}
```
Quinn-With-Two-Ns, line 2410: You could argue this validation should be post-interceptor, but I am fine with it as is for now

bergundy: Hmm... yeah, potentially, but I think getting the line number in the stack trace is better for debugging.

PR #1662, opened 2024-10-08
Error if history contains unexpected events after the StartedEventId

Quinn-With-Two-Ns:
Error if history contains unexpected events after the StartedEventId

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -138,11 +138,14 @@ type (
 	}
 
 	historyIteratorImpl struct {
-		iteratorFunc   func(nextPageToken []byte) (*historypb.History, []byte, error)
-		execution      *commonpb.WorkflowExecution
-		nextPageToken  []byte
-		namespace      string
-		service        workflowservice.WorkflowServiceClient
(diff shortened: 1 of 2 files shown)


internal/internal_task_pollers.go
```diff
@@ -941,6 +947,19 @@ func newGetHistoryPageFunc(
...
 		}
+
+		size := len(h.Events)
+		// While the SDK is processing a workflow task, the workflow task could timeout and server would start
+		// a new workflow task or the server looses the workflow task if it is a speculative workflow task. In either
+		// case, the new workflow task could have events that are beyond the last event ID that the SDK expects to process.
+		// In such cases, the SDK should return error indicating that the workflow task is stale since the result will not be used.
+		if size > 0 && lastEventID > 0 && 
+			h.Events[size-1].GetEventId() > lastEventID {
+			return nil, nil, fmt.Errorf("history contains events past expected last event ID (%v) "+
```
cretz, line 958: Can you confirm that an error iterating here is treated as task failure (eviction and all of that)?

Quinn-With-Two-Ns: Yes

cretz (approved):

PR #1664, opened 2024-10-08
Fix nexus_task_execution_failed to include OperationError outcome in start requests

bergundy:
## Why?

Consistency with other failure metrics.

### DIFF
--- a/internal/internal_nexus_task_poller.go
+++ b/internal/internal_nexus_task_poller.go
@@ -152,7 +152,12 @@ func (ntp *nexusTaskPoller) ProcessTask(task interface{}) error {
 
 	// Execution latency (in-SDK processing time).
 	metricsHandler.Timer(metrics.NexusTaskExecutionLatency).Record(time.Since(executionStartTime))
-	if err != nil || failure != nil {
+
+	// Increment failure in all forms of errors:
+	// Internal error processing the task.
+	// Failure from user handler.
+	// Special case for the start response with operation error.
+	if err != nil || failure != nil || res.Response.GetStartOperation().GetOperationError() != nil {
 		metricsHandler.Counter(metrics.NexusTaskExecutionFailedCounter).Inc(1)
 	}
 


Quinn-With-Two-Ns:
Do we not have any tests for metrics emitted that need to be updated?

bergundy:
> Do we not have any tests for metrics emitted that need to be updated?

No tests for metrics in this case. I guess I can add some.

internal/internal_nexus_task_poller.go
```diff
@@ -152,7 +152,12 @@ func (ntp *nexusTaskPoller) ProcessTask(task interface{}) error {
 
 	// Execution latency (in-SDK processing time).
 	metricsHandler.Timer(metrics.NexusTaskExecutionLatency).Record(time.Since(executionStartTime))
-	if err != nil || failure != nil {
+
+	// Increment failure in all forms of errors:
+	// Internal error processing the task.
+	// Failure from user handler.
+	// Special case for the start response with operation error.
```
cretz, line 159: Is this a task failure if the process task call doesn't return a failure? I wouldn't expect failure to start things like a child workflow to increment this value.

bergundy: This is when a handler returns an UnsuccessfulOperationError. A user would consider it a failure.

Quinn-With-Two-Ns: I'd be fine renaming the metric to something like `temporal_nexus_operation_execution_failed ` not sure why we need two metrics if activity only needs one

bergundy: My issue with your suggestion is that there are other ways that the execution would fail, e.g. by returning a non retryable error in your handler or the workflow in the workflow run operation fails asynchronously.
For activities we don't distinguish between retryable and non retryable errors, so I figured this model is good enough.
What we have now maps to whether a handler method returned an error or not (or there was an internal error within the SDK).

cretz: Can you clarify the definition of `temporal_nexus_task_execution_failed`? Is it "any task that fails unexpectedly or sync operation start fails expectedly"? If so, it is confusing to a user that sync start operation RPC returning an operation error causes this counter to increase but async get operation result returning an operation error does not.

bergundy: Definition is, it's incremented if any of the following are true:
1. The user defined handler returns any error (or panics)
2. There's an internal error in the SDK (e.g. bug) while processing the task

cretz: > The user defined handler returns any error (or panics)

Does this include errors on `GetResult`? Do we really want a metric that can't differentiate between a proper failure and an internal one? Or should we not count operation failures?

bergundy: We haven't implemented `GetResult` but I think we could exclude `ErrOperationStillRunning`. This change includes `UnsuccessfulOperationError` in the metric, where it didn't used to be included before.

The rationale is that all of `HandlerErrorTypeBadRequest`, non-retryable `ApplicationError`, and `UnsuccessfulOperationError` are non-retryable errors from the caller perspective and essentially fail the operation.

bergundy: But I agree that it's not perfect. It wasn't perfect before either.

Quinn-With-Two-Ns (approved):

internal/internal_nexus_task_poller.go
cretz: I think the metrics should be split between user-returned operation errors and unexpected failures similar to workflows. People often just stop using activity failure metrics because they are using intentional failures for some things. They have to create their own metrics to differentiate. The difference is more obvious in other languages where unexpected failures/exceptions are more common and the languages are used to differentiating explicit RPC failure from accidental one.

bergundy:
Discussed offline, we'll add a label to this metric to help distinguish between the different error types in a followup PR.
I can do a fast follow next week.

PR #1666, opened 2024-10-10
Support for mocking nexus operations

rodrigozhou:
## What was changed

Support for mocking nexus operations in the test framework.
The user can mock using either the `nexus.Operation` itself, or using `nexus.OperationReference` through the `nexus.NewOperationReference` helper. See the docs in the `OnNexusOperation` for description and example.

## Why?

Be able to mock nexus operation without needing to have access to the operation implementation itself or creating dummy operations.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -3194,8 +3194,8 @@ func (r *testNexusHandler) StartOperation(
 			service,
 		))
 	}
-	op, ok := s.GetOperation(operation)
-	if !ok {
+	op := s.Operation(operation)
+	if op == nil {
 		panic(fmt.Sprintf(
 			"nexus operation %q is not registered in service %q with the TestWorkflowEnvironment",
 			operation,
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -2552,6 +2552,11 @@ func (env *testWorkflowEnvironmentImpl) RegisterNexusAsyncOperationCompletion(
 			return encodeErr
 		}
 	}
+
+	// Getting the locker to prevent race condition if this function is called while
+	// the test env is already running.
+	env.locker.Lock()
+	defer env.locker.Unlock()
 	env.setNexusAsyncOperationCompletionHandle(
 		service,
 		operation,
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -112,6 +112,13 @@ type (
 		done            bool
 		onCompleted     func(*commonpb.Payload, error)
 		onStarted       func(opID string, e error)
+		isMocked        bool
+	}
+
+	testNexusAsyncOperationHandle struct {
+		result *commonpb.Payload
+		err    error
+		delay  time.Duration
 	}
 
 	testCallbackHandle struct {
@@ -152,6 +159,7 @@ type (
 
 		workflowMock              *mock.Mock
 		activityMock              *mock.Mock
+		nexusMock                 *mock.Mock
 		service                   workflowservice.WorkflowServiceClient
 		logger                    log.Logger
 		metricsHandler            metrics.Handler
@@ -172,11 +180,14 @@ type (
 		timers                 map[string]*testTimerHandle
 		runningWorkflows       map[string]*testWorkflowHandle
 		runningNexusOperations map[int64]*testNexusOperationHandle
+		nexusAsyncOpHandle     map[string]*testNexusAsyncOperationHandle
+		nexusOperationID2SeqID map[string]int64
 
 		runningCount int
 
 		expectedWorkflowMockCalls map[string]struct{}
 		expectedActivityMockCalls map[string]struct{}
+		expectedNexusMockCalls    map[string]struct{}
 
 		onActivityStartedListener        func(activityInfo *ActivityInfo, ctx context.Context, args converter.EncodedValues)
 		onActivityCompletedListener      func(activityInfo *ActivityInfo, result converter.EncodedValue, err error)
@@ -259,10 +270,13 @@ func newTestWorkflowEnvironmentImpl(s *WorkflowTestSuite, parentRegistry *regist
 			localActivities:           make(map[string]*localActivityTask),
 			runningWorkflows:          make(map[string]*testWorkflowHandle),
 			runningNexusOperations:    make(map[int64]*testNexusOperationHandle),
+			nexusAsyncOpHandle:        make(map[string]*testNexusAsyncOperationHandle),
+			nexusOperationID2SeqID:    make(map[string]int64),
 			callbackChannel:           make(chan testCallbackHandle, 1000),
 			testTimeout:               3 * time.Second,
 			expectedWorkflowMockCalls: make(map[string]struct{}),
 			expectedActivityMockCalls: make(map[string]struct{}),
+			expectedNexusMockCalls:    make(map[string]struct{}),
 		},
 
 		workflowInfo: &WorkflowInfo{
@@ -1845,6 +1859,9 @@ func (w *workflowExecutorWrapper) Execute(ctx Context, input *commonpb.Payloads)
 }
 
 func (m *mockWrapper) getCtxArg(ctx interface{}) []interface{} {
+	if m.fn == nil {
+		return nil
+	}
 	fnType := reflect.TypeOf(m.fn)
 	if fnType.NumIn() > 0 {
 		if (!m.isWorkflow && isActivityContext(fnType.In(0))) ||
@@ -1871,6 +1888,23 @@ func (m *mockWrapper) getWorkflowMockReturn(ctx interface{}, input *commonpb.Pay
 	}
 
 	return m.getMockReturn(ctx, input, m.env.workflowMock)
+}
+
+func (m *mockWrapper) getNexusMockReturn(
+	ctx interface{},
+	operation string,
+	input interface{},
+	options interface{},
+) (retArgs mock.Arguments) {
+	if _, ok := m.env.expectedNexusMockCalls[m.name]; !ok {
+		// no mock
+		return nil
+	}
+	return m.getMockReturnWithActualArgs(
+		ctx,
+		[]interface{}{operation, input, options},
+		m.env.nexusMock,
+	)
 }
 
 func (m *mockWrapper) getMockReturn(ctx interface{}, input *commonpb.Payloads, envMock *mock.Mock) (retArgs mock.Arguments) {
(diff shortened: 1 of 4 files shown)


internal/workflow_testsuite.go
```diff
@@ -543,6 +544,63 @@ func (e *TestWorkflowEnvironment) OnUpsertMemo(attributes interface{}) *MockCall
...
+) *MockCallWrapper {
+	var call *mock.Call
+	s, ok := e.impl.registry.nexusServices[service]
+	if !ok {
+		panic(fmt.Sprintf(
+			"nexus service %q is not registered with the TestWorkflowEnvironment",
+			service,
+		))
+	}
+	if _, ok := s.GetOperation(operation); !ok {
```
cretz, line 561: I am not sure you should have to pre-register an operation to mock it any more than you have to register an activity to mock it.

internal/workflow_testsuite.go
```diff
@@ -543,6 +544,63 @@ func (e *TestWorkflowEnvironment) OnUpsertMemo(attributes interface{}) *MockCall
 	return e.wrapWorkflowCall(call)
 }
 
+func (e *TestWorkflowEnvironment) OnNexusOperation(
+	service string,
+	operation string,
```
cretz, line 549: Arguably this should accept a func the same way mocking an activity does and when a func is present instead of a string you should not have to have pre-registered it.

bergundy:
This PR could use some tests and there would be an example in the documentation showing how this may be used. I want to understand the experience you had in mind better before diving deeper into the review.

internal/internal_workflow_testsuite.go
```diff
@@ -2483,9 +2533,66 @@ func (env *testWorkflowEnvironmentImpl) RequestCancelNexusOperation(seq int64) {
 	}
 }
 
+func (env *testWorkflowEnvironmentImpl) RegisterNexusAsyncOperationCompletion(
+	service string,
+	operation string,
+	operationID string,
```
bergundy, line 2539: I wonder if this would be intuitive for users to specify the operationID at registration time.
Also don't we need locks to protect the map here?

internal/internal_workflow_testsuite.go
```diff
@@ -2970,3 +3150,161 @@ func (h *testNexusOperationHandle) cancel() {
...
+		))
+	}
+	fn, _ := reflect.TypeOf(op).MethodByName("Start")
+	inputType := fn.Type.In(2)
+	ptr := reflect.New(inputType)
+	if err := input.Consume(ptr.Interface()); err != nil {
+		panic(fmt.Sprintf("mock of ExecuteNexusOperation failed to deserialize input"))
+	}
+
+	// rebuild the input as *nexus.LazyValue
```
bergundy, line 3212: Why?

internal/internal_workflow_testsuite.go
rodrigozhou: > I wonder if this would be intuitive for users to specify the operationID at registration time.
The user needs to specify the operation id when specifying the return of the mock call.

> Also don't we need locks to protect the map here?
Why do we need to protect this map?

internal/internal_workflow_testsuite.go
rodrigozhou: Because, after I call `input.Consume` on L3208, it closed the reader, so it can't call again, which is needed if the user didn't mock the operation.

rodrigozhou: Well, actually, I can see that the `reader` isn't really being used at this moment, so maybe I could just skip this...

cretz (approved):
LGTM, but would wait until @Quinn-With-Two-Ns takes a look

internal/workflow_testsuite.go
```diff
@@ -543,6 +544,143 @@ func (e *TestWorkflowEnvironment) OnUpsertMemo(attributes interface{}) *MockCall
...
+//	)
+//
+//	t.RegisterNexusAsyncOperationCompletion(
+//		service,
+//		HelloOperation.Name(),
+//		"hello-operation-id",
+//		HelloOutput{Message: "Hello Temporal"},
+//		nil,
+//		1*time.Second,
+//	)
```
rodrigozhou, line 598: In addition to this, maybe we could have helper functions to mock sync and async operations. Example: for async operation, we could have something like this:
```
func (e *TestWorkflowEnvironment) MockNexusAsyncOperation(
	service any,
	operation any,
	input any,
	options any,
	result any,
	err error,
	delay time.Duration,
) *MockCallWrapper {
	operationID := uuid.New() // maybe also input?
	m := e.OnNexusOperation(service, operation, input, options).Return(
		&nexus.HandlerStartOperationResultAsync{
			OperationID: operationID,
		},
		nil,
	)
	e.RegisterNexusAsyncOperationCompletion(service, operation, operationID, result, err, delay)
	return m
}
```

And we could also have something similar for the sync which basically abstract the returned value type `nexus.HandlerStartOperationResultSync`.

internal/internal_workflow_testsuite.go
```diff
@@ -2967,6 +3161,183 @@ func (h *testNexusOperationHandle) cancel() {
 				h.completedCallback(nil, fmt.Errorf("operation cancelation handler failed: %v", failure.GetError().GetFailure().GetMessage()))
 			}
 			h.env.runningCount--
+			if h.env.onNexusOperationCanceledListener != nil {
+				h.env.onNexusOperationCanceledListener(h.params.client.Service(), h.params.operation)
+			}
```
rodrigozhou, line 3166: I'm not sure how useful is the canceled listener here. It doesn't seem that the user can cancel a Nexus operation through the SDK.

internal/workflow_testsuite.go
bergundy: Alternatively, if a user sets `After` on the returned `MockCallWrapper`, it sort of makes the operation async already (with the exception of generating a separate started event).
Can you config `After` worker with the `MockCallWrapper` returned here?

bergundy (commented):
Overall LGTM. There's just one conversation to resolve and this would be good to merge AFAIC.

internal/internal_workflow_testsuite.go
bergundy: Yeah, you can avoid this for now since we control both implementations. Might want to just put a comment here but either way it's not critical.

internal/internal_workflow_testsuite.go
bergundy: I think this is helpful for users to assert that an operation cancelation was requested by the workflow.

internal/internal_workflow_testsuite.go
bergundy: I can see users registering a completion handler while the test env is running so I would like to avoid data races in those cases.

internal/workflow_testsuite.go
rodrigozhou: Yes, you can, but as you said, it would not represent the actual behavior of async operation.

internal/internal_workflow_testsuite.go
rodrigozhou: I'd rather leave as it to avoid forgetting about this later if things change.

internal/workflow_testsuite.go
```diff
@@ -66,6 +66,7 @@ type (
 	TestWorkflowEnvironment struct {
 		workflowMock mock.Mock
 		activityMock mock.Mock
+		nexusMock    mock.Mock
```
Quinn-With-Two-Ns, line 69: It doesn't look like you are calling `AssertExpectations` on this mock like we do for `workflowMock` or `activityMock` is that correct? I think you also didn't add a `AssertNexusOperationCalled` or update the other `Assert*Called`

Quinn-With-Two-Ns (changes requested):
I don't think this PR handles expectation failure , If I am wrong then could we add a test to verify.

rodrigozhou:
@Quinn-With-Two-Ns I added `AssertNexusOperation*` functions.

internal/workflow_testsuite.go
```diff
@@ -999,10 +1167,10 @@ func (e *TestWorkflowEnvironment) AssertExpectations(t mock.TestingT) bool {
 // It can produce a false result when an argument is a pointer type and the underlying value changed after calling the mocked method.
 func (e *TestWorkflowEnvironment) AssertCalled(t mock.TestingT, methodName string, arguments ...interface{}) bool {
 	dummyT := &testing.T{}
-	if !(e.workflowMock.AssertCalled(dummyT, methodName, arguments...) || e.activityMock.AssertCalled(dummyT, methodName, arguments...)) {
-		return e.workflowMock.AssertCalled(t, methodName, arguments...) && e.activityMock.AssertCalled(t, methodName, arguments...)
-	}
-	return true
```
rodrigozhou, line 1005: I think this code is wrong, as well as `AssertNotCalled` and `AssertNumberOfCalls`.

Quinn-With-Two-Ns: What do you mean?

internal/workflow_testsuite.go
```diff
@@ -827,6 +972,27 @@ func (e *TestWorkflowEnvironment) SetOnLocalActivityCanceledListener(
...
+}
+
+func (e *TestWorkflowEnvironment) SetOnNexusOperationCompletedListener(
+	listener func(service string, operation string, result converter.EncodedValue, err error),
+) *TestWorkflowEnvironment {
+	e.impl.onNexusOperationCompletedListener = listener
+	return e
+}
+
+func (e *TestWorkflowEnvironment) SetOnNexusOperationCanceledListener(
```
Quinn-With-Two-Ns, line 989: nit: do we have a test for `SetOnNexusOperationCanceledListener`?

test/nexus_test.go
```diff
@@ -1184,6 +1186,229 @@ func TestWorkflowTestSuite_NexusSyncOperation_ClientMethods_Panic(t *testing.T)
 	require.Equal(t, "not implemented in the test environment", panicReason)
 }
 
+func TestWorkflowTestSuite_MockNexusOperation(t *testing.T) {
```
Quinn-With-Two-Ns, line 1189: nit: I don't think we test mocking cancel?

Quinn-With-Two-Ns (approved):
LGTM! few pieces of test coverage missing but not blocking address if you can. Please point at a released version of the nexus Go SDK before you merge

internal/workflow_testsuite.go
rodrigozhou: L1003 checks if both `workflowMock` and `activityMock` was called. I think it was supposed to be if either of them was called. The other functions have similar mistakes.

internal/workflow_testsuite.go
rodrigozhou: No, I'm not sure how to test it since the SDK itself doesn't support canceling a Nexus operation.

internal/workflow_testsuite.go
```diff
@@ -999,10 +1167,10 @@ func (e *TestWorkflowEnvironment) AssertExpectations(t mock.TestingT) bool {
 // It can produce a false result when an argument is a pointer type and the underlying value changed after calling the mocked method.
 func (e *TestWorkflowEnvironment) AssertCalled(t mock.TestingT, methodName string, arguments ...interface{}) bool {
 	dummyT := &testing.T{}
-	if !(e.workflowMock.AssertCalled(dummyT, methodName, arguments...) || e.activityMock.AssertCalled(dummyT, methodName, arguments...)) {
-		return e.workflowMock.AssertCalled(t, methodName, arguments...) && e.activityMock.AssertCalled(t, methodName, arguments...)
-	}
-	return true
+	return e.AssertWorkflowCalled(dummyT, methodName, arguments...) ||
+		e.AssertActivityCalled(dummyT, methodName, arguments...) ||
```
Quinn-With-Two-Ns, line 1171: Why are some of these called with `dummyT` and some `t`?

internal/workflow_testsuite.go
Quinn-With-Two-Ns: You can cancel a nexus operation from the SDK, the SDK can send a cancelation command from a workflow task.

internal/workflow_testsuite.go
rodrigozhou: That's a great question, and I have no idea. I just copied the existing calls.

Quinn-With-Two-Ns: I think we can just leave it as is, I will note I plan to deprecate these functions in the next minor release because they cause a lot of confusion for users since they awkwards combine workflow and activities so I wouldn't spend a lot of time fussing over them.

PR #1675, opened 2024-10-17
Add Deployment Names to Worker options 

antlai-temporal:
## What was changed

Deployment Names are a new way to group task queues, based on Build IDs, to simplify rollouts in versioning-3.

In this PR we just add a new Worker Option for it, propagate  it during polling, and include it with task completion responses.

### DIFF
--- a/internal/internal_versioning_client.go
+++ b/internal/internal_versioning_client.go
@@ -113,6 +113,8 @@ type (
 		BuildID string
 		// Whether the worker is using the versioning feature.
 		UseVersioning bool
+		// An identifier to group task queues based on Build ID.
+		DeploymentName string
 	}
 
 	// TaskQueuePollerInfo provides information about a worker/client polling a task queue.
@@ -243,8 +245,9 @@ func workerVersionCapabilitiesFromResponse(response *common.WorkerVersionCapabil
 	}
 
 	return &WorkerVersionCapabilities{
-		BuildID:       response.GetBuildId(),
-		UseVersioning: response.GetUseVersioning(),
+		BuildID:        response.GetBuildId(),
+		UseVersioning:  response.GetUseVersioning(),
+		DeploymentName: response.GetDeploymentName(),
 	}
 }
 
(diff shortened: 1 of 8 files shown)


cretz (approved):
Approving since it's just going to a branch

internal/worker.go
```diff
@@ -251,6 +251,12 @@ type (
 		// Note: Cannot be enabled at the same time as EnableSessionWorker
 		UseBuildIDForVersioning bool
 
+		// Optional: Assign a Deployment Name to this worker, an identifier for Worker Versioning that
+		// groups task queues for the given BuildID.
+		// NOTE: Experimental
+		// Note: Both BuildID and UseBuildIDForVersioning need to also be set to enable the new Worker Versioning-3 feature.
```
Quinn-With-Two-Ns, line 257: Nit using `Note` and `NOTE` in the same doc we should be consistent here

Can we validate this note and fail if `DeploymentName` is set but `BuildID` or `UseBuildIDForVersioning` is not set?

antlai-temporal: Thanks. I think sometimes we may want to provide a Deployment Name even if we have `UseBuildIDForVersioning=false`, the same as we can currently provide the BuildID without forcing versioning. For example, when turning versioning on dynamically with an env property, or using it with visibility queries, or with pollers info...

antlai-temporal:
Further comments will be addressed before merging versioning-3 into master

internal/internal_versioning_client.go
```diff
@@ -113,6 +113,8 @@ type (
 		BuildID string
 		// Whether the worker is using the versioning feature.
 		UseVersioning bool
+		// An identifier to group task queues based on Build ID.
```
drewhoskins-temporal, line 116: This is very low-level description.  The point of a deployment is to _abstract_ the task queues.
I would say something like "A Deployment corresponds to a set of Worker instances you want deployed together.  For example, it may represent a Worker service.  Each time you deploy, you'll be changing the build ID for that deployment."

PR #1679, opened 2024-10-18
Make a polling task timeout configurable

kingnand:
An internal task poller currently hardcodes the polling task timeout to 70 seconds. However, it should be set to dynamicconfig.MatchingLongPollExpirationInterval + some delta to account for the full round trip to matching.dynamicconfig.MatchingLongPollExpirationInterval (default is 60 seconds) is the maximum wait time for a new task, after which the server will return empty.

This change will make the polling task timeout configurable when the poller is created.

### DIFF
--- a/internal/internal_nexus_task_poller.go
+++ b/internal/internal_nexus_task_poller.go
@@ -63,6 +63,7 @@ func newNexusTaskPoller(
 			workerBuildID:        params.getBuildID(),
 			useBuildIDVersioning: params.UseBuildIDForVersioning,
 			capabilities:         params.capabilities,
+			pollTaskTimeout:      params.PollTaskTimeout,
 		},
 		taskHandler:     taskHandler,
 		service:         service,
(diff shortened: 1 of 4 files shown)


cretz:
Can you share the use case for altering the polling timeout? If we do allow customizing this option, we have to do this for all SDKs, not just Go.

kingnand:
Connectivity between Temporal workers and the Temporal cluster goes through a few additional layers. One of them that is a gateway has configured a timeout setting that is less than the default value of 70s. This results in a 504 Gateway timeout error when there are no tasks in queue. I believe deploying workers behind a gateway is a common pattern.

kingnand:
thanks @cretz for your clarification and raising this concern to the team. I completely agree with your points that we should prioritize the platform consistency across SDKs and also between self-host/cloud users. I'll discuss alternative solutions with the team internally. I'll close this PR as well.

PR #1682, opened 2024-10-22
Fix lost signal from Selector when Default path blocks

### LINKED ISSUE #1624: Signal sent to Selector can be lost if Default path blocks
## Expected Behavior

Selector can block in Default path without loosing signals

## Actual Behavior

Selector that block in the Default path risk loosing signals

When a signal is sent to the channel while the AddDefault method (long running) is executing, the signal is lost without being received. This was because if AddDefault is specified, the readyFunc is never executed. By storing the value in channel's recValue beforehand, we are making it available for any future receives on that channel without dropping the signal.

  See: https://github.com/altafDevRev/temporal-samples/pull/1

yuandrew:
## What was changed

Fixed a lost signal, gated behind a new SdkFlag. Currently the flag is added, but never set anywhere, so existing behavior is unchanged. A separate PR will be added to enable the flag.

## Why?

We shouldn't be losing a signal in our selector

## Checklist

#1624

2. How was this tested:

Added tests


internal/internal_coroutines_test.go
```diff
@@ -551,6 +551,191 @@ func TestBlockingSelect(t *testing.T) {
 	require.EqualValues(t, expected, history)
 }
 
+func TestSelectBlockingDefault(t *testing.T) {
```
yuandrew, line 554: We probably don't want the same test both here and in `internal_workflow_testsuite_test.go` right? Which file should hold this test?

yuandrew: Actually maybe this is fine, one set tests w/ and w/o flags, and the other does default behavior

test/replaytests/workflows.go
```diff
@@ -610,3 +610,49 @@ func ListAndDescribeWorkflow(ctx workflow.Context) (int, error) {
...
+
+	selector := workflow.NewSelector(ctx)
+	var s string
+	selector.AddReceive(ch1, func(c workflow.ReceiveChannel, more bool) {
+		c.Receive(ctx, &s)
+	})
+	selector.AddDefault(func() {
+		ch2.Receive(ctx, &s)
+	})
+	selector.Select(ctx)
```
Quinn-With-Two-Ns, line 641: How does this assert the signal was lost? I think you would need to call select at least twice to verify the signal was lost no?

internal/internal_workflow.go
```diff
@@ -1407,8 +1407,17 @@ func (s *selectorImpl) Select(ctx Context) {
 					if readyBranch != nil {
 						return false
 					}
-					readyBranch = func() {
+					// readyBranch is not executed when AddDefault is specified,
+					// setting the value here prevents the signal from being dropped
+					dropSignalFlag := getWorkflowEnvironment(ctx).TryUse(SDKFlagBlockedSelectorSignalReceive)
```
Quinn-With-Two-Ns, line 1412: Won't the `TryUse` here set the flag?

test/replaytests/workflows.go
yuandrew: Good callout, added an check for `HasPending` to control activity execution

internal/internal_workflow.go
yuandrew: It didn't seem like there was an existing method to query flags without setting them, created a new `GetFlag` method for this

internal/internal_workflow.go
```diff
@@ -1407,8 +1407,17 @@ func (s *selectorImpl) Select(ctx Context) {
 					if readyBranch != nil {
 						return false
 					}
-					readyBranch = func() {
+					// readyBranch is not executed when AddDefault is specified,
+					// setting the value here prevents the signal from being dropped
+					dropSignalFlag := getWorkflowEnvironment(ctx).GetFlag(SDKFlagBlockedSelectorSignalReceive)
```
Quinn-With-Two-Ns, line 1412: What test did you add that test the `true` path here?

Quinn-With-Two-Ns: ah I see since the test environment always assumes the flag is true, hm I am not sure if we should assume that since the test environment would diverge from the real workflow environment, what do you think?

yuandrew: I agree that the test environment shouldn't assume flags are true by default. That seems like a change that belongs in its own separate PR, I can create an issue for this

internal/internal_coroutines_test.go
```diff
@@ -551,6 +551,138 @@ func TestBlockingSelect(t *testing.T) {
...
+		"add-two",
+		"add-two-done",
+		"c2-two",
+		"select2",
+		"done",
+	}
+	require.EqualValues(t, expected, history)
+}
+
+func TestSelectBlockingDefaultWithFlag(t *testing.T) {
```
Quinn-With-Two-Ns, line 624: The reported bug was that blocking in the default case of a selector could cause signals to be lost, when I last looked at these tests we didn't seem to have any coverage for blocking in one selector case while a signal is received. Can we add tests to verify their is no bugs if a signal is received while blocking in another case of a selector, not just default?

internal/internal_workflow.go
yuandrew: #1700

Quinn-With-Two-Ns: To be clear I don't think SDK flags should be off by default in the test environment, I meant SDK flags used in the test environment should match what we enable when running a new workflow.

yuandrew: Oops, I misunderstood, I agree. 

Is there anywhere flags are set by default for a workflow? From looking at the code, it seems like `TryUse` are scattered around the code for different scenarios.

Quinn-With-Two-Ns: >Is there anywhere flags are set by default for a workflow? From looking at the code, it seems like TryUse are scattered around the code for different scenarios.

Yeah so how you can think of that is for new workflows all current flags are set to true. However for the new flag that you are adding here we don't want to set it to true by default because that would make it difficult for users to rollback their SDK version.

Quinn-With-Two-Ns:
Do you think it would be feasibly to add some debug API to set the flag to true and an integration test + replay test that tests a real workflow with the flag set?

test/workflow_test.go
```diff
@@ -3174,6 +3174,49 @@ func (w *Workflows) RunsLocalAndNonlocalActsWithRetries(ctx workflow.Context, nu
...
 
+func (w *Workflows) SelectorBlockSignal(ctx workflow.Context, enableFlag bool) (string, error) {
+	ctx = workflow.WithActivityOptions(ctx, w.defaultActivityOptions())
+	var logger = workflow.GetLogger(ctx)
+	logger.Info("calling ExecuteActivity")
+	ch1 := workflow.NewChannel(ctx)
+	ch2 := workflow.NewChannel(ctx)
+
+	if enableFlag {
+		internal.SetUnblockSelectorSignal()
```
Quinn-With-Two-Ns, line 3185: I think we should do this as part of the test, not the workflow. More importantly though can you confirm setting this won't effect any test that runs after this test?

yuandrew: > we should do this as part of the test, not the workflow

What is there difference between the two? or is this more of a style/preference?

> can you confirm setting this won't effect any test that runs after this test?

Looks like it does affect subsequent tests :( Is it enough to unset this value? Go tests run in parallel, so that seems insufficient?

Quinn-With-Two-Ns: >What is there difference between the two? or is this more of a style/preference?

I think it is clearer when we are testing a workflow if all the setup is run before hand. 

> Go tests run in parallel, so that seems insufficient?

Only if [t.Parallel()](https://pkg.go.dev/testing#T.Parallel) is called

test/integration_test.go
```diff
@@ -6602,6 +6602,32 @@ func (ts *IntegrationTestSuite) getReportedOperationCount(metricName string, ope
...
+	ts.NoError(run.Get(ctx, &result))
+	ts.Equal("hello", result)
+}
+
+func (ts *IntegrationTestSuite) TestSelectorNoBlock() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	options := ts.startWorkflowOptions("test-selector-block")
+
+	internal.SetUnblockSelectorSignal()
```
Quinn-With-Two-Ns, line 6621: nit: could just have one function that takes a bool: `SetUnblockSelectorSignal(bool)`

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
LGTM! Thanks for putting up with all my requests for more tests

PR #1683, opened 2024-10-23
Support mocking Nexus operation with operation reference

rodrigozhou:
## What was changed

Support mocking Nexus operation using `nexus.OperationReference`.

## Why?

The user won't need to create a dummy implementation of Nexus operation to mock it.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -768,6 +768,22 @@ func (r *registry) getWorkflowDefinition(wt WorkflowType) (WorkflowDefinition, e
 	return newSyncWorkflowDefinition(executor), nil
 }
 
+func (r *registry) getNexusService(service string) *nexus.Service {
+	r.Lock()
+	defer r.Unlock()
+	return r.nexusServices[service]
+}
+
+func (r *registry) getRegisteredNexusServices() []*nexus.Service {
+	r.Lock()
+	defer r.Unlock()
+	result := make([]*nexus.Service, 0, len(r.nexusServices))
+	for _, s := range r.nexusServices {
+		result = append(result, s)
+	}
+	return result
+}
+
 // Validate function parameters.
 func validateFnFormat(fnType reflect.Type, isWorkflow bool) error {
 	if fnType.Kind() != reflect.Func {
@@ -1058,7 +1074,7 @@ func (aw *AggregatedWorker) start() error {
 			return err
 		}
 	}
-	nexusServices := aw.registry.nexusServices
+	nexusServices := aw.registry.getRegisteredNexusServices()
 	if len(nexusServices) > 0 {
 		reg := nexus.NewServiceRegistry()
 		for _, service := range nexusServices {
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -2370,10 +2370,6 @@ func (env *testWorkflowEnvironmentImpl) executeChildWorkflowWithDelay(delayStart
 func (env *testWorkflowEnvironmentImpl) newTestNexusTaskHandler(
 	opHandle *testNexusOperationHandle,
 ) *nexusTaskHandler {
-	if len(env.registry.nexusServices) == 0 {
-		panic(fmt.Errorf("no nexus services registered"))
-	}
-
 	handler, err := newTestNexusHandler(env, opHandle)
 	if err != nil {
 		panic(fmt.Errorf("failed to create nexus handler: %w", err))
@@ -2552,20 +2548,16 @@ func (env *testWorkflowEnvironmentImpl) RegisterNexusAsyncOperationCompletion(
 ) error {
 	opRef := env.nexusOperationRefs[service][operation]
 	if opRef == nil {
-		panic(fmt.Sprintf(
-			"nexus service %q operation %q not mocked",
-			service,
-			operation,
-		))
+		return fmt.Errorf("nexus service %q operation %q not mocked", service, operation)
 	}
 	if reflect.TypeOf(result) != opRef.OutputType() {
-		panic(fmt.Sprintf(
+		return fmt.Errorf(
 			"nexus service %q operation %q expected result type %s, got %T",
 			service,
 			operation,
 			opRef.OutputType(),
 			result,
-		))
+		)
 	}
 
 	var data *commonpb.Payload
@@ -2575,6 +2567,11 @@ func (env *testWorkflowEnvironmentImpl) RegisterNexusAsyncOperationCompletion(
 		if encodeErr != nil {
 			return encodeErr
 		}
+	}
+
+	if err != nil {
+		// The handler workflow error needs to wrapped so it can be passed to the caller correctly.
+		err = NewApplicationError(err.Error(), "", true, err)
 	}
 
 	// Getting the locker to prevent race condition if this function is called while
@@ -3174,15 +3171,13 @@ func newTestNexusHandler(
 	env *testWorkflowEnvironmentImpl,
 	opHandle *testNexusOperationHandle,
 ) (nexus.Handler, error) {
+	nexusServices := env.registry.getRegisteredNexusServices()
+	if len(nexusServices) == 0 {
+		panic(fmt.Errorf("no nexus services registered"))
+	}
+
 	reg := nexus.NewServiceRegistry()
-	for _, service := range env.registry.nexusServices {
-		// register a dummy operation to make sure the service has at least 1 operation
-		_ = service.Register(nexus.NewSyncOperation(
-			"__internal__dummy",
-			func(_ context.Context, _ nexus.NoValue, _ nexus.StartOperationOptions) (nexus.NoValue, error) {
-				return nil, nil
-			},
-		))
+	for _, service := range nexusServices {
 		if err := reg.Register(service); err != nil {
 			return nil, fmt.Errorf("failed to register nexus service '%v': %w", service, err)
 		}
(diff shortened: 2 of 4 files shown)


Quinn-With-Two-Ns:
It would be nice to have some tests using this new approach so we can see how it would be used.

rodrigozhou:
@Quinn-With-Two-Ns Added a test using it.

internal/internal_workflow_testsuite.go
```diff
@@ -3149,6 +3176,13 @@ func newTestNexusHandler(
 ) (nexus.Handler, error) {
 	reg := nexus.NewServiceRegistry()
 	for _, service := range env.registry.nexusServices {
+		// register a dummy operation to make sure the service has at least 1 operation
```
Quinn-With-Two-Ns, line 3179: Can you explain why we need this code? I am curious why we are getting a service that has no operation and why we even bother to register it if it has no operation?

rodrigozhou: Hm, because it's auto registering a service when you mock a Nexus operation with a string. It was just to make the code a little bit simpler, but not required at all.
And in the nexus-rpc/sdk-go internals, if a service has no operations, it returns an error when creating the Nexus handler.

Quinn-With-Two-Ns: hm but when you register the service with a mock operation shouldn't you use the mock operation here?

rodrigozhou: It's mostly a conveniency. Not that we need this at all.
But it covers cases like the user registered a service, and then mock an operation without adding it to the service. Without this, the test will fail saying that the service has no operation.

Quinn-With-Two-Ns: Right, but shouldn't the service have the correct operations?

internal/internal_workflow_testsuite.go
```diff
@@ -2541,6 +2550,24 @@ func (env *testWorkflowEnvironmentImpl) RegisterNexusAsyncOperationCompletion(
 	err error,
 	delay time.Duration,
 ) error {
+	opRef := env.nexusOperationRefs[service][operation]
```
bergundy, line 2553: Better do this in two steps and provide better output in case the service isn't registered.

Also, why panic if this method returns an error? Is it standard in the rest of these `Register` methods? (didn't look into it, trust you to check).

internal/internal_workflow_testsuite.go
bergundy: +1 to @Quinn-With-Two-Ns's concern.

internal/internal_workflow_testsuite.go
```diff
@@ -3229,47 +3268,61 @@ func (r *testNexusHandler) StartOperation(
...
+		operation,
+		inputPtr.Elem().Interface(),
+		r.opHandle.params.options,
+	)
 	if mockRet != nil {
 		mockRetLen := len(mockRet)
 		if mockRetLen != 2 {
 			panic(fmt.Sprintf(
-				"mock of ExecuteNexusOperation has incorrect number of returns, expected 2, but actual is %d",
+				"mock of ExecuteNexusOperation has incorrect number of returns, expected 2, got %d",
```
bergundy, line 3285:
```suggestion
				"mock of ExecuteNexusOperation has incorrect number of return values, expected 2, got %d",
```

internal/internal_workflow_testsuite.go
```diff
@@ -3229,47 +3268,61 @@ func (r *testNexusHandler) StartOperation(
...
-				"mock of ExecuteNexusOperation has incorrect return type, expected error, but actual is %T (%v)",
-				mockErr,
+				"mock of ExecuteNexusOperation has incorrect return type, expected error, got %T",
 				mockErr,
 			))
 		}
 
 		mockResult := mockRet[0]
-		ret, ok := mockResult.(nexus.HandlerStartOperationResult[any])
+		result, ok := mockResult.(nexus.HandlerStartOperationResult[any])
```
bergundy, line 3303: I'm surprised this works in Go as `nexus.HandlerStartOperationResult[T]` isn't assignable to `nexus.HandlerStartOperationResult[any]`. But looks like your test below is passing.

internal/internal_workflow_testsuite.go
```diff
@@ -3229,47 +3268,61 @@ func (r *testNexusHandler) StartOperation(
...
 		if mockResult != nil && !ok {
 			panic(fmt.Sprintf(
-				"mock of ExecuteNexusOperation has incorrect return type, expected nexus.HandlerStartOperationResult[T], but actual is %T (%v)",
-				mockResult,
+				"mock of ExecuteNexusOperation has incorrect return type, expected nexus.HandlerStartOperationResult[T], but actual is %T",
```
bergundy, line 3306: nit: Is it a mock of `ExecuteNexusOperation`? I think of it as a mock of the operation handler.

bergundy (commented):
Very minor comments left. I'll leave it up to @Quinn-With-Two-Ns to approve.

internal/internal_workflow_testsuite.go
rodrigozhou: Only operations that are being mocked are being added to `nexusOperationRefs` map. So, if the user is not explicitly registering the service, and using `OnNexusOperation("service-name", ...)`, the service won't be registered (taking into consideration the other comment to not auto register services and operations).

internal/internal_workflow_testsuite.go
rodrigozhou: It's more like a mock of `ExecuteNexusOperation` because we are checking its input. Eg: we check the `NexusOperationOptions` which is not in the handler.

internal/internal_workflow_testsuite.go
bergundy: This response doesn't address my comment that we should either return an error and not panic (if consistent with other APIs on this struct) and if we do panic, we should provide a more helpful message.

rodrigozhou: Oh, right. Here I can return an error. I'll update.

internal/internal_workflow_testsuite.go
rodrigozhou: We can't create the Nexus operation from the `OperationReference` because of generics: the `reflect` doesn't have a good support for creating instances of generic type with `reflect.Type`.

Quinn-With-Two-Ns: hm but to register an operation don't you just need to satisfy the `RegisterableOperation` interface?

rodrigozhou: Yes, but it needs to embed `nexus.UnimplementedOperation[I, O]`.
I implemented `testNexusOperation` that implements the interface, and works as dummy operation that returns an error if it's actually called (that's because of `nexus.UnimplementedOperation`.
With this, I can create `testNexusOperation` from `testNexusOperationReference` so I can register the operation in the service.

internal/internal_worker.go
```diff
@@ -768,6 +768,22 @@ func (r *registry) getWorkflowDefinition(wt WorkflowType) (WorkflowDefinition, e
 	return newSyncWorkflowDefinition(executor), nil
 }
 
+func (r *registry) getNexusService(service string) *nexus.Service {
+	r.Lock()
+	defer r.Unlock()
+	return r.nexusServices[service]
+}
+
+func (r *registry) getRegisteredNexusServices() []*nexus.Service {
```
bergundy, line 777: nit: Don't name Go functions `GetXXX` it's not idiomatic Go: https://go.dev/doc/effective_go#Getters.

internal/internal_workflow_testsuite.go
```diff
@@ -2550,6 +2569,11 @@ func (env *testWorkflowEnvironmentImpl) RegisterNexusAsyncOperationCompletion(
 		}
 	}
 
+	if err != nil {
+		// The handler workflow error needs to wrapped so it can be passed to the caller correctly.
```
bergundy, line 2573:
```suggestion
		// The handler workflow error needs to be wrapped so it can be passed to the caller correctly.
```

internal/internal_workflow_testsuite.go
```diff
@@ -2550,6 +2569,11 @@ func (env *testWorkflowEnvironmentImpl) RegisterNexusAsyncOperationCompletion(
...
+		err = NewApplicationError(err.Error(), "", true, err)
```
bergundy, line 2574: Hmm.. should the cause here be the original error or a Temporal error? I would double check and compare to how we construct these in other places.

bergundy (approved):
Approved with comments. Nothing too major. Still want @Quinn-With-Two-Ns to do the last pass.

internal/internal_worker.go
rodrigozhou: I was just following the conventions in the file: `getRegisteredActivities`, etc.

internal/internal_workflow_testsuite.go
rodrigozhou: Hm. Actually, it's just nil. I'll update it.

test/nexus_test.go
```diff
@@ -1282,15 +1284,37 @@ func TestWorkflowTestSuite_MockNexusOperation(t *testing.T) {
...
+		require.NoError(t, env.GetWorkflowError())
+		var res string
+		require.NoError(t, env.GetWorkflowResult(&res))
+		require.Equal(t, "fake result", res)
+	})
 
+	t.Run("mock operation reference", func(t *testing.T) {
+		suite := testsuite.WorkflowTestSuite{}
+		env := suite.NewTestWorkflowEnvironment()
+		env.OnNexusOperation(
```
Quinn-With-Two-Ns, line 1307: Do we have a test that mixes `env.RegisterNexusService(service)` with `env.OnNexusOperation`  registering an operation reference?

Quinn-With-Two-Ns (approved):

test/nexus_test.go
rodrigozhou: Added another test mixing it.

PR #1684, opened 2024-10-23
Fix failure_reason label for nexus_task_execution_failed metric on task timeout

bergundy:
We mistakenly reported timeouts as `internal_sdk_error`. I'll update the docs site as well.


test/nexus_test.go
```diff
@@ -350,6 +353,20 @@ func TestNexusSyncOperation(t *testing.T) {
...
+		_, err := nexus.ExecuteOperation(ctx, nc, syncOp, "timeout", nexus.ExecuteOperationOptions{
+			// Force shorter timeout to speed up the test and get a response back.
+			Header: nexus.Header{nexus.HeaderRequestTimeout: "300ms"},
+		})
+		var unexpectedResponseErr *nexus.UnexpectedResponseError
+		require.ErrorAs(t, err, &unexpectedResponseErr)
+		require.Equal(t, nexus.StatusDownstreamTimeout, unexpectedResponseErr.Response.StatusCode)
+
+		require.EventuallyWithT(t, func(t *assert.CollectT) {
+			tc.requireFailureCounter(t, service.Name, syncOp.Name(), "timeout")
```
Quinn-With-Two-Ns, line 367: Why only assert the failure counter? shouldn't `NexusTaskScheduleToStartLatency` be published?

bergundy: I figured it's covered in the other tests.

Quinn-With-Two-Ns: hm I think this case is different because we don't report E2E latency on timeout

bergundy: I know, I meant the other two metrics. In any case, I've added the tests.

Quinn-With-Two-Ns (approved):

bergundy:
We mistakenly reported timeouts as `internal_sdk_error`.

PR #1690, opened 2024-10-27
Document Update-with-Start incompatible options

stephanos:
WISOTT

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -687,6 +687,7 @@ type (
 		// │ │ │ │ │
 		// │ │ │ │ │
 		// * * * * *
+		// Cannot be set the same time as a StartDelay or WithStartOperation.
 		CronSchedule string
 
 		// Memo - Optional non-indexed info that will be shown in list workflow.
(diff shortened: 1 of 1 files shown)


internal/client.go
```diff
@@ -721,7 +723,7 @@ type (
 
 		// StartDelay - Time to wait before dispatching the first workflow task.
 		// A signal from signal with start will not trigger a workflow task.
-		// Cannot be set the same time as a CronSchedule.
+		// Cannot be set the same time as a CronSchedule or WithStartOperation.
```
Quinn-With-Two-Ns, line 726: Do we have a runtime check for these?

stephanos: The server implements these checks: see https://github.com/temporalio/temporal/blob/main/service/frontend/workflow_handler.go#L603-L614

I could also add a client-side check? I think we had discussed client-side vs server-side in a previous PR and couldn't quite find a good rule of thumb for when to use which.

Quinn-With-Two-Ns (approved):

internal/client.go
Quinn-With-Two-Ns: Yeah having client side validation is nice to have, but I think it is well out of scope of this PR

Issue #1663, opened 2024-10-08
Add workflow annotations for versioning


antlai-temporal:
See the parent feature issue in https://github.com/temporalio/features/issues/548

The goal is to support the new versioning APIs that rely on workflow annotations.

PR #1692, opened 2024-10-28
Support annotations for versioning behavior

antlai-temporal:
## What was changed

This PR supports providing a default versioning behavior in Worker Options, and a programmatic API to override these defaults:

For example:
``` go
	workflow.SetVersioningBehavior(ctx, workflow.VersioningBehaviorPinned)
```
called  during the first workflow task sets the versioning mode for that workflow to `Pinned`, eliminating the need of patching.

And 
```go
	w := worker.New(c, ts.taskQueueName, worker.Options{
		DeploymentOptions: worker.DeploymentOptions{
                            BuildID:                   "1.0",
		            UseBuildIDForVersioning:   true,
		            DeploymentName:            "deploy-test1",
		            DefaultVersioningBehavior: workflow.VersioningBehaviorAutoUpgrade,
              }
	})
```
sets the default versioning behavior to `AutoUpgrade`. Note that `DeploymentName`, `UseBuildIDForVersioning` and `BuildID` need to be set to enable versioning-3.
If versioning-3 is on, but no default or programmatic versioning behavior has been set for a workflow, the server will fail its first workflow task. 

## Why?

A hint that a workflow is short-running, and therefore, does not need to be moved to the new Build ID on the next task, eliminates the need of patching for many workflow types. 

## Checklist

1. Closes 
#1663 
2. How was this tested:
Integration test that uses versioning-2 and a grpc interceptor to show that versioning behaviors are propagated with responses.
The dependency on versioning-2 is just to set a default Build ID.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -170,6 +170,9 @@ type (
 		// The worker's deployment name, an identifier in versioning-3 to group Task Queues for a given build ID.
 		DeploymentName string
 
+		// The Versioning Behavior for workflows that do not set one in their first task.
+		DefaultVersioningBehavior VersioningBehavior
+
 		MetricsHandler metrics.Handler
 
 		Logger log.Logger
@@ -1668,6 +1671,7 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 		WorkerBuildID:                         options.BuildID,
 		UseBuildIDForVersioning:               options.UseBuildIDForVersioning,
 		DeploymentName:                        options.DeploymentName,
+		DefaultVersioningBehavior:             options.DefaultVersioningBehavior,
 		MetricsHandler:                        client.metricsHandler.WithTags(metrics.TaskQueueTags(taskQueue)),
 		Logger:                                client.logger,
 		EnableLoggingInReplay:                 options.EnableLoggingInReplay,
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1624,7 +1624,7 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 
 	// Sessions are not currently compatible with worker versioning
 	// See: https://github.com/temporalio/sdk-go/issues/1227
-	if options.EnableSessionWorker && options.UseBuildIDForVersioning {
+	if options.EnableSessionWorker && options.DeploymentOptions.UseBuildIDForVersioning {
 		panic("cannot set both EnableSessionWorker and UseBuildIDForVersioning")
 	}
 
@@ -1668,10 +1668,10 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 		MaxConcurrentWorkflowTaskQueuePollers: options.MaxConcurrentWorkflowTaskPollers,
 		MaxConcurrentNexusTaskQueuePollers:    options.MaxConcurrentNexusTaskPollers,
 		Identity:                              client.identity,
-		WorkerBuildID:                         options.BuildID,
-		UseBuildIDForVersioning:               options.UseBuildIDForVersioning,
-		DeploymentName:                        options.DeploymentName,
-		DefaultVersioningBehavior:             options.DefaultVersioningBehavior,
+		WorkerBuildID:                         options.DeploymentOptions.BuildID,
+		UseBuildIDForVersioning:               options.DeploymentOptions.UseBuildIDForVersioning,
+		DeploymentName:                        options.DeploymentOptions.DeploymentName,
+		DefaultVersioningBehavior:             options.DeploymentOptions.DefaultVersioningBehavior,
 		MetricsHandler:                        client.metricsHandler.WithTags(metrics.TaskQueueTags(taskQueue)),
 		Logger:                                client.logger,
 		EnableLoggingInReplay:                 options.EnableLoggingInReplay,
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -1620,6 +1620,14 @@ func SetVersioningBehavior(ctx Context, behavior VersioningBehavior) {
 	env.WorkflowInfo().currentVersioningBehavior = behavior
 }
 
+// GetVersioningBehavior returns the last versioning behavior set with SetVersioningBehavior.
+//
+// NOTE: Experimental
+func GetVersioningBehavior(ctx Context) VersioningBehavior {
+	env := getWorkflowEnvironment(ctx)
+	return env.WorkflowInfo().currentVersioningBehavior
+}
+
 func getWorkflowMetadata(ctx Context) (*sdk.WorkflowMetadata, error) {
 	info := GetWorkflowInfo(ctx)
 	eo := getWorkflowEnvOptions(ctx)
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -1611,6 +1611,15 @@ func SetCurrentDetails(ctx Context, details string) {
 	getWorkflowEnvOptions(ctx).currentDetails = details
 }
 
+// SetVersioningBehavior sets the strategy to upgrade this workflow when the default Build ID
+// has changed, and Worker Versioning-3 has been enabled.
+//
+// NOTE: Experimental
+func SetVersioningBehavior(ctx Context, behavior VersioningBehavior) {
+	env := getWorkflowEnvironment(ctx)
+	env.WorkflowInfo().currentVersioningBehavior = behavior
+}
+
 func getWorkflowMetadata(ctx Context) (*sdk.WorkflowMetadata, error) {
 	info := GetWorkflowInfo(ctx)
 	eo := getWorkflowEnvOptions(ctx)
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -2555,7 +2555,7 @@ func (wc *workflowEnvironmentInterceptor) RequestCancelNexusOperation(ctx Contex
 	wc.env.RequestCancelNexusOperation(input.seq)
 }
 
-func VersioningBehaviorToProto(t VersioningBehavior) enumspb.VersioningBehavior {
+func versioningBehaviorToProto(t VersioningBehavior) enumspb.VersioningBehavior {
 	switch t {
 	case VersioningBehaviorUnspecified:
 		return enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED
@@ -2565,18 +2565,5 @@ func VersioningBehaviorToProto(t VersioningBehavior) enumspb.VersioningBehavior 
 		return enumspb.VERSIONING_BEHAVIOR_AUTO_UPGRADE
 	default:
 		panic("unknown versioning behavior type")
-	}
-}
-
-func VersioningBehaviorFromProto(t enumspb.VersioningBehavior) VersioningBehavior {
-	switch t {
-	case enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED:
-		return VersioningBehaviorUnspecified
-	case enumspb.VERSIONING_BEHAVIOR_PINNED:
-		return VersioningBehaviorPinned
-	case enumspb.VERSIONING_BEHAVIOR_AUTO_UPGRADE:
-		return VersioningBehaviorAutoUpgrade
-	default:
-		panic("unknown versioning behavior type from proto")
 	}
 }
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -59,6 +59,23 @@ const (
 	HandlerUnfinishedPolicyAbandon
 )
 
+// VersioningBehavior specifies when existing workflows could change their Build ID.
+// NOTE: Experimental
+type VersioningBehavior int
+
+const (
+	// Workflow versioning policy unknown. A default VersioningBehaviorUnspecified policy forces
+	// every workflow to explicitly set a VersioningBehavior different from VersioningBehaviorUnspecified.
+	VersioningBehaviorUnspecified VersioningBehavior = iota
+
+	// Workflow should be pinned to the current Build ID until manually moved.
+	VersioningBehaviorPinned
+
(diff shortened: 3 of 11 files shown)


workflow/workflow.go
```diff
@@ -607,6 +623,22 @@ func SetCurrentDetails(ctx Context, details string) {
...
+// SetVersioningBehavior should be called during the first workflow task, and the context `ctx` should
+// be the original context workflow argument, or derived from this context.
+//
+// If not set, a default behavior provided in worker.Options.DefaultVersioningBehavior will be used.
+//
+// If not set, and the default behavior is also unspecified in the Worker, the first task of this workflow
+// will fail when Worker Versioning-3 has been enabled.
+//
+// NOTE: Experimental
+func SetVersioningBehavior(ctx Context, behavior VersioningBehavior) {
```
cretz, line 638: I do not believe this is needed. Just use the worker option. I don't think we need global default for this worker option any more than any other (if a user wants to reference a global `EnableLoggingInReplay` or `VersioningBehavior` or whatever when they make their workers they can). This will because especially true as you get to other SDKs where we don't like customizable global defaults.

internal/workflow.go
```diff
@@ -59,6 +59,23 @@ const (
...
 )
 
+// VersioningBehavior specifies when existing workflows could change their Build ID.
+// NOTE: Experimental
+type VersioningBehavior int
+
+const (
+	// Workflow versioning policy unknown. A default VersioningBehaviorUnspecified policy forces
+	// every workflow to explicitly set a VersioningBehavior different from VersioningBehaviorUnspecified.
+	VersioningBehaviorUnspecified VersioningBehavior = iota
```
cretz, line 69: Hrmm, I wonder if we want to expose the raw proto enum like we do with `StartWorkflowOptions.WorkflowIDConflictPolicy` or not like I guess we chose not with `WorkflowUpdateStage`?

test/integration_test.go
```diff
@@ -6174,6 +6174,91 @@ func (ts *IntegrationTestSuite) TestScheduleUpdateWorkflowActionMemo() {
...
+					method string,
+					req interface{},
+					reply interface{},
+					cc *grpc.ClientConn,
+					invoker grpc.UnaryInvoker,
+					opts ...grpc.CallOption,
+				) error {
+					if method == "/temporal.api.workflowservice.v1.WorkflowService/RespondWorkflowTaskCompleted" {
+						asReq := req.(*workflowservice.RespondWorkflowTaskCompletedRequest)
+						behavior := internal.VersioningBehaviorFromProto(asReq.VersioningBehavior)
```
cretz, line 6219: This seems to be the only use of the "from proto" exposed utility, but there doesn't seem to be much benefit converting in this test. Can you just use the raw proto enum in this test for this slice and not have the utility?

internal/workflow.go
```diff
@@ -2531,3 +2554,29 @@ func (wc *workflowEnvironmentInterceptor) ExecuteNexusOperation(ctx Context, inp
 func (wc *workflowEnvironmentInterceptor) RequestCancelNexusOperation(ctx Context, input RequestCancelNexusOperationInput) {
 	wc.env.RequestCancelNexusOperation(input.seq)
 }
+
+func VersioningBehaviorToProto(t VersioningBehavior) enumspb.VersioningBehavior {
```
cretz, line 2558:
```suggestion
func versioningBehaviorToProto(t VersioningBehavior) enumspb.VersioningBehavior {
```

internal/internal_workflow.go
```diff
@@ -1611,6 +1611,15 @@ func SetCurrentDetails(ctx Context, details string) {
 	getWorkflowEnvOptions(ctx).currentDetails = details
 }
 
+// SetVersioningBehavior sets the strategy to upgrade this workflow when the default Build ID
```
Quinn-With-Two-Ns, line 1614: Did we intentionally omit a getter here?

workflow/workflow.go
Sushisource: Also agree

antlai-temporal: This is not global, it only changes the running workflow options, so that we can override  the default for a particular workflow. I agree about no process-level defaults...

internal/workflow.go
antlai-temporal: We could, what do we want to do moving forward @Quinn-With-Two-Ns , wrap enums or raw protos?

test/integration_test.go
antlai-temporal: I think the bigger question is your previous comment, whether we just expose the enum protos and be done with it. If not I think sooner or later users may want to go back and forth, so more boiler plate...

internal/internal_workflow.go
antlai-temporal: Yes, versioning behavior  can be dependent on worker options so if app code branches based on it they will get an NDE. And in general, we wanted to be able to change annotations without needing patching, so no-getter helps there...

internal/workflow.go
Quinn-With-Two-Ns: Yeah there is pros and cons to both and the SDK has historically not been consistent. I would bias towards wrapping the enums.

internal/internal_workflow.go
Quinn-With-Two-Ns: Is the versioning behaviour written into history?

internal/internal_worker.go
```diff
@@ -1668,6 +1671,7 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 		WorkerBuildID:                         options.BuildID,
 		UseBuildIDForVersioning:               options.UseBuildIDForVersioning,
 		DeploymentName:                        options.DeploymentName,
+		DefaultVersioningBehavior:             options.DefaultVersioningBehavior,
```
ShahabT, line 1674: As also discussed offline this morning, I think it's better to wrap these options in a struct called `DeploymentOption` or something. The list may grow in the future and we don't want the user to have to repeat all the fields individually in each TQ. Example:
```
deploymentOptions := worker.DeploymentOptions{
	DeploymentName: "my-app",
	BuildID: os.Getenv("BUILD_ID"),
	UseBuildIDForVersioning: true,
	DefaultVersioningBehavior: workflow.VersioningBehaviorAutoUpgrade,
}

w1 := worker.NewWorker(client, "foo", worker.WorkerOptions{
	DeploymentOptions: deploymentOptions,
})

w2 := worker.NewWorker(client, "bar", worker.WorkerOptions{
	DeploymentOptions: deploymentOptions,
})
```

workflow/workflow.go
cretz: I see, my mistake. I see this accepts a workflow context. I misread, sorry.

internal/workflow.go
cretz: Works for me

internal/internal_workflow.go
cretz: I think you need to be able to get something you may have set. I think this should either be in workflow register options and/or this should be called `SetVersioningBehaviorOverride` (and `GetVersioningBehaviorOverride` will just return what you have called with the "set", not the resolved value).

internal/internal_worker.go
antlai-temporal: I'm happy with this. @cretz and @Quinn-With-Two-Ns shall I add the struct and deprecate the original BuildID and UseBuildIdForVersioning?

internal/internal_workflow.go
antlai-temporal: @Quinn-With-Two-Ns My understanding is that it is not part of the history, at least I do nothing special during replay, @ShahabT can you confirm?

antlai-temporal: @cretz we could add the Override suffix, but the emphasis was that most customers will annotate all workflows, and do not set defaults in Workers, since that is safer.
 The Get override, could be useful in the future when we enable changing workflow properties externally (like a reset), maybe we could add it then.
I think the problem of adding to workflow register is that I don't have the context of the running workflow to latch on, so I cannot change the behavior based on arguments, or for dynamic workflows. Also, it may be more clear to have this "annotation" the first line of the workflow definition, together with the code... 
@drewhoskins-temporal any preferences?

internal/internal_worker.go
Quinn-With-Two-Ns: All these fields are marked experimental right? If so I have no strong opinion here. If we do this in Go we should probably group them in all SDKs though.

test/integration_test.go
antlai-temporal: Ok, I don't think implementing the FromProto if they need it is a big deal. I'm getting rid of it, using proto in the test, and making the toProto private.

cretz: Works for me. Yeah the high-level wrapper is nicer for users, and no need to convert the other way.

internal/internal_workflow.go
antlai-temporal: Looking at the proto changes that @ShahabT did, the versioning behavior has been added to WorkflowTaskCompletedEventAttributes in history. I don't do anything with it (maybe I should?) but that would give the server a placeholder to record the versioning behavior that I send after each task...

cretz: I do wonder whether the getter should take that into account. I'll defer to y'all on desired behavior of the getter, but I do believe we should not have a setter without a getter (even if we have to add caveats to the behavior of the getter).

carlydf: The server stores the current build id and versioning behavior in Mutable State and in visibility. The previously-used build ids are also stored in visibility -- I'm not sure if previously-used build ids are also stores in Mutable State.

"getting" the current versioning behavior of a given workflow execution could be done by reading the mutable state or visibility entry.

Querying a bunch of workflow executions for a specific versioning behavior and/or build id combination would be a visibility query (we will use this for reachability).

It seems to me that to support a workflow-execution-level versioning behavior override, the server would need to store not just the (versioning behavior, build id) tuple, but also an `is_override` (or some other name) bit, that basically means, `if is_override, ignore whatever the sdk says about my workflow type and do task routing based on what is currently written in my mutable state as an override`. To me it seems like it could be handled on the server side, since that's where the task routing decision is made.

antlai-temporal:
I think the latest commit addresses most of the issues raised, let me know if it is ready to merge...

internal/internal_workflow.go
cretz: > I think of versioning behavior as primarily a workflow type level thing, with a workflow worker level default in case it's not specified at the type level (but from a product perspective, we want to encourage explicitly setting it for each workflow type

If this is the case, it also needs to be added to `workflow.RegisterOptions`.

> "getting" the current versioning behavior of a given workflow execution could be done by reading the mutable state or visibility entry.

We don't want to get from server what's not already in history. We just need to make sure any `SetWhatever` _inside_ a workflow has a corresponding `GetWhatever`. If this just returns the value that was "set" before in the same workflow code and not the actual server value, that's fine (though it does seem like task complete may have some form of the value). The docs of the getter just need to clarify.

cretz:
@antlai-temporal - based on thread above, I wonder if this needs to be on `worker.RegisterOptions` too if that is the most common way this should be set (at the workflow type level).

antlai-temporal:
> @antlai-temporal - based on thread above, I wonder if this needs to be on `worker.RegisterOptions` too if that is the most common way this should be set (at the workflow type level).

We could have a type-level default in RegisterOptions that can be overridden by the programmatic API. That would be more consistent with the Java annotation+programmatic override I was thinking on. The Getter will probably return just the programmatic to be consistent that we can change worker defaults without patching.
I'm trying to unblock testing for the server team, so I may do that in a separate PR if that's ok...

workflow/workflow.go
```diff
@@ -607,6 +623,34 @@ func SetCurrentDetails(ctx Context, details string) {
 	internal.SetCurrentDetails(ctx, details)
 }
 
+// SetVersioningBehavior sets the strategy to upgrade this workflow when the default Build ID
+// has changed, and Worker Versioning-3 has been enabled.
+//
+// SetVersioningBehavior should be called during the first workflow task, and the context `ctx` should
```
Quinn-With-Two-Ns, line 629: What if it is called, but not in the first workflow task?

antlai-temporal: If there was no worker level default, the server will fail the first workflow task, so there is no progress until a worker sets a default.
If tthere was a worker level default, it will change it ok.
In the comment I'm encouraging to set it in the first task to avoid the first case...

cretz:
> I'm trying to unblock testing for the server team, so I may do that in a separate PR if that's ok...

I think it's probably good (will defer to @Quinn-With-Two-Ns) but probably want to create an issue to track the work. I do think we should be encouraging that way to set worker-type level value and this should be a lesser-used in-workflow override.

>  In the comment I'm encouraging to set it in the first task to avoid the first case...

This is why we should be encouraging the workflow register options approach over the inline approach (can even change the name of the inline one to something less wieldy, e.g. `SetVersioningBehaviorOverride`).

Quinn-With-Two-Ns:
Yeah OK merging into the branch, +1 to everything Chad said.

Once we are further along I would also like to see documentation as to how this workflow API is supposed to interact with versioning info set externally described here https://github.com/temporalio/sdk-go/pull/1692#discussion_r1821431989.

Quinn-With-Two-Ns (approved):

antlai-temporal:
> Yeah OK merging into the branch, +1 to everything Chad said.
> 
> Once we are further along I would also like to see documentation as to how this workflow API is supposed to interact with versioning info set externally described here [#1692 (comment)](https://github.com/temporalio/sdk-go/pull/1692#discussion_r1821431989).

@cretz @Quinn-With-Two-Ns 
I added the issue https://github.com/temporalio/sdk-go/issues/1698 . Merging as it is to unblock server team, but will add another PR before merging to main.
We will clarify the interactions with the external workflow property setting (after discussing today with @carlydf they seem orthogonal). The keyword "override" is used by this external mechanism, I will stick to default vs non-default to avoid confusion. The getter will only return what it was set (as it does right now).

PR #1695, opened 2024-10-29
Dedup duplicate update IDs for test environment

### LINKED ISSUE #1638: Workflow Update in Test Environment should dedup updates by ID
## Expected Behavior
Workflow Updates in Test Environment behave like Updates against a real server and either dupes for fails if an update with a duplicate ID is sent

## Actual Behavior
Test Environment accepts update with duplicate ID

yuandrew:
## What was changed

Added functionality in test environment to cache update 

## Why?

Match non-test behavior

## Checklist

1. Closes #1638

2. How was this tested:
new test written

3. Any docs updates needed?


internal/workflow_testsuite_test.go
```diff
@@ -491,6 +491,63 @@ func TestWorkflowUpdateOrderAcceptReject(t *testing.T) {
...
+			},
+			complete: func(result interface{}, err error) {
+				intResult, ok := result.(int)
+				if !ok {
+					require.Fail(t, "result should be int")
+				} else {
+					require.Equal(t, 0, intResult)
+				}
+			},
+			env: env,
```
yuandrew, line 513: The current solution requires `env` to be passed to updateCallback, not ideal, but struggling to come up with an alternate solution

internal/internal_workflow_test.go
```diff
@@ -1417,6 +1418,10 @@ func (uc *updateCallback) Reject(err error) {
 }
 
 func (uc *updateCallback) Complete(success interface{}, err error) {
+	// cache update result so we can dedup duplicate update IDs
+	if uc.env.impl.updateMap != nil {
+		uc.env.impl.updateMap[uc.env.impl.currentUpdateId] = success
```
Quinn-With-Two-Ns, line 1423: The error would also need to be cached if we want to go with the caching approach

Quinn-With-Two-Ns: the error here represents the users error they returned from the update handler

internal/internal_workflow_testsuite.go
```diff
@@ -229,6 +235,8 @@ type (
 
 		workflowFunctionExecuting bool
 		bufferedUpdateRequests    map[string][]func()
+
+		currentUpdateId string
```
Quinn-With-Two-Ns, line 239: What if there are multiple concurrent updates sent to the workflow? It looks like those would be lost.

yuandrew: hmm, good point

internal/workflow_testsuite_test.go
```diff
@@ -273,14 +273,17 @@ func TestWorkflowIDUpdateWorkflowByID(t *testing.T) {
 	var suite WorkflowTestSuite
 	// Test UpdateWorkflowByID works with custom ID
 	env := suite.NewTestWorkflowEnvironment()
+	updateID := "id"
 	env.RegisterDelayedCallback(func() {
-		err := env.UpdateWorkflowByID("my-workflow-id", "update", "id", &updateCallback{
-			reject: func(err error) {
+		err := env.UpdateWorkflowByID("my-workflow-id", "update", updateID, newUpdateCallback(
+			env,
+			updateID,
```
yuandrew, line 280: @Quinn-With-Two-Ns how do you feel about solution? It doesn't feel ideal that we have to pass `updateID` in 2 places, but I can't think of any other way to handle async updates. `updateCallback` needs to know what the updateID to cache its result to `env` is, and this seems like the only way to accomplish that.

internal/internal_workflow_test.go
```diff
@@ -1406,6 +1406,19 @@ type updateCallback struct {
 	accept   func()
 	reject   func(error)
 	complete func(interface{}, error)
+	env      *TestWorkflowEnvironment
```
Quinn-With-Two-Ns, line 1409: Hmm so this class is not for users, this whole file is for testing since it is post-fixed with `*_test.go`. The logic can't be in the updateCallback as we take an impl. from user. The logic needs to be in `(env *testWorkflowEnvironmentImpl) updateWorkflow`. We could potentially add a wrapper around the user interface. Does that make sense?

yuandrew: Is `user` in this context referring to the user of the test suite?

yuandrew: > this whole file is for testing since it is post-fixed with *_test.go. The logic can't be in the updateCallback as we take an impl. from user
And would it be accurate to rephrase this as "we need to implement this within the test suite, not the individual test logic. `updateCallback` is a test specific implementation"?

yuandrew: > We could potentially add a wrapper around the user interface. 
What interface are you referring to here?

Quinn-With-Two-Ns: >Is user in this context referring to the user of the test suite?

Yeah 

>And would it be accurate to rephrase this as "we need to implement this within the test suite, not the individual test logic. updateCallback is a test specific implementation"?

yeah, exactly

> What interface are you referring to here?

`UpdateCallbacks`

internal/workflow_testsuite_test.go
```diff
@@ -491,6 +491,61 @@ func TestWorkflowUpdateOrderAcceptReject(t *testing.T) {
...
+					require.Fail(t, "result should be int")
+				} else {
+					require.Equal(t, 0, intResult)
+				}
+			},
+		}, 0)
+	}, 0)
+
+	env.RegisterDelayedCallback(func() {
+		env.UpdateWorkflow("update", "id", &updateCallback{
```
Quinn-With-Two-Ns, line 517: What if this update is sent before the first update completes?

yuandrew: good point, added a mutex to the map to ensure the 2nd update must wait for the 1st update to complete

internal/internal_workflow_testsuite.go
```diff
@@ -2910,10 +2924,32 @@ func (env *testWorkflowEnvironmentImpl) updateWorkflow(name string, id string, u
...
+
+	if env.updateMap == nil {
+		env.updateMap = make(map[string]updateResult)
+	}
+
+	var ucWrapper = updateCallbacksWrapper{uc: uc, env: env, updateID: id}
+
+	// check for duplicate update ID
+	if result, ok := env.updateMap[id]; ok {
+		result.mu.Lock()
```
Quinn-With-Two-Ns, line 2936: This doesn't really address the problem because now your `updateWorkflow` call will block which will block other callbacks. To properly dedup you would need to keep track of the inflight updates requests and attach any other requests to them.

Given that is very complex I think it would be simpler to just fail if we detect a duplicate update in the test environment because there is no reason a user should ever send one.

yuandrew: from a high level, could inflight updates be as easy as `updateResults` being
```
 updateResult struct {
		success interface{}
		err     error
		callbacks []func()
	}
```
and in the callback of the first request, we `defer` with a `.complete_callbacks` to `postCallback` all pending duplicate updates?

I'm gonna try spending ~20 minutes this afternoon seeing if this can work, but lmk if anything sounds wrong here at a high level. And if I can't get it to work, I'll go the "fail duplicate updates` route

yuandrew: I think I got a working solution, but the only downside is maybe the `TestWorkflowDuplicateIDDedupInterweave` test isn't obvious (maybe not guaranteed?) that this exact order of code is happening. 

I've confirmed through debug print logs that 
1. both updates get scheduled
2. caches the result
3. first update completes
4. 2nd update then finishes with cached result 
 
Screenshot 2024-11-21 at 3 14 59 PM

not sure if there's a more explicit way of showing/guaranteeing that this exact order is happening.

Quinn-With-Two-Ns (approved):

PR #1699, opened 2024-10-31
bump api to v1.41.0

lina-temporal:
## What was changed
- bumps API to v1.41.0

## Checklist

2. How was this tested:
      `cd internal/cmd/build && go run . check`


Quinn-With-Two-Ns:
Appears like the API repo is not properly tagged https://github.com/temporalio/api/tags , the latest tag there is `1.40.0`

PR #1707, opened 2024-11-12
Don't heartbeat if poller is stopped

### LINKED ISSUE #1706: ProcessWorkflowTask is not stopped on worker.Stop()
## Expected Behavior

## Actual Behavior

```
```

```
```

```go
package main

import (
	"context"
	"fmt"
	"time"

	"go.temporal.io/api/enums/v1"
	"go.temporal.io/sdk/client"
	"go.temporal.io/sdk/worker"
	"go.temporal.io/sdk/workflow"
)

const (
	hostPort   = "localhost:7233"
	namespace  = "default"
	taskQueue  = "default"
	workflowID = "test-workflow"

	nActivities   = 10
)

type Worker struct {
	parentClient client.Client
	client       client.Client
	worker       worker.Worker
}

}

	if err != nil {
	}
	tw := newTemporalWorker(c)
	err = tw.Start()
	if err != nil {
	}
	w.client = c
	w.worker = tw
	return nil
}

func (w *Worker) Stop() {
	w.worker.Stop()
	w.client.Close()
}

		HostPort:  hostPort,
		Namespace: namespace,
	})
}

}

	w.RegisterWorkflow(Workflow)
	w.RegisterActivity(Activity)
	return w
}

	for {
		for range nActivities {
			})
			if err != nil {
			}
		}
		}
	}
}

	time.Sleep(sleepDuration)
	return ctx.Err()
}

	}
	return err
}

func main() {
	ctx := context.Background()
	c, err := newTemporalClient()
	if err != nil {
	}

	err = startWorkflow(ctx, c)
	if err != nil {
		return
	}

	w := NewWorker(c)
	for {
		err = w.Start()
		if err != nil {
			return
		}
		w.Stop()
	}
}
```

## Specifications

  - Version:
  - Platform:

ndtretyak:
## What was changed

`ProcessWorkflowTask` will now stop sending heartbeats if the worker is stopped.

## Why?

Without this change, workflows with local activities could hang indefinitely, waiting for a local activity to execute even when the local activity worker has already stopped.

## Checklist

1. Closes #1706

2. How was this tested:

New unit test

3. Any docs updates needed?

No

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -415,6 +415,9 @@ func (wtp *workflowTaskPoller) processWorkflowTask(task *workflowTask) (retErr e
 			wfctx,
 			func(response interface{}, startTime time.Time) (*workflowTask, error) {
 				wtp.logger.Debug("Force RespondWorkflowTaskCompleted.", "TaskStartedEventID", task.task.GetStartedEventId())
+				if wtp.stopping() {
+					return nil, errStop
+				}
 				heartbeatResponse, err := wtp.RespondTaskCompletedWithMetrics(response, nil, task.task, startTime)
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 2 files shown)


cretz:
Added comment on #1706 mentioning that we don't want to automatically make local activities start failing on poller stop, but maybe there is some other issue here with worker stop and local activities.

Issue #1698, opened 2024-10-30
Enable type-scoped Versioning Behavior for workflows  


antlai-temporal:
Versioning-3 introduces behavior annotations for workflows  that could change what Build ID will be used for a workflow next task.

The current API is programmatic, and requires invoking *SetVersioningBehavior()* during the first workflow task. This gives extra 
flexibility for dynamic workflows but can be error-prone.

This issue suggests an alternative approach which associates a default behavior to a workflow type, using a new option for workflow registration.

**Describe the solution you'd like**

Add the option `DefaultVersioningBehavior` to `workflow.RegisterOptions`, providing a default behavior for all workflows of a particular type that do not use the programmatic API.

PR #1708, opened 2024-11-13
Annotate versioning behavior per Workflow type

antlai-temporal:
## What was changed

Provides an alternative mechanism to provide a default versioning behavior per workflow type using workflow registration options.
```go
	w.RegisterWorkflowWithOptions(ts.workflows.Basic, workflow.RegisterOptions{
		VersioningBehavior: workflow.VersioningBehaviorPinned,
	})
```

## Why?

The programmatic  API to set Versioning Behavior is brittle because it needs to be called during the first workflow task.

## Checklist

1. Closes 
#1698 
2. How was this tested:

An integration test using gRPC interceptors

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -530,12 +530,13 @@ func (aw *activityWorker) Stop() {
 
 type registry struct {
 	sync.Mutex
-	nexusServices    map[string]*nexus.Service
-	workflowFuncMap  map[string]interface{}
-	workflowAliasMap map[string]string
-	activityFuncMap  map[string]activity
-	activityAliasMap map[string]string
-	interceptors     []WorkerInterceptor
+	nexusServices                 map[string]*nexus.Service
+	workflowFuncMap               map[string]interface{}
+	workflowAliasMap              map[string]string
+	workflowVersioningBehaviorMap map[string]VersioningBehavior
+	activityFuncMap               map[string]activity
+	activityAliasMap              map[string]string
+	interceptors                  []WorkerInterceptor
 }
 
 type registryOptions struct {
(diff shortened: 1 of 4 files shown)


internal/internal_worker.go
```diff
@@ -580,6 +584,9 @@ func (r *registry) RegisterWorkflowWithOptions(
 		}
 	}
 	r.workflowFuncMap[registerName] = wf
+	if options.VersioningBehavior != VersioningBehaviorUnspecified {
+		r.workflowVersioningBehaviorMap[registerName] = options.VersioningBehavior
+	}
```
cretz, line 589:
```suggestion
	r.workflowVersioningBehaviorMap[registerName] = options.VersioningBehavior
```

Could just put it in the map always and remove the `, ok` approach from the getter and just rely on unspecified to be equivalent to unset. Not a big deal, but I think it makes for cleaner code.

cretz (approved):

test/integration_test.go
```diff
@@ -6271,6 +6271,100 @@ func (ts *IntegrationTestSuite) TestVersioningBehaviorInRespondWorkflowTaskCompl
...
+					}
+					return invoker(ctx, method, req, reply, cc, opts...)
+				}),
+			},
+		},
+	})
+	ts.NoError(err)
+	defer c.Close()
+
+	ts.worker.Stop()
```
yuandrew, line 6326: 2 questions here

1. What code previously started a worker? I'm guessing there's other ways to start a worker besides `ExecuteWorkflow`?
2. What does stopping the worker here do for us? Why is it necessary for the worker to be stopped here?

antlai-temporal: The `IntegrationTestSuite: SetupTest()` always starts a worker for each test, see line 286 in integration_test.go

We need a worker with specific options, so we need to stop that one to guarantee that the new one will be the only one executing tasks.

yuandrew: got it! Thanks for the explanation :)

internal/internal_worker.go
antlai-temporal: In general I don't like adding unnecessary entries to maps, but here there is no performance/space issue and it is cleaner...
Done

PR #1709, opened 2024-11-13
added method to set workflowID when testing workflows

Shivs11:
## What was changed

- Added a method to be able to set the `workflowID` in the workflow test environment

## Why?

- This is a direct dependency for some of the work I have been doing with respect to `versioning-3`
- During unit testing, I will require the ability to alter workflowID's when starting `deployment` (versioning-3 concept) workflows and having this method in proves to be handy.

## Checklist

1. How was this tested:

- Current unit tests. No new tests were added since this is a getter.

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -491,6 +491,10 @@ func (env *testWorkflowEnvironmentImpl) setIdentity(identity string) {
 	env.identity = identity
 }
 
+func (env *testWorkflowEnvironmentImpl) setWorkflowID(workflowID string) {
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns:
You should be able to use `testWorkflowEnvironmentImpl.SetStartWorkflowOptions` to set the Workflow ID

Shivs11:
> You should be able to use `testWorkflowEnvironmentImpl.SetStartWorkflowOptions` to set the Workflow ID

Ah, I completely seemed to have missed this. Thought this use-case was never required given there are other setter methods (`setXXX`) but not this one. Closing this one.

PR #1710, opened 2024-11-15
Set Nexus operation ID on callback headers

pdoerner:
## What was changed

Setting `nexus-operation-id` header on callback headers.
Depends on https://github.com/nexus-rpc/sdk-go/pull/26
Returns an error is the workflow ID is unset on the start options for a Nexus operation

## Why?

This information is necessary for fabricating a NexusOperationStarted event if the completion is received before the response to a StartOperation request. This ID is set to the workflow ID unless the user provides a value for this header; user-provided values for this header are not overwritten.

## Checklist

1. Closes 

2. How was this tested:

Existing tests

3. Any docs updates needed?


temporalnexus/operation.go
```diff
@@ -324,12 +325,21 @@ func ExecuteUntypedWorkflow[R any](
 	if startWorkflowOptions.TaskQueue == "" {
 		startWorkflowOptions.TaskQueue = nctx.TaskQueue
 	}
+	if startWorkflowOptions.ID == "" {
```
bergundy, line 328: I'm wondering if we want to require setting workflow ID for Nexus operations. We require it in all Core based SDKs, maybe we can at least protect Nexus users. Not blocking this PR though. @cretz, @Quinn-With-Two-Ns to weigh in on this.

bergundy (commented):
LGTM.

Another non-blocking request is to always have the server generate the links (SDK can generate as a fallback when used with an old server).
That way operation ID is always coming from the SDK and the link from the server.

We should also document the implications of changing the operation ID returned from `WorkflowRunOperation.Start`. Let's tell them to never do that because the ID is used for cancelation and is reporting completion from the server.

pdoerner:
@bergundy re this comment

> Another non-blocking request is to always have the server generate the links (SDK can generate as a fallback when used with an old server). That way operation ID is always coming from the SDK and the link from the server.

I didn't add any generation for links here; are you suggesting I should add the fallback?

temporalnexus/operation.go
bergundy: I would change to ensure that the client doesn't auto generate an ID if one isn't set.

We should note this change in the release notes. I would consider it a backwards incompatible behavior change.

```go
	if startWorkflowOptions.ID == "" {
		return nil, ErrMissingWorkflowID
	}

```

bergundy (commented):
Please change the behavior when there's a missing workflow ID.

bergundy:
> @bergundy re this comment
> 
> > Another non-blocking request is to always have the server generate the links (SDK can generate as a fallback when used with an old server). That way operation ID is always coming from the SDK and the link from the server.
> 
> I didn't add any generation for links here; are you suggesting I should add the fallback?


I'm saying we modify the server to return links and only generate them in the SDK in if the server doesn't return them in the StartWorkflowExecutionResponse. We'll need to consider when to return these links though. I opened an issue to track this, it's not blocking.

temporalnexus/operation.go
bergundy: @pdoerner this already changed for Java, please make sure to disallow empty workflow ID for nexus ops before merging this.

internal/error.go
```diff
@@ -338,6 +338,9 @@ var (
 
 	// ErrSkipScheduleUpdate is used by a user if they want to skip updating a schedule.
 	ErrSkipScheduleUpdate = errors.New("skip schedule update")
+
+	// ErrMissingWorkflowID is returned when trying to start an async Nexus operation but no workflow ID is set on the request.
+	ErrMissingWorkflowID = errors.New("workflow ID is unset")
```
Quinn-With-Two-Ns, line 343: nit can the error message make it clear that it was unset for a Nexus operation

temporalnexus/operation.go
```diff
@@ -324,12 +327,21 @@ func ExecuteUntypedWorkflow[R any](
...
 
 	if nexusOptions.RequestID != "" {
 		internal.SetRequestIDOnStartWorkflowOptions(&startWorkflowOptions, nexusOptions.RequestID)
 	}
 
 	if nexusOptions.CallbackURL != "" {
+		if nexusOptions.CallbackHeader == nil {
+			nexusOptions.CallbackHeader = make(nexus.Header)
+		}
+		if _, set := nexusOptions.CallbackHeader[nexus.HeaderOperationID]; !set {
```
Quinn-With-Two-Ns, line 342: nit: should we use the case insensitive operations on the headers? I guess `nexus.HeaderOperationID` is now lower case so it isn't that important.

PR #1711, opened 2024-11-19
Added SpanFromWorkflowContext function

### LINKED ISSUE #1701: go.temporal.io/sdk/contrib/datadog to export TracerSpan and allow custom SpanContextKey
In order to be able to access spans from `workflow.Context` to be able to do things like add custom tags, it would be helpful if the `go.temporal.io/sdk/contrib/datadog/tracing` package exported `TracerSpan` and allowed setting a custom `SpanContextKey` in the tracer options.

This would enable code like the following to be written:
```go
// Assuming the following import:
// import 	"go.temporal.io/sdk/contrib/datadog/tracing"

val := ctx.Value(DD_ACTIVE_SPAN_CONTEXT_KEY)
if val == nil {
	return
}

span := *val.(*tracing.TracerSpan)
span.SetTag("myCustomTag", myCustomValue)
```

go.temporal.io/sdk/contrib/datadog to export the TracerSpan struct (currently named `tracerSpan` so it isn't exported) and allow custom SpanContextKey in the `TracerOptions` struct.

I have prepared some changes that solve this for my particular use-case which you can see here. Happy to open a PR if you agree that this feature makes sense for the library:
https://github.com/temporalio/sdk-go/compare/master...cdimitroulas:temporal-sdk-go:christos/export-datadog-tracer-span-struct?expand=1

I was not able to find an alternative solution that supports this use-case.

**Additional context**
N/A

cdimitroulas:
## What was changed
Added a `SpanFromWorkflowContext` function to allow access to the DataDog span object from within a workflow.

## Why?
This allows users to access the Span from within a workflow to (for example) add custom tags on the span.

## Checklist

1. Closes #1701 
2. How was this tested: New unit test added

3. Any docs updates needed? Would be nice, please let me know where I could add these if this is desired.

### DIFF
--- a/contrib/datadog/tracing/interceptor.go
+++ b/contrib/datadog/tracing/interceptor.go
@@ -143,6 +143,20 @@ func (t *tracerImpl) ContextWithSpan(ctx context.Context, span interceptor.Trace
 	return tracer.ContextWithSpan(ctx, span.(*tracerSpan).Span)
 }
 
+// SpanFromWorkflowContext extracts the DataDog Span object from the workflow context.
+func SpanFromWorkflowContext(ctx workflow.Context) (ddtrace.Span, bool) {
+	val := ctx.Value(activeSpanContextKey)
+	if val == nil {
+		return tracer.SpanFromContext(nil)
+	}
+
+	if span, ok := val.(*tracerSpan); ok {
+		return span.Span, true
+	}
+
+	return tracer.SpanFromContext(nil)
+}
+
 func genSpanID(idempotencyKey string) uint64 {
 	h := fnv.New64()
 	// Write() always writes all bytes and never fails; the count and error result are for implementing io.Writer.
--- a/contrib/datadog/tracing/interceptor_test.go
+++ b/contrib/datadog/tracing/interceptor_test.go
@@ -22,6 +22,7 @@
 package tracing
 
 import (
+	"errors"
 	"strings"
 	"testing"
 
@@ -31,6 +32,9 @@ import (
 
 	"go.temporal.io/sdk/interceptor"
 	"go.temporal.io/sdk/internal/interceptortest"
+	"go.temporal.io/sdk/testsuite"
+	"go.temporal.io/sdk/worker"
+	"go.temporal.io/sdk/workflow"
 )
 
 type testTracer struct {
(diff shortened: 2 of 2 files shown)


contrib/datadog/tracing/interceptor.go
```diff
@@ -143,6 +143,18 @@ func (t *tracerImpl) ContextWithSpan(ctx context.Context, span interceptor.Trace
 	return tracer.ContextWithSpan(ctx, span.(*tracerSpan).Span)
 }
 
+// SpanFromWorkflowContext extracts the DataDog Span object from the workflow context.
+func SpanFromWorkflowContext(ctx workflow.Context) (ddtrace.Span, bool) {
```
cretz, line 147: Not sure the bool is needed. We can just use `nil` to represent not present.

cdimitroulas: Agreed - I've updated it now.

cretz (approved):
LGTM, will let @Quinn-With-Two-Ns review

contrib/datadog/tracing/interceptor.go
```diff
@@ -143,6 +143,18 @@ func (t *tracerImpl) ContextWithSpan(ctx context.Context, span interceptor.Trace
 	return tracer.ContextWithSpan(ctx, span.(*tracerSpan).Span)
 }
 
+// SpanFromWorkflowContext extracts the DataDog Span object from the workflow context.
+func SpanFromWorkflowContext(ctx workflow.Context) ddtrace.Span {
```
Quinn-With-Two-Ns, line 147: We should match the equivalent [SpanFromContext](https://github.com/DataDog/dd-trace-go/blob/58e1e634e199676dd2ec5ce89600da91a282b253/ddtrace/tracer/context.go#L25) signature here, so return a ` (Span, bool)` and return a noop span if none is set

Quinn-With-Two-Ns (changes requested):

contrib/datadog/tracing/interceptor.go
cdimitroulas: @Quinn-With-Two-Ns how can I created a noop span? The `NoopSpan` is exported from `gopkg.in/DataDog/dd-trace-go.v1/ddtrace/internal` so it's not accessible outside the datadog tracing package

cdimitroulas: We could implement our own NoopSpan here but this seems like a lot of extra code for little benefit

contrib/datadog/tracing/interceptor.go
cretz: Actually, I agree with @Quinn-With-Two-Ns below concerning noop span

contrib/datadog/tracing/interceptor.go
cretz: I think `ddtrace.SpanFromContext(nil)` will do it

cdimitroulas: Thank you for the suggestion, I have updated it to use this approach now

Quinn-With-Two-Ns (approved):

contrib/datadog/tracing/interceptor.go
Quinn-With-Two-Ns: Great thank you!

PR #1718, opened 2024-11-20
When evicting workflows from cache make sure all go routines are evicted serially

Quinn-With-Two-Ns:
When evicting workflows from cache make sure all go routines are evicted serially.

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -47,6 +47,7 @@ import (
 
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/internal/common/metrics"
+	"go.temporal.io/sdk/log"
 )
 
 const (
@@ -193,6 +194,7 @@ type (
 		mutex            sync.Mutex // used to synchronize executing
 		closed           bool
 		interceptor      WorkflowOutboundInterceptor
+		logger           log.Logger
 		deadlockDetector *deadlockDetector
 		readOnly         bool
 		// allBlockedCallback is called when all coroutines are blocked,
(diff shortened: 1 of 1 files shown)


cretz (approved):

internal/internal_workflow.go
```diff
@@ -1158,21 +1163,31 @@ func (s *coroutineState) close() {
...
+			st, err := getCoroStackTrace(s, "running", 0)
+			if err != nil {
+				st = fmt.Sprintf("<%s>", err)
+			}
+
+			logger.Warn(fmt.Sprintf("Workflow coroutine %q didn't exit within %v", s.name, warnTimeout), "stackTrace", st)
 		}
+		// We need to make sure the coroutine is closed, otherwise we risk concurrent coroutines running
+		// at the same time causing a race condition.
+		<-s.aboutToBlock
```
yuandrew, line 1190: How does this accomplish the comment above? If we receive a signal from `aboutToBlock` we exit cleanly, and otherwise if the timer fire first, why do we send a signal to `aboutToBlock` when the timer expires? 

Is it just to trigger the next coroutine exiting that it can exit cleanly?

Quinn-With-Two-Ns: We don't send a signal to `aboutToBlock`, we wait on receiving a signal from `aboutToBlock`, which is sent on exit

yuandrew: oops, misread this statement as sending a signal.. thanks for clarifying. So this function's behavior is the same, with an added log statement on timeout.

I'm a little confused by the PR title, this seems more like adding logs to help identify if evicted co-routines are timing out/exiting in the same order, but there is no code to actually make sure this is happening?

Quinn-With-Two-Ns: It behaves a bit differently because previously we didn't have this second `<-s.aboutToBlock` to if the coroutine didn't exit fast enough we would end up evicting multiple coroutines in parallel. Now we always wait for `<-s.aboutToBlock` to fire before we continue

yuandrew: You're right.. Maybe I need more coffee.. thanks for clarifying!

yuandrew (approved):

PR #1721, opened 2024-11-21
Improve Update Workflow testing UX

### LINKED ISSUE #1639: Improve UX for testing Workflow Update with the Workflow Test Environment
Improve UX for testing Workflow Update with the Workflow Test Environment. Currently users using the TestEnviorment have to implement the `UpdateCallbacks` to assert the result of an update call

Provide helpers in the SDK for common update testing scenarios

**Additional context**
https://github.com/temporalio/samples-go/blob/main/reqrespupdate/workflow_test.go#L40

yuandrew:
## What was changed

Added default implementation `TestUpdateCallbacks` for basic test scenarios

## Why?

so tests don't need to implement `UpdateCallbacks` themselves. To make testing a little easier

## Checklist

1. Closes #1639

2. How was this tested:

existing unit tests that were effected still pass

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/nexus-rpc/sdk-go/nexus"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 
@@ -82,6 +83,15 @@ type (
 
 		runFn        func(args mock.Arguments)
 		waitDuration func() time.Duration
+	}
+
+	// TestUpdateCallback is a basic implementation of the UpdateCallbacks interface for testing purposes.
+	// Tests are welcome to implement their own version of this interface if they need to test more complex
+	// update logic. This is a simple implementation to make testing basic Workflow Updates easier.
+	TestUpdateCallback struct {
+		accept   func()
+		reject   func(error)
+		complete func(interface{}, error)
 	}
 )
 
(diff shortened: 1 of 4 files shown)


yuandrew:
@Quinn-With-Two-Ns  lmk if there's other improvements you were thinking about when you filed the issue that I didn't cover

internal/workflow_testsuite.go
```diff
@@ -83,6 +84,15 @@ type (
 		runFn        func(args mock.Arguments)
 		waitDuration func() time.Duration
 	}
+
+	// TestUpdateCallback is a basic implementation of the UpdateCallbacks interface for testing purposes.
+	// Tests are welcome to implement their own version of this interface if they need to test more complex
+	// update logic. This is a simple implementation to make testing basic Workflow Updates easier.
+	TestUpdateCallback struct {
```
Quinn-With-Two-Ns, line 91: The issue was intended to improve the UX for users so we would need to expose this https://github.com/temporalio/sdk-go/blob/master/testsuite/testsuite.go#L35

yuandrew: oops! didn't realize this was needed. Thanks for pointing it out

internal/workflow_testsuite.go
```diff
@@ -83,6 +84,15 @@ type (
...
+		accept   func()
```
Quinn-With-Two-Ns, line 92: These fields are private so you also need to make these public so the type can be constructed. To make them public you need to capitalize the first letter. I would also suggest prefixing them with `On` to be a bit more consistent with the other test environment API.

yuandrew: Thanks :) tested with https://github.com/temporalio/samples-go/pull/371, hopefully not missing anything else

Quinn-With-Two-Ns (approved):

Issue #1712, opened 2024-11-19
Add support for the UpdateWorkflowExecutionOptions API


antlai-temporal:
A new API https://github.com/temporalio/api/pull/466 has been added that enables external tools to override versioning attributes for an existing workflow.

In the future, it will be extended to override other attributes, such as memo.

The ultimate goal is to provide support in the CLI, but we need to add SDK-go support first.

antlai-temporal:
Merged in versioning-3 branch

PR #1722, opened 2024-11-22
Update workflow execution options (versioning)

antlai-temporal:
## What was changed

The new versioning-3 introduces a config override API that enables external tools to change versioning behavior of
specific existing workflows. See #1712 for details. 

This PR exposes that API with a new client method:
```go
	UpdateWorkflowExecutionOptions(ctx context.Context, options UpdateWorkflowExecutionOptionsRequest) (WorkflowExecutionOptions, error)
```
and relies on 
```go
	DescribeWorkflowExecution(ctx context.Context, workflowID, runID string (*workflowservice.DescribeWorkflowExecutionResponse, error)
```
as getter.
## Checklist

1. Closes 
#1712 
2. How was this tested:
There is currently no server implementation that supports this API. Integration tests will be added later, before merge into main. Only a basic protobuf unit test is included.

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -1057,6 +1057,54 @@ func (wc *WorkflowClient) GetWorkerTaskReachability(ctx context.Context, options
 	return converted, nil
 }
 
+// GetWorkflowExecutionOptions returns the current WorkflowExecutionOptions for an existing workflow.
+// NOTE: Experimental
+func (wc *WorkflowClient) GetWorkflowExecutionOptions(ctx context.Context, request GetWorkflowExecutionOptionsRequest) (WorkflowExecutionOptions, error) {
+	// no side-effects, i.e., GET, when UpdateMask is empty and WorkflowExecutionOptions is not nil
+	return wc.UpdateWorkflowExecutionOptions(ctx, UpdateWorkflowExecutionOptionsRequest{
+		WorkflowId:               request.WorkflowId,
+		RunId:                    request.RunId,
+		RequestId:                uuid.New(),
+		WorkflowExecutionOptions: WorkflowExecutionOptions{},
+		UpdatedFields:            []string{},
+	})
+}
+
+// UpdateWorkflowExecutionOptions partially overrides the WorkflowExecutionOptions of an existing workflow execution,
+// and returns the new WorkflowExecutionOptions after applying the changes.
+// It is intended for building tools that can selectively apply ad-hoc workflow configuration changes.
+// NOTE: Experimental
+func (wc *WorkflowClient) UpdateWorkflowExecutionOptions(ctx context.Context, request UpdateWorkflowExecutionOptionsRequest) (WorkflowExecutionOptions, error) {
+	if err := wc.ensureInitialized(ctx); err != nil {
+		return WorkflowExecutionOptions{}, err
+	}
+
+	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
+	defer cancel()
+
+	if request.RequestId == "" {
+		request.RequestId = uuid.New()
+	}
+
+	requestMsg := &workflowservice.UpdateWorkflowExecutionOptionsRequest{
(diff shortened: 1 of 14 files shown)


internal/internal_workflow_client.go
```diff
@@ -1057,6 +1057,48 @@ func (wc *WorkflowClient) GetWorkerTaskReachability(ctx context.Context, options
 	return converted, nil
 }
 
+// GetWorkflowExecutionOptions returns the current WorkflowExecutionOptions of an existing workflow.
+// NOTE: Experimental
+func (wc *WorkflowClient) GetWorkflowExecutionOptions(ctx context.Context, request GetWorkflowExecutionOptionsRequest) (WorkflowExecutionOptions, error) {
+	// no side-effects, i.e., GET, when UpdateMask is empty and WorkflowExecutionOptions is not nil
```
cretz, line 1063: Eek, I didn't notice in https://github.com/temporalio/api/pull/466 that we didn't provide a getter and we were expecting people to use an update call with an empty mask as a getter. I think we need an explicit getter and we need to fail on empty update mask server side. We need to clearly mark mutable RPC vs immutable where we can (it can also affect action recording logic) and it'll be silly for someone to have to `POST` to get options when using the HTTP API.

@carlydf - can we add a getter, or even better, can we just provide this information on describe response?

internal/internal_workflow_client.go
```diff
@@ -1057,6 +1057,48 @@ func (wc *WorkflowClient) GetWorkerTaskReachability(ctx context.Context, options
...
+		WorkflowExecutionOptions: WorkflowExecutionOptions{},
+		UpdatedFields:            []string{},
+	})
+}
+
+// UpdateWorkflowExecutionOptions partially overrides the WorkflowExecutionOptions of an existing workflow execution,
+// and returns the new WorkflowExecutionOptions after applying the changes.
+// It is intended for building tools that can selectively apply ad-hoc workflow configuration changes.
+// NOTE: Experimental
+func (wc *WorkflowClient) UpdateWorkflowExecutionOptions(ctx context.Context, request UpdateWorkflowExecutionOptionsRequest) (WorkflowExecutionOptions, error) {
```
cretz, line 1076: Calling this a "request" is unique amongst our high-level client calls (i.e. ones that don't just forward the protos). I understand the apprehension for having `OptionsOptions`, but we are the ones that chose to use the word "options" in the API when the "properties" event was available in history. I don't have a strong opinion here, it's just a bit confusing to have a new, inconsistent concept (using a "request" vs "options") for naming reasons.

antlai-temporal: I was coping what we did for `QueryWorkflowWithOptionsRequest` but happy to call it to something else. Definetly not OptionsOptions, any suggestions?

internal/internal_workflow_client.go
antlai-temporal: All these are good reasons, in particular the HTTP API, @carlydf are we late to add the getter?

antlai-temporal: I discussed with @carlydf  and @ShahabT . `DescribeWorkflowExecution` will return that info and it is not more expensive than returning the options in a getter. So I will remove the getter from the sdk and add a comment in Update to use `DescribeWorkflowExecution` as a getter... Is that ok @cretz ?

cretz: > So I will remove the getter from the sdk and add a comment in Update to use DescribeWorkflowExecution as a getter

Great! I think this is the best of all worlds.

internal/internal_workflow_client.go
cretz: Yeah, we are unfortunately inconsistent here (same with whether this is a pointer or not, though I think it should remain not a pointer). I am fine with what you have here as a request, though would like to wait for @Quinn-With-Two-Ns's opinion too.

antlai-temporal: I need to move on, trying to make a dependent PR today, but we can revisit naming before merging into main...

PR #1723, opened 2024-11-23
Fix worker leak in eager dispatcher

Quinn-With-Two-Ns:
Fix worker leak in eager dispatcher, before workers would not be removed from the `eagerWorkflowDispatcher` task queue causing a subtle memory leak if workers were started and stopped in a loop.


internal/internal_eager_workflow.go
```diff
@@ -33,24 +33,37 @@ import (
...
+		e.workersByTaskQueue[taskQueue] = make(map[eagerWorker]struct{})
+	}
+	e.workersByTaskQueue[taskQueue][worker.worker] = struct{}{}
+}
+
+// deregisterWorker deregister a worker so that it will not be used for eager workflow dispatch
+func (e *eagerWorkflowDispatcher) deregisterWorker(worker *workflowWorker) {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	delete(e.workersByTaskQueue[worker.executionParameters.TaskQueue], worker.worker)
```
cretz, line 54: IIRC it is unfortunate but known that Go does not reclaim memory for map entries on `delete`. At least it was the case in the older Go versions, but maybe it has changed (see issue https://github.com/golang/go/issues/ + 20135).

internal/internal_eager_workflow.go
```diff
@@ -33,24 +33,37 @@ import (
...
 	e.lock.RLock()
 	workers := e.workersByTaskQueue[request.GetTaskQueue().Name]
-	randWorkers := make([]eagerWorker, len(workers))
-	// Copy the slice so we can release the lock.
-	copy(randWorkers, workers)
+	randWorkers := make([]eagerWorker, 0, len(workers))
+	// Copy the workers so we can release the lock.
+	for worker := range workers {
+		randWorkers = append(randWorkers, worker)
+	}
```
cretz, line 66: Since deregister is rare I figure, it could be better maybe to burden it with the heavier work than this call, since I suspect `copy` is cheaper than building this from keys. But this probably doesn't matter since I can't imagine a good use case where a user would ever have more than one workflow worker for a task queue (versioning notwithstanding since this isn't built for versioning). So nothing needed, just thinking out loud.

internal/internal_eager_workflow.go
Quinn-With-Two-Ns: This issue is about the memory used by the map, what we care about is making sure nothing is holding a reference to `worker.worker` so it can be GC'd

yuandrew (approved):

PR #1731, opened 2024-11-28
New Update-With-Start API

dandavison:
New Update-With-Start API.

- Instead of using `ExecuteWorkflow`, update-with-start is invoked via a new method `UpdateWithStartWorkflow` on `Client` which returns an `UpdateHandle`
- The method receives a `WithStartWorkflowOperation` struct which defines the start operation, and exposes a method to obtain the workflow run:

The snippet below is from the accompanying sample: https://github.com/temporalio/samples-go/pull/370:

```go
tx := earlyreturn.Transaction{ID: uuid.New(), SourceAccount: "Bob", TargetAccount: "Alice", Amount: 100}

startWorkflowOp := c.NewWithStartWorkflowOperation(client.StartWorkflowOptions{
	ID:                       "early-return-workflow-ID-" + tx.ID,
	WorkflowIDConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_FAIL,
	TaskQueue:                earlyreturn.TaskQueueName,
}, earlyreturn.Workflow, tx)

updateHandle, err := c.UpdateWithStartWorkflow(ctxWithTimeout, client.UpdateWithStartWorkflowOptions{
	UpdateOptions: client.UpdateWorkflowOptions{
		UpdateName:   earlyreturn.UpdateName,
		WaitForStage: client.WorkflowUpdateStageCompleted,
	},
	StartOperation: startWorkflowOp,
})
if err != nil {
	log.Fatalln("Error issuing update-with-start:", err)
}
var earlyReturnResult any
err = updateHandle.Get(ctxWithTimeout, &earlyReturnResult)
if err != nil {
	// The workflow will continue running, cancelling the transaction.

	// NOTE: If the error is retryable, a retry attempt must use a unique workflow ID.
	log.Fatalln("Error obtaining update result:", err)
}

workflowRun, err := startWorkflowOp.Get(ctxWithTimeout)
```


internal/client.go
```diff
@@ -135,6 +134,10 @@ type (
 		SignalWithStartWorkflow(ctx context.Context, workflowID string, signalName string, signalArg interface{},
 			options StartWorkflowOptions, workflow interface{}, workflowArgs ...interface{}) (WorkflowRun, error)
 
+		// NewWithStartWorkflowOperation returns a WithStartWorkflowOperation to perform Update-with-Start.
+		// NOTE: Experimental
+		NewWithStartWorkflowOperation(options StartWorkflowOptions, workflow any, args ...interface{}) (*WithStartWorkflowOperation, error)
```
Quinn-With-Two-Ns, line 139: For mock-ability `WithStartWorkflowOperation` should be an interface

dandavison: Right thanks -- I completely forgot to make that change after our discussion.

client/client.go
```diff
@@ -564,6 +559,11 @@ type (
 		SignalWithStartWorkflow(ctx context.Context, workflowID string, signalName string, signalArg interface{},
 			options StartWorkflowOptions, workflow interface{}, workflowArgs ...interface{}) (WorkflowRun, error)
 
+		// NewWithStartWorkflowOperation returns a WithStartWorkflowOperation to perform Update-with-Start.
+		// Returns an error if the WorkflowIDConflictPolicy is not set, or if the workflow or arguments are invalid.
+		// NOTE: Experimental
+		NewWithStartWorkflowOperation(options StartWorkflowOptions, workflow any, args ...interface{}) (*WithStartWorkflowOperation, error)
```
Quinn-With-Two-Ns, line 565: Why have one parameter take `any` and other take `interface{}`? They are alias's, but we should be consistent. I think the Go SDK generally takes `interface{}`

client/client.go
```diff
@@ -162,16 +162,11 @@ type (
 	// StartWorkflowOptions configuration parameters for starting a workflow execution.
 	StartWorkflowOptions = internal.StartWorkflowOptions
 
-	// WithStartWorkflowOperation is a type of operation that can be executed as part of a workflow start.
-	// For example, use NewUpdateWithStartWorkflowOperation to perform Update-with-Start.
+	// WithStartWorkflowOperation defines how to start a workflow when using Update-With-Start.
+	// See NewWithStartWorkflowOperation for details.
```
Quinn-With-Two-Ns, line 166:
```suggestion
	// See [Client.NewWithStartWorkflowOperation] for details.
```

I think this will make it a link

dandavison: Done

client/client.go
dandavison: It was unintended that they weren't consistent. Switched to `interface{}`. I personally prefer to use `any` but agree, we should be consistent thrughout the SDK.

client/client.go
```diff
@@ -564,6 +559,11 @@ type (
 		SignalWithStartWorkflow(ctx context.Context, workflowID string, signalName string, signalArg interface{},
 			options StartWorkflowOptions, workflow interface{}, workflowArgs ...interface{}) (WorkflowRun, error)
 
+		// NewWithStartWorkflowOperation returns a WithStartWorkflowOperation to perform Update-with-Start.
+		// Returns an error if the WorkflowIDConflictPolicy is not set, or if the workflow or arguments are invalid.
+		// NOTE: Experimental
+		NewWithStartWorkflowOperation(options StartWorkflowOptions, workflow interface{}, args ...interface{}) (*WithStartWorkflowOperation, error)
```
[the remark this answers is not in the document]
dandavison: Yes, the SDK has a workflow name aliasing feature which uses a registry on the client instance: https://github.com/temporalio/sdk-go/blob/158b8231aef6bd73f4e632420899ad1b7f36cb89/internal/internal_worker.go#L1859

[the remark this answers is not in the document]
cretz: Can't the workflow input be created as part of `UpdateWithStartWorkflow` instead of on instantiation of this? One might expect the actual call to do all of the validation instead of half here and half on the actual call. Arguably this should not return an error.

client/client.go
```diff
@@ -564,6 +559,11 @@ type (
 		SignalWithStartWorkflow(ctx context.Context, workflowID string, signalName string, signalArg interface{},
 			options StartWorkflowOptions, workflow interface{}, workflowArgs ...interface{}) (WorkflowRun, error)
 
+		// NewWithStartWorkflowOperation returns a WithStartWorkflowOperation to perform Update-with-Start.
+		// Returns an error if the WorkflowIDConflictPolicy is not set, or if the workflow or arguments are invalid.
+		// NOTE: Experimental
+		NewWithStartWorkflowOperation(options StartWorkflowOptions, workflow interface{}, args ...interface{}) WithStartWorkflowOperation
```
cretz, line 565:
```suggestion
		NewWithStartWorkflowOperation(options WithStartWorkflowOperationOptions) WithStartWorkflowOperation
```

Arguably this is more future proof, but understood if we don't want it

client/client.go
```diff
@@ -840,6 +840,11 @@ type (
 		// NOTE: Experimental
 		UpdateWorkflow(ctx context.Context, options UpdateWorkflowOptions) (WorkflowUpdateHandle, error)
 
+		// UpdateWithStartWorkflow intercepts client.Client.UpdateWithStartWorkflow.
+		//
+		// NOTE: Experimental
+		UpdateWithStartWorkflow(ctx context.Context, options UpdateWorkflowOptions, startOperation WithStartWorkflowOperation) (WorkflowUpdateHandle, error)
```
cretz, line 846:
```suggestion
		UpdateWithStartWorkflow(ctx context.Context, options UpdateWithStartWorkflowOptions) (WorkflowUpdateHandle, error)
```

I think an options structure here accepting the two possible parameters here is more future proof

client/client.go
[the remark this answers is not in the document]
cretz: After internal discussion, to allow users to provide their own implementation of this stateful thing, we are making the result an interface and keeping it on the client

client/client.go
dandavison: This seems to be going too far to me: it doesn't seem right to have to create a  `WithStartWorkflowOperationOptions` struct in order to create a `WithStartWorkflowOperation` given that the latter essentially is just a stateful version of the former.

client/client.go
dandavison: Thanks, discussed internally and taken this suggestion: 88e4896b3ae8b035af6c9c481601660ae6502134

client/client.go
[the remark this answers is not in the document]
dandavison: Now returns interface

internal/internal_workflow_client.go
```diff
@@ -1180,6 +1186,37 @@ func (wc *WorkflowClient) UpdateWorkflow(
...
+	ctx context.Context,
+	options UpdateWithStartWorkflowOptions,
+) (WorkflowUpdateHandle, error) {
+	startOp, ok := options.StartOperation.(*WithStartWorkflowOperationImpl)
+	if !ok {
+		return nil, errors.New("startOperation must be created by NewWithStartWorkflowOperation")
+	}
+	if startOp.err != nil {
+		return nil, startOp.err
+	}
```
cretz, line 1199: This negates the benefit of returning the interface. Can just make it optional at this level.

```suggestion
	if startOp, _ := options.StartOperation.(*WithStartWorkflowOperationImpl); startOp != nil && startOp.err != nil {
		return nil, startOp.err
	}
```

But it is understood that in the root interceptor the raw type is needed

internal/internal_workflow_client.go
```diff
@@ -1669,55 +1734,105 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
-	startOp := &workflowservice.ExecuteMultiOperationRequest_Operation{
-		Operation: &workflowservice.ExecuteMultiOperationRequest_Operation_StartWorkflow{
-			StartWorkflow: startRequest,
-		},
+	updateInput *ClientUpdateWorkflowInput,
+	startOperation WithStartWorkflowOperation,
+) (WorkflowUpdateHandle, error) {
+	startOp, ok := startOperation.(*WithStartWorkflowOperationImpl)
+	if !ok {
+		panic("startOperation must be created by NewWithStartWorkflowOperation")
```
cretz, line 1744: Might as well just return an error here same as you had pre-interception

internal/internal_workflow_client.go
```diff
@@ -1669,55 +1734,105 @@ func (w *workflowClientInterceptor) ExecuteWorkflow(
...
-		if err := t.markExecuted(); err != nil {
-			return nil, fmt.Errorf("%w: %w", errInvalidWorkflowOperation, err)
-		}
+	// Create start request
+	if err := startOp.markExecuted(); err != nil {
+		return nil, fmt.Errorf("%w: %w", errInvalidWithStartWorkflowOperation, err)
+	}
+	if startOp.err != nil {
+		return nil, fmt.Errorf("%w: %w", errInvalidWithStartWorkflowOperation, startOp.err)
+	}
```
cretz, line 1753: What situations might this error be non-nil here? Does this check need to occur before the `markExecuted` above here?

cretz (commented):
Mostly little stuff

client/client.go
```diff
@@ -564,6 +564,11 @@ type (
 		SignalWithStartWorkflow(ctx context.Context, workflowID string, signalName string, signalArg interface{},
 			options StartWorkflowOptions, workflow interface{}, workflowArgs ...interface{}) (WorkflowRun, error)
 
+		// NewWithStartWorkflowOperation returns a WithStartWorkflowOperation for use with UpdateWithStartWorkflow.
+		// See [Client.UpdateWithStartWorkflow].
+		// NOTE: Experimental
+		NewWithStartWorkflowOperation(options StartWorkflowOptions, workflow interface{}, args ...interface{}) WithStartWorkflowOperation
```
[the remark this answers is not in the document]
dandavison: > I personally liked that this returned a validation error, but it's not that important to me

Hey! I changed it to not return an error in response to you [saying](https://github.com/temporalio/sdk-go/pull/1731#discussion_r1865922908) this:

> Can't the workflow input be created as part of UpdateWithStartWorkflow instead of on instantiation of this? One might expect the actual call to do all of the validation instead of half here and half on the actual call. Arguably this should not return an error.

internal/internal_workflow_client.go
dandavison: Thanks I had intended to convert all those panics to errors, but I missed this one.

internal/client.go
```diff
@@ -751,22 +753,22 @@ type (
 		links []*commonpb.Link
 	}
 
-	// WithStartWorkflowOperation is a type of operation that can be executed as part of a workflow start.
+	// WithStartWorkflowOperation defines how to start a workflow when using UpdateWithStartWorkflow.
+	// See [NewWithStartWorkflowOperation] and [UpdateWithStartWorkflow].
+	// NOTE: Experimental
 	WithStartWorkflowOperation interface {
-		isWithStartWorkflowOperation()
+		Get(ctx context.Context) (WorkflowRun, error)
```
Quinn-With-Two-Ns, line 760: docs?

internal/internal_workflow_client.go
Quinn-With-Two-Ns: If a users is mocking the client then this code is not relevant

cretz: Technically another benefit of the interface here is that it could be a different impl and be made into the proper impl by an interceptor, but I agree that is far fetched. This code doesn't have to change.

client/client.go
dandavison: > which disables the ability to discuss further

Apparently not :)

>  I am ok if we all agree this isn't needed

Indeed, the consensus with others was that it wasn't needed.

> I fear we will add some server-side option that only applies to multiop start workflow and we won't be able to represent it here.

We would be able to add such an option to `UpdateWithStartWorkflowOptions`.

internal/client.go
```diff
@@ -751,22 +753,22 @@ type (
...
 	}
 
-	// UpdateWithStartWorkflowOperation is used to perform Update-with-Start.
-	// See NewUpdateWithStartWorkflowOperation for details.
-	UpdateWithStartWorkflowOperation struct {
-		input *ClientUpdateWorkflowInput
+	WithStartWorkflowOperationImpl struct {
```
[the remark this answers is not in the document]
dandavison: Thanks, good call.

internal/internal_workflow_client.go
dandavison: Thanks, I agree, it makes more sense above `markExecuted`. Changed. It is non-nil if the conflict policy is not set.

internal/interceptor.go
```diff
@@ -393,6 +393,11 @@ type ClientOutboundInterceptor interface {
 	// NOTE: Experimental
 	UpdateWorkflow(context.Context, *ClientUpdateWorkflowInput) (WorkflowUpdateHandle, error)
 
+	// UpdateWithStartWorkflow intercepts client.Client.UpdateWithStartWorkflow.
+	//
+	// NOTE: Experimental
+	UpdateWithStartWorkflow(context.Context, *ClientUpdateWorkflowInput, WithStartWorkflowOperation) (WorkflowUpdateHandle, error)
```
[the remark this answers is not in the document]
dandavison: Thanks, done.

internal/client.go
dandavison: Thanks, added.

internal/interceptor.go
```diff
@@ -416,6 +421,11 @@ type ClientUpdateWorkflowInput struct {
 	WaitForStage        WorkflowUpdateStage
 }
 
+type ClientUpdateWithStartWorkflowInput struct {
+	UpdateInput            *ClientUpdateWorkflowInput
+	StartWorkflowOperation WithStartWorkflowOperation
```
cretz, line 426: May be worth noting that the interceptor can't do much with this. They can't even know what workflow it applies to. We don't have to do anything now, but I would expect at some point for interceptor authors to ask how to get details of the start the same way they can get details of the update.

internal/interceptor.go
```diff
@@ -416,6 +421,11 @@ type ClientUpdateWorkflowInput struct {
 	WaitForStage        WorkflowUpdateStage
 }
 
+type ClientUpdateWithStartWorkflowInput struct {
+	UpdateInput            *ClientUpdateWorkflowInput
```
cretz, line 425: This is input for another interceptor call, I don't think we should reuse it for this one. The interceptor input really just needs to be the values passed in by the user. Maybe something like:

```go
type ClientUpdateWithStartWorkflowInput struct {
    StartOptions  *StartWorkflowOptions
    StartWorkflowType string
    StartWorkflowArgs []any
    UpdateID string
    UpdateName string
    UpdateArgs []any
    UpdateWaitForStage WorkflowUpdateStage
}
```

Or something like:

```go
type ClientUpdateWithStartWorkflowInput struct {
    StartOptions  *StartWorkflowOptions
    StartWorkflowType string
    StartWorkflowArgs []any
    UpdateOptions *UpdateWorkflowOptions
}
```

Or something like:

```go
type ClientUpdateWithStartWorkflowInput struct {
    UpdateOptions *UpdateWorkflowOptions
    StartOperation WithStartWorkflowOperation
}
```

and add `Options() StartWorkflowOptions` and `WorkflowType() string` and `WorkflowArgs() []any` to `WithStartWorkflowOperation` interface.

dandavison: What problem are you trying to solve? This is how it's done in Java:

```java
    public WorkflowUpdateWithStartInput(
        WorkflowStartInput workflowStartInput, StartUpdateInput<R> workflowUpdateInput) {
```

Quinn-With-Two-Ns: I'll note Java does reuse, but I think it is the only SDK that takes such an approach 

https://github.com/temporalio/sdk-java/blob/c7fcf12f4fc1965cde8a70feb0a109f198298088/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java#L223

dandavison: Ah right I see, thank you @Quinn-With-Two-Ns and @cretz for the suggested schemas. I've switched to 

```go
type ClientUpdateWithStartWorkflowInput struct {
	UpdateOptions          *UpdateWorkflowOptions
	StartWorkflowOperation WithStartWorkflowOperation
}
```

9562e0e4e64f6062aeb56fccde9362b610eaa913

client/client.go
cretz: > We would be able to add such an option to UpdateWithStartWorkflowOptions.

Until we use this `WithStartWorkflowOperation` for more than just update with start, which is why I figured it was named to not be related to update. But it's not a big deal.

cretz (approved):

Quinn-With-Two-Ns (approved):

client/client.go
dandavison: I might be missing something but if we

> add some server-side option that only applies to multiop start workflow

we would at least be able to add as a top-level field here:

```go
type UpdateWithStartWorkflowOptions struct {
	StartWorkflowOperation WithStartWorkflowOperation
	UpdateOptions          UpdateWorkflowOptions
}
```

Issue #1713, opened 2024-11-19
Support the Deployment API


antlai-temporal:
The Deployment API https://github.com/temporalio/api/pull/479 is a low-level API that enables Worker versioning-3 by providing a CRUD interface to manage deployments. 

It also enables directly setting the current deployment target, forcing redirection or creation of workflows to a different worker pool.

This is an experimental API part of the versioning-3 effort.

PR #1733, opened 2024-12-02
Add deployment API

antlai-temporal:
## What was changed

This PR implements the new deployment API  needed for worker versioning. It provides a new `DeploymentClient` that has methods:

```go
Describe(ctx context.Context, deploymentID Deployment) (DeploymentInfo, error)
List(ctx context.Context, options DeploymentListOptions) (DeploymentListIterator, error)
GetReachability(ctx context.Context, deploymentID Deployment) (DeploymentReachabilityInfo, error)
GetCurrent(ctx context.Context, seriesName string) (DeploymentInfo, error)
SetCurrent(ctx context.Context, options DeploymentSetCurrentOptions) (DeploymentSetCurrentResponse, error)
```

## Checklist

1. Closes 
#1713 
2. How was this tested:

Unit tests following the ScheduleClient template. System tests in a separate PR when there is a server that supports it.


internal/deployment_client.go
```diff
@@ -0,0 +1,225 @@
...
+		// Previous - Information about the last current deployment, i.e., before this operation.
+		Previous DeploymentInfo
+	}
+
+	// DeploymentClient is the client that manages deployments.
+	// NOTE: Experimental
+	DeploymentClient interface {
+		// Describes an existing deployment.
+		// NOTE: Experimental
+		Describe(ctx context.Context, deploymentID Deployment) (DeploymentInfo, error)
```
cretz, line 203: I wonder if for `Describe` and `GetCurrent` you'd rather have options structures just to be future proof (Go is a bit annoying here w/ no overloads or optional params).

cretz (commented):
Mostly LGTM, one slight suggestion

internal/deployment_client.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	// DeploymentMetadataUpdate modifies user-defined metadata entries that describe
+	// a deployment.
+	// NOTE: Experimental
+	DeploymentMetadataUpdate struct {
+		// UpsertEntries - Metadata entries inserted or modified. When values are not
+		// of type *commonpb.Payload, the default data converter will be used to generate
+		// payloads.
+		UpsertEntries map[string]interface{}
```
Quinn-With-Two-Ns, line 168: > the default data converter will be used to generate payloads.

Why? does the Temporal server need to read these?

antlai-temporal: The idea of payloads was to support encryption. Converting to payloads transparently in UpsertEntries was just for convenience, they can always pass payloads if they want...

internal/deployment_client.go
antlai-temporal: Makes sense, it will be a bit more annoying to use but it is worth avoiding a `DescribeWithOptions`...

internal/deployment_client.go
Quinn-With-Two-Ns: We should use the clients data converter here then

Quinn-With-Two-Ns: We should only use the default if the server needs to read the payload

internal/deployment_client.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	// DeploymentSetCurrentOptions provides options for DeploymentClient.SetCurrent.
+	// NOTE: Experimental
+	DeploymentSetCurrentOptions struct {
+		// Deployment - An identifier for this deployment.
+		Deployment Deployment
+
+		// ClientIdentity - Optional: The identity of the client who initiated this request.
+		ClientIdentity string
```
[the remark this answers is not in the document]
antlai-temporal: Good point,  I'll send the client identitiy as default, they may want to override it for versioning...

internal/deployment_client.go
antlai-temporal: I'm using `converter.GetDefaultDataConverter().ToPayload`, I thought that would give me client data converter, which one should I be using?

antlai-temporal: I see `getDataConverterFromWorkflowContext()` may give me what I need, it defaults to the default data converter if none provided...

Quinn-With-Two-Ns: `WorkflowClient.dataConverter`

Quinn-With-Two-Ns: `GetDefaultDataConverter()` is a program wide data converter, not the clients

ShahabT: Server does not need to read these.

internal/deployment_client.go
antlai-temporal: `Deployment` is already a struct, so that will give us a escape hatch if we need a deployment id with other fields. Adding `DeploymentGetCurrentOptions` that was just a string argument...

internal/deployment_client.go
antlai-temporal: Done

internal/client.go
```diff
@@ -417,6 +417,9 @@ type (
 		// Schedule creates a new shedule client with the same gRPC connection as this client.
 		ScheduleClient() ScheduleClient
 
+		// DeploymentClient creates a new deployment client with the same gRPC connection as this client.
+		DeploymentClient() DeploymentClient
```
Quinn-With-Two-Ns, line 421: nit: should be experimental

internal/deployment_client.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: No other call in the SDK needed an override, I would leave off until there is an ask

Quinn-With-Two-Ns (approved):

internal/deployment_client.go
```diff
@@ -0,0 +1,243 @@
...
+	// NOTE: Experimental
+	DeploymentInfo struct {
+		// Deployment - An identifier for this deployment.
+		Deployment Deployment
+
+		// CreateTime - When this deployment was created.
+		CreateTime time.Time
+
+		// Current - Whether this deployment is the current one for its deployment series.
+		Current bool
```
ShahabT, line 103: Nit: wouldn't `IsCurrent` be more more idiomatic?

internal/deployment_client.go
```diff
@@ -0,0 +1,243 @@
...
+	// NOTE: Experimental
+	DeploymentListEntry struct {
+		// An identifier for this deployment.
+		Deployment Deployment
+
+		// When this deployment was created.
+		CreateTime time.Time
+
+		// Whether this deployment is the current one for its deployment series.
+		Current bool
```
ShahabT, line 122: same

internal/deployment_client.go
```diff
@@ -0,0 +1,243 @@
...
+	DeploymentReachabilityInfo struct {
+		DeploymentInfo
+
+		// Reachability - Kind of tasks that may reach a worker
+		// associated with a deployment.
+		Reachability DeploymentReachability
+
+		// LastUpdateTime - When reachability was last computed. Computing reachability
+		// is an expensive operation, and the server caches results.
+		LastUpdateTime time.Time
```
ShahabT, line 158: hmm.. if we embed `DeploymentInfo` there would be timestamps such `CreateTime` besides this `LastUpdateTime`, user wouldn't know if the field only refers to reachability status update time. Later more timestamps and other fields will be added to Deployment Info (and maybe to Reachability response) and I worry it'd be confusing for users if we mix them. Should we make deployment info just a nested field instead of an embed?

PR #1734, opened 2024-12-02
Make TraceID and SpanID Field Names Configurable in OpenTelemetry Integration

elaiboudimane:
This PR updates the Temporal Go SDK to make TraceID and SpanID field names configurable in the OpenTelemetry integration. By default, the fields remain TraceID and SpanID, ensuring backward compatibility. This change resolves inconsistencies when integrating Temporal logs with systems using different conventions.

Please review and let me know if additional changes or tests are needed.

### DIFF
--- a/contrib/opentelemetry/tracing_interceptor.go
+++ b/contrib/opentelemetry/tracing_interceptor.go
@@ -86,6 +86,14 @@ type TracerOptions struct {
 	// SpanStarter is a callback to create spans. If not set, this creates normal
 	// OpenTelemetry spans calling Tracer.Start.
 	SpanStarter func(ctx context.Context, t trace.Tracer, spanName string, opts ...trace.SpanStartOption) trace.Span
+
+	// TraceIDFieldName is the name of the field used to represent the Trace ID
+	// If empty, a default name will be used.
+	TraceIDFieldName string
+
+	// SpanIDFieldName is the name of the field used to represent the Span ID
+	// If empty, a default name will be used.
+	SpanIDFieldName string
 }
 
 type spanContextKey struct{}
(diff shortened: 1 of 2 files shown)


contrib/opentelemetry/tracing_interceptor.go
```diff
@@ -86,6 +86,14 @@ type TracerOptions struct {
...
 	// OpenTelemetry spans calling Tracer.Start.
 	SpanStarter func(ctx context.Context, t trace.Tracer, spanName string, opts ...trace.SpanStartOption) trace.Span
+
+	// TraceIDFieldName is the name of the field used to represent the Trace ID
+	// If empty, a default name will be used.
+	TraceIDFieldName string
+
+	// SpanIDFieldName is the name of the field used to represent the Span ID
+	// If empty, a default name will be used.
+	SpanIDFieldName string
```
cretz, line 96:
```suggestion
	// TraceIDLoggerFieldName is the name of the field used to represent the Trace ID in logger.
	// If empty, a default name will be used.
	TraceIDLoggerFieldName string

	// SpanIDLoggerFieldName is the name of the field used to represent the Span ID in logger.
	// If empty, a default name will be used.
	SpanIDLoggerFieldName string
```

Not clear this is only for logger and nothing else

cretz (commented):
@Quinn-With-Two-Ns - thoughts? We have several fields we add to logger in this SDK that we don't allow customizing of the names of. But I don't have any strong objections.

Quinn-With-Two-Ns:
In general to rename attributes we recommend doing replacement at the logger level. How depends on your logger, but here is an example for the standard library logger.

https://programmingpercy.tech/blog/structured-logging-in-go-using-std-lib-slog/#replacing-the-default-logger-and-replacing-attributes

elaiboudimane:
> In general to rename attributes we recommend doing replacement at the logger level. How depends on your logger, but here is an example for the standard library logger.
> 
> https://programmingpercy.tech/blog/structured-logging-in-go-using-std-lib-slog/#replacing-the-default-logger-and-replacing-attributes

Thank you for the suggestion! I agree, it’s a great idea. I’ll make the changes at the logger level as recommended and close the PR. Thanks for the helpful link!

elaiboudimane:
Closing this PR as the changes will be implemented at the logger level following the recommended approach. Thanks again for the feedback!

PR #1735, opened 2024-12-02
Update Godoc for internal structs/funcs to clarify the package they are exposed under

### LINKED ISSUE #1681: Update Godoc for `internal` structs/funcs to clarify the package they are exposed under

At least for all non-obvious `internal` structs/funcs, mention in their Godoc the package they are exposed/aliased at.

yuandrew:
## What was changed

Added a new tool that identifies public structs/funcs that map to internal fields, and detects if the internal fields have a doc comment referring to where it's publicly exposed as.

Passing by with no flags will only output missing mappings to stdout, but passing in `-fix` will automatically generate the mapping comments.

 
## Why?

Clarity

## Checklist

1. Closes #1681

2. How was this tested:

Manually inspected output


internal/activity.go
```diff
@@ -39,11 +39,15 @@ import (
 
 type (
 	// ActivityType identifies an activity type.
+	//
+	// Exposed as: activity:Type
```
Quinn-With-Two-Ns, line 43: hm Can we make these doc links https://tip.golang.org/doc/comment#doclinks?

Quinn-With-Two-Ns: That way they are navigable

internal/error.go
```diff
@@ -43,71 +43,71 @@ If activity fails then *ActivityError is returned to the workflow code. The erro
...
-	*ApplicationError can be returned in two cases:
-		- If activity implementation returns *ApplicationError by using NewApplicationError()/NewNonRetryableApplicationError() API.
-		  The error would contain a message and optional details. Workflow code could extract details to string typed variable, determine
-		  what kind of error it was, and take actions based on it. The details are encoded payload therefore, workflow code needs to know what
-          the types of the encoded details are before extracting them.
-		- If activity implementation returns errors other than from NewApplicationError() API. In this case GetOriginalType()
-		  will return original type of error represented as string. Workflow code could check this type to determine what kind of error it was
-		  and take actions based on the type. These errors are retryable by default, unless error type is specified in retry policy.
+*ApplicationError can be returned in two cases:
+- If activity implementation returns *ApplicationError by using NewApplicationError()/NewNonRetryableApplicationError() API.
```
Quinn-With-Two-Ns, line 47: There appears to be some unnecessary whitespace changes

internal/activity.go
yuandrew: yeah, good idea. Unfortunately it's not as clean because we have to import the entire package name, i.e. `go.temporal.io/sdk/temporal.ApplicationError`, but clickable is better

internal/error.go
yuandrew: oops, this was because I was stripping whitespace at the beginning and ending of lines, and `format.Source([]byte(newFile))` won't automatically add back spacing to comments. Fixed!

internal/cmd/tools/doclink/doclink.go
```diff
@@ -0,0 +1,513 @@
...
+			if len(isWrapper) > 0 {
+				publicToInternal[funcDecl.Name.Name] = isWrapper
+			}
+		}
+		return true
+	})
+	return publicToInternal, nil
+}
+
+func isPublic(name string) bool {
```
cretz, line 205: https://pkg.go.dev/go/ast#IsExported exists

internal/cmd/tools/doclink/doclink.go
```diff
@@ -0,0 +1,513 @@
...
+)
+
+var missing = false
+
+func main() {
+	if err := run(); err != nil {
+		log.Fatal(err)
+	}
+	if missing {
+		log.Fatal("Missing documentation, see previous stdout for which objects. Re-run command with -fix to auto-generate missing docs.")
```
cretz, line 54: This is great that it gives a bad exit code on missing docs. Can you add this step to CI?

yuandrew: yep! This was actually added to the check command https://github.com/temporalio/sdk-go/pull/1735/files#diff-265865b1cb4e06ecf08b23b205f709fb4281ba4d4804cbe46f8b91a725073fd5

cretz: Ah, great, sorry I missed that

cretz (approved):

Quinn-With-Two-Ns (approved):

Issue #1739, opened 2024-12-03
Fail registration of workflow types without versioning behavior


antlai-temporal:
Worker Versioning-3 adds the notion of workflow behavior to decide whether to
migrate existing workflows to new Build IDs. The worker can specify a default, 
but if there is no default, and it is not provided as an option for that workflow, 
it should fail. 

The current server implementation ignores the `UseVersioning` flag
and relies on missing behavior info to disable versioning. Therefore,
the server cannot fail the task, it has to be the SDK...

The solution proposed does two things:
- Removes the programmatic API to set versioning behavior. It is fragile, there 
is not a clear use case for the flexibility added, and makes it hard to fail workflows early (without side-effects).
- Add a check at workflow type registration to fail registration when there is no default, and no value provided.

PR #1742, opened 2024-12-03
Fail registration without versioning behavior

antlai-temporal:
## What was changed

Remove programmatic SetVersioningBehavior API, and fail workflow registration when behavior missing.See #1739 for details.

1. Closes 
#1739 
2. How was this tested:

One system test

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -413,7 +413,9 @@ type (
 		// inside a workflow as a child workflow.
 		Name                          string
 		DisableAlreadyRegisteredCheck bool
-		// Optional: Provides a Versioning Behavior to workflows of this type.
+		// Optional: Provides a Versioning Behavior to workflows of this type. It is required
+		// when WorkerOptions does not specify DeploymentOptions.DefaultVersioningBehavior,
+		// DeploymentOptions.DeploymentSeriesName is set, and UseBuildIDForVersioning is true.
 		// NOTE: Experimental
 		VersioningBehavior VersioningBehavior
 	}
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -413,8 +413,7 @@ type (
 		// inside a workflow as a child workflow.
 		Name                          string
 		DisableAlreadyRegisteredCheck bool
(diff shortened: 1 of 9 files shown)


Quinn-With-Two-Ns (approved):

internal/workflow.go
```diff
@@ -413,8 +413,7 @@ type (
 		// inside a workflow as a child workflow.
 		Name                          string
 		DisableAlreadyRegisteredCheck bool
-		// Optional: Provides a default Versioning Behavior to workflows of this type.
-		// See workflow.SetVersioningBehavior to override this default.
+		// Optional: Provides a Versioning Behavior to workflows of this type.
```
ShahabT, line 416: Should we mention that this is not optional of there is no default behavior and `DeploymentOptions.UseBuildIDForVersioning==true`?

test/integration_test.go
```diff
@@ -6351,68 +6326,41 @@ func (ts *IntegrationTestSuite) TestVersioningBehaviorPerWorkflowType() {
...
 	})
 	ts.NoError(err)
+	defer c.Close()
 
 	ts.worker.Stop()
 	ts.workerStopped = true
-	w := worker.New(ts.client, ts.taskQueueName, worker.Options{
+	w := worker.New(c, ts.taskQueueName, worker.Options{
 		BuildID:                 "1.0",
 		UseBuildIDForVersioning: true,
```
ShahabT, line 6346: is there any different between `UseBuildIDForVersioning` and `DeploymentOptions.UseBuildIDForVersioning`? Do we want to make `UseBuildIDForVersioning` exclusively for versioning 1-2 and the latter for versioning 3?

antlai-temporal: There is no `DeploymentOptions.UseBuildIDForVersioning` until we deprecate versioning-2, otherwise it is just too confusing to have same field twice without deprecation.

internal/workflow.go
antlai-temporal: Good point, I'll make this more clear.

PR #1743, opened 2024-12-04
mock: in order mock calls using new testify mock.InOrder() method

ReyOrtiz:
## What was changed

Upgrading testify to `v1.10.0`.

Adding an InOrderMockCalls(calls ...*MockCallWrapper) helper method to setup mock calls in order. It implements a similar helper that was recently added and released in to https://github.com/stretchr/testify. 

I had opened a similar PR in the past: https://github.com/temporalio/sdk-go/pull/1623 but was canceled because the helper methods implemented in temporal are similar to the ones offered by "testify", and it was already added to "testify".

## Why?

A more intuitive and cleaner way to define mock calls in order. For example:

```
env.InOrderMockCalls(
	env.OnActivity(namedActivity, mock.Anything, "call1").Return("result1", nil),
	env.OnActivity(namedActivity, mock.Anything, "call2").Return("result2", nil),
	env.OnActivity(namedActivity, mock.Anything, "call3").Return("result3", nil),
)
```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/contrib/datadog/go.mod
+++ b/contrib/datadog/go.mod
@@ -5,7 +5,7 @@ go 1.21
 toolchain go1.21.1
 
 require (
-	github.com/stretchr/testify v1.9.0
+	github.com/stretchr/testify v1.10.0
 	go.temporal.io/sdk v1.25.1
 	gopkg.in/DataDog/dd-trace-go.v1 v1.58.1
 )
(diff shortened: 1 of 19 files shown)


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Looks like your branch isn't up to date and there are merge conflicts, please rebase and then we can run CI, thanks for the contribution

ReyOrtiz:
Ok, branch rebased and merge conflicts resolved. Thanks!

ReyOrtiz:
is there anything I need to do to fix the pipeline failures?

cretz:
You will want to run `go mod tidy` in all directories of the repo that have a `go.mod` file

ReyOrtiz:
Thanks and sorry I missed that. I ran `go mod tidy` on them.

Quinn-With-Two-Ns:
Just waiting on the CLA to be signed, then we can merge

ReyOrtiz:
Ok, I've signed a couple of times. It seems it doesn't refresh automatically

PR #1744, opened 2024-12-05
Versioning 3

antlai-temporal:
## What was changed

This PR implements a new deployment based approach for Worker Versioning in go. 

This is an umbrella PR, with several PRs in the versioning-3 branch already reviewed:

#1675 
#1692 
#1708 
#1722
#1727
#1733
#1742 
 
It is marked as draft, missing some system tests, and waiting on a CLI release that provides the new server.

But even in draft mode, it should be reasonably stable, please give feedback... Thanks!


internal/deployment_client.go
```diff
@@ -0,0 +1,243 @@
...
+		// is an expensive operation, and the server caches results.
+		LastUpdateTime time.Time
+	}
+
+	// DeploymentMetadataUpdate modifies user-defined metadata entries that describe
+	// a deployment.
+	// NOTE: Experimental
+	DeploymentMetadataUpdate struct {
+		// UpsertEntries - Metadata entries inserted or modified. When values are not
+		// of type *commonpb.Payload, the default data converter will be used to generate
```
Quinn-With-Two-Ns, line 166:
```suggestion
		// of type *commonpb.Payload, the clients converter will be used to generate
```

internal/internal_workflow_execution_options.go
```diff
@@ -0,0 +1,149 @@
...
+		RunId string
+		// WorkflowExecutionOptions specifies options for a target workflow execution. Fields not in
+		// UpdatedFields are ignored.
+		WorkflowExecutionOptions WorkflowExecutionOptions
+		// Field names in WorkflowExecutionOptions that will be updated.
+		// When it includes a field name, but the corresponding WorkflowExecutionOptions field has not been set,
+		// it will remove previous overrides for that field.
+		// It panics when it includes a field name not in WorkflowExecutionOptions.
+		// An empty UpdatedFields never modifies WorkflowExecutionOptions.
+		UpdatedFields []string
```
Quinn-With-Two-Ns, line 50: Is there a reason we chose this approach to exclude fields over say making the members of `WorkflowExecutionOptions` pointers and treating `nil` as unset

internal/deployment_client.go
antlai-temporal: Thanks, done

internal/internal_workflow_execution_options.go
antlai-temporal: I'm following closely the proto api that uses a FieldMask. In general, I've been trying to avoid pointers in the api, and be more explicit about what to change, it is sometimes easy to forget that nil means delete otherwise...

Quinn-With-Two-Ns: My suggestion was a `nil` pointer would leave a field unchanged , not clear it.

client/client.go
```diff
@@ -840,6 +948,13 @@ type (
 		// NOTE: Experimental
 		UpdateWorkflow(ctx context.Context, options UpdateWorkflowOptions) (WorkflowUpdateHandle, error)
 
+		// UpdateWorkflowExecutionOptions partially overrides the WorkflowExecutionOptions of an existing workflow execution
+		// and returns the new WorkflowExecutionOptions after applying the changes.
+		// It is intended for building tools that can selectively apply ad-hoc workflow configuration changes.
+		// Use DescribeWorkflowExecution to get the current WorkflowExecutionOptions without modifying them.
+		// NOTE: Experimental
+		UpdateWorkflowExecutionOptions(ctx context.Context, options UpdateWorkflowExecutionOptionsRequest) (WorkflowExecutionOptions, error)
```
Quinn-With-Two-Ns, line 956: I fear this API may get confused with `UpdateWorkflow`, but nothing you can do in this PR as your just copying the naming from the grpc API

internal/schedule_client.go
```diff
@@ -282,6 +282,14 @@ type (
 		// Deprecated - This is only for update of older search attributes. This may be removed in a future version.
 		UntypedSearchAttributes map[string]*commonpb.Payload
 
+		// VersioningOverride - Sets the versioning configuration of a specific workflow execution, ignoring current
+		// server or worker default policies. This enables running canary tests without affecting existing workflows.
+		// To unset the override after the workflow is running, use Client.UpdateWorkflowExecutionOptions.
```
Quinn-With-Two-Ns, line 287: Nit:

```suggestion
		// To unset the override after the workflow is running, use [Client.UpdateWorkflowExecutionOptions].
```

There are a few other godocs that could be links

antlai-temporal: Yes, I missed quite a few, I will made them links

internal/internal_workflow_execution_options.go
antlai-temporal: How do you clear things? Do you use a tombstone?

Quinn-With-Two-Ns: Well here you say 

> // When it includes a field name, but the corresponding WorkflowExecutionOptions field has not been set,
		// it will remove previous overrides for that field.
		
So you are OK relying on the default value to mean clear the field , you can do the same if it takes a pointer.

antlai-temporal: Sure, but I have the other safety net of listing keys that I will delete. I can see people reusing a data structure with the pointers, and making mistakes because they did not clear all of them... @cretz What do you think about a separate selector array vs pointers?

Quinn-With-Two-Ns:
Overall the client API makes sense to me, I didn't review the underlying gRPC api to ensure the API is being invoked or mocked correctly.

internal/worker.go
```diff
@@ -30,6 +30,25 @@ import (
 )
 
 type (
+	// WorkerDeploymentOptions provides configuration for Worker Versioning.
+	// NOTE: Both [WorkerOptions.BuildID] and [WorkerOptions.UseBuildIDForVersioning] need to be set for enabling
```
ShahabT, line 34: Let's discuss these options through on Mon sync before merging to main.

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+
+// DeploymentReachability specifies which category of tasks may reach a worker
+// associated with a deployment, simplifying safe decommission.
+// NOTE: Experimental
+type DeploymentReachability int
+
+const (
+	// DeploymentReachabilityUnspecified - Reachability level not specified.
+	// NOTE: Experimental
+	DeploymentReachabilityUnspecified = iota
```
cretz, line 40: Can you confirm where this value is needed by users?

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+	// NOTE: Experimental
+	DeploymentInfo struct {
+		// Deployment - An identifier for this deployment.
+		Deployment Deployment
+
+		// CreateTime - When this deployment was created.
+		CreateTime time.Time
+
+		// IsCurrent - Whether this deployment is the current one for its deployment series.
+		IsCurrent bool
```
cretz, line 103:
```suggestion
		Current bool
```

Pedantic though, don't need to change

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+		// Next - Returns the next deployment and error
+		Next() (*DeploymentListEntry, error)
+	}
+
+	// DeploymentListOptions are the parameters for configuring listing deployments.
+	// NOTE: Experimental
+	DeploymentListOptions struct {
+		// PageSize - How many results to fetch from the Server at a time.
+		// Optional: defaulted to 1000
+		PageSize int
```
cretz, line 140: Is this needed?

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// DeploymentReachabilityInfo extends [DeploymentInfo] with reachability information.
+	// NOTE: Experimental
+	DeploymentReachabilityInfo struct {
+		// DeploymentInfo - Information about the deployment.
+		DeploymentInfo DeploymentInfo
+
+		// Reachability - Kind of tasks that may reach a worker
+		// associated with a deployment.
+		Reachability DeploymentReachability
```
cretz, line 155: You are inconsistent in this type with whether you want to prefix the field name with `Deployment` when the type is prefixed with `Deployment`. But I can see a case for intentional inconsistency to disambiguate the multiple uses of the term "info".

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+		Deployment Deployment
+
+		// CreateTime - When this deployment was created.
+		CreateTime time.Time
+
+		// IsCurrent - Whether this deployment is the current one for its deployment series.
+		IsCurrent bool
+
+		// TaskQueuesInfo - List of task queues polled by workers in this deployment.
+		TaskQueuesInfo []DeploymentTaskQueueInfo
```
cretz, line 106:
```suggestion
		TaskQueueInfos []DeploymentTaskQueueInfo
```

I think this is clearer, or just `TaskQueues`

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+type (
+	// Deployment identifies a set of workers. This identifier combines
+	// the deployment series name with their Build ID.
+	// NOTE: Experimental
+	Deployment struct {
+		// SeriesName - Name of the deployment series. Different versions of the same worker
+		// service/application are linked together by sharing a series name.
+		SeriesName string
+
+		// BuildID - identifies the worker's code and configuration version.
```
cretz, line 71:
```suggestion
		// BuildID - Identifies the worker's code and configuration version.
```

Pedantic, but you are inconsistent with whether you capitalize the doc after ` - `

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+		// List returns an iterator to enumerate deployments in the client's namespace.
+		// It can also optionally filter deployments by series name.
+		// NOTE: Experimental
+		List(ctx context.Context, options DeploymentListOptions) (DeploymentListIterator, error)
+
+		// GetReachability returns reachability information for a deployment. This operation is
+		// expensive, and results may be cached. Use the returned [DeploymentReachabilityInfo.LastUpdateTime]
+		// to estimate cache staleness.
+		// NOTE: Experimental
+		GetReachability(ctx context.Context, options DeploymentGetReachabilityOptions) (DeploymentReachabilityInfo, error)
```
cretz, line 233: Why would I ever call `Describe` when this has everything describe does and more? Why even have `Describe` or better yet, why not make `Describe` return reachability information?

internal/internal_deployment_client.go
```diff
@@ -0,0 +1,255 @@
...
+
+		// paginate - Function to get the next page of deployment from server.
+		paginate func(nexttoken []byte) (*workflowservice.ListDeploymentsResponse, error)
+	}
+)
+
+func (iter *deploymentListIteratorImpl) HasNext() bool {
+	if iter.err == nil {
+		if iter.response == nil ||
+			(iter.nextDeploymentIndex >= len(iter.response.Deployments) && len(iter.response.NextPageToken) > 0) {
```
cretz, line 61: Combine the conditionals into a single `if`

internal/internal_deployment_client.go
```diff
@@ -0,0 +1,255 @@
...
+			iter.response, iter.err = iter.paginate(iter.response.GetNextPageToken())
+			iter.nextDeploymentIndex = 0
```
cretz, line 63: Can you confirm that the API will never return an empty list with a next page token present? Sometimes some list implementations with filters will apply the filter post-page fetch which means it could filter out all items on the page but still expect another page call.

internal/internal_deployment_client.go
```diff
@@ -0,0 +1,255 @@
...
+		return dc.workflowClient.workflowService.ListDeployments(grpcCtx, request)
+	}
+
+	return &deploymentListIteratorImpl{
+		paginate: paginate,
+	}, nil
+}
+
+func (dc *deploymentClient) Describe(ctx context.Context, options DeploymentDescribeOptions) (DeploymentInfo, error) {
+	if err := dc.workflowClient.ensureInitialized(ctx); err != nil {
```
cretz, line 177: Should add this to the top of `List` too

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+		// SeriesName - Name of the deployment series.
+		SeriesName string
+	}
+
+	// DeploymentClient is the client that manages deployments.
+	// NOTE: Experimental
+	DeploymentClient interface {
+		// Describes an existing deployment.
+		// NOTE: Experimental
+		Describe(ctx context.Context, options DeploymentDescribeOptions) (DeploymentInfo, error)
```
cretz, line 222: How confident are you that we will never add anything else to `DescribeDeploymentResponse`?

internal/deployment_client.go
```diff
@@ -0,0 +1,244 @@
...
+
+		// GetReachability returns reachability information for a deployment. This operation is
+		// expensive, and results may be cached. Use the returned [DeploymentReachabilityInfo.LastUpdateTime]
+		// to estimate cache staleness.
+		// NOTE: Experimental
+		GetReachability(ctx context.Context, options DeploymentGetReachabilityOptions) (DeploymentReachabilityInfo, error)
+
+		// GetCurrent returns the current deployment for a given deployment series.
+		// NOTE: Experimental
+		GetCurrent(ctx context.Context, options DeploymentGetCurrentOptions) (DeploymentInfo, error)
```
cretz, line 237: How confident are you that we will never add anything else to `GetCurrentDeploymentResponse`?

internal/internal_workflow_execution_options.go
cretz: We need to decide now how we're going to to do partial updates because `UpdateActivityOptions` (or `UpdateActivityOptionsById`) is coming with the same problem. I would do partial updates in the way the language handles it best. In Go unfortunately there is not one way, but this `UpdatedFields` is probably the least common and IMO we should not expose it.

I would either recommend using pointers or making an update options builder. Pointers works fine.

> I can see people reusing a data structure with the pointers, and making mistakes because they did not clear all of them

We should not use existing data structures in a partial update model because of this concern. So every data type for updating should not reuse the non-updating form on partial update things (this is not a full PUT-with-conflict-token type of API). So either inline the options update fields here or make the data type update-options specific.

.github/workflows/ci.yml
```diff
@@ -88,6 +88,9 @@ jobs:
       - name: Docker compose - integration tests
         if: ${{ matrix.testDockerCompose }}
         run: go run . integration-test
+        env:
+          # TODO(antlai-temporal): Remove this flag once server 1.26.2 released.
+          DISABLE_DEPLOYMENT_TESTS: "1"
```
cretz, line 93: We should not merge unless/until we can remove this flag IMO. If that means we need an RC release of CLI we can, though hopefully CLI for 1.26 is released before this so users of this SDK can test too.

antlai-temporal: We will have an RC for CLI and run  integration test in CI before merging, see: 
https://github.com/temporalio/sdk-go/blob/7e976b2a877e1fd95c98750a8a0e955989e0fd4f/internal/cmd/build/main.go#L143-L145

This flag only disables the tests for the docker image, which unfortunately only gets built for real releases, but the integration tests do run in against that CLI with -dev-server:
https://github.com/temporalio/sdk-go/blob/7e976b2a877e1fd95c98750a8a0e955989e0fd4f/.github/workflows/ci.yml#L47-L59

internal/deployment_client.go
antlai-temporal: I think fairly confident, that API just returns info on a particular deployment, and we control the struct `DeploymentInfo`, since it is only used for `Describe` and `GetCurrent`, so we could extend it if needed...

internal/deployment_client.go
antlai-temporal: Same argument as above, I wanted it to be symmetrical with `Describe`,  I think we can extend the struct `DeploymentInfo` without needing a wrapper...

internal/deployment_client.go
antlai-temporal: It is an optimization to reduce the number of roundtrips vs memory available. Frankly, doubt many people will use it, but it was useful at least for testing (forcing several roundtrips...).

client/client.go
```diff
@@ -840,6 +948,13 @@ type (
 		// NOTE: Experimental
 		UpdateWorkflow(ctx context.Context, options UpdateWorkflowOptions) (WorkflowUpdateHandle, error)
 
+		// UpdateWorkflowExecutionOptions partially overrides the [WorkflowExecutionOptions] of an existing workflow execution
+		// and returns the new [WorkflowExecutionOptions] after applying the changes.
+		// It is intended for building tools that can selectively apply ad-hoc workflow configuration changes.
+		// Use [DescribeWorkflowExecution] to get the current [WorkflowExecutionOptions] without modifying them.
```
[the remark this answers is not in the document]
antlai-temporal: Good point, the same info is available in the response protos but it does not return that type...

client/client.go
```diff
@@ -840,6 +948,13 @@ type (
...
+		// NOTE: Experimental
+		UpdateWorkflowExecutionOptions(ctx context.Context, options UpdateWorkflowExecutionOptionsRequest) (WorkflowExecutionOptions, error)
```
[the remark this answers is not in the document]
antlai-temporal: Yes, I agree is a bit confusing, I'm just trying to be consistent with the API that would have been better with `*Properties`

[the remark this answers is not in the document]
antlai-temporal: Yes, I also think that "update" is too tainted with our signal/update/... and it is not always clear it is a verb coming from CRUD, and not a noun (or adjective) from our updates...

internal/deployment_client.go
antlai-temporal: That value should always come from the server. Reachability is expensive and it is only run every so often, so it could be the server not having a cached value, and deciding not to run it either... I'll ask the server folks...

internal/deployment_client.go
antlai-temporal: Thanks, fixed.

internal/deployment_client.go
antlai-temporal: It was `Current`, @ShahabT asked to change it to `IsCurrent`, it is a bool so it should be clear, but I don't have strong opinions...

internal/deployment_client.go
antlai-temporal: Thanks, much better plural.

internal/deployment_client.go
antlai-temporal: Yes, that was my thinking, just "Info" was too overloaded...

internal/deployment_client.go
antlai-temporal: Getting Reachability info is a very expensive operation, so it may be stale and/or rate limited, and that's why they provided a "light" version with `Describe` that only returns the easy bits... I will add a comment.

internal/internal_deployment_client.go
antlai-temporal: Done

internal/internal_deployment_client.go
antlai-temporal: Good point, thanks. I added it inside `paginate()`, `List` just returns the iterator struct...

internal/internal_workflow_execution_options.go
antlai-temporal: I'm rewriting it with pointers. We only allow "top level object" partial updates, not arbitrary zipping, so I think that as long as the top level container is a different type we should be ok blocking reuse...

client/client.go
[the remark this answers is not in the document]
cretz: The problem is we have `UpdateActivityOptions` coming (or `UpdateActivityOptionsById`). But I still think what's here is probably the best.

internal/deployment_client.go
cretz: It is a bit strange to use `Is` prefixes in Go fields for these kinds of values IMO (unlike Java where of course you will)

internal/deployment_client.go
cretz: Hrmm, wonder if we should remove it. Will defer to others' opinions here.

internal/deployment_client.go
cretz: Hrmm, to me it sounds like overlapping concerns and to obtain reachability should just be an option on describe. I assume this was discussed at the API level and decided to separate `DescribeDeployment` and `GetDeploymentReachability` instead of just make this an option on `Describe` (like we have on `DescribeTaskQueue`)? If it was discussed, can you share the reasoning there?

internal/internal_deployment_client.go
cretz: Right, I think for schedules it was confirmed. If we can just get server team confirmation that no matter what is in `ListDeploymentsRequest` (e.g. `series_name`) it is impossible to get an empty page with a next page token we're good.

internal/deployment_client.go
cretz: For instance, like with `DescribeTaskQueue`, what if we decided to add the ability to request more things (which arguably should have been done with reachability information)? IMO in a perfect world `Describe` would return a `DeploymentDescription` and that description would have info and (opt-in) reachability info today and can grow to whatever else in the future.

internal/deployment_client.go
cretz: We often add things to the gRPC response message and not the single info field within. We have done this before on other calls and regret reusing types as the results on these methods because we are forced to make more.

.github/workflows/ci.yml
cretz: Works for me. I wonder it the 1.26 release (and subsequent CLI release) will come before the RC.

internal/internal_workflow_client.go
```diff
@@ -1057,6 +1057,37 @@ func (wc *WorkflowClient) GetWorkerTaskReachability(ctx context.Context, options
...
+// NOTE: Experimental
+func (wc *WorkflowClient) UpdateWorkflowExecutionOptions(ctx context.Context, request UpdateWorkflowExecutionOptionsRequest) (WorkflowExecutionOptions, error) {
+	if err := wc.ensureInitialized(ctx); err != nil {
+		return WorkflowExecutionOptions{}, err
+	}
+
+	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
+	defer cancel()
+
+	workflowExecutionOptions, updateMask := workflowExecutionOptionsChangesToProto(request.WorkflowExecutionOptionsChanges)
```
cretz, line 1072: I think we should fail if there are no changes and/or a missing workflow ID (general validation like this applies to other calls too of course)

internal/deployment_client.go
antlai-temporal: Yes, it was discussed, they can only do rate limiting at the method level, and that's why they want a separate method.

internal/deployment_client.go
antlai-temporal: If we wanted we could make reachability an optional field in DeploymentInfo, I think extending is not a problem. Note that in versioning-1 there was a separate method for GetReachability, they combine it with DescribeTaskQueue in versioning-2 and that created many problems because that was a popular method, we wanted to abstract away from TQ,  and they could not rate limit cleanly, so we are back to the versioning-1 approach...
But I see your point that reachability is just info about the deployment, maybe we could group both with a flag in the cli?

Alternatively, I could wrap this in the standard "*Response" boring proto and kick the can down the road...

internal/deployment_client.go
antlai-temporal: OK, wrapping the responses

internal/internal_workflow_client.go
antlai-temporal: Adding more validation

internal/deployment_client.go
antlai-temporal: We have it in ScheduleListOptions, I think we should be consistent...

internal/deployment_client.go
cretz: Hrmm, that is unfortunate to burden a user API all the way out through the SDK with our server limitations, but ok. I am guessing they decided it was ok not to need to separately rate limit the different `DescribeTaskQueueRequest` options.

internal/deployment_client.go
cretz: I was thinking this would return a `DeploymentDescription` that would have `Info` as its only field today. That way it can grow if the gRPC API does. If we wanted to though, we could combine this call with `GetReachability` and then `DeploymentDescription` would have two fields, but I understand from the other thread it was decided elsewhere to keep them separate unfortunately because of some internal server limitation.

antlai-temporal: OK, adding a DeploymentDescription and also wrapping the "*Response"

internal/deployment_client.go
cretz: We don't for `GetWorkflowHistory` which returns an iterator (even though we could), so we're going to be inconsistent regardless. But I am ok with leaving it.

internal/internal_deployment_client.go
```diff
@@ -0,0 +1,313 @@
...
+
+		// paginate - Function to get the next page of deployment from server.
+		paginate func(nexttoken []byte) (*workflowservice.ListDeploymentsResponse, error)
+	}
+)
+
+func (iter *deploymentListIteratorImpl) HasNext() bool {
+	if iter.err == nil &&
+		(iter.response == nil ||
+			(iter.nextDeploymentIndex >= len(iter.response.Deployments) && len(iter.response.NextPageToken) > 0)) {
```
carlydf, line 62: `ListDeployments` uses `visibility.ListWorkflowExecutions` under the hood directly, the same way that `ListSchedules` does, so the contents of a page and the `nextPageToken` are handled the same for `ListDeployments` and `ListSchedules`. Because of this, I believe that there is no risk of an empty page followed by a non-empty page.

carlydf: In the elastic search visibility store implementation, next page token is only non-nil if the number of entries on the page == pageSize: https://github.com/temporalio/temporal/blob/main/common/persistence/visibility/store/elasticsearch/visibility_store.go#L839-L848

carlydf: @cretz @antlai-temporal hope this helps

carlydf: same in the sql visibility store implementation, next page token is only non-nil if the number of entries on the page == pageSize: https://github.com/temporalio/temporal/blob/main/common/persistence/visibility/store/sql/visibility_store.go#L233-L248

cretz: > I believe that there is no risk of an empty page followed by a non-empty page

Sounds good. Just wanted to make sure there was no post-visibility item filtering done server side that would affect this.

cretz (commented):
Looks good, just waiting for CLI to be updated properly

antlai-temporal:
@cretz Integration tests against the new cli ok...

cretz (approved):

PR #1745, opened 2024-12-05
Tweak UpdateWorfklow docs

dandavison:
Some proposed edits to the `UpdateWorkflow` docstring. These make it consistent with the `UpdateWithStartWorkflow` docstring, and also make some other edits that I think are slight improvements, but open for discussion.


client/client.go
```diff
@@ -826,14 +826,16 @@ type (
...
-		// from the server will be exposed through the return value of
-		// WorkflowUpdateHandle.Get(). Errors that occur before the
-		// update is requested (e.g. if the required workflow ID field is
-		// missing from the UpdateWorkflowOptions) are returned
-		// directly from this function call.
+		// UpdateWorkflow issues an update request to the specified workflow and
+		// returns a handle to the update. The call will block until the update
+		// has reached the WaitForStage in the options. Note that this means
+		// that the call will not return successfully until the update has been
+		// delivered to the worker service. Errors returned from the update
```
Quinn-With-Two-Ns, line 833:
```suggestion
		// delivered to a worker. Errors returned from the update
```

I would avoid the word worker service here since it has a different meaning in a Temporal cluster https://docs.temporal.io/clusters. Here we mean the user worker.

dandavison: Thanks, I agree "a worker" is better than "worker service". (I don't think I agree that "worker service" should in general be reserved as referring to the internal server implementation detail though)

Quinn-With-Two-Ns (approved):

PR #1746, opened 2024-12-06
Tweak Update-with-Start error reporting

stephanos:
## What was changed

Tweak error reporting of Update-with-Start.

## Why?

Communicate errors better to user.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -1761,7 +1761,7 @@ func (w *workflowClientInterceptor) UpdateWithStartWorkflow(
 	// Create update request
 	updateReq, err := w.createUpdateWorkflowRequest(ctx, updateInput)
 	if err != nil {
-		return nil, fmt.Errorf("%w: %w", errInvalidWithStartWorkflowOperation, err)
+		return nil, err
 	}
 	if updateReq.WorkflowExecution.WorkflowId == "" {
 		updateReq.WorkflowExecution.WorkflowId = startReq.WorkflowId
@@ -1800,7 +1800,7 @@ func (w *workflowClientInterceptor) UpdateWithStartWorkflow(
 	}
 	handle, err := w.updateHandleFromResponse(ctx, updateReq.WaitPolicy.LifecycleStage, updateResp)
 	if err != nil {
-		return nil, fmt.Errorf("%w: %w", errInvalidWithStartWorkflowOperation, err)
+		return nil, err
 	}
 	return handle, nil
 }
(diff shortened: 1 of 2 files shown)


internal/internal_workflow_client.go
```diff
@@ -1800,7 +1800,7 @@ func (w *workflowClientInterceptor) UpdateWithStartWorkflow(
 	}
 	handle, err := w.updateHandleFromResponse(ctx, updateReq.WaitPolicy.LifecycleStage, updateResp)
 	if err != nil {
-		return nil, fmt.Errorf("%w: %w", errInvalidWithStartWorkflowOperation, err)
+		return nil, err
```
stephanos, line 1803: Wrapping with `errInvalidWithStartWorkflowOperation` ("invalid WithStartWorkflowOperation") is inappropriate IMO, since it's not related to the start operation.

internal/internal_workflow_client.go
```diff
@@ -1862,27 +1862,31 @@ func (w *workflowClientInterceptor) updateWithStartWorkflow(
...
 					}
 				default:
 					// this would only happen if a case statement for a newly added operation is missing above
 					return nil, fmt.Errorf("%w: %T", errUnsupportedOperation, t)
 				}
 			}
-			return nil, startErr
+
+			// this should never happen
+			return nil, errors.New(multiErr.Error())
```
stephanos, line 1889: This would only happen if the server sent back `[nil, nil]`; which would be a bug. But in case it does, this is the fallback.

internal/internal_workflow_client_test.go
```diff
@@ -1167,32 +1217,6 @@ func (s *workflowRunSuite) TestExecuteWorkflowWithUpdate_ServerResponseCountMism
 	s.ErrorContains(err, "invalid server response: 0 instead of 2 operation results")
 }
 
-func (s *workflowRunSuite) TestExecuteWorkflowWithUpdate_ServerErrorResponseCountMismatch() {
```
stephanos, line 1170: Moved into the above.

internal/internal_workflow_client.go
```diff
@@ -1862,27 +1862,31 @@ func (w *workflowClientInterceptor) updateWithStartWorkflow(
...
 
 				switch t := opReq.Operation.(type) {
 				case *workflowservice.ExecuteMultiOperationRequest_Operation_StartWorkflow:
 					if !errors.As(opErr, &abortedErr) {
-						startErr = opErr
+						return nil, fmt.Errorf("failed workflow start: %w", opErr)
 					}
 				case *workflowservice.ExecuteMultiOperationRequest_Operation_UpdateWorkflow:
 					if !errors.As(opErr, &abortedErr) {
-						startErr = fmt.Errorf("%w: %w", errInvalidWithStartWorkflowOperation, opErr)
```
stephanos, line 1878: "invalid WithStartWorkflowOperation" isn't quite accurate anyway. It could have just "failed" despite the operation itself being valid.

internal/internal_workflow_client.go
dandavison: Should we be honest and return a message like "This should never happen, please report a bug at ..."?

dandavison (approved):
LGTM, thank you!

internal/internal_workflow_client.go
stephanos: Oh, I only saw this comment now. I'm not sure; is that something we do?

This scenario would mean the server decided to create a MultiOps error, but messed up the inner errors. So the intention is clear; but there aren't enough details sent back to the user. Would it be reasonable to assume that a user will be confused and ask for better errors?

PR #1749, opened 2024-12-09
GHA to check api-go compatibility

rodrigozhou:
## What was changed

GHA to check compatibility of api-go

## Why?

Automation

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/check-api-go.yml
+++ b/.github/workflows/check-api-go.yml
@@ -3,12 +3,22 @@ name: Check api-go compatibility
 on:
   workflow_dispatch:
     inputs:
-      sdk_commit:
-        description: sdk-go commit to check
+      sdk_ref:
+        description: sdk-go ref to check ("latest" for latest release tag)
+        required: true
+        default: latest
+      api_ref:
+        description: api-go ref to check
         required: true
         default: master
-      api_commit:
-        description: api-go commit to check
+  workflow_call:
+    inputs:
+      sdk_ref:
+        description: sdk-go ref to check ("latest" for latest release tag)
+        required: true
+        default: latest
+      api_ref:
+        description: api-go ref to check
         required: true
         default: master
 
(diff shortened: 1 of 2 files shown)


.github/workflows/check-api-go.yml
```diff
@@ -0,0 +1,68 @@
+name: Check api-go compatibility
+
+on:
+  workflow_dispatch:
+    inputs:
+      sdk_commit:
+        description: sdk-go commit to check
+        required: true
+        default: master
```
cretz, line 9: This needs to be the latest tag. The goal is to make sure the latest tag of SDK Go works with the latest (coming) tag of API Go.

cretz (commented):
Thanks! Note, when making the `api` side, you may want the ability to bypass. The reason one might want to bypass is because we know the API is incompatible and we plan on tagging a new Go SDK release with the new API release immediately after API release is complete.

.github/workflows/check-api-go.yml
cretz, line 1: Question, can we just put this whole workflow inside of `api-go`? Is there any reason it needs to be inside the SDK repo?

cretz (commented):
LGTM, but a general question about moving this to `api-go` as there seems to be no benefit to it being in the SDK.

rodrigozhou:
@cretz Sure, I had initially put in the `api` repo because it's where the release process starts. Any particular reason to put in `api-go` instead of `api`?

cretz:
> Any particular reason to put in api-go instead of api?

Yes, this check is very Go specific. I know the API repo has some Go things, but in theory the API repo is supposed to be language agnostic. Having said that, if it is easier to fail the release from `api` instead of `api-go`, no problem.

rodrigozhou:
Makes sense, moved to `api-go` repo: https://github.com/temporalio/api-go/pull/200

PR #1751, opened 2024-12-09
Nexus error rehydration

bergundy:
## What was changed

Add support for full Temporal error rehydration.
Any Temporal errors returned over the Nexus boundary will now automatically be fully rehydrated on the caller side.

Some notable points:
- The Nexus SDK was upgraded to `v0.1.0`, which is a breaking change. Temporal Nexus users should refer to the [release notes](https://github.com/nexus-rpc/sdk-go/releases/tag/v0.1.0) before upgrading the Temporal SDK.
- `nexus.HandlerError` now has a `Failure` representation and is visible to the caller workflow.
- Arbitrary errors are translated to `ApplicationError` as they do in the rest of the Temporal SDK.
- `WorkflowExecutionError`, `QueryRejectedError`, and non retryable `ApplicationError`s are no longer translated to bad request handler errors, instead they're translated to `OperationError`s.

## Why?

Provide a more consistent experience for Temporal users and on par debugging experience with workflows and activities.

## For reviewers

I've added the tests here to illustrate the E2E behavior but I'll have to wait for a server release that supports the new behavior before merging those tests. I may commit the tests in a separate PR just to get this code in or do a server RC for the full test coverage.


test/go.mod
```diff
@@ -67,6 +67,7 @@ require (
 )
 
 replace (
+	github.com/nexus-rpc/sdk-go => ../../nexus-sdk-go
```
bergundy, line 70: Missed this

internal/internal_nexus_task_handler.go
```diff
@@ -200,51 +191,62 @@ func (h *nexusTaskHandler) handleStartOperation(
...
 	if ctx.Err() != nil {
+		if !panic {
+			nctx.Log.Error("Context error while processing Nexus task", "error", ctx.Err())
+		}
 		return nil, nil, errNexusTaskTimeout
 	}
 	if err != nil {
+		if !panic {
+			nctx.Log.Error("Handler returned error while processing Nexus task", "error", err)
+		}
```
cretz, line 220:
```suggestion
		if !panic {
			nctx.Log.Error("Context error while processing Nexus task", tagError, ctx.Err())
		}
		return nil, nil, errNexusTaskTimeout
	}
	if err != nil {
		if !panic {
			nctx.Log.Error("Handler returned error while processing Nexus task", tagError, err)
		}
```

Or is it intentionally not using that tag?

internal/internal_nexus_task_handler.go
```diff
@@ -200,51 +191,62 @@ func (h *nexusTaskHandler) handleStartOperation(
...
-				nctx.Log.Error("Panic captured while handling nexus task", tagStackTrace, string(debug.Stack()), tagError, err)
+				nctx.Log.Error("Panic captured while handling Nexus task", tagStackTrace, string(debug.Stack()), tagError, err)
 			}
 		}()
 		opres, err = h.nexusHandler.StartOperation(ctx, req.GetService(), req.GetOperation(), input, startOptions)
 	}()
 	if ctx.Err() != nil {
+		if !panic {
+			nctx.Log.Error("Context error while processing Nexus task", "error", ctx.Err())
+		}
```
cretz, line 214: Arguably this should be logged regardless

internal/internal_nexus_task_handler.go
cretz, line 220: Arguably this log could just be after the actual call in its own `if err != nil` block, but not a big deal to have the bool

cretz (approved):
Nothing blocking, especially since Nexus is not yet GA

internal/internal_nexus_task_handler.go
bergundy: Yeah, I was debating but figured it'd be nicer to avoid the spam.

internal/internal_nexus_task_handler.go
bergundy: err would not be nil here, it's filled in from the panic handler.

test/nexus_test.go
```diff
@@ -427,16 +422,19 @@ func TestSyncOperationFromWorkflow(t *testing.T) {
 		switch outcome {
 		case "successful":
 			return outcome, nil
-		case "failed":
-			return "", &nexus.UnsuccessfulOperationError{
-				State:   nexus.OperationStateFailed,
-				Failure: nexus.Failure{Message: "failed for test"},
+		case "failed-plain-error":
```
Quinn-With-Two-Ns, line 425: What are these cases for? They don't seem to be used by any test

test/nexus_test.go
```diff
@@ -427,16 +422,19 @@ func TestSyncOperationFromWorkflow(t *testing.T) {
...
+			return "", nexus.NewFailedOperationError(errors.New("failed for test"))
+		case "failed-app-error":
```
Quinn-With-Two-Ns, line 427: What are these cases for? They don't appear to be called by any test

bergundy: I left these and didn't commit the tests since we don't have server support yet. The tests are in a branch on my machine. I can start a draft PR that we'll merge once CLI with 1.26.2 is out.

PR #1754, opened 2024-12-12
Remove "Experimental" notices on Update APIs

dandavison:


cretz (approved):
Saw this also present for Java. I assume this is coming for the TS, Python, and .NET too?

dandavison:
> Saw this also present for Java. I assume this is coming for the TS, Python, and .NET too?

Yes, we'll do the same for them before their next releases.

PR #1755, opened 2024-12-12
Use latest CLI for dev server in tests

rodrigozhou:
## What was changed

Use latest CLI for dev server in tests

## Why?

Pinned CLI version might be outdated, and fail tests.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/cmd/build/main.go
+++ b/internal/cmd/build/main.go
@@ -144,9 +144,6 @@ func (b *builder) integrationTest() error {
 				HostPort:  "127.0.0.1:7233",
 				Namespace: "integration-test-namespace",
 			},
-			CachedDownload: testsuite.CachedDownload{
-				Version: "v1.1.0",
-			},
 			LogLevel: "warn",
 			ExtraArgs: []string{
 				"--dynamic-config-value", "frontend.enableExecuteMultiOperation=true",


internal/cmd/build/main.go
```diff
@@ -144,9 +144,6 @@ func (b *builder) integrationTest() error {
 				HostPort:  "127.0.0.1:7233",
 				Namespace: "integration-test-namespace",
 			},
-			CachedDownload: testsuite.CachedDownload{
-				Version: "v1.1.0",
```
cretz, line 148: Arguably this should just be at the fixed version to pass the test. If a test is added in a newer version, then the CLI needs to be upgraded. This is especially true because for some projects they will set this to an RC version that won't come back on cached download.

I guess we _technically_ can keep it on latest-stable always (we probably did before), but we shouldn't need to and I fear the need to may be hiding some other issue. Can you give some more reasoning behind why it can't be a fixed value?

rodrigozhou: So, the problem I see is that sdk-go is already in v1.31.0, but CLI v1.1.0 is using sdk-go v1.29.0. So, when the sdk-go CICD is running, it's using an already outdated version of CLI which could fail due to staleness.

cretz: > the problem I see is that sdk-go is already in v1.31.0, but CLI v1.1.0 is using sdk-go v1.29.0

The SDK that the CLI happens to be using is wholly unrelated to the SDK that the user or integration tests are using. It is very normal for instance for our users to use an old CLI with newer SDK versions. If the CLI fails due to staleness, that shows up here in CI at which point it can be upgraded as needed.

PR #1762, opened 2025-01-07
Enable SDKFlagBlockedSelectorSignalReceive by default

yuandrew:
## What was changed

Flip the switch to enable this fixed behavior by default. Tests have been updated to reflect this change in default.

## Why?

This is a follow up to #1682, now that there has been a release since introducing this flag, it should be safe to flip the switch 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/internal_flags.go
```diff
@@ -53,7 +53,7 @@ const (
 	SDKFlagUnknown                      = math.MaxUint32
 )
 
-var unblockSelectorSignal bool
+var unblockSelectorSignal = true
```
Quinn-With-Two-Ns, line 56: Missed this in the last PR, can we add a commend explaining what this does flag does and why it exists? I can help you on the wording if you need.

yuandrew: yes! good idea

Quinn-With-Two-Ns (approved):

Issue #1764, opened 2025-01-08
Versioning override AutoUpgrade using a ref to an empty deployment struct instead of a nil pointer


antlai-temporal:
The new versioning override API (`UpdateWorkflowExecutionOptions`) requires a Deployment proto field with a pointer to nil to set 
AutoUpgrade behavior, and instead,  the SDK provides a pointer to an empty Deployment struct.

The result is that the server always fails the request.

Quinn-With-Two-Ns:
@antlai-temporal Can we close this issue?

antlai-temporal:
Sure, closing...

PR #1765, opened 2025-01-08
Fix versioning override with AutoUpgrade behavior

antlai-temporal:
## What was changed

The current implementation for versioning override fails for AutoUpgrade
because it uses an empty Deployment struct instead of a nil pointer.

See #1764 for details.

Also missing `gofmt` of some unrelated files corrected.

## Why?

This is a blocker for the new versioning-3 implementation.

## Checklist

1. Closes 
#1764 
2. How was this tested:

Integration test

### DIFF
--- a/internal/internal_workflow_execution_options.go
+++ b/internal/internal_workflow_execution_options.go
@@ -105,6 +105,12 @@ func workflowExecutionOptionsMaskToProto(mask []string) *fieldmaskpb.FieldMask {
 }
 
 func workerDeploymentToProto(d Deployment) *deploymentpb.Deployment {
+	// Server 1.26.2 requires a nil Deployment pointer, and not just a pointer to an empty Deployment,
+	// to indicate that there is no Deployment.
+	// It is a server error to override versioning behavior to AutoUpgrade while providing a Deployment,
+	// and we need to replace it by nil. See https://github.com/temporalio/sdk-go/issues/1764.
+	//
+	// Future server versions may relax this requirement.
 	if (Deployment{}) == d {
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


internal/internal_workflow_execution_options.go
```diff
@@ -105,6 +105,9 @@ func workflowExecutionOptionsMaskToProto(mask []string) *fieldmaskpb.FieldMask {
 }
 
 func workerDeploymentToProto(d Deployment) *deploymentpb.Deployment {
```
cretz, line 107: Why not make the source of this a pointer if unset is a valid state? Meaning, make `VersioningOverride.Deployment` accept a pointer.

antlai-temporal: I thought it was cleaner not to expose pointers and just use an empty struct. Otherwise you ended up with what "a pointer to an empty struct" means vs a nil pointer. It is also consistent with the other apis that use Deployment.

cretz: > Otherwise you ended up with what "a pointer to an empty struct" means vs a nil pointer.

If there is a difference between empty and null in the server-side API and we need to represent that difference, we should do the same here. If the server errors on empty, good, that tells the user it was invalid like it might for anything else invalid.

> It is also consistent with the other apis that use Deployment.

Are there other fields where the zero value of a struct means nil when converting to proto? I think we should consider fixing those too.

antlai-temporal: Yes, other empty structs related to deployment are ok, but this one was not... They may eventually fix that, but this is 
blocking the autoUpgrade override...

antlai-temporal: @cretz There was a difference but it was not intentional, there should be no difference. It is just that the OSS server release is out, and it is better to fix it internally in the first SDK release without exposing it, and eventually change it in the server...
All the other values that I can think of would take either empty or nil. I mapped  VersioningOverride{} to nil mostly to keep things more clear, but I don't think it is strictly needed.

antlai-temporal: Comment added

Quinn-With-Two-Ns (approved):

PR #1766, opened 2025-01-08
Resolve workflow name alias in Nexus WorkflowRunOperation

bergundy:
This is similar to how child workflow name resolution works. Without this, we don't properly support alias based registration.

### DIFF
--- a/internal/internal_nexus_task_handler.go
+++ b/internal/internal_nexus_task_handler.go
@@ -71,6 +71,7 @@ type nexusTaskHandler struct {
 	failureConverter converter.FailureConverter
 	logger           log.Logger
 	metricsHandler   metrics.Handler
+	registry         *registry
 }
 
 func newNexusTaskHandler(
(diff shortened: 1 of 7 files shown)


rodrigozhou (approved):

internal/internal_nexus_task_handler.go
```diff
@@ -393,6 +396,9 @@ func (h *nexusTaskHandler) newNexusOperationContext(response *workflowservice.Po
 		TaskQueue:      h.taskQueueName,
 		MetricsHandler: metricsHandler,
 		Log:            logger,
+		ResolveWorkflowName: func(a any) (string, error) {
```
cretz, line 399: This would be clearer as a method instead of a lambda property IMO (if you need to store the registry on the context on a non-exported field, no prob)

Quinn-With-Two-Ns (approved):

cretz (approved):

PR #1769, opened 2025-01-09
small docs/rendering fixes

GSmithApps:
small docs/rendering fixes

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -246,7 +246,8 @@ func (bp *basePoller) stopping() bool {
 }
 
 // doPoll runs the given pollFunc in a separate go routine. Returns when either of the conditions are met:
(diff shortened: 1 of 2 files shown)


internal/internal_task_pollers.go
```diff
@@ -246,7 +246,8 @@ func (bp *basePoller) stopping() bool {
 }
 
 // doPoll runs the given pollFunc in a separate go routine. Returns when either of the conditions are met:
-// - poll succeeds, poll fails or worker is stopping
+//  - poll succeeds
```
Quinn-With-Two-Ns, line 249: This should be a list of three if we really want to break it into a list

GSmithApps: Fixed. The "either" made it sound like there were two instead of three

Quinn-With-Two-Ns (approved):

PR #1770, opened 2025-01-10
Add Nexus SignalWorkflowOperation

pdoerner:
## What was changed

Added a new type of synchronous Nexus operation for sending signals.

## Why?

Convenience and request ID + links handling.

## Checklist

1. Closes 

2. How was this tested:

New functional tests

3. Any docs updates needed?

### DIFF
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -518,6 +518,8 @@ type ClientExecuteWorkflowInput struct {
 // Exposed as: [go.temporal.io/sdk/interceptor.ClientSignalWorkflowInput]
 //
 // Exposed as: [go.temporal.io/sdk/interceptor.ClientSignalWorkflowInput]
+//
+// Exposed as: [go.temporal.io/sdk/temporalnexus.SignalWorkflowInput]
 type ClientSignalWorkflowInput struct {
 	WorkflowID string
 	RunID      string
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -41,6 +41,8 @@ import (
 // Exposed as: [go.temporal.io/sdk/interceptor.Interceptor]
 //
 // Exposed as: [go.temporal.io/sdk/interceptor.Interceptor]
+//
+// Exposed as: [go.temporal.io/sdk/interceptor.Interceptor]
 type Interceptor interface {
 	ClientInterceptor
 	WorkerInterceptor
@@ -48,6 +50,8 @@ type Interceptor interface {
 
 // WorkerInterceptor is a common interface for all interceptors. See
 // documentation in the interceptor package for more details.
+//
+// Exposed as: [go.temporal.io/sdk/interceptor.WorkerInterceptor]
 //
 // Exposed as: [go.temporal.io/sdk/interceptor.WorkerInterceptor]
 //
@@ -71,6 +75,8 @@ type WorkerInterceptor interface {
 // Exposed as: [go.temporal.io/sdk/interceptor.ActivityInboundInterceptor]
 //
 // Exposed as: [go.temporal.io/sdk/interceptor.ActivityInboundInterceptor]
+//
+// Exposed as: [go.temporal.io/sdk/interceptor.ActivityInboundInterceptor]
 type ActivityInboundInterceptor interface {
 	// Init is the first call of this interceptor. Implementations can change/wrap
 	// the outbound interceptor before calling Init on the next interceptor.
@@ -88,6 +94,8 @@ type ActivityInboundInterceptor interface {
 // Exposed as: [go.temporal.io/sdk/interceptor.ExecuteActivityInput]
 //
 // Exposed as: [go.temporal.io/sdk/interceptor.ExecuteActivityInput]
+//
+// Exposed as: [go.temporal.io/sdk/interceptor.ExecuteActivityInput]
 type ExecuteActivityInput struct {
 	Args []interface{}
 }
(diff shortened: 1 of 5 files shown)


temporalnexus/operation.go
```diff
@@ -94,6 +93,20 @@ func NewSyncOperation[I any, O any](
...
+//
+// NOTE: Experimental
+func NewSignalWorkflowOperation[I any](
+	name string,
+	workflowID string,
+	runID string,
+	signalName string,
+) nexus.Operation[I, nexus.NoValue] {
+	return NewSyncOperation(name, func(ctx context.Context, c client.Client, i I, options nexus.StartOperationOptions) (nexus.NoValue, error) {
+		return nil, c.SignalWorkflow(ctx, workflowID, runID, signalName, i)
```
Quinn-With-Two-Ns, line 106: Don't we need to set the request ID here as well for idempotency

internal/internal_workflow_client.go
```diff
@@ -1972,9 +1972,10 @@ func (w *workflowClientInterceptor) SignalWorkflow(ctx context.Context, in *Clie
 		return err
 	}
 
+	links, _ := ctx.Value(NexusOperationRequestIDKey).([]*commonpb.Link)
```
bergundy, line 1975: This should be a link key, not the request ID key.

internal/interceptor.go
```diff
@@ -41,6 +41,8 @@ import (
 // Exposed as: [go.temporal.io/sdk/interceptor.Interceptor]
 //
 // Exposed as: [go.temporal.io/sdk/interceptor.Interceptor]
+//
+// Exposed as: [go.temporal.io/sdk/interceptor.Interceptor]
```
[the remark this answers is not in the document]
pdoerner: I originally aliased `temporalnexus.SignalWorkflowInput` to this type and added it to the comment but then undid that and aliased it to the `internal` type. The automated tools complained about the checks until I ran with `-fix` which I guess just added another line again.

temporalnexus/operation.go
```diff
@@ -94,6 +93,28 @@ func NewSyncOperation[I any, O any](
 	}
 }
 
+// SignalWorkflowInput is the input to a NewSignalWorkflowOperation.
+type SignalWorkflowInput = internal.ClientSignalWorkflowInput
+
+// NewSignalWorkflowOperation is a helper for creating a synchronous nexus.Operation to deliver a signal.
+//
+// NOTE: Experimental
+func NewSignalWorkflowOperation(name string) nexus.Operation[SignalWorkflowInput, nexus.NoValue] {
```
bergundy, line 102:
```suggestion
func NewWorkflowSignalOperation[T any](name string, func(ctx context.Context, input T, options nexus.StartOpertionOptions) SignalWorkflowInput) nexus.Operation[T, nexus.NoValue] {
```

test/nexus_test.go
```diff
@@ -371,6 +378,56 @@ func TestNexusSyncOperation(t *testing.T) {
...
+			CallbackURL:    "http://localhost/test",
+			CallbackHeader: nexus.Header{"test": "ok"},
+		})
+	require.NoError(t, err)
+	require.Nil(t, result)
+
+	var out string
+	require.NoError(t, run.Get(ctx, &out))
+	require.Equal(t, "nexus", out)
+
```
bergundy, line 423: Test the request ID and links are properly attached.

test/nexus_test.go
```diff
@@ -371,6 +378,56 @@ func TestNexusSyncOperation(t *testing.T) {
 	})
 }
 
+func TestNexusSignalWorkflowOperation(t *testing.T) {
```
bergundy, line 381: Can probably just test from workflow.

internal/interceptor.go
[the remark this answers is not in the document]
pdoerner: I just went ahead and manually removed all the duplicates

temporalnexus/operation.go
```diff
@@ -94,6 +93,28 @@ func NewSyncOperation[I any, O any](
 	}
 }
 
+// SignalWorkflowInput is the input to a NewSignalWorkflowOperation.
+type SignalWorkflowInput = internal.ClientSignalWorkflowInput
```
bergundy, line 97: Don't reuse the type from the interceptors, you want a type that contains the workflow ID, optional run ID and signal name.

Quinn-With-Two-Ns:
Did we test what happens in a failure case like we signal a workflow that does not exist? I would think we would want that to fail the operation?

temporalnexus/operation.go
bergundy: The caller doesn't care that this is a signal, that's the handler's call.

bergundy (commented):
This is almost good to go AFAIC.

Something to consider here is that the behavior is different from the WorkflowRunOperation where the default is to pass through the operation input as workflow input. I'm wondering if users will find this inconsistency confusing.

@Quinn-With-Two-Ns you may want to comment on this.

Quinn-With-Two-Ns:
>Something to consider here is that the behavior is different from the WorkflowRunOperation where the default is to pass through the operation input as workflow input. I'm wondering if users will find this inconsistency confusing.

I think the difference here is signal needs to know the workflow ID, where as starting a workflow generates the workflow ID. I am not strongly apposed to adding both for signal, I just worry it would encourage a design where uses are passing their workflow IDs in their signal inputs

Quinn-With-Two-Ns:
@bergundy I sync'd with the rest of the SDK team and we all agreed that, at first, we just want to expose just the mapping approach.

pdoerner:
Planning to upgrade api, nexus SDK, and add handling for returned links in a followup PR. Other than that, I think I addressed all comments. @Quinn-With-Two-Ns and/or @bergundy LMK if this looks good to merge.

temporalnexus/operation.go
```diff
@@ -94,6 +93,44 @@ func NewSyncOperation[I any, O any](
 	}
 }
 
+// SignalWorkflowInput encapsulates the values required to send a signal to a workflow.
+type SignalWorkflowInput struct {
```
Quinn-With-Two-Ns, line 97: Nit: Experimental

Quinn-With-Two-Ns (approved):

PR #1783, opened 2025-01-17
Expose workflow client from Activity

### LINKED ISSUE #1761: Provide client from activity context

Allow users to access Temporal client from within an activity. This is likely a `GetClient(context.Context)` function in the `activity` package, but there may be circular import concerns. If so, maybe this is a `GetClientInActivity(context.Context)` function in the `client` package. Up to implementer. See https://github.com/temporalio/features/issues/203.

yuandrew:
## What was changed

Added a new `GetClient` function in the `activity` package

## Why?

Allow users to access Temporal client from within an activity

## Checklist

1. Closes #1761

2. How was this tested:
Add unit and integration test

3. Any docs updates needed?


internal/activity.go
```diff
@@ -264,6 +264,13 @@ func RecordActivityHeartbeat(ctx context.Context, details ...interface{}) {
 	getActivityOutboundInterceptor(ctx).RecordHeartbeat(ctx, details...)
 }
 
+// GetClient returns a client that can be used to interact with the Temporal
+// service from an activity.
+func GetClient(ctx context.Context) Client {
+	activityEnv := getActivityEnv(ctx)
+	return activityEnv.serviceInvoker.GetClient(ClientOptions{Namespace: activityEnv.workflowNamespace})
```
Quinn-With-Two-Ns, line 271: hm, but you are not using the client , or client options the worker was started with. So if the worker client was configured with a data converter or interceptor I don't think this client will have it.

Quinn-With-Two-Ns: I think you might need to just wire the client from worker constructor down to here.

activity/activity.go
```diff
@@ -26,7 +26,7 @@ package activity
 
 import (
 	"context"
-
+	"go.temporal.io/sdk/client"
```
cretz, line 29: I am slightly concerned about this package dependency. We now will never be able to have any package that `client` depends on depend on the `activity` package. Maybe that's ok? I wonder if it'd be ok to just do just have `GetClient` return `internal.Client` instead. Thoughts?

yuandrew: `internal.Client` seems fine with me since it's the same object, @Quinn-With-Two-Ns do you have any objections?

Quinn-With-Two-Ns: Yeah, we should just clarify in the Docs that the proper exported type is

internal/activity.go
```diff
@@ -264,6 +264,12 @@ func RecordActivityHeartbeat(ctx context.Context, details ...interface{}) {
 	getActivityOutboundInterceptor(ctx).RecordHeartbeat(ctx, details...)
 }
 
+// GetClient returns a client that can be used to interact with the Temporal
+// service from an activity.
+func GetClient(ctx context.Context) Client {
+	return getActivityEnv(ctx).client
```
Quinn-With-Two-Ns, line 270: @cretz should this be intercept-able?

Quinn-With-Two-Ns: My allow intercepting things like getting the loggers or metrics so it seems like we would allow this, but if we didn't in other SDKs I am fine to leave off.

test/workflow_test.go
```diff
@@ -3249,6 +3249,17 @@ func (w *Workflows) SelectorBlockSignal(ctx workflow.Context) (string, error) {
 	return hello, nil
 }
 
+func (w *Workflows) WorkflowClientFromActivity(ctx workflow.Context) error {
+	ctx = workflow.WithActivityOptions(ctx, w.defaultActivityOptions())
+	var activities *Activities
+	var result string
+	err := workflow.ExecuteActivity(ctx, activities.ClientFromActivity).Get(ctx, &result)
```
Quinn-With-Two-Ns, line 3256: Can we test local activity as well

yuandrew: oops, added the impl/test

Quinn-With-Two-Ns (approved):

PR #1790, opened 2025-01-24
Nexus WorkflowSignalWithStart Operation

pdoerner:
## What was changed

Added a new Nexus operation wrapping SignalWithStart

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalnexus/operation.go
+++ b/temporalnexus/operation.go
@@ -285,6 +285,97 @@ func (o *workflowRunOperation[I, O]) Start(
 	}, nil
 }
 
+type WorkflowSignalWithStartOperationInput[W, S any] struct {
+	WorkflowOptions client.StartWorkflowOptions
+	WorkflowArg     W
+	SignalName      string
+	SignalArg       S
+}
+
+type WorkflowSignalWithStartOperationOptions[I, W, S, O any] struct {
+	Name     string
+	Workflow func(workflow.Context, W) (O, error)
+	GetInput func(context.Context, I, nexus.StartOperationOptions) (WorkflowSignalWithStartOperationInput[W, S], error)
+	Handler  func(context.Context, I, nexus.StartOperationOptions) (WorkflowHandle[O], error)
+}
+
+type workflowSignalWithStartOperation[I, W, S, O any] struct {
+	nexus.UnimplementedOperation[I, O]
+
+	options WorkflowSignalWithStartOperationOptions[I, W, S, O]
+}
+
+func NewWorkflowSignalWithStartOperation[I, W, S, O any](
+	name string,
+	workflow func(workflow.Context, W) (O, error),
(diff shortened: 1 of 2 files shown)


temporalnexus/operation.go
```diff
@@ -285,6 +285,97 @@ func (o *workflowRunOperation[I, O]) Start(
 	}, nil
 }
 
+type WorkflowSignalWithStartOperationInput[W, S any] struct {
```
pdoerner, line 288: Not sure whether enforcing types for the workflow and signal args is really necessary. Could also mirror what was done for `SignalWorkflowInput` and declare them as `any`. Open to opinions on which is preferable.

temporalnexus/operation.go
```diff
@@ -285,6 +285,97 @@ func (o *workflowRunOperation[I, O]) Start(
...
+	nexus.UnimplementedOperation[I, O]
+
+	options WorkflowSignalWithStartOperationOptions[I, W, S, O]
+}
+
+func NewWorkflowSignalWithStartOperation[I, W, S, O any](
+	name string,
+	workflow func(workflow.Context, W) (O, error),
+	getInput func(context.Context, I, nexus.StartOperationOptions) (WorkflowSignalWithStartOperationInput[W, S], error),
+) nexus.Operation[I, O] {
```
bergundy, line 312: This LGTM.

temporalnexus/operation.go
```diff
@@ -285,6 +285,97 @@ func (o *workflowRunOperation[I, O]) Start(
...
+	WorkflowArg     W
+	SignalName      string
+	SignalArg       S
+}
+
+type WorkflowSignalWithStartOperationOptions[I, W, S, O any] struct {
+	Name     string
+	Workflow func(workflow.Context, W) (O, error)
+	GetInput func(context.Context, I, nexus.StartOperationOptions) (WorkflowSignalWithStartOperationInput[W, S], error)
+	Handler  func(context.Context, I, nexus.StartOperationOptions) (WorkflowHandle[O], error)
```
bergundy, line 299: I wonder if this would be easy for users to leverage. I'm not sure returning a `WorkflowHandle[0]` is desirable here, maybe we need a different return type.
We need a `Handler` function for when users have non-standard workflow signatures, it could leverage a helper method to attach the relevant fields to the signal-with-start request.

Here's an example of what this might look like
```go
func SignalWithStartUntypedWorkflow[R any](
	ctx context.Context,
	nexusOptions nexus.StartOperationOptions,
	startWorkflowOptions client.StartWorkflowOptions,
	signalName string,
	signalArg any,
	workflow any,
	workflowArgs ...any,
) error {
```

pdoerner:
Discussed design further offline, closing this for now.

PR #1791, opened 2025-01-24
Implement SignalWithStart as a synchronous Nexus operation

pdoerner:
## What was changed

Implemented a Nexus operation to wrap SignalWithStart
Refactored the SignalWorkflow Nexus operation to use the same options as SignalWithStart

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalnexus/operation.go
+++ b/temporalnexus/operation.go
@@ -93,46 +93,6 @@ func NewSyncOperation[I any, O any](
 	}
 }
 
-// SignalWorkflowInput encapsulates the values required to send a signal to a workflow.
-//
-// NOTE: Experimental
-type SignalWorkflowInput struct {
-	// WorkflowID is the ID of the workflow which will receive the signal. Required.
-	WorkflowID string
-	// RunID is the run ID of the workflow which will receive the signal. Optional. If empty, the signal will be
-	// delivered to the running execution of the indicated workflow ID.
-	RunID string
-	// SignalName is the name of the signal. Required.
-	SignalName string
-	// Arg is the payload attached to the signal. Optional.
-	Arg any
-}
-
-// NewWorkflowSignalOperation is a helper for creating a synchronous nexus.Operation to deliver a signal, linking the
-// signal to a Nexus operation. Request ID from the Nexus options is propagated to the workflow to ensure idempotency.
-//
-// NOTE: Experimental
-func NewWorkflowSignalOperation[T any](
-	name string,
-	getSignalInput func(context.Context, T, nexus.StartOperationOptions) SignalWorkflowInput,
-) nexus.Operation[T, nexus.NoValue] {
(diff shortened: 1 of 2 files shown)


temporalnexus/operation.go
```diff
@@ -108,45 +124,159 @@ type SignalWorkflowInput struct {
...
-			return nil, err
-		}
-		ctx = context.WithValue(ctx, internal.NexusOperationLinksKey, links)
+// NewWorkflowSignalWithStartOperation is a helper for creating a synchronous nexus.Operation to deliver a signal to a
+// workflow. If the indicated workflow is not running, a new run will be started. The workflow execution chain and
+// signal will be linked to a Nexus operation. Request ID from the Nexus options is propagated to the workflow to
+// ensure idempotency. The operation is complete as soon as the signal is delivered and returns no value.
+//
+// NOTE: Experimental
+func NewWorkflowSignalWithStartOperation[I any](
```
bergundy, line 182: I think it'd be more convenient to have separate options for signal and signal-with-start and only require a single handler method.

I'd go for something like:

```go
type WorkflowSignalWithStartInput struct {
	Workflow any
	Signal string
	WorkflowArgs []any
	SignalArg any
	StartWorkflowOptions client.StartWorkflowOptions
}

func NewWorkflowSignalWithStartOperation[I any](
	name string,
	handler func(context.Context, I, nexus.StartOperationOptions) (WorkflowSignalWithStartInput, error),
) nexus.Operation[I, nexus.NoValue]
```

The input of the operation is unlikely to map directly to the workflow input, and the workflow output doesn't get linked to the operation output.
There's no way to get a signal's type signature in the Go SDK, so there's not much value in exposing having this method be generic apart from just passing the operation input to the handler function.

temporalnexus/operation.go
```diff
@@ -108,45 +124,159 @@ type SignalWorkflowInput struct {
...
+		},
+	}
+}
 
-		links, err := convertNexusLinks(options.Links, GetLogger(ctx))
-		if err != nil {
-			return nil, err
-		}
-		ctx = context.WithValue(ctx, internal.NexusOperationLinksKey, links)
+// NewWorkflowSignalWithStartOperation is a helper for creating a synchronous nexus.Operation to deliver a signal to a
```
Quinn-With-Two-Ns, line 176: I think having `NewWorkflowSignalWithStartOperation` be synchronous is not very intuitive for users, I would expect a signalWithStart operation to be an async operation that attaches to the existing workflow

bergundy: To me, signal-with-start is about the signal, not about the workflow run. Users will typically use this API to ensure that there's a workflow running to receive the signal.

pdoerner:
Discussed design further offline, closing this for now.

PR #1792, opened 2025-01-27
[WiP] Add priority annotation

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/activity.go
```diff
@@ -165,6 +165,10 @@ type (
 		//
 		// NOTE: Experimental
 		Summary string
+
+		// Priority - Optional priority and fairness settings that control relative ordering of
+		// task processing when tasks are backed up in a queue.
```
stephanos, line 170: The docs on `commonpb.Priority` are already very detailed, no need to duplicate them here.

internal/internal_workflow_client.go
```diff
@@ -943,8 +944,9 @@ func (wc *WorkflowClient) QueryWorkflowWithOptions(ctx context.Context, request
 
 // DescribeTaskQueue returns information about the target taskqueue, right now this API returns the
 // pollers which polled this taskqueue in last few minutes.
-//  - taskqueue name of taskqueue
-//  - taskqueueType type of taskqueue, can be workflow or activity
+//   - taskqueue name of taskqueue
+//   - taskqueueType type of taskqueue, can be workflow or activity
+//
```
stephanos, line 949: Auto-formatted by my IDE; seems better now?

internal/activity.go
Quinn-With-Two-Ns: Just a note, I know this is in draft, but we don't reference raw proto types in SDK apis. We generally create a native SDK type and translate

Quinn-With-Two-Ns: See, for example, `RetryPolicy` and `commonpb.RetryPolicy` in these options

internal/client.go
```diff
@@ -835,6 +839,116 @@ type (
 		NonRetryableErrorTypes []string
 	}
 
+	// Priority contains metadata that controls relative ordering of task processing
```
stephanos, line 842: This is _almost_ verbatim the [(excellent) documentation from the protos](https://github.com/temporalio/api/pull/513/files#diff-de60adc512f72d87dc1e38c0f30b41a16dede9243dfaf6209b4af9b5447fa551R250). Only the first and last sentence were slightly edited.

internal/workflow.go
```diff
@@ -2513,6 +2521,34 @@ func convertFromPBRetryPolicy(retryPolicy *commonpb.RetryPolicy) *RetryPolicy {
...
+	return &commonpb.Priority{
+		PriorityKey:       priority.PriorityKey,
+		FairnessKey:       priority.FairnessKey,
+		FairnessWeight:    priority.FairnessWeight,
+		FairnessRateLimit: priority.FairnessRateLimit,
+		OrderingKey:       priority.OrderingKey,
+	}
+}
+
+func convertFromPBPolicy(priority *commonpb.Priority) *Priority {
```
dnr, line 2538:
```suggestion
func convertFromPBPrority(priority *commonpb.Priority) *Priority {
```

dnr (commented):
I think you need a type alias to make `internal.Priority` usable from user code? not sure what package it should go in.. probably `temporal`?

internal/client.go
```diff
@@ -835,6 +839,116 @@ type (
...
+	//  1. First, consider "priority": higher priority (lower number) goes first.
+	//  2. Next, consider fairness: try to dispatch tasks for different fairness keys
+	//     in proportion to their weight.
+	//  3. Finally, tasks may be ordered by an additional ordering key.
+	//
+	// Applications may use any subset of mechanisms that are useful to them and
+	// leave the other fields to use default values.
+	//
+	// NOTE: Not all queues in the system may support the "full" semantics of all priority
+	// fields. (Currently only support in matching task queues is planned.)
```
Quinn-With-Two-Ns, line 872: > (Currently only support in matching task queues is planned.)

Can we leave out server implementation details? The average SDK users does not know what a "matching task queue" is

internal/internal_workflow_client.go
```diff
@@ -1684,6 +1686,7 @@ func (w *workflowClientInterceptor) createStartWorkflowRequest(
 		CompletionCallbacks:      in.Options.callbacks,
 		Links:                    in.Options.links,
 		VersioningOverride:       versioningOverrideToProto(in.Options.VersioningOverride),
+		Priority:                 convertToPBPriority(in.Options.Priority),
```
Quinn-With-Two-Ns, line 1689: Just double checking, multi-op start will support priority right?

internal/internal_activity.go
```diff
@@ -74,6 +74,7 @@ type (
 		DisableEagerExecution  bool
 		VersioningIntent       VersioningIntent
 		Summary                string
+		Priority               *commonpb.Priority
```
Quinn-With-Two-Ns, line 77: What is the inheritance logic for `Priority` in a Workflow? If I call a workflow with a given priority and I don't set a priority when I start a child workflow or activity do they inherit the workflows?

internal/client.go
```diff
@@ -835,6 +839,116 @@ type (
...
+		// Rate limits are specified in tasks dispatched per second.
+		// As with weights, rate limits can come from different sources:
+		//  1. Rate limits for a small set of keys can be overridden in task queue
+		//     configuration with an API.
+		//  2. It can be attached to the workflow/activity in this field.
+		//  3. The "default" rate limit for keys that are not overridden can be set
+		//     in task queue configuration also.
+		//  4. Otherwise, a very high rate limit will be applied.
+
+		FairnessRateLimit float32
```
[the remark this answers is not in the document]
stephanos: Ah, I thought the idea was to have everything there on the feature branch right now - and only delete it shortly before we merge/release. But I'm happy do also do that now.

internal/internal_activity.go
stephanos: David's proto docs define it like this:
> // Priority is attached to workflows and activities. By default, activities
	// inherit Priority from the workflow that created them, but may override fields
	// when an activity is started or modified.

Yes, a child workflow would inherit the calling workflow's priority. Unless it overrides it.

For posterity, this is where that happens in server: https://github.com/temporalio/temporal/pull/7134/files#diff-bd46ccf32fa0605754c1426971b2e886c9a17d09ed7412474b51584ec7fee46fR904

internal/internal_workflow.go
```diff
@@ -218,6 +218,7 @@ type (
 		WorkflowIDConflictPolicy enumspb.WorkflowIdConflictPolicy
 		DataConverter            converter.DataConverter
 		RetryPolicy              *commonpb.RetryPolicy
+		Priority                 *commonpb.Priority
```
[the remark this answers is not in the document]
stephanos: As of right now, `Priority` is not present in `PollWorkflowTaskQueueResponse`. I don't know if there are future plans, @dnr?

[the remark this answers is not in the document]
Quinn-With-Two-Ns: It doesn't need to be on the `PollWorkflowTaskQueueResponse`, the SDK can extract it from the started event. Workflow Info already includes the `RetryPolicy` and it extracts that from the started event.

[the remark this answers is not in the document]
stephanos: Good point; seems reasonable to me then to expose it.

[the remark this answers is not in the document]
dnr: Yes, it's in the blueprint. Oh.. but the part about WorkflowInfo is wrong and corrected in a comment only. I'll fix it.

So yeah, I think the next PR should do ActivityInfo and WorkflowInfo

internal/client.go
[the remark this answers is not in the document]
dnr: We could do it either way, it's easy to add back and easy to remove

dnr (approved):

internal/client.go
```diff
@@ -835,6 +839,33 @@ type (
...
+	// For all fields, the field not present or equal to zero/empty string means to
+	// inherit the value from the calling workflow, or if there is no calling
+	// workflow, then use the default value.
+	//
+	// Exposed as: [go.temporal.io/sdk/temporal.Priority]
+	Priority struct {
+		// PriorityKey is a positive integer from 1 to n, where smaller integers
+		// correspond to higher priorities (tasks run sooner). In general, tasks in
+		// a queue should be processed in close to priority order, although small
+		// deviations are possible.
```
Sushisource, line 859: Probably worth mentioning what will happen if it's set <1. My vote would be to just clamp it to 1.

dnr: `0` means "use default". I'm not sure what negative would mean, we could say also default

internal/internal_workflow.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I think we will need to update the proto API to add `commonpb.Priority` to the `PollActivityTaskQueueResponse` t do activties

internal/client.go
```diff
@@ -835,6 +839,33 @@ type (
 		NonRetryableErrorTypes []string
 	}
 
+	// Priority contains metadata that controls the relative ordering of task processing
+	// when tasks are backed up in a queue. The affected queues depend on the
+	// server version.
+	//
+	// Priority is attached to workflows and activities. By default, activities
```
Quinn-With-Two-Ns, line 846: @stephanos based of our previous conversation https://github.com/temporalio/sdk-go/pull/1792/files#r1934130657 should this say

> By default, activities and child workflows inherit Priority from the workflow that created them

?

stephanos: I agree, that makes it more explicit 👍

internal/client.go
stephanos: What do you think about returning "invalid argument" instead? Removes the ambiguity. Users can't know what `n` is since it's based on the server config, but they do know that <1 is invalid.

internal/internal_workflow.go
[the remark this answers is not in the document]
stephanos: `PollActivityTaskQueueResponse` already has the Priority field, actually.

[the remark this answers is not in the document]
stephanos: I added Priority to ActivityInfo and WorkflowInfo here now, seemed easy enough (unless I missed sth).

internal/internal_activity.go
Quinn-With-Two-Ns: Ah OK It happens in the server, I wasn't sure where the logic was supposed to live.

cretz:
@stephanos - I see this is Go SDK only? Do you have plans to author this for other SDKs? We don't consider Go any more important than others. If we want this feature in all SDKs, without favoritism, can you open an issue at https://github.com/temporalio/features and we can add this to all SDKs (including Go) when prioritized?

internal/activity.go
```diff
@@ -62,6 +62,7 @@ type (
 		Deadline          time.Time     // Time of activity timeout
 		Attempt           int32         // Attempt starts from 1, and increased by 1 for every retry if retry policy is specified.
 		IsLocalActivity   bool          // true if it is a local activity
+		Priority          *Priority
```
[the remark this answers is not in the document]
dnr: It's meant to be optional. Many (most) users won't use this relatively advanced feature, and sending these priority attachments everywhere when they're not being used is additional overhead that we can avoid.

If it was not a pointer, then the sdk would have to attach the message for each command/api call, right? Are you suggesting it compare the whole struct to the zero value and attach it only if non-zero (i.e. at least one field non-zero)? That would work.

internal/client.go
dnr: Are you suggesting returning invalid argument only for < 0, and accepting > n? That's fine.

I don't want to reject > n (where n is the current max) because: 1. n belongs to matching service and frontend doesn't necessarily know about it, 2. n can change over time

internal/activity.go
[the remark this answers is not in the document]
cretz: > Are you suggesting it compare the whole struct to the zero value and attach it only if non-zero (i.e. at least one field non-zero)? 

Yes, I think that's the best way to differentiate empty from nil. I agree there should be only one representation of this concept proto side so hopefully the converter from Go type to proto type does not write the proto if empty (or nil today).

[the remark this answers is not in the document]
dnr: That works for me. @stephanos?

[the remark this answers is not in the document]
Quinn-With-Two-Ns: So we are saying there is no difference between a `nil` and zero value struct from the Server POV?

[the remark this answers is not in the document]
dnr: Yes, I tried to make that clear in the api docs. If it's not, is there different wording there that you think would make it more clear?

internal/client.go
stephanos: @dnr yup, that's exactly what I thought 👍

internal/activity.go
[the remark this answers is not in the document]
stephanos: I can make that change 👍

stephanos:
Looks like Spencer created https://github.com/temporalio/features/issues/593 now 👍

internal/client.go
stephanos: I'm not sure we need to call out the "negative number" case here explicitly, though, as the docs already state it's either default (zero) or a positive number - implying that a negative is not allowed.

Looking at other doc strings, such as `MaximumAttempts` or `BackoffCoefficient`, they aren't spelling that out either.

dnr: I agree it's not worth mentioning in comments, "positive integer" is enough. We'll do basic validation on these in frontend and can reject negative there.

Sushisource: Rejecting in the frontend sounds great

internal/activity.go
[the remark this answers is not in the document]
stephanos: https://github.com/temporalio/sdk-go/pull/1792/commits/b3a899fc71858b029439b978487f62704be81a24 changes all user-facing fields to be of type `Priority` (no pointer) and only sends the Priority proto to the server if the fields are non-zero/default.

internal/workflow.go
```diff
@@ -2522,8 +2522,10 @@ func convertFromPBRetryPolicy(retryPolicy *commonpb.RetryPolicy) *RetryPolicy {
 	return &p
 }
 
-func convertToPBPriority(priority *Priority) *commonpb.Priority {
-	if priority == nil {
+func convertToPBPriority(priority Priority) *commonpb.Priority {
+	// If the priority only contains default values, return nil instead
+	// - since there's no need to send the default values to the server.
+	if priority.PriorityKey == 0 {
```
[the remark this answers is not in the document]
stephanos: Oh right; I forgot that this works when all fields are primitives!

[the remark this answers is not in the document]
stephanos: FWIW, oddly Go doesn't like the struct literal inline:
```
./workflow.go:2528:28: syntax error: unexpected { at end of statement
./workflow.go:2532:2: syntax error: non-declaration statement outside function body
```

Moved it into a variable.

temporal/priority.go
```diff
@@ -0,0 +1,28 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package temporal
+
+import "go.temporal.io/sdk/internal"
+
+// Priority defines the priority for activity/workflow.
+type Priority = internal.Priority
```
cretz, line 28: A bit apprehensive about taking the English term "priority" for the entire SDK to mean this specific use of priorities (similar to my apprehensiveness in the API of doing the same, though it's more pronounced in the SDK), but I have no better term so it's probably ok.

internal/client.go
```diff
@@ -835,6 +839,33 @@ type (
...
+	Priority struct {
+		// PriorityKey is a positive integer from 1 to n, where smaller integers
+		// correspond to higher priorities (tasks run sooner). In general, tasks in
+		// a queue should be processed in close to priority order, although small
+		// deviations are possible.
+		//
+		// The maximum priority value (minimum priority) is determined by server
+		// configuration, and defaults to 5.
+		//
+		// The default value is calculated by (min+max)/2. With the
```
cretz, line 864:
```suggestion
		// The default value when unset or 0 is calculated by (min+max)/2. With the
```

internal/workflow_test.go
cretz, line 1: Before this hits `master`, we may want a test confirming behavior (e.g. creating workflows of different priorities while worker is not running, then worker of max workflow tasks 1 gets the higher priority first or something). But since this is feature branch for now, no prob.

cretz (approved):
LGTM, nothing blocking

Quinn-With-Two-Ns (approved):

internal/workflow_test.go
Quinn-With-Two-Ns: I do not think the SDK should try to test the server behaviour, the only tests I would have for the SDK is to test that if the priority is set on a certain request the priority is set on the object. So if a workflow is started with a certain priority then the workflow info should have the same priority.

cretz: I think it'd be nice to test that it even works not that we just set the value. We don't have to test all variations but to know it even works before releasing to users is reasonable (and might as well put it in a place where it is continually tested instead of done manually).

internal/workflow.go
```diff
@@ -2513,6 +2522,32 @@ func convertFromPBRetryPolicy(retryPolicy *commonpb.RetryPolicy) *RetryPolicy {
 	return &p
 }
 
+func convertToPBPriority(priority Priority) *commonpb.Priority {
+	// If the priority only contains default values, return nil instead
+	// - since there's no need to send the default values to the server.
+	//
+	// Exposed as: [go.temporal.io/sdk/temporal.Priority]
```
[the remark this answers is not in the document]
stephanos: Yup, it is! I wasn't sure whether this was intentional behavior or not. I can file an issue.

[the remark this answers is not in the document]
cretz: Thanks! Though you may want to transfer that issue over to this repo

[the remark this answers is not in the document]
stephanos: Oops, I must have used the wrong tab!

internal/workflow_test.go
stephanos: @cretz @Quinn-With-Two-Ns to pick this up again, what's the level of testing we can/should achieve in the SDK?

I can add tests like these [https://github.com/temporalio/sdk-go/pull/1859] to show that the priority options are correctly passed around.

I can also extend that test and mock all the server responses to test the whole flow end-to-end (ie start workflow, process workflow, process activity etc).

stephanos: Right; I'm asking because I'd like to remove any remaining obstacles before we _want to_ merge to master; so that we're not blocked on anything once everything else is ready to go. We have end-to-end tests including API, Server and SDK already running regularly and are at the quality-control stage. But ofc if we drop a line here somewhere, it would break and the SDK tests wouldn't catch that. I thought that's the concern that was voiced here; and I'd like to address it.

Quinn-With-Two-Ns: To merge to master we need tests in the SDK repo for the feature. In my opinion, those tests don't need to cover the server side implementation. By that I mean they don't need to verify the priority values are actually respected. 

The SDK tests just need to verify, the values the user passed are properly "serialized" and sent on all relevant requests, workflows and activities are properly deserializing and exposing the values, and setting the values from a workflow behaves correctly and defaults are respected. Does that make sense?

Issue #1788, opened 2025-01-24
Allow customization of dev server UI port


cretz:
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/588

PR #1793, opened 2025-01-28
Add option to override UI port

Quinn-With-Two-Ns:
Add option to override UI port

closes: https://github.com/temporalio/sdk-go/issues/1788

### DIFF
--- a/testsuite/devserver.go
+++ b/testsuite/devserver.go
@@ -71,6 +71,9 @@ type DevServerOptions struct {
 	DBFilename string
 	// Whether to enable the UI.
 	EnableUI bool
+	// Override UI port if EnableUI is true.
+	// If not provided, a free port will be used.
+	UIPort string
 	// Log format - defaults to "pretty".
 	LogFormat string
 	// Log level - defaults to "warn".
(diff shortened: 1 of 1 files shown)


testsuite/devserver.go
```diff
@@ -158,6 +161,9 @@ func prepareCommand(options *DevServerOptions, host, port, namespace string) []s
 	if options.DBFilename != "" {
 		args = append(args, "--db-filename", options.DBFilename)
 	}
+	if options.UIPort != "" {
+		args = append(args, "--ui-port", options.UIPort)
```
Sushisource, line 165: Does this also enable if not enabled? Docstring should mention that too

Quinn-With-Two-Ns: No it does not enable the UI

Quinn-With-Two-Ns: I am matching the behaviour of the dev server here

cretz (approved):

testsuite/devserver.go
cretz: I think in a lot of cases they are independently set via env vars and you may disable UI without changing port especially in hierarchical config cases like k8s/helm.

Issue #1756, opened 2024-12-15
Malformed code links in API docs


dandavison:
There are various places in the API docs where the intention was to create a link, but the API docs renderer has not created a link. It looks like relevant documentation for creating links is at https://pkg.go.dev/github.com/fluhus/godoc-tricks#hdr-Documentation_Links-Links. We should do a pass  replacing malformed links with correct links.

Examples:

https://pkg.go.dev/go.temporal.io/sdk@v1.31.0/internal#UpdateWorkerVersioningRulesOptions
```
UpdateWorkerVersioningRulesOptions is the input to [Client.UpdateWorkerVersioningRules]
```

https://pkg.go.dev/go.temporal.io/sdk@v1.31.0/client#UpdateWithStartWorkflowOptions
```
UpdateWorkerVersioningRulesOptions is the input to [Client.UpdateWorkerVersioningRules]
```

https://pkg.go.dev/go.temporal.io/sdk@v1.31.0/internal#WithStartWorkflowOperation
```
WithStartWorkflowOperation defines how to start a workflow when using UpdateWithStartWorkflow. See [NewWithStartWorkflowOperation] and [UpdateWithStartWorkflow]
```

PR #1795, opened 2025-01-28
Fix doc links on pkg.go.dev/

yuandrew:
## What was changed

Fixed all of the broken links that wouldn't generate on https://pkg.go.dev/go.temporal.io/sdk

## Why?

Most of the links I changed are actually correct, there is a bug that's been around for over 2 years in go/doc https://github.com/golang/go/issues/54033. Current workaround solution is quite verbose, but i think it's better to have proper links. We can easily revert back when this issue is fixed. (There were a few incorrect links I fixed)

## Checklist

1. Closes  #1756 

2. How was this tested:

Built docs locally with `pkgsite -http=:6060` and verified broken links are fixed.

3. Any docs updates needed?

### DIFF
--- a/activity/activity.go
+++ b/activity/activity.go
@@ -73,7 +73,7 @@ func GetMetricsHandler(ctx context.Context) metrics.Handler {
 // can check error with TimeoutType()/Details().
 //
 // Note: If using asynchronous activity completion,
-// after returning [ErrResultPending] users should heartbeat with [client.Client.RecordActivityHeartbeat]
+// after returning [ErrResultPending] users should heartbeat with [github.com/temporalio/sdk-go/client.Client.RecordActivityHeartbeat]
 func RecordHeartbeat(ctx context.Context, details ...interface{}) {
 	internal.RecordActivityHeartbeat(ctx, details...)
 }
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ If you're unsure then you should specify it anyways as there's no harm in doing 
 ### Incompatible proto/json encoding
 
 Proto enums will, when formatted to JSON, now be in SCREAMING_SNAKE_CASE rather than PascalCase.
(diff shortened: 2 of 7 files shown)


activity/activity.go
```diff
@@ -73,7 +73,7 @@ func GetMetricsHandler(ctx context.Context) metrics.Handler {
 // can check error with TimeoutType()/Details().
 //
 // Note: If using asynchronous activity completion,
-// after returning [ErrResultPending] users should heartbeat with [client.Client.RecordActivityHeartbeat]
+// after returning [ErrResultPending] users should heartbeat with [github.com/temporalio/sdk-go/client.Client.RecordActivityHeartbeat]
```
Quinn-With-Two-Ns, line 76: When I tested this locally the link did not work, is that expected?

yuandrew: locally, as in your IDE? In my IDE, both my solution and the original `[client.Client.RecordActivityHeartbeat]` don't link to anything. Do you see something different? I'm using GoLand

yuandrew: ahh, this is because the `activity` package never imports the `client` package. So if we want this to show locally, we would need to import the `client` package. but not sure we want to add that dependency just for this link? And it seems like we can't have both the workaround for the docs site as well as local linking, would need to prioritize one over the other

Quinn-With-Two-Ns: by locally I mean by running `pkgsite -http=:6060`. Oh I see, how do you know thiis will work on the docs site then?

yuandrew: oh, actually good catch, turns out fixing the link to `go.temporal.io/sdk/` allows IDE rendering as well!

yuandrew: Screenshot 2025-02-03 at 1 25 17 PM

Quinn-With-Two-Ns (approved):

PR #1797, opened 2025-01-30
[Nexus] Set OnConflictOptions for WorkflowRunOperation

rodrigozhou:
## What was changed

Add `OnConflictOptions` to `StartWorkflowOptions`, and set it for Nexus `WorkflowRunOperation`.

Tests are done in https://github.com/temporalio/sdk-go/pull/1828

## Why?

Enabling attaching callbacks to running Nexus workflows.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -1208,9 +1208,6 @@ func SetLinksOnStartWorkflowOptions(opts *StartWorkflowOptions, links []*commonp
 // options on StartWorkflowOptions.
 // OnConflictOptions are purposefully not exposed to users for the time being.
 func SetOnConflictOptionsOnStartWorkflowOptions(opts *StartWorkflowOptions) {
-	if opts.WorkflowIDConflictPolicy == enumspb.WORKFLOW_ID_CONFLICT_POLICY_UNSPECIFIED {
-		opts.WorkflowIDConflictPolicy = enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING
-	}
 	opts.onConflictOptions = &OnConflictOptions{
 		AttachRequestID:           true,
 		AttachCompletionCallbacks: true,
--- a/internal/client.go
+++ b/internal/client.go
@@ -647,8 +647,11 @@ type (
 		// When WorkflowExecutionErrorWhenAlreadyStarted is true, Client.ExecuteWorkflow will return an error if the
 		// workflow id has already been used and WorkflowIDReusePolicy or WorkflowIDConflictPolicy would
 		// disallow a re-run. If it is set to false, rather than erroring a WorkflowRun instance representing
-		// the current or last run will be returned. However, when WithStartOperation is set, this field is ignored and
-		// the WorkflowIDConflictPolicy UseExisting must be used instead to prevent erroring.
+		// the current or last run will be returned. However, this field is ignored in the following cases:
+		// - when WithStartOperation is set;
+		// - in the Nexus WorkflowRunOperation.
+		// When this field is ignored, you must set WorkflowIDConflictPolicy to UseExisting to prevent
+		// erroring.
 		//
 		// Optional: defaults to false
 		WorkflowExecutionErrorWhenAlreadyStarted bool
--- a/internal/client.go
+++ b/internal/client.go
@@ -742,6 +742,14 @@ type (
 		callbacks []*commonpb.Callback
 		// links. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
 		links []*commonpb.Link
+
+		// OnConflictOptions - Optional workflow ID conflict options used in conjunction with conflict policy
+		// WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING. If onConflictOptions is set and a workflow is already
+		// running, the options specifies the actions to be taken on the running workflow. If not set or use
+		// together with any other WorkflowIDConflictPolicy, this parameter is ignored.
+		//
+		// NOTE: Only settable by the SDK -- e.g. [temporalnexus.workflowRunOperation].
+		onConflictOptions *OnConflictOptions
 	}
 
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/pborman/uuid v1.2.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.10.0
-	go.temporal.io/api v1.43.2
```
Quinn-With-Two-Ns, line 16: This will need to be a released version of the API to release this change.

internal/client.go
```diff
@@ -1288,3 +1296,17 @@ func SetCallbacksOnStartWorkflowOptions(opts *StartWorkflowOptions, callbacks []
...
 	opts.links = links
 }
+
+// SetOnConflictOptionsOnStartWorkflowOptions is an internal only method for setting conflict
+// options on StartWorkflowOptions.
+// OnConflictOptions are purposefully not exposed to users for the time being.
+func SetOnConflictOptionsOnStartWorkflowOptions(opts *StartWorkflowOptions) {
+	if opts.WorkflowIDConflictPolicy == enumspb.WORKFLOW_ID_CONFLICT_POLICY_UNSPECIFIED {
+		opts.WorkflowIDConflictPolicy = enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING
+	}
```
bergundy, line 1306: Not convinced we want to change the default conflict policy, especially not in a way that's hard to discover.
Without reading this function's implementation, I'd never guess that the conflict policy changes implicitly here.
Let's revert this change to the default behavior and let the user opt into this.

internal/internal_workflow_execution_options.go
```diff
@@ -144,6 +154,17 @@ func versioningOverrideFromProto(versioningOverride *workflowpb.VersioningOverri
 	}
 }
 
+func onConflictOptionsToProto(onConflictOptions *OnConflictOptions) *workflowpb.OnConflictOptions {
```
bergundy, line 157: nit: this could be a method on the `OnConflictOptions` struct but I'd just use the proto directly here since we're not exposing this API to users yet.

internal/client.go
```diff
@@ -1288,3 +1296,17 @@ func SetCallbacksOnStartWorkflowOptions(opts *StartWorkflowOptions, callbacks []
...
+	opts.onConflictOptions = &OnConflictOptions{
+		AttachRequestID:           true,
+		AttachCompletionCallbacks: true,
+		AttachLinks:               true,
+	}
```
bergundy, line 1311: You can keep this on all start requests regardless of the conflict policy, the server should ignore it unless the request specifies "use existing".

internal/client.go
rodrigozhou: Sure, I thought we wanted to make Nexus requests to always use this conflict policy.

cretz:
@rodrigozhou - I see this is Go SDK only? Do you have plans to author this for other SDKs? We don't consider Go any more important than others. If we want this feature in all SDKs, without favoritism, can you open an issue at https://github.com/temporalio/features and we can add this to all SDKs (including Go) when prioritized?

bergundy (approved):
LGTM but we'll need a tagged api before merging this.

go.mod
cretz: I'd argue we need a released version of the server and a test proving this works to even merge

cretz:
> but we'll need a tagged api before merging this.

I'd argue we need a test confirming this even works before merging. If it's not going to be in a stable dev server release anytime soon, we can discuss a dev server RC release.

bergundy:
> > but we'll need a tagged api before merging this.
> 
> I'd argue we need a test confirming this even works before merging. If it's not going to be in a stable dev server release anytime soon, we can discuss a dev server RC release.

@rodrigozhou can you add a test in the server for now with this SDK version? There's going to be a server 1.27.0 RC fairly soon, we can open an issue to add a test SDK side once that's ready.

Quinn-With-Two-Ns:
Don't we need workflow test environment support?

bergundy:
> Maybe I just missed it, but shouldn't some user facing SDK docs be updated as part of this PR? This is a new feature, but nothing in docs describe this new behaviour.

Yeah, good callout, this should be documented on the `temporalnexus` methods (e.g. `NewWorkflowRunOperation+WithOptions`, `ExecuteWorkflow`, `ExecuteUntypedWorkflow`).

bergundy:
> Don't we need workflow test environment support?

I'd say yes but low priority.

Quinn-With-Two-Ns:
I consider workflow test environment support a requirement

cretz:
> they just don't cover the e2e flow with SDK.
I think we need an e2e flow with the SDK (should be no rush to ship something to our users that doesn't work in a server they can use anyways). Obviously it's more of a smoke test than full coverage, but it reasonable to confirm it works.

rodrigozhou:
@cretz @Quinn-With-Two-Ns I created a separate PR that adds supports for this in the test framework, and added some tests there as well.

cretz (approved):
Still need @Quinn-With-Two-Ns approval

Quinn-With-Two-Ns:
Now that we have the CLI release can we add integration tests?

rodrigozhou:
@Quinn-With-Two-Ns I added in https://github.com/temporalio/sdk-go/pull/1828.
However, it needs the latest server tag, which is not in the latest CLI tag.

Quinn-With-Two-Ns:
@rodrigozhou you mean `v1.3.0-versioning.0` does not work?

rodrigozhou:
Yes, `v1.3.0-versioning.0` uses `v1.27.0-128.4`, but it needs a fix that in the `main` branch, but not in the cloud branch.

cretz:
Mentioned in comment above (though I marked approved since thinking it was there), I think for this to get to `master` I think we need a working integration test

PR #1799, opened 2025-02-03
Expand arguments in UpdateWorkflow call when args equal one

disco07:
## What was changed
Ensure UpdateWorkflowNoRejection is correctly triggered with one or variadic arguments.

## Why?
It's a bug in unit test.

## Checklist
1. Closes
2. How was this tested
3. Any docs updates needed?

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -1133,7 +1133,12 @@ func (e *TestWorkflowEnvironment) UpdateWorkflowNoRejection(updateName string, u
 		OnAccept:   func() {},
 		OnComplete: func(interface{}, error) {},
 	}
-	e.UpdateWorkflow(updateName, updateID, uc, args)
+
+	if len(args) == 1 {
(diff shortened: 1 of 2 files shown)


internal/workflow_testsuite.go
```diff
@@ -1133,7 +1133,12 @@ func (e *TestWorkflowEnvironment) UpdateWorkflowNoRejection(updateName string, u
 		OnAccept:   func() {},
 		OnComplete: func(interface{}, error) {},
 	}
-	e.UpdateWorkflow(updateName, updateID, uc, args)
+
+	if len(args) == 1 {
```
Quinn-With-Two-Ns, line 1137: Shouldn't you always expand it?

disco07: You're right, I've made one change

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
@yuandrew Do you mind taking a look as well?

yuandrew (approved):
Thanks for this fix!

PR #1802, opened 2025-02-05
Upgrade to Nexus SDK 0.2.0

bergundy:
Do not merge before we update the dependencies to use a published api-go and Nexus Go SDK.

Tested against server 1.26.2 and changes made in https://github.com/temporalio/temporal/pull/7241.


internal/failure_converter.go
```diff
@@ -286,7 +295,8 @@ func (dfc *DefaultFailureConverter) FailureToError(failure *failurepb.Failure) e
 			Endpoint:         info.GetEndpoint(),
 			Service:          info.GetService(),
 			Operation:        info.GetOperation(),
-			OperationID:      info.GetOperationId(),
+			OperationToken:   token,
+			OperationID:      token,
 		}
 	} else if info := failure.GetNexusHandlerFailureInfo(); info != nil {
```
Quinn-With-Two-Ns, line 301: Don't we need to handle the retry policy here?

bergundy: Yes, thanks!

bergundy: I forgot to add this to the api.

temporalnexus/operation.go
```diff
@@ -52,6 +56,18 @@ import (
...
 )
 
+// tokenSource is used as the "source" operation token header.
+const tokenSource = "io.temporal.sdk"
+
+// workflowRunOperationTokenType is used as the "type" operation token header for workflow run operations.
+const workflowRunOperationTokenType = "WorkflowRunOperation"
+
+// workflowRunOperationToken is the decoded form of the workflow run operation token.
+type workflowRunOperationToken struct {
```
Quinn-With-Two-Ns, line 66: Why not the run ID as well?

temporalnexus/operation.go
```diff
@@ -340,13 +384,19 @@ func ExecuteUntypedWorkflow[R any](
...
 	}
 
+	var encodedToken string
 	if nexusOptions.CallbackURL != "" {
 		if nexusOptions.CallbackHeader == nil {
 			nexusOptions.CallbackHeader = make(nexus.Header)
 		}
-		if idHeader := nexusOptions.CallbackHeader.Get(nexus.HeaderOperationID); idHeader == "" {
-			nexusOptions.CallbackHeader.Set(nexus.HeaderOperationID, startWorkflowOptions.ID)
+		encodedToken, err = generateWorkflowRunOperationToken(nctx.Namespace, startWorkflowOptions.ID)
```
Quinn-With-Two-Ns, line 402: What if the `encodedToken` exceeds the server limit, but `startWorkflowOptions.ID` doesn't? So the workflow gets started but the start request cannot succeed?

temporalnexus/operation.go
```diff
@@ -52,6 +56,18 @@ import (
...
 
+// tokenSource is used as the "source" operation token header.
+const tokenSource = "io.temporal.sdk"
+
+// workflowRunOperationTokenType is used as the "type" operation token header for workflow run operations.
+const workflowRunOperationTokenType = "WorkflowRunOperation"
+
+// workflowRunOperationToken is the decoded form of the workflow run operation token.
+type workflowRunOperationToken struct {
+	NamespaceName string `json:"ns"`
```
Quinn-With-Two-Ns, line 67: Is this format expected to be consistent across SDKs?

temporalnexus/operation.go
bergundy: We generate the operation token before we have a run ID because we send it with the start request callback headers for complete-before-start.

temporalnexus/operation.go
bergundy: Good point but we should just bump the server limits for the token.

temporalnexus/operation.go
bergundy: I'd expect it to. Happy for others to chime in.

temporalnexus/operation.go
```diff
@@ -396,3 +457,53 @@ func convertNexusLinks(nexusLinks []nexus.Link, log log.Logger) ([]*common.Link,
...
+		Data:   data,
+	}, nil
+}
+
+func generateWorkflowRunOperationToken(namespace, workflowID string) (string, error) {
+	token := workflowRunOperationToken{
+		NamespaceName: namespace,
+		WorkflowID:    workflowID,
+	}
+	data, err := json.Marshal(token)
```
Quinn-With-Two-Ns, line 498: Do we need the mapping from `namespace` + `workflowID` -> token to be deterministic? I am hopping no, but wanted to double check. If the operation token sent from start mismatched with the one in the start request callback header what would happen?

bergundy: Based on our call today we concluded that it's not required to be deterministic.

internal/internal_nexus_task_handler.go
```diff
@@ -444,9 +467,17 @@ func (h *nexusTaskHandler) nexusHandlerErrorToProto(handlerErr *nexus.HandlerErr
 	if err != nil {
 		return nil, err
 	}
+	var retryBehavior enumspb.NexusHandlerErrorRetryBehavior
```
Quinn-With-Two-Ns, line 470: nit: duplicate code with https://github.com/temporalio/sdk-go/pull/1802/files#diff-7fd7ede0bd4abb4be242c028e1a1e768f47c4ea217de23174aa9868e0f9cc686R188

internal/internal_nexus_task_handler.go
```diff
@@ -469,16 +500,13 @@ var emptyReaderNopCloser = io.NopCloser(bytes.NewReader([]byte{}))
...
-		return nexus.NewFailedOperationError(workflowErr)
-	}
-	if queryRejectedErr, ok := err.(*QueryRejectedError); ok {
-		return nexus.NewFailedOperationError(queryRejectedErr)
-	}
 	// Not using errors.As to be consistent ApplicationError checking with the rest of the SDK.
 	if appErr, ok := err.(*ApplicationError); ok && appErr.NonRetryable() {
-		return nexus.NewFailedOperationError(appErr)
+		return &nexus.HandlerError{
+			// TODO(bergundy): Change this to a non retryable internal error after the 1.27.0 server release.
```
Quinn-With-Two-Ns, line 506: Can we create to track this? Also doesn't this need to be done before GA since it is a breaking change?

internal/internal_nexus_task_handler.go
```diff
@@ -444,9 +467,17 @@ func (h *nexusTaskHandler) nexusHandlerErrorToProto(handlerErr *nexus.HandlerErr
 	if err != nil {
```
Quinn-With-Two-Ns, line 467: Nit: it seems like we have all the Nexus error conversion logic here https://github.com/temporalio/sdk-go/pull/1802/files#diff-82a19cd3836c5a75e55d9c5d4400c0bfdd09029eb29fd1fa9755cf1a724f4654R86 except this one function. Not blocking but might be nice to group it.

temporalnexus/token.go
```diff
@@ -0,0 +1,79 @@
...
+		WorkflowID:    workflowID,
+	}
+	data, err := json.Marshal(token)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal workflow run operation token: %w", err)
+	}
+	return base64.URLEncoding.WithPadding(base64.NoPadding).EncodeToString(data), nil
+}
+
+func loadWorkflowRunOperationToken(data string) (workflowRunOperationToken, error) {
```
Quinn-With-Two-Ns, line 59: During the meeting yesterday @cretz suggested we reserve `v` for a versioning field and fail if it is present. Did we decided not to do that?

internal/failure_converter.go
Quinn-With-Two-Ns: I understand we can't have a full integration test, but can we at least have a test that the retry-able flag is being properly serialized and deserialized?

bergundy (commented):
I think we need an interceptor to get the client, I can add it here or in a separate PR.

internal/failure_converter.go
Quinn-With-Two-Ns: Yeah for this PR could we just have a unit test for the failure converter, we should have that anyway regardless of our E2E test coverage

Quinn-With-Two-Ns:
> I think we need an interceptor to get the client, I can add it here or in a separate PR.

Not sure I understand what do you mean exactly by "need an interceptor" here? We don't have Temporal Nexus inbound or outbound interceptors that is on my plate. When we add them `GetClient` would be part of the outbound interceptor.

internal/failure_converter.go
bergundy: Coming in the next commit.

internal/internal_nexus_task_handler.go
bergundy: There's maybe a slight duplication but I don't think it's very critical apart from the grouping.

internal/internal_nexus_task_handler.go
bergundy: https://github.com/temporalio/sdk-go/issues/1810

temporalnexus/token.go
bergundy: Coming in the next commit.

Quinn-With-Two-Ns (approved):

PR #1803, opened 2025-02-05
Fix child WF ID generation

gow:
## What was changed
We now use `workflowInfo.OriginalRunID` in parent when it is generating a child workflow ID.

## Why?
We need this change to enable resetting of workflows that have pending children. What this means is Go SDK should be able to successfully replay `StartChildWorkflowExecutionInitiated` event in a parent that was reset.
Without this change Go SDK will run into non-determinism error since the child ID it generates (based on new run ID) doesn't match with the ID available in the `StartChildWorkflowExecutionInitiated` event. Since `OriginalRunID` stays the same across resets, with this change the SDK can successfully replay `StartChildWorkflowExecutionInitiated` events.

## Checklist

1. Closes 
N/A

2. How was this tested:
Manually tested it. Added integration tests and replay tests.

3. Any docs updates needed?
N/A


gow:
> Questions here:
> 
> 1. Can we confirm what Java does here?

Java seems to be generating UUIDs when an ID is not provided in ChildWorkflow options. Additionally when I replayed an existing `StartChildWorkflowExecutionInitiated` event (by resetting to a point *after* the child completion) it didn't run into non-determinism error. Go on the other hand ran into non-det error complaining about workflowID. (`"Error":"[TMPRL1100] unknown command CommandType: ChildWorkflow, ID: 0174afb2-0213-4644-9214-8b977310cc9f_14, possible causes are nondeterministic workflow definition code or incompatible change in the workflow definition"`)
 
> 2. IIUC Core-based languages make child workflow IDs with a seeded random that changes seed on reset midway through execution, meaning post-reset child workflow IDs would be different than they were on original run but pre-reset ones are the same. This was done very much by intention. Should we consider a similar approach in Go/Java that has a "current run ID" that can change midway through execution?

Absolutely yes. Java seems to be already doing this (at least in my testing). Looks like it changes the seed at reset-point?
If this is an easier change in Go-SDK, I'd rather do it. Otherwise, we still need this fix since Go-SDK is currently broken (See below)

> 3. Can we assess the impact of behavior changes like this on existing users with existing expectations? I am guessing there is no impact because you could not reset to before child workflows?

Yes. So far all server versions have prevented resetting of workflows when the reset point falls in between child Init & Completed events (i.e within the lifespan of child). So no potential user impact or backward compatibility issues there.
However the server has always allowed workflows to be reset when the reset-point chosen is **outside** the child init & completed events. In cases where reset-point is after child complete event, workers with Go SDK run into non-det error while other SDKs seems to simply reapply events (child Init, Start & complete) without any issues. In that regard this PR is more of a bug fix in Go-SDK.

> 4. Can you confirm `OriginalRunID` always matches `RunID` at this point of code today? Meaning since resets were never allowed over child workflows, it is impossible to have these two values different at child workflow creation time today?

Yes it is possible to have these two values different at child creation time. When the reset-point is **outside** the child's lifespan (Init - Completed) then the server has always allowed resets. So in cases when the reset-point is before the child-init, the parent will try to create a child again and this time `OriginalRunID` is different from `RunID`.

cretz:
@gow - thanks for this info! So reading all of that I am seeing two things:

1. We need to fix Go to be at least a little like other SDKs. We have two options:
    1. Start offering a re-seeded random to users and we use it for child IDs in Go with an SDK flag to only apply to new workflows
    2. Make available (maybe only to us) a "current run ID" that is updated when reset that is used for child workflow IDs (no SDK flag needed for compatibility)
3. We are not concerned with backwards compatibility of only changing Go since it never worked anyways after reset, so long as the non-reset logic remains working

For 1, I will start an internal discussion and invite you. Regardless, I am not sure the desired outcome is what is in this PR.

gow:
The desired outcome of this PR is for Go-SDK to be able to replay `StartChildWorkflowExecutionInitiated` events like other SDKs. Currently it fails. Using `OriginalRunID` will fix this issue and unblock reset feature development. But I understand it doesn't bring Go-SDK to be fully be at parity with other SDKs.

gow:
Putting this on hold. Will discuss internally and work on a better fix.

gow:
After an internal discussion, we decided to bring Go SDK on par with the rest of the SDKs. So instead of simply using OriginalRunID, we are using the `FirstExecutionRunId` as the initial seed and updating it with the currentRunID whenever we encounter a reset.

internal/internal_event_handlers.go
```diff
@@ -1220,6 +1220,11 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	case enumspb.EVENT_TYPE_WORKFLOW_TASK_TIMED_OUT:
 		// No Operation
 	case enumspb.EVENT_TYPE_WORKFLOW_TASK_FAILED:
+		// update the childWorkflowIDSeed if the workflow was reset at this point.
+		attr := event.GetWorkflowTaskFailedEventAttributes()
+		if attr.GetCause() == enumspb.WORKFLOW_TASK_FAILED_CAUSE_RESET_WORKFLOW {
+			weh.workflowInfo.childWorkflowIDSeed = attr.GetNewRunId()
+		}
 		// No Operation
```
yuandrew, line 1228: stale comment

internal/internal_task_handlers_interfaces_test.go
```diff
@@ -176,7 +176,7 @@ func (s *PollLayerInterfacesTestSuite) TestGetNextCommands() {
 		createTestEventWorkflowTaskStarted(3),
 		{
 			EventId:   4,
-			EventType: enumspb.EVENT_TYPE_WORKFLOW_TASK_FAILED,
+			EventType: enumspb.EVENT_TYPE_WORKFLOW_TASK_TIMED_OUT,
```
yuandrew, line 179: why was this changed from failed to timed_out?

yuandrew (commented):
overall LGTM, deferring to @Quinn-With-Two-Ns  or @cretz  for approval. I vaguely remember there was a test they were hoping was added, but can't remember exactly what that test was.

internal/internal_task_handlers_interfaces_test.go
gow: This is because we are now not skipping workflow task failed events in `history.prepareTask()` ([internal/internal_task_handlers.go](https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_handlers.go#L496)) so that they can be replayed. Some of the existing tests relied on this event being skipped. So I changed them to use other events that are still being skipped.

internal/internal_event_handlers.go
```diff
@@ -1457,6 +1462,11 @@ func (weh *workflowExecutionEventHandlerImpl) handleWorkflowExecutionStarted(
 	// as the value varies by WFT)
 	weh.sdkFlags.tryUse(SDKFlagProtocolMessageCommand, !weh.isReplay)
 
+	// Use the first execution run ID from the start event as the initial seed.
+	// First execution run ID stays the same for the entire chain of workflow resets.
+	// This helps us keep child workflow IDs consistent up until the next reset point.
+	weh.workflowInfo.childWorkflowIDSeed = attributes.GetFirstExecutionRunId()
```
gow, line 1468: Can FirstExecutionRunID be empty ever? The current version of the server is guaranteed to include it in the start event. But could there be some long running WFs with this field being empty in start event?
Asking to see if we want to be defensive here and fall back to RunID if FirstExecutionRunID is empty.

gow (commented):
> I vaguely remember there was a test they were hoping was added, but can't remember exactly what that test was.

We need to add some integration tests here. I'm trying to figure out the structuring and setup of them in this codebase. I'll probably sync with you all offline to get some pointers.

internal/internal_event_handlers.go
cretz: We need to set `childWorkflowIDSeed` on `WorkflowInfo` in the same place where we set the `WorkflowExecution` (`createWorkflowContext` which has the started event).

I think whether `attributes.FirstExecutionRunID` is always present is a server-side question. But I do not think it is accurate to use here because it is carried over in other situations like cron and workflow retry. You may need to confirm with Java and/or Core on what they use as the first run ID here. It may need to be `OriginalExecutionRunId` and you may need to fall back to the task's workflow execution run ID if that's not present.

Quinn-With-Two-Ns: Java uses `original_execution_run_id`
```
    // This is the run id when the WorkflowExecutionStarted event was written.
    // A workflow reset changes the execution run_id, but preserves this field.
    string original_execution_run_id = 14;
```

gow: ok. Using `original_execution_run_id` and moved the initialization to `createWorkflowContext`

cretz (commented):
LGTM, though now we get into the more fun challenge of adding tests, heh

test/replaytests/replay_test.go
```diff
@@ -493,6 +493,24 @@ func (s *replayTestSuite) TestPartialReplayNonCommandEvent() {
 	require.NoError(s.T(), err)
 }
 
+func (s *replayTestSuite) TestResetWorkflowWithChildSuccess() {
```
cretz, line 496: These don't seem to have resets in them. Here's my best guess on two main tests needed:

* Capture a JSON of a before-this-PR-change workflow that completed that has a reset after child workflow, and confirm it replays properly
* Capture a JSON of a before-this-PR-change workflow that completed that has a before-this-PR-change reset before child workflow, and confirm it replays properly because it created a whole new child workflow
* Make workflow with two child workflows, run it to completion, reset to in-between those two child workflows and let that reset one run to completion too. Confirm both replay successfully. Confirm that the original workflow's first child ID (i.e. the pre-reset-point one) is the same as the new workflow's first child ID by looking at history. Confirm the original workflow's second child ID (i.e. the post-reset-point one) is _not_ the same as the new workflow's second child ID.

Those are the primary three I can think of, @Quinn-With-Two-Ns may have more. And you may want more to test certain behaviors such as the fact that the default parent close policy of terminate does what is expected.

Quinn-With-Two-Ns: I think Chads point here was even though it doesn't replay properly with the current SDK, your change should cause it to replay properly

Quinn-With-Two-Ns:
Also this `TestNumPollersCounter ` test failure is not your fault. Seems to be very flaky after Go 1.24

test/replaytests/replay_test.go
gow: Agree. I've added both cases.
Adding integration tests now.

Quinn-With-Two-Ns: So you should be able to add 3 test since you can test it

gow: The third case he mentioned above is an integration tests isn't it?

Quinn-With-Two-Ns: It would be an integration test and a replay test. You have two scenarios:

1. Workflow was started with an old SDK, workflow had a child workflow and was reset after the child workflow finished. Currently that workflow is stuck due to NDE. Your change here should fix that workflow

2. Workflow started with a new SDK, workflow had a child workflow and was reset after the child workflow finished. Workflow should complete successfully

Quinn-With-Two-Ns: 1. is a replay test, 2 is an integration test

test/integration_test.go
```diff
@@ -2040,6 +2040,101 @@ func (ts *IntegrationTestSuite) TestResetWorkflowExecution() {
 	ts.Equal(originalResult, newResult)
 }
 
+// TestResetWorkflowExecutionWithChildren tests the behavior of child workflow ID generation when a workflow with children is reset.
+// It repeatedly resets the workflow at different points in its execution and verifies that the child workflow IDs are generated correctly.
+func (ts *IntegrationTestSuite) TestResetWorkflowExecutionWithChildren() {
```
cretz, line 2045: This is a great test

cretz (approved):
LGTM, but wait for @Quinn-With-Two-Ns to be able to look at test coverage and scenarios, he may be aware of more/better ones

internal/workflow.go
```diff
@@ -1281,6 +1281,8 @@ type WorkflowInfo struct {
 	continueAsNewSuggested bool
 	currentHistorySize     int
 	currentHistoryLength   int
+	// childWorkflowIDSeed is used to generate unique child workflow IDs
+	childWorkflowIDSeed string
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I would prefer this is named by what it is, not how it is just currently used

test/replaytests/replay_test.go
Quinn-With-Two-Ns: @gow one test I would like to see with this PR is a workflow with multiple resets. I think we should have a test that has 3 child workflows and between each we do a reset.

internal/workflow.go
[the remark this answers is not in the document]
gow: I had initially named it currentRunID but thought that was not accurate since it's always going to be originalRunID except when a WF is reset (rare). Since this variable represents different possible values I named it based on its usage which is not great either. Will rename it to `currentRunID`

test/replaytests/replay_test.go
gow: @Quinn-With-Two-Ns how would it be different from what I added now? Would it cover any additional cases? Maybe I missed something.

Quinn-With-Two-Ns: It covers the `currentRunID` changing multiple times in the same workflow history

gow: Got it. Will modify that test to reset the resulting runs.

gow: Done. Modified it to start 3 children and reset the latest run each time at different points. Total 4 resets resulting in 5 different runs.

Quinn-With-Two-Ns:
@gow One of the tests you added `TestResetWorkflowExecutionWithChildren` is failing https://github.com/temporalio/sdk-go/actions/runs/13406853712/job/37448243290?pr=1803

gow:
> @gow One of the tests you added `TestResetWorkflowExecutionWithChildren` is failing https://github.com/temporalio/sdk-go/actions/runs/13406853712/job/37448243290?pr=1803

hmmm... the reset points I've chosen are static based on what I observed locally. It fails when there are additional events added due to timeouts/retries. Let me try to pick the correct reset-point dynamically by analyzing the history.

Quinn-With-Two-Ns (approved):

Issue #1779, opened 2025-01-15
Special behavior for Temporal built-in prefixes


cretz:
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/576

PR #1806, opened 2025-02-07
Reserve "__temporal_" prefix 

yuandrew:
## What was changed

Prevent users from using `__temporal_` across workflows, activities, nexus operations/services, and signals.

Queries and updates already reserve the `__` prefix 

## Why?

We want to reserve a prefix to throw any special behavior under.

## Checklist
 

1. Closes  #1779 

2. How was this tested:

Added tests

3. Any docs updates needed?


cretz (approved):
LGTM, may want @Quinn-With-Two-Ns to peek too

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Checking the [issue](https://github.com/temporalio/features/issues/576) again are we also supposed to reserve task queue as well? @cretz @yuandrew

PR #1818, opened 2025-02-11
Add note that SetOnActivityHeartbeatListener may be called concurrently

### LINKED ISSUE #908: Heartbeats in test environment are not concurrency-safe
## Expected Behavior
Expect multiple goroutines of an activity can heartbeat in the test framework

## Actual Behavior
Race detector says this is clashing. Specifically at https://github.com/temporalio/sdk-go/blob/c1ce1d25238b7bb15095f0767497fa93215c2589/internal/internal_workflow_testsuite.go#L321 which calls getActivityHandle, that is not concurrency safe.

This was commented on before but never fixed, see https://github.com/temporalio/sdk-go/pull/780/files#r852206758

yuandrew:
## What was changed

Added a note that SetOnActivityHeartbeatListener may be called concurrently

## Why?

Users have noted that this can occasionally hit a data race issue. Wasn't able to repro, so adding a note of this to the SDK for now.

## Checklist

1. Closes #908  

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -281,6 +281,7 @@ func (t *TestActivityEnvironment) SetWorkerStopChannel(c chan struct{}) {
 // SetOnActivityHeartbeatListener sets a listener that will be called when
 // activity heartbeat is called. ActivityInfo is defined in internal package,
 // use public type activity.Info instead.
+// Note: SetOnActivityHeartbeatListener may be called concurrently.
 //
 // Note: Due to internal caching by the activity system, this may not get called
 // for every heartbeat recorded. This is only called when the heartbeat would be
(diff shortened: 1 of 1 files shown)


internal/workflow_testsuite.go
```diff
@@ -281,6 +281,7 @@ func (t *TestActivityEnvironment) SetWorkerStopChannel(c chan struct{}) {
 // SetOnActivityHeartbeatListener sets a listener that will be called when
 // activity heartbeat is called. ActivityInfo is defined in internal package,
 // use public type activity.Info instead.
+// Note: SetOnActivityHeartbeatListener may be called concurrently.
```
cretz, line 284:
```suggestion
// Note: The provided listener may be called concurrently.
```

To clarify, we are saying we may call the listener callback concurrently, not related to when `SetOnActivityHeartbeatListener` is called

cretz (approved):

PR #1822, opened 2025-02-13
Properly convert Nexus HandlerErrors in the test env

bergundy:
## What was changed

Use similar logic as the real server to convert Nexus HandlerErrors.

## Why?

Make the test env behave as close as possible to the real server.

### DIFF
--- a/internal/nexus_operations.go
+++ b/internal/nexus_operations.go
@@ -174,6 +174,29 @@ func apiOperationErrorToNexusOperationError(opErr *nexuspb.UnsuccessfulOperation
 	}
 }
 
+func apiHandlerErrorToNexusHandlerError(apiErr *nexuspb.HandlerError, failureConverter converter.FailureConverter) (*nexus.HandlerError, error) {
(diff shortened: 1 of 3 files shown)


internal/nexus_operations.go
```diff
@@ -174,6 +174,29 @@ func apiOperationErrorToNexusOperationError(opErr *nexuspb.UnsuccessfulOperation
 	}
 }
 
+func apiHandlerErrorToNexusHandlerError(apiErr *nexuspb.HandlerError, failureConverter converter.FailureConverter) (*nexus.HandlerError, error) {
```
rodrigozhou, line 177: Add a comment that this is a copy from the server code.
I wonder if it's possible to share this code.

rodrigozhou (approved):

Quinn-With-Two-Ns (approved):

internal/nexus_operations.go
bergundy: The whole block of functions is surrounded by a comment that says it's copied from the server code.
I'm hesitant to share to avoid exporting these functions from the SDK.

PR #1824, opened 2025-02-13
Catchup tasks for Nexus features in the test env

bergundy:
- Propagate operation timeout to the handler via header.
- Handle operation complete-before-start.

### DIFF
--- a/internal/nexus_operations.go
+++ b/internal/nexus_operations.go
@@ -185,7 +185,7 @@ func apiHandlerErrorToNexusHandlerError(apiErr *nexuspb.HandlerError, failureCon
 	}
 
 	nexusErr := &nexus.HandlerError{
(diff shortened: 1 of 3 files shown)


internal/nexus_operations.go
```diff
@@ -174,6 +174,29 @@ func apiOperationErrorToNexusOperationError(opErr *nexuspb.UnsuccessfulOperation
 	}
 }
 
+func apiHandlerErrorToNexusHandlerError(apiErr *nexuspb.HandlerError, failureConverter converter.FailureConverter) (*nexus.HandlerError, error) {
```
bergundy, line 177: This is a leftover from #1822, I plan on merging that first.

Quinn-With-Two-Ns (approved):

PR #1826, opened 2025-02-14
temporarily disable TestNumPollersCounter while debugging flakiness

yuandrew:
## What was changed

temporarily disable TestNumPollersCounter

## Why?

test flake making it very difficult to pass CI. I am looking into the cause for this flake, and will uncomment once a fix is found

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):
Thanks!

cretz (approved):
Ideally we would remove instead of commenting out, but this is ok since we are actively working on fixing and so may not live long commented out anyways.

PR #1828, opened 2025-02-14
Support conflict options for starting Nexus operations in test framework

rodrigozhou:
## What was changed

Support OnConflictOptions when starting Nexus operation in test framework.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/internal_workflow.go
```diff
@@ -216,6 +216,7 @@ type (
 		WaitForCancellation      bool
 		WorkflowIDReusePolicy    enumspb.WorkflowIdReusePolicy
 		WorkflowIDConflictPolicy enumspb.WorkflowIdConflictPolicy
+		OnConflictOptions        *OnConflictOptions
```
rodrigozhou, line 219: I'm adding this `WorkflowOptions`, but not sure if I should. I need because the Nexus handler workflow is executed in the test framework as a child workflow, and I need this to resolve workflow ID conflict, ie., starting the same operation multiple times.

I can see this is used in certain operations. For example, I saw `WorkflowOptions` is built for `ExecuteChildWorkflow`, and `OnConflictOptions` is not an options for executing child workflow.

Btw, `WorkflowIDConflictPolicy` was added here a while ago, but I didn't see any usage of this field anywhere... I'm using it now here though.

cc: @Quinn-With-Two-Ns @cretz

Quinn-With-Two-Ns: `WorkflowIDConflictPolicy` is not supported for child workflows https://github.com/temporalio/temporal/issues/6799 so yeah it probably shouldn't have been added here. 

Is it reasonable to refactor the Nexus handler workflow to not be a "child" in the test framework?

cretz: Agreed, we shouldn't add an unused field here (even if there happens to be another unused field here)

Quinn-With-Two-Ns: I think Rodrigo is saying it is no longer an unused field, but I guess it is unused when executing outside the test enviorment?

rodrigozhou: Yes, `WorkflowIDConflictPolicy` is unused since it was added in https://github.com/temporalio/sdk-go/pull/1563, but I need it now with this PR.

As for not being a child workflow, I'm not sure how we would be able to run the Nexus workflow without it since the test environment only allows to run one main/root workflow at a time. cc: @bergundy

cretz: Oh, my mistake, I thought this was user-facing child workflow options or something. If this is an `internal` only thing I have no opinion/preference, will defer to Quinn.

rodrigozhou: `WorkflowOptions` is somewhat user facing, isn't it? The user could get it from the workflow context.
The two fields I mentioned, it's only used in the test environment. Maybe I could just make them unexported?

Quinn-With-Two-Ns: How can users get `WorkflowOptions` from the workflow context?

rodrigozhou: Well, it's not exported, but technically, the user could write directly `ctx.Value("wfEnvOptions")`.
Otherwise, it seems it's only for `internal` only.

Quinn-With-Two-Ns: Yeah I think the only reason the type is even exported anyway is for the nexus test environment and maybe the PHP SDK

internal/internal_workflow_testsuite.go
```diff
@@ -2686,8 +2701,9 @@ func (env *testWorkflowEnvironmentImpl) resolveNexusOperation(seq int64, token s
 			panic(fmt.Errorf("no running operation found for sequence: %d", seq))
 		}
 		if err != nil {
+			err = convertKnownErrors(err)
 			failure := env.failureConverter.ErrorToFailure(err)
-			err = env.failureConverter.FailureToError(nexusOperationFailure(handle.params, handle.operationToken, failure.GetCause()))
+			err = env.failureConverter.FailureToError(nexusOperationFailure(handle.params, handle.operationToken, failure))
```
rodrigozhou, line 2706: @bergundy I think this was missed in https://github.com/temporalio/sdk-go/pull/1822

test/nexus_test.go
```diff
@@ -1264,6 +1265,137 @@ func TestWorkflowTestSuite_WorkflowRunOperation_WithCancel(t *testing.T) {
 	}
 }
 
+func TestWorkflowTestSuite_WorkflowRunOperation_MultipleCallers_NoAttachCallback(t *testing.T) {
```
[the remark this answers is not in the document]
rodrigozhou: There was a bug in the Nexus error handling that caused the issue. I fixed in this PR.

cretz (approved):

internal/internal_workflow.go
```diff
@@ -215,6 +215,7 @@ type (
...
```
Quinn-With-Two-Ns, line 218: nit: can we add a comment explaining these are only used for the test environment?

internal/internal_workflow_testsuite.go
bergundy: I don't think we need to convert known errors here, this is just used when a handler workflow completes and it should already include a properly structured failure error.
I think the reason I use `GetCause()` here is that the workflow error was wrapping the failure and I didn't want that to be part of the cause chain to mimic the server behavior.
Arguably unwrapping should be done where this function is called though.
Worth double checking this.

rodrigozhou: This is also called when the workflow failed to start.
When executing a child workflow, there's a `callback` and a `startedHandler`, and they are both called in case the child workflow fails to start.

bergundy: So the unwrapping should probably be done at the call site.

Quinn-With-Two-Ns:
@rodrigozhou Now that we have the CLI release can we add integration tests as part of this PR ?

PR #1830, opened 2025-02-18
Remove operation and service labels from Nexus schedule to start metric

bergundy:
The labels are not relevant when polling from the Nexus task queue where any request (task) may be served.


Quinn-With-Two-Ns (commented):
Hopefully some test would need to be updated based on the tags changing?

bergundy:
Just fixed the tests.

Quinn-With-Two-Ns (approved):

Issue #1778, opened 2025-01-14
Rename Worker Deployment API interfaces


antlai-temporal:
Some interfaces of the Worker Deployment API are changing to reflect the embedding of the Rollout API within a Deployment.

Issue #1777, opened 2025-01-14
Add ramp to Worker Deployments


antlai-temporal:
Add support for the ramp API to Worker Deployments. A ramp gracefully transitions Workflows and
Activity tasks between two implementations.

PR #1832, opened 2025-02-18
Support Worker Deployments 3.1

antlai-temporal:
## What was changed

This PR merges the previously reviewed https://github.com/temporalio/sdk-go/pull/1814 into main and activates 
system test with the dev-server 1.27.0-128.4

It introduces Worker Deployments using the versioning 3.1 API

## Checklist

1. Closes 
#1778 #1777 

2. How was this tested:

Both system and unit tests


.github/workflows/ci.yml
```diff
@@ -87,6 +87,10 @@ jobs:
       - name: Docker compose - integration tests
         if: ${{ matrix.testDockerCompose }}
         run: go run . integration-test
+        env:
+          # TODO(antlai-temporal): Remove this flag once server 1.27 released.
+          DISABLE_WORKER_DEPLOYMENT_TESTS: "1"
```
Quinn-With-Two-Ns, line 92: Why do we have two flags? My suggestion in the last PR was to just rename this one because it will be used by more then just worker deployment tests.

cretz (approved):
(need @Quinn-With-Two-Ns's approval too before merge)

contrib/opentracing/go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/robfig/cron v1.2.0 // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
 	go.temporal.io/api v1.44.0 // indirect
+	golang.org/x/exp v0.0.0-20231127185646-65229373498e // indirect
```
Quinn-With-Two-Ns, line 27: why did this change?

.github/workflows/ci.yml
antlai-temporal: I misunderstood, I though there were already other tests using that flag and you wanted both. Changing to one with your suggestion...

contrib/opentracing/go.mod
antlai-temporal: I need to `go mod tidy` again...  A few dependencies may be upgrade, I hope that's fine...

Quinn-With-Two-Ns: I am confused because I don't see any other dependencies upgraded, the ones in `cmd/build` wouldn't effect this package. Can you double check if these are actually needed?

Quinn-With-Two-Ns (approved):

contrib/opentracing/go.mod
antlai-temporal: I think that package holds experimental and deprecated packages, and when we upgraded other packages they no longer need it... The problem is that I did a conflict merge on the `go.mod` and forgot to go mod tidy...

antlai-temporal: I see what the problem is for  `internal/cmd/build/go.mod`. I accidentally downgraded `github.com/grpc-ecosystem/grpc-gateway/v2 v2.26.1` in the merge, and that requires the exp package...

antlai-temporal: The problem is that in main we are importing api v1.44.0 at the top level, but inside internal/cmd/build/go.mod we have api v1.44.1.  My PR was using 1.44.0 everywhere and that's why the go.mod was different. 
I'm upgrading everywhere to v1.44.1, if that's ok...

Quinn-With-Two-Ns: >I'm upgrading everywhere to 1.44.1 so they get the fix...

Yes please do

internal/internal_worker.go
```diff
@@ -1746,8 +1757,9 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 		MaxConcurrentNexusTaskQueuePollers:    options.MaxConcurrentNexusTaskPollers,
 		Identity:                              client.identity,
 		WorkerBuildID:                         options.BuildID,
-		UseBuildIDForVersioning:               options.UseBuildIDForVersioning,
+		UseBuildIDForVersioning:               options.UseBuildIDForVersioning || options.DeploymentOptions.UseVersioning,
 		DeploymentSeriesName:                  options.DeploymentOptions.DeploymentSeriesName,
+		WorkerDeploymentVersion:               options.DeploymentOptions.Version.String(),
```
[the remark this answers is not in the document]
antlai-temporal: Reverting, further discussion needed...

internal/worker.go
```diff
@@ -339,3 +379,66 @@ func NewWorker(
...
+// by parsing a string with format `YourDeploymentName.YourBuildID`.
+// The separator "." is a reserved character that cannot be part of a
+// Deployment Name.
+//
+// NOTE: Experimental
+//
+// Exposed as: [go.temporal.io/sdk/worker.NewDeploymentVersionFromString]
+func NewWorkerDeploymentVersionFromString(version string) WorkerDeploymentVersion {
+	splitVersion := strings.SplitN(version, ".", 2)
+	if len(splitVersion) != 2 {
```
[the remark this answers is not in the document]
antlai-temporal: Ditto, reverting last commit...

PR #1833, opened 2025-02-18
Translate relevant errors to non retryable internal Nexus handler errors

bergundy:
Temporarily using the pre-release CLI for testing.

### DIFF
--- a/internal/cmd/build/main.go
+++ b/internal/cmd/build/main.go
@@ -188,6 +188,7 @@ func (b *builder) integrationTest() error {
 	cmd := b.cmdFromRoot(args...)
 	cmd.Dir = filepath.Join(cmd.Dir, "test")
 	cmd.Env = append(os.Environ(), "DISABLE_SERVER_1_25_TESTS=1")
+	cmd.Env = append(os.Environ(), "ENABLE_SERVER_1_27_TESTS=1")
 	if err := b.runCmd(cmd); err != nil {
 		return fmt.Errorf("integration test failed: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


test/nexus_test.go
```diff
@@ -302,7 +302,8 @@ func TestNexusSyncOperation(t *testing.T) {
 		_, err := nexus.ExecuteOperation(ctx, nc, syncOp, "already-started", nexus.ExecuteOperationOptions{})
 		var handlerErr *nexus.HandlerError
 		require.ErrorAs(t, err, &handlerErr)
-		require.Equal(t, nexus.HandlerErrorTypeBadRequest, handlerErr.Type)
```
Quinn-With-Two-Ns, line 305: Don't you need to flag these if you are not using the prerelease test server?

bergundy: Is this for the cloud tests? I see that the integration tests are passing (apart from flakes).

rodrigozhou (approved):

test/nexus_test.go
bergundy: Saw the CI failures, added a flag to skip.

internal/cmd/build/main.go
```diff
@@ -187,6 +188,7 @@ func (b *builder) integrationTest() error {
 	cmd := b.cmdFromRoot(args...)
 	cmd.Dir = filepath.Join(cmd.Dir, "test")
 	cmd.Env = append(os.Environ(), "DISABLE_SERVER_1_25_TESTS=1")
+	cmd.Env = append(os.Environ(), "ENABLE_SERVER_1_27_TESTS=1")
```
Quinn-With-Two-Ns, line 191: This PR https://github.com/temporalio/sdk-go/pull/1832/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR92 also adds a flag for the same purpose can we standardized on this approach?

bergundy: I will go with @antlai-temporal's approach then.

Quinn-With-Two-Ns (approved):
Approving assuming we change the test flag approach

PR #1837, opened 2025-02-20
More user-friendly NDE error message

dandavison:
**before**
> [TMPRL1100] unknown command CommandType: Timer, ID: 8, possible causes are nondeterministic workflow definition code or incompatible change in the workflow definition

**after**

> [TMPRL1100] During replay, a matching Timer command was expected in history event position 8. However, the replayed code did not produce that. Possible causes are nondeterministic workflow definition code, or an incompatible change in the workflow definition.

### DIFF
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -1075,8 +1075,8 @@ func (h *commandsHelper) getCommand(id commandID) commandStateMachine {
 	command, ok := h.commands[id]
 	if !ok {
 		panicMsg := fmt.Sprintf(
-			"[TMPRL1100] During replay, workflow history (event ID %s) implies that a %v command should have been emitted by the replayed code, but no such command was emitted. "+
-				"Possible causes are nondeterministic workflow definition code, or an incompatible change in the workflow definition.", id.id, id.commandType)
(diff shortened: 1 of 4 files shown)


internal/internal_command_state_machine.go
```diff
@@ -1074,8 +1074,9 @@ func (h *commandsHelper) incrementNextCommandEventIDIfVersionMarker() {
 func (h *commandsHelper) getCommand(id commandID) commandStateMachine {
 	command, ok := h.commands[id]
 	if !ok {
-		panicMsg := fmt.Sprintf("[TMPRL1100] unknown command %v, possible causes are nondeterministic workflow definition code"+
-			" or incompatible change in the workflow definition", id)
+		panicMsg := fmt.Sprintf(
+			"[TMPRL1100] During replay, a matching %v command was expected in history event position %s. However, the replayed code did not produce that. "+
+				"Possible causes are nondeterministic workflow definition code, or an incompatible change in the workflow definition.", id.commandType, id.id)
```
Sushisource, line 1079: Core also says what the command that was produced was too, but, honestly we can just wait to do the more structured thing.

Sushisource (approved):

Quinn-With-Two-Ns (approved):

PR #1841, opened 2025-02-21
Nexus interceptors

bergundy:
## What was changed

Added support for Nexus operation inbound and outbound interceptors. The implementation was a bit awkward and required a couple of adapters from and to the Nexus SDK middleware interfaces but that seems reasonable if we want to leverage shared code from that SDK.

## Why?

Support tracing and other use cases.

## Checklist

- [x] Upgrade to a stable Nexus SDK
- [x] Figure out if we want to also intercept logs and metrics emitted by the SDK outside of the "handler" scope - won't do
- [ ] Add tracing interceptor support - will do in a followup PR


interceptor/interceptor.go
```diff
@@ -222,6 +222,47 @@ type ClientUpdateWorkflowInput = internal.ClientUpdateWorkflowInput
 // ClientOutboundInterceptor.UpdateWithStartWorkflow.
 type ClientUpdateWithStartWorkflowInput = internal.ClientUpdateWithStartWorkflowInput
 
+// NexusOperationInboundInterceptor is an interface for intercepting Nexus operation method invocations.
+//
+// All implementations must embed [NexusOperationInboundInterceptorBase] to safely handle future changes.
+//
+// Note: Experimental
```
bergundy, line 229: I marked interceptors as experimental at this stage. Ideally we would have at least one release to let this bake. I don't think this is a blocker for GA.

Quinn-With-Two-Ns (approved):

PR #1844, opened 2025-02-22
Add Nexus tracing interceptor

bergundy:


internal/interceptortest/tracing.go
```diff
@@ -151,6 +164,8 @@ func AssertSpanPropagation(t *testing.T, tracer TestTracer) {
 				Span(tracer.SpanName(&interceptor.TracerStartSpanOptions{Operation: "RunActivity", Name: "testActivity"}))),
 			Span(tracer.SpanName(&interceptor.TracerStartSpanOptions{Operation: "StartActivity", Name: "testActivityLocal"}),
 				Span(tracer.SpanName(&interceptor.TracerStartSpanOptions{Operation: "RunActivity", Name: "testActivityLocal"})))),
+		// This is the workflow that gets started from the nexus workflow run operation.
+		Span(tracer.SpanName(&interceptor.TracerStartSpanOptions{Operation: "RunWorkflow", Name: "testWaitForCancelWorkflow"})),
```
bergundy, line 168: I figure it would have been better to test this with the client but that's not part of our testing setup here. We can consider propagating the parent child relationship in the test env too.

bergundy: Hmmm... I see there's also a tracing integration test, I will add to that.

interceptor/tracing_interceptor.go
```diff
@@ -27,11 +27,13 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/nexus-rpc/sdk-go/nexus"
 	commonpb "go.temporal.io/api/common/v1"
 
 	"go.temporal.io/sdk/activity"
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/internal"
```
Quinn-With-Two-Ns, line 36: The tracing_interceptor shouldn't need to depend on internal

interceptor/tracing_interceptor.go
```diff
@@ -886,14 +1010,38 @@ func (t *tracingInterceptor) writeSpanToHeader(span TracerSpan, header map[strin
...
+		header.Set(k, v)
+	}
+	return nil
+}
+
+func (t *tracingInterceptor) readSpanFromNexusHeader(header nexus.Header) (TracerSpanRef, error) {
+	return t.tracer.UnmarshalSpan(header)
+}
+
+type anyWorkflowFuture interface {
```
Quinn-With-Two-Ns, line 1030: I wouldn't do this approach, I would create a different type to wrap the `NexusOperationFuture` and the `ChildWorkflowFuture`

interceptor/tracing_interceptor.go
bergundy: Ah this was autocomplete by the IDE, I removed the direct dependency.
Just note that it doesn't matter since the interceptor package does depend on internal, it aliases the internal types for export.

interceptor/tracing_interceptor.go
bergundy: Do you think this matters enough to block the PR? Seems like the approaches are equivalent and the one I took doesn't require changing any existing code.

interceptor/tracing_interceptor.go
Quinn-With-Two-Ns: The tracing interceptor should depend on the aliased types, there are still references to the internal types here

interceptor/tracing_interceptor.go
```diff
@@ -847,14 +931,54 @@ func (t *tracingInterceptor) startSpan(
...
+
+func (t *tracingInterceptor) nexusHeaderReader(header nexus.Header) func() (TracerSpanRef, error) {
+	return func() (TracerSpanRef, error) {
+		return t.readSpanFromNexusHeader(header)
+	}
+}
+
+func (t *tracingInterceptor) nexusHeaderWriter(header nexus.Header) func(TracerSpan) error {
+	return func(span TracerSpan) error {
+		return t.writeSpanToNexusHeader(span, header)
```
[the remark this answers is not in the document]
bergundy: Yup, we did the same in Java and this would be compatible with tracing header serialization for HTTP (which is how we transmit these headers over the wire).

interceptor/tracing_interceptor.go
Quinn-With-Two-Ns: I don't think It is correct to return a type implementing multiple different future interfaces from here , is is visible to users using reflection

interceptor/tracing_interceptor.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Will the headers keys and values actually by HTTP compatible ? In Java I had to explicitly opt into HTTP-header-compatible Strings. I don't think the interceptor implemetation in Go are.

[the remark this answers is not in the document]
bergundy: Looks like we're using the text map propagator for this which is how otel HTTP instrumentation works:

```go
func (t *tracer) MarshalSpan(span interceptor.TracerSpan) (map[string]string, error) {
	data := textMapCarrier{}
	tSpan := span.(*tracerSpan)
	ctx := context.Background()
	if !t.options.DisableBaggage {
		ctx = baggage.ContextWithBaggage(ctx, tSpan.Baggage)
	}
	t.options.TextMapPropagator.Inject(trace.ContextWithSpan(ctx, tSpan.Span), data)
	return data, nil
}
```

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/1ae01635b4bd869d7a985b7de063e0bca1b04317/instrumentation/net/http/otelhttp/transport.go#L129
https://github.com/open-telemetry/opentelemetry-go/blob/8f4a5c69cefc94048e7e3bfe262ff81a9d1bce55/propagation/propagation.go#L59-L78

interceptor/tracing_interceptor.go
bergundy: I can refactor if you think this is critical.

interceptor/tracing_interceptor.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Looks like the open tracing interceptor won't as we use `TextMap` https://github.com/temporalio/sdk-go/blob/master/contrib/opentracing/interceptor.go#L122 instead of `HTTPHeaders`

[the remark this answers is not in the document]
Quinn-With-Two-Ns: I think that is fine though as long as open-telemetry works , would like @cretz opinion

interceptor/tracing_interceptor.go
```diff
@@ -886,16 +1009,47 @@ func (t *tracingInterceptor) writeSpanToHeader(span TracerSpan, header map[strin
...
+	return nil
+}
+
+func (t *tracingInterceptor) readSpanFromNexusHeader(header nexus.Header) (TracerSpanRef, error) {
+	return t.tracer.UnmarshalSpan(header)
+}
+
+type errFut struct{ inner workflow.Future }
+
+func newErrFut(ctx workflow.Context, err error) *errFut {
```
Quinn-With-Two-Ns, line 1031: To be clear my suggestion was keep `newErrFut` returning a type that implements `workflow.ChildWorkflowFuture` and add a new function `newNexusErrFut` that returns a `nexusOperationFuture`

Quinn-With-Two-Ns: can also rename `newErrFut ` to something like `newWorkflowErrFut` , but I have no strong opinion there

interceptor/tracing_interceptor.go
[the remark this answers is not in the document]
bergundy: I'm seeing integration test failures with otel right now, let me figure out why this isn't working as expected and also see what happens with opentracing.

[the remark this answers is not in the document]
bergundy: > Looks like the open tracing interceptor won't as we use `TextMap` https://github.com/temporalio/sdk-go/blob/master/contrib/opentracing/interceptor.go#L122 instead of `HTTPHeaders`

It doesn't seem like there's a fundamental different except for how keys are normalized in the HTTP header map.

interceptor/tracing_interceptor.go
bergundy: I simplified this a bit and I think what we have now is a bit cleaner.

interceptor/tracing_interceptor.go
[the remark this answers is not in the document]
bergundy: Had to allow `UnmarshalSpan` to fail since it expects to always construct a valid span. LMK if you have other opinions.

test/nexus_test.go
```diff
@@ -2322,3 +2343,50 @@ func TestInterceptors(t *testing.T) {
...
+	t.Cleanup(w.Stop)
+
+	run, err := tc.client.ExecuteWorkflow(ctx, client.StartWorkflowOptions{
+		TaskQueue: tc.taskQueue,
+		// The endpoint registry may take a bit to propagate to the history service, use a shorter workflow task
+		// timeout to speed up the attempts.
+		WorkflowTaskTimeout: time.Second,
+	}, wf)
+	require.NoError(t, err)
+	require.NoError(t, run.Get(ctx, nil))
```
bergundy, line 2391: Oops, forgot to add the span assertions.

interceptor/tracing_interceptor.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Sorry I don't understand the justification for ignoring the error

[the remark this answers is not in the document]
Quinn-With-Two-Ns: >It doesn't seem like there's a fundamental different except for how keys are normalized in the HTTP header map.

Is that not an important difference if the goal is compatibility with HTTP?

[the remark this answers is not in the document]
bergundy: No because Nexus headers are always lower case and we translate to canonical HTTP case when transmitting the HTTP request.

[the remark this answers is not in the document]
bergundy: The only issue I can think of is if the tracer expects to read back headers that are not in lower case form.
But I would also make sure that the headers the tracer adds are written to the nexus header in lower case just in case.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: I believe HTTP headers in Go are written in canonical form which is not lower case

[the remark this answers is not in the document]
Quinn-With-Two-Ns: again not critical if we are not GA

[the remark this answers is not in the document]
bergundy: Yup but that's taken care of in the Nexus SDK when we construct the HTTP request.

interceptor/tracing_interceptor.go
```diff
@@ -847,14 +930,54 @@ func (t *tracingInterceptor) startSpan(
...
-		if err := t.writeSpanToHeader(span, header); err != nil {
+	if options.ToHeader {
+		if err := headerWriter(span); err != nil {
 			return nil, err
 		}
 	}
 	return span, nil
 }
 
+func (t *tracingInterceptor) headerReader(ctx context.Context) func() (TracerSpanRef, error) {
```
cretz, line 941: Pedantic since it's not exported, but usually Go would prefer an interface to passing multiple functions around, e.g.

```go
type spanMarshaler interface {
	readSpan(*tracingInterceptor) (TracerSpanRef, error)
	writeSpan(*tracingInterceptor, TracerSpan) error
}

type payloadSpanMarshaler map[string]*commonpb.Payload
func (p payloadSpanMarshaler) readSpan(t *tracingInterceptor) (TracerSpanRef, error) {
	return t.readSpanFromHeader(p)
}
func (p payloadSpanMarshaler) writeSpan(t *tracingInterceptor, span TracerSpan) error {
	return t.writeSpanToHeader(span, p)
}

type stringSpanMarshaler map[string]string

// ...etc
```

May be able to just inline `readSpanFromHeader` and `writeSpanToHeader` too, but pedantic

interceptor/tracing_interceptor.go
```diff
@@ -788,20 +818,71 @@ func (t *tracingWorkflowOutboundInterceptor) startNonReplaySpan(
...
+	NexusOperationInboundInterceptorBase
+	root *tracingInterceptor
+}
+
+// CancelOperation implements internal.NexusOperationInboundInterceptor.
+func (t *tracingNexusOperationInboundInterceptor) CancelOperation(ctx context.Context, input NexusCancelOperationInput) error {
+	info := nexus.ExtractHandlerInfo(ctx)
+	// Start span reading from header
+	span, ctx, err := t.root.startSpanFromContext(ctx, &TracerStartSpanOptions{
+		Operation:  "RunCancelNexusOperationHandler",
```
cretz, line 838: There are all sorts of other name options we can bikeshed over here (`RunCancelNexusOperation`, `HandleNexusCancelOperation`, etc), but so long as consistent across SDKs, I don't have a strong preference.

interceptor/tracing_interceptor.go
```diff
@@ -788,20 +818,71 @@ func (t *tracingWorkflowOutboundInterceptor) startNonReplaySpan(
...
+	finishOpts.Error = err
+	return err
+}
+
+// StartOperation implements internal.NexusOperationInboundInterceptor.
+func (t *tracingNexusOperationInboundInterceptor) StartOperation(ctx context.Context, input NexusStartOperationInput) (nexus.HandlerStartOperationResult[any], error) {
+	info := nexus.ExtractHandlerInfo(ctx)
+	// Start span reading from header
+	span, ctx, err := t.root.startSpanFromContext(ctx, &TracerStartSpanOptions{
+		Operation:  "RunStartNexusOperationHandler",
```
cretz, line 860: There are all sorts of other name options we can bikeshed over here (`RunCancelNexusOperation`, `HandleNexusCancelOperation`, etc), but so long as consistent across SDKs, I don't have a strong preference. I do think `Start<whatever>` for outbound and `Run<whatever>` or `Handle<whatever>` for inbound may make more sense, but doesn't matter much.

interceptor/tracing_interceptor.go
```diff
@@ -886,16 +1009,41 @@ func (t *tracingInterceptor) writeSpanToHeader(span TracerSpan, header map[strin
...
+		header.Set(k, v)
+	}
+	return nil
+}
+
+func (t *tracingInterceptor) readSpanFromNexusHeader(header nexus.Header) (TracerSpanRef, error) {
+	// Ignore unmarshal errors, e.g. when an expected header isn't set.
+	// This works differently from non Nexus headers where the trace context is serialized entirely into a single
+	// payload and the presence of that payload implies a valid header.
+	span, _ := t.tracer.UnmarshalSpan(header)
```
cretz, line 1029: Per off-PR discussion, let's see if we can make `UnmarshalSpan` calls in the tracing impls return `nil` if the proper header is not present (already done for datadog, OpenTracing may have a certain error, and have to hardcode "traceparent" lookup for OTel)

Quinn-With-Two-Ns (approved):

PR #1845, opened 2025-02-25
Block using conflict policy UseExisting for Nexus WorkflowRunOperation

rodrigozhou:
## What was changed

Block using conflict policy `UseExisting` for Nexus `WorkflowRunOperation`.
Returning a non-retryable internal error at this moment.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalnexus/operation.go
+++ b/temporalnexus/operation.go
@@ -53,6 +53,8 @@ import (
 	"go.temporal.io/sdk/workflow"
 )
 
+var errConflictPolicyUseExistingNotSupported = errors.New("workflow ID conflict policy UseExisting is not supported for Nexus WorkflowRunOperation")
+
 // GetMetricsHandler returns a metrics handler to be used in a Nexus operation's context.
 func GetMetricsHandler(ctx context.Context) metrics.Handler {
 	return internal.GetNexusOperationMetricsHandler(ctx)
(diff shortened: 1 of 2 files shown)


test/nexus_test.go
```diff
@@ -1726,19 +1742,24 @@ func TestWorkflowTestSuite_WorkflowRunOperation_MultipleCallers(t *testing.T) {
...
 		{
 			input: "conflict-policy-use-existing",
-			checkOutput: func(t *testing.T, numCalls int, res CallerWfOutput) {
-				require.EqualValues(t, numCalls, res.CntOk)
+			checkOutput: func(t *testing.T, numCalls int, res CallerWfOutput, err error) {
+				// TODO(rodrigozhou): assert NotError and remove the comments below after UseExisting is
+				// unblocked.
+				require.ErrorContains(t, err, "workflow ID conflict policy UseExisting is not supported for Nexus WorkflowRunOperation")
+				// require.EqualValues(t, numCalls, res.CntOk)
+				// require.EqualValues(t, 0, res.CntErr)
```
bergundy, line 1762: I would just delete this instead of commenting it out.

temporalnexus/operation.go
```diff
@@ -53,6 +53,8 @@ import (
 	"go.temporal.io/sdk/workflow"
 )
 
+var errConflictPolicyUseExistingNotSupported = errors.New("workflow ID conflict policy UseExisting is not supported for Nexus WorkflowRunOperation")
```
bergundy, line 56: nit: you can just define this inline, no need for this variable.

bergundy (commented):
No blocking comments from me. Will defer to @Quinn-With-Two-Ns for an approval.

temporalnexus/operation.go
rodrigozhou: Isn't there a lint that complains about inline errors? Did we remove that?

Quinn-With-Two-Ns: Yeah no need if it isn't even exported. We have no lint error for inline errors

Quinn-With-Two-Ns (approved):

Issue #1846, opened 2025-02-25
AsTime() on a nil pointer converts to  Jan. 1st, 1970 at midnight UTC


antlai-temporal:
The protobuf library maps a nil timestamp to Jan. 1st, 1970 at midnight UTC and makes IsZero() false. One side effect is that the cli  `cardOmitEmpty` and `omitEmpty` no longer filter timestamps that have not been initialized. This is also a problem for users of the raw grpc API.

`(*timestamppb.Timestamp)(nil).AsTime().IsZero() // FALSE!`

The solution is to check with `IsValid()` before calling `AsTime()`

PR #1847, opened 2025-02-25
Versioning: Fix AsTime() for a nil proto timestamp

antlai-temporal:
## What was changed

A nil Google protobuf for a timestamp returns with `AsTime()` a year 1970, making IsZero() false, since 
Go `time.Time` expects year 1.
This PR checks for nil timestamps and returns `time.Time{}` for them in the new versioning API. 

## Checklist

1. Closes 
#1846 
2. How was this tested:

A unit-test added.

### DIFF
--- a/internal/internal_worker_deployment_client.go
+++ b/internal/internal_worker_deployment_client.go
@@ -27,11 +27,13 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"go.temporal.io/api/common/v1"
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns (approved):

k8scat:
Please have a look at #1953 which completely fix #1846 , and add safeAsDuration to avoid AsDuration for a nil proto timestamp @Quinn-With-Two-Ns

PR #1849, opened 2025-02-25
Support environment configuration

### LINKED ISSUE #1815: External client configuration

See https://github.com/temporalio/features/issues/441 and https://github.com/temporalio/proposals/blob/master/all-sdk/external-client-configuration.md

cretz:
## What was changed

* Added `go.temporal.io/sdk/contrib/envconfig` module that supports loading
  * Supports conversion to/from toml
  * Supports file loading and env var overrides
  * Supports loading client options

See https://github.com/temporalio/features/issues/441, this is mostly based on that proposal with some slight changes.

This SDK may support a few more shortcuts than other SDKs due to its use in CLI. Documentation about file format and which env vars do what will be made available on docs.temporal.io before this launches GA.

## Checklist

1. Closes #1815


contrib/envconfig/client_config.go
```diff
@@ -0,0 +1,210 @@
...
+
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/converter"
+)
+
+// ClientConfig represents a client config file.
+//
+// WARNING: Environment configuration is currently experimental.
+type ClientConfig struct {
+	Profiles map[string]*ClientConfigProfile
```
Sushisource, line 47: Worth saying what the key is here.

contrib/envconfig/client_config.go
```diff
@@ -0,0 +1,210 @@
...
+	ServerCACertPath        string
+	ServerCACertData        []byte
+	ServerName              string
+	DisableHostVerification bool
+}
+
+// ClientConfigCodec is codec configuration for a client.
+//
+// WARNING: Environment configuration is currently experimental.
+type ClientConfigCodec struct {
```
Sushisource, line 83: Maybe call this ClientConfigRemoteCodec? What, if anything, could ever be configured here for local codec? And, even then, maybe it'd make sense to have them be different sub-configs

contrib/envconfig/client_config_load.go
```diff
@@ -0,0 +1,412 @@
...
+	// defaults to [EnvLookupOS].
+	EnvLookup EnvLookup
+}
+
+// LoadClientConfig loads the client configuration structure from TOML. Does not load values from environment variables
+// (but may use environment variables to get which config file to load). This will not fail if the file does not exist.
+// See [ClientConfig.FromTOML] for details on format.
+//
+// WARNING: Environment configuration is currently experimental.
+func LoadClientConfig(options LoadClientConfigOptions) (ClientConfig, error) {
```
Sushisource, line 140: This and `LoadClientOptions` don't have distinct enough names, IMO.

Maybe `LoadClientOptionsFromFile` ?

contrib/envconfig/client_config_load.go
```diff
@@ -0,0 +1,412 @@
...
+func (envLookupOS) Environ() []string                   { return os.Environ() }
+func (envLookupOS) LookupEnv(key string) (string, bool) { return os.LookupEnv(key) }
+
+// ApplyEnvVars overwrites any values in the profile with environment variables if the environment variables are set and
+// non-empty. If env lookup is nil, defaults to [EnvLookupOS]
+func (c *ClientConfigProfile) ApplyEnvVars(env EnvLookup) error {
+	if env == nil {
+		env = EnvLookupOS
+	}
+	if s, ok := env.LookupEnv("TEMPORAL_ADDRESS"); ok {
```
Sushisource, line 308: I think it probably makes sense to make these all public constants. It helps with both documentation and, since people can technically implement their own lookup, they might want to programmatically use them.

contrib/envconfig/client_config_load.go
```diff
@@ -0,0 +1,412 @@
...
+		conf, err := LoadClientConfig(LoadClientConfigOptions{
+			ConfigFilePath:   options.ConfigFilePath,
+			ConfigFileData:   options.ConfigFileData,
+			ConfigFileStrict: options.ConfigFileStrict,
+			EnvLookup:        options.EnvLookup,
+		})
+		if err != nil {
+			return ClientConfigProfile{}, err
+		}
+		// Find profile or fail. The one situation where we don't fail here is if the profile is not user-set
```
yuandrew, line 238: This comment took me a sec to understand, I think something like this could be easier to understand
```suggestion
		// Find user-set profile or use default. Fail if user uses invalid profile.
```

contrib/envconfig/client_config.go
cretz: It is the `codec` field in config and `TEMPORAL_CODEC_` prefixed env vars. So I want it to match those. If we want though, I could change the config field name, env var prefix, struct field name, and this type name.

contrib/envconfig/client_config_load.go
cretz: I think it's distinct enough given its return type and such, but I can change and change the param structure to `LoadClientConfigFromFileOptions`. Would like to wait on other opinions if anyone has any.

contrib/envconfig/client_config.go
```diff
@@ -0,0 +1,210 @@
...
+//
+// WARNING: Environment configuration is currently experimental.
+type ClientConfigProfile struct {
+	Address   string
+	Namespace string
+	APIKey    string
+	TLS       *ClientConfigTLS
+	Codec     *ClientConfigCodec
+	// When loading from TOML and env var, or writing to TOML, the keys are lowercased and hyphens are replaced with
+	// underscores. This is used for deduplicating/overriding too, so manually set values that are not normalized may
```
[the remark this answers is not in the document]
cretz: Hrmm, "manually set values" makes sense to me, maybe "manually-set values"?

tomwheeler (commented):
@cretz I updated my local copy of your Go SDK to commit [#362043e](https://github.com/cretz/temporal-sdk-go/commit/362043eeb9b08fa7a98d9a588f94726795175f38) and then tried this again.

Following [the approach previously described](https://github.com/cretz/temporal-sdk-go/commit/a10d1ec2f286eaf8c87805498dac25ec2a591143#commitcomment-152763562), I tested this with mTLS and found that it now works as expected.

I then attempted to do the same with API Key authentication. I unset the previous four environment variables and then set the following:

* `TEMPORAL_API_KEY=<redacted>`
* `TEMPORAL_NAMESPACE=tom-dev.a2dd6`
* `TEMPORAL_ADDRESS=us-east-1.aws.api.temporal.io:7233`


When I started the Worker, it failed:

```
$ go run worker/main.go
2025/02/25 13:53:31 INFO  No logger configured for temporal client. Created default one.
2025/02/25 13:53:32 Unable to create Temporal client. failed reaching server: rpc error: code = Unauthenticated desc = Jwt is expired
exit status 1
```

I then looked at the code and realized that perhaps I needed to run this:

```
$ export TEMPORAL_TLS=true
```

I then repeated the previous command and found that the Worker now started. I did the same in the other terminal before running the command to start the Workflow, which then completed successfully.

In the end, it seems to work fine, although I think we'd be wise to emphasize in the docs that you must set `TEMPORAL_TLS=true` when using API Key authentication with Temporal Cloud, because I think people might overlook it otherwise.

contrib/envconfig/client_config.go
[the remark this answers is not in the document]
yuandrew: ahh re-reading now, you're right, this doesn't need to be changed

cretz:
> Jwt is expired

@tomwheeler - that is the error you got originally, I am not sure how `TEMPORAL_TLS=true` affects that. For env config, we are defaulting TLS to true if you set the API key (all of this is not documented yet because so much may change during this review process), so you should not have to set it manually. I will confirm this is the case. Also, feel free to contact via internal channels if that'd be easier.

contrib/envconfig/client_config.go
```diff
@@ -0,0 +1,210 @@
...
+// WARNING: Environment configuration is currently experimental.
+type ClientConfig struct {
+	Profiles map[string]*ClientConfigProfile
+}
+
+// ClientConfigProfile is profile-level configuration for a client.
+//
+// WARNING: Environment configuration is currently experimental.
+type ClientConfigProfile struct {
+	Address   string
```
Quinn-With-Two-Ns, line 54: Don't we generally want all public structs documented?

contrib/envconfig/client_config.go
```diff
@@ -0,0 +1,210 @@
...
+	// underscores. This is used for deduplicating/overriding too, so manually set values that are not normalized may
+	// not get overridden with [ClientConfigProfile.ApplyEnvVars].
+	GRPCMeta map[string]string
+}
+
+// ClientConfigTLS is TLS configuration for a client.
+//
+// WARNING: Environment configuration is currently experimental.
+type ClientConfigTLS struct {
+	Disabled                bool
```
Quinn-With-Two-Ns, line 69: So the default will be TLS enabled?

Quinn-With-Two-Ns: Or are we treating `nil` different then the zero value?

contrib/envconfig/client_config.go
cretz: Yes, though I am not sure it necessarily applies to all public struct fields. I can document it, though I will likely just say it applies to certain client option fields and not give all the details.

contrib/envconfig/client_config.go
cretz: > So the default will be TLS enabled?

No, the default is "depends on other fields". If you give an API key the default is enabled, but if you don't the default is disabled (but you can make enabled with any TLS value including TLS itself being set). So this is effectively for those rare cases that have an API key but want to explicitly disable TLS (e.g. self-hosted using JWTs). This is what allows us to just accept an API key env var for cloud and assume TLS.

contrib/envconfig/client_config_load.go
```diff
@@ -0,0 +1,412 @@
...
+// This just delegates to [LoadClientOptions] with the default options set. See that function and associated options for
+// where/how values are loaded.
+//
+// WARNING: Environment configuration is currently experimental.
+func LoadDefaultClientOptions() (client.Options, error) {
+	return LoadClientOptions(LoadClientOptionsOptions{})
+}
+
+// LoadClientOptionsOptions are options for [LoadClientOptions].
+type LoadClientOptionsOptions struct {
```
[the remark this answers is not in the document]
cretz: I didn't think we had settled on this. I thought request was for actual API requests (i.e. RPC calls). But I can change `ToClientOptionsOptions` to `ToClientOptionsRequest` if you think that's better?

contrib/envconfig/client_config_load.go
Sushisource: Why not? They could be doing it as part of some infra-as-code stuff

cretz: Because the constants are part of the spec and I am not sure it has enough value making them part of the public API in every SDK (nor do I think that with the other environment variable names such as `TEMPORAL_DEBUG`).

contrib/envconfig/client_config.go
cretz: Added comment

contrib/envconfig/client_config_load.go
cretz: Updated comment

contrib/envconfig/client_config_load.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I am fine with it being `Request` even though it is not a client call. OK leaving it as `Option` as well since we are consistent in this module at least

contrib/envconfig/client_config.go
cretz: Unset/false here is "not disabled", not necessarily "enabled". I agree it can be confusing, but I think any alternative is more confusing.

contrib/envconfig/client_config_load.go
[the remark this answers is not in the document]
cretz: Changed to `OptionsRequest`

contrib/envconfig/client_config.go
cretz: Documented struct fields

Sushisource (approved):

Quinn-With-Two-Ns (approved):

PR #1851, opened 2025-02-25
Enable TestNexusTracingInterceptor if workflow cache is enabled

bergundy:


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
This does not look like a flaky testhttps://github.com/temporalio/sdk-go/actions/runs/13551018738/job/37874374899?pr=1851

bergundy:
> https://github.com/temporalio/sdk-go/actions/runs/13551018738/job/37874374899?pr=1851

Ah this is because I haven't overridden the dynamic config for the Nexus registry refresh backoff in docker-compose.

Note that this won't be required in the next server release since I've changed the behavior: https://github.com/temporalio/temporal/commit/5c3e16fcbe6d82717a4e3ab691716aa320668631

PR #1853, opened 2025-02-26
Update documentation for `Next` history iterator

timsexperiments:
## What was changed

This change:
- Removes comment that contradicts behavior
- Add documentation comment to make the functions behavior and dependency to `HasNext` clear.
- Adds test to ensure that and breaking changes to this behavior is caught

## Why?

While I was exploring the documentation for the get history calls at work, I was trying to figure out what happens with the iterator if `HasNext()` is false but the `Next()` function was called, and I noticed that there was this line where the comment contradicted it.

## Checklist

1. How was this tested:

Added unit test.

2. Any docs updates needed?

I don't think so

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -1454,7 +1454,7 @@ func (iter *historyEventIteratorImpl) HasNext() bool {
 func (iter *historyEventIteratorImpl) Next() (*historypb.HistoryEvent, error) {
 	// if caller call the Next() when iteration is over, just return nil, nil
 	if !iter.HasNext() {
-		panic("HistoryEventIterator Next() called without checking HasNext()")
+		return nil, nil
 	}
 
 	// we have cached events
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -1451,10 +1451,12 @@ func (iter *historyEventIteratorImpl) HasNext() bool {
 	return false
 }
 
+// Next returns the next history event.
+// If next is called with not more events, it will panic.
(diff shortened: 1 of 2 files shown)


internal/internal_workflow_client.go
```diff
@@ -1450,7 +1450,7 @@ func (iter *historyEventIteratorImpl) HasNext() bool {
 func (iter *historyEventIteratorImpl) Next() (*historypb.HistoryEvent, error) {
 	// if caller call the Next() when iteration is over, just return nil, nil
 	if !iter.HasNext() {
-		panic("HistoryEventIterator Next() called without checking HasNext()")
```
Quinn-With-Two-Ns, line 1453: The code behaviour is correct the comment is incorrect, if you want you can submit a PR to update the comment

timsexperiments: Sg, updated to keep the panic and added a documentation comment to make the behavior and dependency on `HasNext` clear.

As a user of the API, my initial expectation was that reaching the end of the events would return an error like ErrNoMoreResults, rather than triggering a panic. I think for most go users' expectations, panics are generally an indicator for unrecoverable errors, not for a normal control flow situation like running out of events.

Quinn-With-Two-Ns: >Sg, updated to keep the panic and added a documentation comment to make the behavior and dependency on HasNext clear.

Yes

Not calling `HasNext` before `Next ` is a logic error in your code like trying send to a closed channel, it is not control flow

Quinn-With-Two-Ns (approved):

Issue #1868, opened 2025-03-13
ResourceBasedSlotSupplier.ReleaseSlot not implemented


captchanjack-anzx:
## Expected Behavior
Activity SlotSupplier created with `resourcetuner.NewResourceBasedSlotSupplier` should release activity slots when they are completed.

## Actual Behavior
Activity slots are not released after task is completed. This causes scheduled activities to forever hand in `PENDING_ACTIVITY_STATE_SCHEDULED` state.

I suspect because the SlotSupplier interface method ReleaseSlot is not implemented:

https://github.com/temporalio/sdk-go/blob/master/contrib/resourcetuner/resourcetuner.go#L204

## Steps to Reproduce the Problem

  1. Create a tuner with `worker.NewCompositeTuner` and `resourcetuner.NewResourceBasedSlotSupplier` as suggested in [go-sdk docs](https://docs.temporal.io/develop/worker-performance#go-sdk)
  1. Set activity `ResourceBasedSlotSupplierOptions.MaxSlots` to something small like 4.
  1. Start a worker with tuner set in worker.Options
  1. Execute a workflow that starts a large amount of activities - few activities should be scheduled, and others should be schedule/stuck in pending state.

## Specifications

  - Version:
  ```
        go.temporal.io/api v1.44.1
	go.temporal.io/sdk v1.33.0
	go.temporal.io/sdk/contrib/resourcetuner v0.0.0-20250227195219-884681f7392e
```
  - Platform: GCP Cloud Run, 8 cores 8GiB per instance, unsure chip used (i think amd based)

PR #1870, opened 2025-03-14
Fix possibly failing to reserve slots due to stale issued number

Sushisource:
## What was changed
Fix a bug where resource based slot supplier was looking at a stale issued slots number while trying to reserve

## Why?
Could cause it to be stuck failing to reserve

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/1868

2. How was this tested:
Added test

3. Any docs updates needed?

### DIFF
--- a/internal/tuning.go
+++ b/internal/tuning.go
@@ -328,7 +328,7 @@ type slotReserveInfoImpl struct {
 	taskQueue      string
 	workerBuildId  string
 	workerIdentity string
-	issuedSlots    int
+	issuedSlots    *atomic.Int32
 	logger         log.Logger
 	metrics        metrics.Handler
 }
@@ -346,7 +346,7 @@ func (s slotReserveInfoImpl) WorkerIdentity() string {
 }
 
 func (s slotReserveInfoImpl) NumIssuedSlots() int {
-	return s.issuedSlots
+	return int(s.issuedSlots.Load())
 }
 
 func (s slotReserveInfoImpl) Logger() log.Logger {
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns:
What is the reason for using the `SlotInfo` count instead of having `ResourceBasedSlotSupplier` keep track? It should be fine but it is hard to reason about since the `SlotInfo` count is updated outside the `mutex` of the `ResourceBasedSlotSupplier`?

Quinn-With-Two-Ns:
Also is there a reason this isn't implemented if max slots is provided? https://github.com/temporalio/sdk-go/blob/master/contrib/resourcetuner/resourcetuner.go#L205

Sushisource:
> What is the reason for using the `SlotInfo` count instead of having `ResourceBasedSlotSupplier` keep track? It should be fine but it is hard to reason about since the `SlotInfo` count is updated outside the `mutex` of the `ResourceBasedSlotSupplier`?

Because internally we always need to know the issued count, so the tracking supplier does that so it's not duplicated in all implementations. The mutex in the resource based supplier is purely for the last issued time / PID decisions.

Sushisource:
> Also is there a reason this isn't implemented if max slots is provided? https://github.com/temporalio/sdk-go/blob/master/contrib/resourcetuner/resourcetuner.go#L205

It's debatable. I left it unimplemented because it maps to the "available slots" metric, but, you can easily set a max here that isn't actually attainable in any real sense.

Quinn-With-Two-Ns:
> Because internally we always need to know the issued count

To confrim we only update our internal count **after** the slot supplier `ReleaseSlot` returns right?

Sushisource:
> > Because internally we always need to know the issued count
> 
> To confrim we only update our internal count **after** the slot supplier `ReleaseSlot` returns right?

Yup, exactly

PR #1875, opened 2025-03-18
Allow local activities to finish running when stopping a worker

### LINKED ISSUE #1706: ProcessWorkflowTask is not stopped on worker.Stop()

yuandrew:
## What was changed

shutdown the local activity worker after the worker. Also created a separate stop channel to allow pending local activities to finish running when stopping a worker.

## Why?

Previously, local activities would immediately be stopped when a worker was shut down, but we want to allow pending local activities to complete during shutdown. This behavior should now match core.

## Checklist

1. Closes #1706 

2. How was this tested:

Added a new `TestGracefulLocalActivityCompletion` test that validates when the worker stop is initiated, a LA currently processing, as well as a future LA in the same WFT will be completed.

Also added `TestLocalActivityTaskTimeoutHeartbeat`, which validates a LA can still heartbeat and complete after worker shutdown has been initiated.

3. Any docs updates needed?


test/integration_test.go
cretz, line 1: Do we expect local activities to complete if they cross the task timeout and use workflow task heartbeating (an advanced mechanism that allows local activities to go beyond one workflow task). I wonder if we should have a test that sets workflow task timeout to 2s, and has a local activity that initiates worker shutdown when called + sleeps for 4s before returning. I don't really like tests that take 4s, but I can't think of a better way to confirm workflow task heartbeating works on worker shutdown. @Quinn-With-Two-Ns - thoughts?

yuandrew: Good call, I'll add a test for this

test/integration_test.go
```diff
@@ -2391,6 +2393,141 @@ func (ts *IntegrationTestSuite) TestGracefulActivityCompletion() {
...
+	// Start workflow
+	run, err := ts.client.ExecuteWorkflow(ctx, startOptions, workflowFn)
+	ts.NoError(err)
+
+	//// Stop the worker
+	time.Sleep(100 * time.Millisecond)
+	ts.worker.Stop()
+	ts.workerStopped = true
+	// Allow time for LA to complete
+	time.Sleep(4 * time.Second)
```
cretz, line 2507: May be able to just sit on history iterator waiting for the LA marker to appear (basically remove this sleep and change `false` to `true` on the `GetWorkflowHistory` call)

test/integration_test.go
```diff
@@ -2391,6 +2393,141 @@ func (ts *IntegrationTestSuite) TestGracefulActivityCompletion() {
...
+}
+
+func (ts *IntegrationTestSuite) TestLocalActivityTaskTimeoutHeartbeat() {
+	// FYI, setup of this test allows the worker to wait to stop for 10 seconds
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	localActivityFn := func(ctx context.Context) error {
+		// wait for worker shutdown to be started and WorkflowTaskTimeout to be hit
+		time.Sleep(3 * time.Second)
```
cretz, line 2472: I feel bad increasing Go test times like this. @Quinn-With-Two-Ns any ideas on how to do this test effectively using up any less time? I figure 2s wft timeout is enough to prevent flakes, but it is annoying.

test/integration_test.go
```diff
@@ -2391,6 +2393,141 @@ func (ts *IntegrationTestSuite) TestGracefulActivityCompletion() {
...
+	ts.workerStopped = true
+	// Allow time for LA to complete
+	time.Sleep(4 * time.Second)
+
+	// Look for activity completed from the history
+	var laCompleted int
+	var wfeCompleted bool
+	iter := ts.client.GetWorkflowHistory(ctx, run.GetID(), run.GetRunID(),
+		false, enumspb.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT)
+	for iter.HasNext() {
```
cretz, line 2514: Can you do me a favor and assert that a workflow task heartbeat occurred? This could be done by basically counting that there are >= 2 workflow task start events I think (workflow start and then wft heartbeat).

test/integration_test.go
Quinn-With-Two-Ns: No I mean you have to heartbeat, could lower the last timeout or have a debug to lower the heartbeat interval

test/integration_test.go
yuandrew: Ahh yes, that would be important for this test 😅

test/integration_test.go
```diff
@@ -2391,6 +2393,141 @@ func (ts *IntegrationTestSuite) TestGracefulActivityCompletion() {
...
+		} else if event.EventType == enumspb.EVENT_TYPE_WORKFLOW_TASK_STARTED {
+			started++
+		} else if event.EventType == enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED {
+			wfeCompleted = true
+		}
+	}
+
+	// Confirm local activity and WFE completed
+	ts.Equal(1, laCompleted)
+	ts.Equal(2, started)
```
cretz, line 2527:
```suggestion
	ts.GreaterOrEqual(started, 2)
```

_Technically_ something could happen server side to cause this to be more than two. Not a big deal though.

cretz (approved):

internal/internal_task_pollers.go
```diff
@@ -625,7 +625,7 @@ func (latp *localActivityTaskPoller) PollTask() (taskForWorker, error) {
 }
 
 func (latp *localActivityTaskPoller) ProcessTask(task interface{}) error {
-	if latp.stopping() {
+	if latp.stopping() && (task == nil || task.(*localActivityTask) == nil) {
```
Quinn-With-Two-Ns, line 628: Can this still happen? I remember when we were debugging this it could happen because we hadn't made a separate stop channel for the local activity tunnel

yuandrew: oh you're right, confirmed this doesn't hit anymore

yuandrew: scratch that, this is still needed because `latp.stopping()` uses the stop channel that's shared with the worker. the tunnel will continue to send tasks even when `latp.stopping()` is true, allowing us to complete pending local activities

Quinn-With-Two-Ns: Didn't we want to use a different channel for the `localActivityTaskPoller `?b Or is there a reason we need to use the stop channel that's shared with the worker?

Quinn-With-Two-Ns (approved):

PR #1879, opened 2025-03-19
Add RawValue support

### LINKED ISSUE #1231: Add RawValue type
Sometimes users just want to get the payload from something without attempting any data conversion.

Add a type like pythons [RawValue](https://python.temporal.io/temporalio.common.RawValue.html) to the Go SDK

**Additional context**
This would be useful to handle corrupted signals because the user could read the signal value without putting it through the data converter

yuandrew:
## What was changed
Added new RawValue type and add support in codec and composite data converter.

## Why?
SDKs need to support a payload wrapper called "raw value" that is just a payload but is passed through the conversion stage (but still applies to codec).

## Checklist

1. Closes #1231 

2. How was this tested:
Added unit tests for both codec and composite data converter, and an integration test that pass a payload to and from a workflow and activity

3. Any docs updates needed?
No

### DIFF
--- a/converter/codec.go
+++ b/converter/codec.go
@@ -133,6 +133,9 @@ func (*zlibCodec) Decode(payloads []*commonpb.Payload) ([]*commonpb.Payload, err
 // CodecDataConverter is a DataConverter that wraps an underlying data
 // converter and supports chained encoding of just the payload without regard
 // for serialization to/from actual types.
+//
+// CodecDataConverter provides support for RawValue handling, where it skips the
+// parent data converter and directly encodes/decodes the RawValue payload.
 type CodecDataConverter struct {
 	parent DataConverter
 	codecs []PayloadCodec
@@ -200,7 +203,6 @@ func (e *CodecDataConverter) ToPayload(value interface{}) (*commonpb.Payload, er
 func (e *CodecDataConverter) ToPayloads(value ...interface{}) (*commonpb.Payloads, error) {
 	var payloads *commonpb.Payloads
 	var rawValuePayloads []*commonpb.Payload
-	fmt.Println("len(value)", value)
 	for _, v := range value {
 		rawValue, ok := v.(RawValue)
 		if ok {
--- a/converter/codec.go
+++ b/converter/codec.go
@@ -176,7 +176,7 @@ func (e *CodecDataConverter) ToPayload(value interface{}) (*commonpb.Payload, er
 
 	var payload *commonpb.Payload
 	if ok {
-		payload = rawValue.Payload
+		payload = rawValue.Payload()
 	} else {
 		var err error
 		payload, err = e.parent.ToPayload(value)
@@ -204,7 +204,7 @@ func (e *CodecDataConverter) ToPayloads(value ...interface{}) (*commonpb.Payload
 	for _, v := range value {
 		rawValue, ok := v.(RawValue)
 		if ok {
-			rawValuePayloads = append(rawValuePayloads, rawValue.Payload)
+			rawValuePayloads = append(rawValuePayloads, rawValue.Payload())
 		}
 	}
 
--- a/converter/codec.go
+++ b/converter/codec.go
@@ -172,9 +172,17 @@ func (e *CodecDataConverter) decode(payloads []*commonpb.Payload) ([]*commonpb.P
 // ToPayload implements DataConverter.ToPayload performing encoding on the
 // result of the parent's ToPayload call.
 func (e *CodecDataConverter) ToPayload(value interface{}) (*commonpb.Payload, error) {
-	payload, err := e.parent.ToPayload(value)
-	if payload == nil || err != nil {
-		return payload, err
+	rawValue, ok := value.(RawValue)
+
+	var payload *commonpb.Payload
+	if ok {
+		payload = rawValue.Payload
+	} else {
+		var err error
+		payload, err = e.parent.ToPayload(value)
+		if payload == nil || err != nil {
+			return payload, err
+		}
 	}
 
 	encodedPayloads, err := e.encode([]*commonpb.Payload{payload})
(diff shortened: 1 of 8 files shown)


converter/codec.go
```diff
@@ -225,6 +255,22 @@ func (e *CodecDataConverter) FromPayloads(payloads *commonpb.Payloads, valuePtrs
 	if err != nil {
 		return err
 	}
+
+	var isRawValue bool
+	for i, payload := range decodedPayloads {
+		if i >= len(valuePtrs) {
+			break
+		}
+		rawValue, ok := valuePtrs[i].(*RawValue)
```
yuandrew, line 264: My assumption is `valuePtrs` will always be a slice of pointers, please correct me if I'm wrong.

yuandrew:
the work item also mentions "This would be useful to handle corrupted signals because the user could read the signal value without putting it through the data converter", I'm inclined to have it as a separate issue and PR

converter/codec.go
```diff
@@ -172,9 +172,17 @@ func (e *CodecDataConverter) decode(payloads []*commonpb.Payload) ([]*commonpb.P
 // ToPayload implements DataConverter.ToPayload performing encoding on the
 // result of the parent's ToPayload call.
 func (e *CodecDataConverter) ToPayload(value interface{}) (*commonpb.Payload, error) {
```
cretz, line 174: I don't think raw value logic should be in the `CodecDataConverter`, I think `CompositeDataConverter` is enough. Codecs have nothing to do with raw value, raw value is about type conversion, raw value handling should be done by the parent/underlying data converter.

converter/composite_data_converter.go
cretz, line 1: I think on the `DataConverter` interface, `ToPayload` and `FromPayload` should have docs saying that when converting to/from actual Go types, `RawValue` must be respected.

converter/value.go
```diff
@@ -48,4 +50,17 @@ type (
 		// json.Unmarshal.
 		Get(valuePtr ...interface{}) error
 	}
+
+	// RawValue is a representation of an unconverted, raw payload.
+	//
+	// This type can be used as a paramter or return type in workflows and activities to pass through
+	// a raw payload. Encoding/decoding of the payload is still done by the system.
+	RawValue struct {
```
cretz, line 58: I wonder if we should define a `MarshalJSON` and `UnmarshalJSON` for this type that returns an error saying it is not JSON serializable.

cretz, line 58: I wonder if we should consider `PayloadMarshaler` and `PayloadUnmarshaler` interfaces similar to `json.Marshaler` and `json.Unmarshaler` and just have `RawValue` implement both.

converter/value.go
```diff
@@ -48,4 +50,17 @@ type (
...
+		Payload *commonpb.Payload
```
cretz, line 59: You might consider not exporting this field and having a `Payload()` method instead. This discourages people from creating the struct themselves. The benefit of discouraging this is that it is less likely for people to use a `*RawValue` instead of a `RawValue` since the must use `NewRawValue` and see its result type is non-pointer.

cretz (commented):
Added some thoughts, but interested in @Quinn-With-Two-Ns's thoughts as well.

converter/codec.go
Quinn-With-Two-Ns: I disagree, from the issue `Encoding/decoding of the payload is still done by the system` so the `CodecDataConverter` has to be aware of it

converter/value.go
cretz: If stack trace opt in, no prob. As for the interfaces, I think that's fairly uniquely Go, but ok if we don't want it.

converter/codec.go
Quinn-With-Two-Ns: You can't assume that the parent data convert will have support for `RawValue`. There is no reason not to build in support for `RawValue` into the `CodecDataConverter`

cretz: > You can't assume that the parent data convert will have support for RawValue

Right, same as if the codec was not present, but adding new type conversion behavior whether you have the codec on or off is confusing. It's not like, due to the inability to "assume that the parent data convert will have support for RawValue" we're going to everywhere the data converter is using and altering the code based on that assumption, we are just hoping that type conversion properly respects this value.

Quinn-With-Two-Ns: We can clarify in the `CodecDataConverter` docs, but the `CodecDataConverter` should handle `RawValue` absolutely. It is not a responsibility we should push onto our users .

converter/codec.go
Quinn-With-Two-Ns: Well users can pass anything , normally the data converters return `ErrValuePtrIsNotPointer`. I am no sure if there is a case to not pass a pointer. I think I would just add a check here and if it is not a pointer you can assume it is not a raw value and forward it like normal.

yuandrew: Cool, we should be okay with the current code then, if it's not a `*RawValue`, all of the RawValue behavior should be skipped.

converter/codec.go
cretz: It is a responsibility we should push onto users not using a composite payload converter, but ok if you want to _also_ add to codec data converter (it's a bit confusing/inconsistent that only codec users don't have to properly handle this type in non-composite payload converters they write). There are a lot of people not using codecs, so we need to be consistent in our expectation of user responsibility for handling this value regardless of codec.

converter/value.go
```diff
@@ -48,4 +53,29 @@ type (
...
+	//
+	// This type can be used as a paramter or return type in workflows and activities to pass through
+	// a raw payload. Encoding/decoding of the payload is still done by the system.
+	RawValue struct {
+		payload *commonpb.Payload
+	}
 )
+
+// NewRawValue creates a new RawValue instance.
+func NewRawValue(payload *commonpb.Payload) RawValue {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Since once something starts getting used in workflows we have to consider history compatibility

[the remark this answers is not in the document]
yuandrew: Do you mean we should test that `RawValue` returned from a workflow/activity goes through the data converter properly?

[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah like an end to end test where a client uses a `RawValue ` as input and `RawValue ` as the output for a workflow and activity. Does that seem resonable

Quinn-With-Two-Ns (approved):

Issue #1882, opened 2025-03-25
Caller skips is incorrectly calculated when a custom logger is provided to the SDK client


AhmadElsagheer:
## Context
We have, at Payrails, a custom logger (a wrapper of zap)  that implements the temporal SDK `log.Logger` interface and is provided to the SDK client. Recently, we wanted to sanitize the logs to make the zap log fields `caller` and `stacktrace` more representative of the actual log line skipping all logging and temporal SDK internal details. So, we additionally implemented `log.WithSkipCallers` interface but looks like there is an overcount in the skip calls.

Without implementing the skipping interface, an error log looks like
```json
{
    "level": "error",
    "ts": 1742899256.974098,
    "caller": "log/with_logger.go:79",    // <--- Current (without the skipping interface)
    "msg": "HelloWorld: this is an error inside the workflow",
    "Namespace": "default",
    "TaskQueue": "main_queue",
    "WorkerID": "37387@KCJ74K6WJ2@",
    "WorkflowType": "HelloWorld",
    "WorkflowID": "989f5c44-9311-4bd1-be2d-5477c157392c",
    "RunID": "84ecd3c3-6ab2-4cd9-82f6-ca14ed470f58",
    "Attempt": 1,
[...]

PR #1883, opened 2025-03-25
Fix withLogger caller skip overcount

AhmadElsagheer:
## What was changed
Decrementing caller skip depth by 1 to offset the increment by 1 in `newWithLogger`.

## Why?
- The `withLogger` logger, whose `WithCallerSkip()` function is called, is replaced by a new `withLogger` using the same underlying logger `l.logger`. 
- Each new `withLogger` adds 1 skip in `newWithLogger`.
- When a `withLogger` is removed, its added skip should be decremented as well to fix the overcounting issue.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/1882

2. How was this tested:
Ran the same code as in the issue and got the expected behavior:
```
--- skip: 0
--- skip: 1
--- skip: -1
--- skip: 1
--- skip: 0
--- skip: 1
{"level":"error","ts":1742900242.772918,"caller":"play/business.go:11","msg":"HelloWorld: this is an error inside the workflow","Namespace":"default","TaskQueue":"main_queue","WorkerID":"60774@KCJ74K6WJ2@","WorkflowType":"HelloWorld","WorkflowID":"3827b52a-a075-4f93-887d-9347c7b933d5","RunID":"210f13ef-f403-454d-9da9-d3161a5ec078","Attempt":1,"stacktrace":"main.HelloWorld\n\t/payrails/sdk-go/play/business.go:11\nreflect.Value.call\n\t/usr/local/go/src/reflect/value.go:581\nreflect.Value.Call\n\t/usr/local/go/src/reflect/value.go:365\ngo.temporal.io/sdk/internal.executeFunction\n\t/payrails/sdk-go/internal/internal_worker.go:2087\ngo.temporal.io/sdk/internal.(*workflowEnvironmentInterceptor).ExecuteWorkflow\n\t/payrails/sdk-go/internal/workflow.go:797\ngo.temporal.io/sdk/internal.(*workflowExecutor).Execute\n\t/payrails/sdk-go/internal/internal_worker.go:922\ngo.temporal.io/sdk/internal.(*syncWorkflowDefinition).Execute.func1\n\t/payrails/sdk-go/internal/internal_workflow.go:574\ngo.temporal.io/sdk/internal.(*coroutineState).run\n\t/payrails/sdk-go/internal/internal_workflow.go:1217"}
```

3. Any docs updates needed?
No

### DIFF
--- a/log/with_logger.go
+++ b/log/with_logger.go
@@ -81,7 +81,7 @@ func (l *withLogger) Error(msg string, keyvals ...interface{}) {
 
 func (l *withLogger) WithCallerSkip(depth int) Logger {
 	if sl, ok := l.logger.(WithSkipCallers); ok {
-		return newWithLogger(sl.WithCallerSkip(depth), l.keyvals...)
+		return newWithLogger(sl.WithCallerSkip(depth-1), l.keyvals...)
 	}
 	return l
 }


log/with_logger.go
```diff
@@ -81,7 +81,7 @@ func (l *withLogger) Error(msg string, keyvals ...interface{}) {
 
 func (l *withLogger) WithCallerSkip(depth int) Logger {
 	if sl, ok := l.logger.(WithSkipCallers); ok {
-		return newWithLogger(sl.WithCallerSkip(depth), l.keyvals...)
+		return newWithLogger(sl.WithCallerSkip(depth-1), l.keyvals...)
```
Quinn-With-Two-Ns, line 84: I'm sorry, this doesn't quite make sense. This output
```
--- skip: 0
--- skip: 1
--- skip: -1
--- skip: 1
--- skip: 0
--- skip: 1
```
Is not expected, a logger should never see a negative depth.

AhmadElsagheer: Yeah I was unsure about that as well. Also, there is a double skipping here and inside newWithLogger. If combined (with a bit of refactoring of `newWithLogger`), the logger will never see a negative depth.

Looks like it's assumed that `WithSkipCallers` is meant to be always implemented with `WithLogger`, so this scenario is not applicable.

PR #1885, opened 2025-03-25
Reproduce heartbeats

ndtretyak:
Test reproducing https://github.com/temporalio/sdk-go/issues/1706

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -250,7 +250,7 @@ func (ts *IntegrationTestSuite) SetupTest() {
 	}
 
 	if strings.Contains(ts.T().Name(), "GracefulActivityCompletion") ||
-		strings.Contains(ts.T().Name(), "GracefulLocalActivityCompletion") ||
+		// strings.Contains(ts.T().Name(), "GracefulLocalActivityCompletion") ||
 		strings.Contains(ts.T().Name(), "TestLocalActivityTaskTimeoutHeartbeat") {
 		options.WorkerStopTimeout = 10 * time.Second
 	}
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
Hi https://github.com/temporalio/sdk-go/issues/1706 is now resolved and a fix will be included in the next release. The SDK will not correctly wait for the workflow tasks to finish before shutting down the local activity worker (assuming a graceful stop time is provided)

ndtretyak:
In my opinion, it's fine not to wait for the workflow tasks to finish and let them fail. However, they currently end up in an invalid state and [continue heartbeating](https://github.com/temporalio/sdk-go/actions/runs/14067722219/job/39394562700?pr=1885#step:6:811). Is it possible to stop heartbeating if the local activity worker has already stopped?

Also, there is no `WorkerStopTimeout` in the server's code https://github.com/search?q=repo%3Atemporalio%2Ftemporal%20WorkerStopTimeout&type=code, so this fix alone won't prevent server's scheduler from getting stuck.

Quinn-With-Two-Ns:
>However, they currently end up in an invalid state and [continue heartbeating](https://github.com/temporalio/sdk-go/actions/runs/14067722219/job/39394562700?pr=1885#step:6:811).

The worker is not in any invalid state, it is waiting for the local activity to finish or timeout

ndtretyak:
I added another commit setting ScheduleToCloseTimeout to 2 seconds and StartToCloseTimeout to 1 second, but the workflow task continues heartbeating for the entire duration of the test (15 seconds).

Quinn-With-Two-Ns:
>I added another commit setting ScheduleToCloseTimeout to 2 seconds and StartToCloseTimeout to 1 second, but the workflow task continues heartbeating for the entire duration of the test (15 seconds).

Ah okay that is not correct I agree. Thank you for explaining. @yuandrew can you please take a look

PR #1891, opened 2025-03-27
Fix `WorkflowId` being wrongly capitalized on example comments

Kamefrede:
## What was changed

This PR changes the capitalization of `WorkflowID` to `WorkflowId` in the comments of the example code for `ListWorkflow`

## Why?

I copied the example code to use in a query like so:

```go
temporalClient.ListWorkflow(context.Background(), &workflowservice.ListWorkflowExecutionsRequest{
		Namespace: TemporalNamespace,
		Query:     fmt.Sprintf("(WorkflowID = '%s' or WorkflowID = '%s')", workflow1, workflow2),
	})
```

And was greeted with the following error:

`invalid query: invalid expression: column name 'WorkflowID' is not a valid search attribute`

Changing `WorkflowID` to `WorkflowId` fixed the issue, so I'm opening this PR so others after me don't fall into the same mistake.

## Checklist

1. How was this tested:
- Ran a `ListWorkflow` query with the updated example code.
1. Any docs updates needed?
I don't think so? At least I'm not sure if this example is in the docs.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1036,7 +1036,7 @@ type (
 		ListOpenWorkflow(ctx context.Context, request *workflowservice.ListOpenWorkflowExecutionsRequest) (*workflowservice.ListOpenWorkflowExecutionsResponse, error)
 
 		// ListWorkflow gets workflow executions based on query. The query is basically the SQL WHERE clause, examples:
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns (approved):
Thanks!

Issue #1887, opened 2025-03-26
enabling FIPS only mode with go 1.24.1 raises panic related to MD5 usage


muralisrini:
## Expected Behavior
FIPS compatible protocols are in use and go runtime should not raise FIPS related panic

## Actual Behavior
Panic raised in GO runtime:

```
panic: crypto/md5: use of MD5 is not allowed in FIPS 140-only mode

goroutine 49 [running]:
go.temporal.io/sdk/internal.getBinaryChecksum()
        /home/muralisr/manetu/pkg/mod/go.temporal.io/sdk@v1.33.0/internal/internal_worker.go:1246 +0xca
go.temporal.io/sdk/internal.(*workerExecutionParameters).getBuildID(...)
        /home/muralisr/manetu/pkg/mod/go.temporal.io/sdk@v1.33.0/internal/internal_worker.go:291
go.temporal.io/sdk/internal.newWorkflowTaskHandler({{0x1d42a8c, 0x7}, {0x1dd9731, 0x18}, {0x2383840, 0xc000a894f0}, 0x40f86a0000000000, 0x5, 0x2, 0x40f86a0000000000, ...}, ...)
        /home/muralisr/manetu/pkg/mod/go.temporal.io/sdk@v1.33.0/internal/internal_task_handlers.go:562 +0x54
go.temporal.io/sdk/internal.newWorkflowWorkerInternal(_, {{0x1d42a8c, 0x7}, {0x1dd9731, 0x18}, {0x2383840, 0xc000a894f0}, 0x40f86a0000000000, 0x5, 0x2, ...}, ...)
        /home/muralisr/manetu/pkg/mod/go.temporal.io/sdk@v1.33.0/internal/internal_worker.go:320 +0xd0
go.temporal.io/sdk/internal.newWorkflowWorker(...)
        /home/muralisr/manetu/pkg/mod/go.temporal.io/sdk@v1.33.0/internal/internal_worker.go:250
go.temporal.io/sdk/internal.NewAggregatedWorker(_, {_, _}, {0xc8, 0x0, 0xc8, 0x0, 0x0, 0x5, 0xc8, ...})
        /home/muralisr/manetu/pkg/mod/go.temporal.io/sdk@v1.33.0/internal/internal_worker.go:1828 +0xeea
go.temporal.io/sdk/internal.NewWorker(...)
        /home/muralisr/manetu/pkg/mod/go.temporal.io/sdk@v1.33.0/internal/worker.go:360
[...]

PR #1892, opened 2025-03-27
Add FIPs to CI

Quinn-With-Two-Ns:
Switch from using `md5` to `sha256` to in `initBinaryChecksumLocked` so the Go SDK will work with `fips140` mode. I also added this to our CI when we connect to Temporal cloud over TLS

see also https://go.dev/doc/security/fips140

closes https://github.com/temporalio/sdk-go/issues/1887

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -28,7 +28,7 @@ package internal
 
 import (
 	"context"
-	"crypto/md5"
+	"crypto/sha256"
 	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


cretz (approved):

internal/internal_worker.go
```diff
@@ -1235,7 +1235,7 @@ func initBinaryChecksumLocked() error {
 		_ = f.Close() // error is unimportant as it is read-only
 	}()
 
-	h := md5.New()
+	h := sha256.New()
```
cretz, line 1238: Doubles the checksum size, but I don't think that's a big deal, nor do I think performance switching from md5 to sha256 is a concern

PR #1894, opened 2025-03-27
Use RawValue for built-in query responses

### LINKED ISSUE #1843: Built-in query responses should use "RawValue"

See https://github.com/temporalio/features/issues/604. When #1231 is done, make sure build-in queries return a raw value

yuandrew:
## What was changed

Use RawValue for built-in query response, only `QueryTypeWorkflowMetadata` is being changed due to backwards compatibility issues

## Why?

So user's payload converter is not used

## Checklist

1. Closes #1843 

2. How was this tested:

Added an integration test to test `__temporal_workflow_metadata` 

3. Any docs updates needed?


internal/internal_event_handlers.go
```diff
@@ -1415,9 +1415,19 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessQuery(
 ) (*commonpb.Payloads, error) {
 	switch queryType {
 	case QueryTypeStackTrace:
-		return weh.encodeArg(weh.StackTrace())
+		// Use raw value built from default converter because we don't want to use user-conversion
+		payload, err := converter.GetDefaultDataConverter().ToPayload(weh.StackTrace())
```
Quinn-With-Two-Ns, line 1419: For backwards compatibility I don't think we can change these, only the `QueryTypeWorkflowMetadata` can use `RawValue`

yuandrew: Could you explain more? How come these can't change, but QueryTypeWorkflowMetadata can? Is this because users are already relying on their user payloads to process these built in requests?

If this is a compat issue, shouldn't that be the case for all SDKs?

cretz: > How come these can't change, but QueryTypeWorkflowMetadata can? Is this because users are already relying on their user payloads to process these built in requests?

Yes, and their payload converter may not support RawValue which means stack traces stop working when this is released. We may also need to be prepared for users complaining about workflow metadata queries showing errors in their logs at which point we need to advise them to support raw value in their converter.

> If this is a compat issue, shouldn't that be the case for all SDKs?

It will be the case for most SDKs I suspect, we should probably update the features issue. The two SDKs this is already done in have had raw value support since their inception and have documented that custom-payload-converter users should support them.

internal/internal_workflow.go
```diff
@@ -623,7 +623,13 @@ func (d *syncWorkflowDefinition) Execute(env WorkflowEnvironment, header *common
...
 					return nil, err
 				} else {
-					return encodeArg(getDataConverterFromWorkflowContext(rootCtx), result)
+					// Use raw value built from default converter because we don't want to use
+					// user-conversion
+					resultPayload, err := converter.GetDefaultDataConverter().ToPayload(result)
+					if err != nil {
+						return nil, err
+					}
+					return encodeArg(getDataConverterFromWorkflowContext(rootCtx), converter.NewRawValue(resultPayload))
```
Quinn-With-Two-Ns, line 632: @cretz do you also have special processing on the client side to match the worker side? Since now the worker will use raw value but right now the client doesn't

cretz: The only clients for user metadata at this time are UI and CLI, both of which can read this object.

Quinn-With-Two-Ns: Any SDK can make a issue a system query. I am asking how when you did RawValues in other SDKs did you put special processing on the client side in addition to the worker side?

cretz: Are you asking if we added processing to client side to do special things on `__stack_trace` and `__temporal_workflow_metadata` queries unlike other queries? No we did not. Or are you asking in general when wrote `RawValue` support in SDKs if we did any special processing on _any_ side, then the answer is no, not on client side or worker side, only in the converter (well some SDKs have data type expectations for dynamic things, but not really). So, say in Go, if you have a worker that does not use raw value, you can still `var rawVal RawValue`, `myRun.Get(ctx, &rawVal)` and it will work too.

cretz (approved):
LGTM, but will want to wait for @Quinn-With-Two-Ns's approval

internal/internal_workflow.go
Quinn-With-Two-Ns: >Are you asking if we added processing to client side to do special things on __stack_trace and __temporal_workflow_metadata queries unlike other queries? 

Yes

>No we did not because we don't expect people to manually call these queries 

Well the CLI will call this query so that would need to support `RawValues` No?

internal/internal_workflow.go
```diff
@@ -623,7 +623,13 @@ func (d *syncWorkflowDefinition) Execute(env WorkflowEnvironment, header *common
 				if result, err := getWorkflowMetadata(rootCtx); err != nil {
 					return nil, err
 				} else {
-					return encodeArg(getDataConverterFromWorkflowContext(rootCtx), result)
+					// Use raw value built from default converter because we don't want to use
```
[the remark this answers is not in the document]
yuandrew: How would a data converter not accept `proto`?

internal/internal_workflow.go
cretz: The CLI already uses the default payload converter which supports this proto type. We actually do have a homemade raw-value form in CLI for other things, but that's unrelated. Raw value is not a construct of the payload it is a construct of the code, and in this case we only use raw value because they may not support proto types, but CLI does support them.

PR #1896, opened 2025-03-28
Make sure schedule Update treats CatchupWindow properly

Quinn-With-Two-Ns:
Make sure schedule Update treats CatchupWindow properly. When we create a schedule we treat a `CatchupWindow=0` as unset, so we pass `nil` which tells the server to use the default value. On update we forgot to convert `CatchupWindow=0` causing the server to use the min. value of `10s` instead.

https://github.com/temporalio/sdk-go/blob/31c4c30812adcd53554f08dfc5ca917df718ceae/internal/internal_schedule_client.go#L124

### DIFF
--- a/internal/internal_schedule_client.go
+++ b/internal/internal_schedule_client.go
@@ -547,8 +547,7 @@ func convertToPBSchedule(ctx context.Context, client *WorkflowClient, schedule *
 
 	var catchupWindow *durationpb.Duration
 	if schedule.Policy.CatchupWindow != 0 {
-		// Convert to nil so the server uses the default
-		// catchup window.
+		// Only convert non-zero CatchWindow so server treats 0 as nil and uses the default catchup window.
 		catchupWindow = durationpb.New(schedule.Policy.CatchupWindow)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/internal_schedule_client.go
```diff
@@ -544,12 +544,20 @@ func convertToPBSchedule(ctx context.Context, client *WorkflowClient, schedule *
 	if err != nil {
 		return nil, err
 	}
+
+	var catchupWindow *durationpb.Duration
+	if schedule.Policy.CatchupWindow != 0 {
+		// Convert to nil so the server uses the default
```
yuandrew, line 550: This comment confuses me, shouldn't this be something more like 
`// Only convert non-zero CatchWindow so server treats 0 as nil and uses the default catchup window.`

Since this if block is handling the non-zero case?

yuandrew (approved):
Small comment on code comment, otherwise LGTM

internal/internal_schedule_client.go
Quinn-With-Two-Ns: Yeah that is clearer, agree!

Issue #1874, opened 2025-03-18
Document that local activities should not use anonymous functions


Quinn-With-Two-Ns:
We should document  that local activities should not use anonymous functions, even though it is technically allowed. This is because the function name, and therefore activity type, can be non deterministic and it is easy for the local activity to capture local variable which is also dangerous

Issue #1341, opened 2024-01-09
WorkflowChecker should consider anonymous functions in local activities deterministic, but maybe warn about them anyway 


Quinn-With-Two-Ns:
## Expected Behavior
WorkflowChecker should consider anonymous functions in local activities deterministic.

## Actual Behavior
WorkflowChecker does not consider anonymous functions in local activities deterministic.

## Steps to Reproduce the Problem

```Go
func GreetingSample(ctx workflow.Context) error {
	fut := workflow.ExecuteLocalActivity(ctx, func(ctx context.Context) time.Time { return time.Now() })
}
```

## Note

Anonymous functions as local activities are ripe for data races because they can run in parallel with the workflow function, so maybe we should warn about their use from that perspective.

PR #1897, opened 2025-03-28
Add warning about anonymous functions as local activities

Quinn-With-Two-Ns:
Add warning about anonymous functions as local activities. I tried to see if we could warn if they were used, unfortunatly it doesn't seem like there is a good way to identify an anonymous function in the Go runtime

closes https://github.com/temporalio/sdk-go/issues/1874


cretz (approved):
> unfortunatly it doesn't seem like there is a good way to identify an anonymous function in the Go runtime

We could consider this an error in `workflowcheck` I think (I don't think it's backwards incompatible to start erroring on something in that tool we believe is wrong).

Quinn-With-Two-Ns:
Yeah we have another issue for that https://github.com/temporalio/sdk-go/issues/1341

PR #1902, opened 2025-04-03
Add Task Queue Priority

Quinn-With-Two-Ns:
Add Task Queue Priority


internal/workflow.go
```diff
@@ -1862,6 +1875,15 @@ func WithDataConverter(ctx Context, dc converter.DataConverter) Context {
...
 }
 
+// WithPriority adds a priority to the context.
+//
+// Exposed as: [go.temporal.io/sdk/workflow.WithWorkflowPriority]
+func WithWorkflowPriority(ctx Context, priority Priority) Context {
+	ctx1 := setWorkflowEnvOptionsIfNotExist(ctx)
+	getWorkflowEnvOptions(ctx1).Priority = convertToPBPriority(priority)
+	return ctx1
+}
```
cretz, line 1885: IMO this is advanced enough that users can just set it via `WithChildOptions` (I kinda wish we did that with versioning intent too). Not sure what the bar is for each child option getting its own `With` call, but I figure these settings are surely not more popular than a `ParentClosePolicy`.

internal/workflow.go
```diff
@@ -2497,6 +2521,15 @@ func WithRetryPolicy(ctx Context, retryPolicy RetryPolicy) Context {
...
 }
 
+// WithPriority adds priority to the copy of the context.
+//
+// Exposed as: [go.temporal.io/sdk/workflow.WithPriority]
+func WithPriority(ctx Context, priority Priority) Context {
+	ctx1 := setActivityParametersIfNotExist(ctx)
+	getActivityOptions(ctx1).Priority = convertToPBPriority(priority)
+	return ctx1
+}
```
cretz, line 2531: Same concern here, does this activity option need its own `With` helper? Added benefit of making people use `WithChildOptions` and `WithActivityOptions` to set this value is that it is not confusing which is for activity and which is for child (though I know that confusion already exists with some others like retry policy).

test/integration_test.go
```diff
@@ -6976,6 +6976,25 @@ func (ts *IntegrationTestSuite) TestUserMetadata() {
 	ts.Equal("my-child-wf-details", str)
 }
 
+func (ts *IntegrationTestSuite) TestTaskQueuePriority() {
```
cretz, line 6979: Would be neat to have a simple behavior test to confirm it works, but a test confirming server has received it is good enough

cretz (approved):
Looks great, nothing blocking

internal/workflow.go
Quinn-With-Two-Ns: We have these helpers for almost all options, there is a clear patterns that all `WithWorkflow` are for workflow and `With` are for activity if there is ambiguity

PR #1903, opened 2025-04-04
Add support for activity pause

Quinn-With-Two-Ns:
Add support for activity pause. Makes use of `context.Cause` to tell users why exactly their Activity was cancelled 

closes: https://github.com/temporalio/sdk-go/issues/1899

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -604,6 +604,41 @@ func (ts *IntegrationTestSuite) TestHeartbeatOnActivityFailure() {
 	ts.Equal(6, heartbeatCounts)
 }
 
+func (ts *IntegrationTestSuite) TestActivityPause() {
+	ctx, cancel := context.WithTimeout(context.Background(), ctxTimeout)
+	defer cancel()
+	// Run ActivityHeartbeat workflow, this workflow will call
+	// ActivityToBePaused activity twice, the first call will test pausing an activity successfully
+	// and the second call will test completing the activity after it is resumed
+	run, err := ts.client.ExecuteWorkflow(ctx,
+		ts.startWorkflowOptions("test-activity-pause"), ts.workflows.ActivityHeartbeat)
+	ts.NoError(err)
+	// Wait for the workflow to finish
+	var result string
+	err = run.Get(ctx, &result)
+	ts.NoError(err)
+	// Check the result
+	ts.Equal("I am stopped by Pause", result)
+	// Verify that the activity was called twice
+	expectedActivities := []string{"ActivityToBePaused", "ActivityToBePaused"}
+	ts.EqualValues(expectedActivities, ts.activities.invoked())
+	// Describe the workflow execution
+	desc, err := ts.client.WorkflowService().DescribeWorkflowExecution(ctx, &workflowservice.DescribeWorkflowExecutionRequest{
+		Namespace: ts.config.Namespace,
+		Execution: &commonpb.WorkflowExecution{
+			WorkflowId: run.GetID(),
+			RunId:      run.GetRunID(),
+		},
+	})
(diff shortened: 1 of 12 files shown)


test/integration_test.go
```diff
@@ -604,6 +604,41 @@ func (ts *IntegrationTestSuite) TestHeartbeatOnActivityFailure() {
...
+			WorkflowId: run.GetID(),
+			RunId:      run.GetRunID(),
+		},
+	})
+	ts.NoError(err)
+	// Check the workflow still has one paused pending activity
+	ts.Len(desc.GetPendingActivities(), 1)
+	ts.Equal(desc.GetPendingActivities()[0].GetActivityType().GetName(), "ActivityToBePaused")
+	ts.Equal(desc.GetPendingActivities()[0].GetAttempt(), int32(1))
+	ts.Nil(desc.GetPendingActivities()[0].GetLastFailure()) // Verify that the activity paused successfully
```
cretz, line 638: Hrmm, it has been asked but not decided yet what this looks like. I think we should be a bit more detailed on the assertion here of what the activity failure of a pause interruption looks like. Is it a canceled failure or an application failure, and if the latter, if at all possible to have SDKs match their "type" strings here that'd be ideal.

Quinn-With-Two-Ns: The API described how the SDK should respond to a failure, I don't see how this assertion can be more detailed

test/workflow_test.go
```diff
@@ -493,6 +493,23 @@ func (w *Workflows) ActivityHeartbeatWithRetry(ctx workflow.Context) (heartbeatC
...
 }
 
+func (w *Workflows) ActivityHeartbeat(ctx workflow.Context) (string, error) {
+	ctx = workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		StartToCloseTimeout: 4 * time.Second,
+		HeartbeatTimeout:    2 * time.Second,
+	})
+
+	var activities *Activities
+	_ = workflow.ExecuteActivity(ctx, activities.ActivityToBePaused, false)
```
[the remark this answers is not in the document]
cretz: Ok, confirmed timeouts don't apply while paused, but will apply immediately on unpause.

test/integration_test.go
Quinn-With-Two-Ns: How can I check the type of a nil error?

cretz: Oh, sorry, I misread that as `NotNil`. I am not sure this is right. Hrmm, I am not sure this is correct behavior to have a nil last failure. Do last heartbeat details at least get preserved? Users expect any attempt > 1 to have a last failure, but this should an attempt could be something like 5, get paused, and have no failure.

Can we get confirmation this is the behavior we want?

cretz: Ok, confirmed internally that we don't expect last_failure to be null. Will put server GH issue link in here once available. Want to leave assertion as is (to break on server upgrade) or want to comment out assertion with this link (when available)?

cretz (approved):
Approving. Some things decided during this SDK PR that can apply to others:

* We are imposing no SDK rules on what a pause error looks like (matters more to SDKs that will be able to customize the error more clearly)
* There is a server-side bug where pending activity last failure info is null after pause which will be getting fixed

test/integration_test.go
cretz: Note, this is fixed in https://github.com/temporalio/temporal/pull/7572 and once released/used, tests relying on last failure to be nil will start failing.

Quinn-With-Two-Ns:
>Heads up you may want to incorporate https://github.com/temporalio/api/pull/568 before merging

This API change doesn't effect this PR

Issue #1156, opened 2023-07-06
Activity cancel error returned when not server-side cancelled is rejected by server


cretz:
## Expected Behavior

When a worker stops while an activity is running and that context is cancelled and bubbled out as an activity error, that activity attempt should be recorded as failed server side. This is how it works in most SDKs today.

## Actual Behavior

Today if you stop a worker, the activity tries to return to the server the cancelled error but the server fails the activity task completion with:

> unable to mark activity as canceled without activity being request canceled first

I think we need to send 1) only send activity cancelled if the server requested cancel (otherwise wrap as application error if canceled error received), and 2) write a test proving that worker stop during a non-heartbeating activity does the next attempt on the next worker.

cretz:
This maybe could be done at the same time as #1086

josh-berry:
Am I correct in thinking the impact of this is just performance—that is, it takes longer overall to successfully complete the activity than it otherwise might?

My mental model of this is:

1. Activity cancellation error gets sent to server
2. Server fails the task completion, and therefore doesn't realize the activity was canceled
3. Eventually the activity times out server-side (<-- this is where the extra time goes)
4. Server retries the activity (hopefully on a new worker since the old one died)
5. Eventually the activity completes (if nothing else goes wrong)

lmk what I got wrong. :)

cretz:
Yes, that seems correct. I think in most Core-based SDKs, we already convert non-server-requested-activity-cancellation errors (i.e. worker shutdown) to activity errors. I think we probably just confirm and do this in all SDKs.

dmateusp:
I've just hit the same behavior. I'm cancelling the activity context in an interceptor when the graceful shutdown of workers is initiated, but then I see the same `unable to mark activity as canceled without activity being request canceled first` and Temporal waits for the `StartToCloseTimeout` before attempting the activity again.

Ideally we'd like the next attempt to start as soon as a new worker has rolled out.

This is a snippet from my interceptor code for reference:

```go
	gracefulShutdownChan := activity.GetWorkerStopChannel(ctx)

	activityCtx, activityCtxCancelFunc := context.WithCancel(ctx)
	defer activityCtxCancelFunc()

	go func() {
		<-gracefulShutdownChan
		time.Sleep(10 * time.Second) // Give the running activities some time to finish
		activityCtxCancelFunc()
	}()
	result, err := a.Next.ExecuteActivity(activityCtx, in)
```

[...]

PR #1904, opened 2025-04-05
Add worker shutdown cause to context

Quinn-With-Two-Ns:
Add worker shutdown cause to context. This will help users tell why their activity/local activity context was cancelled. Before this PR when an activity/local activity was shutting down at the end of the shutdown process the SDK would cancel the context, but users couldn't distinguish this cancel vs other reasons the cancel may be cancelled.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -162,7 +162,7 @@ type (
 		client                           *WorkflowClient
 		metricsHandler                   metrics.Handler
 		logger                           log.Logger
-		userContext                      context.Context
+		backgroundContext                context.Context
 		registry                         *registry
 		activityProvider                 activityProvider
 		dataConverter                    converter.DataConverter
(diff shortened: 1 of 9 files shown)


cretz (approved):

internal/internal_task_handlers.go
```diff
@@ -2272,7 +2273,8 @@ func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice
 
 	output, err := activityImplementation.Execute(ctx, t.Input)
 	// Check if context canceled at a higher level before we cancel it ourselves
-	isActivityCancel := ctx.Err() == context.Canceled
+	// TODO : check if the cause of the context cancellation is from the server
```
yuandrew, line 2276: I can probably confirm this when I work on #1156

yuandrew (approved):

Issue #1829, opened 2025-02-18
doclink: Incorrect "Exposed" requirement in method without context


antlai-temporal:
##  Actual Behavior
```
	// WorkerDeploymentHandle is a handle to a Worker Deployment.
	// NOTE: Experimental
	//
	// Exposed as: [go.temporal.io/sdk/client.WorkerDeploymentHandle]
	WorkerDeploymentHandle interface {
        ...
        }
	// WorkerDeploymentClient is the client that manages Worker Deployments.
	// NOTE: Experimental
	//
	// Exposed as: [go.temporal.io/sdk/client.WorkerDeploymentClient]
	WorkerDeploymentClient interface {
		// GetHandle returns a handle to a Worker Deployment.
		//
		// This method does not validate the Worker Deployment Name. If there is no deployment
		// with that name in this namespace, methods like WorkerDeploymentHandle.Describe()
		// will return an error.
		// NOTE: Experimental
		//
		// Exposed as: [go.temporal.io/sdk/client.WorkerDeploymentHandle]    <----- WRONG, not needed, added by -fix
		GetHandle(name string) WorkerDeploymentHandle
       }
```

Missing doc in internal/worker_deployment_client.go for internal:WorkerDeploymentHandle to client:WorkerDeploymentHandle
2025/02/17 20:19:49 Changes needed, see previous stdout for which objects. Re-run command with -fix to auto-generate new docs.

Got above after running `-fix`

## Expected Behavior
It should not add/require Exposed in GetHandle

See https://github.com/temporalio/sdk-go/pull/1814

PR #1909, opened 2025-04-09
[doclink] Fix interface parsing

yuandrew:
## What was changed

Fixed a bug where the tool was checking inside interface definitions, and marking it as exposed publicly.

Also found a bug where our line indentation count was only counting spaces, not tabs.

## Why?

Fix bug

## Checklist

1. Closes  #1829 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/cmd/tools/doclink/doclink.go
+++ b/internal/cmd/tools/doclink/doclink.go
@@ -34,6 +34,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"unicode"
 )
 
 type (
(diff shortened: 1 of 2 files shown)


internal/cmd/tools/doclink/doclink.go
```diff
@@ -578,7 +582,7 @@ func isValidDefinitionWithMatch(line, private string, inGroup string, insideStru
 	}
 
 	if insideStruct {
-		fmt.Println("should never hit")
+		panic("should never hit")
 		return false
```
cretz, line 586:
```suggestion
```

cretz (approved):

PR #1910, opened 2025-04-09
Handle activity cancel due to worker shutdown properly

yuandrew:
## What was changed
Stop sending activity cancellation due to worker shutdown. Now we send an Application error, allowing for another worker to retry the activity.

## Why?

align behavior with core SDKs

## Checklist

1. Closes  #1156

2. How was this tested:

Added integration test, shutdown worker in the middle of activity, and start a new worker to see the workflow complete.

3. Any docs updates needed?

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -2240,7 +2240,7 @@ func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice
 		metricsHandler.Counter(metrics.UnregisteredActivityInvocationCounter).Inc(1)
 		return convertActivityResultToRespondRequest(ath.identity, t.TaskToken, nil,
 			NewActivityNotRegisteredError(activityType, ath.getRegisteredActivityNames()),
-			ath.dataConverter, ath.failureConverter, ath.namespace, false, ath.versionStamp, ath.deployment, ath.workerDeploymentOptions), nil
+			ath.dataConverter, ath.failureConverter, ath.namespace, false, ath.versionStamp, ath.deployment, ath.workerDeploymentOptions, false), nil
 	}
 
 	// panic handler
@@ -2258,7 +2258,7 @@ func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice
 			metricsHandler.Counter(metrics.ActivityTaskErrorCounter).Inc(1)
 			panicErr := newPanicError(p, st)
 			result = convertActivityResultToRespondRequest(ath.identity, t.TaskToken, nil, panicErr,
-				ath.dataConverter, ath.failureConverter, ath.namespace, false, ath.versionStamp, ath.deployment, ath.workerDeploymentOptions)
+				ath.dataConverter, ath.failureConverter, ath.namespace, false, ath.versionStamp, ath.deployment, ath.workerDeploymentOptions, false)
 		}
 	}()
 
@@ -2276,6 +2276,7 @@ func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice
 	// Check if context canceled at a higher level before we cancel it ourselves
 	// TODO : check if the cause of the context cancellation is from the server
 	isActivityCanceled := ctx.Err() == context.Canceled
+	canceledFromWorkerShutdown := errors.Is(context.Cause(ctx), ErrWorkerShutdown)
 
 	dlCancelFunc()
 	if <-ctx.Done(); ctx.Err() == context.DeadlineExceeded {
@@ -2299,7 +2300,7 @@ func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice
 		)
 	}
 	return convertActivityResultToRespondRequest(ath.identity, t.TaskToken, output, err,
-		ath.dataConverter, ath.failureConverter, ath.namespace, isActivityCanceled, ath.versionStamp, ath.deployment, ath.workerDeploymentOptions), nil
+		ath.dataConverter, ath.failureConverter, ath.namespace, isActivityCanceled, ath.versionStamp, ath.deployment, ath.workerDeploymentOptions, canceledFromWorkerShutdown), nil
 }
 
 func (ath *activityTaskHandlerImpl) getActivity(name string) activity {
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -1211,6 +1211,7 @@ func convertActivityResultToRespondRequest(
 	versionStamp *commonpb.WorkerVersionStamp,
 	deployment *deploymentpb.Deployment,
 	workerDeploymentOptions *deploymentpb.WorkerDeploymentOptions,
+	canceledFromWorkerShutdown bool,
 ) interface{} {
 	if err == ErrActivityResultPending {
 		// activity result is pending and will be completed asynchronously.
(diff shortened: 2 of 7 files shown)


internal/internal_task_pollers.go
```diff
@@ -1245,6 +1246,21 @@ func convertActivityResultToRespondRequest(
 			}
 		}
 		if errors.Is(err, context.Canceled) {
+			// Cancelation due to worker shutdown should be wrapped in an
+			// Application Error, so we don't waste time waiting for the
+			// activity to timeout from the server side.
+			if canceledFromWorkerShutdown {
```
Quinn-With-Two-Ns, line 1252: There is already code to handle this in the SDK, look at `cancelAllowed` we shouldn't add another flag, just fix the broken flag

Quinn-With-Two-Ns:
The context can be cancelled for other reasons that are not shutdown, but are also not cancellation from the server (like pause). I don't think this issue can be fixed until https://github.com/temporalio/sdk-go/pull/1903 is merged and then you can inspect the cause of the context and use that to discriminate.

internal/internal_task_pollers.go
```diff
@@ -1207,7 +1207,7 @@ func convertActivityResultToRespondRequest(
 	dataConverter converter.DataConverter,
 	failureConverter converter.FailureConverter,
 	namespace string,
-	cancelAllowed bool,
```
Quinn-With-Two-Ns, line 1210: I don't think you need to change this, you can just change how `cancelAllowed` is computed

internal/internal_task_handlers.go
```diff
@@ -2285,8 +2285,7 @@ func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice
 
 	output, err := activityImplementation.Execute(ctx, t.Input)
 	// Check if context canceled at a higher level before we cancel it ourselves
-	// TODO : check if the cause of the context cancellation is from the server
-	isActivityCanceled := ctx.Err() == context.Canceled
+	isActivityCanceled := ctx.Err() == context.Canceled && errors.Is(context.Cause(ctx), context.Canceled)
```
Quinn-With-Two-Ns, line 2288: The `cause` of a cancelled activity should be a `*CanceledError` no?

test/integration_test.go
```diff
@@ -7510,3 +7510,32 @@ func (ts *IntegrationTestSuite) TestRawValueQueryMetadata() {
 	ts.Equal("Basic", metadata.Definition.Type)
 	ts.Equal(3, len(metadata.Definition.QueryDefinitions))
 }
+
+func (ts *IntegrationTestSuite) TestActivityCancelServer() {
```
Quinn-With-Two-Ns, line 7514: Nit: I don't think this test name is very clear since you are not testing an activity is canceled by the server

test/integration_test.go
```diff
@@ -7510,3 +7510,32 @@ func (ts *IntegrationTestSuite) TestRawValueQueryMetadata() {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	options := client.StartWorkflowOptions{
+		ID:                       "test-activity-cancel" + uuid.NewString(),
+		TaskQueue:                ts.taskQueueName,
+		WorkflowExecutionTimeout: 6 * time.Second,
+		WorkflowTaskTimeout:      time.Second,
+		WorkflowIDReusePolicy:    enumspb.WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE,
+		EnableEagerStart:         true,
```
Quinn-With-Two-Ns, line 7524: Why enable eager start?

test/workflow_test.go
```diff
@@ -3505,6 +3505,20 @@ func (w *Workflows) WorkflowRawValue(ctx workflow.Context, value converter.RawVa
...
+	ctx = workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		ScheduleToCloseTimeout: 30 * time.Second,
+		RetryPolicy: &temporal.RetryPolicy{
+			InitialInterval:    1 * time.Second,
+			BackoffCoefficient: 1.0,
+			MaximumAttempts:    3,
+		},
+	})
+	var activities *Activities
+	_ = workflow.ExecuteActivity(ctx, activities.ReactToCancel, time.Second*2).Get(ctx, nil)
```
Quinn-With-Two-Ns, line 3518: We should also test local activity to

yuandrew: I believe the LA version of this test is not going to work due to #1706, I'm planning on fixing that next, will add a test for that when that issue gets fixed

test/integration_test.go
yuandrew: copy paste fail

test/workflow_test.go
yuandrew: scratch that, I think this test can be added to this PR

internal/internal_task_handlers.go
```diff
@@ -1009,6 +1009,10 @@ processWorkflowLoop:
 					case lar := <-workflowTask.laResultCh:
 						// local activity result ready
 						response, err = workflowContext.ProcessLocalActivityResult(workflowTask, lar)
+						var appErr *ApplicationError
+						if errors.As(lar.err, &appErr) {
+							break processWorkflowLoop
+						}
```
yuandrew, line 1015: This is the fix for LAs heartbeating after they've been canceled. Resolving that bug is required to get Local Activities working for non-server activity cancel

Quinn-With-Two-Ns: I don't think this is correct, just because the activity returned a certain error type doesn't mean it was because of the worker shutting down. Can we resolve #1706 in a separate PR?

internal/session.go
```diff
@@ -441,6 +441,14 @@ func sessionCreationActivity(ctx context.Context, sessionID string) error {
 		select {
 		case <-ctx.Done():
 			sessionEnv.CompleteSession(sessionID)
+			// Because of how session creation configures retryPolicy, we need to wrap context cancels that don't
+			// originate from the server as non-retryable errors. See retrypolicy in createSession() above.
+			if !(ctx.Err() == context.Canceled && errors.Is(context.Cause(ctx), &CanceledError{})) {
+				return NewApplicationErrorWithOptions(
+					"context canceled due to worker shutdown", "sessionCreationActivity",
```
Quinn-With-Two-Ns, line 448: Nit: I don't think `sessionCreationActivity` is a very clear failure type. It is just repeating the activity type. Maybe something like `SessionWorkerShutdown`. I would also clarify in the error message that the session is failing because the worker is shutdown.

test/integration_test.go
```diff
@@ -7522,3 +7522,142 @@ func (ts *IntegrationTestSuite) TestRawValueQueryMetadata() {
...
 }
+
+func (ts *IntegrationTestSuite) TestActivityCancelFromWorkerShutdown() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	options := client.StartWorkflowOptions{
+		ID:                       "test-activity-cancel" + uuid.NewString(),
+		TaskQueue:                ts.taskQueueName,
+		WorkflowExecutionTimeout: 6 * time.Second,
```
Quinn-With-Two-Ns, line 7533: Are the workflow options needed for the test?

test/integration_test.go
```diff
@@ -7522,3 +7522,142 @@ func (ts *IntegrationTestSuite) TestRawValueQueryMetadata() {
 	ts.Equal("Basic", metadata.Definition.Type)
 	ts.Equal(3, len(metadata.Definition.QueryDefinitions))
 }
+
+func (ts *IntegrationTestSuite) TestActivityCancelFromWorkerShutdown() {
```
Quinn-With-Two-Ns, line 7526: Does this test fail before your changes? How does it fail?

Quinn-With-Two-Ns:
@yuandrew Is this PR still active? I think you opened up a different PR for https://github.com/temporalio/sdk-go/pull/1910#discussion_r2054728426?

yuandrew:
@Quinn-With-Two-Ns Yes, I was waiting for https://github.com/temporalio/sdk-go/pull/1933 to be resolved before I ensure Local Activities behave correctly for this

test/integration_test.go
yuandrew: no, these tests don't fail before my changes, removing, will try to come up with a different test

internal/internal_task_handlers.go
yuandrew: was resolved in a separate PR

internal/internal_task_handlers.go
yuandrew: Why would cause be `*CanceledError`? Aren't the only cause today `ErrWorkerShutdown` and `ErrActivityPaused`, which are plain `errors.New()`?

yuandrew: oh, I misunderstood, `isActivityCanceled` is only true when server has canceled the activity, so we don't want to be `true` if shutdown or pause occurred

test/integration_test.go
yuandrew: new tests are confirmed to hit `unable to mark activity as canceled without activity being request canceled first` without fix

Issue #1330, opened 2023-12-20
Infer peer services in tracing interceptors


caramelomartins:
Hey! I'll stick to template.

**Is your feature request related to a problem? Please describe.**

OpenTelemetry allows specifying "a remote service" to which a connection is made [through a peer service name](https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#peer-service-name). This can be helpful to determine interactions between services - namely workers - even if Temporal's architecture is asynchronous.

Currently,  none of the available options in [TracerStartSpanOptions](https://github.com/temporalio/sdk-go/blob/5fdbecc56c8cd3bab8e8e33f7073070b98cefc0c/interceptor/tracing_interceptor.go#L123) allows us to infer if a worker is making a request that will be fulfilled by another worker. This makes it impossible to add peer service information in outbound calls (e.g. `StartActivity` or `StartWorkflow`).

**Describe the solution you'd like**

One potential approach could be to share the `TaskQueue` as a `TracerStartSpanOption`, similar to what is already done with the Name and Operation being executed. This would allow developers to externally infer if a cross-service request is being made based on whether the `TaskQueue` of the outbound request is different than the `TaskQueue` where a worker is executing now.

I'm sure this isn't an appropriate solution for all situations, but that could be left to the developer to decide - depending on the architecture of their services. Just providing the `TaskQueue` there could potentially open space for attempting to infer peer services.

**Describe alternatives you've considered**

I considered using the Name as the peer service but that value changes to often to be able to provide a meaningful signal for cross-service interactions. Nonetheless, I'm very open to other suggestions about how to go about doing this.

I'm happy to open a PR and work on this, but wouldn't want to do that before opening a feature request and discussing it.

Thank you!

cretz:
It's not just task queue, there are a lot of things people may want as tags or customize the span.

I wonder if there is a way to provide the context to `TracerStartSpanOptions` (that is probably `interface{ Value(any) any }`) so that you can extract anything out of the context. For a workflow context this would be easy enough to get the info, but for a client context, you may be expected to provide something to the context before starting.

Another approach may be to have a special context key for storing tags on the context, and you can have an outer interceptor that sets this task queue that is then read by the span starter.

caramelomartins:
@cretz I mentioned `TaskQueue` in attempt to not overload the feature request with a lot of things. Your suggestion is sensible though, I feel users will increasingly want to have more and more data - likely.

[...]

PR #1911, opened 2025-04-11
Add temporal relevant attributes to spans

ankitnayan:
## What was changed
Added `workflowType`, `activityType`, `namespace`, `taskQueue` attributes to `ExecuteActivity` and `ExecuteWorkflow` spans

## Why?
Important attributes relevant to temporal were missing from the tracing data for better analysis

## Checklist

1. How was this tested:
Tested locally by enabling trace intercept and sending data to SigNoz

2. Any docs updates needed?
None

### DIFF
--- a/interceptor/tracing_interceptor.go
+++ b/interceptor/tracing_interceptor.go
@@ -42,6 +42,10 @@ const (
 	runIDTagKey      = "temporalRunID"
 	activityIDTagKey = "temporalActivityID"
 	updateIDTagKey   = "temporalUpdateID"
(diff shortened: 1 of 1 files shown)


cretz:
I am not sure we want to add this information for every user of tracing of the Go SDK. Commercial telemetry providers may charge per attribute on a span, so to sometimes more than double attribute counts on users just because they upgrade SDK versions is a bit rough.

We may be able to discuss options on how you can customize the attributes at span creation time. See #1330 where there has been some discussion here.

PR #1917, opened 2025-04-16
Implement Nexus operation cancellation types

pdoerner:
## What was changed
Implemented cancellation types for Nexus operations which specify what action to take when the caller context is cancelled.
* `ABANDON` -> Do not request cancellation of the operation.
* `TRY_CANCEL` -> Initiate a cancellation request and immediately report cancellation to the caller.
* `WAIT_REQUESTED` -> Request cancellation of the operation and wait for confirmation that the request was received.
* `WAIT_COMPLETED` -> Wait for operation completion. Default.
These options are set on `workflow.NexusOperationOptions` which are passed in when starting an operation.

Depends on https://github.com/temporalio/api/pull/572

## Why?
To give users more control over what happens to their handler workflows when their caller workflows are cancelled.

## Checklist

1. Closes 

2. How was this tested:
Updated functional tests

3. Any docs updates needed?


internal/internal_event_handlers.go
```diff
@@ -666,17 +668,24 @@ func (wc *workflowEnvironmentImpl) ExecuteNexusOperation(params executeNexusOper
...
-			data.startedCallback("", ErrCanceled)
-			data.startedCallback = nil
-		}
-		if data.completedCallback != nil {
-			data.completedCallback(nil, ErrCanceled)
-			data.completedCallback = nil
+	cancelCmd := command.(*nexusOperationStateMachine).cancelation
+
+	// Make sure to unblock the futures if the caller has indicated they do not want to wait.
+	if cancelCmd.getState() == commandStateCreated || cancelCmd.getState() == commandStateCommandSent || cancelCmd.getState() == commandStateInitiated {
```
pdoerner, line 674: I'm pretty sure this was a bug. We were only immediately unblocking the futures when the operation had not been started. Otherwise they would block until operation completion.

internal/workflow.go
```diff
@@ -2630,6 +2664,11 @@ type NexusOperationOptions struct {
 	// Optional: defaults to the maximum allowed by the Temporal server.
 	ScheduleToCloseTimeout time.Duration
 
+	// CancellationType - Indicates what action should be taken when the caller is cancelled.
+	//
+	// Optional: defaults to NexusOperationCancellationTypeTryCancel.
+	CancellationType NexusOperationCancellationType
```
cretz, line 2670: Can you confirm for current released SDK without this PR, the behavior is the same as `TryCancel`?

pdoerner: It is `TryCancel` unless the operation has started, then it is `WaitCompleted`

workflow/workflow.go
```diff
@@ -67,6 +67,28 @@ const (
 	HandlerUnfinishedPolicyAbandon = internal.HandlerUnfinishedPolicyAbandon
 )
 
+// NexusOperationCancellationType specifies what action should be taken for a Nexus operation when the
+// caller is cancelled.
+type NexusOperationCancellationType = internal.NexusOperationCancellationType
+
+const (
+	// Nexus operation cancellation type is unknown.
+	NexusOperationCancellationTypeUnspecified NexusOperationCancellationType = iota
```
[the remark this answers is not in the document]
pdoerner: I don't have a strong opinion here but I thought having an explicit unset value might be useful if we decide to add more types or change the default in the future.

internal/internal_event_handlers.go
```diff
@@ -1984,6 +1996,44 @@ func (weh *workflowExecutionEventHandlerImpl) handleNexusOperationCompleted(even
...
 }
 
+func (weh *workflowExecutionEventHandlerImpl) handleNexusOperationCancelRequestDelivered(event *historypb.HistoryEvent) error {
+	var scheduledEventID int64
+	var failure *failurepb.Failure
+
+	switch event.EventType {
+	case enumspb.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUEST_COMPLETED:
+		attrs := event.GetNexusOperationCancelRequestCompletedEventAttributes()
+		scheduledEventID = attrs.GetScheduledEventId()
```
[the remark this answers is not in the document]
pdoerner: It looks like unfortunately that did happen. I will update this PR to handle that case.

internal/workflow.go
pdoerner: Hmm I guess that might be an issue? The history that is generated should be the same, but the operation future would get unblocked sooner with a canceled error. As I noted in another comment, it looked to me like it was a bug that we were waiting for completion after the operation had started. I'm not 100% certain what the intended behavior was before this PR but I could change the default to `WaitCompleted` which I think would cover that case.

Quinn-With-Two-Ns: Yeah the default has to be backwards compatible so it has to be `WaitCompleted `.

internal/internal_event_handlers.go
Quinn-With-Two-Ns: >I'm pretty sure this was a bug.

No please do not change this, this is the documented behaviour. this is not a backwards compatible change as described

internal/internal_event_handlers.go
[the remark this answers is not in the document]
cretz: 👍 I'd have said something if I knew the release was occurring

internal/workflow.go
cretz: Is `WaitCompleted` compatible even?

Say in a workflow I canceled a Nexus operation before it was shown as started in history. Then the Nexus operation completed successfully. It seems like if `WaitCompleted` becomes default in this PR, before-this-PR SDK would return an error and after-this-PR SDK would show a success.

It sounds like if it was TryCancel-pre-start and WaitCompleted-post-start, we have to retain that behavior. We can do it with an SDK flag or make that the default or something.

internal/internal_event_handlers.go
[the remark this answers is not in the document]
cretz: Yeah, this is the pain of frequently releasing things before we know how they will be used

internal/workflow.go
Quinn-With-Two-Ns: We lack history compatibility tests and probably should add them before we do any changes around here, but yes `WaitCompleted` **should** be backwards compatible

cretz: > yes WaitCompleted should be backwards compatible
I was under the impression that `WaitCompleted` means if it's canceled before start it still waits until completed which may succeed whereas current code before the PR will fail if you cancel before start.

Is it true the behavior today before this PR is "TryCancel-pre-start" and "WaitCompleted-post-start"? If so, that is not the same as "WaitCompleted-pre-and-post-start" there is no enum to represent that, so we have to add a flag or have a way to represent that I think.

Quinn-With-Two-Ns: >WaitCompleted means if it's canceled before start it still waits until completed

Yes that is true, and that should be true today as well. The nexus operation can also be cancelled before the command was ever sent to the server, that will cancel the future instantly

cretz: Ah, ok, @pdoerner mentioned above that it was "TryCancel-pre-start" and "WaitCompleted-post-start", but you're saying it was "WaitCompleted" always? @pdoerner - can you confirm that it is not "TryCancel-pre-start" as mentioned at https://github.com/temporalio/sdk-go/pull/1917#discussion_r2047449249?

Quinn-With-Two-Ns:
@pdoerner are you able to add replay tests before or along with this PR? Happy to help or provide instruction here as well

pdoerner:
> @pdoerner are you able to add replay tests before or along with this PR? Happy to help or provide instruction here as well

Yes, I will add them. Aiming to have this PR ready for re-review later today or tomorrow.

pdoerner:
New functional tests I added will fail in CI (I tested them locally) until we have a newer CLI version that supports the new history events. To that effect I opened https://github.com/temporalio/cli/pull/798 so we can either merge that and upgrade the version in the functional tests or I can disable the tests for now.

internal/workflow.go
```diff
@@ -93,6 +93,40 @@ const (
...
+// NexusOperationCancellationType specifies what action should be taken for a Nexus operation when the
+// caller is cancelled.
+//
+// Exposed as: [go.temporal.io/sdk/workflow.NexusOperationCancellationType]
+type NexusOperationCancellationType int
+
+const (
+	// Nexus operation cancellation type is unknown.
+	//
+	// Exposed as:
```
Quinn-With-Two-Ns, line 105: missing link to where it is exposed

internal/internal_event_handlers.go
```diff
@@ -1358,7 +1360,11 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 		enumspb.EVENT_TYPE_NEXUS_OPERATION_TIMED_OUT:
 		err = weh.handleNexusOperationCompleted(event)
 	case enumspb.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUESTED:
-		weh.commandsHelper.handleNexusOperationCancelRequested(event.GetNexusOperationCancelRequestedEventAttributes().GetScheduledEventId())
+		err = weh.handleNexusOperationCancelRequested(event)
+		//weh.commandsHelper.handleNexusOperationCancelRequested(event.GetNexusOperationCancelRequestedEventAttributes().GetScheduledEventId())
```
Quinn-With-Two-Ns, line 1364: commented out code?

test/replaytests/workflows.go
```diff
@@ -698,3 +701,28 @@ func ResetWorkflowWithChild(ctx workflow.Context) (string, error) {
...
+
+var WaitOnSignalOp = temporalnexus.NewWorkflowRunOperation(
+	"wait-on-signal-op",
+	WaitForCancelWorkflow,
+	func(ctx context.Context, _ nexus.NoValue, soo nexus.StartOperationOptions) (client.StartWorkflowOptions, error) {
+		return client.StartWorkflowOptions{}, nil
+	},
+)
+
+func CancelNexusOperationWorkflow(ctx workflow.Context) error {
```
Quinn-With-Two-Ns, line 717: We need to cover all possible previous state transitions we could have taken here basically:
* cancel before the operation start command is sent to the server
* cancel after the command is sent but before nexus operation is started
* cancel after the operation is started but before finished
* cancel after finished

We also need to make sure the workflow generates different commands on these different code paths so the SDK can detect the non determinism

test/replaytests/replay_test.go
```diff
@@ -511,6 +511,13 @@ func (s *replayTestSuite) TestResetWorkflowAfterChildComplete() {
 	require.NoError(s.T(), err)
 }
 
+func (s *replayTestSuite) TestCancelledNexusOperation() {
+	replayer := worker.NewWorkflowReplayer()
+	replayer.RegisterWorkflow(CancelNexusOperationWorkflow)
+	err := replayer.ReplayWorkflowHistoryFromJSONFile(ilog.NewDefaultLogger(), "nexus-operation-cancelled.json")
+	s.NoError(err)
```
cretz, line 518: For my own understanding, can you confirm that if we had the default as `TryCancel` in our code this test would fail?

cretz (commented):
LGTM though agree with @Quinn-With-Two-Ns about the other replay scenarios. I know it's a bit annoying to add these replay tests (have to generate history first on older server, capture it, etc). Thanks for sticking with it!

test/replaytests/replay_test.go
pdoerner: It would, I did test that manually.

internal/workflow.go
pdoerner: the doclink tool keeps removing them when I run it with `-fix` because it thinks they are stale. not too sure what's going on there.

pdoerner: I guess I will remove them for now to allow the tests to run but LMK how I should resolve this.

Quinn-With-Two-Ns: @yuandrew can you take a look, seems like a bug in the tool

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Looks like CI is failing because it can't download the test server, you may want to follow up that the release was done properly.

pdoerner:
> Looks like CI is failing because it can't download the test server, you may want to follow up that the release was done properly.

Yeah I've tried it a few times. I am able to download it locally so I don't think it is an issue with the release process. Probably some Github cache or something.

internal/workflow.go
yuandrew: hmm, this is technically by design, the doclink tool only maps type aliases, but in this case these are 2 separate definitions

Quinn-With-Two-Ns: OH we don't map these enum type values?

Quinn-With-Two-Ns:
@pdoerner just FYI you need to disable the new tests you added for the docker-compose tests since it uses the latests released OSS server.

PR #1922, opened 2025-04-18
expose default resource based slot suppliers

kevinawoo:
## What was changed
Expose the default slot suppliers options for Workflows and Activities

## Why?
The sample we supply in the [docs](https://docs.temporal.io/develop/worker-performance#go-sdk) show that options can be used.

I'm updating the doc so it can more easily be copy/past-able.

## Checklist
1. How was this tested:
Copy/Paste

2. Any docs updates needed?
https://github.com/temporalio/documentation/pull/3540


Quinn-With-Two-Ns:
Thanks will include this in the next release for the resource tuner

Issue #1667, opened 2024-10-10
Add high-level workflow describe


cretz:
**Describe the solution you'd like**

* Add a `DescribeWorkflow` call on client
  * Probably deprecate `DescribeWorkflowExecution`
* Return a `WorkflowExecutionDescription` that contains some obvious things and access to the raw describe response for now
  * Ideally this should embed the type that just represents the workflow information that can come back from list even though we don't have high-level list call at this time
* Make sure it is interceptable

See TypeScript, Python, and .NET SDKs if looking for design inspiration for this call, or reference https://github.com/temporalio/sdk-java/issues/2265.

PR #1924, opened 2025-04-21
Add high level workflow describe

Quinn-With-Two-Ns:
Add high level workflow describe

close https://github.com/temporalio/sdk-go/issues/1667


client/client.go
```diff
@@ -1133,6 +1136,16 @@ type (
...
 
+		// DescribeWorkflowExecution returns information about the specified workflow execution.
+		//  - runID can be default(empty string). if empty string then it will pick the last running execution of that workflow ID.
+		//
+		// The errors it can return:
+		//  - serviceerror.InvalidArgument
+		//  - serviceerror.Internal
+		//  - serviceerror.Unavailable
+		//  - serviceerror.NotFound
+		DescribeWorkflow(ctx context.Context, workflowID, runID string) (*DescribeWorkflowExecution, error)
```
Quinn-With-Two-Ns, line 1147: Returned a pointer here for consistency with some other APIs

internal/internal_workflow_client.go
```diff
@@ -901,6 +917,27 @@ type QueryWorkflowWithOptionsResponse struct {
 	QueryRejected *querypb.QueryRejected
 }
 
+type WorkflowExecutionMetadata struct {
```
Quinn-With-Two-Ns, line 920: I picked fields to expsoe here based of what dotnet exposed https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Client/WorkflowExecutionDescription.cs#L10

client/client.go
```diff
@@ -269,6 +269,9 @@ type (
 	// QueryWorkflowWithOptionsResponse defines the response to QueryWorkflowWithOptions.
 	QueryWorkflowWithOptionsResponse = internal.QueryWorkflowWithOptionsResponse
 
+	// DescribeWorkflowExecution defines the response to DescribeWorkflow.
+	DescribeWorkflowExecution = internal.WorkflowExecutionDescription
```
cretz, line 273:
```suggestion
	WorkflowExecutionDescription = internal.WorkflowExecutionDescription
```

I think the internal name is a good one to export as

client/client.go
```diff
@@ -1133,6 +1136,16 @@ type (
...
 
+		// DescribeWorkflow returns information about the specified workflow execution.
+		//  - runID can be default(empty string). if empty string then it will pick the last running execution of that workflow ID.
+		//
+		// The errors it can return:
+		//  - serviceerror.InvalidArgument
+		//  - serviceerror.Internal
+		//  - serviceerror.Unavailable
+		//  - serviceerror.NotFound
+		DescribeWorkflow(ctx context.Context, workflowID, runID string) (*DescribeWorkflowExecution, error)
```
cretz, line 1147: Do we want to accept `DescribeWorkflowOptions` for future proofing?

client/client.go
```diff
@@ -1133,6 +1136,16 @@ type (
 		//  - serviceerror.NotFound
 		DescribeWorkflowExecution(ctx context.Context, workflowID, runID string) (*workflowservice.DescribeWorkflowExecutionResponse, error)
```
cretz, line 1137: Do we want to discourage/deprecate this in favor of either using `DescribeWorkflow` or `WorkflowService().DescribeWorkflowExecution`?

internal/internal_workflow_client.go
cretz: Works for me, though not the biggest fan of the term "metadata", but I know it's what Java uses, so up to you. I figure one day we'll deprecate all the bad forms of "list" and have a new "list" call that returns `iter.Seq[WorkflowExecutionMetadata]`, but not sure if that affects your name choice.

internal/internal_workflow_client.go
```diff
@@ -901,6 +917,67 @@ type QueryWorkflowWithOptionsResponse struct {
...
+	WorkflowExecutionMetadata
+	dc                   converter.DataConverter
+	staticSummaryPayload *commonpb.Payload
+	staticDetailsPayload *commonpb.Payload
+}
+
+// GetStaticSummary returns the summary set on workflow start.
+//
+// NOTE: Experimental
+func (w *WorkflowExecutionDescription) GetStaticSummary() (string, error) {
```
cretz, line 960: Arguably could be worth memoizing e.g. behind a sync.Once, but no strong opinion

client/client.go
cretz, line 273: Though you don't _technically_ have to, I think you should also export the embedded field type

internal/interceptor_base.go
```diff
@@ -567,6 +567,14 @@ func (c *ClientOutboundInterceptorBase) QueryWorkflow(
 	return c.Next.QueryWorkflow(ctx, in)
 }
 
+// QueryWorkflow implements ClientOutboundInterceptor.DescribeWorkflow.
```
yuandrew, line 570:
```suggestion
// DescribeWorkflow implements ClientOutboundInterceptor.DescribeWorkflow.
```

internal/internal_workflow_client.go
Quinn-With-Two-Ns: If you have a different name I am open to suggestions, when we add list I was going to add another type that would also embedded `WorkflowExecutionMetadata `

client/client.go
Quinn-With-Two-Ns: Yeah not sure why I changed the name here

Quinn-With-Two-Ns: What is the use case for the exported embedded field type?

client/client.go
Quinn-With-Two-Ns: I was thinking we would add `DescribeWorkflowWithOptions` if we ever add more options here

internal/internal_workflow_client.go
cretz: Yeah, since `WorkflowExecution` is already in use elsewhere, this is fine. Not sure at list time we need another type since list contents should be a subset of describe, but can discuss at that time.

client/client.go
Quinn-With-Two-Ns: I wasn't going to as part of this PR, I think if we want to deprecate these calls that just directly call the  `WorkflowService()` we should deprecate them all.

client/client.go
cretz: For tests, people often appreciate the ability to easily instantiate the objects our clients can return. In this case, you have to instantiate and then set one field at a time. I think it's a reasonable rule that "every struct that is exported should also have its exported fields' types exported" (unless there's a good reason to have a non-exported interface, which is possible but rare)

client/client.go
cretz: :+1: Would support deprecating any high-level ones that don't really provide value over direct gRPC call, but definitely for this one we should deprecate the other one (can be separate PR)

cretz (approved):

PR #1925, opened 2025-04-22
Add application err category

### LINKED ISSUE #1908: Apply application failure logging and metrics behaviour according to ApplicationErrorCategory
See corresponding features issue:
https://github.com/temporalio/features/issues/614

THardy98:
## What was changed
Upgrade api-go to 1.48.0
Added support for `ApplicationErrorCategory`, allowing users to specify benign application errors.

## Why?
Part of the benign exceptions work.

1. Closes #1908

2. How was this tested:
Modify existing failure conversion tests.

3. Any docs updates needed?
Not sure


internal/error.go
```diff
@@ -137,6 +140,7 @@ type (
 		//
 		// NOTE: This option is supported by Temporal Server >= v1.24.2 older version will ignore this value.
 		NextRetryDelay time.Duration
+		Category       ApplicationErrorCategory
```
Quinn-With-Two-Ns, line 143: Needs docs

test/integration_test.go
```diff
@@ -7522,3 +7523,198 @@ func (ts *IntegrationTestSuite) TestRawValueQueryMetadata() {
...
+func (ts *IntegrationTestSuite) TestWorkflowTaskFailureMetric_BenignHandling() {
+	wfWithApplicationErr := func(ctx workflow.Context, isBenign bool) error {
+		if !isBenign {
+			return temporal.NewApplicationError("Non-benign failure", "", false, nil)
+		}
+		return temporal.NewApplicationErrorWithOptions(
+			"Benign failure",
+			"",
+			temporal.ApplicationErrorOptions{
+				Category: internal.ErrorCategoryBenign,
```
Quinn-With-Two-Ns, line 7536: test should not refer to `internal` types, you need to export `ErrorCategoryBenign`

internal/error.go
```diff
@@ -1029,3 +1045,42 @@ func getErrType(err error) string {
...
+	case enumspb.APPLICATION_ERROR_CATEGORY_UNSPECIFIED:
+		// Unspecified maps to zero value
+		return ""
+	default:
+		// Fallback to zero value if unknown case
+		return ""
+	}
+}
+
+func IsBenignApplicationError(err error) bool {
```
cretz, line 1075:
```suggestion
func isBenignApplicationError(err error) bool {
```

No need to expose I assume unless it's a helper we really need to expose, but I don't believe we do. A user can easily check for any value on application error without a helper.

internal/error.go
Quinn-With-Two-Ns, line 143: What is the release status of https://github.com/temporalio/features/issues/614?

test/integration_test.go
```diff
@@ -7522,3 +7523,198 @@ func (ts *IntegrationTestSuite) TestRawValueQueryMetadata() {
...
+		false,
+	)
+	ts.NoError(err)
+	// Wait for completion
+	err = runNonBenign.Get(context.Background(), nil)
+	// Expect a non-benign application error.
+	ts.Error(err)
+	var appErr *temporal.ApplicationError
+	ts.True(errors.As(err, &appErr))
+	ts.False(internal.IsBenignApplicationError(err))
```
cretz, line 7557: Try to avoid using `internal` in integration tests. Users can't use it.

internal/internal_task_pollers.go
```diff
@@ -474,19 +474,21 @@ func (wtp *workflowTaskPoller) RespondTaskCompletedWithMetrics(
 ) (response *workflowservice.RespondWorkflowTaskCompletedResponse, err error) {
 	metricsHandler := wtp.metricsHandler.WithTags(metrics.WorkflowTags(task.WorkflowType.GetName()))
 	if taskErr != nil {
-		wtp.logger.Warn("Failed to process workflow task.",
```
cretz, line 477: Are any of the changes made inside of this method needed?

internal/internal_task_pollers.go
```diff
@@ -474,19 +474,21 @@ func (wtp *workflowTaskPoller) RespondTaskCompletedWithMetrics(
...
+		completedRequest = failWorkflowTask
+
 		failureReason := "WorkflowError"
 		if failWorkflowTask.Cause == enumspb.WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR {
 			failureReason = "NonDeterminismError"
 		}
+
 		incrementWorkflowTaskFailureCounter(metricsHandler, failureReason)
-		completedRequest = failWorkflowTask
+		wtp.logger.Warn("Failed to process workflow task.",
```
Quinn-With-Two-Ns, line 486: What is the goal of this change?

internal/error.go
```diff
@@ -119,6 +119,9 @@ Workflow consumers will get an instance of *WorkflowExecutionError. This error w
 */
 
 type (
+	// Category of the error. Maps to logging/metrics behaviours.
+	ApplicationErrorCategory string
```
Quinn-With-Two-Ns, line 123: Is `ApplicationErrorCategory` intended to be an enum?

test/integration_test.go
THardy98: Exported/exposed in common error.

internal/error.go
THardy98: That's a good question - I'll ask drew

internal/internal_task_pollers.go
THardy98: old change that I forgot to revert, reverted.

internal/error.go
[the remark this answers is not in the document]
THardy98: Exposed in common error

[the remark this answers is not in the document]
THardy98: Yeah not sure why I opted for a string... changed to an int enum

internal/error.go
```diff
@@ -380,6 +385,11 @@ var (
 	ErrMissingWorkflowID = errors.New("workflow ID is unset for Nexus operation")
 )
 
+const (
+	// ErrorCategoryBenign indicates an error that is expected under normal operation and should not trigger alerts.
+	ErrorCategoryBenign ApplicationErrorCategory = "BENIGN"
```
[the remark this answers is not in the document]
THardy98: Done

internal/error.go
```diff
@@ -1029,3 +1045,42 @@ func getErrType(err error) string {
...
 }
+
+func applicationErrorCategoryToProto(category ApplicationErrorCategory) enumspb.ApplicationErrorCategory {
+	switch category {
+	case ErrorCategoryBenign:
+		return enumspb.APPLICATION_ERROR_CATEGORY_BENIGN
+	case "":
+		// Zero value maps to unspecified
+		return enumspb.APPLICATION_ERROR_CATEGORY_UNSPECIFIED
+	default:
```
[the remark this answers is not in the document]
THardy98: Yup, removed

internal/error.go
THardy98: Didn't mean to export this, unexported.

test/integration_test.go
THardy98: Done - inline boolean check.

internal/error.go
THardy98: Did you want this removed btw? Just do the check inline wherever it's used?
Alternatively, should we put this check under a different name (i.e. `shouldRecordFailureMetric` or something similar, probably move it to metrics or something), in case we adopt more conditional logic for failure metrics in the future?

internal/internal_task_pollers.go
THardy98: Nope, old change I forgot to revert

internal/error.go
THardy98: Yes

internal/error.go
```diff
@@ -1029,3 +1055,16 @@ func getErrType(err error) string {
 
 	return t.Name()
 }
+
+func isBenignApplicationError(err error) bool {
+	var appError *ApplicationError
+	return errors.As(err, &appError) && appError.Category() == ApplicationErrorCategoryBenign
```
cretz, line 1061: I don't think we should assume causes no matter the depth as benign. Explained at https://github.com/temporalio/sdk-java/pull/2485#discussion_r2059154968.

```suggestion
	appError, _ := err.(*ApplicationError)
	return appError != nil && appError.Category() == ApplicationErrorCategoryBenign
```

internal/error.go
```diff
@@ -137,6 +137,8 @@ type (
 		//
 		// NOTE: This option is supported by Temporal Server >= v1.24.2 older version will ignore this value.
 		NextRetryDelay time.Duration
+		// Category of the error. Maps to logging/metrics behaviours.
```
Quinn-With-Two-Ns, line 140:
```suggestion
		// Category of the error. Maps to SDK logging/metrics behaviours.
```
This `Category` does not effect server side metrics right?

internal/error.go
Quinn-With-Two-Ns: Yes we should only look at the top level error

internal/error.go
THardy98: Correct

cretz (approved):
Look great! Would like @Quinn-With-Two-Ns's approval before merge though.

internal/error.go
Quinn-With-Two-Ns: Can you update the comment then if this is correct?

THardy98: Done

Quinn-With-Two-Ns (approved):

PR #1933, opened 2025-04-25
Don't heartbeat local activity on non-graceful worker shutdown

### LINKED ISSUE #1706: ProcessWorkflowTask is not stopped on worker.Stop()

yuandrew:
## What was changed

Stop local activity from heartbeating after worker has shutdown

## Why?

They shouldn't heartbeat after they've been canceled due to non-graceful shutdown

## Checklist

1. Closes #1706 

2. How was this tested:

added integration test

3. Any docs updates needed?


internal/internal_task_handlers.go
```diff
@@ -998,6 +998,17 @@ processWorkflowLoop:
...
+						// to avoid heartbeating after local activity is no longer being run
+						var appErr *ApplicationError
+						if errors.As(lar.err, &appErr) {
+							// AppErr with ErrCanceled errType are non-server initiated cancellations.
+							if appErr.errType == ErrCanceled.Error() {
+								break processWorkflowLoop
+							}
+						}
+
 						response, err = workflowContext.ProcessLocalActivityResult(workflowTask, lar)
```
cretz, line 1012: I need some help understanding the issue. Why don't we want the workflow to heartbeat until the local activity has completed? Can you help me understand what logic below without this new change was causing `ProcessLocalActivityResult` to return both a nil `response` and `err` causing us to continue to wait for local activity? Wouldn't the `retryLocalActivity` be false since this is a canceled error and therefore `err` be non-nil?

yuandrew: `retryLocalActivity` is false, but `w.hasPendingLocalActivityWork()` in `CompleteWorkflowTask` is true, causing `response` and `err` to be nil and to loop back

cretz: Why is `hasPendingLocalActivityWork` still true when the local activity is done? Hrmm, what if `hasPendingLocalActivityWork` _is_ legitimately true because another local activity is still pending. My concern now is what if I have two local activities running and only one has responded with cancel and the other is still processing its cleanup logic? This breaks out of the whole loop even though another local activity may finish with success. I wonder if we should just consider _this_ local activity as no longer pending.

I am now generally curious about concurrent local activities with one failing with the cancel and one succeeding what would be recorded in the workflow task? What if the success came before the cancel one but both before any heartbeat occurred, does that mean only the one success marker is captured? Is that safe? (sorry, my heading is getting a bit lost thinking about it).

internal/internal_task_handlers.go
```diff
@@ -998,6 +998,17 @@ processWorkflowLoop:
 
 					case lar := <-workflowTask.laResultCh:
 						// local activity result ready
+
+						// When local activity is canceled due to non-server cancel, we break loop here
+						// to avoid heartbeating after local activity is no longer being run
+						var appErr *ApplicationError
+						if errors.As(lar.err, &appErr) {
+							// AppErr with ErrCanceled errType are non-server initiated cancellations.
+							if appErr.errType == ErrCanceled.Error() {
```
Quinn-With-Two-Ns, line 1007: A user can return this error to no? and a user doesn't have to return this error when the local activity worker is shut down so I am not sure this is the correct way to detect this condition.

yuandrew: Do you think it's a viable fix to somehow pass a custom error through `laResultCh` altogether? That seems like the easiest way to communicate that the worker has shutdown and we can stop processing this WFT

> a user doesn't have to return this error when the local activity worker is shut down
My understanding is when worker shuts down, it cancels context and the user's code just stops running. Are you saying the user can handle that context cancellation and return their own custom error?

yuandrew: > shutdown comes from the top down to both workers

I don't think it does in this scenario. If I don't return to `laResultCh`, the `heartbeatTimer.C` keeps firing and the test continues to run for the full 15 seconds.

yuandrew: oh actually, I can probably use `workflowContext.laTunnel.stopCh`

internal/internal_task_handlers.go
yuandrew: Do multiple local activities matter if the worker is shutting down? Isn't it all moot, unless the WFT happens to finish as the worker shuts down?

internal/internal_task_handlers.go
```diff
@@ -939,17 +939,24 @@ processWorkflowLoop:
...
-							workflowContext.CompleteWorkflowTask(workflowTask, false),
-							startTime,
-						)
-						if err != nil {
-							errRet = &workflowTaskHeartbeatError{Message: fmt.Sprintf("error sending workflow task heartbeat %v", err)}
-							return
-						}
-						if workflowTask == nil {
+						select {
+						case <-workflowContext.laTunnel.stopCh:
```
Quinn-With-Two-Ns, line 943: Doesn't this channel close when the worker is shutting down during graceful shutdown? What happens when a local activity competes during graceful shutdown? Will the result be ignored?

yuandrew: no, this channel should now only close after local activities finish running, from https://github.com/temporalio/sdk-go/pull/1875.
```
    // We want a separate stop channel for local activities because when a worker shuts down,
	// we need to allow pending local activities to finish running for that workflow task.
	// After all pending local activities are handled, we then close the local activity stop channel.
	laStopChannel := make(chan struct{})
	laParams := params
	laParams.WorkerStopChannel = laStopChannel

	// laTunnel is the glue that hookup 3 parts
	laTunnel := newLocalActivityTunnel(getReadOnlyChannel(laStopChannel))
```

Also added another integration test validating this behavior

test/integration_test.go
```diff
@@ -7707,3 +7708,141 @@ func (ts *IntegrationTestSuite) TestLocalActivityFailureMetric_BenignHandling()
 	// Expect count to remain unchanged
 	ts.assertMetricCount(metrics.LocalActivityExecutionFailedCounter, currCount)
 }
+
+func (ts *IntegrationTestSuite) TestLocalActivityWorkerShutdownNoHeartbeat() {
+	// FYI, setup of this test allows the worker to wait to stop for 10 seconds
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	localActivityFn := func(ctx context.Context) error {
+		time.Sleep(300 * time.Millisecond)
```
Quinn-With-Two-Ns, line 7717: nit: can we use the worker shutdown channel here to avoid any potential timing race? Basically complete after the shutdown channel is closed so we know the activity completed after the worker was requested to shutdown

yuandrew: looks like because of https://github.com/temporalio/sdk-go/pull/1875, now the LA worker stop channel is `ww.localActivityStopC`, which only closes after pending LAs are complete. 

I can work on a separate PR to see if I can fix `GetWorkerStopChannel` to return the worker shutdown channel instead of the LA shutdown channel, and in that PR switch these and some prev tests i wrote to use GetWorkerStopChannel instead of a manual wait

Quinn-With-Two-Ns: Can you open up an issue for this and add a TODO here? We will see if we can pull it in next sprint.

internal/internal_task_handlers.go
Quinn-With-Two-Ns: Talked offline this should be fine, maybe just add a comment explaining what is happening here for future readers.

Quinn-With-Two-Ns (approved):

cretz (commented):
Btw, no need to wait for my review/approval as this is a bit too deep for me to dig back into, @Quinn-With-Two-Ns's review is enough IMO

PR #1934, opened 2025-04-25
Handle links from ExecuteWorkflow in Nexus WorkflowRunOperation

rodrigozhou:
## What was changed

Introduce `StartWorkflowResponseInfo` to `StartWorkflowOptions`.
Handle links from `ExecuteWorkflow` in Nexus `WorkflowRunOperation`.

## Why?

Be able to pass information from `StartWorkflowExecutionResponse` to `ExecuteWorkflow` call through `StartWorkflowOptions`.
Use the link from the server instead of generating one in `WorkflowRunOperation`.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -762,6 +762,10 @@ type (
 		// WARNING: Task queue priority is currently experimental.
 		Priority Priority
 
+		// responseInfo - Optional pointer to store information of StartWorkflowExecution response.
+		// Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+		responseInfo *startWorkflowResponseInfo
+
 		// request ID. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
 		requestID string
 		// workflow completion callback. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
@@ -776,6 +780,13 @@ type (
 		//
 		// NOTE: Only settable by the SDK -- e.g. [temporalnexus.workflowRunOperation].
 		onConflictOptions *OnConflictOptions
+	}
+
+	// startWorkflowResponseInfo can be passed to StartWorkflowOptions to receive additional information
+	// of StartWorkflowExecution response.
+	startWorkflowResponseInfo struct {
+		// Link to the workflow event.
+		Link *commonpb.Link
 	}
 
 	// WithStartWorkflowOperation defines how to start a workflow when using UpdateWithStartWorkflow.
@@ -1275,4 +1286,14 @@ func SetOnConflictOptionsOnStartWorkflowOptions(opts *StartWorkflowOptions) {
 		AttachCompletionCallbacks: true,
 		AttachLinks:               true,
 	}
+}
+
+// SetResponseInfoOnStartWorkflowOptions is an internal only method for setting start workflow
+// response info object pointer on StartWorkflowOptions and return the object pointer.
+// StartWorkflowResponseInfo is purposefully not exposed to users for the time being.
+func SetResponseInfoOnStartWorkflowOptions(opts *StartWorkflowOptions) *startWorkflowResponseInfo {
+	if opts.responseInfo == nil {
+		opts.responseInfo = &startWorkflowResponseInfo{}
+	}
+	return opts.responseInfo
 }
--- a/temporalnexus/operation.go
+++ b/temporalnexus/operation.go
@@ -214,6 +214,7 @@ type workflowHandle[T any] struct {
 	namespace   string
 	id          string
 	runID       string
+	wfEventLink *common.Link
 	cachedToken string
 }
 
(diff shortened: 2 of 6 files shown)


internal/client.go
```diff
@@ -778,6 +778,10 @@ type (
 		// WARNING: Task queue priority is currently experimental.
 		Priority Priority
 
+		// resultInfo - Optional pointer to store information of StartWorkflowExecution response.
+		// Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+		resultInfo *startWorkflowResultInfo
```
cretz, line 783: I would expect this field and the result info to be exported to be usable by anyone outside of this library

internal/client.go
```diff
@@ -794,6 +798,13 @@ type (
 		onConflictOptions *OnConflictOptions
 	}
 
+	// startWorkflowResultInfo can be passed to StartWorkflowOptions to receive additional information
+	// of StartWorkflowExecution response.
+	startWorkflowResultInfo struct {
```
cretz, line 803: While you are here, would like two more booleans: `Started` and `StartedEagerly`. `Started` is very helpful for users that do use-existing, update with start, and those not setting `WorkflowExecutionErrorWhenAlreadyStarted`. `StartedEagerly` is something users may want to know, and in general our tests would like to know this too. If we want to do this in a separate issue, no prob.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/nexus-rpc/sdk-go v0.3.0
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.10.0
-	go.temporal.io/api v1.46.0
+	go.temporal.io/api v1.48.1-0.20250423210630-eb8d7eac6218
```
cretz, line 17: Can't merge this PR until this is tagged of course

internal/client.go
cretz, line 803: Can we change this to "ResponseInfo" instead of "ResultInfo" so it's not confused with the concept of "workflow result"?

internal/client.go
```diff
@@ -794,6 +798,13 @@ type (
...
+		// Link to the workflow event.
+		Link *commonpb.Link
```
cretz, line 805: I think it's ok, but would like to know other SDK team member's thoughts about exposing the raw proto here vs creating a high-level Go representation of a link (cc @Quinn-With-Two-Ns). If/when we expose the ability for users to set links on workflow start, what form would we accept?

Quinn-With-Two-Ns: No thoughts since this isn't exposed outside of the SDK

internal/client.go
Quinn-With-Two-Ns: Seems unnecessary in this PR since these fields won't be accessible to users anyway

internal/client.go
Quinn-With-Two-Ns: This should not be exposed

cretz: Oh, I didn't see it being used by Nexus, I though it was just a mistake that it wasn't exposed

internal/client.go
cretz: :+1: I didn't realize it is intentionally unexposed.

internal/client.go
cretz: 👍 I didn't realize it is intentionally unexposed.

internal/client.go
rodrigozhou: I'll have a separate PR for the Nexus change. I can do at once, just thought it would be easier to review as two PRs

cretz:
@rodrigozhou - if nothing is using this private code, why does it need to get into main until it's used? Let's wait until it will be actually used to add to the repo, there seems to be 0 benefit to getting it there ahead of time.

temporalnexus/operation.go
```diff
@@ -383,6 +387,7 @@ func ExecuteUntypedWorkflow[R any](
 	}
 	internal.SetLinksOnStartWorkflowOptions(&startWorkflowOptions, links)
 	internal.SetOnConflictOptionsOnStartWorkflowOptions(&startWorkflowOptions)
+	responseInfo := internal.SetResponseInfoOnStartWorkflowOptions(&startWorkflowOptions)
```
cretz, line 390: Is there an end-to-end test to prove this works?

cretz (commented):
LGTM (waiting to approve pending API release and potentially an end-to-end test to make sure it works as expected for users). Will you also do the Java one?

temporalnexus/operation.go
rodrigozhou: Modified existing tests to check this.

cretz (approved):
I think at some point we should expose start info to users with booleans telling you whether the workflow started and/or is running as a result of the call. But since this is just links and since there is actually code in here using this unexposed piece, LGTM.

PR #1937, opened 2025-04-28
Handle mix of RawValue and non-RawValue inputs for CompositeDataConverter

yuandrew:
## What was changed

For data converter, support handling a mix of RawValue and non-RawValue input.

For codec, Raw Value support has been removed, as we'd only be able to conditionally support RawValue (if all inputs are RawValue), not worth the complexity to kinda support it in certain scenarios

## Why?

Match python/dotnet

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):

converter/codec.go
```diff
@@ -210,7 +214,7 @@ func (e *CodecDataConverter) ToPayloads(value ...interface{}) (*commonpb.Payload
 		}
 	}
 
-	if len(rawValuePayloads) > 0 {
```
Quinn-With-Two-Ns, line 213: It is a bit confusing to only handle raw values here is all of the inputs are raw value. Maybe we should just rely on the underlying data converter? Originally I was hopping we could make `CodecDataConverter` support raw value, but it looks like that may not be possible to handle it consistently. What do you think?

cretz (approved):
LGTM. May want to note on the raw value type itself that you must have a raw-value enabled payload converter for it to work. Also may want to change this PR title (shows up in git blame, release notes, etc).

Quinn-With-Two-Ns (commented):
Can we add a test case for this? Testing `CompositeDataConverter.ToPayloads` and `CompositeDataConverter.FromPayloads` can handle a mix of values should be sufficient.

Quinn-With-Two-Ns (approved):

PR #1938, opened 2025-04-28
Add missing lock to direct registration of WorkflowDefinition

yuandrew:
## What was changed

Added a lock before mutating shared state

## Why?

Came across this while looking at dynamic activities, seems like we should be locking here.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/internal_worker.go
```diff
@@ -583,6 +583,8 @@ func (r *registry) RegisterWorkflowWithOptions(
 		if strings.HasPrefix(options.Name, temporalPrefix) {
 			panic(temporalPrefixError)
 		}
+		r.Lock()
+		defer r.Unlock()
```
cretz, line 587: Not a big deal, but can probably just move the lock/unlock from several lines down to the top of this function so it covers future code too

cretz (approved):

Quinn-With-Two-Ns (approved):

PR #1945, opened 2025-05-01
Set Nexus links in the callback

rodrigozhou:
## What was changed

Set Nexus links in the callback

## Why?

All links from the Nexus request must be associated with the Nexus callback.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalnexus/operation.go
+++ b/temporalnexus/operation.go
@@ -323,6 +323,8 @@ func ExecuteUntypedWorkflow[R any](
 		})
 	}
 
+	// Links are duplicated in startWorkflowOptions to backwards compatibility with older servers that
+	// don't support links in callbacks.
 	internal.SetLinksOnStartWorkflowOptions(&startWorkflowOptions, links)
 	internal.SetOnConflictOptionsOnStartWorkflowOptions(&startWorkflowOptions)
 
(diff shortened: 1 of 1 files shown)


temporalnexus/operation.go
```diff
@@ -332,17 +340,11 @@ func ExecuteUntypedWorkflow[R any](
...
 	}
 
-	links, err := convertNexusLinks(nexusOptions.Links, GetLogger(ctx))
-	if err != nil {
-		return nil, &nexus.HandlerError{
-			Type:  nexus.HandlerErrorTypeBadRequest,
-			Cause: err,
-		}
-	}
 	internal.SetLinksOnStartWorkflowOptions(&startWorkflowOptions, links)
```
bergundy, line 348: Please add a note saying that the links are sent twice for older servers that don't accept links in the callback.

bergundy (approved):

Quinn-With-Two-Ns (approved):

cretz (approved):
Is there a Java equivalent needed?

rodrigozhou:
@cretz Yes, we do need an equivalent in Java.

Issue #997, opened 2023-01-05
Support dynamic query handler


cretz:
**Describe the solution you'd like**

Need to allow a dynamic query handler that is called when no other query handler is found. It's signature must accept a string name and a slice of `converter.EncodedValue`

PR #1946, opened 2025-05-02
Dynamic workflows and activities

### LINKED ISSUE #543: Support dynamic activity and workflow implementations
In some cases, activity and workflow implementations should be loaded dynamically.

Expose an extension point for dynamic activity loading. This issue is created to ensure feature parity among SDK's where https://github.com/temporalio/sdk-java/issues/245 already implements this for Java.

**Additional context**
In [Dynamic activities registration at runtime and worker splitting](https://community.temporal.io/t/dynamic-activities-registration-at-runtime-and-worker-splitting/930/12?u=aschrijver) on the community forum @mfateev requested another user to create this issue, but that never happened AFAIK. Hence this issue.

yuandrew:
## What was changed
Added support for dynamic workflows and activities.

Wrote a sample-go sample https://github.com/temporalio/samples-go/pull/404, will merge once this feature goes in

## Why?

Feature parity with other SDKs

## Checklist

1. Closes #543
Queries, signals, updates are tracked in a different issue: https://github.com/temporalio/sdk-go/issues/997

2. How was this tested:
Added new unit tests and integration tests

3. Any docs updates needed?


internal/internal_worker.go
```diff
@@ -1076,6 +1178,12 @@ func (aw *AggregatedWorker) RegisterActivityWithOptions(a interface{}, options R
 	aw.registry.RegisterActivityWithOptions(a, options)
 }
 
+// RegisterDynamicActivity registers the dynamic activity function with options.
+// Registering activities via a structure is not supported for dynamic activities.
+func (aw *AggregatedWorker) RegisterDynamicActivity(a interface{}) {
```
yuandrew, line 1183: Is it worth adding an `options` param now, even though it'd be empty? for future-proofing

internal/workflow.go
```diff
@@ -472,6 +472,26 @@ type (
 		VersioningBehavior VersioningBehavior
 	}
 
+	// DynamicRegisterOptions consists of options for registering a dynamic workflow
+	//
+	// Exposed as: [go.temporal.io/sdk/workflow.DynamicRegisterOptions], [go.temporal.io/sdk/activity.DynamicRegisterOptions]
```
yuandrew, line 477: will fix the tool

internal/encode_args.go
```diff
@@ -80,6 +80,11 @@ func decodeArgsToRawValues(dc converter.DataConverter, fnType reflect.Type, data
 	}
 
 	// Unmarshal
+
+	// Dynamic workflows take in a single EncodedValues, encode all data into single EncodedValues
```
cretz, line 84: I thought it would be varargs of `RawValue`, but encoded values makes sense too

internal/encode_args.go
```diff
@@ -80,6 +80,11 @@ func decodeArgsToRawValues(dc converter.DataConverter, fnType reflect.Type, data
...
+	if fnType.NumIn() == 2 && len(pointers) == 1 && fnType.In(1) == reflect.TypeOf((*converter.EncodedValues)(nil)).Elem() {
```
cretz, line 85: Hrmm, wonder if it's worth a more specific check whether it's a dynamic workflow/activity instead of assuming so based on signature? Maybe not worth it.

internal/internal_worker.go
```diff
@@ -867,6 +938,17 @@ func validateFnFormat(fnType reflect.Type, isWorkflow bool) error {
 		}
 	}
 
+	if isDynamic {
+		if fnType.NumIn() != 2 {
+			return fmt.Errorf(
+				"expected function to have two arguments, first being workflow.Context and second being an EncodedValues type, found %d arguments", fnType.NumIn(),
+			)
+		}
+		if !fnType.In(1).Implements(reflect.TypeOf((*converter.EncodedValues)(nil)).Elem()) {
```
cretz, line 947:
```suggestion
		if fnType.In(1) != reflect.TypeOf((*converter.EncodedValues)(nil)).Elem() {
```

I think simple equality check is good here

internal/internal_worker.go
```diff
@@ -619,6 +622,33 @@ func (r *registry) RegisterWorkflowWithOptions(
...
+		return
+	}
+
+	// Validate that it is a function
+	fnType := reflect.TypeOf(wf)
+	if err := validateFnFormat(fnType, true, true); err != nil {
+		panic(err)
+	}
+
+	if r.dynamicWorkflow != nil {
```
cretz, line 642: Not concurrency safe. There's nothing blocking in this function, I'd just move the lock/unlock statements to be the first two statements in the function, no need to try to narrow it

internal/internal_worker.go
```diff
@@ -713,6 +743,29 @@ func (r *registry) registerActivityStructWithOptions(aStruct interface{}, option
...
+	if err := validateFnFormat(fnType, false, true); err != nil {
+		panic(err)
+	}
+	r.Lock()
+	defer r.Unlock()
+	if r.dynamicActivity != nil {
+		panic("dynamic activity already registered")
+	}
+	r.dynamicActivity = &activityExecutor{name: "", fn: af}
+
```
cretz, line 766:
```suggestion
```

internal/internal_worker.go
```diff
@@ -929,7 +1011,10 @@ func (we *workflowExecutor) Execute(ctx Context, input *commonpb.Payloads) (*com
 	// Execute and serialize result
 	result, err := envInterceptor.inboundInterceptor.ExecuteWorkflow(ctx, &ExecuteWorkflowInput{Args: args})
 	var serializedResult *commonpb.Payloads
-	if err == nil && result != nil {
+	// Dynamic workflows always return EncodedValues, skip encoding because result should already be encoded
```
cretz, line 1014: Do they always return encoded values? There is no validation of this return type at registration time, and they don't really _need_ to. For many, it may make more sense to return an `any` or a `RawValue`. Also, even if we do support encoded value as a return type, one might expect a single encoded value instead of plural (or at least have the option).

internal/internal_worker.go
cretz: IMO yes it's worth it

internal/workflow.go
```diff
@@ -472,6 +472,26 @@ type (
...
+	DynamicRegisterOptions struct {
+		// Allows dynamic options to be loaded for a workflow.
+		LoadDynamicOptions func(name string) (DynamicWorkflowOptions, error)
```
cretz, line 480:
```suggestion
		LoadDynamicOptions func(LoadDynamicOptionsDetails) (DynamicWorkflowOptions, error)
```

Need to be more future proof on the param IMO

internal/workflow.go
```diff
@@ -472,6 +472,26 @@ type (
...
+	}
+
+	// DynamicWorkflowOptions are options for a dynamic workflow.
+	//
+	// Exposed as: [go.temporal.io/sdk/workflow.DynamicWorkflowOptions]
+	DynamicWorkflowOptions struct {
```
cretz, line 486: I wonder if we need a different name here. The concept of "workflow options" is traditionally used by workflow callers to set options about the workflow they are calling (e.g. `StartWorkflowOptions` and `ChildWorkflowOptions`).

yuandrew: yeah I see what you're saying, maybe `config`?

internal/encode_args.go
Quinn-With-Two-Ns: It is unfortunate we have both concepts, but `EncodedValues ` is more appropriate for dynamic workflows

internal/encode_args.go
Quinn-With-Two-Ns: Yeah I would consider a more specific check or just not sue this function for dynamic workflows/activities

internal/internal_worker.go
```diff
@@ -559,6 +559,9 @@ type registry struct {
 	workflowVersioningBehaviorMap map[string]VersioningBehavior
 	activityFuncMap               map[string]activity
 	activityAliasMap              map[string]string
+	dynamicWorkflow               interface{}
+	dynamicOptions                DynamicRegisterOptions
```
Quinn-With-Two-Ns, line 563:
```suggestion
	dynamicWorkflowOptions                DynamicRegisterOptions
```
We may have activity options in the future

internal/internal_worker.go
Quinn-With-Two-Ns: Yeah it is a good question, In Java we require dynamic workflows to return an `Object` so that is like `any` in Go.

Quinn-With-Two-Ns:
I think this is a great case to try adding a sample and link to it in the PR, what do you think @yuandrew ?

internal/encode_args.go
cretz: Hrmm. I think they're similar from a user POV for params (I personally prefer codec to apply outside workflow boundary), but we should allow any return value IMO.

internal/internal_worker.go
cretz: Instead of `any`, I think we should have no requirement on return type at all. Returning a `RawValue` should be just fine for example.

yuandrew: I'm in favor of matching the other non-Java SDKs and not requiring a return value. I don't think adding a return type requirement adds any guard rails for the users, so I wanna say matching behavior of the majority of the SDKs would be better.

yuandrew:
Good idea, added a sample and linked in PR description

internal/encode_args.go
Quinn-With-Two-Ns: Yeah the dynamic workflow should be allowed to return anything I agree, I was referring to the input type

internal/workflow_testsuite.go
```diff
@@ -323,7 +331,15 @@ func (e *TestWorkflowEnvironment) RegisterActivityWithOptions(a interface{}, opt
 	e.impl.RegisterActivityWithOptions(a, options)
 }
 
-// RegisterWorkflow registers a Nexus Service with the TestWorkflowEnvironment.
+// RegisterDynamicActivity registers the dynamic activity implementation with the TestWorkflowEnvironment
+func (e *TestWorkflowEnvironment) RegisterDynamicActivity(a interface{}, options DynamicRegisterActivityOptions) {
```
Quinn-With-Two-Ns, line 335: Are there tests for this and `RegisterDynamicWorkflow` that tests a succesful workflow run?

yuandrew: Added a test

internal/workflow.go
```diff
@@ -472,6 +472,36 @@ type (
...
+	LoadDynamicOptionsDetails struct {
+		WorkflowType WorkflowType
+	}
+
+	// DynamicRegisterOptions consists of options for registering a dynamic workflow
+	//
+	// Exposed as: [go.temporal.io/sdk/workflow.DynamicRegisterOptions]
+	DynamicRegisterWorkflowOptions struct {
+		// Allows dynamic options to be loaded for a workflow.
+		LoadDynamicOptions func(details LoadDynamicOptionsDetails) (RegisterDynamicWorkflowOptions, error)
```
cretz, line 487: The more I think about this, the concept of `DynamicRegisterWorkflowOptions` (even if exposed as `DynamicRegisterOptions`) and `RegisterDynamicWorkflowOptions` is just too confusingly named.

I think we should use the word "runtime" in here maybe, e.g.:

```suggestion
		LoadDynamicRuntimeOptions func(details LoadDynamicRuntimeOptionsDetails) (DynamicRuntimeWorkflowOptions, error)
```

cretz (commented):
Everything LGTM but still hung up on the naming of the type returned from `LoadDynamicOptions`

internal/internal_worker.go
```diff
@@ -559,6 +559,9 @@ type registry struct {
 	workflowVersioningBehaviorMap map[string]VersioningBehavior
 	activityFuncMap               map[string]activity
 	activityAliasMap              map[string]string
+	dynamicWorkflow               interface{}
+	dynamicWorkflowOptions        DynamicRegisterWorkflowOptions
+	dynamicActivity               activity
```
Quinn-With-Two-Ns, line 564: nit: can wire the dynamic activity options now since you added them

internal/internal_worker.go
```diff
@@ -978,8 +1082,11 @@ func (ae *activityExecutor) ExecuteWithActualArgs(ctx context.Context, args []in
 	result, resultErr := interceptor.ExecuteActivity(ctx, &ExecuteActivityInput{Args: args})
 	var serializedResult *commonpb.Payloads
 	if result != nil {
+		// Dynamic activities always return EncodedValues, skip encoding because result should already be encoded
```
Quinn-With-Two-Ns, line 1085: Didn't we say we wanted dynamic workflows and activities to be able to return any value?

internal/workflow.go
Quinn-With-Two-Ns: Fine with me, whatever we do here we should also change the dynamic activity options name

cretz: This is only specifically concerning runtime loading of workflow options by workflow type and has no activity anallog at this time

Quinn-With-Two-Ns (approved):

Issue #1930, opened 2025-04-23
Have implicit activity context timeout start-to-close be relative to system time not server time


cretz:
**Describe the solution you'd like**

Currently the code at https://github.com/temporalio/sdk-go/blob/a92189071ec5931c88a97284d60b8b06f1024fc3/internal/activity.go#L444-L454 has start-to-close calculated from server-provided start time instead of current system time which means it may be subject to skew (ref #1912). Have it based on `time.Now()` instead. Note, nothing can be done about schedule-to-start equivalent because there is no local-time equivalent for "scheduled".

Issue #1912, opened 2025-04-13
activity Schedule To Close timeout（Activity complete after timeout）


Pharaohsk:
## Expected Behavior
In a three-node cluster using Temporal, the time on each node is not synchronized. The Temporal server is deployed on the main control node, where Service A starts a workflow and sets configurations such as ScheduleToCloseTimeout and other timeout settings. When the task is scheduled to execute on Service B on node B, the local time on node B is later than that on node A. As a result, when Service B executes the task, its local time has already exceeded the timeout. It is expected that no error will be reported.

## Actual Behavior
return err : Activity complete after timeout..
File : internal/internal_task_handlers.go
```go
	info := getActivityEnv(ctx)
	ctx, dlCancelFunc := context.WithDeadline(ctx, info.deadline)
	defer dlCancelFunc()

	output, err := activityImplementation.Execute(ctx, t.Input)
	// Check if context canceled at a higher level before we cancel it ourselves
	isActivityCancel := ctx.Err() == context.Canceled

	dlCancelFunc()
	if <-ctx.Done(); ctx.Err() == context.DeadlineExceeded {
		ath.logger.Info("Activity complete after timeout.",
			tagWorkflowID, t.WorkflowExecution.GetWorkflowId(),
			tagRunID, t.WorkflowExecution.GetRunId(),
			tagActivityType, activityType,
			tagAttempt, t.Attempt,
			tagResult, output,
			tagError, err,
		)
		return nil, ctx.Err()
	}
```
I think the code should use context.WithTimeout instead of context.WithDeadline(ctx, info.deadline).

## Steps to Reproduce the Problem

  1.Use Temporal in a cluster with two or more nodes, with the Temporal server running on the main control node.
  2.Configure the task timeout and allow the task to be scheduled for execution on another node.
  3.Modify the time on all hosts in the cluster to be unsynchronized, with the time difference exceeding the configured task timeout.

## Specifications

  - Version: v1.22.0
  - Platform: go version go1.18 linux

cretz:
Interesting. The deadline calculated on the worker is based on the server start-time of the task because that's how server uses it. The server will timeout the activity for the same period of time, this is just a worker-side to keep activities from running forever. The time is started when the server starts the task, not when the worker starts the task. I am not sure we support situations of significant clock skew. I think there is an expectation that server and worker clocks must be reasonably accurate and that timeouts should be set high enough to not be hit except in rare/failure scenarios.

[...]

PR #1952, opened 2025-05-10
Use system time as started time

Pharaohsk:
## What was changed

Use system time as started time

## Why?

Have implicit activity context timeout start-to-close be relative to system time not server time

## Checklist

1. Closes 
#1930
2. How was this tested:

The time is not consistent across hosts in the construct cluster.
The local time of the host on which the construct executes the activity has exceeded the timeout set by the server.
#1912

3. Any docs updates needed?

N/A

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -287,7 +287,7 @@ func WithActivityTask(
 	client *WorkflowClient,
 ) (context.Context, error) {
 	scheduled := task.GetScheduledTime().AsTime()
-	started := task.GetStartedTime().AsTime()
+	started := time.Now()
 	scheduleToCloseTimeout := task.GetScheduleToCloseTimeout().AsDuration()
 	startToCloseTimeout := task.GetStartToCloseTimeout().AsDuration()
 	heartbeatTimeout := task.GetHeartbeatTimeout().AsDuration()


cretz:
I am not sure backwards-incompatibly changing the activity start time to no longer be the server-reported start time is the correct answer to #1930. #1930 refers to the timeout calculation which can be done with system time. People may expect the activity info's `StartedTime` to match exactly what the server reports for the activity and this would break that expectation.

Pharaohsk:
> 我不确定将活动开始时间向后不兼容地更改为不再使用服务器报告的开始时间是否是#1930的正确答案。#1930指的是可以使用系统时间进行的超时计算。人们可能期望活动信息与`StartedTime`服务器报告的活动信息完全匹配，但这会打破这种期望。活动开始的时间实际上归服务器所有，并且与活动到达工作器的时间不同。

already modified

internal/activity.go
```diff
@@ -287,7 +287,7 @@ func WithActivityTask(
 	client *WorkflowClient,
 ) (context.Context, error) {
 	scheduled := task.GetScheduledTime().AsTime()
-	started := task.GetStartedTime().AsTime()
+	started := time.Now()
```
cretz, line 290: I don't think we should change this variable, just leave it as it was and adjust `calculateActivityDeadline` to no longer accept a `started` time and just use `time.Now()` within it. It's effectively the same thing, but I think it's clearer not change anything in this function (except it'll be one less parameter to `calculateActivityDeadline`).

Pharaohsk: thanks for your review. I have modified it
Use time.Now as started in calculateActivityDeadline

cretz (approved):
This looks good to me, @Quinn-With-Two-Ns?

yuandrew:
@Pharaohsk looks like the unit test `TestActivityExecutionDeadline` needs to be updated with this change

Pharaohsk:
> thanks for your reminder. I have modified it

thanks for your reminder. I have modified it

internal/internal_task_handlers_test.go
```diff
@@ -2067,8 +2067,8 @@ func (t *TaskHandlersTestSuite) TestActivityExecutionDeadline() {
 		{0, time.Now(), 3 * time.Second, time.Now(), 4 * time.Second, "test", nil},
 		{0, time.Now(), 3 * time.Second, time.Now(), 4 * time.Second, "unknown", nil},
 		{0, time.Now().Add(-1 * time.Second), 1 * time.Second, time.Now(), 1 * time.Second, "test", context.DeadlineExceeded},
-		{0, time.Now(), 1 * time.Second, time.Now().Add(-1 * time.Second), 1 * time.Second, "test", context.DeadlineExceeded},
-		{0, time.Now().Add(-1 * time.Second), 1, time.Now().Add(-1 * time.Second), 1 * time.Second, "test", context.DeadlineExceeded},
+		{0, time.Now().Add(-1 * time.Second), 1 * time.Second, time.Now(), 1 * time.Second, "test", context.DeadlineExceeded},
```
yuandrew: all 3 of these are now the same test, can remove the duplicates

Pharaohsk: thanks for your reminder. I have removed duplicated tests

yuandrew:
Thanks for the contribution!

PR #1953, opened 2025-05-10
Use safeAsTime and safeAsDuration

### LINKED ISSUE #1846: AsTime() on a nil pointer converts to  Jan. 1st, 1970 at midnight UTC

k8scat:
Completely fix #1846

### DIFF
--- a/internal/internal_worker_deployment_client.go
+++ b/internal/internal_worker_deployment_client.go
@@ -11,7 +11,6 @@ import (
 	"go.temporal.io/api/deployment/v1"
 	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/converter"
-	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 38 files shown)


k8scat:
@cretz Do you remind spending some time to have a look on this PR 😄

internal/internal_worker_deployment_client.go
```diff
@@ -29,6 +30,15 @@ func safeAsTime(timestamp *timestamppb.Timestamp) time.Time {
 	}
 }
 
+// safeAsDuration ensures that a nil proto duration makes `AsDuration()` return 0.
+func safeAsDuration(duration *durationpb.Duration) time.Duration {
```
Quinn-With-Two-Ns, line 34: Why was this added? `.AsDuration()` should already be safe for nil

k8scat: > Why was this added? `.AsDuration()` should already be safe for nil

thanks for review, I will check this soon

k8scat: > Why was this added? `.AsDuration()` should already be safe for nil

`.AsDuration` is a func on `*durationpb.Duration`, and will panic if call it from nil varible

image

image

Quinn-With-Two-Ns: >and will panic if call it from nil varible,

That is not true

```
	var d *durationpb.Duration
	fmt.Println(d.AsDuration())
```

prints 0s

> so it is like the .AsTime

That is also not true

k8scat:
@Quinn-With-Two-Ns any other question？

PR #1956, opened 2025-05-14
Add poller autoscaler

Quinn-With-Two-Ns:
Add poller autoscaler

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -39,19 +39,36 @@ const (
 	ratioToForceCompleteWorkflowTaskComplete = 0.8
 )
 
+type workflowTaskPollerMode int
+
+const (
+	Mixed workflowTaskPollerMode = iota
+	NonSticky
+	Sticky
+)
+
 type (
-	// taskPoller interface to poll and process for task
+	// taskPoller interface to poll for tasks
 	taskPoller interface {
 		// PollTask polls for one new task
 		PollTask() (taskForWorker, error)
-		// ProcessTask processes a task
-		ProcessTask(interface{}) error
 		// Called when the poller will no longer be polled. Presently only useful for
 		// workflow workers.
 		Cleanup() error
 	}
 
+	// taskProcessor interface to process tasks
+	taskProcessor interface {
+		// ProcessTask processes a task
+		ProcessTask(interface{}) error
+	}
+
+	pollerScaleDecision struct {
+		pollRequestDeltaSuggestion int
+	}
+
 	taskForWorker interface {
+		scaleDecision() (pollerScaleDecision, bool)
 		isEmpty() bool
 	}
 
@@ -79,8 +96,36 @@ type (
 		gauge      metrics.Gauge
 	}
 
-	// workflowTaskPoller implements polling/processing a workflow task
 	workflowTaskPoller struct {
+		basePoller
+		mode             workflowTaskPollerMode
+		namespace        string
+		taskQueueName    string
+		identity         string
+		service          workflowservice.WorkflowServiceClient
+		taskHandler      WorkflowTaskHandler
+		contextManager   WorkflowContextManager
+		logger           log.Logger
+		dataConverter    converter.DataConverter
+		failureConverter converter.FailureConverter
+
+		stickyUUID                   string
+		StickyScheduleToStartTimeout time.Duration
+
+		pendingRegularPollCount int
+		pendingStickyPollCount  int
+		stickyBacklog           int64
+		requestLock             sync.Mutex
+		stickyCacheSize         int
+		eagerActivityExecutor   *eagerActivityExecutor
+
+		numNormalPollerMetric *numPollerMetric
+		numStickyPollerMetric *numPollerMetric
+	}
+
+	// workflowTaskProcessor implements processing of a workflow task and can create
+	// workflow task pollers
+	workflowTaskProcessor struct {
 		basePoller
 		namespace        string
 		taskQueueName    string
@@ -261,14 +306,14 @@ func (bp *basePoller) getCapabilities() *workflowservice.GetSystemInfoResponse_C
 	return bp.capabilities
 }
 
-// newWorkflowTaskPoller creates a new workflow task poller which must have a one to one relationship to workflow worker
-func newWorkflowTaskPoller(
+// newWorkflowTaskProcessor creates a new workflow task poller which must have a one to one relationship to workflow worker
+func newWorkflowTaskProcessor(
 	taskHandler WorkflowTaskHandler,
 	contextManager WorkflowContextManager,
 	service workflowservice.WorkflowServiceClient,
 	params workerExecutionParameters,
-) *workflowTaskPoller {
-	return &workflowTaskPoller{
+) *workflowTaskProcessor {
+	return &workflowTaskProcessor{
 		basePoller: basePoller{
 			metricsHandler:          params.MetricsHandler,
 			stopC:                   params.WorkerStopChannel,
@@ -337,8 +382,33 @@ func (wtp *workflowTaskPoller) PollTask() (taskForWorker, error) {
 	return workflowTask, nil
 }
 
+func (wtp *workflowTaskProcessor) createPoller(mode workflowTaskPollerMode) taskPoller {
+	return &workflowTaskPoller{
+		basePoller:                   wtp.basePoller,
+		mode:                         mode,
+		namespace:                    wtp.namespace,
+		taskQueueName:                wtp.taskQueueName,
+		identity:                     wtp.identity,
+		service:                      wtp.service,
+		taskHandler:                  wtp.taskHandler,
+		contextManager:               wtp.contextManager,
+		logger:                       wtp.logger,
+		dataConverter:                wtp.dataConverter,
+		failureConverter:             wtp.failureConverter,
+		stickyUUID:                   wtp.stickyUUID,
+		StickyScheduleToStartTimeout: wtp.StickyScheduleToStartTimeout,
+		pendingRegularPollCount:      wtp.pendingRegularPollCount,
+		pendingStickyPollCount:       wtp.pendingStickyPollCount,
+		stickyBacklog:                wtp.stickyBacklog,
+		stickyCacheSize:              wtp.stickyCacheSize,
+		eagerActivityExecutor:        wtp.eagerActivityExecutor,
+		numNormalPollerMetric:        wtp.numNormalPollerMetric,
+		numStickyPollerMetric:        wtp.numStickyPollerMetric,
+	}
+}
+
 // ProcessTask processes a task which could be workflow task or local activity result
-func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
+func (wtp *workflowTaskProcessor) ProcessTask(task interface{}) error {
 	if wtp.stopping() {
 		return errStop
 	}
@@ -353,7 +423,7 @@ func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
 	}
 }
 
-func (wtp *workflowTaskPoller) processWorkflowTask(task *workflowTask) (retErr error) {
+func (wtp *workflowTaskProcessor) processWorkflowTask(task *workflowTask) (retErr error) {
 	if task.task == nil {
 		// We didn't have task, poll might have timeout.
 		traceLog(func() {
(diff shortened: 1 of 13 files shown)


worker/worker.go
```diff
@@ -303,3 +308,24 @@ func SetBinaryChecksum(checksum string) {
 func InterruptCh() <-chan interface{} {
 	return internal.InterruptCh()
 }
+
+// NewPollerBehaviorSimpleMaximum creates a PollerBehavior that allows the worker to start up to a maximum number of pollers.
+//
+// NOTE: Experimental
+func NewPollerBehaviorSimpleMaximum(
```
Quinn-With-Two-Ns, line 325: Open question if these should take options instead

internal/worker.go
```diff
@@ -303,6 +331,24 @@ type (
 		//
 		// NOTE: Experimental
 		Tuner WorkerTuner
+
+		// Optional: If set, the worker will use the provided poller behavior when polling for workflow tasks.
+		// This is mutually exclusive with MaxConcurrentWorkflowTaskPollers.
```
yuandrew, line 377: Should we add to the docs of `MaxConcurrentWorkflowTaskPollers` that it's mutually exclusive with `WorkflowTaskPollerBehavior` too?

internal/internal_task_pollers.go
```diff
@@ -771,20 +840,28 @@ func (wtp *workflowTaskPoller) updateBacklog(taskQueueKind enumspb.TaskQueueKind
...
-//  1. if sticky execution is disabled, always poll for regular task queue
-//  2. otherwise:
-//     2.1) if sticky task queue has backlog, always prefer to process sticky task first
-//     2.2) poll from the task queue that has less pending requests (prefer sticky when they are the same).
-//
-// TODO: make this more smart to auto adjust based on poll latency
+//  1. if mode is NonSticky, always poll from regular task queue
+//  2. if mode is Sticky, always poll from sticky task queue
+//  3. if mode is Mixed
+//     3.1. sticky execution is disabled, always poll for regular task queue
```
yuandrew:
```suggestion
//     3.1. if sticky execution is disabled, always poll for regular task queue
```

internal/internal_worker.go
```diff
@@ -362,17 +377,19 @@ func newWorkflowTaskWorkerInternal(
...
-		identity:         laParams.Identity,
-		buildId:          laParams.getBuildID(),
-		logger:           laParams.Logger,
-		stopTimeout:      laParams.WorkerStopTimeout,
-		fatalErrCb:       laParams.WorkerFatalErrorCallback,
-		metricsHandler:   laParams.MetricsHandler,
+		taskPollers: []scalableTaskPoller{
+			newScalableTaskPoller(localActivityTaskPoller, params.Logger, NewPollerBehaviorSimpleMaximum(1)),
+		},
+		taskProcessor:  localActivityTaskPoller,
```
yuandrew, line 393: we're setting a "poller" as taskProcessor? That feels confusing

internal/internal_task_pollers.go
```diff
@@ -103,6 +120,33 @@ type (
 		numStickyPollerMetric *numPollerMetric
 	}
 
+	// workflowTaskProcessor implements processing of a workflow task and can create
```
yuandrew, line 123: I feel like the differentiation between Processor/Poller could be clearer. This comment says processor can create pollers, but `newWorkflowTaskProcessor` says it creates a new poller

internal/internal_worker_test.go
```diff
@@ -2794,8 +2794,8 @@ func assertWorkerExecutionParamsEqual(t *testing.T, paramsA workerExecutionParam
 	require.Equal(t, paramsA.WorkerActivitiesPerSecond, paramsB.WorkerActivitiesPerSecond)
 	require.Equal(t, paramsA.TaskQueueActivitiesPerSecond, paramsB.TaskQueueActivitiesPerSecond)
 	require.Equal(t, paramsA.StickyScheduleToStartTimeout, paramsB.StickyScheduleToStartTimeout)
-	require.Equal(t, paramsA.MaxConcurrentWorkflowTaskQueuePollers, paramsB.MaxConcurrentWorkflowTaskQueuePollers)
-	require.Equal(t, paramsA.MaxConcurrentActivityTaskQueuePollers, paramsB.MaxConcurrentActivityTaskQueuePollers)
+	//require.Equal(t, paramsA.MaxConcurrentWorkflowTaskQueuePollers, paramsB.MaxConcurrentWorkflowTaskQueuePollers)
+	//require.Equal(t, paramsA.MaxConcurrentActivityTaskQueuePollers, paramsB.MaxConcurrentActivityTaskQueuePollers)
```
yuandrew: need to find a new way to compare poller behavior?

internal/internal_worker.go
```diff
@@ -323,13 +322,29 @@ func newWorkflowTaskWorkerInternal(
 	if client != nil {
 		service = client.workflowService
 	}
-	poller := newWorkflowTaskPoller(taskHandler, contextManager, service, params)
-	worker := newBaseWorker(baseWorkerOptions{
-		pollerCount:      params.MaxConcurrentWorkflowTaskQueuePollers,
+	taskProcessor := newWorkflowTaskProcessor(taskHandler, contextManager, service, params)
+
+	var taskWorkers []scalableTaskPoller
```
yuandrew: There seem to be 3 terms that are all related but slightly different: task pollers, task workers, and task processor. 

Looking at the code here, `taskWorkers == taskPollers`, and looking down in the file, `taskProcessor: poller,`. Do we need all 3 terms separately? Or can we combine taskWorkers and taskPollers, and differentiate them more clearly from taskProcessor?

internal/internal_worker_base_test.go
```diff
@@ -0,0 +1,21 @@
...
+	PollScalerReportHandleSuite struct {
+		suite.Suite
+	}
+)
+
+func TestPollScalerReportHandleSuite(t *testing.T) {
+	suite.Run(t, new(PollScalerReportHandleSuite))
+}
+
+func (s *PollScalerReportHandleSuite) TestCreateAndFree() {
```
yuandrew: missing poller autoscaling tests?

internal/internal_worker.go
Quinn-With-Two-Ns: The `localActivityTaskPoller ` implements the poller interface and the task processing interface. I am not too inclined to split them , but do you have a better naming suggestion?

yuandrew: hmm, not really 😅 it's probably less important for the Local Activity case

worker/worker.go
Sushisource: Don't think so. Worst case scenario if we ever need to add anything it can be a new behavior.

internal/internal_worker.go
Quinn-With-Two-Ns: Yeah let me get rid of `taskWorkers`, it was a term I was using, but don't think it fits well

Quinn-With-Two-Ns: So we have two interfaces task pollers and task processor that do two different things (poll tasks and process those tasks). For some task types the same struct implements both interfaces.

internal/internal_worker.go
```diff
@@ -1897,6 +1911,36 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 		capabilities: &capabilities,
 	}
 
+	if options.MaxConcurrentWorkflowTaskPollers != 0 && options.WorkflowTaskPollerBehavior != nil {
+		panic("cannot set both MaxConcurrentWorkflowTaskPollers and WorkflowTaskPollerBehavior")
```
Sushisource: Feels like these should bubble up as errors to worker start rather than be panics, but, not part of this PR.

internal/internal_worker_base.go
```diff
@@ -344,52 +414,79 @@ func (bw *baseWorker) isStop() bool {
 	}
 }
 
-func (bw *baseWorker) runPoller() {
+func (bw *baseWorker) runPoller(taskWorker scalableTaskPoller) {
 	defer bw.stopWG.Done()
+	// TODO(quinn): with poller autoscaling, this metric doesn't make a lot of sense
+	// since the number of pollers can go up and down.
 	bw.metricsHandler.Counter(metrics.PollerStartCounter).Inc(1)
```
Sushisource: Does Go not have `num_pollers`? We should add if not.

internal/internal_worker_base.go
```diff
@@ -344,52 +414,79 @@ func (bw *baseWorker) isStop() bool {
...
-			if err != nil {
-				if !errors.Is(err, context.Canceled) {
-					bw.logger.Error(fmt.Sprintf("Error while trying to reserve slot: %v", err))
-					select {
-					case reserveChan <- nil:
-					case <-ctx.Done():
-						return
-					}
+		// TODO(quinn): is there a way we can refactor this to avoid the function?
+		if func() bool {
```
Sushisource: You could set a var which is this fn's return value and `continue` if false, but it makes the defer stuff harder too... I think this is totally fine.

internal/internal_worker_base.go
```diff
@@ -600,3 +706,246 @@ func (bw *baseWorker) Stop() {
...
+	if prh.scaleCallback != nil {
+		prh.logger.Debug("Updating number of permits", "permits", permits)
+		prh.scaleCallback(permits)
+	}
+}
+
+func (prh *pollScalerReportHandle) handleError(err error) {
+	_, resourceExhausted := err.(*serviceerror.ResourceExhausted)
+	if resourceExhausted {
+		prh.updateTarget(func(target int64) int64 {
```
Sushisource: As discussed, want to add the "ever saw scale decision" check here

worker/worker.go
```diff
@@ -303,3 +308,24 @@ func SetBinaryChecksum(checksum string) {
...
+	maximumConcurrentPollers int,
+) PollerBehavior {
+	return internal.NewPollerBehaviorSimpleMaximum(maximumConcurrentPollers)
+}
+
+// NewPollerBehaviorAutoscaling creates a PollerBehavior that allows the worker to scale the number of pollers within a given range.
+// based on the workflow and feedback from the server.
+//
+// NOTE: Experimental
+func NewPollerBehaviorAutoscaling(
```
Sushisource, line 335: We'll want some defaults here too per slack convo

Sushisource (commented):
Worth reviewing James' fix here https://github.com/temporalio/sdk-core/pull/941/files#diff-a65538899b187dda6f4128ecad6a38b908f9491bb63a4828f488e0e1d97f548fR121

To ensure that's not a problem here with the balancer either.

internal/internal_worker.go
Quinn-With-Two-Ns: Worker start can't fail so we have to panic unfortunately

internal/internal_worker_base.go
Quinn-With-Two-Ns: We do

internal/internal_worker.go
Sushisource: Ah, dang.

internal/worker.go
```diff
@@ -375,3 +468,55 @@ func workerDeploymentOptionsToProto(useVersioning bool, version WorkerDeployment
...
+//
+// NOTE: Experimental
+//
+// Exposed as: [go.temporal.io/sdk/worker.NewPollerBehaviorAutoscaling]
+func NewPollerBehaviorAutoscaling(
+	options PollerBehaviorAutoscalingOptions,
+) PollerBehavior {
+	initialNumberOfPollers := options.InitialNumberOfPollers
+	if initialNumberOfPollers <= 0 {
+		initialNumberOfPollers = defaultAutoscalingInitialNumberOfPollers // Default initial number of pollers.
```
yuandrew, line 507: Would it be helpful to log something here so users are aware their configuration is invalid?

yuandrew (approved):

internal/worker.go
Quinn-With-Two-Ns: This is more to catch a non-set options then an invalid configuration. Setting to below zero is invalid and we could error here, but we don't for the old poller setting so I kept the same logic here.

internal/internal_worker_base.go
```diff
@@ -604,3 +709,255 @@ func (bw *baseWorker) Stop() {
...
+		})
+	case *pollerBehaviorSimpleMaximum:
+		tw.pollerCount = p.maximumNumberOfPollers
+	}
+	return tw
+}
+
+// balance checks if the poller type is balanced with other poller types. The goal is to ensure that
+// at least one poller of each type is running before allowing any poller of the given type to increase.
+func (pb *pollerBalancer) balance(ctx context.Context, pollerType string) error {
```
Sushisource, line 900: I think this could potentially be susceptible to some of the stuff fixed in https://github.com/temporalio/sdk-core/pull/941 -- but I'm not sure. Adding a test is probably worth it.

Quinn-With-Two-Ns: Looking at the core issue, This code actually won't apply in the `SimpleMaximum` case. We didn't change any behaviour in the `SimpleMaximum ` case for Go

PR #1957, opened 2025-05-14
fix: use 127.0.0.1 instead of localhost as default HostPort for IPv4 consistency

AmitThakkar:
## What was changed
Updated the `defaultHostPort` constant in `internal/grpc_dialer.go` from:
```go
const defaultHostPort = "localhost:7233"
```
to:
```go
const defaultHostPort = "127.0.0.1:7233"
```
to ensure consistent and reliable connectivity in local development environments across platforms.

## Why?
The default HostPort value used when connecting to the Temporal server is set to "localhost:7233". However, "localhost" may resolve to ::1 (IPv6) on some systems, while temporal server start-dev binds only to 127.0.0.1 (IPv4). 

This mismatch leads to errors like:
> failed reaching server: context deadline exceeded

### DIFF
--- a/internal/grpc_dialer.go
+++ b/internal/grpc_dialer.go
@@ -32,7 +32,7 @@ const (
 	// LocalHostPort is a default host:port for worker and client to connect to.
 	//
 	// Exposed as: [go.temporal.io/sdk/client.DefaultHostPort]
-	LocalHostPort = "localhost:7233"
+	LocalHostPort = "127.0.0.1:7233"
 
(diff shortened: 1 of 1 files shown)


cretz:
I am concerned this is a backwards incompatible change. If we had the SDK to design again, we wouldn't even have a default host:port, we'd _require_ it be set by users (and this is what we do in newer SDKs). In the meantime, we encourage explicitly setting this value to what you need it to be and not relying on the default.

PR #1958, opened 2025-05-14
Add SearchAttributes.Copy exception to workflowcheck

### LINKED ISSUE #1943: `SearchAttributes.Copy` is getting flagged by `workflowcheck`

Need to mark `SearchAttributes.Copy` as ignored by `workflowcheck`

yuandrew:
## What was changed

Added an exception for `SearchAttributes.Copy`

## Why?
The map iteration inside is used deterministically

## Checklist

1. Closes #1943

2. How was this tested:
Tested locally, with the following code in a workflow
```
sa := temporal.SearchAttributes{}
	_ = sa.Copy()
```

3. Any docs updates needed?


Quinn-With-Two-Ns (approved):

contrib/tools/workflowcheck/determinism/ident_refs.go
```diff
@@ -26,6 +26,7 @@ var DefaultIdentRefs = IdentRefs{
 	"fmt.Sscan":                 false,
 	"fmt.Sscanf":                false,
 	"fmt.Sscanln":               false,
+	"(go.temporal.io/sdk/internal.SearchAttributes).Copy": false,
```
cretz, line 29: Since this is our code, can't we just put `//workflowcheck:ignore` on the func (or offending line)? Also, can we have a test that would fail before this exclusion is added?

yuandrew: good point, will add directly.

For writing a test, since the tests use `analysistest` which uses dummy data, I made a pseudo copy of SearchAttributes to test

cretz: :+1: And I assume that test would fail if you didn't have `//workflowcheck:ignore` on the function?

cretz (approved):

contrib/tools/workflowcheck/determinism/ident_refs.go
yuandrew: affirmative

PR #1962, opened 2025-05-30
:boom: [Breaking] Update versioning APIs to use struct

Sushisource:
DRAFT just to make sure tests pass & give myself a review pass

## What was changed
Change APIs to use struct form instead of string form for worker deployment versions & update protos & use new fields in them.

## Why?
Much more usable and less error prone, consistent with other SDKs

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?


internal/internal_worker_base.go
```diff
@@ -538,6 +541,14 @@ func (bw *baseWorker) logPollTaskError(err error) {
 		bw.lastPollTaskErrStarted = time.Now()
 		return
 	}
+
+	// Ignore connection loss on server shutdown. This helps with quiescing spurious error messages
+	// upon server shutdown (where server is using the SDK).
+	st, ok := status.FromError(err)
+	if ok && st.Code() == codes.Unavailable && strings.Contains(st.Message(), "graceful_stop") {
+		return
+	}
```
Sushisource: This is unrelated, but, just wanted to do it while I was in here - helps shut up some pointless logspew in the dev server.

internal/worker_deployments.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+
+	deploymentpb "go.temporal.io/api/deployment/v1"
+)
+
+// Represents the version of a specific worker deployment.
+//
+// NOTE: Experimental
+//
+// Exposed as: [go.temporal.io/sdk/client.WorkerDeploymentVersion], [go.temporal.io/sdk/worker.WorkerDeploymentVersion]
```
Quinn-With-Two-Ns: Why is this exposed twice?

internal/worker_deployments.go
```diff
@@ -0,0 +1,58 @@
+package internal
```
Quinn-With-Two-Ns: This doesn't need to be in its own file

internal/internal_worker_base.go
Quinn-With-Two-Ns: but a normal user would want to get logs if the poller received errors from the server is shutting down

internal/internal_workflow_execution_options.go
```diff
@@ -32,36 +34,48 @@ type (
...
-	// An entry with a `nil` pointer means do not change.
-	// An entry with a pointer to an empty value means delete the entry, i.e., the empty value is a tombstone.
-	// An entry with a pointer to a non-empty value means replace the entry, i.e., there is no deep merging.
+	// WorkflowExecutionOptionsChanges describes changes to the options of a workflow execution in
+	// [WorkflowExecutionOptions]. An entry with a `nil` pointer means do not change.
 	//
 	// NOTE: Experimental
 	WorkflowExecutionOptionsChanges struct {
-		VersioningOverride *VersioningOverride
+		VersioningOverride *VersioningOverrideChange
```
Quinn-With-Two-Ns, line 43: nit: Add docs and move the comment about `nil` down here

internal/worker_deployments.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (wd *WorkerDeploymentVersion) toProto() *deploymentpb.WorkerDeploymentVersion {
+	return &deploymentpb.WorkerDeploymentVersion{
+		DeploymentName: wd.DeploymentName,
+		BuildId:        wd.BuildId,
+	}
+}
+
+func (wd *WorkerDeploymentVersion) ToCanonicalString() string {
```
ShahabT: should this be also private the same way that `workerDeploymentVersionFromString` is? it is only useful for the deprecated fields, right?

internal/internal_worker_base.go
Sushisource: Yeah, that's a good point. We'd need some kind of indication we're in the server environment.

internal/worker_deployments.go
Sushisource: Just because it applies in both packages and that seems kinda useful. It show up in various client responses, but also as a parameter to worker options. If you'd prefer it in just one, no problem, I suppose client?

internal/internal_workflow_execution_options.go
Sushisource: I think you still want it up top for the future, but added it here too.

internal/internal_worker_base.go
Sushisource: Got some advice from David on what to check

internal/worker_deployments.go
Quinn-With-Two-Ns: We shouldn't expose it twice, that is just duplicating the object and confusing for users. I would just have the one in `worker`

.gitignore
```diff
@@ -21,4 +21,5 @@ test.log
 .gobincache
 go.work
 go.work.sum
-*~
\ No newline at end of file
+*~
+internal/cmd/build/*.sqlite*
```
cretz, line 25: Is something creating this file? (no comment on the review itself, LGTM)

Sushisource: Yeah, when you run the integ tests w/ dev server mode

cretz: @Quinn-With-Two-Ns - I see we changed from in-memory mode to DB filename mode via `DBFilename: "temporal.sqlite"` in #1893. Was that intentional?

Quinn-With-Two-Ns: yes

cretz: Why? I am concerned successive tests will reuse a DB from a previous test (and in general our tests should be able to run via in memory dev server).

PR #1965, opened 2025-06-05
Fix GetWorkerStopChannel() for Local Activities

### LINKED ISSUE #1963: GetWorkerStopChannel no longer indicates worker shutdown for Local Activities
## Expected Behavior
`GetWorkerStopChannel()` should indicate that a worker has begun shutting down

## Actual Behavior
`GetWorkerStopChannel()` for local activities now indicates when the local activity specific stop channel has closed. This is due to https://github.com/temporalio/sdk-go/pull/1875.

The solution should be to also keep the worker stop channel in the Local Activity poller and have GetWorkerStopChannel() return that, while also ensuring shutdown behaves properly and the LA specific stop channel is closed at the right time.

Once this is fixed, see TODO comment linking this issue in integration tests and have Local Activity wait on that channel closure, instead of a manual time wait, to avoid potential race conditions

  1.
  1.
  1.

## Specifications

  - Version:
  - Platform:

yuandrew:
## What was changed

Plumb WorkerStopChannel through for Local Activities

## Why?

Looks like this was never implemented to begin with

## Checklist

1. Closes #1963

2. How was this tested:

Changed a few tests to use this to avoid a race condition

3. Any docs updates needed?

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -7267,6 +7267,10 @@ func (ts *IntegrationTestSuite) TestSelectorBlock() {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 	options := ts.startWorkflowOptions("test-selector-block")
+
+	internal.SetUnblockSelectorSignal(false)
+	defer internal.SetUnblockSelectorSignal(true)
+
 	run, err := ts.client.ExecuteWorkflow(ctx, options, ts.workflows.SelectorBlockSignal)
 	ts.NoError(err)
 	var result string
@@ -7278,9 +7282,6 @@ func (ts *IntegrationTestSuite) TestSelectorNoBlock() {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 	options := ts.startWorkflowOptions("test-selector-block")
-
-	internal.SetUnblockSelectorSignal(true)
-	defer internal.SetUnblockSelectorSignal(false)
 
 	run, err := ts.client.ExecuteWorkflow(ctx, options, ts.workflows.SelectorBlockSignal)
 	ts.NoError(err)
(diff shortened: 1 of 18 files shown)


test/integration_test.go
```diff
@@ -2414,6 +2414,8 @@ func (ts *IntegrationTestSuite) TestGracefulLocalActivityCompletion() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	localActivityFn := func(ctx context.Context) error {
+		// There is a slight delay between when workerStopChannel closes and when backgroundContextCancel is canceled
```
Quinn-With-Two-Ns: Well the delay should be the graceful shutdown time no?

Quinn-With-Two-Ns: `workerStopChannel` should be closed during graceful shutdown and then the context should be cancelled when the graceful period elapse or the shutdown finished

yuandrew: hmm, looking at the graceful shutdown case, there is no delay.

but looking at `TestLocalActivityWorkerShutdownNoHeartbeat`, which I realized is non-graceful shutdown, we need a sleep. Otherwise, both LAs are able to complete before shutdown stops the LA.

Quinn-With-Two-Ns: hm for graceful shutdown there should be a delay between the channel closing and the context closing. The context shouldn't close during the graceful shutdown period

yuandrew: I was getting confused because the tests are doing different things. Added some comments to the tests to hopefully avoid any future confusion

For the non-graceful test, we want the LA to return `context canceled` err in order to test that we don't heartbeat on failure.

For the graceful tests, we want the LA to wait for shutdown to start before we complete, so when ctx is canceled doesn't matter in these tests

> The context shouldn't close during the graceful shutdown period

I've separately confirmed that this is true.

PR #1982, opened 2025-06-20
Add cause to RespondQueryTaskCompletedRequest

### LINKED ISSUE #1971: [go] Add `WorkflowTaskFailedCause` to `RespondQueryTaskCompletedRequest`
Should work the same as for workflow task failures

yuandrew:
## What was changed

Support adding `Cause` to RespondQueryTaskCompletedRequest

Need to wait for next `api-go` release before merging

## Why?

API change

## Checklist

1. Closes #1971

2. How was this tested:

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1822,9 +1822,11 @@ func (wth *workflowTaskHandlerImpl) completeWorkflow(
 ) interface{} {
 	// for query task
 	if task.Query != nil {
+		cause := errToWorkflowTaskFailedCause(workflowContext.err)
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -1822,9 +1822,11 @@ func (wth *workflowTaskHandlerImpl) completeWorkflow(
 ) interface{} {
 	// for query task
 	if task.Query != nil {
+		cause := errToWorkflowTaskFailedCause(workflowContext.err)
```
Quinn-With-Two-Ns, line 1825: Nit: To test you might want to use a gRPC interceptor to verify the cause is sent since it won't be written to history

PR #1984, opened 2025-06-25
fix: add timestamp to trigger immediately signal

spkane31:
## What was changed
add a timestamp to TriggerImmediatelyRequest

## Why?
We can make the wft “more idempotent”: put the desired trigger time in the trigger immediately request itself, so it goes in the signal payload, so all wfts attempt to start the target with the same id.

## Checklist

1. Closes 

2. How was this tested:

Covered by existing functional tests

3. Any docs updates needed?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/nexus-rpc/sdk-go v0.3.0
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.10.0
-	go.temporal.io/api v1.49.1
+	go.temporal.io/api v1.50.1-0.20250625152634-34d23b4cdd1f
 	golang.org/x/sync v0.13.0
 	golang.org/x/sys v0.32.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/nexus-rpc/sdk-go v0.3.0
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.10.0
-	go.temporal.io/api v1.49.1
+	go.temporal.io/api v1.50.1-0.20250625152634-34d23b4cdd1f
```
yuandrew, line 17: We should wait for an `api` release and point to an official release version to ingest this change

dnr (commented):
There's not really any reason to make any change in sdk-go at all. The server can fill in the current time. No one has asked for the ability to override the timestamp for trigger immediately. If someone does we could add it then, otherwise better to just avoid all the extra work in all sdks

PR #1996, opened 2025-07-08
Add warning when registering struct functions as workflows

### LINKED ISSUE #1986: SDK should warn when using structures to define a workflow

I've implemented my workflows on a struct and the workflow funcs have a pointer receiver. This may have lead to errors and it's possible to detect this with static code analysis to avoid this programming error in the future.

The `workflowcheck` command should report this.

Implementing a custom linter seems unnecessary since `workflowcheck` already exists. Its name also leaves room for analysis beyond determinism.

**Additional context**

As discussed here https://community.temporal.io/t/go-sdk-non-deterministic-error/12393

yuandrew:
## What was changed

Added a warning log saying we suggest only using structs for activity registration

## Why?

Help prevent users from hitting potential NDEs

## Checklist

1. Closes #1986 

2. How was this tested:

N/A, only added a log statement

3. Any docs updates needed?

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1147,6 +1147,10 @@ func (aw *AggregatedWorker) RegisterWorkflow(w interface{}) {
 		aw.executionParams.DefaultVersioningBehavior == VersioningBehaviorUnspecified {
 		panic("workflow type does not have a versioning behavior")
 	}
+	_, ok := w.(WorkflowDefinitionFactory)
+	if ok {
+		aw.logger.Warn("It is recommended to only use structures to define activity functions. In some cases, struct workflow functions may cause NDE errors.")
+	}
 	aw.registry.RegisterWorkflow(w)
 }
 
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1147,9 +1147,9 @@ func (aw *AggregatedWorker) RegisterWorkflow(w interface{}) {
 		aw.executionParams.DefaultVersioningBehavior == VersioningBehaviorUnspecified {
 		panic("workflow type does not have a versioning behavior")
 	}
-	_, ok := w.(WorkflowDefinitionFactory)
-	if ok {
-		aw.logger.Warn("It is recommended to only use structures to define activity functions. In some cases, struct workflow functions may cause NDE errors.")
+	fnName := runtime.FuncForPC(reflect.ValueOf(w).Pointer()).Name()
+	if strings.HasSuffix(fnName, "-fm") {
+		aw.logger.Warn("It is recommended to only use struct methods to define activity functions. In some cases, struct methods as workflows may cause NDE errors.")
 	}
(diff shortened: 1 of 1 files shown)


internal/internal_worker.go
```diff
@@ -1147,6 +1147,10 @@ func (aw *AggregatedWorker) RegisterWorkflow(w interface{}) {
 		aw.executionParams.DefaultVersioningBehavior == VersioningBehaviorUnspecified {
 		panic("workflow type does not have a versioning behavior")
 	}
+	_, ok := w.(WorkflowDefinitionFactory)
```
Quinn-With-Two-Ns: Will this effect the PHP SDK? I think this functionality is used by them

yuandrew: I misunderstood how the user was registering struct methods, the new commit should be correct, manually repro'd the issue.

Is PHP SDK built on top of Go?

Quinn-With-Two-Ns: >Is PHP SDK built on top of Go?

Yep

internal/internal_worker.go
```diff
@@ -1147,6 +1147,10 @@ func (aw *AggregatedWorker) RegisterWorkflow(w interface{}) {
 		aw.executionParams.DefaultVersioningBehavior == VersioningBehaviorUnspecified {
 		panic("workflow type does not have a versioning behavior")
 	}
+	fnName := runtime.FuncForPC(reflect.ValueOf(w).Pointer()).Name()
+	if strings.HasSuffix(fnName, "-fm") {
+		aw.logger.Warn("It is recommended to only use struct methods to define activity functions. In some cases, struct methods as workflows may cause NDE errors.")
```
Quinn-With-Two-Ns: nit could we log the function they tried to register here too?

internal/internal_worker.go
```diff
@@ -1147,6 +1147,10 @@ func (aw *AggregatedWorker) RegisterWorkflow(w interface{}) {
 		aw.executionParams.DefaultVersioningBehavior == VersioningBehaviorUnspecified {
 		panic("workflow type does not have a versioning behavior")
 	}
+	fnName := runtime.FuncForPC(reflect.ValueOf(w).Pointer()).Name()
+	if strings.HasSuffix(fnName, "-fm") {
+		aw.logger.Warn("Registering workflow" + fnName + ". It is recommended to only use struct methods to define activity functions. In some cases, struct methods as workflows may cause NDE errors.")
```
cretz: Pedantic, but would avoid the acronym NDE in user-facing contexts

yuandrew: thanks for the suggestion :)

yuandrew:
Decided internally that since this spams our test logs (our test workflows are all methods under the same struct), for now we'll add something in the documentation warning against this.

Our tests can be changed to not share the same struct, but that's a rather large refactor.

PR #2007, opened 2025-07-30
Add `ErrGroup` and `Once` types

opeik:
## What was changed
- Adds `ErrGroup` and `Once` types
  - `ErrGroup` ported from [`x/sync.Group`](https://cs.opensource.google/go/x/sync/+/refs/tags/v0.15.0:errgroup/errgroup.go) v0.15.0
  - `Once` ported from [`sync.Once`](https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/sync/once.go) v1.24.4
- Adds `workflow.NewErrGroup` and `workflow.NewOnce`

## Why?
- Because `WaitGroup` can't handle errors and has no concurrency limits

## Checklist
1. How was this tested:
- All the original tests were ported

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"reflect"
 	"runtime"
+	"runtime/debug"
 	"slices"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


Quinn-With-Two-Ns:
Hey thanks for the contribution, but we have already discussed internally adding these helpers and decided against it. `ErrGroup` is not a stable interface and the original author has some major issues with the API. Once `ErrGroup` is added to the Go standard library as a stable feature then we will add a Temporal wrapper for it.

opeik:
Ah well, thanks for taking a look.

PR #2009, opened 2025-07-31
Move Nexus operation token generation outside of callback conditional

pdoerner:
## What was changed
Moved Nexus operation token generation outside the conditional the checks whether a callback URL has been set.

## Why?
Callbacks are optional in the Nexus protocol and many non-workflow callers will start operations without them but still need to get the operation token.


bergundy (approved):
Thanks! No reason why the code was in the conditional.

PR #2012, opened 2025-08-04
Add Fairness Keys/Weights

Sushisource:
> [!WARNING]
> DRAFT until we have a CLI RC with this in it that can be used for tests 

## What was changed
Added fairness keys/weights to priority

## Why?
New feature

## Checklist

1. Closes 

2. How was this tested:
Added test

3. Any docs updates needed?

### DIFF
--- a/test/worker_versioning_test.go
+++ b/test/worker_versioning_test.go
@@ -832,7 +832,7 @@ func (ts *WorkerVersioningTestSuite) TestTaskQueueStats() {
 		ts.NoError(err)
 		ts.Equal(1, len(taskQueueInfo.VersionsInfo))
 
-		// TODO: Fix to work with newer response format
+		// TODO: Fix to work with newer response format - https://github.com/temporalio/sdk-go/issues/2025
 		// ts.validateTaskQueueStats(expectedWorkflowStats, taskQueueInfo.VersionsInfo[""].TypesInfo[client.TaskQueueTypeWorkflow].Stats)
 		// ts.validateTaskQueueStats(expectedActivityStats, taskQueueInfo.VersionsInfo[""].TypesInfo[client.TaskQueueTypeActivity].Stats)
 	}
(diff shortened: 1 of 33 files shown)


test/worker_versioning_test.go
```diff
@@ -832,8 +832,9 @@ func (ts *WorkerVersioningTestSuite) TestTaskQueueStats() {
 		ts.NoError(err)
 		ts.Equal(1, len(taskQueueInfo.VersionsInfo))
 
-		ts.validateTaskQueueStats(expectedWorkflowStats, taskQueueInfo.VersionsInfo[""].TypesInfo[client.TaskQueueTypeWorkflow].Stats)
-		ts.validateTaskQueueStats(expectedActivityStats, taskQueueInfo.VersionsInfo[""].TypesInfo[client.TaskQueueTypeActivity].Stats)
+		// TODO: Fix to work with newer response format
```
Quinn-With-Two-Ns: Are you going to address this in this PR

Sushisource: No, I think it needs a more comprehensive change and I don't want to combine them together. I'll make an item for it though and reference it.

test/workflow_test.go
```diff
@@ -3242,54 +3242,66 @@ func (w *Workflows) PriorityWorkflow(ctx workflow.Context) (int, error) {
...
-	// Verify the activity returned the expected priority
-	if result != 5 {
-		return 0, fmt.Errorf("activity did not return expected value %d != %d", 5, result)
+	if result.PriorityKey != 5 {
+		return 0, fmt.Errorf("activity did not return expected priority %d != %d", 5, result.PriorityKey)
+	}
+	if result.FairnessKey != "fair-activity" {
+		return 0, fmt.Errorf("activity did not return expected fairness key %s != %s", "fair-activity", result.FairnessKey)
+	}
+	if result.FairnessWeight != 4.20 {
```
yuandrew, line 3262: Is the extra 0 here needed for correctness? I see the test below doesn't have an extra 0 in the decimal,

Sushisource: No, it's pointless. I just wrote it that way lol

PR #2016, opened 2025-08-07
Improve Nexus cancellation type test assertions

pdoerner:
## What was changed
Small improvements to Nexus cancellation types test assertions

### DIFF
--- a/test/nexus_test.go
+++ b/test/nexus_test.go
@@ -1061,24 +1061,48 @@ func TestAsyncOperationFromWorkflow(t *testing.T) {
 	})
 }
 
-func runCancellationTypeTest(ctx context.Context, tc *testContext, cancellationType workflow.NexusOperationCancellationType, t *testing.T) (client.WorkflowRun, string, time.Time) {
+// cancelTypeOp is a wrapper for a workflow run operation that delays responding to the cancel request so that time
+// based assertions aren't flakey.
+type cancelTypeOp struct {
+	nexus.UnimplementedOperation[string, string]
+	workflowRunOp nexus.Operation[string, string]
+	cancelDelay   time.Duration
+}
+
+func (o *cancelTypeOp) Name() string {
+	return o.workflowRunOp.Name()
+}
+
+func (o *cancelTypeOp) Start(ctx context.Context, input string, options nexus.StartOperationOptions) (nexus.HandlerStartOperationResult[string], error) {
+	return o.workflowRunOp.Start(ctx, input, options)
+}
+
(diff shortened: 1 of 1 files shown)


test/nexus_test.go
```diff
@@ -1061,24 +1061,48 @@ func TestAsyncOperationFromWorkflow(t *testing.T) {
 	})
 }
 
-func runCancellationTypeTest(ctx context.Context, tc *testContext, cancellationType workflow.NexusOperationCancellationType, t *testing.T) (client.WorkflowRun, string, time.Time) {
+// cancelTypeOp is a wrapper for a workflow run operation that delays responding to the cancel request so that time
+// based assertions aren't flakey.
+type cancelTypeOp struct {
+	nexus.UnimplementedOperation[string, string]
+	workflowRunOp nexus.Operation[string, string]
+	cancelDelay   time.Duration
```
bergundy: Why would you even need a timer, that's unclear to me. Timers in tests increase chance of flakiness, please find a better alternative.

pdoerner: We have to delay the cancellation request. The only other option would be channels I guess. This seemed simpler and the test was already using timers.

test/nexus_test.go
```diff
@@ -1061,24 +1060,50 @@ func TestAsyncOperationFromWorkflow(t *testing.T) {
...
+func (o *cancelTypeOp) Start(ctx context.Context, input string, options nexus.StartOperationOptions) (nexus.HandlerStartOperationResult[string], error) {
+	return o.workflowRunOp.Start(ctx, input, options)
+}
+
+func (o *cancelTypeOp) Cancel(ctx context.Context, token string, options nexus.CancelOperationOptions) error {
+	if o.unblockCancelCh != nil {
+		// Should only be non-nil in the TRY_CANCEL case.
+		<-o.unblockCancelCh
+	}
+	return o.workflowRunOp.Cancel(ctx, token, options)
```
bergundy, line 1084: Why do you need to also eventually cancel in the TRY_CANCEL case? What's the expected behavior when the workflow is eventually canceled? Do we assert anything?

bergundy (approved):
I don't have a lot of context on these tests but definitely endorse using predictable signals and channels over timers.

test/nexus_test.go
pdoerner: I just need to delay the cancel request. It doesn't have to eventually cancel, but immediately returning an error would just cause a `NEXUS_OPERATION_CANCEL_REQUEST_FAILED` event which would unblock `WAIT_REQUESTED` making TRY_CANCEL and WAIT_REQUESTED the same.

PR #2026, opened 2025-08-20
Do not auto-retry gRPC-message-size-too-large errors

### LINKED ISSUE #995: Do not auto-retry gRPC-message-size-too-large errors
## Expected Behavior

When server gets > 4MB message, it returns `ResourceExhausted` error with something like:

> grpc: received message larger than max (10248653 vs. 4194304)

Which is retried by our gRPC interceptor as are all other `ResourceExhausted`. We need to stop retrying this one in particular but still retry others. If we can prevent it from going over the wire without pre-calculating the size that would be best, but otherwise we may need to hack something to change the error type for this particular error before it gets to the retry interceptor. Yes that may mean a message string check which is admittedly gross.

Also, add an integration test with a workflow whose activity payload will trigger this and confirm behavior (can be in sdk-features).

maciejdudko:
## What was changed

- Added an interceptor to `grpc_dialer` that detects errors caused by gRPC message being too large.
- Changed gRPC retry logic to not retry on that kind of error, but continue retrying other RESOURCE_EXHAUSTED errors.
- Made `RespondTaskCompletedWithMetrics` detect gRPC-message-too-large error and report it to server as a separate WFT failure request.
- As a side effect, fixed the "is error retryable" check in `temporalInvoker.internalHeartBeat` to correctly handle internal errors.

## Why?

Feature request: https://github.com/temporalio/features/issues/624

## Checklist

1. Closes #995 

2. How was this tested:

Added `TestSignalWithStartWorkflowTypedSearchAttributes` to `test/integration_test.go`.


internal/common/retry/interceptor.go
```diff
@@ -145,12 +142,56 @@ func NewRetryOptionsInterceptor(excludeInternal *atomic.Bool) grpc.UnaryClientIn
...
+			if failure, ok := details[0].(*errordetails.ResourceExhaustedFailure); ok {
+				return failure.Cause != RESOURCE_EXHAUSTED_CAUSE_EXT_GRPC_MESSAGE_TOO_LARGE
+			}
+		}
+	}
 	return false
 }
+
+// RESOURCE_EXHAUSTED_CAUSE_EXT_GRPC_MESSAGE_TOO_LARGE is an extension to the ResourceExhaustedCause enum to mark gRPC message too large errors.
+const RESOURCE_EXHAUSTED_CAUSE_EXT_GRPC_MESSAGE_TOO_LARGE enumspb.ResourceExhaustedCause = 101 // TODO: add the cause to the upstream API repo and remove this
```
yuandrew: Is there an issue # we can attach to this TODO comment?

yuandrew (approved):
LGTM

internal/common/retry/interceptor.go
maciejdudko: Created issue: https://github.com/temporalio/sdk-go/issues/2030

Will add it to the TODO comment.

PR #2041, opened 2025-08-28
Allow Workflows to Be Started with Signals in Nexus Operations

elb3k:
## What was changed
Added an option to start the workflow with a signal.

## Why?
One of our use cases requires starting a workflow with a signal. I could not find any other way to achieve that.


Quinn-With-Two-Ns:
You can already do this in the SDK by setting `Handler` in [WorkflowRunOperationOptions](https://pkg.go.dev/go.temporal.io/sdk@v1.36.0/temporalnexus#WorkflowRunOperationOptions)

Quinn-With-Two-Ns:
Note: this is not the same as signal with start, there is a potential race doing it client side, we do want to add support for starting a Nexus operation backed by a workflow using signal with start, but I do not have a timeline for that support. To follow Nexus development I'd encourage you to follow out [Nexus](https://temporal.io/community) community slack channel

PR #2042, opened 2025-08-28
Changes to gRPC message too large error handling

### LINKED ISSUE #2030: Move RESOURCE_EXHAUSTED_CAUSE_EXT_GRPC_MESSAGE_TOO_LARGE to upstream API
#2026 added a constant `RESOURCE_EXHAUSTED_CAUSE_EXT_GRPC_MESSAGE_TOO_LARGE` as an extension of the ResourceExhaustedCause enum for better handling of gRPC message too large errors. To avoid compatibility issues, the upstream API definition of the enum should have this constant added (with the next ordinal integer value and without `EXT` in the name). After that, the SDK code should be changed to remove the extension constant and use the constant in upstream API instead.

### LINKED ISSUE #2033: Flaky tests: TestIntegrationSuite/TestGrpcMessageTooLarge
#2026 added `TestIntegrationSuite/TestGrpcMessageTooLarge` test suite. It's sometimes failing in CI.

The cause is unknown, but one potential theory is that it's due to timeouts. The test should be rewritten so that it does not rely on timing of errors.

maciejdudko:
## What was changed
1. Changed gRPC message too large error handling to wrap error in a new type instead of using non-standard cause enum.
2. Improved `TestGrpcMessageTooLarge` to not rely on timeouts.

If this PR is merged, we should close the API PR temporalio/api#635 as it's no longer needed.

## Why?
1. The code delivered in #2026 required extending upstream API. This PR presents an alternative approach that does not require changes to upstream API.
2. The tests should stop flaking now.

## Checklist

1. Closes #2030
Closes #2033 

2. How was this tested:

Updated `TestIntegrationSuite/TestGrpcMessageTooLarge`

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -7751,23 +7751,17 @@ func (ts *IntegrationTestSuite) TestLocalActivitySummary() {
 }
 
 func (ts *IntegrationTestSuite) TestGrpcMessageTooLarge() {
-	ts.T().Skip("issue-2033: Test is flaky")
-	assertGrpcErrorInHistoryOnce := func(ctx context.Context, run client.WorkflowRun) {
-		found := false
+	assertGrpcErrorInHistory := func(ctx context.Context, run client.WorkflowRun) {
 		iter := ts.client.GetWorkflowHistory(ctx, run.GetID(), run.GetRunID(), true, enumspb.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT)
 		for iter.HasNext() {
 			event, err := iter.Next()
 			ts.NoError(err)
 			if event.EventType == enumspb.EVENT_TYPE_WORKFLOW_TASK_FAILED {
-				if found {
-					ts.Fail("Found more than 1 workflow task failed event in history")
-				} else {
-					found = true
-				}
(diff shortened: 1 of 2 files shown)


test/integration_test.go
```diff
@@ -7801,25 +7795,22 @@ func (ts *IntegrationTestSuite) TestGrpcMessageTooLarge() {
 	ts.worker.RegisterWorkflow(failureInQueryTaskWorkflowFn)
 	ts.worker.RegisterActivity(activityFn)
 	startOptions := client.StartWorkflowOptions{
-		TaskQueue:           ts.taskQueueName,
-		WorkflowTaskTimeout: 300 * time.Millisecond,
-		WorkflowRunTimeout:  1 * time.Second,
```
yuandrew, line 7806: How was this test previously relying on these timeouts? Seems like they're just taken out

yuandrew (approved):
agree with what's been said, if this is SDK-only, makes sense to avoid API changes

test/integration_test.go
maciejdudko: The test waited until the whole workflow timed out and then checked if the history contained the expected error. The low task timeout forced multiple tasks to be issued, to verify the error is only recorded once. But if the first attempt task timed out (which happened in CI presumably due to server load), the test would randomly fail. The new test doesn't wait for timeout at all and just reads history of a still running task.

Issue #2034, opened 2025-08-26
[Feature Request] Reclassify Benign Application errors in OpenTelemetry


THardy98:
See issue description here:

https://github.com/temporalio/sdk-python/issues/1041

PR #2048, opened 2025-09-10
OpenTelemetry: Specify span kind and don't set error type for benign exception

### LINKED ISSUE #2024: Set Span Kinds in Otel tracer interceptors
Python, .NET, and Ruby all set span kinds in their OTel tracer interceptors

Go should also set their span kinds. `client` as the outbound span kind, and `server` as the inbound span kind

**Additional context**

yuandrew:
## What was changed

1. Set span kind, `server` for inbound, and `client` for outbound
  1. Also removed `ToHeader` and `FromHeader`, they appear to directly be associated with if inbound or outbound 
2. Don't mark benign exceptions as Errors for otel status. Also avoid setting error tag for opentracing

## Why?

1. Match other SDKs
2. benign exceptions are expected errors, shouldn't be marking otel status as actual errors

## Checklist

1. Closes #2024 and #2034 

4. How was this tested:

Added tests

5. Any docs updates needed?

### DIFF
--- a/contrib/opentelemetry/tracing_interceptor.go
+++ b/contrib/opentelemetry/tracing_interceptor.go
@@ -196,8 +196,13 @@ func (t *tracer) StartSpan(opts *interceptor.TracerStartSpanOptions) (intercepto
 		}
 	}
 
+	spanKind := trace.SpanKindServer
+	if opts.Outbound {
+		spanKind = trace.SpanKindClient
+	}
+
 	// Create span
-	span := t.options.SpanStarter(ctx, t.options.Tracer, opts.Operation+":"+opts.Name, trace.WithTimestamp(opts.Time))
+	span := t.options.SpanStarter(ctx, t.options.Tracer, opts.Operation+":"+opts.Name, trace.WithTimestamp(opts.Time), trace.WithSpanKind(spanKind))
 
 	// Set tags
 	if len(opts.Tags) > 0 {
--- a/contrib/opentelemetry/tracing_interceptor.go
+++ b/contrib/opentelemetry/tracing_interceptor.go
@@ -3,6 +3,7 @@ package opentelemetry
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"go.opentelemetry.io/otel"
@@ -14,6 +15,7 @@ import (
 
 	"go.temporal.io/sdk/interceptor"
 	"go.temporal.io/sdk/log"
+	"go.temporal.io/sdk/temporal"
 )
 
 // DefaultTextMapPropagator is the default OpenTelemetry TextMapPropagator used
@@ -246,10 +248,20 @@ type tracerSpan struct {
 }
 
 func (t *tracerSpan) Finish(opts *interceptor.TracerFinishSpanOptions) {
-	if opts.Error != nil {
+	if opts.Error != nil && !isBenignApplicationError(opts.Error) {
 		t.SetStatus(codes.Error, opts.Error.Error())
 	}
 	t.End()
+}
+
+func isBenignApplicationError(err error) bool {
+	var appErr *temporal.ApplicationError
+	if temporal.IsApplicationError(err) {
+		if errors.As(err, &appErr) {
+			return appErr.Category() == temporal.ApplicationErrorCategoryBenign
+		}
+	}
+	return false
 }
 
 type textMapCarrier map[string]string
--- a/contrib/opentracing/interceptor.go
+++ b/contrib/opentracing/interceptor.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"go.temporal.io/sdk/temporal"
 
 	"github.com/opentracing/opentracing-go"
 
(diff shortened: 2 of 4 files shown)


contrib/opentracing/interceptor.go
```diff
@@ -164,9 +165,19 @@ type tracerSpanRef struct{ opentracing.SpanContext }
 type tracerSpan struct{ opentracing.Span }
 
 func (t *tracerSpan) Finish(opts *interceptor.TracerFinishSpanOptions) {
-	if opts.Error != nil {
+	if opts.Error != nil && !isBenignApplicationError(opts.Error) {
```
Quinn-With-Two-Ns: Open tracing is [deprecated](https://www.cncf.io/blog/2022/01/31/cncf-archives-the-opentracing-project/), not sure we should bother fixing it here/ do a release to publish this

interceptor/tracing_interceptor.go
```diff
@@ -147,6 +139,14 @@ type TracerStartSpanOptions struct {
...
 	// Do not attempt to parse it, as the format is subject to change.
 	IdempotencyKey string
+
+	// Outbound is true if the span is outbound and false if it is inbound.
+	//
+	// This is used internally by the interceptor to determine whether to read
+	// from the header or write to the header, where outbound means the span should
+	// be placed on the Temporal header, and inbound means the parent span can be
+	// retrieved from the Temporal header.
+	Outbound bool
```
Quinn-With-Two-Ns: Note: A change like this would break compatibility between new SDK and old tracing modules since they are versioned independently

contrib/opentelemetry/tracing_interceptor.go
```diff
@@ -196,8 +198,13 @@ func (t *tracer) StartSpan(opts *interceptor.TracerStartSpanOptions) (intercepto
 		}
 	}
 
+	spanKind := trace.SpanKindServer
```
Quinn-With-Two-Ns, line 204: Does this line up roughly with what other SDKs like dotnet or python do?

interceptor/tracing_interceptor.go
Quinn-With-Two-Ns: This is also part of the SDK public API so technically our backwards compatibility guarantees apply.

contrib/opentelemetry/tracing_interceptor.go
cretz: I _think_ so (though I think I just spotted a bug in Python for outbound signal child workflow)

contrib/opentelemetry/tracing_interceptor.go
```diff
@@ -241,12 +248,22 @@ type tracerSpan struct {
 }
 
 func (t *tracerSpan) Finish(opts *interceptor.TracerFinishSpanOptions) {
-	if opts.Error != nil {
+	if opts.Error != nil && !isBenignApplicationError(opts.Error) {
```
cretz, line 256: I think our implementation may not have been doing the right thing here because we are not calling `.RecordError`, which I think we should always do on error regardless of whether benign (but status as error only for benign like you have here). I don't know if we consider now starting to record errors as a breaking/dangerous change, but probably not.

contrib/opentelemetry/tracing_interceptor.go
```diff
@@ -241,12 +248,22 @@ type tracerSpan struct {
...
 		t.SetStatus(codes.Error, opts.Error.Error())
 	}
 	t.End()
 }
 
+func isBenignApplicationError(err error) bool {
+	var appErr *temporal.ApplicationError
+	if temporal.IsApplicationError(err) {
+		if errors.As(err, &appErr) {
```
cretz: Per https://github.com/temporalio/sdk-go/pull/1925#discussion_r2059166881, we should only look at the top-level error to determine whether benign. Mimic the `isBenignApplicationError` already in `internal/error.go`.

contrib/opentracing/interceptor.go
yuandrew: sounds good, reverted changing `contrib/opentracing/`

interceptor/tracing_interceptor.go
yuandrew: gotcha, reverting this change, although I added a check that ToHeader and FromHeader aren't both set. Technically that can be a breaking change, but this is correcting invalid usage, so should be fine?

contrib/opentelemetry/tracing_interceptor.go
yuandrew: should be, 

see https://github.com/temporalio/sdk-python/blob/ba3d50f750ea8e1219eb30234a3a9d6a58572351/temporalio/contrib/opentelemetry.py#L345

https://github.com/temporalio/sdk-dotnet/blob/092ed33e3a2bacf59ec10d93b13bd0d7a8861e22/src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs#L759

contrib/opentelemetry/tracing_interceptor.go
yuandrew: Looking at the godoc of `RecordError`, I think you're right. This doesn't seem like a dangerous change, but kinda breaking? But I feel like something we probably want to do, to fix behavior

contrib/opentelemetry/tracing_interceptor.go
yuandrew: Thanks!

Quinn-With-Two-Ns (approved):

Issue #2053, opened 2025-09-13
Expose Temporal Nexus operation info to Temporal Nexus operation authors


Quinn-With-Two-Ns:
https://github.com/temporalio/features/issues/674

PR #2054, opened 2025-09-13
Add NexusInfo

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-go/issues/2053


temporalnexus/operation.go
```diff
@@ -31,6 +31,19 @@ import (
 	"go.temporal.io/sdk/workflow"
 )
 
+// Info contains information about a currently executing Nexus operation.
+type Info = internal.NexusInfo
```
cretz: Arguably this could be called `OperationInfo`, but meh

Issue #1208, opened 2023-08-24
Canceling a context with multiple child contexts can be non deterministic


Quinn-With-Two-Ns:
## Expected Behavior
Child context cancellation propagates in a deterministic order 

## Actual Behavior
Child context cancellation does not propagates in a deterministic order.

When a parent context cancels its' children it loops [through a map](https://github.com/temporalio/sdk-go/blob/b9e5e241869fdb39b03ef8b6ef043ce75c78344e/internal/context.go#L338). Map order in Go is not deterministic so the order child contexts are canceled is not deterministic.

## Steps to Reproduce the Problem
```
func TestContextCancellationOrderDeterminism(t *testing.T) {
	/*
		Previously, child-contexts were stored in a map, preventing deterministic order when propagating cancellation.
		The order of branches being selected in this test was random, both for the first event and in following ones.

		In principle this should be fine, but it's possible for the effects of cancellation to trigger a selector's
		future-done callback, which currently records the *real-time*-first event as the branch to unblock, rather than
		doing something more safe by design (e.g. choosing based on state when the selector's goroutine is unblocked).

		Unfortunately, we cannot change the selector's behavior without introducing non-backwards-compatible changes to
		currently-working workflows.

		So the workaround for now is to maintain child-context order, so they are canceled in a consistent order.
		As this order was not controlled before, and Go does a pretty good job at randomizing map iteration order,
		converting non-determinism to determinism should be strictly no worse for backwards compatibility, and it
		fixes the issue for future executions.
	*/
	check := func(t *testing.T, separateStart, separateSelect bool) {
		env := newTestWorkflowEnv(t)
		act := func(ctx context.Context) error {
			return nil // will be mocked
		}
		wf := func(ctx Context) ([]int, error) {
			ctx, cancel := WithCancel(ctx)
			Go(ctx, func(ctx Context) {
				_ = Sleep(ctx, time.Minute)
				cancel()
			})

			// start some activities, which will not complete before the timeout cancels them
			ctx = WithActivityOptions(ctx, ActivityOptions{
				TaskQueue:              "",
				ScheduleToCloseTimeout: time.Hour,
				ScheduleToStartTimeout: time.Hour,
				StartToCloseTimeout:    time.Hour,
			})
			s := NewSelector(ctx)
			var result []int
			for i := 0; i < 10; i++ {
				i := i
				// need a child context, a future alone is not enough as it does not become a child
				cctx, ccancel := WithCancel(ctx)

[...]

PR #2055, opened 2025-09-16
Cancellation of child cancelers is non-deterministic

mattpercy-anz:
## What was changed
Moved from using a map internally to represent child contexts to cancel, moved to using a slice (but added similar functionality for checking existing items aren't added twice)

## Why?
We had noticed one of our replay tests failing intermittently, so had a closer look. The replay was specifically to do with cancellation of a workflow.
In having a deeper dive, I realised that the order in which child contexts are cancelled is non-deterministic (due to map iteration). I had thought of iterating the map in an ordered fashion, but wasn't sure how we would want to order the cancelers, so instead went with a slice so we can keep order based on the order in which they appear in the slice.

[Replay file](https://github.com/user-attachments/files/22373001/41ee787c-f851-4fe0-aee4-ecbd6d20689b_events.json)
I've attached an event history from a simple reproduction of the issue I created with some modifications of the cancellation example in samples-go:
```
func YourWorkflow(ctx workflow.Context) error {
		StartToCloseTimeout: 30 * time.Minute,
		HeartbeatTimeout:    5 * time.Second,
		WaitForCancellation: true,
	}
	logger := workflow.GetLogger(ctx)
	logger.Info("cancel workflow started")
	var a *Activities // Used to call Activities by function pointer
	defer func() {

		if !errors.Is(ctx.Err(), workflow.ErrCanceled) {
			return
		}

		// When the Workflow is canceled, it has to get a new disconnected context to execute any Activities
		newCtx, _ := workflow.NewDisconnectedContext(ctx)
		err := workflow.ExecuteActivity(newCtx, a.CleanupActivity).Get(ctx, nil)
		if err != nil {
			logger.Error("CleanupActivity failed", "Error", err)
		}
	}()

	// My changes - BEGIN
	selector := workflow.NewSelector(ctx)
	timer1Ctx, _ := workflow.WithCancel(ctx)
	// Create two new timers
	selector.AddFuture(workflow.NewTimer(timer1Ctx, time.Hour), func(f workflow.Future) {
		logger.Info("Timer 1 completed")
	})
	timer2Ctx, _ := workflow.WithCancel(ctx)
	selector.AddFuture(workflow.NewTimer(timer2Ctx, 2*time.Hour), func(f workflow.Future) {
		logger.Info("Timer 2 completed")
	})
	selector.Select(ctx)
	// My changes - END

	err := workflow.ExecuteActivity(ctx, a.ActivityToBeSkipped).Get(ctx, nil)
	logger.Error("Error from ActivityToBeSkipped", "Error", err)

	logger.Info("Workflow Execution complete.")

	return nil
}
```
Running the replay with the following code on repeat will encounter failures, without the fix in this PR:
```
func TestReplay(t *testing.T) {
	replayOpts := worker.WorkflowReplayerOptions{
		EnableLoggingInReplay: true,
	}
	replayer, err := worker.NewWorkflowReplayerWithOptions(replayOpts)
	assert.NoError(t, err)
	replayer.RegisterWorkflow(YourWorkflow)

	f, err := os.Open("41ee787c-f851-4fe0-aee4-ecbd6d20689b_events.json")
	require.NoError(t, err, "failed to open file")
	defer f.Close()

	h, err := client.HistoryFromJSON(f, client.HistoryJSONOptions{})
	require.NoError(t, err, "failed to parse history from JSON")

	err = replayer.ReplayWorkflowHistory(nil, h)

	assert.NoError(t, err)
}
```

## Checklist

1. How was this tested:
Unit tests cover the changes, and I have ran the code over our intermittently failing replay tests and it resolves the failures.

2. Any docs updates needed?
I don't believe so? This is likely not documented, since it's internal code

### DIFF
--- a/internal/context.go
+++ b/internal/context.go
@@ -218,10 +218,15 @@ func propagateCancel(parent Context, child canceler) {
 			child.cancel(false, parentErr)
 		} else {
 			p.childrenLock.Lock()
-			if p.children == nil {
-				p.children = make(map[canceler]bool)
+			childExists := false
+			for _, c := range p.children {
+				if c == child {
+					childExists = true
+				}
 			}
-			p.children[child] = true
+			if !childExists {
+				p.children = append(p.children, child)
+			}
 			p.childrenLock.Unlock()
 		}
 	} else {
@@ -253,7 +258,16 @@ func removeChild(parent Context, child canceler) {
 	}
 	p.childrenLock.Lock()
 	if p.children != nil {
-		delete(p.children, child)
+		childIndex := -1
+		for i, c := range p.children {
+			if c == child {
+				childIndex = i
+				break
+			}
+		}
+		if childIndex >= 0 {
+			p.children = append(p.children[:childIndex], p.children[childIndex+1:]...)
+		}
 	}
 	p.childrenLock.Unlock()
 }
@@ -272,7 +286,7 @@ type cancelCtx struct {
 
 	done Channel // closed by the first cancel call.
 
-	children     map[canceler]bool // set to nil by the first cancel call
+	children     []canceler // set to nil by the first cancel call
 	childrenLock sync.Mutex
 	err          error // set to non-nil by the first cancel call
 	errLock      sync.RWMutex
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns:
This is a know [issue](https://github.com/temporalio/sdk-go/issues/1208) this needs to be done very carefully to preserver backwards compatibility with existing workflows which the current change does not due. Given the sensitive of this change this is probably something best left to the Temporal team to fix

PR #2056, opened 2025-09-17
Activity reset

### LINKED ISSUE #1931: [go] SDK support for activity reset
See https://github.com/temporalio/features/issues/620

THardy98:
## What was changed
Add support for activity reset

## Why?
Allow users to reset their activity

1. Closes #1931  

2. How was this tested:
Integration test

3. Any docs updates needed?
Maybe?

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -588,7 +588,7 @@ func (ts *IntegrationTestSuite) TestActivityPause() {
 	// ActivityToBePaused activity twice, the first call will test pausing an activity successfully
 	// and the second call will test completing the activity after it is resumed
 	run, err := ts.client.ExecuteWorkflow(ctx,
-		ts.startWorkflowOptions("test-activity-pause"), ts.workflows.ActivityHeartbeat)
+		ts.startWorkflowOptions("test-activity-pause"), ts.workflows.ActivityHeartbeatPause)
 	ts.NoError(err)
 	// Wait for the workflow to finish
 	var result string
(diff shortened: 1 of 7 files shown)


test/integration_test.go
```diff
@@ -619,6 +619,26 @@ func (ts *IntegrationTestSuite) TestActivityPause() {
...
+	// ActivityToBeReset activity twice, the first call will test resetting an activity successfully
+	// and the second call will test completing the activity after it has been reset
+	run, err := ts.client.ExecuteWorkflow(ctx,
+		ts.startWorkflowOptions("test-activity-reset"), ts.workflows.ActivityHeartbeatReset)
+	ts.NoError(err)
+	// Wait for the workflow to finish
+	var result string
+	err = run.Get(ctx, &result)
+	ts.NoError(err)
+	// Check the result - ensure ErrActivityReset was thrown, heartbeat details were reset, and num attempts were reset
```
Quinn-With-Two-Ns: Does this actually check the details were reset? Maybe I am misreading the test, but it looks like the activity only runs once no?

THardy98: It should, activity reset/pause is propagated via heartbeat response. I think running the activity once should be enough to do:

`Reset` -> `Heartbeat` (propagate the reset response, transform + emit an activity task cancellation) -> `Receive cancellation in activity (i.e. ErrActivityReset)`

Which is enough to check the reset flow, unless I'm missing something

Quinn-With-Two-Ns: How does that confirm the heartbeat details were reset, and num attempts were reset? you would need another attempt to verify that no?

THardy98: Thought heartbeat details were updated via heartbeat responses, but they are only updated through subsequent activity tasks (like all other activity info). Updated test.

PR #2057, opened 2025-09-22
Added retry policy to activity info

### LINKED ISSUE #1919: Show retry policy on activity info

See https://github.com/temporalio/features/issues/615

maciejdudko:
## What was changed

Added retry policy to activity info.

## Why?

Feature request: https://github.com/temporalio/features/issues/615

## Checklist

1. Closes #1919

2. How was this tested: Added new check to `InspectActivityInfo` tests.

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -46,7 +46,7 @@ type (
 		// WARNING: Task queue priority is currently experimental.
 		Priority Priority
 		// Retry policy for the activity. Note that the server may have set a different policy than the one provided
-		// when scheduling the activity. If the value is None, it means the server didn't send information about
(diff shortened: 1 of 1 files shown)


internal/activity.go
```diff
@@ -45,6 +45,10 @@ type (
 		//
 		// WARNING: Task queue priority is currently experimental.
 		Priority Priority
+		// Retry policy for the activity. Note that the server may have set a different policy than the one provided
+		// when scheduling the activity. If the value is None, it means the server didn't send information about
```
Quinn-With-Two-Ns:
```suggestion
		// when scheduling the activity. If the value is nil, it means the server didn't send information about
```

Generally Go docs use `nil` not `None`

PR #2059, opened 2025-09-26
`ResetWorkflow` workflow panic policy

elb3k:
## What was changed
Added a new panic policy called `ResetWorkflow`, if this option is set, the workflow will be reset from the beginning.

## Why?
New panic policy to reset the workflow from the start in case of panic/non-determinism is needed for our application. Full discussion [here](https://community.temporal.io/t/catching-nondeterministicexception/5020/10?u=_elbek_khoshimjonov).

## Checklist

1. Closes https://github.com/temporalio/features/issues/417 (at least for go-sdk)

2. How was this tested: [unit test here](https://github.com/elb3k/temporal-sdk-go/commit/658ecb5d3f55c8ef21069e34b43fd38c855d8671), and using this sample workflow to test:
```golang
func Workflow(ctx workflow.Context, params string) error {
	logger := workflow.GetLogger(ctx)
	logger.Info("Starting workflow... ")

	var a *Activities
	}

	retryCtx := workflow.WithRetryPolicy(ctx, temporal.RetryPolicy{
		MaximumAttempts: 1,
	})

	workflow.Sleep(ctx, 1*time.Second)
	var ret int
	if err := workflow.ExecuteActivity(retryCtx, a.Rulet).Get(retryCtx, &ret); err != nil {
		return err
	}
	if ret == 3 {
		return nil
	}
	panic("error")
}

type Activities struct{}

func (*Activities) Rulet(ctx context.Context) (int, error) {
	return rand.Intn(7), nil
}
```

4. Any docs updates needed?
`ResetWorkflow` option needs to be added alongside `BlockWorkflow` and `FailWorkflow`

5. Caveats - this is a rough and simple implementation and has a few caveats:
- Any panic in the workflow will result in a restart, and if the workflow code does not handle the issue, there might be an infinite loop. (Some kind of maximum limit should be set.)

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -133,6 +133,7 @@ type (
 		cache                     *WorkerCache
 		deadlockDetectionTimeout  time.Duration
 		capabilities              *workflowservice.GetSystemInfoResponse_Capabilities
+		client                    *WorkflowClient // Added for ResetWorkflowExecution support
 	}
 
 	activityProvider func(name string) activity
@@ -546,7 +547,7 @@ func inferMessageFromAcceptedEvent(attrs *historypb.WorkflowExecutionUpdateAccep
 }
 
 // newWorkflowTaskHandler returns an implementation of workflow task handler.
-func newWorkflowTaskHandler(params workerExecutionParameters, ppMgr pressurePointMgr, registry *registry) WorkflowTaskHandler {
+func newWorkflowTaskHandler(params workerExecutionParameters, ppMgr pressurePointMgr, registry *registry, client *WorkflowClient) WorkflowTaskHandler {
 	ensureRequiredParams(&params)
 	return &workflowTaskHandlerImpl{
 		namespace:                 params.Namespace,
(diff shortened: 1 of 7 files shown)


Quinn-With-Two-Ns:
Hi @elb3k thank you for the contribution but this is not the correct approach to implement this. This needs sever side support to add a command for the SDK to request a workflow be reset after a workflow task. That is how all SDK workflow completions work like Failed, Continue as new, Success etc work. In the future please open an issue first to understand how something should be implemented and if there is interest.

You can make an issue on the [Temporal Server ](https://github.com/temporalio/temporal) repo to discuss if this is a feature the server is interested in supporting and how to go about adding it to the Temporal server

elb3k:
> Hi @elb3k thank you for the contribution but this is not the correct approach to implement this. This needs sever side support to add a command for the SDK to request a workflow be reset after a workflow task. That is how all SDK workflow completions work like Failed, Continue as new, Success etc work. In the future please open an issue first to understand how something should be implemented and if there is interest.
> 
> You can make an issue on the [Temporal Server ](https://github.com/temporalio/temporal) repo to discuss if this is a feature the server is interested in supporting and how to go about adding it to the Temporal server

Thanks for the feedback @Quinn-With-Two-Ns. I understand this issue is not Go SDK exclusive. The issue was created last year here: https://github.com/temporalio/features/issues/417
We wanted this feature for our projects pretty soon, and this was my attempt at solving the issue. How can we implement this in [temporal-server](https://github.com/temporalio/temporal)? I think some of the concerns you had about child workflows are mitigated (since [this](https://github.com/temporalio/temporal/issues/3437) is closed).

PR #2060, opened 2025-09-29
fix: inconsistent interceptor span emission when workflow is in replay [WIP]

sourcec0de:
## What was changed
This PR addresses an issue that causes `RunWorkflow:*` spans to be emitted when workflows are viewed in the UI (i.e in replay). To make the behavior consistent, `startNonReplaySpan` was extracted from `tracingWorkflowOutboundInterceptor` and simplified to check the workflow context and return a no-op span when the workflow is in replay mode. Then `startNonReplaySpan` is reused in all places where a workflow context is in scope.

## Why?
[Discern](https://discern.com/) uses the `RunWorkflow:*` otel span to alert on complete workflow failures (i.e a workflow that has died indefinitely). When viewing the temporal UI in the cloud console, our monitoring is re-triggered because the query to the worker replays the workflow history and re-emits the `RunWorkflow` span.

[Historical context in slack](https://temporalio.slack.com/archives/CTDTU3J4T/p1736795189388109)

### How was this tested:
I am still in the process of testing this using go workspaces by aliasing my branch of `temporalio/sdk-go` in our Go environment.

### DIFF
--- a/interceptor/tracing_interceptor.go
+++ b/interceptor/tracing_interceptor.go
@@ -63,7 +63,7 @@ type Tracer interface {
 // BaseTracer is a default implementation of Tracer meant for embedding.
 type BaseTracer struct{}
 
-func (BaseTracer) GetLogger(logger log.Logger, ref TracerSpanRef) log.Logger {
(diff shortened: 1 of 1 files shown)


sourcec0de:
cc: @cretz

PR #2073, opened 2025-10-09
Add WorkflowType tag to activity log

wilianto:
## What was changed
Add WorkflowType tag to all activity logs

## Why?
We have an activity that’s reused across multiple workflows. This tag will help us debug production issues more efficiently by allowing filtering based on workflow type.

## Checklist

1. Closes 

2. How was this tested:

Manually by pointing the SDK to local build.

Sample log:
```
6:33PM ERR Activity error. ActivityType=CreateJournalEntry Attempt=10429 Error="invalid_argument: validation failed: amount: value 0 is less than minimum of 1." Namespace=default RunID=d063adf7-bfc5-4803-b70c-11153049699b TaskQueue=journal WorkerID=35819@Wiliantos-MacBook-Pro.local@ WorkflowID=chargefeev2_fee_bjyopigj5n1ga8n4hleazzes56 WorkflowType=ChargeFeeWorkflow
```

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -2229,6 +2229,7 @@ func newServiceInvoker(
 func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice.PollActivityTaskQueueResponse) (result interface{}, err error) {
 	traceLog(func() {
 		ath.logger.Debug("Processing new activity task",
+			tagWorkflowType, t.WorkflowType.GetName(),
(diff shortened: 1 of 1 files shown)


yuandrew:
We don't want to add this key/label pair to activity logs for ALL Go SDK users. Instead, you can use the `activityOutboundInterceptor` to add extra tags, see the SDK sample: https://github.com/temporalio/samples-go/blob/6c38ecf1c880e19906e17eacb1ebb069a3ee2108/logger-interceptor/interceptor.go#L50

PR #2075, opened 2025-10-09
Set worker version on eager start requests

Sushisource:
## What was changed
Add worker deployment options to eager workflow start requests

## Why?
Needed so server has versioning info available to it when processing eager start requests

## Checklist

1. Closes 

2. How was this tested:
Unit tests

3. Any docs updates needed?

### DIFF
--- a/internal/internal_eager_workflow_test.go
+++ b/internal/internal_eager_workflow_test.go
@@ -104,7 +104,6 @@ func TestEagerWorkflowDispatchWithDeploymentOptions(t *testing.T) {
 	require.Equal(t, workerWithDeployment, exec.worker)
 	require.True(t, request.GetRequestEagerExecution())
 
-	// Verify that deployment options were set correctly
 	require.NotNil(t, request.EagerWorkerDeploymentOptions)
 	ewdo := request.EagerWorkerDeploymentOptions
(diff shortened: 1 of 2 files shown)


internal/internal_eager_workflow_test.go
```diff
@@ -68,6 +74,77 @@ func TestEagerWorkflowDispatchAvailableWorker(t *testing.T) {
...
+	dispatcher := &eagerWorkflowDispatcher{
+		workersByTaskQueue: make(map[string]map[eagerWorker]struct{}),
+	}
+
+	// Worker without deployment versioning enabled
+	workerWithoutVersioning := &eagerWorkerMock{
+		tryReserveSlotCallback: func() *SlotPermit { return &SlotPermit{} },
+		deploymentOptions: WorkerDeploymentOptions{
+			UseVersioning: false,
+			Version: WorkerDeploymentVersion{
```
yuandrew, line 124: For my own learning, when would this scenario happen, where `WorkerDeploymentVersion` is specified but `UseVersioning` is turned off?

Sushisource: You just want the build id to show up in history.

PR #2081, opened 2025-10-17
Remove Nexus deprecated APIs

bergundy:
Left the compatibility with older servers for now that may use `OperationID`.

### DIFF
--- a/temporalnexus/operation.go
+++ b/temporalnexus/operation.go
@@ -321,6 +321,7 @@ func ExecuteUntypedWorkflow[R any](
 			nexusOptions.CallbackHeader = make(nexus.Header)
 		}
 
+		// This field is expected to be populated by servers older than 1.27.0.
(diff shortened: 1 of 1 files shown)


temporalnexus/operation.go
```diff
@@ -323,8 +321,7 @@ func ExecuteUntypedWorkflow[R any](
 			nexusOptions.CallbackHeader = make(nexus.Header)
 		}
 
-		//lint:ignore SA1019 this field is expected to be populated by servers older than 1.27.0.
-		nexusOptions.CallbackHeader.Set(nexus.HeaderOperationID, encodedToken)
+		nexusOptions.CallbackHeader.Set("nexus-operation-id", encodedToken)
```
bergundy, line 325:
```suggestion
		// This field is expected to be populated by servers older than 1.27.0.
		nexusOptions.CallbackHeader.Set("nexus-operation-id", encodedToken)
```

yuandrew (approved):

PR #2087, opened 2025-10-21
Add nightly throughput stress

THardy98:
## What was changed
Added a nightly workflow to run Omes throughput stress scenario.

Runs throughput stress with a minimum throughput expectation.

Uploads worker logs artifacts and sends a Slack notification on failure.

## Why?
Provides basic smoke/stress testing, with a base benchmark throughput expectation.

### DIFF
--- a/.github/workflows/nightly-throughput-stress.yml
+++ b/.github/workflows/nightly-throughput-stress.yml
@@ -15,7 +15,7 @@ on:
         default: '6h'
         type: string
       timeout:
-        description: 'Scenario timeout (should always be 30m more than duration)'
+        description: 'Scenario timeout (should be greater than duration)'
         required: false
         default: '6h30m'
         type: string


.github/workflows/nightly-throughput-stress.yml
```diff
@@ -0,0 +1,156 @@
...
+  TEST_DURATION: ${{ inputs.duration || vars.NIGHTLY_TEST_DURATION || '6h' }}
+  TEST_TIMEOUT: ${{ inputs.timeout || vars.NIGHTLY_TEST_TIMEOUT || '6h30m' }}
+
+  # Logging and artifacts
+  WORKER_LOG_DIR: /tmp/throughput-stress-logs
+
+  # Omes configuration
+  OMES_REPO: temporalio/omes
+  OMES_REF: main
+  RUN_ID: ${{ github.run_id }}-throughput-stress
```
Quinn-With-Two-Ns, line 39: May want to amend the SDK to the run ID, but if we do we should do for all SDKs

Quinn-With-Two-Ns (approved):

PR #2088, opened 2025-10-22
Stop sending deployment info for session per-worker task queue

Quinn-With-Two-Ns:
Stop sending deployment info for session per-worker task queue

---

> [!NOTE]
> Temporarily disables deployment/versioning for the session activity worker on the resource-specific task queue (restored for creation worker) and adds a session build ID test.
> 
> - **Worker/session handling**:
>   - In `newSessionWorker`, save `params.DeploymentOptions`/`params.UseBuildIDForVersioning`, set `DeploymentOptions.UseVersioning=false` and `UseBuildIDForVersioning=false` for the session activity worker on the resource-specific task queue, then restore for the creation worker.
>   - Minor refactor of `workerOverrides` construction.
> - **Tests**:
>   - Add `TestBuildIDWithSession` in `test/worker_deployment_test.go` to validate running a session-enabled worker with deployment versioning; registers session workflow and activity using `activity.RegisterOptions` and verifies workflow completion.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6ef51b01fb675281d585c8f83c89f155f99ea252. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -452,7 +452,8 @@ func newSessionWorker(client *WorkflowClient, params workerExecutionParameters, 
 	// Save them to restore later
 	deployments := params.DeploymentOptions
 	useBuildIDForVersioning := params.UseBuildIDForVersioning
-	params.DeploymentOptions = WorkerDeploymentOptions{}
+	// Disable versioning for activity worker within session, but still send deployment name for debug purpose
+	params.DeploymentOptions.UseVersioning = false
 	params.UseBuildIDForVersioning = false
 	activityWorker := newActivityWorker(client, params,
 		&workerOverrides{


test/worker_deployment_test.go
```diff
@@ -236,6 +236,57 @@ func (ts *WorkerDeploymentTestSuite) TestBuildIDChangesOverWorkflowLifetime() {
 	ts.Equal("2.0", lastBuildID)
 }
 
+func (ts *WorkerDeploymentTestSuite) TestBuildIDSession() {
+	ts.T().Skip("issue-1650: Build ID integration tests are flaky")
```
Sushisource: Seems like this new guy shouldn't be skipped, no?

internal/internal_worker.go
```diff
@@ -448,15 +448,22 @@ func newSessionWorker(client *WorkflowClient, params workerExecutionParameters,
 	creationTaskqueue := getCreationTaskqueue(params.TaskQueue)
 	params.BackgroundContext = context.WithValue(params.BackgroundContext, sessionEnvironmentContextKey, sessionEnvironment)
 	params.TaskQueue = sessionEnvironment.GetResourceSpecificTaskqueue()
+	// For the resource specific task queue, we don't need to include deployment options
+	// Save them to restore later
+	deployments := params.DeploymentOptions
+	params.DeploymentOptions = WorkerDeploymentOptions{}
```
ShahabT: you need to disable v1-2 versioning too because server still supports that.
```
	params.UseBuildIDForVersioning = false
```

internal/internal_worker.go
```diff
@@ -448,15 +448,25 @@ func newSessionWorker(client *WorkflowClient, params workerExecutionParameters,
 	creationTaskqueue := getCreationTaskqueue(params.TaskQueue)
 	params.BackgroundContext = context.WithValue(params.BackgroundContext, sessionEnvironmentContextKey, sessionEnvironment)
 	params.TaskQueue = sessionEnvironment.GetResourceSpecificTaskqueue()
+	// For the resource specific task queue, we don't need to include deployment options
+	// Save them to restore later
+	deployments := params.DeploymentOptions
```
ShahabT, line 453: Only disable versioning instead of removing the whole thing? better to keep the deployment name and build id if user has given it (to keep in the events).

Quinn-With-Two-Ns: Does it make any difference for an activity only task queue? I don't see any difference in the history.

ShahabT: we might add the deployment options in activity events later

test/worker_deployment_test.go
```diff
@@ -236,6 +237,60 @@ func (ts *WorkerDeploymentTestSuite) TestBuildIDChangesOverWorkflowLifetime() {
 	ts.Equal("2.0", lastBuildID)
 }
 
+func (ts *WorkerDeploymentTestSuite) TestBuildIDWithSession() {
+
```
yuandrew, line 241: Do we want to gate this? All other tests here have this req

```
if os.Getenv("DISABLE_SERVER_1_27_TESTS") != "" {
		ts.T().Skip("temporal server 1.27+ required")
	}
```

test/worker_deployment_test.go
```diff
@@ -236,6 +237,60 @@ func (ts *WorkerDeploymentTestSuite) TestBuildIDChangesOverWorkflowLifetime() {
...
+		BuildID:       v1.BuildID,
+		ConflictToken: response1.ConflictToken,
+	})
+	ts.NoError(err)
+
+	// start workflow1 with 1.0, BasicSession, auto-upgrade
+	wfHandle, err := ts.client.ExecuteWorkflow(ctx, ts.startWorkflowOptions("evolving-wf-1"), "SessionBuildIDWorkflow")
+	ts.NoError(err)
+
+	ts.NoError(wfHandle.Get(ctx, nil))
```
yuandrew, line 291: What is this actually verifying? This seems to pass on `master` without your fix

yuandrew: Should we be asserting that buildId and versioning settings for the session activity worker is properly set?

Quinn-With-Two-Ns: Yeah I am verifying it still works and I didn't break anything.

PR #2105, opened 2025-11-07
fix: run maxNumberOfPollers poller routines instead of initialNumberOfPollers

jjaegerman:
## What was changed
Autoscaling Poller now spawns `maxNumberOfPollers` separate `runPoller` routines instead of previously only `initialNumberOfPollers`. 

## Why?
The `runPoller` ([impl](https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker_base.go#L423), [started here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker_base.go#L388)) routines are the main unit of concurrent polling execution.

They acquire (and block on) a semaphore ([here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker_base.go#L435)) to limit how many can execute a poll ([here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker_base.go#L486)) at a given moment in time

`pollScalerReportHandle` is responsible for deciding on ([here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker_base.go#L738)) and triggering autoscaling of pollers by managing the capacity of this semaphore ([here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker_base.go#L783))

But since the number of `runPoller` routines is fixed and equal to the autoscaling option's `initialNumberOfPollers` ([here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_worker_base.go#L888)), regardless of the capacity on the semaphore, only a maximum `initialNumberOfPollers` will be able to run a poll at a given moment.

Spawning `maxNumberOfPollers` separate `runPoller` routines instead allows the autoscaling to run up to `maxNumberOfPollers` concurrent polls.

1. Closes 

[2106](https://github.com/temporalio/sdk-go/issues/2106)

3. How was this tested:

Reran the experiment described in the issue and observed a significant increase in `temporal_num_pollers` (all the way up to the default maximum of 100) and significant decrease in average workflow task schedule to start latency (from 30s to 3s).

Also added unit tests that isolate the `scalableTaskPoller`'s concurrency logic with a stubbed out `taskPoller` that blocks until signaled.

4. Any docs updates needed?

No

### DIFF
--- a/internal/internal_worker_base_test.go
+++ b/internal/internal_worker_base_test.go
@@ -1,12 +1,18 @@
 package internal
 
 import (
+	"context"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/sdk/internal/common/metrics"
+	ilog "go.temporal.io/sdk/internal/log"
 )
 
 type (
@@ -17,6 +23,16 @@ type (
 
 func TestPollScalerReportHandleSuite(t *testing.T) {
 	suite.Run(t, new(PollScalerReportHandleSuite))
+}
+
+type (
+	ScalableTaskPollerSuite struct {
+		suite.Suite
+	}
+)
+
+func TestScalableTaskPollerSuite(t *testing.T) {
+	suite.Run(t, new(ScalableTaskPollerSuite))
 }
 
 type testTask struct {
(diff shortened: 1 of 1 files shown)


internal/internal_worker_base.go
```diff
@@ -885,7 +885,7 @@ func newScalableTaskPoller(
 	}
 	switch p := pollerBehavior.(type) {
 	case *pollerBehaviorAutoscaling:
-		tw.pollerCount = p.initialNumberOfPollers
+		tw.pollerCount = p.maximumNumberOfPollers
```
Quinn-With-Two-Ns, line 888: If this is a fix for a bug can we add a test to verify the issue and that this actually resolves the issue?

jjaegerman: Added.

PR #2110, opened 2025-11-17
chore: Remove redundant activityInterceptorContextKey assignment

rerorero:
## What was changed

Removes redundant activityInterceptorContextKey assignment in newActivityContext.

## Why?

Noticed that it's redundant with line 437.

## Checklist

1. Closes 

2. How was this tested:

go run . unit-test
(TestIsActivity)

3. Any docs updates needed?

No

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -424,7 +424,6 @@ func newActivityContext(
 	envInterceptor.inboundInterceptor = envInterceptor
 	envInterceptor.outboundInterceptor = envInterceptor
 	ctx = context.WithValue(ctx, activityEnvInterceptorContextKey, envInterceptor)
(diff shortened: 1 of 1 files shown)


internal/activity.go
```diff
@@ -424,7 +424,6 @@ func newActivityContext(
 	envInterceptor.inboundInterceptor = envInterceptor
 	envInterceptor.outboundInterceptor = envInterceptor
 	ctx = context.WithValue(ctx, activityEnvInterceptorContextKey, envInterceptor)
-	ctx = context.WithValue(ctx, activityInterceptorContextKey, envInterceptor.outboundInterceptor)
```
Quinn-With-Two-Ns, line 427: This line should not be removed, the context past this point is expected to have this value set.

PR #2112, opened 2025-11-17
Defer updating workflow completion metrics until completion accepted by server

### LINKED ISSUE #2111: workflow_completed and workflow_continue_as_new counters should be logged after a successful reporting to the server

See https://github.com/temporalio/features/issues/222

cretz:
## What was changed

* Built new `workflowTaskCompletion` struct to represent workflow task completion instead of `interface{}`
* Added optional `applyCompletionMetrics` func to be called when the task completion is accepted
* Set `applyCompletionMetrics` with the workflow completed metrics to be applied on task completion accepted

## Checklist

1. Closes #2111

### DIFF
--- a/internal/cmd/build/main.go
+++ b/internal/cmd/build/main.go
@@ -172,11 +172,11 @@ func (b *builder) integrationTest() error {
 	if *coverageFileFlag != "" {
 		args = append(args, "-coverprofile="+filepath.Join(b.rootDir, coverageDir, *coverageFileFlag), "-coverpkg=./...")
 	}
+	args = append(args, "./...")
 	if *devServerFlag {
-		args = append(args, "-using-cli-dev-server")
+		args = append(args, "--", "-using-cli-dev-server")
 		env = append(env, "TEMPORAL_NAMESPACE=integration-test-namespace")
 	}
-	args = append(args, "./...")
 	// Must run in test dir
 	cmd := b.cmdFromRoot(args...)
 	cmd.Dir = filepath.Join(cmd.Dir, "test")
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1724,6 +1724,7 @@ func (aw *WorkflowReplayer) replayWorkflowHistory(logger log.Logger, service wor
 		return nil
 	}
 
+	var rawRequest proto.Message
 	if resp != nil {
 		completeReq, ok := resp.rawRequest.(*workflowservice.RespondWorkflowTaskCompletedRequest)
 		if ok {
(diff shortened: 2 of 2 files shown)


internal/internal_public.go
```diff
@@ -32,8 +32,8 @@ type (
 	WorkflowExecutionContext interface {
```
cretz, line 32: This is in `internal_public` but I don't see it referenced in `internalbindings` or having any side effects of my changing this interface (some methods already accepted parameters that couldn't be instantiated externally anyways).

@Quinn-With-Two-Ns - any concerns with the altering of this interface?

Quinn-With-Two-Ns: Is this interface even used? My IDE can't find a us of it?

internal/internal_task_handlers.go
```diff
@@ -1847,27 +1853,25 @@ func (wth *workflowTaskHandlerImpl) completeWorkflow(
...
 	// complete workflow task
 	var closeCommand *commandpb.Command
 	var canceledErr *CanceledError
 	var contErr *ContinueAsNewError
+	var metricCounterToIncrement string
 
 	if errors.As(workflowContext.err, &canceledErr) {
 		// Workflow canceled
-		metricsHandler.Counter(metrics.WorkflowCanceledCounter).Inc(1)
+		metricCounterToIncrement = metrics.WorkflowCanceledCounter
```
Quinn-With-Two-Ns, line 1867: Do you check if the existing test coverage for these completion metrics is sufficient to catch any potential bugs here? (I can check if you haven't just let me know)

internal/internal_public.go
cretz: Same, I doubt it is

internal/internal_task_handlers.go
cretz: I did not check, only confirmed existing tests pass and added my check w/ unhandled command that wouldn't have passed before. Doing a quick glance, I don't think any of these completion metrics have any coverage. Do you feel that new tests covering these existing metrics should be in this PR? I don't mind doing it if I must, will just add some time.

Quinn-With-Two-Ns: I think we should have test coverage to make sure these metrics are still being emitted correctly. Unfortunately we lack test coverage of some basic metrics in the Go SDK so we have to keep an eye out. I don't think it adds a lot of complexity to your test to iterate through the few other ways a workflow can complete.

cretz: Will do as part of this PR (may not be right away though)

internal/internal_task_pollers.go
```diff
@@ -577,7 +577,11 @@ func (wtp *workflowTaskProcessor) sendTaskCompletedRequest(
 			metrics.NoneTagValue, metrics.NoneTagValue))),
 		defaultGrpcRetryParameters(ctx))
 	defer cancel()
-	switch request := completedRequest.(type) {
+	if taskCompletion == nil {
+		// should not happen
+		panic("unknown request type from ProcessWorkflowTask()")
```
yuandrew, line 582: Why do we panic here instead of logging an error? Seems like Core's `dbg_panic` would be nice here, where we can panic for tests but only log an error when running an real worker

internal/internal_public.go
yuandrew: +1 don't see it being used anywhere

internal/internal_task_pollers.go
cretz: > Why do we panic here instead of logging an error?

Because we always did, see the `default` case of the type switch below, I just also put it up here to check the situation where `taskCompletion` could be nil. I have no opinion on whether we _should_ panic though.

internal/internal_task_handlers.go
cretz: Added test

PR #2113, opened 2025-11-18
Upgrade Datadog tracer to dd-trace-go v2

yarinzirlin:
- Updated imports from v1 (gopkg.in/DataDog/dd-trace-go.v1) to v2 (github.com/DataDog/dd-trace-go/v2)
- Updated struct types to use pointers (*tracer.Span, *tracer.SpanContext) as required by v2
- Replaced deprecated ChildOf() pattern with modern v2 approach:
  - Use Span.StartChild() for direct parent spans
  - Use tracer.WithStartSpanConfig() for span contexts
- Updated tests to handle v2 API changes (pointer types for mocktracer.Span)
- Removed unused helper functions (newSpanContextReader, spanContextReader)
- All tests passing

This upgrade ensures compatibility with the latest Datadog APM features and removes dependency on deprecated v1 APIs.

## What was changed
Updated dd-trace-go from the deprecated v1 to v2

## Why?
1. dd-trace-go v1 stopped working

## Checklist

1. How was this tested:
I ran the upgraded interceptor against my Datadog instance and saw the expected traces

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -68,6 +68,16 @@ type (
 		callback         ResultHandler
 		activityType     string
 		heartbeatDetails *commonpb.Payloads
+		// Timeout configuration
+		heartbeatTimeout    time.Duration
+		startToCloseTimeout time.Duration
+		// Timeout tracking
+		startTime         time.Time // when activity started executing
+		lastHeartbeatTime time.Time
+		// Timeout result (set by monitoring goroutine)
+		timedOut           bool
+		timeoutType        enumspb.TimeoutType // which timeout occurred
+		cancelTimeoutWatch func()              // cancels the timeout monitoring goroutine
 	}
 
 	testWorkflowHandle struct {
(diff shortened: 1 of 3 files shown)


contrib/datadog/tracing/interceptor.go
```diff
@@ -6,11 +6,9 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
-	"strconv"
 	"strings"
 
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
```
jlegrone, line 13: There are two breaking changes here. They won't impact all users, but:

- Anyone specifying a `TracerOptions.OnFinish` function will need to modify their imports from `gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer` to `github.com/DataDog/dd-trace-go/v2/ddtrace/tracer`. The [old](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.74.8/ddtrace#FinishOption) and [new](https://pkg.go.dev/github.com/DataDog/dd-trace-go/v2/ddtrace/tracer#FinishOption) versions of this type are source compatible however, so no other changes are needed.
- Anyone using the `SpanFromWorkflowContext` function will need to modify their code to reference the new type if they pass the returned span into a function accepting the old `ddtrace.Span` type, or if they were calling the `Context()` method.

Breaking changes are technically ok to do without releasing a v2 of this module, since we are still on a pre-stable version (`v0.4.0`). I think it is acceptable to release this as `v0.5.0` as long as we document these breaking changes in the release notes. If necessary, we should still be able to backport bugfixes by tagging a `v0.4.1` version.

@yarinzirlin would you like to take a stab at writing the release notes for this version?

PR #2118, opened 2025-11-21
Add SpanFromWorkflowContext function for OTel

jessefoster:
## What was changed
Added a function which can extract Open Telemetry Span from Workflow Context.

## Why?
I saw Datadog had this capability already, see https://github.com/temporalio/sdk-go/pull/1711, and my project needs access to Open Telemetry spans.

## Checklist

1. How was this tested:
I've confirmed the feature works within my codebase and added unit tests as well.  I've run all automated testing as prescribed [here](https://github.com/temporalio/sdk-go/blob/master/CONTRIBUTING.md)

2. Any doc updates needed?
None that I'm aware of

### DIFF
--- a/contrib/opentelemetry/tracing_interceptor.go
+++ b/contrib/opentelemetry/tracing_interceptor.go
@@ -15,6 +15,7 @@ import (
 	"go.temporal.io/sdk/interceptor"
 	"go.temporal.io/sdk/log"
 	"go.temporal.io/sdk/temporal"
+	"go.temporal.io/sdk/workflow"
 )
 
 // DefaultTextMapPropagator is the default OpenTelemetry TextMapPropagator used
@@ -172,6 +173,21 @@ func (t *tracer) ContextWithSpan(ctx context.Context, span interceptor.TracerSpa
 		ctx = baggage.ContextWithBaggage(ctx, span.(*tracerSpan).Baggage)
 	}
 	return trace.ContextWithSpan(ctx, span.(*tracerSpan).Span)
+}
+
+// SpanFromWorkflowContext extracts an OpenTelemetry span from the given
+// workflow context.  If no span is found, a no-op span is returned.
+func SpanFromWorkflowContext(ctx workflow.Context) (trace.Span, bool) {
+	val := ctx.Value(spanContextKey{})
+
+	if val != nil {
+		if span, ok := val.(*tracerSpan); ok {
+			return span.Span, true
+		}
+	}
+
+	// Fallback to OpenTelemetry span extraction behavior
+	return trace.SpanFromContext(nil), false
 }
 
 func (t *tracer) StartSpan(opts *interceptor.TracerStartSpanOptions) (interceptor.TracerSpan, error) {
--- a/contrib/opentelemetry/tracing_interceptor_test.go
+++ b/contrib/opentelemetry/tracing_interceptor_test.go
@@ -1,11 +1,14 @@
 package opentelemetry_test
 
 import (
+	"context"
+	"errors"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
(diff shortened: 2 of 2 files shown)


contrib/opentelemetry/tracing_interceptor.go
```diff
@@ -174,6 +175,19 @@ func (t *tracer) ContextWithSpan(ctx context.Context, span interceptor.TracerSpa
 	return trace.ContextWithSpan(ctx, span.(*tracerSpan).Span)
 }
 
+func SpanFromWorkflowContext(ctx workflow.Context) (trace.Span, bool) {
```
Quinn-With-Two-Ns, line 180: Please add a godoc to this function explaining the functionality

jessefoster: Sure, godoc added

contrib/opentelemetry/tracing_interceptor.go
```diff
@@ -174,6 +175,21 @@ func (t *tracer) ContextWithSpan(ctx context.Context, span interceptor.TracerSpa
 	return trace.ContextWithSpan(ctx, span.(*tracerSpan).Span)
 }
 
+// SpanFromWorkflowContext extracts an OpenTelemetry span from the given
+// workflow context.  If no span is found, a no-op span is returned.
+func SpanFromWorkflowContext(ctx workflow.Context) (trace.Span, bool) {
```
cretz, line 180: Why return a bool that is always true?

cretz, line 180: May need to be clarified in the godoc that this is unsafe/non-deterministic because it can technically return different values when replaying vs not

cretz, line 180: I wonder if `SpanContextFromWorkflowContext` is needed/better

contrib/opentelemetry/tracing_interceptor.go
```diff
@@ -174,6 +175,21 @@ func (t *tracer) ContextWithSpan(ctx context.Context, span interceptor.TracerSpa
...
+	val := ctx.Value(spanContextKey{})
+
+	if val != nil {
+		if span, ok := val.(*tracerSpan); ok {
+			return span.Span, true
+		}
+	}
+
+	// Fallback to OpenTelemetry span extraction behavior
+	return trace.SpanFromContext(nil), true
```
Quinn-With-Two-Ns: This should return false

contrib/opentelemetry/tracing_interceptor.go
Quinn-With-Two-Ns: Yeah this look like a bug

contrib/opentelemetry/tracing_interceptor.go
jessefoster: Yeah sorry about that, copy/paste mistake as I had to recreate file and run tests on another box.  Not allowed to push directly from corporate laptop.

contrib/opentelemetry/tracing_interceptor.go
jessefoster: We haven't had any need to use SpanContext, but it is another option. The Datadog interceptor only provides [Span ](https://github.com/temporalio/sdk-go/blob/master/contrib/datadog/tracing/interceptor.go#L125) as well so I followed it's example since it's already in the codebase.

jessefoster: So you'd like a comment warning people they should only use the Span data for observability needs?  That's generally the assumption with observability tools, but I can make it explicit.

cretz: Yes. That assumption is reasonable in code where non-deterministic elements are fine _if_ they chose to use span data for other things, but for Temporal, it's usually better to clearly mark any non-deterministic workflow functions we offer as unsafe if they may be.

cretz: Yeah, this is probably good, as I can't think of a situation where you could get a span context but not a span. I will say that query and update handlers will get a new span on non-replay, or the "run workflow" span on replay since they don't create new spans on replay but outer "run workflow" does IIUC.

jessefoster:
Thank you for yals help with the PR.

PR #2121, opened 2025-12-01
Introduce SDK flag to use user DC on Memo's

### LINKED ISSUE #1045: Memo does not go through user provided data coverter
## Expected Behavior

Go to use the user provided data converter like other SDKs to encode memos when creating workflow and schedules.

## Actual Behavior
https://github.com/temporalio/sdk-go/blob/0b0234f751a218e3d093e7261922dc6f9748387e/internal/internal_workflow_client.go#L1310-L1311

Go uses the default data converter. 

Need to decide how we will fix this as changing is not backwards compatible and could break running workflows

yuandrew:
## What was changed
Introduced an SDK flag, `SDKFlagMemoUserDCEncode`, to configure new behavior, where we actually use the user data converter when encoding Memo's. If user DC errors out, the default DC will be used as a fallback. If that also returns an error, the user DC error will be returned to the caller.

SDK flag is still set to false by default, so no behavior changes. After a Go SDK release or two, will the flag then be enabled by default.

## Why?
Correct behavior to match other SDKs

Something interesting to note about this SDK flag, this is the first time an SDK flag is used with the client and not the workflow. `TryUse(SDKFlagMemoUserDCEncode)` is instances from the workflow, and references to `sdkFlagsAllowed[SDKFlagMemoUserDCEncode]` are when we're using the client and don't have access to workflow's SDKFlag settings.

## Checklist
1. Closes #1045 

2. How was this tested:
Added new tests that flip on/off the SDK flag and validates behavior

Added replay tests for old behavior without flag. (manually tested with the flag turned on, tests do nothing at the moment but will be useful when the flag is flipped in a few releases)
<details>

<summary>Code used to general history </summary>

```
func TestGenerateMemoWorkflowHistoryJSON(t *testing.T) {
	//t.Skip("Manual test: requires a running Temporal server")
	c, _ := client.Dial(client.Options{
		Logger: ilog.NewDefaultLogger(),
	})
	defer c.Close()

	w := worker.New(c, "replay-test-memo-json", worker.Options{})
	w.RegisterWorkflow(MemoChildWorkflowJSON)
	w.RegisterWorkflow(MemoEncodingWorkflowJSON)
	_ = w.Start()
	defer w.Stop()

	we, _ := c.ExecuteWorkflow(context.Background(), client.StartWorkflowOptions{
		ID:        "replay-tests-memo-json",
		TaskQueue: "replay-test-memo-json",
	}, MemoEncodingWorkflowJSON, "test-value")
	var res string
	err := we.Get(context.Background(), &res)
	require.NoError(t, err)
	t.Logf("Workflow completed with result: %s", res)
	t.Log("Now run:")
	t.Log("temporal workflow show --workflow-id replay-tests-memo-json --output json > test/replaytests/memo-json.json")
}


func TestGenerateScheduleMemoWorkflowHistoryJSON(t *testing.T) {
	//t.Skip("Manual test: requires a running Temporal server")
	c, _ := client.Dial(client.Options{
		Logger: ilog.NewDefaultLogger(),
	})
	defer c.Close()

	w := worker.New(c, "replay-test-schedule-memo-json", worker.Options{})
	w.RegisterWorkflow(ScheduleMemoWorkflowJSON)
	_ = w.Start()
	defer w.Stop()

	scheduleID := "replay-test-schedule-memo-json"
	scheduleHandle, err := c.ScheduleClient().Create(context.Background(), client.ScheduleOptions{
		ID: scheduleID,
		Spec: client.ScheduleSpec{
			CronExpressions: []string{"* * * * *"},
		},
		Action: &client.ScheduleWorkflowAction{
			ID:        "replay-test-schedule-memo-json-wf",
			Workflow:  ScheduleMemoWorkflowJSON,
			TaskQueue: "replay-test-schedule-memo-json",
			Memo: map[string]interface{}{
				"schedule-memo-key": "schedule-test-value",
			},
		},
		TriggerImmediately: true,
	})
	require.NoError(t, err)
	defer func() {
		_ = scheduleHandle.Delete(context.Background())
	}()

	ctx := context.Background()
	var desc *client.ScheduleDescription
	var workflowID string
	assert.Eventually(t, func() bool {
		desc, err = scheduleHandle.Describe(ctx)
		require.NoError(t, err)
		if len(desc.Info.RecentActions) == 0 {
			return false
		}
		workflowID = desc.Info.RecentActions[0].StartWorkflowResult.WorkflowID
		return true
	}, 10*time.Second, 200*time.Millisecond)

	t.Log("Wait for the workflow to complete, then run:")
	t.Logf("temporal workflow show --workflow-id %s --output json > test/replaytests/memo-schedule-json.json", workflowID)
}
```
</details>

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Touches memo encoding paths for workflow starts/children/schedules; behavior is gated by a new flag but incorrect flag handling could affect memo serialization compatibility and replay determinism.
> 
> **Overview**
> Introduces new SDK flag `SDKFlagMemoUserDCEncode` (default **off**) to optionally encode workflow `Memo` values using the *user-provided* `DataConverter`, with fallback to the default converter; if both fail, the user-converter error is surfaced.
> 
> Threads this flag through memo encoding for workflow-side operations (`UpsertMemo`, child workflow start) using `TryUse(...)` for replay safety, and client-side operations (workflow start, signal-with-start, schedules) using `sdkFlagsAllowed[...]`. Adds `encodeMemoValue` helper, updates unit tests to cover old/new behaviors and failure cases, and adds replay histories/tests to ensure older JSON-encoded memo workflows (including schedule-started) remain replayable when the flag is absent.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e4a76cab8ce675a6a7c043fc5cad3837e6442d97. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


internal/internal_flags.go
```diff
@@ -29,13 +29,20 @@ const (
...
 )
 
 // unblockSelectorSignal exists to allow us to configure the default behavior of
 // SDKFlagBlockedSelectorSignalReceive. This is primarily useful with tests.
 var unblockSelectorSignal = os.Getenv("UNBLOCK_SIGNAL_SELECTOR") != ""
 
+// memoUserDCEncode exists to allow us to configure the default behavior of
+// SDKFlagMemoUserDCEncode. This is primarily useful with tests.
+var memoUserDCEncode = os.Getenv("MEMO_USER_DC_ENCODE") != ""
+
```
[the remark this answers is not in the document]
yuandrew: Thanks cursor

internal/internal_flags.go
```diff
@@ -29,13 +29,20 @@ const (
...
+	SDKFlagUnknown          = math.MaxUint32
 )
 
 // unblockSelectorSignal exists to allow us to configure the default behavior of
 // SDKFlagBlockedSelectorSignalReceive. This is primarily useful with tests.
 var unblockSelectorSignal = os.Getenv("UNBLOCK_SIGNAL_SELECTOR") != ""
 
+// memoUserDCEncode exists to allow us to configure the default behavior of
+// SDKFlagMemoUserDCEncode. This is primarily useful with tests.
+var memoUserDCEncode = os.Getenv("MEMO_USER_DC_ENCODE") != ""
```
Quinn-With-Two-Ns: The comment says this is useful for tests, but I don't see this used in the SDK anywhere. Why did we need to add this?

yuandrew: It's set and used in `SetMemoUserDCEncode()`, so we can `defer` setting it back to its previous flag

internal/internal_flags.go
```diff
@@ -29,13 +29,20 @@ const (
 	// SDKFlagBlockedSelectorSignalReceive will cause a signal to not be lost
 	// when the Default path is blocked.
 	SDKFlagBlockedSelectorSignalReceive = 5
-	SDKFlagUnknown                      = math.MaxUint32
+	// SDKFlagMemoUserDCEncode will use the user data converter when encoding a memo. If user data converter fails,
+	// we will fallback onto the default data converter. If the default DC fails, the user DC error will be returned.
+	SDKFlagMemoUserDCEncode = 6
```
cretz: Just for my confirmation (didn't review whole PR or anything), this flag will now be set for any workflow that uses `UpsertMemo` or `ExecuteChildWorkflow` but not for any other workflows? But for now this is off by default unless the `MEMO_USER_DC_ENCODE` env var is set? And I presume one day we'll just turn it on for all users (i.e. change `var memoUserDCEncode = os.Getenv("MEMO_USER_DC_ENCODE") != ""` to `os.Getenv("MEMO_USER_DC_ENCODE") != "false"` or something)?

yuandrew: > this flag will now be set for any workflow that uses UpsertMemo or ExecuteChildWorkflow but not for any other workflows

What do you mean "not for any other workflows"? My understanding is this should be set for all workflows, regardless of if they use `UpsertMemo` or `ExecuteChildWorkflow`.

> And I presume one day we'll just turn it on for all users (i.e. change var memoUserDCEncode = os.Getenv("MEMO_USER_DC_ENCODE") != "" to os.Getenv("MEMO_USER_DC_ENCODE") != "false" or something)?

Correct, one day we will flip the switch to turn it on by default. iirc server needs a version or 2 to fall back onto, so we want to wait a bit before flipping the switch on a new SDK flag

cretz: > What do you mean "not for any other workflows"? My understanding is this should be set for all workflows, regardless of if they use UpsertMemo or ExecuteChildWorkflow.

When I looked at when `.TryUse(SDKFlagMemoUserDCEncode)` was invoked, it was only in those two situations, but I may be misreading

internal/internal_flags.go
yuandrew: as well as configuring the default behavior for tests, so we can validate both the old and new behavior

internal/internal_flags.go
yuandrew: ahh no I think you're right, but does it matter if we set the SDK flag if our workflow isn't exercising Memo code? I wanna say no. The flag is primarily for replaying and ensuring this fix doesn't cause NDE with pre-fix histories

cretz: It doesn't matter, in fact I think it's better to not have the flag for everyone, just confirming that is the case.

cretz (approved):
Since this is not on by default and we have replay tests for with flag and without, this should be safe, LGTM

cretz (approved):

PR #2122, opened 2025-12-02
Add resource_slots_cpu/mem_usage metrics

### LINKED ISSUE #2089: Add resource tuner metrics
The CPU/mem usage metrics that are present in Core/Java are absent here. https://docs.temporal.io/references/sdk-metrics#resource_slots_cpu_usage for reference.

Add them (and possibly also the pid value metrics, though those are less obviously useful and could be skipped).

yuandrew:
## What was changed
Add `resource_slots_cpu_usage` and `resource_slots_mem_usage` metrics

## Why?
Match Core and Java

## Checklist

1. Closes #2089 

2. How was this tested:
Added a test

3. Any docs updates needed?
Should update https://docs.temporal.io/references/sdk-metrics#resource_slots_cpu_usage to include Go in the list of supported SDKs

---

> [!NOTE]
> Resource tuner now emits `temporal_resource_slots_cpu_usage` and `temporal_resource_slots_mem_usage` gauges, with tests added and a minor CI workflow cleanup.
> 
> - **Resourcetuner (metrics)**:
>   - Add metric constants: `temporal_resource_slots_cpu_usage`, `temporal_resource_slots_mem_usage`.
>   - `ResourceController.pidDecision` now accepts `client.MetricsHandler` and publishes gauges via `publishResourceMetrics`.
>   - `ResourceBasedSlotSupplier.TryReserveSlot` passes `info.MetricsHandler()` to `pidDecision`.
> - **Tests**:
>   - Update `TestPidDecisions` to supply a metrics handler.
>   - Add `TestPidDecisionEmitsUsageMetrics` to verify gauge emission and values.
> - **CI**:
>   - Remove `features-repo-ref` override in `.github/workflows/ci.yml`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 394c7158f02840536e33024a24aabad1dcdf0bcc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,4 +135,3 @@ jobs:
       go-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       version: ${{github.event.pull_request.head.ref}}
       version-is-repo-ref: true
-      features-repo-ref: "build-id-breaking-change"


contrib/resourcetuner/resourcetuner.go
```diff
@@ -3,6 +3,7 @@ package resourcetuner
 import (
 	"context"
 	"errors"
+	"go.temporal.io/sdk/internal/common/metrics"
```
Quinn-With-Two-Ns: Don't think we should be using internal in a external package

Quinn-With-Two-Ns: TBH I am surprised this isn't failing a lint error

yuandrew: Removed the internal dep, are we okay still using the internal dep in the test file? I saw we were previously already relying on it

contrib/resourcetuner/resourcetuner.go
```diff
@@ -3,18 +3,24 @@ package resourcetuner
...
 	"go.einride.tech/pid"
+	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/log"
 	"go.temporal.io/sdk/worker"
 )
 
+// Metric names emitted by the resource-based tuner
+const (
+	ResourceSlotsCPUUsage = "temporal_resource_slots_cpu_usage"
+	ResourceSlotsMemUsage = "temporal_resource_slots_mem_usage"
```
cretz: Is it intentional to make these part of the user-facing API? Can we uncapitalize these?

log/noop.go
cretz, line 1: Hrmm, do we intend to use this issue to make a new user-facing logger? I guess it's not a big deal, just confirming we're not doing something for our own tests with an unintentional user-facing component. (also, could be `NopLogger()` returning the same instance everytime instead of `NewNopLogger()` but that's pedantic)

contrib/resourcetuner/resourcetuner.go
yuandrew: Wouldn't we want to make them available for users? So they can reference the exact same metric from their side?

log/noop.go
yuandrew: yeah, i guess there's no strong reason to have this new logger, I was thinking of also removing the `internal` dep from tests when making this, but ended up still relying on `internal` for metrics. Might as well continue using the `internal` package here and not expose something new

contrib/resourcetuner/resourcetuner.go
cretz: No, we have not traditionally made our metric constants available programmatically to users in Go. If we do want to do that, I think we can do that in general, but shouldn't be specific to these two IMO.

PR #2126, opened 2025-12-04
💥 [BREAKING] enable TLS if api key specified

THardy98:
## What was changed
DWISOTT

1. Part of https://github.com/temporalio/features/issues/687

2. How was this tested:
`internals/credentials_test.go`

3. Any docs updates needed?
Maybe

### DIFF
--- a/.github/workflows/nightly-throughput-stress.yml
+++ b/.github/workflows/nightly-throughput-stress.yml
@@ -4,10 +4,6 @@ on:
   schedule:
     # Run at 3 AM PST (11:00 UTC) - offset from existing nightly
     - cron: '00 11 * * *'
-  push:
-    branches:
-      - add-nightly-throughput-stress-workflow
-      - cleanup-nightly-tps-workflow
   workflow_dispatch:
     inputs:
       duration:
@@ -25,14 +21,23 @@ on:
         required: false
         default: 360
         type: number
+      is_experiment:
+        description: 'Mark this run as an experiment (excluded from nightly dashboards)'
+        required: false
+        default: false
+        type: boolean
 
 permissions:
   contents: read
+  id-token: write
 
 env:
   # Workflow configuration
   TEST_DURATION: ${{ inputs.duration || vars.NIGHTLY_TEST_DURATION || '5h' }}
   TEST_TIMEOUT: ${{ inputs.timeout || vars.NIGHTLY_TEST_TIMEOUT || '5h30m' }}
+
+  # AWS S3 metrics upload ARN
+  AWS_S3_METRICS_UPLOAD_ROLE_ARN: ${{ vars.AWS_S3_METRICS_UPLOAD_ROLE_ARN }}
 
   # Logging and artifacts
   WORKER_LOG_DIR: /tmp/throughput-stress-logs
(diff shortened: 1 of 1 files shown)


internal/client.go
```diff
@@ -542,6 +542,11 @@ type (
 		// TLS configures connection level security credentials.
 		TLS *tls.Config
 
+		// TLSDisabled explicitly disables TLS. When true, TLS will not be used even
```
cretz, line 545: > When true, TLS will not be used

Reading the code here, this is not true, TLS will still be enable if the `TLS` field is set. IMO you should either 1) disallow both this being true _and_ the `TLS` field being set (my preference), or 2) document that this being true means anything in the `TLS` field is ignored, and then respect that in code.

internal/client.go
```diff
@@ -1210,7 +1215,13 @@ func NewAPIKeyDynamicCredentials(apiKeyCallback func(context.Context) (string, e
 	return apiKeyCredentials(apiKeyCallback)
 }
 
-func (apiKeyCredentials) applyToOptions(*ConnectionOptions) error { return nil }
+func (apiKeyCredentials) applyToOptions(opts *ConnectionOptions) error {
+	// Auto-enable TLS when API key is provided and TLS is not explicitly set/disabled
```
cretz, line 1225: Might as well document to users that we do this where API key is set by users (i.e. in the Godoc on the `NewAPIKey*` functions)

cretz (commented):
(leaving main review to Go SDK owners, just a couple of notes here)

internal/client.go
THardy98: added a little line to the new api key methods noting the auto TLS behavior

internal/client.go
THardy98: Ah I should've tested this. Added validation to make it mutually exclusive + test

client/client.go
```diff
@@ -1514,6 +1514,8 @@ func HistoryFromJSON(r io.Reader, options HistoryJSONOptions) (*historypb.Histor
 // Note, this uses a fixed header value for authentication. Many users that want
 // to rotate this value without reconnecting should use
 // [NewAPIKeyDynamicCredentials].
+//
+// Note, TLS is auto-enabled when API key is provided and TLS is not explicitly set/disabled
```
yuandrew: nit: all other comment lines send in a .
```suggestion
// Note, TLS is auto-enabled when API key is provided and TLS is not explicitly set/disabled.
```

client/client.go
```diff
@@ -1528,6 +1530,8 @@ func NewAPIKeyStaticCredentials(apiKey string) Credentials {
 // "Authorization" header with "Bearer " + the given function result. If the
 // resulting string is non-empty, it will overwrite any "Authorization" header
 // that may be on the context or from existing header provider.
+//
+// Note, TLS is auto-enabled when API key is provided and TLS is not explicitly set/disabled
```
yuandrew:
```suggestion
// Note, TLS is auto-enabled when API key is provided and TLS is not explicitly set/disabled.
```

Quinn-With-Two-Ns (approved):

Issue #2084, opened 2025-10-20
Add support for static summary for side effects and mutable side effects


yuandrew:
**Is your feature request related to a problem? Please describe.**

Side Effect markers in the UI don't have any identifiable information, when there's multiple in quick succession in the UI, we're not able to identify them apart unless we click into the specific item.

**Describe the solution you'd like**

Support static summary for these items, then UI can be changed to display their summaries, similar to activities and local activities.

**Additional context**
https://github.com/temporalio/sdk-go/issues/2076

djeer:
Added PR to implement it here https://github.com/temporalio/sdk-go/pull/2129

Issue #2076, opened 2025-10-10
SideEffects markers are very difficult to navigate in the timeline


djeer:
**Is your feature request related to a problem? Please describe.**
We're using SideEffects a lot and from a user perspective it's very hard to navigate the workflow timeline when there are 10-20 side effects in a row. The only way to see what was called there is to check the payload and try to understand which method could return this payload.

**Describe the solution you'd like**
Add SideEffectOptions with Name parameter 
PR https://github.com/temporalio/sdk-go/pull/2074

**Describe alternatives you've considered**
Checked if that can be implemented on UI side, by using MutableSideEffect with IDs, but those IDs are not shown in the timeline as well. Considered building Chrome extension to fetch MutableSideEffects's IDs in the UI, but overriding marker names in SDK is the cleanest solution.

**Additional context**

Solution that didn't work well - mutable side effects - you have to expand each of them, so there's not much benefit when reading a timeline.

<img width="2650" height="1004" alt="Image" src="https://github.com/user-attachments/assets/42ae6caa-6314-4080-9ebd-aea9bfd47226" />

My [proposed](https://github.com/temporalio/sdk-go/pull/2074) solution - name override, it's shown in the timeline.

<img width="903" height="588" alt="image" src="https://github.com/user-attachments/assets/5cdaebff-8f92-4650-be73-d18898a82d85" />

yuandrew:
Thanks for the issue! I agree having a bunch of `SideEffect`s in a row in the UI is not helpful. We don't want to give `SideEffects` the ability to rename "Markers" themselves, but instead, we can add Summaries to "Markers", and then have the UI display summaries, similar to[ Activities and LocalActivities](https://github.com/temporalio/sdk-go/blob/ad990a3bfe17560c82cf84c80dfdf6c41eb43580/internal/internal_activity.go#L52C3-L52C32).

PR #2129, opened 2025-12-05
Add summary to side effect and mutable side effect

djeer:
Implements https://github.com/temporalio/sdk-go/issues/2084 in Go SDK. It is already implemented in Java SDK here https://github.com/temporalio/sdk-java/pull/2699

## What was changed
  - Adds SideEffectOptions and MutableSideEffectOptions structs with a Summary field
  - Adds SideEffectWithOptions and MutableSideEffectWithOptions functions to the workflow API
  - Wires summary through to UserMetadata on marker commands so it appears in UI/CLI
Very similar to what's already exists in [Activities and LocalActivities](https://github.com/temporalio/sdk-go/blob/ad990a3bfe17560c82cf84c80dfdf6c41eb43580/internal/internal_activity.go#L52C3-L52C32).

## Why?
The justification is in https://github.com/temporalio/sdk-go/issues/2076 , it's difficult to read a timeline with 20 SideEffect's in a row.
image

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/2084

2. How was this tested:
Local run of our application with Options parameter provided in the code
image

3. Any docs updates needed?
Current behaviour is not changed, just a new experimental feature, so I guess no?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,4 +135,3 @@ jobs:
       go-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       version: ${{github.event.pull_request.head.ref}}
       version-is-repo-ref: true
-      features-repo-ref: "build-id-breaking-change"
--- a/contrib/resourcetuner/resourcetuner.go
+++ b/contrib/resourcetuner/resourcetuner.go
@@ -10,8 +10,15 @@ import (
 	"github.com/shirou/gopsutil/v4/cpu"
 	"github.com/shirou/gopsutil/v4/mem"
 	"go.einride.tech/pid"
+	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/log"
 	"go.temporal.io/sdk/worker"
+)
+
+// Metric names emitted by the resource-based tuner
+const (
+	resourceSlotsCPUUsage = "temporal_resource_slots_cpu_usage"
+	resourceSlotsMemUsage = "temporal_resource_slots_mem_usage"
 )
 
 type ResourceBasedTunerOptions struct {
@@ -157,7 +164,7 @@ func (r *ResourceBasedSlotSupplier) TryReserveSlot(info worker.SlotReservationIn
 	numIssued := info.NumIssuedSlots()
 	if numIssued < r.options.MinSlots || (numIssued < r.options.MaxSlots &&
 		time.Since(r.lastSlotIssuedAt) > r.options.RampThrottle) {
-		decision, err := r.controller.pidDecision(info.Logger())
+		decision, err := r.controller.pidDecision(info.Logger(), info.MetricsHandler())
 		if err != nil {
 			info.Logger().Error("Error calculating resource usage", "error", err)
 			return nil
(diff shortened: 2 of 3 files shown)


test/integration_test.go
```diff
@@ -7772,6 +7772,100 @@ func (ts *IntegrationTestSuite) TestLocalActivitySummary() {
...
+	ts.NoError(run.Get(ctx, nil))
+
+	var summary string
+	iter := ts.client.GetWorkflowHistory(ctx, run.GetID(), run.GetRunID(), true, enumspb.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT)
+	for iter.HasNext() {
+		event, err := iter.Next()
+		ts.NoError(err)
+		attributes := event.GetMarkerRecordedEventAttributes()
+		if event.EventType == enumspb.EVENT_TYPE_MARKER_RECORDED && attributes.MarkerName == "SideEffect" {
+			if event.UserMetadata != nil && event.UserMetadata.Summary != nil {
```
yuandrew: I think this should be an assert instead of an `if` statement, since we only expect 1 side effect event and if this statement is false, we never assert anything.

test/integration_test.go
```diff
@@ -7772,6 +7772,100 @@ func (ts *IntegrationTestSuite) TestLocalActivitySummary() {
...
+	ts.NoError(run.Get(ctx, nil))
+
+	var summary string
+	iter := ts.client.GetWorkflowHistory(ctx, run.GetID(), run.GetRunID(), true, enumspb.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT)
+	for iter.HasNext() {
+		event, err := iter.Next()
+		ts.NoError(err)
+		attributes := event.GetMarkerRecordedEventAttributes()
+		if event.EventType == enumspb.EVENT_TYPE_MARKER_RECORDED && attributes.MarkerName == "MutableSideEffect" {
+			if event.UserMetadata != nil && event.UserMetadata.Summary != nil {
```
yuandrew: same comment here as the concern above

internal/workflow.go
```diff
@@ -604,6 +604,36 @@ type (
...
+	// NOTE: Experimental
+	//
+	// Exposed as: [go.temporal.io/sdk/workflow.SideEffectOptions]
+	SideEffectOptions struct {
+		// Summary is a single-line summary of this side effect that will appear in UI/CLI.
+		// This can be in single-line Temporal Markdown format.
+		//
+		// Optional: defaults to none/empty.
+		//
+		// NOTE: Experimental
```
yuandrew, line 617: This should be the only place we mark experimental, the Summary field itself. `SideEffects` and `MutableSideEffects` are not experimental, thus `SideEffectWithOptions` and `MutableSideEffectWithOptions` are not experimental. This also includes the `SideEffectOptions` struct itself not being marked as experimental

yuandrew (commented):
Overall looks good! Left some minor feedback. Also looks like you'll need to add the new interceptor methods to the `TestInterceptorCalls` test

yuandrew:
@djeer thanks for the ping, changes look good! Looks like CI is still failing, once you fix up the test we should be good to merge

yuandrew (approved):
Looks good! Thanks for the contribution

Issue #2066, opened 2025-10-02
Lost messages from workflow.NewChannel(ctx)


yuandrew:
## Expected Behavior

No messages get lost

## Actual Behavior

Messages are getting lost. This bug was introduced in 1.35.0, from #1762

## Steps to Reproduce the Problem
Setting `unblockSelectorSignal` to `true` causes the lost items 

```
func (s *WorkflowTestSuiteUnitTest) TestChannelWorkerPattern() {
	unblockSelectorSignal = true // test passes when set to false

	assert := s.Assert()
	require := s.Require()

	// Test configuration
	numWorkers := 10
	numItems := 50
	items := make([]int, numItems)
	for i := 0; i < numItems; i++ {
		items[i] = i
	}
	// Track which items were processed
	processedItems := make(map[int]int) // maps input -> output
	// Simple workflow that mimics the channel worker pattern
	workflowFn := func(ctx Context) error {
		// Create input channel and feed items
		inputCh := NewChannel(ctx)
		Go(ctx, func(ctx Context) {
			for _, item := range items {
				inputCh.Send(ctx, item)
			}
			inputCh.Close()
		})

		// Start workers
		outputCh := NewChannel(ctx)
		wg := NewWaitGroup(ctx)
		wg.Add(numWorkers)
		// Wait group cleanup goroutine
		Go(ctx, func(ctx Context) {
			wg.Wait(ctx)
			outputCh.Close()
		})
		// Start worker goroutines
		for i := 0; i < numWorkers; i++ {
			Go(ctx, func(ctx Context) {
				defer wg.Done()
				stop := false
				selector := NewSelector(ctx)
				// Stop worker if workflow is cancelled
				selector.AddReceive(ctx.Done(), func(c ReceiveChannel, more bool) {
					stop = true
				})
				// Receive items from input channel
				selector.AddReceive(inputCh, func(ch ReceiveChannel, more bool) {
					stop = !more
					var input int
					if ch.Receive(ctx, &input) {
						output := input * 2
						outputCh.Send(ctx, output)
					}
				})
				// Worker loop
				for !stop {
					selector.Select(ctx)
				}
			})
		}
		// Collect outputs
		for i := 0; i < numItems; i++ {
			var output int
			if outputCh.Receive(ctx, &output) {
				// output = input * 2, so input = output / 2
				processedItems[output/2] = output
			}
		}
		return nil
	}
	env := s.NewTestWorkflowEnvironment()
[...]

PR #2134, opened 2025-12-10
Selector Signal fix with multiple workers

yuandrew:
## What was changed
Fixed a scenario where selectors without default branch was having shared data `c.recValue` overwritten. 

Added a new flag `SDKFlagWorkflowNewChannelLostMessages` to fix this, as there were 2 SDK versions (v1.35.0 and v1.36.0) that enabled `SDKFlagBlockedSelectorSignalReceive`, so we need a new flag so those histories don't break.

## Why?
Fix bug

Both flags will be enabled in a future release (both currently disabled, so no behavior change with this PR), after this new flag's introduction has baked at least for a release or 2. No point in re-enabling `SDKFlagBlockedSelectorSignalReceive` if we don't also enable `SDKFlagWorkflowNewChannelLostMessages`

## Checklist

1. #2066 and #1624 (will close issues when the SDK flag is enabled)

2. How was this tested:
Added new test

Added replay test verifying history with just `SDKFlagBlockedSelectorSignalReceive` doesn't break. This is primarily valuable once we enable `SDKFlagWorkflowNewChannelLostMessages` in a future release

---

> [!NOTE]
> **Medium Risk**
> Touches core deterministic workflow selector/channel internals; while gated behind a new default-off flag and covered by unit + replay tests, mistakes here could impact message delivery semantics or replay behavior.
> 
> **Overview**
> Fixes a workflow `Selector` receive edge case where, with `SDKFlagBlockedSelectorSignalReceive` enabled, multiple blocked selectors on the same channel could overwrite `channelImpl.recValue` and cause a dropped/zero-value receive.
> 
> Introduces `SDKFlagWorkflowNewChannelLostMessages` (default disabled) and updates `selectorImpl.Select` to only pre-store `recValue` in the callback when appropriate (e.g., when a default branch exists), otherwise deferring storage to the per-callback `readyBranch` to prevent cross-selector clobbering. Adds unit coverage for the “two workers on one channel” pattern and a replay test + history JSON to ensure old behavior remains replayable when only the older flag is present in history.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c618c78d3077849707df1cae502f409f3c083304. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/contrib/sysinfo/go.mod
+++ b/contrib/sysinfo/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/power-devops/perfstat v0.0.0-20210106213030-5aafc221ea8c // indirect
 	github.com/robfig/cron v1.2.0 // indirect
 	github.com/shoenig/go-m1cpu v0.1.6 // indirect
-	github.com/sirupsen/logrus v1.9.0 // indirect
+	github.com/sirupsen/logrus v1.9.1 // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
 	github.com/tklauser/go-sysconf v0.3.12 // indirect
 	github.com/tklauser/numcpus v0.6.1 // indirect
--- a/contrib/sysinfo/go.sum
+++ b/contrib/sysinfo/go.sum
@@ -54,8 +54,8 @@ github.com/shoenig/go-m1cpu v0.1.6 h1:nxdKQNcEB6vzgA2E2bvzKIYRuNj7XNJ4S/aRSwKzFt
 github.com/shoenig/go-m1cpu v0.1.6/go.mod h1:1JJMcUBvfNwpq05QDQVAnx3gUHr9IYF7GNg9SUEw2VQ=
 github.com/shoenig/test v0.6.4 h1:kVTaSd7WLz5WZ2IaoM0RSzRsUD+m8wRR+5qvntpn4LU=
 github.com/shoenig/test v0.6.4/go.mod h1:byHiCGXqrVaflBLAMq/srcZIHynQPQgeyvkvXnjqq0k=
-github.com/sirupsen/logrus v1.9.0 h1:trlNQbNUG3OdDrDil03MCb1H2o9nJ1x4/5LYw7byDE0=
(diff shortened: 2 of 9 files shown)


internal/internal_flags.go
```diff
@@ -34,7 +34,7 @@ const (
 
 // unblockSelectorSignal exists to allow us to configure the default behavior of
 // SDKFlagBlockedSelectorSignalReceive. This is primarily useful with tests.
-var unblockSelectorSignal = os.Getenv("UNBLOCK_SIGNAL_SELECTOR") != ""
```
Quinn-With-Two-Ns: Is there any roll back concerns here? Do we need to protect this behaviour by another flag?

yuandrew: Added another flag

test/replaytests/channel-worker-blocked-selector.json
cretz, line 1: Was this history generated from the current SDK before this flag and confirms with the new flag it still works?

I'm wondering if there are 3 SDK replay tests needed here: 1 from before a couple of versions ago, 1 from current released, and 1 from with this PR? If not no worries.

I admit I don't understand the PR or the problem well enough, just confirming coverage/confidence.

yuandrew: Yes, these were generated off of main before this solution (and flag) were introduced. 

Good question, the three scenarios you mentioned are these SDK flag combinations: none, 5, and 5+7.

`TestSelectorNonBlocking` already should cover `none`, previously implemented in the original solution, and this new test `TestChannelWorkerWithBlockedSelectorFlag` now covers flag 5. I'm not sure we need a replay test for 5+7, where we have unit tests that validate the behavior. 

I guess it wouldn't hurt to add, but I'm thinking replay tests are primarily used for ensuring old histories don't hit NDE when we introduce a new flag.

cretz (approved):
I admit I don't understand this very well and may not easily be able to, but if we have replay tests confirming all scenarios are safe, works for me.

PR #2135, opened 2025-12-11
Plugin support

### LINKED ISSUE #2020: Plugin support

See https://github.com/temporalio/features/issues/652

cretz:
## What was changed

Added `client.Plugin`, `worker.Plugin`, and `temporal.SimplePlugin` and everything required to make those work.

Notes:

* It is recognized that this is missing test suite functionality. I think this may need to be solved as a followup because it involves a discussion around worker options design for test suites.
* There are some tradeoffs that had to be made compared to the Python and .NET forms of this feature, but all similar functionality should be captured (e.g. ability to register things in plugins, ability to surround workers and replayers with the same context code in simple plugin, etc)

## Checklist

1. Closes #2020

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -553,6 +553,12 @@ type (
 		// TLS configures connection level security credentials.
 		TLS *tls.Config
 
+		// TLSDisabled explicitly disables TLS. When true, TLS will not be used even
+		// if API key credentials are provided (which would normally auto-enable TLS).
+		// This is not recommended for production use as it sends credentials in plaintext.
+		// This option is mutually exclusive with TLS - an error will be returned if both are set.
+		TLSDisabled bool
+
 		// Authority specifies the value to be used as the :authority pseudo-header.
 		// This value only used when TLS is nil.
 		Authority string
@@ -1004,6 +1010,11 @@ func newClient(ctx context.Context, options ClientOptions, existing Client) (Cli
 		options.Logger.Info("No logger configured for temporal client. Created default one.")
 	}
 
+	// Validate mutually exclusive TLS options
+	if options.ConnectionOptions.TLS != nil && options.ConnectionOptions.TLSDisabled {
+		return nil, fmt.Errorf("cannot set both TLS and TLSDisabled in ConnectionOptions")
+	}
+
 	if options.Credentials != nil {
 		if err := options.Credentials.applyToOptions(&options.ConnectionOptions); err != nil {
 			return nil, err
@@ -1262,7 +1273,13 @@ func NewAPIKeyDynamicCredentials(apiKeyCallback func(context.Context) (string, e
 	return apiKeyCredentials(apiKeyCallback)
 }
 
-func (apiKeyCredentials) applyToOptions(*ConnectionOptions) error { return nil }
+func (apiKeyCredentials) applyToOptions(opts *ConnectionOptions) error {
+	// Auto-enable TLS when API key is provided and TLS is not explicitly set/disabled
+	if opts.TLS == nil && !opts.TLSDisabled {
+		opts.TLS = &tls.Config{}
+	}
+	return nil
+}
 
 func (a apiKeyCredentials) gRPCInterceptor() grpc.UnaryClientInterceptor { return a.gRPCIntercept }
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,4 +135,3 @@ jobs:
       go-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       version: ${{github.event.pull_request.head.ref}}
       version-is-repo-ref: true
-      features-repo-ref: "build-id-breaking-change"
--- a/.github/workflows/nightly-throughput-stress.yml
+++ b/.github/workflows/nightly-throughput-stress.yml
@@ -4,10 +4,6 @@ on:
   schedule:
     # Run at 3 AM PST (11:00 UTC) - offset from existing nightly
     - cron: '00 11 * * *'
-  push:
-    branches:
-      - add-nightly-throughput-stress-workflow
-      - cleanup-nightly-tps-workflow
   workflow_dispatch:
     inputs:
       duration:
@@ -25,14 +21,23 @@ on:
         required: false
         default: 360
         type: number
+      is_experiment:
+        description: 'Mark this run as an experiment (excluded from nightly dashboards)'
+        required: false
+        default: false
+        type: boolean
 
 permissions:
   contents: read
+  id-token: write
 
 env:
   # Workflow configuration
   TEST_DURATION: ${{ inputs.duration || vars.NIGHTLY_TEST_DURATION || '5h' }}
   TEST_TIMEOUT: ${{ inputs.timeout || vars.NIGHTLY_TEST_TIMEOUT || '5h30m' }}
+
+  # AWS S3 metrics upload ARN
+  AWS_S3_METRICS_UPLOAD_ROLE_ARN: ${{ vars.AWS_S3_METRICS_UPLOAD_ROLE_ARN }}
 
   # Logging and artifacts
   WORKER_LOG_DIR: /tmp/throughput-stress-logs
@@ -42,13 +47,16 @@ env:
   OMES_REF: main
   RUN_ID: ${{ github.run_id }}-throughput-stress
 
+  # Prometheus version
+  PROM_VERSION: "3.8.0"
+
+  # Language
+  SDK_LANG: "go"
+
 jobs:
   throughput-stress:
     runs-on: ubuntu-latest-4-cores
     timeout-minutes: ${{ fromJSON(inputs.job_timeout_minutes || (vars.NIGHTLY_JOB_TIMEOUT_MINUTES || '360')) }}
-    permissions:
-      contents: read
-      actions: write
 
     steps:
       - name: Print test configuration
@@ -85,6 +93,13 @@ jobs:
       - name: Install Temporal CLI
         uses: temporalio/setup-temporal@v0
 
+      - name: Install Prometheus
+        run: |
+          wget -q https://github.com/prometheus/prometheus/releases/download/v${PROM_VERSION}/prometheus-${PROM_VERSION}.linux-amd64.tar.gz
+          tar xzf prometheus-${PROM_VERSION}.linux-amd64.tar.gz
+          sudo mv prometheus-${PROM_VERSION}.linux-amd64/prometheus /usr/local/bin/
+          prometheus --version
+
       - name: Setup log directory
         run: mkdir -p $WORKER_LOG_DIR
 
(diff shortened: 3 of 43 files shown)


internal/cmd/tools/doclink/doclink.go
```diff
@@ -545,6 +545,12 @@ func isValidDefinition(line string, inGroup *string, insideStruct *bool) bool {
 
 // Checks if `line` is a valid definition, and that definition is for `private`
 func isValidDefinitionWithMatch(line, private string, inGroup string, insideStruct bool) bool {
+	// Vars with underscores are often used to assert interface validation and
```
yuandrew, line 548: nice

internal/client.go
```diff
@@ -522,6 +522,9 @@ type (
 
 		// If set true, error code labels will not be included on request failure metrics.
 		DisableErrorCodeMetricTags bool
+
+		// TODO(cretz): Document that it is undefined behavior if plugins mutate this option, and it shouldn't be done
```
yuandrew: Will this TODO be tackled in this PR?

worker/plugin.go
```diff
@@ -0,0 +1,45 @@
...
+type PluginBase = internal.WorkerPluginBase
+
+// PluginConfigureWorkerOptions are options for ConfigureWorker on a
+// worker plugin.
+type PluginConfigureWorkerOptions = internal.WorkerPluginConfigureWorkerOptions
+
+// PluginConfigureWorkerRegistryOptions are the set of callbacks that can
+// be adjusted by plugins when configuring workers. If adjusting a callback that
+// is already set, implementers may want to take care to invoke the existing
+// callback inside their own.
```
yuandrew, line 29: I'm not sure I understand the last sentence or why it needs a special callout, this is only if they want the newly adjusted callback to continue to do the old callback's logic, on top of what's new right? If they want to adjust their plugin callback to do something completely different, then they wouldn't want to do this?

internal/worker.go
```diff
@@ -390,6 +390,9 @@ type (
 		//
 		// NOTE: Experimental
 		NexusTaskPollerBehavior PollerBehavior
+
+		// TODO(cretz): Document
```
yuandrew: another TODO that should be addressed

internal/plugin.go
```diff
@@ -0,0 +1,626 @@
...
+	// invoked after any above options are set.
+	ConfigureWorker func(context.Context, WorkerPluginConfigureWorkerOptions) error
+
+	// ConfigureWorkflowReplayer if set is invoked to adjust workflow replayer
+	// options. This is invoked after any above options are set.
+	ConfigureWorkflowReplayer func(context.Context, WorkerPluginConfigureWorkflowReplayerOptions) error
+
+	// RunContextBefore is invoked on worker start or before each workflow
+	// replay of a replayer. Implementers can use this to register items or
+	// simply start something needed.
```
yuandrew, line 383: > register items

refers to registering things onto the Context, right?

> simply start something needed

maybe more of a nit: this phrase feels strange to me, maybe because it feels a little vague and I'm not able to imagine what an example is

internal/plugin.go
```diff
@@ -0,0 +1,626 @@
...
+	// ClientOptions are the set of options used for the client. These should
+	// not be mutated, that should be done via the ConfigureClient method.
+	ClientOptions ClientOptions
+
+	// Lazy is whether the new client call is being invoked lazily or not.
+	Lazy bool
+
+	// FromExisting is set to a value if this client is being created from an
+	// existing client.
+	FromExisting Client
```
yuandrew, line 79: Some reason my intuition hearing `FromExisting` makes me think this var is a `bool`. After thinking for a second I get that it's a client, but I wonder if naming can make it more clear it's a client? 

`ExistingClient`? Maybe the docstring can mention it is the existing client? `to a value` is kinda vague

internal/internal_worker.go
```diff
@@ -1478,6 +1519,9 @@ type WorkflowReplayerOptions struct {
 	// Optional: Disable the default 1 second deadlock detection timeout. This option can be used to step through
 	// workflow code with multiple breakpoints in a debugger.
 	DisableDeadlockDetection bool
+
+	// TODO(cretz): Document
```
yuandrew: todo

internal/plugin.go
```diff
@@ -0,0 +1,626 @@
...
+}
+
+// WorkerPluginConfigureWorkflowReplayerOptions are options for
+// ConfigureWorkflowReplayer on a worker plugin.
+//
+// Exposed as: [go.temporal.io/sdk/worker.PluginConfigureWorkflowReplayerOptions]
+type WorkerPluginConfigureWorkflowReplayerOptions struct {
+	// WorkflowReplayerInstanceKey is the unique, immutable instance key for
+	// this workflow replayer.
+	WorkflowReplayerInstanceKey string
```
yuandrew, line 253: This is new, I don't think any of the core langs have this. Is there any utility of this I'm not seeing besides giving each replay an ID and allowing tests to verify Plugins are being applied to replay in the correct sequence?

Correct me if I'm wrong, but each time a worker replays a workflow, it'll generate a new `WorkflowReplayerInstanceKey`. So A single worker, with its own `WorkerInstanceKey`, can generate many new `WorkflowReplayerInstanceKey`s

internal/plugin.go
```diff
@@ -0,0 +1,626 @@
...
+	WorkflowReplayerRegistryOptions *WorkerPluginConfigureWorkflowReplayerRegistryOptions
+}
+
+// WorkerPluginConfigureWorkflowReplayerRegistryOptions are the set of callbacks
+// that can be adjusted by plugins when configuring workflow replayers. If
+// adjusting a callback that is already set, implementers may want to take care
+// to invoke the existing callback inside their own.
+//
+// Exposed as: [go.temporal.io/sdk/worker.PluginConfigureWorkflowReplayerRegistryOptions]
+type WorkerPluginConfigureWorkflowReplayerRegistryOptions struct {
```
yuandrew, line 274: Is this the longest var name in the SDK? 😂

internal/client.go
cretz: Ah, I completely forgot to add docs to these `Plugin` fields in client and replayer options. Will fix.

worker/plugin.go
cretz: Right. But I figured it was important to call out that if they set this, they overwrite what was set by other plugins, because at first glance it could look like each plugin can set this uniquely unrelated to each other.

internal/plugin.go
cretz: Not sure how to make it more clear that it's a client than the type :-). But yes, I can adjust the doc a bit.

internal/plugin.go
cretz: > refers to registering things onto the Context, right?

No, the options

internal/plugin.go
cretz: The core langs don't have this because they will be getting access to the actual worker/replay instance in their "run worker" or "replay workflow" calls (which has the options and such). This is needed if a user needs to correlate configuration with the instance.

Not specific to this but for all `XInstanceKey`s in this PR, there needs to be a way for workers to correlate start with stop (and configure, though that's less important) and workflow replayers to correlate successive replay runs together. The worker side is the most important, it's just that I went ahead and added it to replayer for clarity there too.

For instance, say your plugin on worker start registered an activity with a newly opened database connection. When should it close that database connection?

I do expect this to be the same value as the actual instance key we're using for worker heartbeating when we do that at some point, but since this is all experimental (eek, I need to go mark them all as such), we can adjust as needed in the future.

internal/plugin.go
cretz, line 1: I need to go back through all of this and mark it experimental (here and in public packages)

cretz: Fixed

internal/internal_worker.go
cretz: Fixed

internal/plugin.go
cretz: Fixed

internal/worker.go
cretz: Fixed

internal/client.go
cretz: Fixed

yuandrew (approved):

internal/plugin.go
```diff
@@ -0,0 +1,660 @@
...
+// WorkerPluginConfigureWorkerRegistryOptions are the set of callbacks that can
+// be adjusted by plugins when configuring workers. If adjusting a callback that
+// is already set, implementers may want to take care to invoke the existing
+// callback inside their own.
+//
+// Exposed as: [go.temporal.io/sdk/worker.PluginConfigureWorkerRegistryOptions]
+//
+// NOTE: Experimental
+type WorkerPluginConfigureWorkerRegistryOptions struct {
+	OnRegisterWorkflow        func(any, RegisterWorkflowOptions)
```
Quinn-With-Two-Ns: A bit pedantic, but could we document these callbacks? They are a bit unclear since they have to take `any` as the first parameter

internal/plugin.go
```diff
@@ -0,0 +1,660 @@
...
+type WorkerPluginStartWorkerOptions struct {
+	// WorkerInstanceKey is the unique, immutable instance key for this worker.
+	WorkerInstanceKey string
+
+	// WorkerRegistry is the worker registry plugins can use to register items
+	// with the worker. Implementers should usually not mutate this before
+	// passing to "next", but instead add register callbacks in ConfigureWorker
+	// if needed. Calls to this registry to invoke the OnX callbacks set in
+	// ConfigureWorker.
+	WorkerRegistry interface {
```
Quinn-With-Two-Ns, line 225: What is the difference between this interface and `WorkerPluginConfigureWorkerRegistryOptions`?

internal/plugin.go
```diff
@@ -0,0 +1,660 @@
...
+		options WorkerPluginStartWorkerOptions,
+		next func(context.Context, WorkerPluginStartWorkerOptions) error,
+	) error
+
+	// StopWorker is called to stop a worker. This is called on Worker.Stop or
+	// if Worker.Run is interrupted via its interrupt channel. However, if a
+	// fatal worker error occurs during Worker.Run, this may not be called.
+	// Implementers can account for this situation by setting OnFatalError in
+	// the worker options. Implementers should invoke next.
+	StopWorker(
```
Quinn-With-Two-Ns, line 131: Is there a reason you didn't allow returning an error here , but did for `ConfigureWorker`?

cretz: Just because I think it's more normal to fail configuration than it is to fail stopping from a Temporal POV. But I can change `ConfigureWorker` to ask users to panic on validation issues if that's what we want.

internal/plugin.go
cretz: That one is workflow only. I could technically make a workflow only one and embed it here, but it didn't seem to have enough value.

internal/plugin.go
cretz: Done

PR #2137, opened 2025-12-18
Update to mockery 3 and run against all public interfaces.

thespags:
## What was changed
Add .mockery file for mocker 3 and ran, currently using all public interfaces but could limit it to the previous mocked interfaces.

## Why?
Mockery 3 adds a lot of helpful additions to make using mocks easier, e.g. NewMockType(t) will auto wire register assertExpectations for you.

## Checklist

1. Closes 
N/A

2. How was this tested:
Currently manually running mockery and building.
You could add CI step to run `mockery` and compare the diff to ensure the mocks stay up to date, that seems like a more invasive change to update the CI.

3. Any docs updates needed?
No current mock docs, you could add docs describing the mocks folder and how to run mockery.


thespags:
@Quinn-With-Two-Ns , that's from 5 years ago, so I assume the issue was out of date and that it wasn't planned anymore.

PR #2145, opened 2026-01-09
Additional fields in env config

stephanos:
## What was changed

Adds ability to encode/decode additional profile keys in env config on top of pre-defined ones. 

## Why?

The Temporal CLI recently [added support for an OAuth profile keys in the env config](https://github.com/temporalio/cli/pull/892). It had to duplicate code from the envconfig package to do that. If the envconfig package supports encoding/decoding additional fields, that code can be deleted.

Also see previous convo: https://github.com/temporalio/cli/pull/892#discussion_r2662163834

See https://github.com/temporalio/cli/pull/901 for how this would be used.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,4 +135,3 @@ jobs:
       go-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       version: ${{github.event.pull_request.head.ref}}
       version-is-repo-ref: true
-      features-repo-ref: "build-id-breaking-change"
--- a/client/plugin.go
+++ b/client/plugin.go
@@ -1,0 +1,32 @@
+package client
+
+import "go.temporal.io/sdk/internal"
+
+// Plugin is a plugin that can configure client options and surround client
+// creation/connection. Many plugin implementers may prefer the simpler
+// [go.temporal.io/sdk/temporal.SimplePlugin] instead.
+//
+// All client plugins must embed [go.temporal.io/sdk/client.PluginBase]. All
+// plugins must implement Name().
+//
+// All client plugins that also implement [go.temporal.io/sdk/worker.Plugin] are
+// automatically configured on workers made from the client.
+//
+// NOTE: Experimental
+type Plugin = internal.ClientPlugin
+
+// PluginBase must be embedded into client plugin implementations.
+//
+// NOTE: Experimental
+type PluginBase = internal.ClientPluginBase
+
+// PluginConfigureClientOptions are options for ConfigureClient on a
+// client plugin.
+//
+// NOTE: Experimental
+type PluginConfigureClientOptions = internal.ClientPluginConfigureClientOptions
+
+// PluginNewClientOptions are options for NewClient on a client plugin.
+//
+// NOTE: Experimental
+type PluginNewClientOptions = internal.ClientPluginNewClientOptions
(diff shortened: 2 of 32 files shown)


contrib/envconfig/client_config_toml_test.go
```diff
@@ -1,10 +1,11 @@
-package envconfig_test
+package envconfig
```
stephanos, line 1: Needed to change this to have access to `knownProfileKeys` in order to have a test that ensures it's in-sync with tags in `tomlClientConfigProfile`. Another option I considered was doing the key lookup at runtime (via reflection); that would remove the need for this change. Let me which tradeoff you prefer.

contrib/envconfig/client_config_toml.go
```diff
@@ -8,49 +8,130 @@ import (
...
+	AdditionalProfileFields map[string]map[string]any
 }
 
 // ToTOML converts the client config to TOML. Note, this may not be byte-for-byte exactly what may have been set in
 // [ClientConfig.FromTOML].
 func (c *ClientConfig) ToTOML(options ClientConfigToTOMLOptions) ([]byte, error) {
 	var conf tomlClientConfig
 	conf.fromClientConfig(c)
+
+	// Encode to TOML then decode to map for merging additional fields
```
stephanos, line 38: There is an optimization opportunity here to only do the re-encoding if `AdditionalProfileFields` is set, but that would add more code and increase complexity. Let me know if you prefer the opposite tradeoff here.

KeithB-Temporal: My opinion: I don't think we need that level of performance optimization for the CLI operations. I'd prefer to minimize complexity and maximize readability and maintainability.

contrib/envconfig/client_config_toml_test.go
KeithB-Temporal: Simple is best and most maintainable.

contrib/envconfig/client_config_toml.go
cretz: Not sure it is much more complexity to add a conditional around this raw stuff, but meh, not a big deal to decode and re-encode for no reason

contrib/envconfig/client_config_toml.go
```diff
@@ -8,49 +8,130 @@ import (
...
 	var buf bytes.Buffer
-	enc := toml.NewEncoder(&buf)
+	if err := toml.NewEncoder(&buf).Encode(&conf); err != nil {
+		return nil, err
+	}
+	var rawConf map[string]any
+	if _, err := toml.Decode(buf.String(), &rawConf); err != nil {
+		return nil, err
+	}
```
cretz, line 46: Could also have the alternative approach of changing `var conf tomlClientConfig` + `conf.fromClientConfig(c)` to `func (*ClientConfigToTOMLOptions) toTomlableMap() map[string]any`, but since you're mutating a couple of layers deep, it might end up being encode/decode instead of explicit map anyways, so this is fine too.

contrib/envconfig/client_config_toml_test.go
cretz: In SDKs, we like to write tests that users might, and therefore in this case if you change the package, it will be hard for us to catch when we accidentally make something internal. Not a big deal though, but allowing all tests to access privates is not ideal for how we like to demonstrate what is something the user can author (matters more in other parts of the Go SDK where integration tests can help us show what we haven't exposed).

yuandrew: Agree with Chad on both points, should be fine, we definitely want `TestKnownProfileKeysInSync` to test we're keeping these 2 in sync

PR #2146, opened 2026-01-13
Nexus caller timeouts

bergundy:
## Overview

Implements two new granular timeout types for Nexus operations, allowing callers to have fine-grained control over different phases of operation execution:

- **Schedule-to-Start Timeout**: Maximum time to wait for an operation to be started (or completed if synchronous) by the handler
- **Start-to-Close Timeout**: Maximum time to wait for an asynchronous operation to complete after it has been started

These timeouts complement the existing **Schedule-to-Close Timeout** to provide better control and diagnostics for Nexus operation execution.

See also https://github.com/temporalio/api/pull/695 and the corresponding server PR: https://github.com/temporalio/temporal/pull/9010.

Do not merge until the API has been released.

### DIFF
--- a/contrib/datadog/go.mod
+++ b/contrib/datadog/go.mod
@@ -81,7 +81,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.37.0 // indirect
 	go.opentelemetry.io/otel/sdk v1.37.0 // indirect
 	go.opentelemetry.io/otel/trace v1.37.0 // indirect
-	go.temporal.io/api v1.62.0 // indirect
+	go.temporal.io/api v1.62.1 // indirect
 	go.uber.org/atomic v1.11.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	go.uber.org/zap v1.27.0 // indirect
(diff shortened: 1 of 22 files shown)


test/nexus_test.go
```diff
@@ -2906,3 +2906,112 @@ func TestNexusTracingInterceptor(t *testing.T) {
 		})
 	}
 }
+
+func TestWorkflowTestSuite_WorkflowRunOperation_ScheduleToStartTimeout(t *testing.T) {
```
Quinn-With-Two-Ns, line 2907: Once the server support is merged can we make a CLI RC release to test against?

bergundy: We don't have tests for any timeouts with the server. I can add those if we want but IMHO it's enough to have the test env coverage because it means the command is properly generated SDK side and the server has functional tests using the gRPC API.

cretz: I think we need to be able to assert the same behavior that a user can assert at runtime. This is a primary reason for integration tests in SDK, to confirm the runtime behavior as users see it is as expected. If a user can't even use this feature, there is no reason to merge/release it yet.

Quinn-With-Two-Ns:
For activity timeouts we expose the timeouts to the callee in the `ActivityInfo`, do we want to do the same for Nexus?

test/nexus_test.go
bergundy: Users will be able to use this feature in cloud, I will cut a new CLI RC so we can test and users can use in local dev.

cretz: :+1: Using in cloud has never been enough for an open-source SDK feature like this, but a CLI RC anyone (including our tests) can use works. Standalone activities will also benefit from this RC release.

test/nexus_test.go
```diff
@@ -2906,3 +2903,224 @@ func TestNexusTracingInterceptor(t *testing.T) {
...
+		opStartDelay           time.Duration
+		expectedTimeoutType    enumspb.TimeoutType
+		expectOperationToken   bool
+	}{
+		{
+			name:                   "schedule-to-start fires first",
+			scheduleToCloseTimeout: 10 * time.Second,
+			scheduleToStartTimeout: 3 * time.Second,
+			startToCloseTimeout:    5 * time.Second,
+			opStartDelay:           4 * time.Second,
```
yuandrew, line 3040: This test adds 9 additional seconds total to our test suite, if I'm reading this correctly, can we lower this? Find a good spot where it's not racy with CI, but a little faster than 9 seconds each run? Maybe 1s for the shortest timeout?

bergundy: There's a config in the server that prevents dispatching nexus operations that have less than 1.5 seconds to run. I worked with that but also wanted to give enough time for a busy CI to eliminate races.
I can decrease the timeout via dynamic config though if we want to trim it down. It's just still going to be in the order of magnitude of seconds to prevent races so I'm not inclined to tweak it too much and prefer more stability.

PR #2147, opened 2026-01-13
Remove nil search attribute value from workflow info

awln-temporal:
## What was changed
Remove nil search attribute value from workflow info.

## Why?
`nil` search attribute values should not be displayed in the UI. These are valid values to provide to erase any prior values upserted in Workflow Executions, but they should not be displayed or recorded in History Events.

There is a bug in the UI when a user intends on deleting search attributes from a workflow through an Upsert command, but also decides to continue the workflow as a new run. When both are specified, the new run displays `CustomSearchAttribute: null` in the Search Attributes UI.

## Checklist

1. How was this tested:
samples-go child-workflow-continue-as-new, upserting search attributes in the child workflow definition before continuing as new.

Screenshot 2026-01-13 at 6 10 41 PM

2. Any docs updates needed?
N/A

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -466,9 +466,23 @@ func mergeSearchAttributes(current, upsert *commonpb.SearchAttributes) *commonpb
 
 	fields := current.IndexedFields
 	for k, v := range upsert.IndexedFields {
-		fields[k] = v
+		// Delete nil payloads from workflow info. Nil payloads indicate user erasing a search attribute value.
+		// In a continue-as-new execution, search attributes are referenced from workflow info. History Events and UI should
+		// not display a nil value. UpsertWorkflowSearchAttributes will still receive the full map of search attributes.
+		if isNilPayload(v) {
+			delete(fields, k)
+		} else {
+			fields[k] = v
+		}
 	}
 	return current
+}
+
+func isNilPayload(p *commonpb.Payload) bool {
+	if p == nil {
+		return true
+	}
+	return string(p.GetMetadata()[converter.MetadataEncoding]) == converter.MetadataEncodingNil
 }
 
 func validateAndSerializeSearchAttributes(attributes map[string]interface{}) (*commonpb.SearchAttributes, error) {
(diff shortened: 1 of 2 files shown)


internal/internal_event_handlers.go
```diff
@@ -466,11 +466,25 @@ func mergeSearchAttributes(current, upsert *commonpb.SearchAttributes) *commonpb
 
 	fields := current.IndexedFields
 	for k, v := range upsert.IndexedFields {
-		fields[k] = v
+		// Delete nil payloads from workflow info. Nil payloads indicate user erasing a search attribute value.
+		// In a continue-as-new execution, search attributes are referenced from workflow info. History Events and UI should
```
cretz, line 470: While I think this may be the accurate fix, I am concerned about breaking existing user code here by changing workflow info for everyone. I think we may have to either use the advanced SDK flagging feature, or trim these off at continue-as-new execution-building time instead of here at merging time. I am leaning towards the latter and just documenting that workflow info may have nil attributes after upsert. @yuandrew - thoughts?

awln-temporal: How do clients typically access WorkflowInfo or view search attributes? My impression is that anywhere that calls `TypedSearchAttributes()` will call `convertToTypedSearchAttributes()` (or places like `DescribeWorkflow()` that directly call `convertToTypedSearchAttributes()` on workflowInfo) which also filters out nil values. Let me know if I'm missing something, will defer to SDK team suggestions.

yuandrew: Trying to brush up on search attributes in Go. People often use `workflow.GetInfo()`  where this would be a breaking change if they did `info.SearchAttributes.IndexedFields["key"]`, so I think it's a valid concern for backwards compat (our tests do this too). 

I may not understand the core issue well enough, but is this something the UI itself can fix? If it can't be a UI-only fix, I also lean towards Chad's latter option.

@cretz should we be concerned with removing these nil entries across the CaN boundary?

cretz: > should we be concerned with removing these nil entries across the CaN boundary?

Now that I think about it, it might just be a UI issue of displaying them and not that they actually get set when the workflow starts. Let me ask around a bit internally.

cretz:
We would have to apply a different style to make this change since it can technically cause non-determinism. However, to address your specific problem here, I have opened up a less-intrusive issue at https://github.com/temporalio/sdk-go/issues/2150. It won't affect upsert, but will affect continue as new.

PR #2148, opened 2026-01-14
Add worker heartbeat support

### LINKED ISSUE #2094: Worker Heartbeating
Implement worker heartbeating, which is sent via a new background nexus worker.

Basically need to design a Go version of https://github.com/temporalio/sdk-core/pull/953 and https://github.com/temporalio/sdk-core/pull/1038

**Additional context**

yuandrew:
## What was changed

Key changes:
- Add HeartbeatMetricsHandler to capture metrics needed for heartbeats
- Add internal_worker_heartbeat.go for heartbeat worker management
- 💥 Removed `contrib/resourcetuner` lib, resource tuner portion has been moved into the SDK and is exposed through the `worker` package, while a new `contrib/sysinfo` package was created to handle system resource measurement
  - We never released a v1.0.0 of `contrib/resourcetuner`, so this should be okay.
- Plumb heartbeat data through workers (workflow, activity, nexus, local activity)
- always send `ShutdownWorker` RPC to give final heartbeat, even if not using sticky queues
- Move `verifyNamespaceExist` calls from each specific worker, and instead call at the `AggregatedWorker` level and cache for worker heartbeating

## Why?

New feature!

## Checklist

1. Closes #2094

2. How was this tested:

Added a bunch of tests to `worker_heartbeat_test.go`
- Add integration tests for worker heartbeat functionality

3. Any docs updates needed?

---

> [!NOTE]
> **High Risk**
> Touches core worker lifecycle, RPCs, and metrics plumbing, adding new background goroutines and shutdown behavior that could affect worker stability and server load if misconfigured.
> 
> **Overview**
> Adds **worker heartbeat support**: the SDK now periodically reports per-worker slot/poller stats, sticky cache counters, plugin names, and host CPU/memory usage to the server via `RecordWorkerHeartbeat`, and sends a final `ShutdownWorker` RPC (with optional sticky queue) on worker stop.
> 
> Introduces a shared per-namespace heartbeat manager and in-memory metric capture (`heartbeatMetricsHandler`), plumbs poll-success timestamps from workflow/activity/nexus pollers, and adds `client.Options.WorkerHeartbeatInterval` with capability gating via `DescribeNamespace`.
> 
> Refactors/migrates resource tuning: removes `contrib/resourcetuner` by moving the resource-based tuner/controller and PID logic into `internal` and re-exporting from `worker`, requiring a `SysInfoProvider`; adds new `contrib/sysinfo` (gopsutil + Linux cgroups) as the recommended provider and updates integration/tests accordingly (including new `worker_heartbeat_test.go`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit da40521420775adea477300718eb31967721d923. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -2243,9 +2243,9 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 		var prevNexusProcessed, prevNexusFailed int64
 
 		populateOpts := &populateHeartbeatOptions{
-			workflowPollerBehavior:     options.WorkflowTaskPollerBehavior,
-			activityPollerBehavior:     options.ActivityTaskPollerBehavior,
-			nexusPollerBehavior:        options.NexusTaskPollerBehavior,
+			workflowPollerBehavior:     workerParams.WorkflowTaskPollerBehavior,
+			activityPollerBehavior:     workerParams.ActivityTaskPollerBehavior,
+			nexusPollerBehavior:        workerParams.NexusTaskPollerBehavior,
 			prevWorkflowProcessed:      &prevWorkflowProcessed,
 			prevWorkflowFailed:         &prevWorkflowFailed,
 			prevActivityProcessed:      &prevActivityProcessed,
--- a/internal/internal_worker_heartbeat.go
+++ b/internal/internal_worker_heartbeat.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"context"
 	"fmt"
+	ilog "go.temporal.io/sdk/internal/log"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -26,6 +27,9 @@ type heartbeatManager struct {
 
 // newHeartbeatManager creates a new heartbeatManager.
 func newHeartbeatManager(client *WorkflowClient, interval time.Duration, logger log.Logger) *heartbeatManager {
+	if logger == nil {
+		logger = ilog.NewDefaultLogger()
+	}
 	return &heartbeatManager{
 		client:   client,
 		interval: interval,
@@ -82,7 +86,9 @@ func (m *heartbeatManager) getOrCreateSharedNamespaceWorker(worker *AggregatedWo
 		return nil, fmt.Errorf("failed to get namespace capabilities: %w", err)
 	}
 	if !capabilities.GetWorkerHeartbeats() {
-		m.logger.Debug("Worker heartbeating configured, but server version does not support it.")
+		if m.logger != nil {
+			m.logger.Debug("Worker heartbeating configured, but server version does not support it.")
+		}
 		return nil, nil
 	}
 	namespace := worker.executionParams.Namespace
--- a/internal/cmd/build/main.go
+++ b/internal/cmd/build/main.go
@@ -162,7 +162,8 @@ func (b *builder) integrationTest() error {
 				"--dynamic-config-value", `component.nexusoperations.callback.endpoint.template="http://localhost:7243/namespaces/{{.NamespaceName}}/nexus/callback"`,
 				"--dynamic-config-value", "frontend.WorkerHeartbeatsEnabled=true",
 				"--dynamic-config-value", "frontend.ListWorkersEnabled=true",
-			}})
+			},
+		})
 		if err != nil {
 			return fmt.Errorf("failed starting dev server: %w", err)
 		}


worker/hostmetrics/hostmetrics.go
```diff
@@ -0,0 +1,121 @@
...
+package hostmetrics
+
+import (
+	"context"
+	"runtime"
+	"sync"
+	"time"
+
+	"github.com/shirou/gopsutil/v4/cpu"
+	"github.com/shirou/gopsutil/v4/mem"
```
cretz: Do we have to bring in new dependencies as part of this project? The whole reason `resourcetuner` is a contrib library is so we don't force every user of this library to have these dependencies they don't really need.

Is it possible to not capture this information by default and either make it opt-in or and/or let users monitor their own resource utilization? I'd argue this is more information than users really need, especially by default if it forces a dependency on everyone whether they need this stuff or not.

internal/common/metrics/heartbeat_handler.go
cretz, line 1: Some high-level thoughts:

* I don't think this needs to be in this package, just put it in the internal package
* Arguably could use a `map[heartbeatMetric]*atomic.Uint64`, but if you must use explicit fields, should be in sub-struct that's easier to ref on with-tags (and don't need `atomic.Uint64` to be a pointer if you use explicit fields, but a map is clearer IMO)...also, if using a map, no need to make those last poll ones as signed unless you really think they'll go negative
* You should have a method on this struct that mutates `WorkerHeartbeat` proto instead of divorced code building it
* For code simplicity reasons, have you considered just a capturing all metrics and then at heartbeat build time taking out what you want?

internal/common/metrics/heartbeat_handler.go
```diff
@@ -0,0 +1,519 @@
...
+	RecordWorkflowPollSuccess()
+	RecordWorkflowStickyPollSuccess()
+	RecordActivityPollSuccess()
+	RecordNexusPollSuccess()
+}
+
+// RecordPollSuccess records a successful poll time if the handler supports it.
+// pollerType should be one of PollerTypeWorkflowTask, PollerTypeWorkflowStickyTask,
+// PollerTypeActivityTask, or PollerTypeNexusTask.
+func RecordPollSuccess(h Handler, pollerType string) {
```
cretz: I wonder if this should be divorced from metrics handler. Specifically, I wonder if the heartbeat metrics stuff has a handler for some metrics, and then `*HeartbeatMetricsHandler` is set on pollers specifically just for recording poll success. If you must keep this, why even have the interface instead of just type asserting that it is a `*HeartbeatMetricsHandler`?

internal/common/metrics/heartbeat_handler.go
```diff
@@ -0,0 +1,519 @@
...
+	workerType := h.workerType
+	if wt, ok := tags[WorkerTypeTagName]; ok {
+		workerType = wt
+	}
+
+	// Track the poller type if present in tags
+	pollerType := h.pollerType
+	if pt, ok := tags[PollerTypeTagName]; ok {
+		pollerType = pt
+	}
```
cretz: I wonder if these two values should not be extracted from `WithTags` but be done more explicitly when they are set which is only once per worker/poller. Not a big deal, but it may clean up some of this abstraction if the worker/poller type specific metrics were in their own struct maybe? And/or a worker or poller type was required to get a handler out here? I can fashion some ideas if needed...

contrib/resourcetuner/go.mod
```diff
@@ -5,15 +5,14 @@ go 1.23.0
 toolchain go1.23.6
```
Sushisource, line 5: There's probably no reason for this to be a separate package now that we take the gopsutil dependency in the SDK - but I guess moving it in would possibly break people's dependencies so no harm in keeping it if that's the case.

internal/common/metrics/heartbeat_handler.go
```diff
@@ -0,0 +1,519 @@
...
+
+	return &HeartbeatMetricsHandler{
```
Sushisource: Can't most of this just be done with?

```
copy := *h
h.underlying = ... yada yada
return &copy
```

internal/internal_worker.go
```diff
@@ -1468,9 +1506,74 @@ func (aw *AggregatedWorker) Stop() {
...
+	_, err := aw.client.workflowService.ShutdownWorker(grpcCtx, &workflowservice.ShutdownWorkerRequest{
+		Namespace:       aw.executionParams.Namespace,
+		StickyTaskQueue: stickyTaskQueue,
+		Identity:        aw.executionParams.Identity,
+		Reason:          "graceful shutdown",
+		WorkerHeartbeat: heartbeat,
+	})
+
+	// Ignore unimplemented (server doesn't support it) and unavailable (server shutting down)
+	if _, isUnimplemented := err.(*serviceerror.Unimplemented); isUnimplemented {
```
Sushisource, line 1528: Doesn't look like this is ignoring unavailable?

internal/internal_workers_test.go
```diff
@@ -76,7 +76,6 @@ func (s *WorkersTestSuite) TestWorkflowWorker() {
 	s.service.EXPECT().DescribeNamespace(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil)
 	s.service.EXPECT().PollWorkflowTaskQueue(gomock.Any(), gomock.Any(), gomock.Any()).Return(&workflowservice.PollWorkflowTaskQueueResponse{}, nil).AnyTimes()
 	s.service.EXPECT().RespondWorkflowTaskCompleted(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil).AnyTimes()
-	s.service.EXPECT().ShutdownWorker(gomock.Any(), gomock.Any(), gomock.Any()).Return(&workflowservice.ShutdownWorkerResponse{}, nil).Times(1)
```
Sushisource, line 79: These ones that went away seem wrong. We should still be calling shutdown worker in cases where we were previously

internal/internal_worker.go
```diff
@@ -2145,19 +2261,190 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
...
+		pluginInfos := collectPluginInfos(client.clientPluginNames, plugins)
+
+		var prevWorkflowProcessed, prevWorkflowFailed int64
+		var prevActivityProcessed, prevActivityFailed int64
+		var prevLocalActivityProcessed, prevLocalActivityFailed int64
+		var prevNexusProcessed, prevNexusFailed int64
+
+		heartbeatCallback = func() *workerpb.WorkerHeartbeat {
+			heartbeatTime := time.Now()
+			elapsedSinceLastHeartbeat := previousHeartbeatTime.Sub(heartbeatTime)
```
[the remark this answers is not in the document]
Sushisource: Seems right

internal/internal_worker.go
```diff
@@ -2145,19 +2261,190 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
...
+					BuildId:        options.DeploymentOptions.Version.BuildID,
+				}
+			}
+
+			var workflowTaskSlotsInfo *workerpb.WorkerSlotsInfo
+			var activityTaskSlotsInfo *workerpb.WorkerSlotsInfo
+			var localActivitySlotsInfo *workerpb.WorkerSlotsInfo
+			var nexusTaskSlotsInfo *workerpb.WorkerSlotsInfo
+
+			if aw.workerHeartbeatManager.heartbeatMetrics != nil {
```
Sushisource: These checks for `heartbeatMetrics != nil` are weirdly scattered in multiple places instead of just one block

internal/internal_worker.go
```diff
@@ -2476,3 +2763,83 @@ func workerDeploymentVersionFromProtoOrString(wd *deploymentpb.WorkerDeploymentV
...
+	}
+	for _, plugin := range workerPlugins {
+		set[plugin.Name()] = struct{}{}
+	}
+
+	result := make([]*workerpb.PluginInfo, 0, len(set))
+	for name := range set {
+		result = append(result, &workerpb.PluginInfo{Name: name})
+	}
+	return result
```
Sushisource, line 2696: I think we sorted this in core?

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,143 @@
...
+		NexusTaskPollerBehavior: NewPollerBehaviorSimpleMaximum(PollerBehaviorSimpleMaximumOptions{MaximumNumberOfPollers: 1}),
+	}
+
+	nw, err := newNexusWorker(nexusWorkerOptions{
+		executionParameters: params,
+		client:              hw.client,
+		workflowService:     hw.client.workflowService,
+	})
+
+	return nw, nil
```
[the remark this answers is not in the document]
Sushisource: both of these seem right too. (for it's other comment:) Even though we don't have any ops implemented yet we don't want old workers to crash once we do start having them. Even though we need to make sure workers advertise which operations they are capable of - we should probably have a catch-all handler that responds to the nexus task with failure as unsupported

I dunno if we missed that in Core too

test/worker_heartbeat_test.go
```diff
@@ -0,0 +1,997 @@
...
+	// Wait for activity to start
+	select {
+	case <-blockingActivityStarted:
+		ts.T().Log("Activity started")
+	case <-time.After(5 * time.Second):
+		ts.Fail("Timeout waiting for activity to start")
+	}
+
+	// Wait for heartbeat to capture the in-flight activity
+	time.Sleep(100 * time.Millisecond)
```
Sushisource: Prefer an `eventually` with one of the initial assertions

test/worker_heartbeat_test.go
```diff
@@ -0,0 +1,997 @@
...
+	ts.Equal(int32(0), nexusTaskSlots.TotalProcessedTasks)
+	ts.Equal(int32(1000), nexusTaskSlots.CurrentAvailableSlots)
+	ts.Equal(int32(0), nexusTaskSlots.CurrentUsedSlots)
+	ts.Equal("Fixed", nexusTaskSlots.SlotSupplierKind)
+	localActivityTaskSlots = workerInfo.LocalActivitySlotsInfo
+	ts.Equal(int32(0), localActivityTaskSlots.TotalProcessedTasks)
+	ts.Equal(int32(1000), localActivityTaskSlots.CurrentAvailableSlots)
+	ts.Equal(int32(0), localActivityTaskSlots.CurrentUsedSlots)
+	ts.Equal("Fixed", localActivityTaskSlots.SlotSupplierKind)
+	//
```
Sushisource: ?

test/worker_heartbeat_test.go
```diff
@@ -0,0 +1,997 @@
...
+	// Timestamp validations - second heartbeat check (after shutdown)
+	// StartTime should be unchanged
+	ts.Equal(firstStartTime, workerInfo.StartTime.AsTime(),
+		"StartTime should not change between heartbeats")
+
+	// HeartbeatTime should have advanced
+	ts.True(workerInfo.HeartbeatTime.AsTime().After(firstHeartbeatTime),
+		"HeartbeatTime should advance between heartbeats")
+
+	fmt.Println("aa")
```
Sushisource: Classic Andrew comments

test/worker_heartbeat_test.go
```diff
@@ -0,0 +1,997 @@
...
+				BuildID:        "test_build_id",
+			},
+			DefaultVersioningBehavior: internal.VersioningBehaviorAutoUpgrade,
+		},
+	})
+	w.RegisterWorkflow(simpleWorkflow)
+	ts.NoError(w.Start())
+	defer w.Stop()
+
+	// Wait for heartbeat to be sent
```
Sushisource, line 286: Same about `eventually`

test/worker_heartbeat_test.go
```diff
@@ -0,0 +1,997 @@
...
+			run, err := ts.client.ExecuteWorkflow(ctx, workflowOptions, simpleWorkflow)
+			ts.NoError(err)
+			err = run.Get(ctx, nil)
+			ts.NoError(err)
+		}(i, tq)
+	}
+	wg.Wait()
+
+	// Wait for heartbeats
+	time.Sleep(150 * time.Millisecond)
```
Sushisource: Same. I won't leave more, all of them should work that way

internal/internal_workflow_client.go
```diff
@@ -1393,6 +1429,89 @@ func (wc *WorkflowClient) WorkerDeploymentClient() WorkerDeploymentClient {
...
+	defer cancel()
+	resp, err := wc.workflowService.RecordWorkerHeartbeat(grpcCtx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+func (wc *WorkflowClient) getOrCreateHeartbeatWorker(namespace string) (*sharedNamespaceWorker, error) {
```
cretz: This is a bit confusing to follow that a client get or creates this worker, but a caller stops it at some point, and a caller registers a callback but a client unregisters one. It's a bit inconsistent.

I wonder if you should consider simplifying and encapsulating this stuff. Just have a "heartbeat manager" on a client that handles all aspects of registering callbacks, starting/stopping the shared worker, collecting metrics, etc. The client can have a simple "register worker" and "unregister worker" aspects that can also provide/mutate things needed like metrics handlers. And just put this all of this in the `internal_worker_heartbeat.go` file.

internal/tuning.go
```diff
@@ -130,6 +131,26 @@ type SlotSupplier interface {
 	MaxSlots() int
 }
 
+// SlotSupplierKinder is an optional interface that slot suppliers can implement to provide
+// a custom kind/type name. If not implemented, getSlotSupplierKind will use reflection.
+type SlotSupplierKinder interface {
```
cretz: Probably doesn't need to be an interface, can just `s.(interface { Kind() string})` below, though if you are going to make it a user-facing interface, should properly alias it in public package. Also, arguably could call the method `SlotSupplierKind`, but not a big deal.

worker/worker.go
```diff
@@ -236,6 +236,13 @@ type (
 
 	// ReplayWorkflowHistoryOptions are options for replaying a workflow.
 	ReplayWorkflowHistoryOptions = internal.ReplayWorkflowHistoryOptions
+
+	// HostMetricsProvider provides host-level CPU and memory metrics for worker heartbeats.
+	// Implement this interface to provide custom metrics collection, or use the default
+	// implementation provided by the SDK in the worker/hostmetrics package.
+	//
+	// NOTE: Experimental
+	HostMetricsProvider = internal.HostMetricsProvider
```
cretz: If this is meant to be implemented by a tuner, IMO docs should say as much and arguably be named that way

internal/internal_workflow_client.go
yuandrew: Re-arranged thing so client only calls `Un/RegisterWorker`, and the heartbeat callback creation still happens on the `AggregatedWorker` creation.

Hmm, is it confusing that the client and worker both have  a `HeartbeatManager` that server different purposes? I feel like this could only get a little confusing when you try to globally search for `HeartbeatManager`.

internal/tuning.go
yuandrew: oh nice, I didn't realize i could do that. Took out this interface, `getSlotSupplierKind`'s doc comment explains that `Kind()` will be used if available.

worker/worker.go
yuandrew: renamed to `TunerHostMetricsProvider` and added docs

internal/common/metrics/heartbeat_handler.go
yuandrew: Moved and changed to use a map. Also created the method `HeartbeatMetricsHandler.PopulateHeartbeat`

> For code simplicity reasons, have you considered just a capturing all metrics and then at heartbeat build time taking out what you want?

Not until now, but I think I still prefer being explicit. We would need additional mutexes for new map entries, and might as well only use memory for what we're tracking, not that it's that much more memory

worker/hostmetrics/hostmetrics.go
yuandrew: Went with trying to vendor `gopsutil`, although we still need to take in a dep on `purego`. Not worth writing and maintaining an assembly trampoline IMO, as the alternative to avoiding `purego`.

If we're okay with this, I will move the `contrib/resourcetuner` package into the SDK code itself, but didn't want to go through the migration before we settle on this.

test/worker_heartbeat_test.go
yuandrew: removed

internal/internal_worker.go
yuandrew: oops, fixed

internal/internal_workers_test.go
yuandrew: The level at where `ShutdownWorker` is called went up a level, so now this test shuts down a workflow worker no longer makes the gRPC request, it's made at the `AggregatedWorker` level

internal/internal_worker.go
yuandrew: not sure what issue this is referring to

internal/common/metrics/heartbeat_handler.go
yuandrew: Is there a specific concern with divorcing from metrics handler? Today, heartbeat metrics should be a transparent layer, so passing in a specific *HeartbeatMetricsHandler doesn't feel too different than just using the metrics handler itself, maybe except being a little clarity, at the cost of passing an object (*HeartbeatMetricsHandler) that we already have access to from `wtp.metricsHandler`.

Thanks for the callout on the interface being unneccesary

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.23.0
 toolchain go1.23.6
 
 require (
+	github.com/ebitengine/purego v0.9.1
```
cretz: I am a bit disappointed to force this dependency on all of our users, even ones that don't care about host metrics in the UI. From the README of that project:

> This is beta software so expect bugs and potentially API breaking changes but each release will be tagged to avoid breaking people's code

What that means is that if a user is using this dependency in their project too, since it's zero ver, unless they are on the same version as us, they very possibly will fail compilation. From https://go.dev/ref/mod#versions:

> A version is considered unstable if its major version is 0 or it has a pre-release suffix. Unstable versions are not subject to compatibility requirements. For example, v0.2.0 may not be compatible with v0.1.0, and v1.5.0-beta may not be compatible with v1.5.0.

I cannot tell from the project's release notes how frequently it makes breaking changes, but they are totally allowed to and have even noted that they may. Maybe a sample project can be created with someone using a different version of this library (or a dependency thereof) to see if this is a problem. I see thousands of dependents of this library [here](https://github.com/ebitengine/purego/network/dependents?package_id=UGFja2FnZS0zMTY5NTUzNTk4), some of which on `v0.8.4`, I wonder if a user of ours on that version will now break if they use our SDK. I guess it depends on if they used anything that broke between https://github.com/ebitengine/purego/compare/v0.8.4...v0.9.0. Regardless, it is likely going to happen one day where some user will want to use newer version of this dep, and some will want to use older, and they will be incompatible and we will have to choose.

I know we put forth a lot of effort to try and get host metrics to the server here, but I am afraid we're so hyper-focused on this one very small part of the SDK that we don't realize it affects everyone, even those that don't care about this one small part. I suppose we can release and hope nobody complains (library users often don't, they sigh and work around), but in general I think our posture of new dependencies should be more compatible with our users.

contrib/resourcetuner/go.mod
cretz: We have never made a release of this library before. Assuming we can get dependencies figured out, I think it'd be acceptable to move all of this verbatim and deprecate it here for a couple of releases before just deleting this dir altogether (with accompanying release notes and such of course).

internal/tuning.go
```diff
@@ -130,6 +131,19 @@ type SlotSupplier interface {
 	MaxSlots() int
 }
 
+// getSlotSupplierKind returns the kind/type name of a slot supplier. If the supplier implements
+// a Kind() string method, it uses that. Otherwise, it falls back to reflection on the type name.
+func getSlotSupplierKind(s SlotSupplier) string {
```
cretz, line 133: I am curious, how are we populating this value in other SDKs? All SDKs have a concept of custom slot suppliers, but none of them have a concept of providing a kind that I can see. If this is for internal use only, I think a simple type switch can probably do it for the 3 kinds we have.

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/nexus-rpc/sdk-go/nexus"
+	workerpb "go.temporal.io/api/worker/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/log"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// HeartbeatManager manages heartbeat workers across namespaces for a client.
+type HeartbeatManager struct {
```
cretz: Does this or any of the other code/methods here need to be exported/public (even though we know `internal` isn't really public)?

internal/common/metrics/heartbeat_handler.go
cretz: > Not until now, but I think I still prefer being explicit

That sure is a lot of copied/specific code in `internal_worker_heartbeat_metrics.go`

internal/internal_worker_heartbeat_metrics.go
```diff
@@ -0,0 +1,559 @@
...
+	workerpb "go.temporal.io/api/worker/v1"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"go.temporal.io/sdk/internal/common/metrics"
+)
+
+type heartbeatMetric int
+
+const (
+	metricStickyCacheHit heartbeatMetric = iota
```
cretz: Keeping all of these as integer indexes is a bit rough and defeats the purpose of a map IMO. I don't think keeping a map (or `sync.Map`) of metrics that we care about based on string names as we see them is that big of an issue, but up to you I suppose. This is one pretty big file to just capture metrics.

Part of me wonders if you should just capture all metrics indiscriminately and then at heartbeat time deduce from what you captured what goes where.

internal/internal_worker_heartbeat_metrics.go
```diff
@@ -0,0 +1,559 @@
...
+
+	return &workerpb.WorkerPollerInfo{
+		CurrentPollers:         currentPollers,
+		LastSuccessfulPollTime: timestamppb.New(lastSuccessfulPollTime),
+		IsAutoscaling:          isAutoscaling,
+	}
+}
+
+// GetStickyCacheHit returns the total number of sticky cache hits.
+func (h *HeartbeatMetricsHandler) GetStickyCacheHit() int32 {
```
cretz: What are these getters used for?

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+		logger:   logger,
+		workers:  make(map[string]*sharedNamespaceWorker),
+	}
+}
+
+// RegisterWorker registers a worker's heartbeat callback with the shared heartbeat worker for the namespace.
+func (m *HeartbeatManager) RegisterWorker(
+	namespace string,
+	workerInstanceKey string,
+	callback func() *workerpb.WorkerHeartbeat,
```
cretz: Is all of this abstraction around callbacks and nested funcs and such needed? This is all internal code. Can this just accept the `*AggregatedWorker` and store that and call a method on that to get metrics and what not? I find the setting a callback on a client `HeartbeatManager` that is a method on `workerHeartbeatManager` that has nested funcs that call metrics populating things to be a quite complicated dance. Can you simplify?

internal/internal_worker.go
```diff
@@ -1468,9 +1505,74 @@ func (aw *AggregatedWorker) Stop() {
...
+	}
+	return aw.client.heartbeatManager.RegisterWorker(
+		aw.executionParams.Namespace,
+		aw.workerInstanceKey,
+		aw.workerHeartbeatManager.heartbeatCallback,
+	)
+}
+
+func (aw *AggregatedWorker) unregisterHeartbeatWorker() {
+	if aw.client.heartbeatManager == nil || aw.workerHeartbeatManager == nil {
```
cretz: In what situations is `aw.woerkerHeartbeatManager` nil here but not nil in the `registerHeartbeatWorker` equivalent?

internal/internal_worker.go
```diff
@@ -2145,19 +2260,128 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
...
+			PrevWorkflowFailed:         &prevWorkflowFailed,
+			PrevActivityProcessed:      &prevActivityProcessed,
+			PrevActivityFailed:         &prevActivityFailed,
+			PrevLocalActivityProcessed: &prevLocalActivityProcessed,
+			PrevLocalActivityFailed:    &prevLocalActivityFailed,
+			PrevNexusProcessed:         &prevNexusProcessed,
+			PrevNexusFailed:            &prevNexusFailed,
+		}
+
+		heartbeatCallback = func() *workerpb.WorkerHeartbeat {
```
cretz, line 2270: The cursor bot has a good point. Can you confirm whether this function is guaranteed to never be run concurrently? Even in shutdown situations?

internal/internal_worker.go
cretz: Sorting of the plugins by name

internal/internal_worker.go
```diff
@@ -2476,3 +2700,37 @@ func workerDeploymentVersionFromProtoOrString(wd *deploymentpb.WorkerDeploymentV
...
+		set[name] = struct{}{}
+	}
+	for _, plugin := range workerPlugins {
+		set[plugin.Name()] = struct{}{}
+	}
+
+	result := make([]*workerpb.PluginInfo, 0, len(set))
+	for name := range set {
+		result = append(result, &workerpb.PluginInfo{Name: name})
+	}
```
cretz: Pedantic, but can just make the slice at the same time during the original iterations, just using the map to prevent a second slice write. Don't have to come back and iterate map again.

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	hw, ok := m.workers[namespace]
+	if !ok {
+		capabilities, err := m.client.loadNamespaceCapabilities(context.Background())
```
cretz: Can we not put this (or any blocking code) under a lock?

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to create nexus worker for heartbeating: %w", err)
+		}
+		hw.nexusWorker = nexusWorker
+
+		m.workers[namespace] = hw
+		go hw.run()
+	}
+
+	hw.mu.Lock()
```
cretz: Curious how this doesn't cause a deadlock with the `h.mu.Lock()` at the start of this function. I guess locks may be reentrant, but it's not necessary to re-lock here unless some strange unlocking is happening within a method elsewhere. Same for unregister code (and maybe others) which does this same thing.

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+	return nw, err
+}
+
+func (hw *sharedNamespaceWorker) run() {
+	defer close(hw.stoppedC)
+
+	hw.started.Store(true)
+
+	if err := hw.nexusWorker.Start(); err != nil {
+		return
```
cretz: Swallowed error

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+
+	return nw, err
+}
+
+func (hw *sharedNamespaceWorker) run() {
+	defer close(hw.stoppedC)
+
+	hw.started.Store(true)
+
+	if err := hw.nexusWorker.Start(); err != nil {
```
cretz: Can you confirm for me on regular worker start (that includes these sub-workers and this shared worker) how many calls to `DescribeNamespace` we are making?

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+
+	mu        sync.RWMutex
+	callbacks map[string]func() *workerpb.WorkerHeartbeat // workerInstanceKey -> callback
+
+	stopC    chan struct{}
+	stoppedC chan struct{}
+	started  atomic.Bool
+}
+
+func (hw *sharedNamespaceWorker) createNexusWorker() (*nexusWorker, error) {
```
cretz: Can you help me understand this a bit, why are we creating a nexus worker here with no Nexus services? Is there any value in a Nexus worker that can't do anything? Are we just adding more poll calls against our system for all of our users unnecessarily? Should this be disabled until there is a Nexus call to be made?

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+		return
+	}
+
+	heartbeats := make([]*workerpb.WorkerHeartbeat, 0, len(callbacks))
+	for _, cb := range callbacks {
+		hb := cb()
+		heartbeats = append(heartbeats, hb)
+	}
+
+	if len(heartbeats) == 0 {
```
cretz: How is this possible?

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
cretz: This is the default timeout for our calls normally which is part of `newGRPCContext` that occurs inside that `RecordWorkerHeartbeat` call on the client

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,229 @@
...
+
+	_, err := hw.client.RecordWorkerHeartbeat(ctx, &workflowservice.RecordWorkerHeartbeatRequest{
+		Namespace:       hw.namespace,
+		WorkerHeartbeat: heartbeats,
+	})
+
+	if err != nil {
+		if status.Code(err) == codes.Unimplemented {
+			// Server doesn't support heartbeats, shutdown worker
+			hw.stop()
```
cretz: Deadlock possibility here if I'm reading the code right. You call stop, stop closes `stopC` and waits on `stoppedC`, but `stoppedC` can't be closed by `run` because it's still waiting on this `sendHeartbeats` call to complete. I would recommend using an error on this function to report back to `run` that it should stop.

internal/internal_workflow_client.go
```diff
@@ -1366,6 +1373,33 @@ func (wc *WorkflowClient) loadCapabilities(ctx context.Context) (*workflowservic
 	return capabilities, nil
 }
 
+// Get namespace capabilities, lazily fetching from server if not already obtained.
+func (wc *WorkflowClient) loadNamespaceCapabilities(ctx context.Context) (*namespacepb.NamespaceInfo_Capabilities, error) {
```
cretz: We already call describe namespace on every worker start in every language, can we reuse that?

internal/internal_workflow_client.go
```diff
@@ -1393,6 +1427,21 @@ func (wc *WorkflowClient) WorkerDeploymentClient() WorkerDeploymentClient {
 	}
 }
 
+func (wc *WorkflowClient) RecordWorkerHeartbeat(ctx context.Context, request *workflowservice.RecordWorkerHeartbeatRequest) (*workflowservice.RecordWorkerHeartbeatResponse, error) {
```
cretz: Mentioned for other things, but will also mention here, does this need to be public?

internal/common/metrics/heartbeat_handler.go
cretz: The main struggle for me is that if you hide everything behind metrics handler interface as if you were a user, it's hard for code readers in the SDK that may make future changes to recognize metrics handler is not just about user metrics, it's leveraged for internal utilities too. But not a big deal to leave as is.

internal/internal_workflow_client.go
cretz: > is it confusing that the client and worker both have a HeartbeatManager that server different purposes?

Yes

go.mod
yuandrew: Good points. We decided we will scrap this vendoring effort, and continue to pull in from `contrib/resourcetuner` so we don't introduce a new dep into the SDK

internal/internal_worker.go
yuandrew: derp, I can't read... some reason I read "sorted this out", was confused what "this" was

contrib/sysinfo/cgroups.go
```diff
@@ -1,137 +0,0 @@
...
-
-type cGroupInfoImpl struct {
-	lastCGroupMemStat *stats.MemoryStat
-	cgroupCpuCalc     cgroupCpuCalc
-}
-
-func (p *cGroupInfoImpl) Update() (bool, error) {
-	err := p.updateCGroupStats()
-	// Stop updates if not in a container. No need to return the error and log it.
-	if !errors.Is(err, fs.ErrNotExist) {
```
yuandrew, line 30: TODO: This check seems backwards

go.mod
cretz: I do think there is a better option here...

Before (and instead of) releasing `resourcetuner`, we can have a `sysinfo` module instead that has a single `DefaultSysInfoProvider()` function. And then we can have client accept `SysInfoProvider` option+interface and move all the resource tuner stuff in `worker`, and if a worker tries to be created with a resource tuner but the client has `SysInfoProvider` as `nil`, we panic, otherwise the resource tuner uses that. And the `resourcetuner` package is deprecated, soon to be deleted.

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,143 @@
...
+		TaskQueue:               hw.taskQueue,
+		Tuner:                   tuner,
+		NexusTaskPollerBehavior: NewPollerBehaviorSimpleMaximum(PollerBehaviorSimpleMaximumOptions{MaximumNumberOfPollers: 1}),
+	}
+
+	nw, err := newNexusWorker(nexusWorkerOptions{
+		executionParameters: params,
+		client:              hw.client,
+		workflowService:     hw.client.workflowService,
+	})
```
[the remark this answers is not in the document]
yuandrew: removed nexus worker for now, since the worker isn't being used yet

internal/common/metrics/heartbeat_handler.go
yuandrew: I still lean towards having this method, primarily for code simplicity sake. I've renamed to `recordPollSuccessIfHeartbeat`, does that help make it more distinct we're doing this for heartbeat, not user metrics?

internal/common/metrics/heartbeat_handler.go
yuandrew: Added an explicit `forWorker` and `forPoller` and made `WithTags` a passthrough

internal/internal_workflow_client.go
yuandrew: Removed the `AggregatedWorker.HeartbeatManager

internal/internal_worker.go
```diff
@@ -2145,19 +2260,128 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
...
+			PrevLocalActivityProcessed: &prevLocalActivityProcessed,
+			PrevLocalActivityFailed:    &prevLocalActivityFailed,
+			PrevNexusProcessed:         &prevNexusProcessed,
+			PrevNexusFailed:            &prevNexusFailed,
+		}
+
+		heartbeatCallback = func() *workerpb.WorkerHeartbeat {
+			heartbeatTime := time.Now()
+			elapsedSinceLastHeartbeat := heartbeatTime.Sub(previousHeartbeatTime)
+			previousHeartbeatTime = heartbeatTime
```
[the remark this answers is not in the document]
yuandrew: Fixed, added a mutex

internal/internal_workflow_client.go
```diff
@@ -1366,6 +1373,33 @@ func (wc *WorkflowClient) loadCapabilities(ctx context.Context) (*workflowservic
...
+		return capabilities, nil
+	}
+
+	grpcCtx, cancel := newGRPCContext(ctx, grpcTimeout(wc.getSystemInfoTimeout))
+	defer cancel()
+	resp, err := wc.workflowService.DescribeNamespace(grpcCtx, &workflowservice.DescribeNamespaceRequest{Namespace: wc.namespace})
+	if _, isUnimplemented := err.(*serviceerror.Unimplemented); err != nil && !isUnimplemented {
+		return nil, fmt.Errorf("failed reaching server: %w", err)
+	}
+	if resp != nil && resp.NamespaceInfo.Capabilities != nil {
```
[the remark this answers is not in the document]
yuandrew: fixed

go.mod
yuandrew: Decided to move `resourcetuner` into `worker`, and now a new `hostinfo` package was created, to more tightly encapsulate the gopsutil dep

internal/tuning.go
yuandrew: Hmmm good callout, don't remember why I wanted to make this more enhanced and return the name of the custom kind.

internal/internal_worker_heartbeat.go
yuandrew: thanks, made private

internal/internal_worker_heartbeat_metrics.go
yuandrew: removed the iota

internal/internal_worker_heartbeat_metrics.go
yuandrew: AI slop 😔 my bad, I didn't catch this bc goland registers the docstring as a usage, didn't notice that it wasn't used anywhere real

internal/internal_worker_heartbeat.go
yuandrew: yes, I'm afraid I'm functioning too much in a Rust lense and translating things too literally over to Go. Good suggestion on passing `*AggregatedWorker` into register/unregister

internal/internal_worker.go
yuandrew: It should never, but removed `aw.workerHeartbeatManager` altogether

internal/internal_worker.go
yuandrew: It could be, but added a mutex to make it concurrency safe

internal/internal_worker_heartbeat.go
yuandrew: changed to only lock when accessing  `m.workers`

internal/internal_worker_heartbeat.go
yuandrew: I don't believe mutexes in go are re-entrant, these are different locks, one is `hw`, the other is `h`

internal/internal_worker_heartbeat.go
yuandrew: I've consolidated all into `loadNamespaceCapabilities`, and they all share the same cache, so this should now only be called once, instead of the 3-4 times previously

internal/internal_worker_heartbeat.go
yuandrew: hmm, that's a good callout, today this is idle. I didn't realize we're just using the client directly to make the gRPC call today, so the nexus worker itself isn't doing anything until the first worker command is implemented.. 

removed for now

internal/internal_worker_heartbeat.go
yuandrew: ahh didn't realize, thanks, removed

internal/internal_worker_heartbeat.go
yuandrew: oops, thanks for pointing out, added `error` return types to avoid deadlock

internal/internal_workflow_client.go
yuandrew: I've consolidated all into loadNamespaceCapabilities with a cache, should only call `DescriebNamespace` once

contrib/sysinfo/cgroups.go
yuandrew: fixed in `contrib/hostinfo/cgroups.go`

contrib/sysinfo/sysinfo.go
cretz, line 1: Pedantic, but IMO we should call this module `sysinfo`

contrib/sysinfo/sysinfo.go
```diff
@@ -0,0 +1,109 @@
...
+// NewSystemInfoSupplier creates a SystemInfoSupplier using gopsutil.
+// Supports cgroup metrics in containerized Linux environments.
+func NewSystemInfoSupplier() worker.SystemInfoSupplier {
+	return &psUtilSystemInfoSupplier{
+		cGroupInfo: newCGroupInfo(),
+	}
+}
+
+// NewResourceBasedTuner creates a resource-based tuner with gopsutil-based system info.
+func NewResourceBasedTuner(opts worker.ResourceBasedTunerOptions) (worker.WorkerTuner, error) {
```
cretz: Not sure this shortcut has that much value, but not a big deal

contrib/hostinfo/hostinfo.go
```diff
@@ -0,0 +1,109 @@
...
+	"time"
+
+	"github.com/shirou/gopsutil/v4/cpu"
+	"github.com/shirou/gopsutil/v4/mem"
+	"go.temporal.io/sdk/worker"
+)
+
+// NewSystemInfoSupplier creates a SystemInfoSupplier using gopsutil.
+// Supports cgroup metrics in containerized Linux environments.
+func NewSystemInfoSupplier() worker.SystemInfoSupplier {
```
cretz: Pedantic, but would recommend calling this `SysInfoProvider` ("sysinfo" to match module name, "provider" to not confuse with other uses of the term "supplier")

cretz: I wonder if this should be `DefaultSysInfoProvider()` and always return the same value for all calls

internal/common/metrics/heartbeat_handler.go
cretz: I think the name change goes wrong in the other direction. We should be recording poll successes regardless of metrics handler types. What I was trying to say is that heartbeats don't need to use metrics handlers for this data since workers already control where it is called. Extracting poll data out of a metrics handler adds for a bunch of confusing logic when you could just call something here that records it more specifically for that use case (and leave metrics handler alone to record it as it always has, no need to intercept).

But if you still want to extract poll metrics from the metrics handler abstraction, ok, but definitely no need to change the name or only do it if it's a certain metrics handler type.

internal/common/metrics/heartbeat_handler.go
cretz: I didn't mean double up calls on the caller side, I meant you can do what `forWorker` and `forPoller` do inside of `WithTags` (i.e. store more explicit state) and/or make these `forWorker` and `forPoller` calls do the `WithTags` themselves. It doesn't make sense to have two always-consecutive calls both independently copy this thing.

internal/internal_worker_heartbeat_metrics.go
```diff
@@ -0,0 +1,328 @@
...
+
+func (h *heartbeatMetricsHandler) get(key string) int64 {
+	if v, ok := h.metrics.Load(key); ok {
+		return v.(*atomic.Int64).Load()
+	}
+	return 0
+}
+
+// PopulateHeartbeatOptions contains external dependencies needed to populate heartbeat metrics.
+type PopulateHeartbeatOptions struct {
```
cretz: Does this struct and all of its fields need to be public?

internal/internal_worker_heartbeat_metrics.go
```diff
@@ -0,0 +1,328 @@
...
+	if capturedCounters[name] {
+		return &capturingCounter{
+			underlying: underlying,
+			value:      h.getOrCreate(name),
+		}
+	}
+	return underlying
+}
+
+func (h *heartbeatMetricsHandler) Gauge(name string) metrics.Gauge {
```
cretz, line 87: A bit unfortunate that gauge couldn't be generic like counter/timer, but I understand if it's not worth it to make it tag specific for the two that are.

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,185 @@
...
+
+// registerWorker registers a worker's heartbeat callback with the shared heartbeat worker for the namespace.
+func (m *heartbeatManager) registerWorker(
+	worker *AggregatedWorker,
+) error {
+	namespace := worker.executionParams.Namespace
+	m.mu.Lock()
+	hw, ok := m.workers[namespace]
+	m.mu.Unlock()
+	if !ok {
```
cretz: Logic would be clearer IMO if you had a `getOrCreateSharedNamespaceWorker()` method, but not a big deal. Also note for get or create situations like these, sometimes a `sync.OnceValue` is clearer code instead of locks.

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,185 @@
...
+		capabilities, err := m.client.loadNamespaceCapabilities(worker.heartbeatMetrics)
+		if err != nil {
+			return fmt.Errorf("failed to get namespace capabilities: %w", err)
+		}
+		if !capabilities.GetWorkerHeartbeats() {
+			m.logger.Debug("Worker heartbeating configured, but server version does not support it.")
+			return nil
+		}
+
+		newHw := &sharedNamespaceWorker{
```
cretz: There's a race I see you account for where you create this thing that's not used. I think you can totally put all of the logic after loading namespace capabilities under lock. The entire concept of creating this struct should be atomic IMO.

internal/internal_worker_heartbeat.go
cretz: Sorry, got confused reading, thought they were the same lock. If possible, if locks are only for a single field, would recommend naming the lock for the field e.g. `workersMutex` or `callbacksMutex` (but no need, was just my mistake reading)

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,185 @@
...
+
+	stopC    chan struct{}
+	stoppedC chan struct{}
+	started  atomic.Bool
+}
+
+func (hw *sharedNamespaceWorker) run() {
+	defer close(hw.stoppedC)
+
+	hw.started.Store(true)
```
cretz: Arguably could be before the defer, could be a `Swap` and if it returns `true`, panic, but this is pedantic

internal/client.go
```diff
@@ -1144,6 +1158,15 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
 		}
 	}
 
+	var heartbeatInterval time.Duration
+	if options.WorkerHeartbeatInterval == nil {
+		heartbeatInterval = time.Second * 60
+	} else if *options.WorkerHeartbeatInterval == 0 {
+		heartbeatInterval = time.Second * 0
```
cretz:
```suggestion
		heartbeatInterval = 0
```

😁

internal/client.go
```diff
@@ -1144,6 +1158,15 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
...
+	} else {
+		heartbeatInterval = *options.WorkerHeartbeatInterval
```
cretz, line 1220: Pedantic, but validate against a reasonable minimum IMO

internal/common/metrics/heartbeat_handler.go
yuandrew: oops, I had an additional `WithTags` call for `forWorker`. I feel better about this, where we always call `WithTags`, and if heartbeating, we supplementally call `forWorker`/`forPoller` before hand.

Don't want to change the function signature of `WithTags`, and now sure how we can be more explicit with `WithTags` without doing so. And I feel like putting the `WithTags` call inside of `forWorker` makes the caller less clear that it's still calling `WithTags`

internal/internal_worker_heartbeat.go
yuandrew: no that's a good point, single/double letter vars are easy to misread/mistake for each other

contrib/sysinfo/sysinfo.go
yuandrew: sure, renamed

contrib/sysinfo/sysinfo.go
yuandrew: I was originally thinking it was a utility to auto-configure the InfoSupplier, but given it sharing the same name as worker.NewResourceBasedTuner, it's not clear of the extra utility it adds.

Maybe it's better just to have folks specify `InfoSupplier` manually. removed this shortcut

contrib/hostinfo/hostinfo.go
yuandrew: good idea

yuandrew: oh nice, 

having a ref and sync.Once is a newer pattern for me, good idea
```
	sysInfoOnce     sync.Once
	sysInfoInstance *psUtilSystemInfoSupplier
```

internal/internal_worker_heartbeat.go
yuandrew: Added getOrCreateSharedNamespaceWorker(), decided to not use `OnceValue` here

internal/internal_worker_heartbeat.go
yuandrew: agree concept of creating this struct should be atomic, but more importantly, we should only create the new `sharedNamespaceWorker` if it doesn't already exist. 
Moved this `newHw` creation into the `else {}` block

internal/internal_worker_heartbeat.go
yuandrew: Decided to actually move the hw.started.Swap to the caller, before we run `go run()`, to fix the race mentioned in https://github.com/temporalio/sdk-go/pull/2148#discussion_r2740353696

internal/client.go
yuandrew: ooo ty for the reminder, this should be between 1s and 60s. That's the range we agreed on to start, and what Core has

contrib/hostinfo/hostinfo.go
cretz: Can even use `sync.OnceValue` to combine those I suspect

internal/common/metrics/heartbeat_handler.go
yuandrew: Created a separate `pollTimeTracker` and plumbed through separately

contrib/sysinfo/cgroups.go
```diff
@@ -27,7 +27,7 @@ type cGroupInfoImpl struct {
 func (p *cGroupInfoImpl) Update() (bool, error) {
 	err := p.updateCGroupStats()
 	// Stop updates if not in a container. No need to return the error and log it.
-	if !errors.Is(err, fs.ErrNotExist) {
+	if errors.Is(err, fs.ErrNotExist) {
```
yuandrew, line 30: This check seems inverted, I switched it around

internal/internal_workflow_client.go
yuandrew: hmm, actually I'm second guessing this cache decision, can we run into problems down the line adding a dependency onto `WorkflowClient` for all of these different worker types? Since technically `WorkflowClient` is only for workflows. Today we pass in this client to each new worker, but not sure if we want to make that a requirement now, just for this cache.

Alternative could be to only have `workflowWorker` share the same cache for this call

contrib/sysinfo/cgroups.go
cretz: That this wasn't caught before makes me concerned this logic isn't being tested properly (but doesn't have to be part of this PR)

contrib/sysinfo/sysinfo.go
```diff
@@ -0,0 +1,117 @@
...
+})
+
+// SysInfoProvider returns a shared SystemInfoSupplier using gopsutil.
+// Supports cgroup metrics in containerized Linux environments.
+func SysInfoProvider() worker.SystemInfoSupplier {
+	return sysInfoProvider()
+}
+
+// NewResourceBasedTuner creates a resource-based tuner with gopsutil-based system info.
+func NewResourceBasedTuner(opts worker.ResourceBasedTunerOptions) (worker.WorkerTuner, error) {
```
cretz: https://github.com/temporalio/sdk-go/pull/2148#discussion_r2752877199 says you removed this shortcut, but it's still here

worker/tuning.go
```diff
@@ -46,3 +46,79 @@ func NewCompositeTuner(options CompositeTunerOptions) (WorkerTuner, error) {
...
+// Use contrib/sysinfo.SysInfoProvider() for a gopsutil-based implementation,
+// or provide your own.
+type SystemInfoSupplier = internal.SystemInfoSupplier
+
+// SystemInfoContext provides context for SystemInfoSupplier calls.
+type SystemInfoContext = internal.SystemInfoContext
+
+// HasSystemInfoSupplier is an optional interface that SlotSupplier implementations can implement
+// to expose their SystemInfoSupplier.
+type HasSystemInfoSupplier = internal.HasSystemInfoSupplier
```
cretz: Recommend use of `SysInfo` instead of `SystemInfo` (and calling the options field `SysInfoX` not `InfoX`). Also, https://github.com/temporalio/sdk-go/pull/2148#discussion_r2752877488 implies this should be called "Provider" not "Supplier" (I know we changed the contrib lib, but we didn't change these iface names)

internal/client.go
```diff
@@ -533,6 +534,14 @@ type (
 		//
 		// NOTE: Experimental
 		Plugins []ClientPlugin
+
+		// WorkerHeartbeatInterval is the interval at which the worker will send heartbeats to the server.
+		// Interval must be between 1s and 60s, inclusive.
+		//
+		// default: 60s. To disable, set to 0.
```
cretz, line 594: Pedantic, but would maybe recommend "negative to disable" and remove the pointer and assume 0/unset means use default. Not a big deal though.

internal/client.go
```diff
@@ -1130,9 +1139,15 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
 		options.ConnectionOptions.GetSystemInfoTimeout = defaultGetSystemInfoTimeout
 	}
 
+	if options.Logger == nil {
```
cretz: Can you help me understand this change? All user-facing code that calls `NewServiceClient` should already be setting this default. Is there a new code path that calls `NewServiceClient`?

internal/internal_worker.go
```diff
@@ -1468,9 +1479,67 @@ func (aw *AggregatedWorker) Stop() {
 		WorkerInstanceKey: aw.workerInstanceKey,
```
cretz, line 1476: Arguably we should go ahead and update API dependency and set this on the poll and shutdown calls, but we can make an issue to do that in successive PR if we want

internal/internal_worker_heartbeat_metrics.go
```diff
@@ -0,0 +1,319 @@
...
+func buildPollerInfo(currentPollers int32, lastSuccessfulPollTime time.Time, pollerBehavior PollerBehavior) *workerpb.WorkerPollerInfo {
+	var isAutoscaling bool
+	switch pollerBehavior.(type) {
+	case *pollerBehaviorAutoscaling:
+		isAutoscaling = true
+	}
+
+	return &workerpb.WorkerPollerInfo{
+		CurrentPollers:         currentPollers,
+		LastSuccessfulPollTime: timestamppb.New(lastSuccessfulPollTime),
```
cretz: May have to check, but this may give the year 1 if there is no last successful poll time yet. I think you should set this to `nil` if `lastSuccessfulPollTime.IsZero()`

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,215 @@
...
+	close(hw.stopC)
+	<-hw.stoppedC
+}
+
+// pollTimeTracker tracks the last successful poll time for each poller type.
+type pollTimeTracker struct {
+	times sync.Map // pollerType (string) -> time.Time (stored as int64 nanos)
+}
+
+func newPollTimeTracker() *pollTimeTracker {
```
cretz: Pedantic, but FWIW, more idiomatic Go would not even have this `newPollTimeTracker` call, and would have those storing it on their structs not use a pointer here (the default/zero value is acceptable, think how `sync.Map` works). EDIT: I see now it can be nil when heartbeating is disabled, but meh, harmless to always have regardless of heartbeating and clearer to understand too.

internal/internal_nexus_worker.go
```diff
@@ -68,19 +69,27 @@ func newNexusWorker(opts nexusWorkerOptions) (*nexusWorker, error) {
 
 	baseWorker := newBaseWorker(bwo)
 
+	// Type assert to get the concrete client for namespace capabilities loading
+	var workflowClient *WorkflowClient
+	if wc, ok := opts.client.(*WorkflowClient); ok {
+		workflowClient = wc
+	}
```
cretz: Should we ever accept a situation where client is not this? But in general, why does `nexusWorker` need a client? Loading capabilities seems like something the outer/aggregate worker would do.

internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+}
+
+func (m *heartbeatManager) getOrCreateSharedNamespaceWorker(worker *AggregatedWorker) (*sharedNamespaceWorker, error) {
+	namespace := worker.executionParams.Namespace
+	m.workersMutex.Lock()
+	hw, ok := m.workers[namespace]
+	m.workersMutex.Unlock()
+	if !ok {
+		capabilities, err := m.client.loadNamespaceCapabilities(worker.heartbeatMetrics)
```
cretz: This should be cheap I assume because it's surely already done? Therefore, I would move this capabilities check to the top of the function before everything else, and use `m.workersMutex.Lock()` + `defer m.workersMutex.Unlock()` right after that to make your locking easier to read and safer from mistakes.

internal/resource_tuner.go
```diff
@@ -21,13 +16,49 @@ const (
 	resourceSlotsMemUsage = "temporal_resource_slots_mem_usage"
 )
 
+// SystemInfoSupplier implementations provide information about system resources.
+//
+// Exposed as: [go.temporal.io/sdk/worker.SystemInfoSupplier]
+type SystemInfoSupplier interface {
+	// GetMemoryUsage returns the current system memory usage as a fraction of total memory between
+	// 0 and 1.
+	GetMemoryUsage(infoContext *SystemInfoContext) (float64, error)
```
cretz: Pedantic, but wouldn't use `Get` prefix in Go, but not that important, same for the getter on `HasSystemInfoSupplier`

internal/resource_tuner.go
```diff
@@ -21,13 +16,49 @@ const (
...
+	// GetCpuUsage returns the current system CPU usage as a fraction of total CPU usage between 0
+	// and 1.
+	GetCpuUsage(infoContext *SystemInfoContext) (float64, error)
+}
+
+// SystemInfoContext provides context for SystemInfoSupplier calls.
+//
+// Exposed as: [go.temporal.io/sdk/worker.SystemInfoContext]
+type SystemInfoContext struct {
```
cretz: Consider embedding `context.Context` in here even if just `context.Background()` (but not that important)

yuandrew: keeping out for now, `localActivityContext`, `NexusOperationContext` and `testContext` are all *Context structs that are missing `context.Context`, if we want to add it in later to use, we can

internal/resource_tuner.go
yuandrew: removed

internal/internal_nexus_worker.go
yuandrew: agree we don't want this requirement on `WorkerClient`, removed all individual specific workers using this and instead `AggregateWorker` makes this call instead

internal/internal_worker_heartbeat.go
yuandrew: sounds good, changed from pointer to directly use the struct

internal/internal_worker_heartbeat_metrics.go
yuandrew: thanks, added this check

internal/internal_worker.go
yuandrew: would prefer to do this in a separate PR

internal/client.go
yuandrew: No new code path, but sessionEnvironmentImpl.SignalCreationResponse calls GetClient without a logger, this seems like it would fix that missing scenario, but also seems like good practice to have in general?

internal/client.go
yuandrew: I remember consulting claude on this design decision, it seemed like this was the more idiomatic choice for Go, 

>   1. HashiCorp consul-template — the most extensive example. Nearly every duration config is
  *time.Duration with a Finalize() method that fills defaults for nil. They even have a
  TimeDurationPresent() helper that returns true only if non-nil AND non-zero, explicitly treating zero
  as a distinct state from nil.

  https://pkg.go.dev/github.com/hashicorp/consul-template/config

worker/tuning.go
yuandrew: sounds good, changed the SDK-side names to match

contrib/sysinfo/cgroups.go
yuandrew: Let's do this in a separate PR, left myself a note to add a test for this

contrib/sysinfo/sysinfo.go
yuandrew: oo thanks for catching, removed

internal/internal_worker.go
yuandrew: https://github.com/temporalio/sdk-go/pull/2178

contrib/sysinfo/cgroups.go
yuandrew: https://github.com/temporalio/sdk-go/pull/2179

internal/client.go
cretz: > this seems like it would fix that missing scenario, but also seems like good practice to have in general?

They may have left logger off intentionally, unsure, but would take some more digging. It's possible/likely you're right, but it seems unrelated to this project and may deserve a separate issue.

internal/resource_tuner.go
cretz: Then IMO we should consider passing a Go context into the calls alongside where this is passed (even if it's not used for anything yet)

yuandrew: I'm planning on leaving it out for now, and adding in the context into the `SysInfoContext` struct when we need the context. It seems a little redundant to pass in a `Context` and `SysInfoContext` together, any reason we'd need to add it into the calls alongside right now?

internal/client.go
yuandrew: yeah that's fair, I'll remove this for now

yuandrew: turns out this was due to `newHeartbeatManager` using the logger, but I added the check into that function itself, so `SignalCreationResponse` is unaffected

PR #2149, opened 2026-01-15
Add Serialization context for codecs and converters for Activities and Child Workflows

### LINKED ISSUE #1352: Context aware data converters cannot get IDs of outbound activities and child workflow
Context aware data converters should have some way to get ID of outbound activities and child workflows when the data converter is being used to serialize activity input or child workflow input. This is useful if the data converter changes behaviour based on the workflow ID, for example using the ID as associated data for encryption.

In the Java SDK this is trivial because the Java SDK will pass the child workflow ID to the data converter when serializing the child workflow input

I can think of acceptable solutions:
1. Users could create an interceptor to manually set the activity ID/ child workflow ID, maybe the SDK can provide some random generator to help create deterministic UUIDs. Other SDK already have helpers for this so maybe it is worth doing regardless.

2. Ideally the ID would be accessible in the `WithContext` or `WithWorkflowContext` call on the data converter. When we generate the ID in question we could set it on the context and pass that context with the ID to the context aware data converter. This would require some minor refactoring to when the SDK generates an activity ID/ child workflow ID

yuandrew:
## What was changed
Moved ID generation to before serialization to allow data converters and codecs to have access to ID.

## Why?
Useful information to have, prevent replay attacks

## Checklist

1. Closes #1352

2. How was this tested:
Added tests

3. Any docs updates needed?

---

> [!NOTE]
> Introduces context-aware outbound metadata and pre-serialization ID generation for activities and child workflows.
> 
> - Add `OutboundInfo` and `GetOutboundInfo` to workflow context so codecs/data converters can read `ActivityID`/`ChildWorkflowID` during input serialization
> - Generate IDs before serialization: activities get `ScheduleID` (new in `ExecuteActivityOptions`) and resolved `ActivityID`; child workflows get resolved `WorkflowID` early; both are set on params and used by commands
> - Update env/interfaces to support sequencing (`GenerateSequence`) and consume provided IDs (`scheduleActivityTask` now takes `ScheduleID`); remove late ID generation paths
> - Expose APIs via `workflow` package; add comprehensive tests validating `OutboundInfo` capture and bindings updates to pass `ScheduleID`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cb75a870ec1ebf63d9ea7f47729c7a973a318684. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -423,6 +423,12 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(
 	if params.TaskQueueName == "" {
 		return nil, serviceerror.NewWorkflowExecutionAlreadyStarted("Empty task queue name", "", "")
 	}
+
+	// Tests set workflow ID slightly differently than prod, the test env doesn't set currentRunID,
+	// so we must detect and change ID to use RunID
+	if strings.HasPrefix(params.WorkflowID, "_") {
+		params.WorkflowID = env.workflowInfo.WorkflowExecution.RunID + params.WorkflowID
+	}
 	var cronSchedule string
 	if len(params.CronSchedule) > 0 {
 		cronSchedule = params.CronSchedule


internal/workflow.go
```diff
@@ -103,6 +103,46 @@ var (
...
+// using the ID as associated data for encryption.
+//
+// Exposed as: [go.temporal.io/sdk/workflow.OutboundInfo]
+type OutboundInfo struct {
+	// ActivityID is set when scheduling an activity.
+	ActivityID string
+	// ChildWorkflowID is set when starting a child workflow.
+	ChildWorkflowID string
+	// NexusOperationSeq is set when executing a Nexus operation (sequence number as string).
+	NexusOperationSeq string
```
Quinn-With-Two-Ns: Note Nexus does not support serialization context in any SDK that support both, and that is intentional due to how results are mapped from async operations. I don't think we should add it to Go at this time.

yuandrew: ah okay, i misunderstood your comment in the GH issues. Thanks for the info! Will remove

internal/internal_workflow_testsuite.go
```diff
@@ -423,10 +423,6 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(
 	if params.TaskQueueName == "" {
 		return nil, serviceerror.NewWorkflowExecutionAlreadyStarted("Empty task queue name", "", "")
 	}
-
```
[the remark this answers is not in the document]
yuandrew: Hmm, this seems legit

internal/internal_workflow_testsuite.go
```diff
@@ -424,8 +424,10 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(
 		return nil, serviceerror.NewWorkflowExecutionAlreadyStarted("Empty task queue name", "", "")
 	}
 
-	if params.WorkflowID == "" {
-		params.WorkflowID = env.workflowInfo.WorkflowExecution.RunID + "_" + getStringID(env.nextID())
+	// Tests set workflow ID slightly differently than prod, the test env doesn't set currentRunID,
+	// so we must detect and change ID to use RunID
```
cretz, line 428: I don't understand this exactly, what happens if we leave the code as it was?

internal/internal_event_handlers.go
```diff
@@ -722,13 +719,7 @@ func (wc *workflowEnvironmentImpl) CreateNewCommand(commandType enumspb.CommandT
 
 func (wc *workflowEnvironmentImpl) ExecuteActivity(parameters ExecuteActivityParams, callback ResultHandler) ActivityID {
 	scheduleTaskAttr := &commandpb.ScheduleActivityTaskCommandAttributes{}
-	scheduleID := wc.GenerateSequence()
```
cretz, line 725: Can you confirm for me by walking the code path during activity execute that the generation of this ID that now happens earlier has no other things that happen between it and this point? Meaning, this sequence can't be changed in this PR?

For example, what if (and obviously it would be bad on their part) a user had made a command-generating workflow call inside their data converter today?

workflow/workflow.go
cretz, line 1: I don't think this is how this should be implemented and I'm not sure all of the features have been properly captured. To match other SDKs, I think we need the following:

* `converters.SerializationContext` as an interface with private method
* `converters.ActivitySerializationContext` and `converters.WorklowSerializationContext` structs that implement `SerializationContext` and contain the limited-but-available-inbound-and-outbound activity and workflow info respectively
* `converters.WithSerializationContext` (or `SerializationContextAware`) interface that has a single `WithSerializationContext(SerializationContext) DataConverter` function
* ^ Call that for both outbound and inbound serialization every time. So that means for outbound activity calls from workflow and inbound activity execute. That also means for workflow outbound calls from client and inbound workflow execute. Basically every outbound call must match an inbound equivalent and vice versa.

This is admittedly a non-trivial effort. In addition to Java, can look at https://github.com/temporalio/sdk-dotnet/pull/446 and https://github.com/temporalio/sdk-python/pull/1102 for inspiration.

PR #2158, opened 2026-01-21
💥 Standalone activities support

### LINKED ISSUE #2124: Support standalone activities

Allow executing activities without a workflow. The clients must support the new (upcoming) server APIs for starting, managing and getting results of activities. The workers must support running activities outside of workflow context.

**Additional context**

Server API PR (WIP): https://github.com/temporalio/api/pull/640

maciejdudko:
## What was changed
- Implemented support for running standalone activity tasks.
- Client calls for standalone activities: execute, get result, describe, cancel, terminate, list, count.
- Changes to `TestActivityEnvironment` to support executing activities as standalone.
- 💥 [activity task tokens for tests have been changed](https://github.com/temporalio/sdk-go/pull/2158#discussion_r2721745352), now `testActivityToken` (containing activityID and runID) are used to keep track of activities, instead of just `activityID`. Most users should not be affected, only if you're manually constructing task tokens in your tests.

## Checklist

1. Closes #2124

2. How was this tested:

New integration tests: `TestIntegrationSuite/TestExecuteActivitySuite`

3. Any docs updates needed?

Standalone activity docs are in progress.

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -2895,6 +2895,9 @@ func (ts *IntegrationTestSuite) TestInterceptorStartWithSignal() {
 }
 
 func (ts *IntegrationTestSuite) TestInterceptorStandaloneActivity() {
+	if os.Getenv("DISABLE_STANDALONE_ACTIVITY_TESTS") != "" {
+		ts.T().SkipNow()
+	}
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 


client/client.go
```diff
@@ -896,6 +897,16 @@ type (
...
 
+	ExecuteActivityOptions       = internal.ClientExecuteActivityOptions
+	ListActivitiesOptions        = internal.ListActivitiesOptions
+	CountActivitiesOptions       = internal.CountActivitiesOptions
+	CountActivitiesResult        = internal.CountActivitiesResult
+	ActivityHandle               = internal.ActivityHandle
+	ActivityExecutionMetadata    = internal.ActivityExecutionMetadata
+	ActivityExecutionDescription = internal.ActivityExecutionDescription
+	CancelActivityOptions        = internal.CancelActivityOptions
+	TerminateActivityOptions     = internal.TerminateActivityOptions
```
cretz: All of these will need docs IMO. Same goes for any user-facing construct in other parts of the PR (many are missing docs).

client/client.go
```diff
@@ -896,6 +897,16 @@ type (
 	// Note, this is not related to any general concept of timing out or cancelling a running update, this is only related to the client call itself.
 	WorkflowUpdateServiceTimeoutOrCanceledError = internal.WorkflowUpdateServiceTimeoutOrCanceledError
 
+	ExecuteActivityOptions       = internal.ClientExecuteActivityOptions
```
cretz: `Client.ExecuteWorkflow` accepts `client.StartWorkflowOptions`, should `Client.ExecuteActivity` accept `client.StartActivityOptions`?

client/client.go
```diff
@@ -1038,19 +1049,40 @@ type (
...
 		//  - ApplicationError
 		//  - TimeoutError
 		//  - CanceledError
 		CompleteActivityByID(ctx context.Context, namespace, workflowID, runID, activityID string, result interface{}, err error) error
 
+		// CompleteActivityByActivityID reports activity completed.
+		// Similar to CompleteActivity, but may save user from keeping taskToken info.
+		// This method works only for standalone activities. To complete a workflow activity, use CompleteActivityByID.
+		//
+		// Activity's Execute method can return activity.ErrResultPending to
```
cretz: I would not expect activity "execute" (which is "start" in Go) to ever return an error on successful start. Actually, I think I understand now that this is just bad copy/paste.

client/client.go
```diff
@@ -1306,6 +1338,14 @@ type (
 		// if not specified the most recent runID will be used.
 		GetWorkflowUpdateHandle(ref GetWorkflowUpdateHandleOptions) WorkflowUpdateHandle
 
+		ExecuteActivity(ctx context.Context, options ExecuteActivityOptions, activity any, args ...any) (ActivityHandle, error)
+
+		GetActivityHandle(activityID string, runID string) ActivityHandle
```
cretz: I think this is probably ok, but I wonder if we may want to accept options here. My concern is that we may need something in the future like "first execution run ID" and we have nowhere to put it in Go.

client/client.go
```diff
@@ -896,6 +897,16 @@ type (
 	// Note, this is not related to any general concept of timing out or cancelling a running update, this is only related to the client call itself.
 	WorkflowUpdateServiceTimeoutOrCanceledError = internal.WorkflowUpdateServiceTimeoutOrCanceledError
 
+	ExecuteActivityOptions       = internal.ClientExecuteActivityOptions
+	ListActivitiesOptions        = internal.ListActivitiesOptions
+	CountActivitiesOptions       = internal.CountActivitiesOptions
+	CountActivitiesResult        = internal.CountActivitiesResult
+	ActivityHandle               = internal.ActivityHandle
+	ActivityExecutionMetadata    = internal.ActivityExecutionMetadata
```
cretz: Why `ActivityExecutionMetadata` instead of just `ActivityExecution` or maybe `ActivityExecutionInfo`? Not sure we have ever used this metadata terminology this way (and can't think of when we will ever anywhere else).

internal/activity.go
```diff
@@ -337,15 +338,41 @@ func WithActivityTask(
 		attempt:                task.GetAttempt(),
 		priority:               task.GetPriority(),
 		heartbeatDetails:       task.HeartbeatDetails,
-		workflowType: &WorkflowType{
+		namespace:              task.WorkflowNamespace,
```
cretz, line 342: Side note, I think we should go back and add a comment to clarify this is the general namespace unrelated to workflow on `workflow_namespace` at https://github.com/temporalio/api/blob/173e50794ca2c3ce2d2f8e98010a3e1eabd145d4/temporal/api/workflowservice/v1/request_response.proto#L449-L450. Otherwise, the name is confusing.

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+	// Exposed as: [go.temporal.io/sdk/client.ExecuteActivityOptions]
+	ClientExecuteActivityOptions struct {
+		ID                     string
+		TaskQueue              string
+		ScheduleToCloseTimeout time.Duration
+		ScheduleToStartTimeout time.Duration
+		StartToCloseTimeout    time.Duration
+		HeartbeatTimeout       time.Duration
+		IDConflictPolicy       enumspb.ActivityIdConflictPolicy
+		IDReusePolicy          enumspb.ActivityIdReusePolicy
```
cretz: We chose to prefix the ID policy field names in `client.ExecuteWorkflowOptions` with `Workflow`, we should do the same here with `Activity`

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+		RetryPolicy            *RetryPolicy
+		SearchAttributes       SearchAttributes
+		Summary                string
+		Priority               Priority
+	}
+
+	ListActivitiesOptions struct {
```
cretz: I think it makes sense to prefix all of these `internal` struct names with `Client` (though can remove that prefix when aliasing of course)

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+	CountActivitiesOptions struct {
+		Query string
+	}
+
+	CountActivitiesResult struct {
+		Count  int64
+		Groups []ActivityAggregationGroup
+	}
+
+	ActivityAggregationGroup struct {
```
cretz: In addition to the `Client` prefix before, I recommend calling this `ClientCountActivitiesAggregationGroup` to properly scope it

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+		Describe(ctx context.Context) (*ActivityExecutionDescription, error)
+		// Cancel requests cancellation of the activity.
+		Cancel(ctx context.Context, options CancelActivityOptions) error
+		// Terminate terminates the activity.
+		Terminate(ctx context.Context, options TerminateActivityOptions) error
+	}
+
+	// DescribeActivityOptions contains options for ActivityHandle.Describe call.
+	// For future compatibility, currently unused.
+	DescribeActivityOptions struct{}
```
cretz: Need to set this as the parameter of `Describe` to provide that future compatibility help. Also, do we need to support getting input/outcome? This is present in the Python one.

client/client.go
```diff
@@ -896,6 +897,16 @@ type (
 	// Note, this is not related to any general concept of timing out or cancelling a running update, this is only related to the client call itself.
 	WorkflowUpdateServiceTimeoutOrCanceledError = internal.WorkflowUpdateServiceTimeoutOrCanceledError
 
+	ExecuteActivityOptions       = internal.ClientExecuteActivityOptions
+	ListActivitiesOptions        = internal.ListActivitiesOptions
+	CountActivitiesOptions       = internal.CountActivitiesOptions
+	CountActivitiesResult        = internal.CountActivitiesResult
```
cretz: You have to expose all types a user may need, this includes the aggregation group, describe options, etc

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+	ActivityExecutionDescription struct {
+		ActivityExecutionMetadata
+		// RawExecutionInfo is the raw PB message this struct was built from.
+		RawExecutionInfo *activitypb.ActivityExecutionInfo
+		// DataConverter is the data converter used for reading certain fields. By default, it's the data converter of
+		// the client used to make the ActivityHandle.Describe call.
+		DataConverter converter.DataConverter
+		// FailureConverter is the failure converter used for reading last failure. By default, it's the data converter
+		// of the client used to make the ActivityHandle.Describe call.
+		FailureConverter        converter.FailureConverter
```
cretz: Not sure these need to be exported or mutable (any more than they would be for getting the result from an activity handle)

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+func (d *ActivityExecutionDescription) GetHeartbeatDetailsCount() int {
+	return len(d.RawExecutionInfo.GetHeartbeatDetails().GetPayloads())
+}
+
+// GetHeartbeatDetails converts heartbeat details using DataConverter and stores them in a slice.
+// If successful, valuesPtr is set to the resulting slice.
+//
+// valuesPtr must be a pointer to slice (*[]T). The type of the slice is retrieved from valuesPtr using reflection;
+// heartbeat details are converted to that type.
+func (d *ActivityExecutionDescription) GetHeartbeatDetails(valuesPtr any) error {
```
cretz: Requiring a slice doesn't work well here, you can't get different types at different indices correct. I think you can follow the model `activity.GetHeartbeatDetails` follows and accept a vararg.

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+		return err
+	}
+	valuesPtrVal.Set(detailsSlice)
+	return nil
+}
+
+// GetLastFailure returns the last failure of the activity execution, using FailureConverter for conversion.
+// Returns nil if there was no failure.
+func (d *ActivityExecutionDescription) GetLastFailure() error {
+	return d.FailureConverter.FailureToError(d.RawExecutionInfo.GetLastFailure())
```
cretz: Pedantic, but would like to eagerly return nil here if there is no failure instead of expect all failure converter implementations handle nil failure properly

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+		ActivityType         string
+		ScheduleTime         time.Time
+		CloseTime            time.Time
+		Status               enumspb.ActivityExecutionStatus
+		SearchAttributes     SearchAttributes
+		TaskQueue            string
+		ExecutionDuration    time.Duration
+	}
+
+	ActivityExecutionDescription struct {
```
cretz: Do we need to model getting outcome (success/failure)? Python PR will have this feature.

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+		TaskQueue              string
+		ScheduleToCloseTimeout time.Duration
+		ScheduleToStartTimeout time.Duration
+		StartToCloseTimeout    time.Duration
+		HeartbeatTimeout       time.Duration
+		IDConflictPolicy       enumspb.ActivityIdConflictPolicy
+		IDReusePolicy          enumspb.ActivityIdReusePolicy
+		RetryPolicy            *RetryPolicy
+		SearchAttributes       SearchAttributes
+		Summary                string
```
cretz: Comment on `StartActivityExecutionRequest.user_summary` implies summary _and_ details are supported

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+		ID                     string
+		TaskQueue              string
+		ScheduleToCloseTimeout time.Duration
+		ScheduleToStartTimeout time.Duration
+		StartToCloseTimeout    time.Duration
+		HeartbeatTimeout       time.Duration
+		IDConflictPolicy       enumspb.ActivityIdConflictPolicy
+		IDReusePolicy          enumspb.ActivityIdReusePolicy
+		RetryPolicy            *RetryPolicy
+		SearchAttributes       SearchAttributes
```
cretz: Pedantic, but would be ok calling this field `TypedSearchAttributes` to match the others, but ok with this too

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+
+func (h *activityHandleImpl) GetID() string {
+	return h.id
+}
+
+func (h *activityHandleImpl) GetRunID() string {
+	return h.runID
+}
+
+func (h *activityHandleImpl) Get(ctx context.Context, valuePtr any) error {
```
cretz: I admit I didn't check integration tests, but can you confirm the behavior of a short context timeout on this call in relation to the workflow equivalent? I suspect it'll return the error properly, just want to confirm.

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+				return nil
+			} else {
+				return h.client.dataConverter.FromPayloads(result, valuePtr)
+			}
+		}
+	}
+}
+
+func (h *activityHandleImpl) pollResult(ctx context.Context) (*workflowservice.PollActivityExecutionResponse, error) {
+	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
```
cretz: There's some things that need to change for some of these contexts specific to activities and/or this call. First, can you confirm that you need or don't need to pass `grpcMetricsHandler`? And can you confirm whether or not you should be setting some RPC tags? Need to check the same for other calls in this PR. Second, I think this needs to have `grpcLongPoll(true)` set on it.

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+		RunId:      h.runID,
+		Identity:   h.client.identity,
+		RequestId:  uuid.NewString(),
+		Reason:     options.Reason,
+	}
+	_, err := h.client.WorkflowService().TerminateActivityExecution(grpcCtx, request)
+	return err
+}
+
+func (wc *WorkflowClient) ExecuteActivity(ctx context.Context, options ClientExecuteActivityOptions, activity any, args ...any) (ActivityHandle, error) {
```
cretz: We need to design interceptors up front IMO

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+	}
+	if options.StartToCloseTimeout < 0 {
+		return errors.New("negative StartToCloseTimeout")
+	}
+	startToCloseTimeout := options.StartToCloseTimeout
+	if startToCloseTimeout == 0 {
+		if options.ScheduleToCloseTimeout == 0 {
+			return errors.New("at least one of ScheduleToCloseTimeout and StartToCloseTimeout is required")
+		} else {
+			startToCloseTimeout = options.ScheduleToCloseTimeout
```
cretz: I am not sure we should do this SDK side (but server can do it)

internal/internal_activity_client.go
```diff
@@ -0,0 +1,531 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.Groups == nil {
+		return &CountActivitiesResult{
+			Count:  resp.Count,
+			Groups: nil,
+		}, nil
+	}
```
cretz: Don't really need this part, though mostly harmless. If you must have it, change it to `if len(resp.Groups) == 0` (in Go, readers shouldn't assume difference in empty and nil unless there's a really good reason)

internal/internal_workflow_testsuite.go
```diff
@@ -1186,6 +1197,11 @@ func (env *testWorkflowEnvironmentImpl) CompleteActivity(taskToken []byte, resul
 	if taskToken == nil {
 		return errors.New("nil task token provided")
 	}
+	activityToken, ok := activityTokenFromBytes(taskToken)
+	if !ok {
+		return errors.New("invalid task token provided")
+	}
```
cretz, line 1199: _Technically_ a backwards incompatible change to change token formats for existing tests, though likely harmless

internal/workflow_testsuite.go
```diff
@@ -273,6 +273,15 @@ func (t *TestActivityEnvironment) SetOnActivityHeartbeatListener(
 	return t
 }
 
+// SetExecuteActivitiesInWorkflow controls the simulated environment in which the tested activity is being executed.
+// If set to true, the activity will be executed as if it was started by a workflow.
+// If set to false, the activity will be executed as if it was started directly by a client.
+// Defaults to true.
```
cretz, line 281: Can you clarify the behavior difference in the two? Specifically, for most users, does it matter? Or does it only affect some activity info stuff?

maciejdudko: Currently this is only about activity info, but I want to keep the possibility of there being more differences in the future. I will elaborate more on the activity info being different.

internal/cmd/build/main.go
```diff
@@ -155,6 +155,9 @@ func (b *builder) integrationTest() error {
 				"--dynamic-config-value", `system.refreshNexusEndpointsMinWait="0s"`, // Make Nexus tests faster
 				"--dynamic-config-value", `component.nexusoperations.recordCancelRequestCompletionEvents=true`, // Defaults to false until after OSS 1.28 is released
 				"--dynamic-config-value", `history.enableRequestIdRefLinks=true`,
+				"--dynamic-config-value", "activity.enableStandalone=true",
+				"--dynamic-config-value", "history.enableChasm=true",
```
yuandrew, line 159: Curious, this enables all of Chasm? Are there other unintended side effects of enabling this?

internal/client.go
```diff
@@ -180,19 +181,40 @@ type (
...
 
+		// CompleteActivityByActivityID reports activity completed.
+		// Similar to CompleteActivity, but may save user from keeping taskToken info.
+		// This method works only for standalone activities. To complete a workflow activity, use CompleteActivityByID.
+		//
+		// Activity's Execute method can return activity.ErrResultPending to
+		// indicate the activity is not completed when its Execute method returns. In that case, this CompleteActivityById() method
+		// should be called when that activity is completed with the actual result and error. If err is nil, activity task
+		// completed event will be reported; if err is CanceledError, activity task canceled event will be reported; otherwise,
+		// activity task failed event will be reported.
```
yuandrew: Not sure why GH won't let me select multi lines, but the sentence 
```
If err is nil, activity task
		// completed event will be reported; if err is CanceledError, activity task canceled event will be reported; otherwise,
		// activity task failed event will be reported.
```

feels unnecessary? I don't think this behavior needs to be described

yuandrew: Especially since we include the errors it can return below

test/activity_test.go
```diff
@@ -273,6 +279,80 @@ func (a *Activities) InspectActivityInfo(ctx context.Context, namespace, taskQue
...
+		return fmt.Errorf("expected ActivityRunID to be empty but got %v", info.ActivityRunID)
+	}
+
+	return nil
+}
+
+func (a *Activities) InspectActivityInfoNoWorkflow(ctx context.Context, namespace, activityID, taskQueue string, scheduleToCloseTimeout, startToCloseTimeout time.Duration, retryPolicy *temporal.RetryPolicy) error {
+	a.append("inspectActivityInfoNoWorkflow")
+	if !activity.IsActivity(ctx) {
+		return fmt.Errorf("expected IsActivity to return %v but got %v", true, activity.IsActivity(ctx))
```
yuandrew: nit: seems like this can just be `expected IsActivity to be true`

test/integration_test.go
```diff
@@ -8462,3 +8464,213 @@ func (ts *IntegrationTestSuite) TestSimplePluginDoNothing() {
...
+		ts.EqualValues(1, description.SearchAttributes.Size())
+		attr, hasAttr := description.SearchAttributes.GetString(searchAttrKey)
+		ts.True(hasAttr)
+		ts.Equal(searchAttrValue, attr)
+		summary, err := description.GetSummary()
+		ts.NoError(err)
+		ts.Equal(options.Summary, summary)
+
+		// ensure measurable amount of time passes, then complete activity
+		time.Sleep(100 * time.Millisecond)
```
yuandrew, line 8724: Can we use `ts.Eventually()` here?

.gitignore
```diff
@@ -24,3 +24,4 @@ go.work.sum
 *~
 internal/cmd/build/*.sqlite*
 .claude
+internal/cmd/build/build
```
yuandrew: Why was this added?

internal/workflow_testsuite_test.go
```diff
@@ -1232,3 +1232,45 @@ func TestDynamicWorkflows(t *testing.T) {
...
+		}
+	}
+	return nil
+}
+
+func TestExecuteActivitiesInWorkflow(t *testing.T) {
+	testSuite := &WorkflowTestSuite{}
+	env := testSuite.NewTestActivityEnvironment()
+	env.RegisterActivity(checkActivityInfo)
+	_, err := env.ExecuteActivity(checkActivityInfo, true)
```
yuandrew, line 1263: Why do we have 3 sets of ExecuteActivity here? Shouldn't we only need 2, where `SetExecuteActivitiesInWorkflow` is `true` and `false`?

internal/workflow_testsuite.go
yuandrew, line 281: For places that use this testing feature, do we need them to `defer previousBehavior()`?

internal/internal_task_handlers.go
```diff
@@ -2246,12 +2246,22 @@ func newServiceInvoker(
...
+		if t.WorkflowExecution.GetWorkflowId() != "" {
+			ath.logger.Debug("Processing new activity task",
+				tagActivityID, t.ActivityId,
+				tagActivityRunID, t.ActivityRunId,
+				tagActivityType, t.ActivityType.GetName(),
+				tagAttempt, t.Attempt,
+			)
+		} else {
+			ath.logger.Debug("Processing new activity task",
+				tagWorkflowID, t.WorkflowExecution.GetWorkflowId(),
```
yuandrew, line 2258: I think these are flipped, how it is currently, `t.WorkflowExecution.GetWorkflowId()` is always "" here.

And should we make the new log mention Processing new standalone activity task, just to make it more clear? or do we want the log message itself to be the same

client/client.go
yuandrew: And don't forget to add `// NOTE: Experimental` to the docs too!

internal/activity.go
```diff
@@ -337,15 +338,41 @@ func WithActivityTask(
...
+		retryPolicy:            convertFromPBRetryPolicy(task.RetryPolicy),
+		workerStopChannel:      workerStopChannel,
+		contextPropagators:     contextPropagators,
+		client:                 client,
+	}
+
+	isWorkflowActivity := task.WorkflowExecution.GetWorkflowId() != ""
```
yuandrew: nit: in `internal_task_handlers.go` you use this condition directly, doesn't seem like we need to set this to a variable here

internal/internal_activity_client.go
yuandrew: would prefer this too, for clarity

internal/cmd/build/main.go
yuandrew: I'm completely unaware of the state of Chasm today haha

internal/workflow_testsuite.go
cretz: :+1: The main reason I ask is because then we don't need most users to set this when testing their activities, only in advanced situations where they are asserting something standalone specific

internal/internal_activity_client.go
maciejdudko: No, this feature is not planned for now for any SDK, including Python.

internal/internal_activity_client.go
maciejdudko: I will add options as argument to Describe. Getting input/outcome in Describe call is not planned for now for any SDK, including Python.

internal/cmd/build/main.go
maciejdudko: Standalone Activities are built on top of CHASM, both need to be enabled for it to work. Server team is working on making sure it has no side effects :)

test/integration_test.go
maciejdudko: Execution time is only set one time, on completion. So `Eventually` wouldn't work here.

.gitignore
maciejdudko: This is the built binary of the `build` command, we don't want to commit it.

internal/internal_activity_client.go
maciejdudko: Workflow activities only support summary, not details. I'd rather keep it the same here, at least for now. Probably safe to add details in both places but that's definitely out of scope here.

client/client.go
maciejdudko: "First execution run ID" is unlikely to ever be a thing for activities. We can add `GetActivityHandleWithOptions` in the future if we ever need it.

internal/internal_activity_client.go
cretz: This needs to be clarified in the proto docs, because right now it says details too

internal/activity.go
maciejdudko: Done: https://github.com/temporalio/api/pull/702

internal/workflow_testsuite_test.go
maciejdudko: The first one tests that default is true, the third tests explicitly setting true. Added clarifying comments.

internal/workflow_testsuite.go
maciejdudko: No, we always create local activity test env inside the test.

internal/internal_activity_client.go
yuandrew: Agree we should be using `grpcLongPoll(true)`

> And can you confirm whether or not you should be setting some RPC tags?

What dictates whether or not we should be? If we think it'll be useful or not? I'm wonder if I'm missing context on how this decision is made. 

From my digging into the code + using AI, it seems like grpcMetricsHandler is only worth it if we also use `RPCTags`. So it's either we omit both, or we include both, with  something like `metricsHandler.WithTags(metrics.RPCTags(metrics.NoneTagValue, h.activityType, metrics.NoneTagValue))`?

client/client.go
yuandrew: For future-proofing, switched this to take `GetActivityHandleOptions`, which also mirrors `ClientGetActivityHandleInput`

client/client.go
yuandrew: maciej renamed to `ActivityExecutionInfo`

internal/internal_activity_client.go
yuandrew: agreed, made https://github.com/temporalio/api/pull/707

internal/internal_activity_client.go
yuandrew: looks like this was done

internal/internal_activity_client.go
yuandrew: looks like maciej added this (as a parameter for `Describe`)

internal/internal_activity_client.go
yuandrew: looks like maciej made these private

internal/internal_activity_client.go
yuandrew: maciej completed this

internal/internal_activity_client.go
yuandrew: has been changed to `GetHeartbeatDetails(valuePtrs ...any) `

internal/internal_activity_client.go
yuandrew: This was removed, we now just return an error if `options.StartToCloseTimeout == 0 && options.ScheduleToCloseTimeout == 0 `

internal/internal_activity_client.go
yuandrew: maciej removed this

internal/internal_workflow_testsuite.go
yuandrew: Seems like this break is needed with standalone activity's requirements? Can add a 💥 line in the PR description noting this small breaking change

internal/workflow_testsuite.go
yuandrew: Maciej added a blurb

.gitignore
yuandrew: removing, no instructions cause you to build this binary, so I don't think we need a special ignore for it

internal/internal_activity_client.go
cretz: To clarify, I only meant the field in the SDK, not the underlying proto

internal/internal_activity_client.go
cretz: Yes, if no additional tags needed for metrics on this call, no grpcMetricsHandler needed probably

internal/internal_activity_client.go
yuandrew: gotcha, yeah I gave up on the API efforts, don't wanna go through the pain of coordinating OSS/SDK updates just for the name

the fields themselves have been renamed

client/client.go
yuandrew: All types should be exposed now

internal/internal_activity_client.go
yuandrew: Sorry, what level is this decision made? Do you know if this is a crew level decision (with PM input), or just something we decide for SDK?

client/client.go
yuandrew: All public facing APIs should have docs now, and marked experimental

internal/internal_activity_client.go
yuandrew: I saw Maciej added ExecuteActivity and GetActivityHandle.

Am I correct in understanding we only need interceptor support for publicly exposed client-level methods? I ask because Workflow has `Describe/Cancel/TerminateWorkflow`, but those are all exposed through the public interface, whereas standalone is exposed through `ActivityHandle`

internal/internal_activity_client.go
yuandrew: Since this is a new high-level thing, added details support

internal/internal_activity_client.go
yuandrew: yes, looks like there is already a test for this

```
	ts.Run("Activity result timeout", func() {
		ctx, cancel := context.WithTimeout(context.Background(), ctxTimeout)
		defer cancel()
		handle, err := ts.client.ExecuteActivity(ctx, makeOptions(), "readFromChannelActivity")
		ts.NoError(err)

		getCtx, getCancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
		defer getCancel()
		err = handle.Get(getCtx, nil)
		var serviceErr *serviceerror.DeadlineExceeded
		ts.True(errors.Is(err, context.DeadlineExceeded) || errors.As(err, &serviceErr))
		activityResultChan <- "" // allow activity to complete
	})
```

internal/internal_activity_client.go
yuandrew: agree that grpcMetricsHandler isn't needed, there are no meaningful tags to add

client/client.go
cretz: Bump, this doesn't make sense here right?

client/client.go
```diff
@@ -1306,6 +1397,49 @@ type (
...
+		// The user can use this to start using a function or activity type name.
+		// Either by
+		//     ExecuteActivity(ctx, options, "activityTypeName", arg1, arg2, arg3)
+		//     or
+		//     ExecuteActivity(ctx, options, activityFn, arg1, arg2, arg3)
+		//
+		// Returns an ActivityExecutionAlreadyStarted error if an activity with the same ID already exists
+		// in this namespace, unless permitted by the specified ID conflict policy.
+		//
+		// ActivityHandle has the following methods:
```
cretz: Not sure we should have to keep an up-to-date list of the return type's interface methods here

client/client.go
```diff
@@ -1306,6 +1397,49 @@ type (
...
+
+		// GetActivityHandle creates a handle to the referenced activity.
+		//
+		// NOTE: Experimental
+		GetActivityHandle(options GetActivityHandleOptions) ActivityHandle
+
+		// ListActivities lists activity executions based on query.
+		//
+		// NOTE: Experimental
+		ListActivities(ctx context.Context, options ListActivitiesOptions) iter.Seq2[*ActivityExecutionInfo, error]
```
cretz: I think we should consider putting the iterator inside a response struct. We never know when we might want to also return something general (e.g. total count). Other langs can get away with this since they have overloads and such.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/nexus-rpc/sdk-go v0.5.1
 	github.com/robfig/cron v1.2.0
 	github.com/stretchr/testify v1.10.0
-	go.temporal.io/api v1.59.0
+	go.temporal.io/api v1.61.1-0.20260122222125-7b2c423d8b2f
```
cretz: Will have to adjust this to a real tag before merging

internal/interceptor.go
```diff
@@ -538,14 +571,104 @@ type ClientDescribeWorkflowInput struct {
...
 }
 
+// ClientExecuteActivityInput is the input to
+// ClientOutboundInterceptor.ExecuteActivity.
+//
+// NOTE: Experimental
+//
+// Exposed as: [go.temporal.io/sdk/interceptor.ClientExecuteActivityInput]
+type ClientExecuteActivityInput struct {
+	Options      *ClientStartActivityOptions
```
cretz, line 588: FWIW, it is arguable that this should not be a pointer, but since this matches other interceptor calls, no prob

internal/interceptor.go
```diff
@@ -538,14 +571,104 @@ type ClientDescribeWorkflowInput struct {
...
+	ActivityType *ActivityType
```
cretz: Any reason this shouldn't just be a string like it is for workflow?

internal/interceptor.go
```diff
@@ -538,14 +571,104 @@ type ClientDescribeWorkflowInput struct {
...
+
+// ClientPollActivityResultOutput is the output of
+// ClientOutboundInterceptor.PollActivityResult.
+//
+// NOTE: Experimental
+//
+// Exposed as: [go.temporal.io/sdk/interceptor.ClientPollActivityResultOutput]
+type ClientPollActivityResultOutput struct {
+	Result  *commonpb.Payloads
+	Failure *failurepb.Failure
```
cretz: I wonder if it makes more sense to match `ClientPollWorkflowUpdateOutput` (e.g. encoded value + error)

internal/internal_activity_client.go
```diff
@@ -0,0 +1,661 @@
...
+		// GetID returns the ID of the activity this handle points to.
+		GetID() string
+		// GetRunID returns the run ID that this handle was created with.
+		//
+		// Handle returned by [client.Client] has it set to run ID of the started execution.
+		//
+		// Handle returned by client.Client.GetActivityHandle has it set to the provided run ID.
+		// If empty run ID was provided, then this function returns empty string and the handle points to the most
+		// recent execution with matching activity ID. The run ID of this execution can be retrieved by calling Describe.
+		GetRunID() string
```
cretz, line 190: I think we should remove `Get` from in front of these

internal/internal_activity_client.go
```diff
@@ -0,0 +1,661 @@
...
+	// clientActivityHandleImpl is the default implementation of ClientActivityHandle.
+	clientActivityHandleImpl struct {
+		client *WorkflowClient
+		id     string
+		runID  string
+	}
+)
+
+// HasHeartbeatDetailsCount returns whether heartbeat details are present. Use GetHeartbeatDetails to retrieve them.
+func (d *ClientActivityExecutionDescription) HasHeartbeatDetailsCount() bool {
```
cretz:
```suggestion
func (d *ClientActivityExecutionDescription) HasHeartbeatDetails() bool {
```

internal/internal_activity_client.go
```diff
@@ -0,0 +1,661 @@
...
+	if failure == nil {
+		return nil
+	}
+	return d.failureConverter.FailureToError(failure)
+}
+
+// GetSummary returns summary of the activity. See ClientStartActivityOptions.Summary. Returns empty string if there is no summary.
+// Uses the data converter of the client used to make the Describe call. Returns error if data conversion fails.
+func (d *ClientActivityExecutionDescription) GetSummary() (string, error) {
+	payload := d.RawExecutionInfo.GetUserMetadata().GetSummary()
```
cretz, line 322: Arguably no need to continually convert this on every call, but not a big deal

internal/internal_activity_client.go
```diff
@@ -0,0 +1,661 @@
...
+}
+
+func (h *clientActivityHandleImpl) Get(ctx context.Context, valuePtr any) error {
+	if err := h.client.ensureInitialized(ctx); err != nil {
+		return err
+	}
+
+	// repeatedly poll, the loop repeats until there's an outcome
+	for {
+		resp, err := h.client.interceptor.PollActivityResult(ctx, &ClientPollActivityResultInput{
```
cretz, line 380: Arguably this should also need to be called once and stored otherwise, but not a big deal

internal/internal_activity_client.go
```diff
@@ -0,0 +1,661 @@
...
+	}
+
+	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))
+	defer cancel()
+
+	resp, err := w.client.WorkflowService().StartActivityExecution(grpcCtx, request)
+
+	var runID string
+	if e, ok := err.(*serviceerror.ActivityExecutionAlreadyStarted); ok &&
+		in.Options.ActivityIDConflictPolicy == enumspb.ACTIVITY_ID_CONFLICT_POLICY_USE_EXISTING {
```
cretz: This conditional doesn't really make sense. If the conflict policy was use existing, you'd never get this error. Rather, we need to confirm if, like workflows, we default to already-exists returning the handle (w/ `WorkflowExecutionErrorWhenAlreadyStarted`-like option to change it) or not.

internal/internal_activity_client.go
```diff
@@ -0,0 +1,661 @@
...
+	grpcCtx, cancel := newGRPCContext(ctx, defaultGrpcRetryParameters(ctx), grpcLongPoll(true))
+	defer cancel()
+
+	request := &workflowservice.PollActivityExecutionRequest{
+		Namespace:  w.client.namespace,
+		ActivityId: in.ActivityID,
+		RunId:      in.RunID,
+	}
+
+	resp, err := w.client.WorkflowService().PollActivityExecution(grpcCtx, request)
```
cretz: I think this needs to loop while there is no outcome

yuandrew: added a loop `for resp.GetOutcome() != nil`

internal/internal_activity_client.go
yuandrew: Thanks for the callout, added `ActivityExecutionErrorWhenAlreadyStarted` to mirror workflow

internal/internal_activity_client.go
yuandrew: Sure, caching the result

internal/internal_activity_client.go
yuandrew: this mirrors `workflowRun` and `ScheduleHandle`, I think we should keep it for consistency

internal/interceptor.go
yuandrew: mirroring ClientPollWorkflowUpdateOutput

internal/interceptor.go
yuandrew: don't think so, made this string

go.mod
yuandrew: Bumped to 1.62.0

client/client.go
yuandrew: We do this for [ExecuteWorkflow](https://github.com/temporalio/sdk-go/blob/6c2f815cad201a71ff1c456db64a0b2b9e9dd163/client/client.go#L916), I think it's fine to keep it for consistency

client/client.go
yuandrew: Looks like this was an issue that's been around since forever, clarified in `CompleteActivity`/`CompleteActivityByID`/`CompleteActivityByActivityID` that we're referring to the activity's implementation, matching the doc for `ErrResultPending`

client/client.go
yuandrew: Sure, that sounds fine. I'm on the fence if we want to also have error return type, we wouldn't use it today, but for future proofing. LMK if you think we should add it in, I've kept it out for now

internal/internal_activity_client.go
yuandrew: caching these in `ClientActivityExecutionDescription`

client/client.go
cretz: `ActivityHandle` is not the same as `WorkflowRun`. When we add new operations for workflows, we make top-level methods, so we don't have to update that list.

Regardless, this is a perfect example of what I'm talking about. That list is missing `GetWithOptions`, but I wasn't around to predict that would happen then like I am now :-) It shouldn't be there either.

internal/internal_activity_client.go
```diff
@@ -0,0 +1,779 @@
...
+	ClientListActivitiesOptions struct {
+		Query string
+	}
+
+	// ClientListActivitiesResult contains the result of the ListActivities call.
+	//
+	// NOTE: Experimental
+	//
+	// Exposed as: [go.temporal.io/sdk/client.ListActivitiesResult]
+	ClientListActivitiesResult struct {
```
cretz, line 137: Pedantic, but could be called `ClientActivityList`, but meh, this is fine

client/client.go
cretz: I admit I like the future proofing of having error return type, even if we document that all errors currently come back in the iterator

internal/internal_activity_client.go
cretz: Makes sense, totally agree, I forget where I saw the inconsistency that prompted this

internal/internal_activity_client.go
cretz: This works for me, though may want to confirm with others on the standalone activity team they have no concerns

internal/internal_activity_client.go
yuandrew: Ended up adding interceptor support for describe/cancel/terminate

internal/internal_activity_client.go
yuandrew: Discussed offline, removing `ActivityExecutionErrorWhenAlreadyStarted` and instead relying on conflict policy, as that was added into workflows after `WorkflowExecutionErrorWhenAlreadyStarted`, we're diverging from mirroring workflows, but conflict policy should be sufficient here.

client/client.go
yuandrew: sounds good, I've added it in

client/client.go
yuandrew: yeah, that's fair, one less thing to keep synchronized, removed

PR #2164, opened 2026-01-22
Enhance SDK flags so flags are off by default when first introduced

yuandrew:
## What was changed
- Enhance the sdk flag logic to require manual flipping to enable a flag by default. 
- Added logic to automatically support env var overrides for SDK flags, as an emergancy mechanism for customers on release upgrades that flip a flag and reveal a new bug.
- centralized default flag enablement into `internal_flags.go`, instead of at the flag call site

## Why?
We require at least 1 release in between when a flag is introduced and when it can be flipped on, for backwards compatibility. Make it requirement more obvious/ergonomic. Also encapsulates flag flipping into a singular place (`internal_flags.go`), instead of wherever the flag is being used.

## Checklist

1. Closes 

2. How was this tested:

Added new tests

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes internal workflow flag gating/recording logic that influences deterministic behavior and history metadata; mistakes could alter when new semantics activate or are recorded across replay.
> 
> **Overview**
> Introduces a centralized `sdkFlagsAllowed` defaults map (with init-time `TEMPORAL_SDK_FLAG_<ID>=1|0` overrides) so **new SDK flags are OFF by default** unless the flag is already present in workflow history.
> 
> Refactors flag plumbing by renaming `newSDKFlags` to `newSDKFlagSet`, removing `GetFlag`, and tightening `TryUse` semantics to only *record/enable* flags when allowed; call sites for selector signal-drop behavior and `AwaitWithTimeout` timer-cancel now rely on `TryUse`.
> 
> Updates and expands unit/tests to cover env overrides, `TryUse` behavior, and selector signal-loss vs non-loss scenarios; integration selector tests are adjusted (one skipped) to reflect the new default-off rollout behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a000e1eeb2fb9b2c4df59b1c16b5ad968e1404cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/workflow_testsuite_test.go
+++ b/internal/workflow_testsuite_test.go
@@ -1113,6 +1113,7 @@ func TestMockCallWrapperNotBefore(t *testing.T) {
 		ctx = WithLocalActivityOptions(ctx, LocalActivityOptions{
 			ScheduleToCloseTimeout: time.Hour,
 			StartToCloseTimeout:    time.Hour,
+			RetryPolicy:            &RetryPolicy{MaximumAttempts: 1},
 		})
 		var result string
 		return ExecuteLocalActivity(ctx, "namedActivity", "call2").Get(ctx, &result)


internal/internal_flags.go
```diff
@@ -32,9 +35,87 @@ const (
...
+		if !flag.isValid() {
+			continue
+		}
+		envKey := fmt.Sprintf("TEMPORAL_SDK_FLAG_%d", i)
+		if val := os.Getenv(envKey); val != "" {
+			if parsed, err := strconv.ParseBool(val); err == nil {
+				sdkFlagOverrides[flag] = parsed
+			}
+		}
+	}
```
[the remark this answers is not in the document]
yuandrew: yikes, good catch

internal/internal_flags.go
```diff
@@ -32,9 +35,88 @@ const (
...
+
+// Env var format: TEMPORAL_SDK_FLAG_<FLAG_ID>=true|false
+// Example: TEMPORAL_SDK_FLAG_5=true (enables SDKFlagBlockedSelectorSignalReceive)
+//
+// Bulk override: TEMPORAL_SDK_FLAGS_ENABLE=1,5,6 or TEMPORAL_SDK_FLAGS_DISABLE=1,5
+//
+// NOTE: Using env vars to set flags is strongly discouraged, but this utility is built in
+// as an emergency mechanism in case there is an unanticipated bug with a flag flip, where
+// users would not have to wait until the next release to upgrade.
+func loadFlagOverrides() {
```
cretz: Would recommend just a simpler single env var per flag that can be set to 1 or 0 to enable/disable. Also, if you want it testable, have it accept the map to mutate so you can test it without having global side effects

internal/internal_flags.go
```diff
@@ -32,9 +35,88 @@ const (
...
+		envKey := fmt.Sprintf("TEMPORAL_SDK_FLAG_%d", i)
+		if val := os.Getenv(envKey); val != "" {
+			if parsed, err := strconv.ParseBool(val); err == nil {
+				sdkFlagOverrides[flag] = parsed
+			}
+		}
+	}
+}
+
+func getFlagOverride(flag sdkFlag) (enabled bool, hasOverride bool) {
```
cretz: Is an entire function that just does what Go does for reading from a map worth it?

internal/internal_flags.go
```diff
@@ -32,9 +35,88 @@ const (
...
-// unblockSelectorSignal exists to allow us to configure the default behavior of
-// SDKFlagBlockedSelectorSignalReceive. This is primarily useful with tests.
-var unblockSelectorSignal = os.Getenv("UNBLOCK_SIGNAL_SELECTOR") != ""
+// enabledByDefault returns whether this flag should be enabled by default for new
+// workflows. Flags must be explicitly listed to be enabled - unlisted flags default
+// to false. This ensures new flags don't accidentally ship enabled.
+//
+// We need to wait at least one release between a new SDK flag being introduced
+// and when it's enabled, for to backwards compatibility
+func (f sdkFlag) enabledByDefault() bool {
```
cretz: I'd make this simpler by just having an `var sdkFlagsEnabled map[sdkFlag]bool` and set a map literal with the defaults and then have an `init()` just loop and adjust as needed. Much less code, much simpler to comprehend.

internal/internal_flags.go
```diff
@@ -125,8 +222,23 @@ func (sf *sdkFlags) gatherNewSDKFlags() []sdkFlag {
 	return flags
 }
 
-// SetUnblockSelectorSignal toggles the flag to unblock the selector signal.
+// SetFlagOverrideForTest sets a flag override for testing. Returns a cleanup function.
+// For test use only.
+func SetFlagOverrideForTest(flag sdkFlag, enabled bool) func() {
```
cretz: You don't need this or the next func, the test files can set the map directly and defer unset what they did

internal/internal_flags.go
```diff
@@ -12,51 +12,72 @@ import (
 type sdkFlag uint32
 
 const (
-	SDKFlagUnset sdkFlag = 0
+	SDKFlagUnset sdkFlag = iota
```
cretz: Hrmm, I don't believe we should lean on iota here, it's easy to make a mistake. Can we leave all of this code untouched?

internal/internal_flags.go
```diff
@@ -12,51 +12,72 @@ import (
...
+
+// loadFlagOverridesFromEnv loads flag overrides from environment variables into the provided map.
+// Env var format: TEMPORAL_SDK_FLAG_<FLAG_ID>=1|0
+// Example: TEMPORAL_SDK_FLAG_5=1 (enables SDKFlagBlockedSelectorSignalReceive)
+//
+// NOTE: Using env vars to set flags is strongly discouraged, but this utility is built in
+// as an emergency mechanism in case there is an unanticipated bug with a flag flip, so
+// users would not have to wait until the next release to upgrade.
+func loadFlagOverridesFromEnv(overrides map[sdkFlag]bool) {
+	for flag := SDKFlagUnset + 1; flag < SDKFlagUnknown; flag++ {
```
cretz: IMO should change the param name to `defaults`, and just loop over those keys, checking their flag instead of looping over all possible

internal/internal_flags.go
```diff
@@ -12,51 +12,72 @@ import (
...
-	case uint32(SDKPriorityUpdateHandling):
-		return SDKPriorityUpdateHandling
-	case uint32(SDKFlagBlockedSelectorSignalReceive):
-		return SDKFlagBlockedSelectorSignalReceive
-	default:
-		return SDKFlagUnknown
+	if value > 0 && value < uint32(SDKFlagUnknown) {
+		return sdkFlag(value)
 	}
+	return SDKFlagUnknown
```
cretz: Just use your map. In fact, could call the map `sdkFlags` instead of `sdkFlagDefaults` to make that clearer

internal/internal_flags.go
yuandrew: ahh good point, that would also get rid of the desire to move these vars to `iota`

internal/internal_flags.go
yuandrew: easy to make a mistake as in accidentally adding something in the wrong spot and shifting element values?

but also reverted from iota, since your suggestion below gets rid of the reason I moved to iota in the first place

internal/internal_flags.go
yuandrew: sure, calling this `sdkFlags` and below renamed the existing `sdkFlags` over to `sdkFlagSet`

yuandrew: actually, I like keeping `sdkFlags` to the existing struct, and this new map could be `sdkFlagsAllowed`

internal/internal_flags_test.go
```diff
@@ -14,26 +14,108 @@ var (
...
+		wantSet bool
+	}{
+		{"1 enables", "1", true, true},
+		{"0 disables", "0", false, true},
+		{"invalid value ignored", "true", false, false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv("TEMPORAL_SDK_FLAG_5", tt.envVal)
```
cretz, line 30: Pedantic, but should unset side-effecting things like this at end of test IMO

yuandrew: TIL, `t.Setenv` does this 😄 

> Setenv calls os.Setenv(key, value) and uses Cleanup to restore the environment variable to its original value after the test.

PR #2171, opened 2026-01-28
Versioning/upgrade on can

THardy98:
## What was changed
Expose `InitialVersioningBehavior` on
- `ContinueAsNewError`
- `ContinueAsNewErrorOptions`
this specifies the versioning behavior that the first task of the new run should use.

Expose `	continueAsNewSuggestedReasons` on `WorkflowInfo`, which describes the reason(s) that the CAN is suggested.

## Why?
Provides workflow version control on CAN boundaries. This change in particular allows users to auto-upgrade on CAN.

1. Part of: https://github.com/temporalio/features/issues/738
2. How was this tested:
Integration tests

3. Any docs updates needed?
Yes.

### DIFF
--- a/test/worker_deployment_test.go
+++ b/test/worker_deployment_test.go
@@ -1331,7 +1331,7 @@ func (ts *WorkerDeploymentTestSuite) TestDeleteDeployment() {
 	}, 305*time.Second, 1000*time.Millisecond)
 }
 
-func (ts *WorkerDeploymentTestSuite) TestContinueAsNewWithWithVersionUpgrade() {
+func (ts *WorkerDeploymentTestSuite) TestContinueAsNewWithVersionUpgrade() {
 	ctx, cancel := context.WithTimeout(context.Background(), ctxTimeout)
 	defer cancel()
 


internal/error.go
```diff
@@ -193,6 +193,10 @@ type (
 		// compatible build ID or not. See VersioningIntent.
 		VersioningIntent VersioningIntent
 
+		// InitialVersioningBehavior specifies the versioning behavior that the first task of the new run should use.
+		// For example, choose to AutoUpgrade on continue-as-new instead of inheriting the pinned version of the previous run.
+		InitialVersioningBehavior ContinueAsNewVersioningBehavior
```
cretz, line 201: This needs to be marked experimental IMO

internal/workflow.go
```diff
@@ -1450,6 +1512,13 @@ func (wInfo *WorkflowInfo) GetContinueAsNewSuggested() bool {
 	return wInfo.continueAsNewSuggested
 }
 
+// GetContinueAsNewSuggestedReasons returns a list of reasons why continue as new is suggested,
+// if the server is configured to suggest continue as new and it is suggested.
+// This value may change throughout the life of the workflow.
+func (wInfo *WorkflowInfo) GetContinueAsNewSuggestedReasons() []ContinueAsNewSuggestedReason {
```
cretz, line 1518: This needs to be marked experimental IMO

test/worker_deployment_test.go
```diff
@@ -86,6 +86,38 @@ func (ts *WorkerDeploymentTestSuite) waitForWorkerDeploymentVersion(
...
+		switch resp.GetWorkerDeploymentInfo().GetRoutingConfigUpdateState() {
+		case enumspb.ROUTING_CONFIG_UPDATE_STATE_COMPLETED:
+			return true
+		case enumspb.ROUTING_CONFIG_UPDATE_STATE_UNSPECIFIED:
+			return true // not implemented
+		case enumspb.ROUTING_CONFIG_UPDATE_STATE_IN_PROGRESS:
+			return false
+		}
+		return false
+	}, 10*time.Second, 300*time.Millisecond)
```
cretz: Would like to see all of the worker deployment "eventually" stuff reduces to max timeout 5s and interval 100ms if we can, but not blocking

internal/internal_event_handlers.go
```diff
@@ -2162,3 +2165,13 @@ func (weh *workflowExecutionEventHandlerImpl) protocolConstructorForMessage(
 	}
 	return nil, fmt.Errorf("unsupported protocol: %v", protoName)
 }
+
+func convertContinueAsNewSuggestedReasonsFromProto(
+	reasons []enumspb.SuggestContinueAsNewReason,
+) []ContinueAsNewSuggestedReason {
+	var converted []ContinueAsNewSuggestedReason
```
cretz: Pedantic, but you should pre-set the capacity here if you know it

internal/error.go
```diff
@@ -555,6 +563,9 @@ func NewContinueAsNewErrorWithOptions(ctx Context, options ContinueAsNewErrorOpt
 		if options.RetryPolicy != nil {
 			continueAsNewErr.RetryPolicy = options.RetryPolicy
 		}
+		if options.InitialVersioningBehavior != ContinueAsNewVersioningBehaviorUnspecified {
```
yuandrew: nit: I don't think we need this if check, since the default of both `continueAsNewErr.InitialVersioningBehavior` and `options.InitialVersioningBehavior` are the same

contrib/datadog/go.mod
```diff
@@ -81,7 +81,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.37.0 // indirect
 	go.opentelemetry.io/otel/sdk v1.37.0 // indirect
 	go.opentelemetry.io/otel/trace v1.37.0 // indirect
-	go.temporal.io/api v1.59.0 // indirect
+	go.temporal.io/api v1.60.0 // indirect
```
carlydf, line 84: should pull in v1.62.2 to implement this API change https://github.com/temporalio/api/pull/709

PR #2175, opened 2026-02-04
💥 Payload limit options and validation

### LINKED ISSUE #2165: SDK should fail workflow task if payloads size it known to be too large - Go
Parent issue: https://github.com/temporalio/features/issues/723

### LINKED ISSUE #2167: Warn if the SDK tried to send a payload above a specific size - Go
Parent issue: https://github.com/temporalio/features/issues/701

jmaeagle99:
## What was changed

- 💥 Update worker to get payload error limits from Temporal server and enforce those limits by default. The SDK will now eagerly fail a task that has too large payloads instead of uploading them to the server and having the server fail the workflow. If the server doesn't report error limits, error limits are not enforced in the worker.
- Add client payload limits options for setting payload size warning thresholds. Warnings are emitted in client, worker, and replayer for any payloads that are above these thresholds.

### 💥 Behavioral change details

Within workers, if a payload exceeds the server limits, the worker will eagerly fail the current task instead of uploading the object with the too large payload. This allows the task to be retried instead of entirely failing the workflow from within the server.

Customers who use gRPC proxies that alter payloads before they are passed to the server (encryption in the proxy, offloading to external storage within proxy) should disable this new behavior on the worker using the new `DisablePayloadErrorLimit` option on the worker. For example:

```go
c, err := client.Dial(...)

w := worker.New(c, "queue-name", worker.Options{
    DisablePayloadErrorLimit: true,
})
```

### Examples
Log output when an activity attempts to return a result that exceeds the warning limit:
```
WARN  [TMPRL1103] Attempted to upload payloads with size that exceeded the warning limit. Namespace integration-test-namespace TaskQueue tq-238573ce-2bd2-4ee3-a455-aaf22d640463-TestIntegrationSuite/TestPayloadSizeWarningCustomSize WorkerID <worker_id> WorkerType payload-size-warning-activity-result WorkflowID TestIntegrationSuite/TestPayloadSizeWarningCustomSize RunID 019c2a6b-3833-758d-a30c-4796c230ad05 ActivityID 5 Attempt 1 ActivityType large-payload-activity
```
The above example will upload the payload to the server as normal.

Log output when a workflow attempts to return a result or provide an activity some input that exceeds the warning limit:
```
WARN  [TMPRL1103] Attempted to upload payloads with size that exceeded the warning limit. Namespace integration-test-namespace TaskQueue tq-238573ce-2bd2-4ee3-a455-aaf22d640463-TestIntegrationSuite/TestPayloadSizeWarningCustomSize WorkerID <worker_id> WorkflowType payload-size-warning-activity-result WorkflowID TestIntegrationSuite/TestPayloadSizeWarningCustomSize RunID 019c2a6b-3833-758d-a30c-4796c230ad05
```
The above example will upload the payload to the server as normal.

Log output when a client attempts to provide input to a workflow that exceeds the warning limit:
```
WARN  [TMPRL1103] Attempted to upload payloads with size that exceeded the warning limit.
```
The above example will upload the payload to the server as normal.

## Why?

Users need to know when payload sizes are approaching or have exceeded size limits. This will help prevent workflow outages and inform users to adjust their workflows to make use of alternate storage methods or to break down their payloads more granularly.

## Checklist

1. Closes #2165
2. Closes #2167

3. How was this tested: Unit tests

4. Any docs updates needed? Yes


internal/payload_limits.go
```diff
@@ -0,0 +1,11 @@
+package internal
+
+type (
+	// Options for when payload sizes exceed limits.
+	//
+	// Exposed as: [go.temporal.io/sdk/client.PayloadLimitOptions]
+	PayloadLimitOptions struct {
```
jmaeagle99: This is supposed to be similar to https://github.com/temporalio/sdk-python/pull/1288/changes#diff-dca556db4f4a9a3950b33cf90d75886e8c3216b0955fa950e71cf60aadee3946R1211, but the payload visitor in Go doesn't allow access to Memo fields. The visitor (https://github.com/temporalio/api-go/blob/master/proxy/interceptor.go#L652) will pull out the fields of the Memo and pass that `map[string]*common.Payload` back to the visitor and is handled at https://github.com/temporalio/api-go/blob/master/proxy/interceptor.go#L266. The visitor will see each payload individually rather than as a collection. The collection is what is needed to validate the the memo size.

Thinking of some options to allow this alternative behavior for memos:
- Add a new bool field to VisitPayloadsOptions that means "I want all payloads of a memo at once". Maybe call it `VisitMemoPayloadsAsSequence bool`. This would allow current uses of the VisitPayloadsOptions` struct to work as-is if they are using named parameters, but would break those using position parameters (less likely).
- Replicate the visitor behavior in the current repo but just for memo fields. Not a great option from a maintenance perspective.
- Don't check memo fields for their limits. This would be a behavioral difference across SDKs.

Looking for other suggestions or alternatives.

internal/payload_limit_data_converter.go
cretz, line 1: Does all of this have to be exported outside of the internal package? It may seem innocuous, but sadly our type alias posture makes it worth not adding to the public https://pkg.go.dev/go.temporal.io/sdk/internal API surface without a good reason.

Or is this expected to be exposed? It does beg the question - for the significant number of our users that use data converters directly for whatever reason (e.g. in a payload), how can they apply the same payload limits we do internally?

internal/payload_limits.go
cretz: > The visitor will see each payload individually rather than as a collection. The collection is what is needed to validate the the memo size.

Why is the collection what is needed? Are we sure we're not tying payload size validation expectations, which is meant as a good best practice for a user, too much to some really old internal server algorithms? Large memo is not that commonly used, not sure any error/warnings need to have relation to server.

Granted I think this is general. IMO none of this project should have anything to do with server or what server happens to have set in the past except as a way to come up with sensible client-side defaults. I don't believe we should receive the limits from server either just to help the rare self-hosted users that have customized the size.

internal/client.go
```diff
@@ -1144,6 +1147,8 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
 		}
 	}
 
+	dataConverter, _ := NewPayloadLimitDataConverterWithOptions(options.DataConverter, options.Logger, options.PayloadLimits)
```
cretz: Can this wait for external storage? I would like to see the extension points applied for external storage to understand if payload limits can leverage them. We may find that data converter is not the place and do the significant refactoring to pass around a thing that contains data converter, external storage, and payload limits.

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+func (e PayloadSizeError) Error() string {
+	return e.message
+}
+
+type PayloadErrorLimits struct {
+	PayloadSizeError int64
+}
+
+// NOTE: This func returns the data converter and a func callback allowing the setting of error limits after the converter has been created.
+func NewPayloadLimitDataConverter(innerConverter converter.DataConverter, logger log.Logger) (converter.DataConverter, func(*PayloadErrorLimits)) {
```
cretz: This is internal only code right? Don't need "overloads" or compatibility, just have a single `newPayloadLimitDataConverter` that accepts what is needed or accepts a `payloadLimitDataConverterOptions` with everything needed. And it could just return the actual `*PayloadLimitDataConverter` and the caller can do something with the `SetErrorLimits` func it if wants (or just make their own func).

internal/internal_nexus_worker.go
```diff
@@ -78,10 +78,6 @@ func newNexusWorker(opts nexusWorkerOptions) (*nexusWorker, error) {
 
 // Start the worker.
 func (w *nexusWorker) Start() error {
-	err := verifyNamespaceExist(w.workflowService, w.executionParameters.MetricsHandler, w.executionParameters.Namespace, w.worker.logger)
```
cretz, line 81: Please coordinate with @yuandrew who is also adjusting some of this

internal/internal_task_handlers.go
```diff
@@ -1895,6 +1901,36 @@ func (wth *workflowTaskHandlerImpl) completeWorkflow(
 			RetryPolicy:         convertToPBRetryPolicy(retryPolicy),
 			InheritBuildId:      useCompat,
 		}}
+	} else if errors.As(workflowContext.err, &payloadSizeErr) {
```
cretz: This logic being here signifies to me that arguably a payload size issue in a workflow should be a panic not a traditional error. Task failures in Go are panics so they are not reacted to as if they were normal errors that might complete the workflow (which means they may trigger user cleanup code that swallows this error or something).

internal/internal_worker.go
```diff
@@ -287,16 +290,21 @@ func verifyNamespaceExist(
...
-	_, err := client.DescribeNamespace(grpcCtx, &workflowservice.DescribeNamespaceRequest{Namespace: namespace})
-	return err
+	response, err := client.DescribeNamespace(grpcCtx, &workflowservice.DescribeNamespaceRequest{Namespace: namespace})
+	if err != nil {
+		return nil, err
+	}
+	if response != nil && response.NamespaceInfo != nil {
+		return response.NamespaceInfo.Limits, nil
+	}
+	return nil, nil
```
cretz: Pedantic, but can shorten to `return response.GetNamespaceInfo().GetLimits(), err` if you wanted

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+	errorLimits    *PayloadErrorLimits
+	options        PayloadLimitDataConverterOptions
+}
+
+type PayloadLimitDataConverterOptions struct {
+	Logger             log.Logger
+	PayloadSizeWarning int
+}
+
+type PayloadSizeError struct {
```
cretz: This should be `ErrPayloadSize` and if it's something that can be handled by a user, it needs to be public. But see my other note about this needing to be a panic in workflows.

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+		options: PayloadLimitDataConverterOptions{
+			Logger:             logger,
+			PayloadSizeWarning: payloadSizeWarning,
+		},
+	}
+	return dataConverter, dataConverter.SetErrorLimits
+}
+
+func (converter *PayloadLimitDataConverter) SetErrorLimits(errorLimits *PayloadErrorLimits) {
+	converter.errorLimits = errorLimits
```
cretz: So while _technically_ this should be set before ever used, it is a bit more comfortable from a code safety/reader POV if this is guarded by a mutex (or `atomic.Pointer[PayloadErrorLimits]` or whatever)

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (converter *PayloadLimitDataConverter) ToPayload(value interface{}) (*commonpb.Payload, error) {
+	payload, err := converter.innerConverter.ToPayload(value)
+	if err != nil {
+		return nil, err
+	}
+	payloads := &commonpb.Payloads{
+		Payloads: []*commonpb.Payload{payload},
+	}
+	err = proxy.VisitPayloads(context.TODO(), payloads, proxy.VisitPayloadsOptions{
```
cretz: I do not believe we should use the visitor here, I believe we should invoke the size check explicitly on single/multiple payloads. The indirection of calling the visitor here adds overhead (albeit likely negligible) and potential future complication I don't think is necessary. Just have a simple `func (*PayloadLimitDataConverter) check(...*commonpb.Payload) error`, and if you need to have a visitor form of that, you can and have it invoke this one.

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+	"context"
+
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/api/proxy"
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/log"
+)
+
+type PayloadLimitDataConverter struct {
+	innerConverter converter.DataConverter
```
cretz: FWIW, you could embed this and therefore only need to explicitly provide functions you want to "override"

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (converter *PayloadLimitDataConverter) WithWorkflowContext(ctx Context) converter.DataConverter {
+	logger := converter.options.Logger
+	if workflowEnvironmentAny := ctx.Value(workflowEnvironmentContextKey); workflowEnvironmentAny != nil {
+		workflowInfo := workflowEnvironmentAny.(WorkflowEnvironment).WorkflowInfo()
+		logger = log.With(logger,
+			tagWorkflowType, workflowInfo.WorkflowType.Name,
+			tagWorkflowID, workflowInfo.WorkflowExecution.ID,
+			tagRunID, workflowInfo.WorkflowExecution.RunID)
+	}
```
cretz: Can probably just `logger := GetLogger(ctx)`. This also provides replay protection.

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+
+	innerConverter := converter.innerConverter
+	if contextAwareInnerConverter, ok := converter.innerConverter.(ContextAware); ok {
+		innerConverter = contextAwareInnerConverter.WithWorkflowContext(ctx)
+	}
+
+	if logger == converter.options.Logger && innerConverter == converter.innerConverter {
```
cretz: With how Go equality rules work, this is not straightforward identity comparison. Also, basically all of the time the logger will be different. It is cheap to just always return a new converter from this function. Same for activity form.

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+		errorLimits:    converter.errorLimits,
+		options: PayloadLimitDataConverterOptions{
+			Logger:             logger,
+			PayloadSizeWarning: converter.options.PayloadSizeWarning,
+		},
+	}
+}
+
+func (converter *PayloadLimitDataConverter) WithContext(ctx context.Context) converter.DataConverter {
+	logger := converter.options.Logger
```
cretz: Same here as before, can probably just `logger := GetActivityLogger(ctx)`

internal/payload_limit_data_converter.go
```diff
@@ -0,0 +1,191 @@
...
+	innerConverter := converter.innerConverter
+	if contextAwareInnerConverter, ok := converter.innerConverter.(ContextAware); ok {
+		innerConverter = contextAwareInnerConverter.WithWorkflowContext(ctx)
+	}
+
+	if logger == converter.options.Logger && innerConverter == converter.innerConverter {
+		return converter
+	}
+
+	return &PayloadLimitDataConverter{
```
cretz: Note, it may be here that you can configure payload data converter to panic on too-large payload. I admit I didn't dig into code to see if everywhere we do conversion is safe for panic.

internal/payload_limit_data_converter.go
jmaeagle99: It does not need to be expose. I made everything except for PayloadLimitOptions internal.

internal/payload_limit_data_converter.go
jmaeagle99: I removed the "overload" and just pass empty options.

internal/payload_limit_data_converter.go
jmaeagle99: Yeah, you're right, not sure why I was using the visitor here.

internal/payload_limit_data_converter.go
jmaeagle99: WithContext is getting called outside of activities, so I cannot use `GetActivityLogger` here without a panic occurring.

internal/payload_limit_data_converter.go
jmaeagle99: I still have the converter return an error but that is specially handled (as needed) within workflow tasks to cause a panic. Updated errorToFailWorkflowTask to specifically handle the `payloadSizeError` type and set the corresponding cause.

internal/payload_limit_data_converter.go
jmaeagle99: I've updated the error to carry the payload size and limit information for logging purposes, so I won't be able to create a reusable instance of the error.

jmaeagle99: And the error type is now internal as well.

internal/client.go
jmaeagle99: External storage and payload handles _can_ be data converters that are chained together. Not proposing that's what needs to be done, but it looks to be plausible. See https://github.com/jmaeagle99/temporalio-sdk-go/blob/payloads-all-up/internal/client.go#L1206 where I create this chain in the client and an integration test that uses payload handles in https://github.com/jmaeagle99/temporalio-sdk-go/blob/payloads-all-up/test/integration_test.go#L8299; these integration tests were updated to configure external storage and payload limits.

cretz: Makes sense. I think understanding that can help us confirm what is best here. We may find that we have some kind of super-object that holds data converter, external payload provider, and payload limits, and refactor everything to use that instead of the user-facing data converter interface.

But maybe the user-facing data converter interface does make sense, but at the least, might as well have the Temporal-internal wrapper be a single one handling both limits and external payloads.

internal/internal_worker.go
```diff
@@ -1062,6 +1059,9 @@ func (we *workflowExecutor) Execute(ctx Context, input *commonpb.Payloads) (*com
 	if err == nil && result != nil {
 		serializedResult, err = encodeArg(dataConverter, result)
 	}
+	if errPayloadSize, payloadSize := err.(payloadSizeError); payloadSize {
```
cretz: Payload conversion is done in many places in a workflow. Granted since we only care about the encoding side and not decoding, that cuts it in half, but it still applies to workflow result, update result, activity arg, nexus arg, child workflow arg, external signal arg, memo upsert, and maybe others I am forgetting. Query result can still be error technically, but might as well make it panic too, we `recover()` those and make them errors to caller.

I think it may be wiser to adjust the `WithWorkflowContext` to tell the converter itself to panic henceforth instead of error. Either way, due to how this touches users, it may be worth an integration tests confirming those ~7 places all fail tasks on too-large payload.

jmaeagle99: There are integration tests that cover a few of these but I can add the rest. Good point in knowing that if WithWorkflowContext is called, then it can panic.

I do recall seeing some comment about MemoUpsert possibly erroneously only using the default data converter. I'll link to it when I find it again.

internal/client.go
jmaeagle99: Makes sense. And I was thinking of the internal super-object or super data converter, just didn't want to invest into that just yet. But I think this shows we may be able to change it in these few places (if just super converter) or trickle it down by replacing the internal converter fields (if doing super object).

internal/payload_processing_converter.go
```diff
@@ -0,0 +1,166 @@
...
+	converter.DataConverter
+	errorLimits        atomic.Pointer[payloadErrorLimits]
+	logger             log.Logger
+	payloadSizeWarning int
+}
+
+// NOTE: This func returns the data converter and a func callback allowing the setting of error limits after the converter has been created.
+func newPayloadProcessingDataConverter(innerConverter converter.DataConverter, logger log.Logger, options PayloadLimitOptions) (converter.DataConverter, func(*payloadErrorLimits)) {
+	payloadSizeWarning := 512 * kb
+	if options.PayloadSizeWarning != 0 {
```
jmaeagle99, line 55: Is there a better way of having a default value (Go doesn't allow default values on structs) and allowing customers to specify their own? I would like to use 0 to mean "disabled". Do we need a `DisablePayloadSizeWarning" field to actually disable it?

In Python, we set the field to the default value and allow customers to set it to 0 to mean "disabled".

internal/payload_limits.go
jmaeagle99: I'm not sure why I involved the visitor. It may have been left over from an earlier attempt and I didn't think about whether it was truly necessary.

I think what I can do is extend the `payloadProcessorDataConverter` (this will be the one that contains all of external storage, handles, and limits eventually) to be able to convert memos via new functions on this converter. I'd rather to either type test the converter interface at the memo conversion locations or propagate the converter type through the internal code. Obviously open to other ideas.

internal/internal_worker.go
jmaeagle99: I've change the converter to panic when used within a workflow context and added some more tests. Still need to add one for query result since the panic handle is not handling data converter panics. I also had to fix memo serialization since it was not using the converter passed to it nor was the workflow converter being passed in. I can make more adjustments to that if it needs to be the default data converter but wrapped with this new converter.

cretz: > I also had to fix memo serialization since it was not using the converter passed to it nor was the workflow converter being passed in

Even if it's the right thing, I am afraid this is a backwards incompatible change. We may need to bring it up with team, because in many cases, we recognize that even though something is right, it's important not breaking existing code.

internal/payload_processing_converter.go
cretz: In Go, it's most idiomatic to make 0 (i.e. unset) represent your default, and therefore you need to use something besides 0 for any magic value besides the default. Documenting that "any value less than zero disables this check" is plenty normal in Go. _Technically_ you could accept an `*int` as a "nilable int", but it is not as idiomatic.

internal/payload_processing_converter.go
```diff
@@ -0,0 +1,170 @@
...
+
+// payloadProcessingDataConverter is a converter that wraps another data converter and applies post conversion operations
+// to payloads, such as payload size limit logging and enforcement. Future operations may be added to create an internal
+// centralized pipeline of transformations and validations in one converter.
+type payloadProcessingDataConverter struct {
+	converter.DataConverter
+	errorLimits        atomic.Pointer[payloadErrorLimits]
+	logger             log.Logger
+	payloadSizeWarning int
+	errorBehavior      func(error) error
```
cretz: A bit clearer in Go to just make this a `panicOnError` bool and have `checkPayloadsSize` use that compared to passing around callables and using super flexible anon funcs for internal-only constructs that can only have two states.

cretz (commented):
The only blocker for me is memo compatibility thing. That this is in a common processing data converter that we are pretty sure the external storage provider stuff will use, and almost all of this is private, that's enough for me.

internal/payload_limit_data_converter.go
cretz: You can call `IsActivity(ctx)` first, and only override this default value with the `GetActivityLogger(ctx)` if true

internal/payload_limits.go
cretz: 👍 I vote checking whole-memo size at the memo serialization sites (still leaving per-payload tests as is which is fine)

internal/payload_processing_converter.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// NOTE: This func returns the data converter and a func callback allowing the setting of error limits after the converter has been created.
```
yuandrew: nit: do we need the `NOTE:` part? There isn't any other part of a comment to draw emphasis from

client/client.go
```diff
@@ -218,6 +218,9 @@ type (
 	// ConnectionOptions are optional parameters that can be specified in ClientOptions
 	ConnectionOptions = internal.ConnectionOptions
 
+	// PayloadLimitOptions are optional payload size limits that can be specified in ClientOptions.
```
yuandrew, line 221: Do we need to mark any of these fields as experimental?

internal/payload_processing_converter.go
```diff
@@ -0,0 +1,170 @@
...
+func (c *payloadProcessingDataConverter) SetErrorLimits(errorLimits *payloadErrorLimits) {
+	c.errorLimits.Store(errorLimits)
+}
+
+func (c *payloadProcessingDataConverter) ToPayload(value interface{}) (*commonpb.Payload, error) {
+	payload, err := c.DataConverter.ToPayload(value)
+	if err != nil {
+		return nil, err
+	}
+	// Other operations can be inserted here.
```
yuandrew, line 77: Does `other operations` here refer to future work we're adding in later?

internal/internal_nexus_worker.go
yuandrew: https://github.com/temporalio/sdk-go/pull/2186/files#r2794889332

nice, had similar ideas. We'll probably want to cache the `NamespaceInfo_Limits` onto the client

internal/internal_workflow_client.go
```diff
@@ -1662,8 +1664,7 @@ func getWorkflowMemo(input map[string]interface{}, dc converter.DataConverter) (
 
 	memo := make(map[string]*commonpb.Payload)
 	for k, v := range input {
-		// TODO (shtin): use dc here???
-		memoBytes, err := converter.GetDefaultDataConverter().ToPayload(v)
+		memoBytes, err := dc.ToPayload(v)
```
yuandrew: Like Chad mentioned, this is a breaking change that requires an SDK flag to change its behavior, I have https://github.com/temporalio/sdk-go/pull/2121/files open. I haven't had a chance to do some final validation to merge, just want to be extra sure we don't break everyone's histories, but we have to be pretty careful here

internal/payload_processing_converter.go
jmaeagle99: Yes

internal/payload_processing_converter.go
jmaeagle99: I'll remove it.

internal/internal_workflow_client.go
jmaeagle99: I'll work on adding the memo size warning limit now that I have a path forward. I'll think about whether to wait for your SDK flag PR to be finished or be able to merge without changing the data converter but have it light up when the the flag is available.

internal/internal_worker.go
jmaeagle99: I've reverted my change to use the configured data converter instead of the default and was able to have the memo participate in size checks. So this should preserve the existing behavior with the limit check added.

internal/internal_workflow_client.go
jmaeagle99: I've reverted my change to use the configured data converter instead of the default and was able to have the memo participate in size checks. So this should preserve the existing behavior with the limit check added.

When you complete the SDK flag PR, this should then be able to validate both the individual payloads and the overall memo.

client/client.go
jmaeagle99: I wasn't planning on doing so. This is an on-by-default behavior that customers can only opt-out of for error limit checks by setting `DisablePayloadErrorLimit` to `true` on the worker.

cretz (commented):
This PR looks good to me.

Not marking approved yet because 1) I want to let @yuandrew review, and 2) not sure we need to urgently merge before we get a PR on top of this for external payload stuff (but not a strong opinion, if team decides we can merge before waiting for that PR, I'm ok with that too).

PR #2176, opened 2026-02-05
Remove experimental warning for worker deployment versioning and envconfig

THardy98:
## What was changed
DWISOTT:
- remove experimental warnings for worker deployment API
- remove experimental warnings for (client) environment configuration
- add deprecated warnings/markings to old worker deployment APIs

## Why?
Worker deployment versioning and envconfig are GA

1. Part of https://github.com/temporalio/features/issues/744

3. Any docs updates needed?
Yes?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,15 +81,15 @@ jobs:
         if: ${{ matrix.testDockerCompose }}
         uses: actions/checkout@v4
         with:
-          repository: temporalio/docker-compose
-          path: ./docker-compose
+          repository: temporalio/samples-server
+          path: ./samples-server
 
       - name: Docker compose - start
         if: ${{ matrix.testDockerCompose }}
         run: |
-          cp ./.github/workflows/docker/docker-compose.override.yaml ./docker-compose/docker-compose.override.yaml
-          cp ./.github/workflows/docker/dynamic-config-custom.yaml ./docker-compose/dynamicconfig/dynamic-config-custom.yaml
-          docker compose --project-directory ./docker-compose up &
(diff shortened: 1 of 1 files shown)


client/client.go
cretz, line 1: All of the worker deployment client stuff should stay experimental. It is only in Go and arguably shouldn't even be there (but if we want it to remain, it is experimental until in all SDKs). Only the literal versioning options that are set on existing constructs and are available in all SDKs should be non-experimental.

THardy98: Ah yes - I overlooked this after removing these tags across all SDKs, reverted.

client/client.go
```diff
@@ -588,46 +588,32 @@ type (
 	DeploymentClient = internal.DeploymentClient
 
 	// UpdateWorkflowExecutionOptionsRequest is a request for [client.Client.UpdateWorkflowExecutionOptions].
-	//
-	// NOTE: Experimental
 	UpdateWorkflowExecutionOptionsRequest = internal.UpdateWorkflowExecutionOptionsRequest
```
cretz, line 592: I think workflow execution option updating should still be considered experimental too until it's either in all SDKs or decided to be removed from this one. Not sure it's on SDK's radar, may deserve a feature repo issue or internal SDK team issue to decide its fate.

THardy98: I've added it back to update wf exec options, was it just update in particular?

cretz: It also applies to the objects present only for the purpose of those worker deployment client calls.

Anything relating to the worker deployment client remains experimental. The best way to think about it, as you're working across SDKs, is that if anything here is not present in other SDKs, it should remain experimental.

THardy98: done - basically meant reverted the changes in `client.go` . Marked the `internal` usages as experimental

Issue #1496, opened 2024-05-30
Workflow outcome is inconsistent between test env and real server


bergundy:
When running a workflow that starts a child and cancels it:

```go
wf := func(ctx workflow.Context) error {
	childCtx, cancel := workflow.WithCancel(ctx)
	defer cancel()
	fut := workflow.ExecuteChildWorkflow(childCtx, waitForCancelWorkflow)
	if err := fut.GetChildWorkflowExecution().Get(ctx, nil); err != nil {
		return err
	}
	cancel()
	return fut.Get(ctx, nil)
}

func waitForCancelWorkflow(ctx workflow.Context) error {
	return workflow.Await(ctx, func() bool { return false })
}
```

I'm getting inconsistent outcomes:

With server:

```
workflow execution error (type: func1, workflowID: 1eb1f045-b804-429f-b99e-57e45023a6e8, runID: 4821bd7c-b9f6-4684-92be-ab7354eefc36): canceled
```

With test env:

```
workflow execution error (type: func1, workflowID: default-test-workflow-id, runID: default-test-run-id): child workflow execution error (type: waitForCancelWorkflow, workflowID: default-test-run-id_1, runID: default-test-run-id_1_RunID, initiatedEventID: 0, startedEventID: 0): canceled
```

The test env is consistent with the behavior of other SDKs, but it's likely too late to change the cancelation behavior with the real server for consistency. I propose we change the test env to be consistent with the "real" env.

PR #2181, opened 2026-02-06
💥 Use Temporal Failures for Nexus Error Serialization

Quinn-With-Two-Ns:
Nexus error changes to use `Temporal` failures to respond to task failures and operation failures. This change now allows handler error and operation error to have their own message independent of the cause.

---

> [!NOTE]
> **Medium Risk**
> Changes Nexus task completion/failure payloads and error conversion logic (including message/cause/stack trace handling) with backward-compat paths for older servers, which could affect interoperability and error semantics.
> 
> **Overview**
> **Switches Nexus error serialization to use Temporal `Failure` objects** so operation and handler errors can carry their own message/stack trace independent of the underlying cause.
> 
> Updates Nexus task handling/polling and the failure converter to round-trip `nexus.HandlerError`/`nexus.OperationError` via Temporal failures (including token/id mapping for `NexusOperationError`, preserving original failures, and optional encoding of common attributes), while adding a capability/override path to fall back to the legacy proto error format.
> 
> Adds `CanceledErrorOptions` + `NewCanceledErrorWithOptions` (message, details, cause + `Unwrap`) and expands tests/CI gating (new `DISABLE_NEW_NEXUS_ERROR_FORMAT_TESTS` flag) alongside bumping `github.com/nexus-rpc/sdk-go` to `v0.6.0` across modules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e064e8c5ec3dc50a5457be9bcf155e04549a9558. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/nexus_test.go
+++ b/test/nexus_test.go
@@ -1938,9 +1938,9 @@ func TestAsyncOperationCompletionCustomFailureConverter(t *testing.T) {
 	require.ErrorAs(t, err, &appErr)
 	require.Equal(t, "async failure", appErr.Message())
 	if os.Getenv("DISABLE_NEW_NEXUS_ERROR_FORMAT_TESTS") != "" {
-		require.Equal(t, "", appErr.Type())
-	} else {
 		require.Equal(t, "NexusFailure", appErr.Type())
+	} else {
+		require.Equal(t, "", appErr.Type())
 	}
 	var details nexus.Failure
 	require.NoError(t, appErr.Details(&details))


internal/failure_converter.go
```diff
@@ -294,10 +312,24 @@ func (dfc *DefaultFailureConverter) FailureToError(failure *failurepb.Failure) e
 		case enumspb.NEXUS_HANDLER_ERROR_RETRY_BEHAVIOR_NON_RETRYABLE:
 			retryBehavior = nexus.HandlerErrorRetryBehaviorNonRetryable
 		}
+		// Check if the message is the default message or if it was set by the user. If it's the default message,
+		// we will replace it with an empty string to avoid confusion.
+		cause := dfc.FailureToError(failure.GetCause())
+		typ := nexus.HandlerErrorType(info.Type)
+		if strings.HasPrefix(message, fmt.Sprintf("handler error (%s)", typ)) {
+			message = ""
+		}
```
VegetarianOrc, line 324: Is this still necessary after the most recent server changes? I haven't noticed this string being present in Python, but that could mean I've done something wrong on that side 😅.

internal/internal_nexus_task_handler.go
```diff
@@ -406,32 +456,149 @@ func (h *nexusTaskHandler) newNexusOperationContext(response *workflowservice.Po
...
+		f := res.GetStartOperation().GetFailure()
+		var state string
+		if cf := f.GetCanceledFailureInfo(); cf != nil {
+			state = string(nexus.OperationStateCanceled)
+		} else if af := f.GetApplicationFailureInfo(); af != nil {
+			state = string(nexus.OperationStateFailed)
+		} else {
+			return nil, fmt.Errorf("cannot convert failure to operation error without failure reason support")
+		}
+		failure, err := h.temporalFailureToNexusFailure(convertUnsupportedFailures(f.GetCause()))
```
Quinn-With-Two-Ns: @bergundy Do you think we still need this change since we are no longer adding new failure types?

bergundy: No need.

internal/failure_converter.go
Quinn-With-Two-Ns: @bergundy Thoughts? I thought we agreeded to keep the error message the same for old errors but I think the current nexus go SDK does change it.

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.3.2
-	github.com/nexus-rpc/sdk-go v0.5.1
+	github.com/nexus-rpc/sdk-go v0.5.2-0.20260211051645-26b0b4c584e5
```
cretz: Will defer to @yuandrew for review, but we will want to release Nexus SDK before we merge this

Quinn-With-Two-Ns: Yep agree, just holding off releasing the Nexus SDK in case we come across anything in this PR that makes us want to adjust behaviour

internal/error.go
```diff
@@ -142,11 +142,26 @@ type (
...
 
+	// CanceledErrorOptions should be used to set all the desired attributes of a new CanceledError
+	//
+	// Exposed as: [go.temporal.io/sdk/temporal.CanceledErrorOptions]
+	CanceledErrorOptions struct {
+		// Message is the error message.
+		// Defaults to "canceled" if not set.
+		Message string
+		// Details is a list of arbitrary values that can be used to provide additional context to the error.
+		Details []interface{}
```
bergundy: Please:

```suggestion
		Details []any
```

internal/error_test.go
```diff
@@ -447,6 +447,98 @@ func Test_IsCanceledError(t *testing.T) {
...
+	require.False(t, canceledErr.HasDetails())
+
+	// Test with default message (empty string)
+	err = NewCanceledErrorWithOptions(CanceledErrorOptions{})
+	require.True(t, errors.As(err, &canceledErr))
+	require.Equal(t, "canceled", canceledErr.Error())
+
+	// Test with details
+	err = NewCanceledErrorWithOptions(CanceledErrorOptions{
+		Details: []interface{}{testErrorDetails1, testErrorDetails2},
```
bergundy: More nitpicks:

```suggestion
		Details: []any{testErrorDetails1, testErrorDetails2},
```

internal/error_test.go
```diff
@@ -1020,7 +1112,7 @@ func Test_convertErrorToFailure_NexusHandlerError(t *testing.T) {
 		Cause:         errors.New("custom cause"),
 		RetryBehavior: nexus.HandlerErrorRetryBehaviorNonRetryable,
 	})
-	require.Equal("handler error (INTERNAL): custom cause", f.GetMessage())
+	require.Equal("handler error (INTERNAL)", f.GetMessage())
```
bergundy: Seems worth putting a message on the handler error in this test?

internal/error_test.go
```diff
@@ -1266,3 +1358,155 @@ func Test_convertFailureToError_SaveFailure(t *testing.T) {
...
+	require.NotNil(t, failure1)
+	require.NotNil(t, failure1.GetEncodedAttributes(), "Round 1: encoded attributes should be present")
+	require.Equal(t, "Encoded failure", failure1.GetMessage(), "Round 1: message should be encoded")
+	require.Equal(t, "", failure1.GetStackTrace(), "Round 1: stack trace should be cleared")
+
+	// Round 2: Failure -> HandlerError (with decoding)
+	err2 := middleConverter.FailureToError(failure1)
+	require.NotNil(t, err2)
+	he2, ok := err2.(*nexus.HandlerError)
+	require.True(t, ok, "Round 2: should be HandlerError")
```
bergundy: nit: Use `require.ErrorAs`.

internal/error_test.go
```diff
@@ -1266,3 +1358,155 @@ func Test_convertFailureToError_SaveFailure(t *testing.T) {
...
+	}
+
+	// Round 1: HandlerError -> Failure (with encoding)
+	failure1 := encodingConverter.ErrorToFailure(originalErr)
+	require.NotNil(t, failure1)
+	require.NotNil(t, failure1.GetEncodedAttributes(), "Round 1: encoded attributes should be present")
+	require.Equal(t, "Encoded failure", failure1.GetMessage(), "Round 1: message should be encoded")
+	require.Equal(t, "", failure1.GetStackTrace(), "Round 1: stack trace should be cleared")
+
+	// Round 2: Failure -> HandlerError (with decoding)
```
bergundy: This isn't decoding though?

internal/failure_converter.go
```diff
@@ -163,6 +165,12 @@ func (dfc *DefaultFailureConverter) ErrorToFailure(err error) *failurepb.Failure
 		}
 		failure.FailureInfo = &failurepb.Failure_NexusOperationExecutionFailureInfo{NexusOperationExecutionFailureInfo: failureInfo}
 	case *nexus.HandlerError:
+		if err.OriginalFailure != nil {
+			f, err := nexusFailureToAPIFailure(*err.OriginalFailure, true)
+			if err == nil {
```
bergundy, line 172: Worth a comment explaining why we are ignoring errors here.

temporal/error.go
```diff
@@ -188,12 +188,22 @@ func NewNonRetryableApplicationError(message, errType string, cause error, detai
 	)
 }
 
+// CanceledErrorOptions should be used to set all the desired attributes of a new CanceledError
+// To get a new instance use CanceledErrorAttributes function
```
yuandrew: These sentences seem to be missing punctuation

test/nexus_test.go
```diff
@@ -284,7 +285,7 @@ func TestNexusSyncOperation(t *testing.T) {
 		var opErr *nexus.OperationError
 		require.ErrorAs(t, err, &opErr)
 		require.Equal(t, nexus.OperationStateFailed, opErr.State)
-		require.Equal(t, "fail", opErr.Cause.Error())
+		//require.Equal(t, "fail", opErr.Cause.Error())
```
yuandrew: should this be uncommented or removed?

test/nexus_test.go
```diff
@@ -699,7 +767,9 @@ func TestSyncOperationFromWorkflow(t *testing.T) {
 		err = opErr.Unwrap()
 		var appErr *temporal.ApplicationError
 		require.ErrorAs(t, err, &appErr)
-		require.Equal(t, "failed for test", appErr.Message())
+		if temporalFailureResp {
+			//require.Equal(t, "failed for test", appErr.Message())
```
yuandrew: uncomment or remove?

test/nexus_test.go
```diff
@@ -835,6 +911,239 @@ func TestSyncOperationFromWorkflow(t *testing.T) {
...
+		require.ErrorAs(t, err, &execErr)
+		var canceledErr *temporal.CanceledError
+		err = execErr.Unwrap()
+		require.ErrorAs(t, err, &canceledErr)
+		// The Go SDK doesn't preserve the error chain for canceled errors, so we can't check the message or cause,
+		// but we can check that it's a canceled error.
+		//
+		// See: https://github.com/temporalio/sdk-go/issues/1496
+		//
+		// require.Equal(t, "canceled", canceledErr.Error())
```
yuandrew, line 978: uncomment or remove?

test/nexus_test.go
```diff
@@ -1628,7 +1937,7 @@ func TestAsyncOperationCompletionCustomFailureConverter(t *testing.T) {
 	var appErr *temporal.ApplicationError
 	require.ErrorAs(t, err, &appErr)
 	require.Equal(t, "async failure", appErr.Message())
-	require.Equal(t, "NexusFailure", appErr.Type())
+	// require.Equal(t, "NexusFailure", appErr.Type())
```
yuandrew: uncomment or remove?

internal/nexus_operations.go
```diff
@@ -274,6 +274,28 @@ func nexusOperationFailure(params executeNexusOperationParams, token string, cau
...
+	b, err := protojson.Marshal(failure)
+	if err != nil {
+		return nil, err
+	}
+	failure.Message = message
+	failure.StackTrace = stackTrack
+	return &nexus.Failure{
+		Message:  message,
+		Metadata: nexusFailureMetadata,
+		Details:  b,
```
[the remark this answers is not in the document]
yuandrew: codex seems right here, missing a `StackTrace: stacktrace,` line?

internal/failure_converter.go
```diff
@@ -175,6 +183,10 @@ func (dfc *DefaultFailureConverter) ErrorToFailure(err error) *failurepb.Failure
 			RetryBehavior: retryBehavior,
 		}
 		failure.FailureInfo = &failurepb.Failure_NexusHandlerFailureInfo{NexusHandlerFailureInfo: failureInfo}
+		failure.StackTrace = err.StackTrace
+		if len(err.Message) > 0 {
```
bergundy, line 190: Why?

internal/failure_converter.go
bergundy: I think we need this to avoid repetition for when this error is generated by an old SDK?

internal/failure_converter.go
```diff
@@ -163,6 +165,12 @@ func (dfc *DefaultFailureConverter) ErrorToFailure(err error) *failurepb.Failure
 		}
 		failure.FailureInfo = &failurepb.Failure_NexusOperationExecutionFailureInfo{NexusOperationExecutionFailureInfo: failureInfo}
 	case *nexus.HandlerError:
+		if err.OriginalFailure != nil {
+			f, err := nexusFailureToAPIFailure(*err.OriginalFailure, true)
```
bergundy: Can we rename this to `nexusFailureToTemporalFailure` for symmetry with the inverse function?

internal/internal_nexus_task_handler.go
```diff
@@ -31,13 +32,25 @@ import (
...
 
-func nexusHandlerError(t nexus.HandlerErrorType, message string) *nexuspb.HandlerError {
-	return &nexuspb.HandlerError{
-		ErrorType: string(t),
-		Failure: &nexuspb.Failure{
-			Message: message,
-		},
+// debugDisableTemporalFailureResponses is an internal debug switch to force the SDK
+// to treat GetTemporalFailureResponses() as always returning false.
+// This is only intended for testing purposes.
```
yuandrew, line 37: This var itself isn't just for testing purposes, right? i see `SetDebugDisableTemporalFailureResponses` only called from tests, but `getEffectiveTemporalFailureResponses` uses this var itself depending on server support, that seems to always apply, not just for tests?

Quinn-With-Two-Ns: Yes, because the server may not support the new format so we need to know if we should fall back to the old response format

test/nexus_test.go
Quinn-With-Two-Ns: This can't be uncommented until https://github.com/temporalio/sdk-go/issues/1496 is resolved, would you rather I remove this for now?

internal/nexus_operations.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah I originally left it off since Go errors don't normally have a stack trace, but I guess for completion I might as well.

test/nexus_test.go
yuandrew: right, that makes sense. More of a nit, could we say something 
`// TODO: uncomment once issue above is resolved` so it's more obviously clear

Quinn-With-Two-Ns: Works for me, i'll update the comment

internal/internal_nexus_task_handler.go
```diff
@@ -208,32 +246,50 @@ func (h *nexusTaskHandler) handleStartOperation(
 		var unsuccessfulOperationErr *nexus.OperationError
 		err = convertKnownErrors(err)
 		if errors.As(err, &unsuccessfulOperationErr) {
-			failure, err := h.errorToFailure(unsuccessfulOperationErr.Cause)
-			if err != nil {
-				return nil, nil, err
+			// Convert OperationError to a Temporal failure
```
bergundy, line 249: Shouldn't this be part of the Temporal failure converter? Seems like we should consistently do this conversion regardless of where operation errors appear.

internal/internal_nexus_task_handler.go
```diff
@@ -208,32 +246,50 @@ func (h *nexusTaskHandler) handleStartOperation(
...
+					"OperationError",
+					ApplicationErrorOptions{
+						NonRetryable: true,
+						Cause:        unsuccessfulOperationErr.Cause,
+					},
+				)
+			default:
+				return nil, nexus.NewHandlerErrorf(
+					nexus.HandlerErrorTypeInternal,
+					"invalid operation state in OperationError: %s",
```
bergundy: Maybe give more details here so it's clearer to our users that this is something they caused?

internal/internal_nexus_task_handler.go
```diff
@@ -334,12 +392,13 @@ func (h *nexusTaskHandler) handleCancelOperation(ctx context.Context, nctx *Nexu
...
 		var handlerErr *nexus.HandlerError
 		if errors.As(err, &handlerErr) {
-			protoErr, err := h.nexusHandlerErrorToProto(handlerErr)
-			return nil, protoErr, err
+			return nil, handlerErr, nil
 		}
 		// Default to internal error.
-		protoErr, err := h.internalError(err)
-		return nil, protoErr, err
+		return nil, &nexus.HandlerError{
```
bergundy, line 398: I thought we said we would put a message on handler errors like we did in .NET?

temporal/error.go
```diff
@@ -188,12 +188,22 @@ func NewNonRetryableApplicationError(message, errType string, cause error, detai
 	)
 }
 
+// CanceledErrorOptions should be used to set all the desired attributes of a new CanceledError
+// To get a new instance use CanceledErrorAttributes function.
+type CanceledErrorOptions = internal.CanceledErrorOptions
```
bergundy, line 193: I hate this type aliasing but it's consistent with the rest of the SDK.

bergundy (commented):
Would you run this suite again with all of the recent changes to verify everything works as expected still? https://github.com/bergundy/nexus-error-compat-tests

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: I don't agree this is a special case for Nexus operation , we don't want this conversion in other cases

temporal/error.go
Quinn-With-Two-Ns: Yeah we aren't going to remove type aliasing in this PR

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: I think dotnet just uses the message "Internal handler error", you want to use that here?

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: Any suggestions? The error message already states the problem is in the `OperationError` that the user returned? The only thing I can think to make is clearer is maybe add "invalid operation state in OperationError: %s, only Failed or Canceled is allowed"?

internal/failure_converter.go
Quinn-With-Two-Ns: If the error doesn't have a message we don't want to wipe the default message we set at the top of the converter. I'll add a comment to explain why.

internal/internal_nexus_task_handler.go
```diff
@@ -406,22 +457,59 @@ func (h *nexusTaskHandler) newNexusOperationContext(response *workflowservice.Po
...
+			StartOperation: &nexuspb.StartOperationResponse{
+				Variant: &nexuspb.StartOperationResponse_OperationError{
+					OperationError: &nexuspb.UnsuccessfulOperationError{
+						OperationState: state,
+						Failure:        failure,
+					},
+				},
+			},
+		}
+	}
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: This is intentional, the legacy format does loose information.

internal/internal_nexus_task_handler.go
bergundy: IMHO it'd be best to be consistent here so we can document this behavior without asterisks. This is also the server behavior. It will convert these failures no matter how deep in the cause chain. There's a subtle difference, the server only does this for communication over nexus since operation errors aren't representable as a Temporal failure.

https://github.com/temporalio/temporal/blob/ee8b8411102e55c34ac024da0a6a916d9b27d1b7/common/nexus/failure.go#L197

internal/internal_nexus_task_handler.go
bergundy: `handler returned an invalid operation error type: %s`?

internal/internal_nexus_task_handler.go
bergundy: Yeah, let's put error messages on all of the handler errors the SDK produces. I think this is the only place that's missing but I haven't scanned the rest of the code to tell for sure.

bergundy (approved):
Approved with comments.

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: This is a non standard way to convert an error to a Temporal failure in the SDK so I would strongly think we need to keep this outside the failure converter. Failure converter is also user controlled so doing it at a higher level is safer and more consistent.

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: Sure, done!

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: Went through and added message to a couple other obscure handler errors as well

internal/internal_nexus_task_handler.go
Quinn-With-Two-Ns: We do a lot of special conversion for errors from the handler already this is aligned with that approach

bergundy: IMHO a user might be surprised if we document that operation error is transformed to application failure or canceled failure and not have this work for a nested error.

PR #2186, opened 2026-02-10
💥Add worker heartbeat support

### LINKED ISSUE #2094: Worker Heartbeating

yuandrew:
NOTE: This is a re-creation of https://github.com/temporalio/sdk-go/pull/2148, making a newer, clean PR as the resolved comment count was getting very high and making it hard to navigate. 

## What was changed

Key changes:
- Add HeartbeatMetricsHandler to capture metrics needed for heartbeats
- Add internal_worker_heartbeat.go for heartbeat worker management
- 💥 Removed `contrib/resourcetuner` lib, resource tuner portion has been moved into the SDK and is exposed through the `worker` package, while a new `contrib/sysinfo` package was created to handle system resource measurement
  - We never released a v1.0.0 of `contrib/resourcetuner`, so this should be okay.
- Plumb heartbeat data through workers (workflow, activity, nexus, local activity)
- always send `ShutdownWorker` RPC to give final heartbeat, even if not using sticky queues
- Move `verifyNamespaceExist` calls from each specific worker, and instead call at the `AggregatedWorker` level and cache for worker heartbeating

## Why?

New feature!

## Checklist

1. Closes #2094

2. How was this tested:

Added a bunch of tests to `worker_heartbeat_test.go`
- Add integration tests for worker heartbeat functionality

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Touches core worker start/stop and polling paths and adds new background goroutines/RPCs for heartbeating; failures could impact worker shutdown behavior or add load if misconfigured, but the feature is gated by an experimental interval option and server capabilities.
> 
> **Overview**
> Adds **experimental worker heartbeats**: clients can now be configured with `WorkerHeartbeatInterval` to periodically call `RecordWorkerHeartbeat` and to send a final heartbeat via `ShutdownWorker` on stop, including worker/poller/slot metrics, plugin names, deployment version, and optional CPU/memory usage.
> 
> Refactors tuning/sysinfo: moves the resource-based tuner into the SDK `worker` API (requiring an injected `SysInfoProvider`), introduces new `contrib/sysinfo` (gopsutil + cgroup-aware) implementation, and wires poll-success tracking plus heartbeat-specific metrics capturing across workflow/activity/nexus pollers.
> 
> Updates dev/test configs to enable `WorkerHeartbeatsEnabled`/`ListWorkersEnabled`, adjusts integration tests accordingly, and adds a large new `worker_heartbeat_test.go` coverage suite.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1cb7c909c67792b545a9dab335417299405c3082. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/docker/dynamic-config-custom.yaml
+++ b/.github/workflows/docker/dynamic-config-custom.yaml
@@ -50,5 +50,7 @@ history.enableTransitionHistory:
   - value: true
 component.nexusoperations.useSystemCallbackURL:
   - value: false
+frontend.WorkerHeartbeatsEnabled:
+  - value: true
 frontend.ListWorkersEnabled:
   - value: true


internal/internal_worker_heartbeat.go
```diff
@@ -0,0 +1,200 @@
...
+		return fmt.Errorf("failed to get namespace capabilities: %w", err)
+	}
+	if !capabilities.GetWorkerHeartbeats() {
+		if m.logger != nil {
+			m.logger.Debug("Worker heartbeating configured, but server version does not support it.")
+		}
+		return nil
+	}
+
+	namespace := worker.executionParams.Namespace
```
yuandrew, line 56: https://github.com/temporalio/sdk-go/pull/2148#discussion_r2742009640

Chad previously suggested moving this into a `getOrCreateSharedNamespaceWorker`, but I just moved things back because we want to keep `workersMutex` locked until after we update the `hw.callbacks` so avoid a race where worker A removes the "empty" worker before worker B is able to register its callback

internal/client.go
```diff
@@ -601,6 +602,14 @@ type (
...
 		// NOTE: Experimental
 		Plugins []ClientPlugin
+
+		// WorkerHeartbeatInterval is the interval at which the worker will send heartbeats to the server.
+		// Interval must be between 1s and 60s, inclusive.
+		//
+		// default: 60s. To disable, set to 0.
+		//
+		// NOTE: Experimental
+		WorkerHeartbeatInterval *time.Duration
```
yuandrew: https://github.com/temporalio/sdk-go/pull/2148#discussion_r2764590514

cretz
Pedantic, but would maybe recommend "negative to disable" and remove the pointer and assume 0/unset means use default. Not a big deal though.

yuandrew
I remember consulting claude on this design decision, it seemed like this was the more idiomatic choice for Go,

        HashiCorp consul-template — the most extensive example. Nearly every duration config is
        *time.Duration with a Finalize() method that fills defaults for nil. They even have a
        TimeDurationPresent() helper that returns true only if non-nil AND non-zero, explicitly treating zero
        as a distinct state from nil.

https://pkg.go.dev/github.com/hashicorp/consul-template/config

internal/client.go
```diff
@@ -1198,7 +1207,9 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
```
yuandrew, line 1207: https://github.com/temporalio/sdk-go/pull/2148#discussion_r2764603338

cretz
Can you help me understand this change? All user-facing code that calls NewServiceClient should already be setting this default. Is there a new code path that calls NewServiceClient?

yuandrew
No new code path, but sessionEnvironmentImpl.SignalCreationResponse calls GetClient without a logger, this seems like it would fix that missing scenario, but also seems like good practice to have in general?

cretz
    this seems like it would fix that missing scenario, but also seems like good practice to have in general?

They may have left logger off intentionally, unsure, but would take some more digging. It's possible/likely you're right, but it seems unrelated to this project and may deserve a separate issue.

yuandrew
turns out this was due to newHeartbeatManager using the logger, but I added the check into that function itself, so SignalCreationResponse is unaffected

internal/resource_tuner.go
```diff
@@ -21,13 +16,47 @@ const (
...
+	MemoryUsage(infoContext *SysInfoContext) (float64, error)
+	// CpuUsage returns the current system CPU usage as a fraction of total CPU usage between 0
+	// and 1.
+	CpuUsage(infoContext *SysInfoContext) (float64, error)
+}
+
+// SysInfoContext provides context for SysInfoProvider calls.
+//
+// Exposed as: [go.temporal.io/sdk/worker.SysInfoContext]
+type SysInfoContext struct {
```
yuandrew, line 34: https://github.com/temporalio/sdk-go/pull/2148#discussion_r2764801380

cretz
Consider embedding context.Context in here even if just context.Background() (but not that important)

yuandrew
keeping out for now, localActivityContext, NexusOperationContext and testContext are all *Context structs that are missing context.Context, if we want to add it in later to use, we can

cretz
Then IMO we should consider passing a Go context into the calls alongside where this is passed (even if it's not used for anything yet)

yuandrew
I'm planning on leaving it out for now, and adding in the context into the SysInfoContext struct when we need the context. It seems a little redundant to pass in a Context and SysInfoContext together, any reason we'd need to add it into the calls alongside right now?

internal/internal_worker.go
```diff
@@ -1286,6 +1278,10 @@ func (aw *AggregatedWorker) start() error {
 	}
 	proto.Merge(aw.capabilities, capabilities)
 
+	if _, err := aw.client.loadNamespaceCapabilities(aw.executionParams.MetricsHandler); err != nil {
```
yuandrew, line 1284: @jmaeagle99 this is similar to your change in external storage, we can coordinate

internal/client.go
cretz: Claude is wrong IMO, pointers to scalars like duration (which is just an integer basically) are rare, and Hashicorp's code is a poor example here IMO. Having `nil` and `0` be different things is not very Go like IMO.

internal/internal_worker.go
```diff
@@ -1470,9 +1474,64 @@ func (aw *AggregatedWorker) Stop() {
 		WorkerInstanceKey: aw.workerInstanceKey,
 	})
 
+	aw.unregisterHeartbeatWorker()
```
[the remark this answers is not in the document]
cretz: Should be heeded. Can you issue a heartbeat after shutdown worker?

internal/internal_worker.go
```diff
@@ -264,6 +272,9 @@ func ensureRequiredParams(params *workerExecutionParameters) {
 				NumNexusSlots:         defaultMaxConcurrentTaskExecutionSize,
 			})
 	}
+	if params.pollTimeTracker == nil {
```
cretz, line 277: What is the situation where this is not always nil? Should this be set where params is created and not here?

internal/internal_worker.go
```diff
@@ -1470,9 +1474,64 @@ func (aw *AggregatedWorker) Stop() {
...
+		heartbeat = aw.heartbeatCallback()
+		heartbeat.Status = enumspb.WORKER_STATUS_SHUTTING_DOWN
+	}
+
+	var stickyTaskQueue string
+	if aw.workflowWorker != nil && aw.workflowWorker.stickyUUID != "" {
+		stickyTaskQueue = getWorkerTaskQueue(aw.workflowWorker.stickyUUID)
+	}
+
+	_, err := aw.client.workflowService.ShutdownWorker(grpcCtx, &workflowservice.ShutdownWorkerRequest{
```
cretz, line 1521: Can we pass the worker instance key here and in polls?Or if these should be done in separate issue, can we make it and put it up for triage?

internal/internal_worker.go
```diff
@@ -1470,9 +1474,64 @@ func (aw *AggregatedWorker) Stop() {
...
+		WorkerHeartbeat: heartbeat,
+	})
+
+	// Ignore unimplemented (server doesn't support it)
+	if _, isUnimplemented := err.(*serviceerror.Unimplemented); isUnimplemented {
+		return
+	}
+
+	if err != nil {
+		aw.logger.Debug("ShutdownWorker rpc errored during worker shutdown.", tagError, err)
```
cretz, line 1536: I know this was debug in old code, but we should consider making this a warning, but can be done in a separate issue. This failing has negative user effects, so effectively swallowing this error and not telling user can be rough. But we have to think about situations where this is just noise.

internal/resource_tuner.go
cretz: Go API that is invoked that others implement basically all need a `context.Context` always even if you don't do anything with it yet. At this point it might as well be part of Go API heh. You can wait and embed later if you want, though harmless to do now.

internal/internal_worker.go
yuandrew: will address separately

internal/internal_worker.go
yuandrew: yes, added

internal/internal_worker.go
yuandrew: It's always nil, but this is the most common place to nil check, mainly for unit tests sake. I had this set where params are created in `NewAggregatedWorker`, but I'd still like to keep this check here for tests

internal/internal_worker.go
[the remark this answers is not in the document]
yuandrew: mirrored core by making this a shared atomic, and subsequent heartbeats will also be `SHUTTING_DOWN` state

internal/client.go
yuandrew: changed to not be a pointer

test/worker_heartbeat_test.go
```diff
@@ -0,0 +1,996 @@
...
+		ts.T().Logf("  Last Successful Poll Time: %v", workerInfo.NexusPollerInfo.LastSuccessfulPollTime.AsTime())
+		ts.T().Logf("  Is Autoscaling: %v", workerInfo.NexusPollerInfo.IsAutoscaling)
+	}
+	if len(workerInfo.Plugins) > 0 {
+		ts.T().Logf("=== Plugins ===")
+		for _, plugin := range workerInfo.Plugins {
+			ts.T().Logf("  Name: %s", plugin.Name)
+		}
+	}
+}
```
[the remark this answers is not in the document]
Sushisource: Is true

internal/internal_worker.go
Sushisource: It looks like it's still not on shutdown here, but is on polls?

internal/internal_worker_heartbeat_metrics.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func (h *heartbeatMetricsHandler) get(key string) int64 {
+	if v, ok := h.metrics.Load(key); ok {
+		return v.(*atomic.Int64).Load()
+	}
+	return 0
+}
+
+// populateHeartbeatOptions contains external dependencies needed to populate heartbeat metrics.
```
Sushisource: "external dependencies" is somewhat odd phrasing

internal/internal_worker_heartbeat_metrics.go
```diff
@@ -0,0 +1,323 @@
...
+
+	return &workerpb.WorkerSlotsInfo{
+		CurrentAvailableSlots:      slotsAvailable,
+		CurrentUsedSlots:           slotsUsed,
+		SlotSupplierKind:           supplierKind,
+		TotalProcessedTasks:        int32(totalProcessed),
+		TotalFailedTasks:           int32(totalFailed),
+		LastIntervalProcessedTasks: int32(intervalProcessed),
+		LastIntervalFailureTasks:   int32(intervalFailed),
+	}
```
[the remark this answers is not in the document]
Sushisource: Alternatively these should just be stored as int32 since that's what we have to send them as eventually anyway

Sushisource (approved):
Overall makes sense to me. I didn't do a super detailed review but the high level and tests look reasonable. A few small things.

PR #2198, opened 2026-02-19
Fix standalone activity result long-poll bug

dandavison:
## What was changed
Use per-RPC context deadlines in Standalone Activity result long-poll loop instead of one deadline for all RPCs. 

## Why?
The long-poll loop was failing to wait for longer than 10s. This meant that `handle.Get()` could not be used for long-running activities.

## Checklist
2. How was this tested:
New integration test

---

> [!NOTE]
> **Medium Risk**
> Changes the standalone activity result long-poll loop to create/cancel a fresh gRPC context each iteration with a fixed 60s timeout, which can affect polling behavior and resource usage but is localized to activity result retrieval.
> 
> **Overview**
> Fixes standalone activity result long-polling so `PollActivityResult` creates a *new* gRPC context per poll iteration (and cancels it immediately after the RPC), instead of reusing a single context across the whole loop.
> 
> Introduces a `pollActivityTimeout` (60s) applied via `grpcTimeout(...)` to allow long-running standalone activities to be awaited reliably, and adds a regression test asserting each poll call receives a fresh context with a deadline.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9248ddc99ac0166fdcf84588332179a004d5547c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/internal_workflow_client_test.go
+++ b/internal/internal_workflow_client_test.go
@@ -2653,7 +2653,7 @@ func TestUpdate(t *testing.T) {
 }
 
 func (s *workflowClientTestSuite) TestPollActivityResultUsesPerIterationContext() {
-	var prevDeadline time.Time
+	var prevCtx context.Context
 	callCount := 0
 	s.service.EXPECT().
 		PollActivityExecution(gomock.Any(), gomock.Any(), gomock.Any()).
@@ -2663,13 +2663,12 @@ func (s *workflowClientTestSuite) TestPollActivityResultUsesPerIterationContext(
 			_ ...grpc.CallOption,
 		) (*workflowservice.PollActivityExecutionResponse, error) {
 			callCount++
-			deadline, ok := ctx.Deadline()
+			_, ok := ctx.Deadline()
 			s.True(ok)
 			if callCount > 1 {
-				s.Greater(deadline, prevDeadline,
-					"each call should get a fresh context, not share one")
+				s.True(prevCtx != ctx, "each call should get a fresh context, not share one")
 			}
-			prevDeadline = deadline
+			prevCtx = ctx
 
 			if callCount < 3 {
 				return &workflowservice.PollActivityExecutionResponse{}, nil


test/integration_test.go
```diff
@@ -8915,4 +8914,23 @@ func (ts *IntegrationTestSuite) TestExecuteActivitySuite() {
...
+		handle, err := ts.client.ExecuteActivity(ctx, options, "readFromChannelActivity", 15*time.Second)
+		ts.NoError(err)
+
+		go func() { activityResultChan <- "done" }()
+
+		var result string
+		err = handle.Get(ctx, &result)
+		ts.NoError(err, "handle.Get() should survive beyond the 10s internal gRPC timeout")
+		ts.Equal("done", result)
+	})
```
dandavison: Not sure if you want to merge such a lengthy test, but it repros the bug; the relevant constants didn't seem amenable to being manipulated in the test.

yuandrew: yeah, i don't think we want such a long test. We could have a unit test that instead inspects the context deadline it receives on each call, 

Maybe something like below (I haven't run this code, largely AI generated). But here we check that each RPC has a different deadline, and if you want would you also assert that the deadline is > 50s or something, so it's closer to the 60s you're setting now

```
  func (s *workflowClientTestSuite) TestPollActivityResultUsesPerIterationContext() {
      var prevDeadline time.Time
      callCount := 0
      s.service.EXPECT().
          PollActivityExecution(gomock.Any(), gomock.Any(), gomock.Any()).
          DoAndReturn(func(
              ctx context.Context,
              _ *workflowservice.PollActivityExecutionRequest,
              _ ...grpc.CallOption,
          ) (*workflowservice.PollActivityExecutionResponse, error) {
              callCount++
              deadline, ok := ctx.Deadline()
              s.True(ok)
              if callCount > 1 {
                  s.Greater(deadline, prevDeadline,
                      "each call should get a fresh context, not share one")
              }
              prevDeadline = deadline

              if callCount < 3 {
                  return &workflowservice.PollActivityExecutionResponse{}, nil
              }
              return &workflowservice.PollActivityExecutionResponse{
                  Outcome: &activitypb.ActivityExecutionOutcome{
                      Value: &activitypb.ActivityExecutionOutcome_Result{
                          Result: &commonpb.Payloads{},
                      },
                  },
              }, nil
          }).Times(3)

      out, err := s.client.interceptor.PollActivityResult(
          context.Background(),
          &ClientPollActivityResultInput{ActivityID: "test-id", RunID: "run-id"},
      )
      s.NoError(err)
      s.NotNil(out.Result)
  }
```

dandavison: Thanks! Replaced with a test like that.

internal/internal_workflow_client_test.go
```diff
@@ -2650,3 +2651,43 @@ func TestUpdate(t *testing.T) {
...
+			deadline, ok := ctx.Deadline()
+			s.True(ok)
+			if callCount > 1 {
+				s.Greater(deadline, prevDeadline,
+					"each call should get a fresh context, not share one")
+			}
+			prevDeadline = deadline
+
+			if callCount < 3 {
+				return &workflowservice.PollActivityExecutionResponse{}, nil
```
yuandrew, line 2674: Windows is failing, Claude says 
> time.Now() has ~15.6ms resolution (the system timer interrupt interval)

maybe add a `time.Sleep(100 * time.Milliseconds)` before returning an empty response, to give the next attempt time for the clock to tick

dandavison: Thanks. I've changed it to assert context pointer inequality rather than increasing deadlines.

PR #2199, opened 2026-02-21
No longer cancel pending polls on shutdown

yuandrew:
## What was changed

Plumbed `worker_instance_key` to all poll calls, and `TaskQueueType` and `TaskQueue` to the `ShutdownWorker` RPC call

If `worker_poll_complete_on_shutdown` namespace capability exists, then don't cancel outstanding poll requests on shutdown. Instead expect server to return an empty result.

Also added a new `noRepoll` to baseWorker that prevents pollers from polling after worker shutdown is initiated. This is needed because there is now a small window where server returns an empty response and `stopCh` is not yet closed. In that window, the poll loop polls again, when we don't want it to.

## Why?

During worker shutdown, the SDK cancels outstanding poll requests client-side by racing them against the shutdown token. This
creates a race condition: the server may dispatch a task to a poll at the same instant the SDK kills the connection, orphaning the
task until start_to_close timeout.

## Checklist

1. Closes 

2. How was this tested:

added test

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes worker shutdown semantics and poller concurrency behavior, which can affect task delivery and shutdown timing across workflow/activity/nexus pollers; mitigated by new unit/integration coverage and capability gating.
> 
> **Overview**
> Updates worker shutdown to be *server-driven* when the namespace advertises `worker_poll_complete_on_shutdown`: pollers no longer cancel in-flight long polls on stop, and instead wait for the server to return empty responses. This is plumbed through all poll RPCs via `WorkerInstanceKey`, and `ShutdownWorker` now includes `TaskQueue` plus active `TaskQueueTypes`.
> 
> Adds a `noRepoll` guard to prevent poll loops from re-polling during the shutdown window, and includes new unit and integration tests to validate graceful shutdown behavior and updated expectations against newer server semantics. Also bumps the embedded dev server version/config in the build tool and refreshes `CONTRIBUTING.md` test-running guidance.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 557f207071c23ebd333d4494c17ec021f51cde76. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -220,6 +220,13 @@ type (
 		workerInstanceKey string
 
 		workerPollCompleteOnShutdown *atomic.Bool
+
+		// Set to true during start() when the namespace has the poller_autoscaling capability.
+		serverSupportsAutoscaling *atomic.Bool
+
+		inboundPayloadVisitor PayloadVisitor
+
+		outboundPayloadVisitor PayloadVisitor
 	}
 
 	// HistoryJSONOptions are options for HistoryFromJSON.
@@ -330,14 +337,14 @@ func newWorkflowTaskWorkerInternal(
 	switch params.WorkflowTaskPollerBehavior.(type) {
 	case *pollerBehaviorSimpleMaximum:
 		scalableTaskPollers = []scalableTaskPoller{
-			newScalableTaskPoller(taskProcessor.createPoller(Mixed), params.Logger, params.WorkflowTaskPollerBehavior),
+			newScalableTaskPoller(taskProcessor.createPoller(Mixed), params.Logger, params.WorkflowTaskPollerBehavior, params.serverSupportsAutoscaling),
 		}
 	case *pollerBehaviorAutoscaling:
 		scalableTaskPollers = []scalableTaskPoller{
-			newScalableTaskPoller(taskProcessor.createPoller(NonSticky), params.Logger, params.WorkflowTaskPollerBehavior),
+			newScalableTaskPoller(taskProcessor.createPoller(NonSticky), params.Logger, params.WorkflowTaskPollerBehavior, params.serverSupportsAutoscaling),
 		}
 		if taskProcessor.stickyCacheSize > 0 {
-			scalableTaskPollers = append(scalableTaskPollers, newScalableTaskPoller(taskProcessor.createPoller(Sticky), params.Logger, params.WorkflowTaskPollerBehavior))
+			scalableTaskPollers = append(scalableTaskPollers, newScalableTaskPoller(taskProcessor.createPoller(Sticky), params.Logger, params.WorkflowTaskPollerBehavior, params.serverSupportsAutoscaling))
 		}
 	}
 
@@ -387,7 +394,7 @@ func newWorkflowTaskWorkerInternal(
 				PollerBehaviorSimpleMaximumOptions{
 					MaximumNumberOfPollers: 2,
 				},
-			)),
+			), params.serverSupportsAutoscaling),
 		},
 		taskProcessor:  localActivityTaskPoller,
 		workerType:     "LocalActivityWorker",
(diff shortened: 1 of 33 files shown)


internal/internal_worker.go
```diff
@@ -433,6 +435,11 @@ func (ww *workflowWorker) Stop() {
 }
 
 func newSessionWorker(client *WorkflowClient, params workerExecutionParameters, env *registry, maxConcurrentSessionExecutionSize int) *sessionWorker {
+	// Session workers poll on resource-specific task queues not included in
```
Sushisource, line 445: Ugh, lovely :( 

We should probably raise this with the server folks. I don't want to block our PRs on it or anything, but, seemingly we could include any task queues the worker is polling on, not just the sticky/nonsticky like we do now.

internal/internal_worker.go
```diff
@@ -1446,6 +1458,19 @@ func (aw *AggregatedWorker) Stop() {
...
+	if !util.IsInterfaceNil(aw.activityWorker) {
+		aw.activityWorker.worker.noRepoll.Store(true)
+	}
+	if !util.IsInterfaceNil(aw.workflowWorker) {
+		aw.workflowWorker.worker.noRepoll.Store(true)
+	}
+	if !util.IsInterfaceNil(aw.nexusWorker) {
+		aw.nexusWorker.worker.noRepoll.Store(true)
+	}
+
```
Sushisource, line 1484: Why do this here instead of in the `Stop` impls? Is it because they happen to late? If not, can we put it in there?

internal/internal_worker_base.go
```diff
@@ -432,6 +434,9 @@ func (bw *baseWorker) runPoller(taskWorker scalableTaskPoller) {
 
 	for {
 		if func() bool {
```
Sushisource, line 438: I know you didn't write this but, wow the whole `if func() bool` thing is wildly confusing

internal/internal_worker_base.go
```diff
@@ -432,6 +434,9 @@ func (bw *baseWorker) runPoller(taskWorker scalableTaskPoller) {
...
+			if bw.noRepoll.Load() {
+				return true
+			}
```
Sushisource, line 441: It feels like we should be able to re-use the stop channel for this?

test/integration_test.go
```diff
@@ -5004,32 +5004,20 @@ func (ts *IntegrationTestSuite) testNonDeterminismFailureCause(historyMismatch b
 	ts.NoError(nextWorker.Start())
 	defer nextWorker.Stop()
 
+	// Give the new worker time to start polling before sending the signal.
+	time.Sleep(100 * time.Millisecond)
```
Sushisource: Why does this matter?  Should be fine to send the signal even if no one is polling yet

yuandrew: I think this as leftover from before I found the actual fix, removing

internal/internal_worker_base.go
yuandrew: Between `shutdownWorker()` returning and `baseWorker.Stop()` closing `stopCh`, a poller that received an empty response from the server can loop back and issue a new poll that no one will ever cancel. So we need an extra knob here to prevent polling before the stop channel is closed.

internal/internal_worker.go
yuandrew: Yeah, there's a small window before the `Stop` impls run, so we need this to be set right before we call `aw.shutdownWorker()` to completely avoid this race

internal/internal_worker.go
yuandrew: Will let Kannan know

internal/internal_worker_base.go
Sushisource: Right, but is it possible we can just move where we check the stop channel?

yuandrew: Not following, the stop channel isn't closed yet, so we can't rely on the stop channel to prevent polling for this race. This has to be a separate mechanism, https://github.com/temporalio/sdk-go/pull/2199#discussion_r2963167743

Sushisource: K, no worries then, just wanted to see if we could re use it for both purposes (IE: close it earlier, and move the check earlier)

internal/internal_worker.go
rkannan82: Yes. What we need is:
```
repeated TaskQueuePolling task_queues

message TaskQueuePolling {
    string name = 1;
    TaskQueueType type = 2;
}
```

PR #2200, opened 2026-02-22
Fix 2090 child workflow goroutine leak

### LINKED ISSUE #2090: Starting a child workflow in tests leaks goroutines
## Expected Behavior

If a child workflow is started and terminates during the test no go routine should be left once the test completes.

## Actual Behavior

A go routine is left running from `executeChildWorkflowWithDelay` which doesn't seem to terminate.

  1. Create a test that runs a Workflow that starts a ChildWorkflow.
  2. Use `defer goleak.VerifyNone(s.T())`

I've added a [repo](https://github.com/mbark/temporal-childwf-reproduction) to illustrate how to reproduce the issue, see the file `do_test.go`:

```go
type TestSuite struct {
	suite.Suite
	testsuite.WorkflowTestSuite
}

func Test(t *testing.T) {
	suite.Run(t, new(TestSuite))
}

func ChildWorkflow(ctx workflow.Context) error {
	return workflow.Sleep(ctx, 24*time.Hour)
}

func ParentWorkflow(ctx workflow.Context) error {
	return workflow.ExecuteChildWorkflow(ctx, "ChildWorkflow", nil).Get(ctx, nil)
}

func (s *TestSuite) TestChildWorkflow() {
	defer goleak.VerifyNone(s.T())
	env.RegisterWorkflow(ChildWorkflow)
	env.RegisterWorkflow(ParentWorkflow)

	env.ExecuteWorkflow(ParentWorkflow)
	err := env.GetWorkflowError()
	s.Require().NoError(err)

	env.AssertExpectations(s.T())
}
```

## Specifications

  - Version: `v1.37.0`
  - Platform: -

kritibehl:
## What was changed
Fix a goroutine leak in the test workflow environment when executing child workflows by registered type name.

Made doneChannel close idempotent using sync.Once.

Ensured doneChannel is closed from both:

startMainLoop() (deferred in the root environment)

Complete() (explicit close on workflow completion)

Consolidated leak regression coverage into a single test under test/ (using test/go.mod, which already includes go.uber.org/goleak).

Removed the previous internal/* leak tests and removed goleak from the root go.mod/go.sum.

## Why?
When executing a child workflow by registered type name, a child testWorkflowEnvironmentImpl could remain blocked in startMainLoop() waiting on doneChannel.

In some completion paths, doneChannel was not guaranteed to be closed for the child environment, causing leaked goroutines.

This change ensures:

doneChannel is always closed exactly once

No race or double-close panic

Child workflow test environments exit cleanly

Regression coverage verifies no goroutines remain after workflow completion
## Checklist

1. Closes #2090
2. How was this tested:
go test ./... -count=1
(cd test && go test -run TestChildWorkflowByName_NoGoroutineLeak -count=1 -v)

Verified no startMainLoop goroutine remains after workflow completion.

3. Any docs updates needed?
No.


yuandrew:
Thanks for the contribution! The fix makes sense to me. But I don't think we want to add goleak as a dependency to our root `go.mod`. Can we instead add a test to `test/`? `test/go.mod` already has the goleak dependency there. And do we need 2 separate tests? Our CI doesn't run tests with `-tags goleak` anyways, seems like we can consolidate into a single test here

PR #2224, opened 2026-03-10
Experimental: External Storage

### LINKED ISSUE #2208: SDK External Payload Storage
Parent issue: https://github.com/temporalio/features/issues/765

jmaeagle99:
## What was changed

Add external storage APIs to allow customers to specify how large payloads or payloads that match some specific customer-defined quality can be stored via customer-supplied external storage systems.
- Customers can choose their external storage endpoints by providing storage drivers. They can either use prebuilt ones or write their own via the `StorageDriver` type. This change does not contain any prebuilt drivers, just provides the primitives to make them.
- Customers can register several storage drivers. This allows workflows and clients to be able to retrieve payloads that have been stored by different driver types.
- Customers can dynamically choose which driver to use for storing each payload by specifying a driver selector. The selector can be a simple callable or could be a `StorageDriverSelector` to participate in serialization context transforms.
- External storage is applied using payload visiting. This visitation occurs just before proto messages are submitted to the server and just after receiving proto messages from the server for run workflow and activity tasks. The application of external storage within a worker happens outside the context of workflow user code.
- Logs are emitted as TMPRL1104 when workflow task execution is above the 5 sec and 10 sec thresholds. The logs include information about payload upload and download metrics as those are likely to have an effect on workflow task duration.
- Logs are emitted as TMPRL1105 when we detect externally stored payloads in workflow history but external storage is not configured.
- External storage is off by default and must be opted into via specifying `ClientOptions.ExternalStorage`. There are no breaking changes regardless of opted into or not.

## Why?

- Allow customers to mitigate situations where payload size may be too large to store into temporal history.
- Allow customers to externally store payloads based on other payload and contextual qualities such a serialization context and payload metadata.

## Checklist

1. Closes #2208 

2. How was this tested: Existing tests and newly added unit and worker tests.

3. Any docs updates needed? Yes, new docs for external storage should be created.

### DIFF
--- a/contrib/datadog/go.mod
+++ b/contrib/datadog/go.mod
@@ -81,7 +81,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.40.0 // indirect
 	go.opentelemetry.io/otel/sdk v1.40.0 // indirect
 	go.opentelemetry.io/otel/trace v1.40.0 // indirect
-	go.temporal.io/api v1.62.2 // indirect
+	go.temporal.io/api v1.62.5 // indirect
 	go.uber.org/atomic v1.11.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	go.uber.org/zap v1.27.0 // indirect
--- a/contrib/datadog/go.sum
+++ b/contrib/datadog/go.sum
@@ -233,8 +233,8 @@ go.opentelemetry.io/proto/slim/otlp/collector/profiles/v1development v0.0.1 h1:T
 go.opentelemetry.io/proto/slim/otlp/collector/profiles/v1development v0.0.1/go.mod h1:riqUmAOJFDFuIAzZu/3V6cOrTyfWzpgNJnG5UwrapCk=
 go.opentelemetry.io/proto/slim/otlp/profiles/v1development v0.0.1 h1:z/oMlrCv3Kopwh/dtdRagJy+qsRRPA86/Ux3g7+zFXM=
 go.opentelemetry.io/proto/slim/otlp/profiles/v1development v0.0.1/go.mod h1:C7EHYSIiaALi9RnNORCVaPCQDuJgJEn/XxkctaTez1E=
-go.temporal.io/api v1.62.2 h1:jFhIzlqNyJsJZTiCRQmTIMv6OTQ5BZ57z8gbgLGMaoo=
-go.temporal.io/api v1.62.2/go.mod h1:iaxoP/9OXMJcQkETTECfwYq4cw/bj4nwov8b3ZLVnXM=
+go.temporal.io/api v1.62.5 h1:9R/9CeyM7xqHSlsNt+QIvapQLcRxCZ38bnXQx4mCN6I=
+go.temporal.io/api v1.62.5/go.mod h1:iaxoP/9OXMJcQkETTECfwYq4cw/bj4nwov8b3ZLVnXM=
 go.uber.org/atomic v1.9.0/go.mod h1:fEN4uk6kAWBTFdckzkM89CLk9XfWZrxpCo0nPH17wJc=
 go.uber.org/atomic v1.11.0 h1:ZvwS0R+56ePWxUNi+Atn9dWONBPp/AUETXlHW0DxSjE=
 go.uber.org/atomic v1.11.0/go.mod h1:LUxbIzbOniOlMKjJjyPfpl4v+PKK2cNJn91OQbhoJI0=
--- a/contrib/envconfig/go.mod
+++ b/contrib/envconfig/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/robfig/cron v1.2.0 // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
-	go.temporal.io/api v1.62.2 // indirect
+	go.temporal.io/api v1.62.5 // indirect
 	golang.org/x/net v0.39.0 // indirect
 	golang.org/x/sync v0.13.0 // indirect
 	golang.org/x/sys v0.32.0 // indirect
--- a/contrib/envconfig/go.sum
+++ b/contrib/envconfig/go.sum
@@ -37,8 +37,8 @@ github.com/stretchr/testify v1.10.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf
 github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.3.5/go.mod h1:mwnBkeHKe2W/ZEtQ+71ViKU8L12m81fl3OWwC1Zlc8k=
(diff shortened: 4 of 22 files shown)


internal/internal_task_pollers.go
```diff
@@ -514,13 +546,55 @@ func (wtp *workflowTaskProcessor) RespondTaskCompletedWithMetrics(
...
+		}
+		return
+	}
+
+	taskDuration := time.Since(startTime)
+	metricsHandler.Timer(metrics.WorkflowTaskExecutionLatency).Record(taskDuration)
 
 	response, err = wtp.sendTaskCompletedRequest(taskCompletion, task)
 
+	if taskDuration > 5*time.Second {
```
jmaeagle99: Add debug level logging to unconditionally produce this information.

converter/extstore.go
```diff
@@ -0,0 +1,114 @@
...
+	// Store persists the given payloads and returns one StorageClaim per
+	// payload in the same order. The returned claims are serialized into the
+	// Temporal event and must contain enough information for Retrieve to locate
+	// the data. Store must not modify the input payloads.
+	Store(ctx StorageDriverContext, payloads []*commonpb.Payload) ([]StorageClaim, error)
+
+	// Retrieve fetches the payloads identified by the given claims, returning
+	// one payload per claim in the same order. It must not modify the input
+	// claims.
+	Retrieve(ctx StorageDriverContext, claims []StorageClaim) ([]*commonpb.Payload, error)
```
jmaeagle99: Consider separate context type for the retrieval path so that it doesn't get ambient contextual information (such as serialization context) since storage claim should contain all information to retrieve. But we'll still want to provide `context.Context` for, at least, cancellation, even if the context doesn't have anything else in it.

converter/extstore.go
```diff
@@ -0,0 +1,114 @@
...
+// payload, or returns nil to leave the payload inline (not stored externally).
+// Use this when different payloads should be routed to different backends. For
+// example, routing large binary blobs to object storage while keeping small
+// JSON payloads inline. If no selector is set, the first driver in
+// StorageOptions.Drivers is used for every payload that exceeds the size
+// threshold.
+//
+// NOTE: Experimental
+type StorageDriverSelector interface {
+	SelectDriver(StorageDriverContext, *commonpb.Payload) (StorageDriver, error)
```
jmaeagle99: Consider returning `string` instead of `StorageDriver`. If we are validating that the driver exists in the driver list by name and we don't actually use the driver returned from the selector, then there isn't much use in providing the full driver. Users might implement another driver with the same name expecting that it would actually be used, but we find one with the same name in the registered driver list and use that instead. Let's use `string` to narrow down to what is actually needed and expected.

converter/extstore.go
```diff
@@ -0,0 +1,114 @@
...
+
+// StorageOptions configures external payload storage for a Temporal client or
+// worker. When set, the SDK intercepts payloads on the way to and from the
+// Temporal server: payloads that exceed PayloadSizeThreshold are offloaded to
+// an external store by the configured driver(s), and storage references are
+// substituted in the history event. The references are resolved back to the
+// original payloads before they reach the data converter or application code.
+//
+// NOTE: Experimental
+type StorageOptions struct {
```
yuandrew: Talked about maybe renaming this to `StorageProvider`, per Quinn's suggestion

internal/extstore.go
```diff
@@ -0,0 +1,359 @@
...
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/protobuf/proto"
+)
+
+const defaultPayloadSizeThreshold = 256 * 1024
+
+type storageParameters struct {
+	driverMap            map[string]converter.StorageDriver
+	driverSelector       converter.StorageDriverSelector
+	firstDriver          converter.StorageDriver
```
yuandrew: hmm, would it be more clear to call `firstDriver` something like `defaultDriver`? That way it's more clear this is the fallback if `driverSelector` is nil? I guess the current name makes it clear that it's literally the first driver in `options.Drivers`

internal/extstore.go
```diff
@@ -0,0 +1,359 @@
...
+		return storageReference{}, fmt.Errorf("payload is not a storage reference: unexpected encoding %q", string(p.GetMetadata()[converter.MetadataEncoding]))
+	}
+	var ref storageReference
+	if err := json.Unmarshal(p.Data, &ref); err != nil {
+		return storageReference{}, fmt.Errorf("failed to unmarshal storage reference: %w", err)
+	}
+	return ref, nil
+}
+
+type storageRetrievalVisitor struct {
```
yuandrew: nit: `store`  in `storageStoreVisitor` is a verb, but `retrieval` in `storageRetrievalVisitor` is a noun. I guess the ideal would be `storage`, but that would be `storageStorageVisitor`. How is `externalStorageVisitor` and `externalRetrievalVisitor`? That also doesn't sound perfect either... maybe this is fine

internal/extstore_test.go
```diff
@@ -0,0 +1,1033 @@
...
+	require.NoError(t, err)
+	require.Equal(t, 1024, params.payloadSizeThreshold)
+}
+
+func TestStorageOptionsToParams_DuplicateDriverNames(t *testing.T) {
+	_, err := StorageOptionsToParams(converter.StorageOptions{
+		Drivers: []converter.StorageDriver{newTestDriver("same"), newTestDriver("same")},
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "same")
```
yuandrew: should we assert the error message? `duplicate storage driver name: same`?

internal/internal_task_pollers.go
```diff
@@ -685,6 +767,32 @@ func (wtp *workflowTaskProcessor) errorToFailWorkflowTaskWithCause(taskToken []b
 	return builtRequest
 }
 
+type workflowTaskStorageMetrics struct {
+	mu                 sync.Mutex
+	PayloadCount       int
```
yuandrew: These fields are exported but `workflowTaskStorageMetrics` itself isn't, these fields probably don't need to be exported

internal/internal_task_pollers.go
```diff
@@ -620,6 +696,9 @@ func (wtp *workflowTaskProcessor) reportGrpcMessageTooLarge(
 		emitFailMetric = true
 		request := wtp.errorToFailWorkflowTask(task.TaskToken, sendErr)
 		request.Cause = enumspb.WORKFLOW_TASK_FAILED_CAUSE_GRPC_MESSAGE_TOO_LARGE
+		if err = visitProtoPayloads(ctx, wtp.outboundPayloadVisitor, request); err != nil {
+			return
```
yuandrew, line 701: should we log the `err` here and below?

internal/payload_visitor.go
```diff
@@ -0,0 +1,39 @@
...
+func visitPayload(ctx context.Context, visitor PayloadVisitor, p *commonpb.Payload) (*commonpb.Payload, error) {
+	if visitor == nil {
+		return p, nil
+	}
+	vpc := &proxy.VisitPayloadsContext{Context: ctx}
+	visited, err := visitor.Visit(vpc, []*commonpb.Payload{p})
+	if err != nil {
+		return nil, err
+	}
+	return visited[0], nil
```
yuandrew, line 41: could `len(visited)` every be 0, causing a panic here?

converter/extstore.go
jmaeagle99: Decided to have the selector return the driver pointer.

converter/extstore.go
jmaeagle99: Renamed to `ExternalStorage`.

yuandrew (approved):
🥳

PR #2225, opened 2026-03-10
Add Serialization context for codecs and converters

### LINKED ISSUE #1352: Context aware data converters cannot get IDs of outbound activities and child workflow

yuandrew:
A re-creation of https://github.com/temporalio/sdk-go/pull/2149, as that did not properly capture the feature.

## What was changed

  Add SerializationContext — an opt-in mechanism that provides metadata (namespace, workflow ID, activity type,
  etc.) to DataConverter, PayloadCodec, and FailureConverter implementations during serialization/deserialization.

  New public API (converter package):
  - `SerializationContext` sealed interface with two implementations:
    - `WorkflowSerializationContext` — for workflow-level payloads (workflow input/result, child workflow, signal,
  query, update, memo, continue-as-new)
    - `ActivitySerializationContext` — for activity-level payloads (activity input/result, heartbeat, failure details)
  - Three opt-in interfaces: `DataConverterWithSerializationContext`, `PayloadCodecWithSerializationContext`,
  `FailureConverterWithSerializationContext`
  - `WithDataConverterSerializationContext()` / `WithFailureConverterSerializationContext()` helper functions
  - `CodecDataConverter.WithSerializationContext` auto-propagates context to both parent DC and codecs
  - `CompleteActivityWithOptions` / `CompleteActivityByIDWithOptions` / `CompleteActivityByActivityIDWithOptions` / `RecordActivityHeartbeatWithOptions` / `RecordActivityHeartbeatByIDWithOptions` - New client functions that allow users to pass in proper activity metadata to pass into codecs. The existing methods don't have sufficient context information.

<details>

<summary>Context is provided at every serialization site:</summary>

  - Workflow execution (input decode/result encode)
  - Activity and local activity (input encode/decode, result, heartbeat, failure)
  - Child workflow (input encode/decode, result)
  - Signal (including external signal)
  - Query (input/result)
  - Update (input/result/rejection)
  - Continue-as-new (input)
  - Side effect / mutable side effect (uses workflow-scoped DC)
  - Client operations (start, signal, query, terminate, update, describe)
  - Schedule client (create/describe actions)
  - Async activity completion by ID (complete, heartbeat)
  - Workflow task failure and query failure (FC wrapping)

</details>

## Why?

  Enables use cases like per-workflow encryption (using workflow ID as associated data for a key), context-dependent
   encoding, or audit logging — without requiring custom interceptors or thread-local hacks.

  Matches the serialization context feature already implemented in Python, .NET, and Java.

## Checklist

1. Closes #1352 

2. How was this tested:
    - Unit tests in converter/serialization_context_test.go — codec propagation matrix, signing round-trips, helper
  behavior
    - Workflow environment tests in internal/serialization_context_test.go — activity, local activity, child
  workflow, side effect, concurrent operations, signal external, query, update
    - Client mock tests in internal/serialization_context_client_test.go — StartWorkflow, SignalWorkflow,
  QueryWorkflow, TerminateWorkflow
    - Integration tests in test/serialization_context_test.go — end-to-end with signing codec against real server
  (side effect + activity + child workflow combined)

3. Any docs updates needed?

---

> [!NOTE]
> **High Risk**
> High risk because it changes core serialization/encoding behavior across workflows, activities, client APIs, schedules, and the test harness; incorrect context propagation could cause payload decode failures or incompatibilities for users with custom converters/codecs.
> 
> **Overview**
> Adds an opt-in `converter.SerializationContext` API (workflow/activity metadata) and plumbs it through `DataConverter`, `PayloadCodec`, and `FailureConverter` via new `With*SerializationContext` interfaces/helpers and `CodecDataConverter.WithSerializationContext` propagation.
> 
> Updates workflow/activity execution and client paths to wrap converters with the correct workflow/activity context (including pre-generating activity/child workflow IDs so codecs can use them), and stores per-activity/per-child-workflow converters so cancellations/failures decode with matching context.
> 
> Extends the public client API with `*WithOptions` variants for async activity completion/heartbeat that accept activity metadata, and adds broad unit/integration test coverage to validate round-trip behavior and context capture.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c4ec0bf6e06d810b6048f479e40b3993e8b3dc06. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -91,6 +91,7 @@ type (
 		unclosedClients        *int32
 		inboundPayloadVisitor  PayloadVisitor
 		outboundPayloadVisitor PayloadVisitor
+		storageDriverTypes     []string
 	}
 
 	// namespaceClient is the client for managing namespaces.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,6 +37,10 @@ All PR titles should start with Upper case.
 
 ## Testing
 
+Tests are managed through the build tool at `internal/cmd/build`. This tool handles starting an embedded Temporal dev
+server with the required dynamic configs and search attributes, enforces consistent test flags (`-race`, `-count 1`, no caching),
+and manages coverage collection — so you don't need to manually configure a server or remember the right flags.
+
 Run all static analysis tools:
 
 ```bash
@@ -44,18 +48,44 @@ cd ./internal/cmd/build
 go run . check
 ```
 
-Run the integration tests (requires local server running, or pass `-dev-server`):
+### Integration Tests
+
+Integration tests live in the `test/` directory and require a Temporal server by default. Use `-dev-server` to start an
+embedded server automatically:
 
 ```bash
 cd ./internal/cmd/build
-go run . integration-test
+go run . integration-test -dev-server
 ```
 
-Run the unit tests:
+Run a specific test with `-run` (uses the same syntax as `go test -run`):
+
+```bash
+# Run a single test within a suite
+cd ./internal/cmd/build
+go run . integration-test -dev-server -run "TestIntegrationSuite/TestMyTest"
+
+# Run all tests in a suite
+cd ./internal/cmd/build
+go run . integration-test -dev-server -run "TestWorkerTunerTestSuite"
+```
+
+Without `-dev-server`, the tests connect to a server already running on `localhost:7233`.
+
+### Unit Tests
+
+Unit tests cover all packages except `test/`:
 
 ```bash
 cd ./internal/cmd/build
 go run . unit-test
+```
+
+Run specific unit tests with `-run`:
+
+```bash
+cd ./internal/cmd/build
+go run . unit-test -run "TestMyFunction"
 ```
 
 ## Updating go mod files
--- a/internal/client.go
+++ b/internal/client.go
@@ -1249,7 +1249,6 @@ func newClient(ctx context.Context, options ClientOptions, existing Client) (Cli
 
 	if options.Logger == nil {
 		options.Logger = ilog.NewDefaultLogger()
-		options.Logger.Info("No logger configured for temporal client. Created default one.")
 	}
 
 	// Validate mutually exclusive TLS options
@@ -1405,6 +1404,8 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
 		panic(fmt.Sprintf("invalid ExternalStorage options: %v", err))
 	}
 
+	storageDriverTypes := collectStorageDriverTypes(options.ExternalStorage.Drivers)
+
 	client := &WorkflowClient{
 		workflowService:          workflowServiceClient,
 		conn:                     conn,
@@ -1428,6 +1429,7 @@ func NewServiceClient(workflowServiceClient workflowservice.WorkflowServiceClien
 		workerGroupingKey:       uuid.NewString(),
 		inboundPayloadVisitor:   NewExternalRetrievalVisitor(storageParams),
 		outboundPayloadVisitor:  NewExternalStorageVisitor(storageParams),
+		storageDriverTypes:      storageDriverTypes,
 	}
 
 	if heartbeatInterval > 0 {
@@ -1651,4 +1653,21 @@ func SetResponseInfoOnStartWorkflowOptions(opts *StartWorkflowOptions) *startWor
 		opts.responseInfo = &startWorkflowResponseInfo{}
 	}
 	return opts.responseInfo
+}
+
+// collectStorageDriverTypes returns deduplicated driver types from the given drivers.
+func collectStorageDriverTypes(drivers []converter.StorageDriver) []string {
+	if len(drivers) == 0 {
+		return nil
+	}
+	seen := make(map[string]struct{}, len(drivers))
+	result := make([]string, 0, len(drivers))
+	for _, d := range drivers {
+		t := d.Type()
+		if _, found := seen[t]; !found {
+			seen[t] = struct{}{}
+			result = append(result, t)
+		}
+	}
+	return result
 }
--- a/internal/cmd/build/main.go
+++ b/internal/cmd/build/main.go
@@ -121,7 +121,7 @@ func (b *builder) integrationTest() error {
 	if *devServerFlag {
 		devServer, err := testsuite.StartDevServer(context.Background(), testsuite.DevServerOptions{
 			CachedDownload: testsuite.CachedDownload{
-				Version: "v1.6.1-server-1.31.0-151.0",
+				Version: "v1.6.2-server-1.31.0-151.6",
 			},
 			ClientOptions: &client.Options{
 				HostPort:  "127.0.0.1:7233",
@@ -161,6 +161,7 @@ func (b *builder) integrationTest() error {
 				"--dynamic-config-value", `component.nexusoperations.useSystemCallbackURL=false`,
 				"--dynamic-config-value", `component.nexusoperations.callback.endpoint.template="http://localhost:7243/namespaces/{{.NamespaceName}}/nexus/callback"`,
 				"--dynamic-config-value", "frontend.ListWorkersEnabled=true",
+				"--dynamic-config-value", "frontend.enableCancelWorkerPollsOnShutdown=true",
 			},
 		})
 		if err != nil {
--- a/internal/internal_nexus_task_poller.go
+++ b/internal/internal_nexus_task_poller.go
@@ -37,13 +37,15 @@ func newNexusTaskPoller(
 ) *nexusTaskPoller {
 	return &nexusTaskPoller{
 		basePoller: basePoller{
-			metricsHandler:          params.MetricsHandler,
-			stopC:                   params.WorkerStopChannel,
-			workerBuildID:           params.getBuildID(),
-			useBuildIDVersioning:    params.UseBuildIDForVersioning,
-			workerDeploymentVersion: params.DeploymentOptions.Version,
-			capabilities:            params.capabilities,
-			pollTimeTracker:         params.pollTimeTracker,
+			metricsHandler:               params.MetricsHandler,
+			stopC:                        params.WorkerStopChannel,
+			workerBuildID:                params.getBuildID(),
+			useBuildIDVersioning:         params.UseBuildIDForVersioning,
+			workerDeploymentVersion:      params.DeploymentOptions.Version,
+			capabilities:                 params.capabilities,
+			pollTimeTracker:              params.pollTimeTracker,
+			workerInstanceKey:            params.workerInstanceKey,
+			workerPollCompleteOnShutdown: params.workerPollCompleteOnShutdown,
 		},
 		taskHandler:     taskHandler,
 		service:         service,
@@ -80,6 +82,7 @@ func (ntp *nexusTaskPoller) poll(ctx context.Context) (taskForWorker, error) {
 			ntp.useBuildIDVersioning,
 			ntp.workerDeploymentVersion,
 		),
+		WorkerInstanceKey: ntp.workerInstanceKey,
 	}
 
 	response, err := ntp.pollNexusTaskQueue(ctx, request)
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"go.temporal.io/sdk/internal/common/retry"
@@ -84,6 +85,8 @@ type (
 		pollTimeTracker *pollTimeTracker
 		// Unique identifier for worker
 		workerInstanceKey string
+		// Server cancels polls on shutdown
+		workerPollCompleteOnShutdown *atomic.Bool
 	}
 
 	// numPollerMetric tracks the number of active pollers and publishes a metric on it.
@@ -306,6 +309,28 @@ func (bp *basePoller) doPoll(pollFunc func(ctx context.Context) (taskForWorker, 
 		close(doneC)
 	}()
 
+	if bp.workerPollCompleteOnShutdown != nil && bp.workerPollCompleteOnShutdown.Load() {
+		// Don't kill the gRPC stream. After ShutdownWorker, the server returns empty responses.
+		select {
+		case <-doneC:
+			return result, err
+		case <-bp.stopC:
+			// TEMP FIX: Give the server a reasonable window to complete the poll after
+			// ShutdownWorker. Fall back to cancelling the poll if it takes too
+			// long, e.g. when the gRPC connection was closed before Stop().
+			timer := time.NewTimer(5 * time.Second)
+			defer timer.Stop()
+			select {
+			case <-doneC:
+			case <-timer.C:
+				cancel()
+				<-doneC
+			}
+			return result, err
+		}
+	}
+
+	// Legacy: cancel in-flight polls immediately on shutdown
 	select {
 	case <-doneC:
 		return result, err
@@ -336,14 +361,15 @@ func newWorkflowTaskProcessor(
 ) *workflowTaskProcessor {
 	return &workflowTaskProcessor{
 		basePoller: basePoller{
-			metricsHandler:          params.MetricsHandler,
-			stopC:                   params.WorkerStopChannel,
-			workerBuildID:           params.getBuildID(),
-			useBuildIDVersioning:    params.UseBuildIDForVersioning,
-			workerDeploymentVersion: params.DeploymentOptions.Version,
-			capabilities:            params.capabilities,
-			pollTimeTracker:         params.pollTimeTracker,
-			workerInstanceKey:       params.workerInstanceKey,
+			metricsHandler:               params.MetricsHandler,
+			stopC:                        params.WorkerStopChannel,
+			workerBuildID:                params.getBuildID(),
+			useBuildIDVersioning:         params.UseBuildIDForVersioning,
+			workerDeploymentVersion:      params.DeploymentOptions.Version,
+			capabilities:                 params.capabilities,
+			pollTimeTracker:              params.pollTimeTracker,
+			workerInstanceKey:            params.workerInstanceKey,
+			workerPollCompleteOnShutdown: params.workerPollCompleteOnShutdown,
 		},
 		service:                      service,
 		namespace:                    params.Namespace,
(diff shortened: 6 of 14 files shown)


converter/serialization_context.go
```diff
@@ -0,0 +1,105 @@
...
+// and may call this method frequently. Avoid recreating expensive objects on every call.
+type FailureConverterWithSerializationContext interface {
+	WithSerializationContext(SerializationContext) FailureConverter
+}
+
+// WithSerializationContext returns a DataConverter that is aware of the given
+// serialization context. If the DataConverter implements
+// [DataConverterWithSerializationContext], it delegates to that implementation;
+// otherwise it returns the original DataConverter unchanged.
+func WithSerializationContext(dc DataConverter, ctx SerializationContext) DataConverter {
```
jmaeagle99: Maybe rename to `WithDataConverterSerialziationContext` to avoid leading users into thinking this _might_ generically handle all `WithSerializationContext` implementers. I know parameters matter, but type completion doesn't show that many times.

converter/serialization_context_test.go
```diff
@@ -0,0 +1,223 @@
...
+	codec := NewZlibCodec(ZlibCodecOptions{AlwaysEncode: true})
+	cdc := NewCodecDataConverter(parentDC, codec)
+
+	ctx := WorkflowSerializationContext{WorkflowID: "wf-4"}
+	wrapped := WithSerializationContext(cdc, ctx)
+
+	require.Same(cdc, wrapped)
+}
+
+func TestCodecDataConverter_SigningRoundTrip(t *testing.T) {
```
jmaeagle99: nit: Rename to indicate the failure that is being tested.

converter/serialization_context_test.go
```diff
@@ -0,0 +1,223 @@
...
+	require.Equal("hello", result)
+
+	// Different context fails to decode
+	wrapped2 := WithSerializationContext(cdc, WorkflowSerializationContext{WorkflowID: "wf-B"})
+	err = wrapped2.FromPayload(payload, &result)
+	require.Error(err)
+	require.Contains(err.Error(), "signature mismatch")
+}
+
+func TestCodecDataConverter_SigningRoundTrip_ActivityContext(t *testing.T) {
```
jmaeagle99: nit: Rename to indicate the failure that is being tested.

converter/serialization_context_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// Different activity type fails
+	otherCtx := ActivitySerializationContext{WorkflowID: "wf-1", ActivityType: "OtherActivity"}
+	wrapped2 := WithSerializationContext(cdc, otherCtx)
+	err = wrapped2.FromPayload(payload, &result)
+	require.Error(err)
+	require.Contains(err.Error(), "signature mismatch")
+}
+
+func TestWithSerializationContextHelper_NoOp(t *testing.T) {
```
jmaeagle99: `TestDefaultDataConverter_NotSerializationContextAware`?

internal/workflow.go
```diff
@@ -1356,14 +1352,21 @@ func (wc *workflowEnvironmentInterceptor) ExecuteChildWorkflow(ctx Context, chil
...
-		attempt:         1,
+		WorkflowOptions:  *options,
+		Input:            input,
+		WorkflowType:     wfType,
+		Header:           header,
+		scheduledTime:    Now(ctx), /* this is needed for test framework, and is not send to server */
+		attempt:          1,
+		failureConverter: failureConverter,
 	}
+	params.WorkflowID = childWorkflowID
```
jmaeagle99: nit: Could this be part of the ExecuteWorkflowParams initialization instead of setting afterwards?

internal/internal_event_handlers.go
```diff
@@ -1495,7 +1494,7 @@ func (weh *workflowExecutionEventHandlerImpl) handleActivityTaskFailed(event *hi
 		&commonpb.ActivityType{Name: activity.activityType.Name},
 		activityID,
 		attributes.GetRetryState(),
-		weh.GetFailureConverter().FailureToError(attributes.GetFailure()),
+		cmp.Or(activity.failureConverter, weh.GetFailureConverter()).FailureToError(attributes.GetFailure()),
```
jmaeagle99: Under what condition is `activity.failureConverter == nil`? Can we update all the callers to ensure it's not nil and assert that somewhere?

converter/serialization_context.go
```diff
@@ -0,0 +1,105 @@
...
+	if sc, ok := dc.(DataConverterWithSerializationContext); ok {
+		return sc.WithSerializationContext(ctx)
```
jmaeagle99: Should we check that the result is not `nil` and return the original converter instead?

converter/serialization_context.go
```diff
@@ -0,0 +1,105 @@
...
+	return dc
+}
+
+// WithFailureConverterSerializationContext returns a FailureConverter that is aware of the given
+// serialization context. If the FailureConverter implements
+// [FailureConverterWithSerializationContext], it delegates to that implementation;
+// otherwise it returns the original FailureConverter unchanged.
+func WithFailureConverterSerializationContext(fc FailureConverter, ctx SerializationContext) FailureConverter {
+	if sc, ok := fc.(FailureConverterWithSerializationContext); ok {
+		return sc.WithSerializationContext(ctx)
```
jmaeagle99: Same comment about checking result is not `nil`.

internal/internal_update.go
```diff
@@ -170,7 +175,7 @@ func (up *updateProtocol) Reject(err error) {
 			RejectedRequestMessageId:         up.requestMsgID,
 			RejectedRequestSequencingEventId: up.requestSeqID,
 			RejectedRequest:                  up.initialRequest,
-			Failure:                          up.env.GetFailureConverter().ErrorToFailure(err),
+			Failure:                          cmp.Or(up.failureConverter, up.env.GetFailureConverter()).ErrorToFailure(err),
```
jmaeagle99: When is `up.failureConverter == nil`? Can we update all the callers to ensure it's not nil and assert that somewhere?

jmaeagle99: Same for `up.dataConverter`

internal/internal_update.go
```diff
@@ -272,8 +278,13 @@ func defaultUpdateHandler(
 			return
 		}
 
+		if up, ok := callbacks.(*updateProtocol); ok {
+			up.dataConverter = handler.dataConverter
```
jmaeagle99, line 283: Why is this needed? And should it be the other way around?

```
handler.dataConverter = up.dataConverter
handler.failureConverter = up.failureConverter
```

seems that `up.dataConverter` and `up.failureConverter` are not used beyond this point.

internal/internal_workflow_client.go
```diff
@@ -561,7 +570,10 @@ func (wc *WorkflowClient) RecordActivityHeartbeatByID(ctx context.Context,
 		return err
 	}
 
-	dataConverter := WithContext(ctx, wc.dataConverter)
+	dataConverter := converter.WithSerializationContext(WithContext(ctx, wc.dataConverter), converter.WorkflowSerializationContext{
```
jmaeagle99: Should this be ActivitySerializationContext?

internal/internal_workflow_client.go
```diff
@@ -497,7 +501,12 @@ func (wc *WorkflowClient) CompleteActivityByID(ctx context.Context, namespace, w
 		return errors.New("empty activity or workflow id or namespace")
 	}
 
-	dataConverter := WithContext(ctx, wc.dataConverter)
+	wfCtx := converter.WorkflowSerializationContext{
```
jmaeagle99: `ActivitySerializationContext`?

internal/internal_workflow_client.go
jmaeagle99: I think this needs `converter.WithSerializationContext` applied.

internal/internal_workflow_client.go
yuandrew: Added `CompleteActivityByIDWithOptions` to get enough context for `ActivitySerializationContext`

internal/internal_workflow_client.go
yuandrew: added RecordActivityHeartbeatByIDWithOptions to get enough context for `ActivitySerializationContext`

internal/internal_update.go
yuandrew: These are used in `func (up *updateProtocol) Reject(err error) {` and `func (up *updateProtocol) Complete(success interface{}, outcomeErr error) {`

internal/internal_update.go
yuandrew: sure, added default initialization from env in `newUpdateProtocol()`

converter/serialization_context.go
yuandrew: Hmm, if result is `nil`, that is clearly a user bug, and I wanna say in this case, a runtime panic is noisier, thus better than encrypting with the wrong key, or potentially no key at all.

Actually, what do you think about us intentionally panic'ing with something like `panic("WithSerializationContext must not return nil")`, that way users don't see `  panic: runtime error: invalid memory address or nil pointer dereference` and have no idea what's causing it? Changed to this idea, but happy to keep discussing

also added a test to verify this and failureConverter being nil

internal/internal_event_handlers.go
yuandrew: Most places using `cmp.Or` only have their `converter == nil` in test paths, but you bring up a good point, it's probably better to ensure failureConverter is set in tests than have this special code to accept special test scenarios.

internal/workflow.go
yuandrew: This would go into WorkflowOptions, but yes you're right, it should be fine to set above

converter/serialization_context_test.go
yuandrew: good idea

converter/serialization_context.go
yuandrew: good point

internal/internal_workflow_client.go
```diff
@@ -463,28 +467,46 @@ func (wc *WorkflowClient) getWorkflowExecutionHistory(ctx context.Context, rpcMe
...
 	}
 
-	if taskToken == nil {
+	if opts.TaskToken == nil {
 		return errors.New("invalid task token provided")
 	}
 
-	dataConverter := WithContext(ctx, wc.dataConverter)
+	actCtx := converter.ActivitySerializationContext{
+		Namespace:    opts.Namespace,
```
yuandrew: Chose to intentionally not default back to `wc.Namespace`, and instead have users pass it in, since it would technically be a different namespace

internal/error.go
```diff
@@ -621,7 +621,13 @@ func (wc *workflowEnvironmentInterceptor) NewContinueAsNewError(
 		panic("context is missing required options for continue as new")
 	}
 	env := getWorkflowEnvironment(ctx)
-	workflowType, input, err := getValidatedWorkflowFunction(wfn, args, options.DataConverter, env.GetRegistry())
+	wfInfo := env.WorkflowInfo()
+	wfCtx := converter.WorkflowSerializationContext{
+		Namespace:  wfInfo.Namespace,
+		WorkflowID: wfInfo.WorkflowExecution.ID,
+	}
+	dc := converter.WithDataConverterSerializationContext(getDataConverterFromWorkflowContext(ctx), wfCtx)
```
jmaeagle99: Cursor bot already pointed this out, but it seems that `getDataConverterFromWorkflowContext` already applies the serialization context, so reapplying it again shouldn't be necessary. This seems to be a common pattern throughout this PR.

internal/workflow.go
```diff
@@ -1696,7 +1756,12 @@ func signalExternalWorkflow(ctx Context, workflowID, runID, signalName string, a
 		return future
 	}
 
-	dataConverter := getDataConverterFromWorkflowContext(ctx)
+	dataConverter := converter.WithSerializationContext(
+		getDataConverterFromWorkflowContext(ctx),
+		converter.WorkflowSerializationContext{
+			Namespace:  options.Namespace,
+			WorkflowID: workflowID,
+		})
```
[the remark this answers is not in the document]
jmaeagle99: I saw that this was resolved, but it looks like the issue still exists? Or was this somehow fixed somewhere else? Please feel free to resolve again if I missed something.

internal/internal_workflow_testsuite.go
jmaeagle99: This might need to be wrapped with serialization context. The SDK seems to do that, so I think this test infra needs it as well.

internal/internal_event_handlers.go
```diff
@@ -754,18 +755,19 @@ func (wc *workflowEnvironmentImpl) ExecuteActivity(parameters ExecuteActivityPar
 		return ActivityID{}
 	}
 
-	command := wc.commandsHelper.scheduleActivityTask(scheduleID, scheduleTaskAttr, startMetadata)
+	command := wc.commandsHelper.scheduleActivityTask(parameters.ScheduleID, scheduleTaskAttr, startMetadata)
```
jmaeagle99, line 798: This looks correct for the Go SDK itself, because `workflowEnvironmentInterceptor.ExecuteActivity` generates the ID. But this affects the `bindings` API, so binding callers now need to call `env.GenerateSequence()` and set `ScheduleID`. As is, I think this _may_ be a breaking change for bindings callers (other SKDs built on top of the Go SDK).

internal/serialization_context_client_test.go
```diff
@@ -0,0 +1,137 @@
+package internal
```
jmaeagle99, line 1: Tests for DescribeWorkflow, PollWorkflowUpdate, and UpdateWithStartWorkflow?

internal/serialization_context_test.go
```diff
@@ -0,0 +1,552 @@
...
+			foundLocal = true
+			s.NotEmpty(actCtx.WorkflowID)
+			s.NotEmpty(actCtx.ActivityType)
+			break
+		}
+	}
+	s.True(foundLocal, "should have captured ActivitySerializationContext with IsLocal=true")
+}
+
+// --- Child Workflow Tests ---
```
jmaeagle99: How about tests for the top-level workflow that has workflow args? I think all of the top-level workflows in these tests do not have args.

converter/serialization_context.go
```diff
@@ -0,0 +1,113 @@
...
+	TaskQueue    string
+	IsLocal      bool
+}
+
+func (ActivitySerializationContext) isSerializationContext() {}
+
+// DataConverterWithSerializationContext is an optional interface that [DataConverter]
+// implementations can implement to receive serialization context.
+//
+// When implemented, the SDK calls WithDataConverterSerializationContext before serializing/deserializing
```
jmaeagle99: nit: While technically correct, I think it's better to say that "the SDK calls WithSerializationContext" since that is the method that `DataConverter` is implementing. The fact that it is indirectly called through `WithDataConverterSerializationContext` shouldn't matter to the developer of the `DataConverter`.

internal/internal_workflow_testsuite.go
yuandrew: good call, added

internal/error.go
yuandrew: yeah.. I decided to just remove serialization context from `getDataConverterFromWorkflowContext` and require each caller to explicitly apply context, that way we avoid this double-applying, and it's more clear.

internal/internal_event_handlers.go
yuandrew: True, this will need to be called out in the PR description/release notes

internal/serialization_context_client_test.go
yuandrew: added

internal/serialization_context_test.go
yuandrew: added

internal/internal_event_handlers.go
yuandrew: Ended up adding in backwards compatible checks that will `GenerateSequence()` in the same places we do today if bindings callers don't mirror this new behavior, so we should be good, no longer a breaking change

internal/internal_workflow_client.go
yuandrew: jk, that would be breaking change behavior, will default back to wc.namespace

internal/internal_workflow_client.go
```diff
@@ -497,80 +520,156 @@ func (wc *WorkflowClient) CompleteActivity(ctx context.Context, taskToken []byte
...
+		RunID:      runID,
+		ActivityID: activityID,
+		Result:     result,
+		Err:        err,
+	})
+}
+
+// CompleteActivityByIDWithOptions reports workflow activity completed with full context options.
+func (wc *WorkflowClient) CompleteActivityByIDWithOptions(ctx context.Context, opts CompleteActivityByIDOptions) error {
+	if opts.ActivityID == "" || opts.WorkflowID == "" || opts.Namespace == "" {
```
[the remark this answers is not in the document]
yuandrew: CompleteActivityByID never had the initialization check

internal/workflow.go
```diff
@@ -1034,6 +1036,30 @@ func (wc *workflowEnvironmentInterceptor) ExecuteActivity(ctx Context, typeName
...
 
+	env := getWorkflowEnvironment(ctx)
+	// Generate activity ID before serialization so it's available to context-aware data converters
+	scheduleID := env.GenerateSequence()
+	var activityID string
+	if options.ActivityID != "" {
+		activityID = options.ActivityID
+	} else {
+		activityID = getStringID(scheduleID)
+	}
```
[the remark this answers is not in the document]
yuandrew: The old call for `GenerateSequence()` happened at the beginning of `workflowEnvironmentImpl.ExecuteActivity()`. now, they're called in the interceptor immediately before calling `env.ExecuteActivity()`. The order should be the same. In between the current `GenerateSequence` call and where it used to be is building context, wrapping converters, encoding args. There should be no sequence consumption.

PR #2226, opened 2026-03-11
Populate resource_id fields for request header routing

tconley1428:
## Summary

Implement logic to populate resource_id fields in request messages to support multi-cell request routing based on resource IDs.

This PR addresses the need to populate the newly added resource_id fields in various request messages so that the proxy can extract them and create appropriate temporal-resource-id headers for routing.

### Workflow Task Requests
- `RespondWorkflowTaskCompletedRequest.resource_id` → populated with workflow ID from task context
- `RespondWorkflowTaskFailedRequest.resource_id` → populated with workflow ID from task context

### Activity Task Requests
- `RespondActivityTaskCompletedRequest.resource_id` → workflow ID for workflow activities, activity ID for standalone
- `RespondActivityTaskCompletedByIdRequest.resource_id` → workflow ID for workflow activities, activity ID for standalone
- `RespondActivityTaskFailedRequest.resource_id` → workflow ID for workflow activities, activity ID for standalone
- `RespondActivityTaskFailedByIdRequest.resource_id` → workflow ID for workflow activities, activity ID for standalone
- `RespondActivityTaskCanceledRequest.resource_id` → workflow ID for workflow activities, activity ID for standalone
- `RespondActivityTaskCanceledByIdRequest.resource_id` → workflow ID for workflow activities, activity ID for standalone

### Activity Heartbeat Requests
- `RecordActivityTaskHeartbeatRequest.resource_id` → extracted from activity context
- `RecordActivityTaskHeartbeatByIdRequest.resource_id` → workflow ID for workflow activities, activity ID for standalone

### Batch Operation Requests
- `ExecuteMultiOperationRequest.resource_id` → workflow ID from the start operation

### Worker Requests
- `RecordWorkerHeartbeatRequest.resource_id` → worker grouping key from client

### Helper Functions Added
- `getActivityResourceId(workflowId, activityId)`: Determines appropriate resource ID (workflow ID if present, otherwise activity ID)
- `getActivityResourceIdFromCtx(ctx)`: Extracts resource ID from activity context using activityEnvironment

### Logic
- **Workflow activities**: Use the workflow ID as resource_id
- **Standalone activities**: Use the activity ID as resource_id  
- **Multi-operation requests**: Use the workflow ID from the first start operation
- **Worker requests**: Use the client's workerGroupingKey

## Test plan

- [x] Project builds successfully
- [x] `TestTemporalHeaderInterceptor` passes (validates header extraction works)
- [x] Activity heartbeat tests continue to pass
- [x] All resource_id fields are properly populated when requests are made

## Notes

- `RespondQueryTaskCompletedRequest` does not have a `resource_id` field
- `FetchWorkerConfigRequest` and `UpdateWorkerConfigRequest` resource_id fields were not implemented as these APIs are not yet implemented in the SDK
- Nexus requests were skipped per discussion
- The existing `TestTemporalHeaderInterceptor` validates that the temporal-resource-id header is correctly set to "workflow:test-workflow-id", confirming the entire pipeline works

🤖 Generated with [Claude Code](https://claude.ai/code)


internal/internal_worker_heartbeat.go
```diff
@@ -168,6 +168,7 @@ func (hw *sharedNamespaceWorker) sendHeartbeats() error {
 	_, err := hw.client.recordWorkerHeartbeat(hw.heartbeatCtx, &workflowservice.RecordWorkerHeartbeatRequest{
 		Namespace:       hw.namespace,
 		WorkerHeartbeat: heartbeats,
+		ResourceId:      hw.client.workerGroupingKey,
```
tconley1428: Follow up on whether this should actually be instance key

internal/internal_task_pollers.go
```diff
@@ -1507,6 +1508,16 @@ func reportActivityCompleteByID(
...
 }
 
+func getActivityResourceId(workflowId, activityId string) string {
+	if workflowId != "" {
+		return fmt.Sprintf("workflow:%s", workflowId)
+	}
+	if activityId != "" {
+		return fmt.Sprintf("activity:%s", activityId)
+	}
+	return ""
```
yuandrew, line 1588: What scenario do we hit this base case? Is it worth a log or something?

internal/internal_task_pollers.go
```diff
@@ -768,17 +768,18 @@ func (wtp *workflowTaskProcessor) errorToFailWorkflowTask(taskToken []byte, err
...
-func (wtp *workflowTaskProcessor) errorToFailWorkflowTaskWithCause(taskToken []byte, err error, cause enumspb.WorkflowTaskFailedCause) *workflowservice.RespondWorkflowTaskFailedRequest {
+func (wtp *workflowTaskProcessor) errorToFailWorkflowTaskWithCause(taskToken []byte, workflowId string, err error, cause enumspb.WorkflowTaskFailedCause) *workflowservice.RespondWorkflowTaskFailedRequest {
 	builtRequest := &workflowservice.RespondWorkflowTaskFailedRequest{
 		TaskToken:      taskToken,
 		Cause:          cause,
 		Failure:        wtp.failureConverter.ErrorToFailure(err),
 		Identity:       wtp.identity,
 		BinaryChecksum: wtp.workerBuildID,
 		Namespace:      wtp.namespace,
+		ResourceId:     fmt.Sprintf("workflow:%s", workflowId),
```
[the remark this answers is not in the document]
yuandrew: yeah, this is probably a good idea

internal/internal_task_pollers.go
```diff
@@ -1519,6 +1530,8 @@ func convertActivityResultToRespondRequest(
 	versionStamp *commonpb.WorkerVersionStamp,
 	deployment *deploymentpb.Deployment,
 	workerDeploymentOptions *deploymentpb.WorkerDeploymentOptions,
+	workflowId string,
```
yuandrew, line 1610: can we just pass in `resourceId` instead here? All uses for these 2 vars are to immediately call `getActivityResourceId`

internal/internal_workflow_client.go
```diff
@@ -512,7 +512,8 @@ func (wc *WorkflowClient) CompleteActivityWithOptions(ctx context.Context, opts
 	// We do allow canceled error to be passed here
 	cancelAllowed := true
 	request := convertActivityResultToRespondRequest(wc.identity, opts.TaskToken,
-		data, opts.Err, dataConverter, failureConverter, wc.namespace, cancelAllowed, nil, nil, nil)
+		data, opts.Err, dataConverter, failureConverter, wc.namespace, cancelAllowed, nil, nil, nil,
+		opts.WorkflowID, "")
```
yuandrew, line 526: `func (wc *WorkflowClient) CompleteActivity(` above never passes in WorkflowID, so this is always false in that case

internal/resource_id_test.go
```diff
@@ -0,0 +1,825 @@
...
+				}
+			case "canceled":
+				var canceledErr *CanceledError
+				if errors.As(tc.simulateError, &canceledErr) {
+					canceledRequest, ok := result.(*workflowservice.RespondActivityTaskCanceledRequest)
+					require.True(t, ok, "Result should be a RespondActivityTaskCanceledRequest")
+					assert.Equal(t, tc.expectedResourceID, canceledRequest.ResourceId, tc.description)
+				}
+			}
+		})
```
yuandrew, line 382: can we add a safety `default` case that fails the test?

internal/resource_id_test.go
```diff
@@ -0,0 +1,825 @@
...
+func testWorkflowTaskCompletedResourceID(t *testing.T) {
+	testCases := []struct {
+		name               string
+		workflowID         string
+		runID              string
+		expectedResourceID string
+	}{
+		{
+			name:               "StandardWorkflowTask",
+			workflowID:         "test-workflow-completed-123",
```
yuandrew, line 150: Is there value in testing with an empty `workflowID` both here and in `testWorkflowTaskFailedResourceID`?

internal/internal_task_handlers.go
```diff
@@ -2446,15 +2450,24 @@ func createNewCommandWithMetadata(commandType enumspb.CommandType, metadata *sdk
...
-		TaskToken: taskToken,
-		Details:   details,
-		Identity:  identity,
-		Namespace: namespace,
+		TaskToken:  taskToken,
+		Details:    details,
+		Identity:   identity,
+		Namespace:  namespace,
+		ResourceId: getActivityResourceIdFromCtx(ctx),
 	}
```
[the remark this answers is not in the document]
yuandrew: sounds legit

internal/internal_task_handlers.go
```diff
@@ -2446,15 +2450,24 @@ func createNewCommandWithMetadata(commandType enumspb.CommandType, metadata *sdk
...
 }
 
+func getActivityResourceIdFromCtx(ctx context.Context) string {
+	env := getActivityEnvironmentFromCtx(ctx)
+	if env == nil {
+		return ""
```
yuandrew, line 2630: do we want test coverage for when env == nil?

internal/resource_id_test.go
```diff
@@ -0,0 +1,825 @@
...
+			runID:              "test-run-completed-456",
+			expectedResourceID: "workflow:test-workflow-completed-123",
+		},
+		{
+			name:               "DifferentWorkflowID",
```
yuandrew: What's the point of this second test case? from my reading it's testing the exact same as the first run?

internal/resource_id_test.go
```diff
@@ -0,0 +1,825 @@
...
+	testCases := []struct {
+		name               string
+		groupingKey        string
+		expectedResourceID string
+	}{
+		{
+			name:               "WorkerHeartbeat",
+			groupingKey:        "test-worker-grouping-key-123",
+			expectedResourceID: "worker:test-worker-grouping-key-123",
+		},
```
yuandrew, line 769: single entry, no need for table style for this test

internal/internal_task_pollers.go
tconley1428: We could, it's not expected.

internal/internal_task_pollers.go
tconley1428: We could, but I thought it made more sense to keep the more actually meaningful values until necessary. They seem more useful than resource id. If you disagree, I can change it.

internal/internal_workflow_client.go
tconley1428: So is the activityserializationcontext here broken as well?

internal/resource_id_test.go
tconley1428: We could validate it, though it is actually an unexpected case. Workflow task without a workflow id?

internal/resource_id_test.go
tconley1428: Agreed

internal/internal_workflow_client.go
tconley1428: It only has a task token, so there's nothing we can do in this case.

yuandrew (commented):
LGTM, once we retarget this PR to `feature-mcn`, good to approve and merge

Issue #2206, opened 2026-02-26
Session activity cancellation causes permanent TMPRL1100 Non-Determinism Error on replay when DataConverter/codec fails


d2army:
## Expected Behavior

When a `DataConverter` (e.g., hitting the remote codec server) returns any error during `encodeArgs` in `ExecuteActivity`, the workflow task should fail and be retryable without triggering an NDE. The workflow should remain recoverable.

## Actual Behavior

Two bugs combine to permanently brick the workflow:

**Bug 1: `encodeArgs` panics instead of returning an error** (`internal/workflow.go:943`):
```go
input, err := encodeArgs(dataConverter, args)
if err != nil {
    panic(err) // Should fail the workflow task gracefully, not panic
}
```

**Bug 2: `handleCancelInitiatedEvent` rejects `Initiated` state on replay** (`internal/internal_command_state_machine.go:566-574`):
```go
func (d *commandStateMachineBase) handleCancelInitiatedEvent() {
    switch d.state {
    case commandStateCancellationCommandSent, commandStateCanceledAfterInitiated:
        // No state change
    default:
        d.failStateTransition(eventCancelInitiated) // Panics for Initiated state
    }
}
```

When a workflow uses `CreateSession` + `defer CompleteSession`, the panic from Bug 1 causes an `UNHANDLED_FAILURE`. On replay, the SDK encounters `ActivityTaskCancelRequested` for the `internalSessionCreationActivity` (the session creation activity scheduled by `CreateSession`, typically the first activity in the workflow, e.g. event ID 7 in our case) before the workflow code has re-executed `CompleteSession()` to transition the state machine to `CancellationCommandSent`. The state machine is still in `Initiated`, so Bug 2 panics with TMPRL1100. Every subsequent replay hits the same panic -- the workflow is permanently unrecoverable.

**Note:** Even if Bug 1 is not fixed (i.e., `encodeArgs` continues to panic), **Bug 2 alone** can still brick a workflow whenever any cancel event hits a session activity in `Initiated` state during replay. The state machine fix is required regardless of whether the panic-on-encode behavior is changed.

**Decoded failure from the bricked workflow:**
```
message: "failed to encode payloads: rpc error: code = DeadlineExceeded desc = context deadline exceeded"
stack_trace: "...workflowEnvironmentInterceptor.ExecuteActivity at workflow.go:943..."
```

**Decoded NDE from replay:**
```
[...]

PR #2228, opened 2026-03-13
Fix NDE due to task failure in sessions + cancellation

Sushisource:
## Summary

- Prevent workflow coroutine from yielding during panic unwinding, which was causing defers (e.g. `CompleteSession`) to commit commands as part of a successful WFT response
- Accept activity cancel in `Initiated` state in the command state machine

## Problem

When a `DataConverter` fails during `encodeArgs` in `ExecuteActivity`, the SDK panics. In a session workflow with `defer CompleteSession`, this panic triggers the defer which calls `.Get()` on a completion activity future. The `.Get()` call **yields the coroutine**, effectively "freezing" the panic and allowing the WFT to complete normally with the defer's commands (cancel session creation + schedule completion activity).

When the `DataConverter` later works again (e.g. transient codec server timeout resolves), the workflow takes a different code path (no panic, no `CompleteSession` commands), causing a permanent `TMPRL1100` nondeterminism error due to command mismatch with the history.

## Fix

Added a `panickedValue` field to `coroutineState`. Before panicking in `encodeArgs`, the error is stored in this field. In `yield()`, if `panickedValue` is set, it re-panics immediately instead of blocking. This prevents defers from yielding the coroutine and inadvertently committing commands. The panic propagates to the coroutine's `recover()`, and `applyWorkflowPanicPolicy` with `BlockWorkflow` fails the WFT cleanly — no commands are sent to the server.

Separately, the command state machine's `handleCancelInitiatedEvent` now accepts `commandStateInitiated`, allowing activity cancellation while in the Initiated state.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/2206

2. How was this tested:
Added UT and IT

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Touches core workflow execution (coroutine `yield`) and command state-machine cancellation transitions; mistakes here could cause new panics or replay/nondeterminism behavior changes, but changes are targeted and covered by new regression tests.
> 
> **Overview**
> Prevents workflow coroutines from **blocking/yielding during panic unwinding** by detecting panics via `runtime.gopanic` on the stack and re-panicking if `yield()` is invoked from a deferred cleanup path, avoiding committing commands on a workflow task that should fail.
> 
> Updates the internal command state machine to **accept `handleCancelInitiatedEvent` while in `commandStateInitiated`** (and treat repeat cancels as no-ops), addressing unusual replay orderings where `ActivityTaskCancelRequested` arrives before the SDK replays the cancel command.
> 
> Adds regression coverage (unit, worker replay, and integration) for issue `#2206`, including a session workflow that triggers a DataConverter-induced panic and a history replay with reordered cancel events.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c2d10f4450ee7282766733ce24b665a62b14d33f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


test/integration_test.go
```diff
@@ -8932,3 +8932,177 @@ func (ts *IntegrationTestSuite) TestExecuteActivitySuite() {
...
+
+func (f *poisonDataConverter) ToPayload(value interface{}) (*commonpb.Payload, error) {
+	if s, ok := value.(string); ok && s == f.poison {
+		return nil, fmt.Errorf("simulated codec server timeout: DeadlineExceeded")
+	}
+	return f.DataConverter.ToPayload(value)
+}
+
+// TestSessionCancelNDE is a regression test for
+// https://github.com/temporalio/sdk-go/issues/2206.
```
yuandrew: I don't think we need to include which issue this resolves, this information is already available from git blame, and looks like the only other time we've done this was earlier this year when Max also used CC to implement a fix

test/integration_test.go
```diff
@@ -8932,3 +8932,177 @@ func (ts *IntegrationTestSuite) TestExecuteActivitySuite() {
 		activityResultChan <- "" // allow activity to complete
 	})
 }
+
+// poisonDataConverter wraps a DataConverter and fails ToPayloads when any of the
+// values matches the poison string. Used by TestSessionCancelNDE to reproduce
+// https://github.com/temporalio/sdk-go/issues/2206.
```
yuandrew: Same comment as below, and even if we wanted to, this is redundant since the test already gives this info

test/integration_test.go
```diff
@@ -8932,3 +8932,177 @@ func (ts *IntegrationTestSuite) TestExecuteActivitySuite() {
...
+// activity. On replay, the ActivityTaskCancelRequested event for that activity
+// hits the state machine in Initiated state, causing a permanent TMPRL1100 NDE.
+func (ts *IntegrationTestSuite) TestSessionCancelNDE() {
+	// Stop the default worker — we need our own with a failing DataConverter.
+	ts.worker.Stop()
+	ts.workerStopped = true
+
+	// Create a DataConverter that fails when encoding the poison value.
+	// Session creation encodes a session ID (UUID), so it succeeds. The
+	// workflow's ExecuteActivity encodes the poison string, triggering the panic.
```
yuandrew: This comment feels unneeded, maybe only worth saying poison happens on `ExecuteActivity`?

test/integration_test.go
```diff
@@ -8932,3 +8932,177 @@ func (ts *IntegrationTestSuite) TestExecuteActivitySuite() {
...
+
+// A workflow panics on its first WFT attempt. During panic unwinding, a deferred function schedules
+// an activity (which yields the coroutine). Without the fix, this yield freezes the panic, the WFT
+// completes with the defer's activity command, and the second attempt (which doesn't panic) takes a
+// different code path causing an NDE.
+func (ts *IntegrationTestSuite) TestPanicWithDeferredYield() {
+	// Stop any existing worker
+	ts.worker.Stop()
+
+	var panicWithDeferYieldCounter atomic.Int32
```
yuandrew, line 9062: Any reason we don't just use `atomic.Bool` here?

test/workflow_test.go
```diff
@@ -3626,6 +3626,31 @@ func (w *Workflows) WorkflowReactToCancel(ctx workflow.Context, localActivity bo
 	return nil
 }
 
+// SessionCancelNDE is a regression test workflow for
```
yuandrew: Don't think we need this comment here, this is already explained in the root test that calls this workflow

test/integration_test.go
yuandrew: not going to comment on every test this does it, but should apply to all new tests here

yuandrew (changes requested):

test/integration_test.go
Sushisource: Yes it's when it equals one. Subsequent iterations should succeed.

internal/internal_workflow.go
```diff
@@ -1035,9 +1035,36 @@ func (s *coroutineState) initialYield(stackDepth int, status string) {
...
+		frame, more := frames.Next()
+		if frame.Function == "runtime.gopanic" {
+			return true
+		}
+		if !more {
+			break
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
Sushisource: True but irrelevant (fixed the comment). `recover()` ing inside of deferred code in a Workflow I think we should just state is UB. It makes no sense. If you're panicking from workflow code, it's because something unaccounted for went wrong and that's a workflow task failure. If you're using it as, essentially, flow control... don't do that.

PR #2230, opened 2026-03-15
feat: Allow concurrent payload visiting

### LINKED ISSUE #2223: Enable concurrent payload visiting
Parent issue: https://github.com/temporalio/features/issues/772

brucearctor:
Closes #2223

This PR introduces a `Concurrency` configuration option inside `PayloadCodecGRPCClientInterceptorOptions` and `NewFailureGRPCClientInterceptorOptions`.

When `Concurrency > 1`, the proxy visitor collects payload pointers sequentially but decodes/encodes payloads concurrently using `errgroup`, reducing latency drastically when the codec includes heavy IO (e.g., KMS client-side encryption calls). 
It avoids deep copies by directly modifying the byte vectors in place on the tree.
Fallback execution remains entirely sequential if `Concurrency=1` or `0`.

A unit test asserting the parallel time reduction is included.

### DIFF
--- a/converter/grpc_interceptor.go
+++ b/converter/grpc_interceptor.go
@@ -1,19 +1,53 @@
 package converter
 
 import (
+	"context"
 	"fmt"
+	"slices"
 
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
 
 	commonpb "go.temporal.io/api/common/v1"
 	failurepb "go.temporal.io/api/failure/v1"
 	"go.temporal.io/api/proxy"
 )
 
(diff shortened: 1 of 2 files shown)


jmaeagle99:
Hi @brucearctor. Thanks for taking time to create a PR for this issue and for your contribution.

The intent of https://github.com/temporalio/features/issues/772 was to update https://github.com/temporalio/api-go/blob/master/proxy/interceptor.go#L101C6-L101C19 to enable concurrent visiting so that it is broadly usable. This would allow concurrency regardless of the visitor callback implementation. gRPC interceptors are one type of these callback implementations.

The concurrent visiting was enabled upstream in https://github.com/temporalio/api-go/pull/244 and integrated into the current repository in https://github.com/temporalio/sdk-go/pull/2294. The gRPC interceptors themselves are not specifically aware of the concurrency enablement, but can be made concurrent using the payload visitor concurrency options.

Since the interceptors do not need concurrent visiting explicitly enabled within their implementations and can leverage the generalized concurrent visiting options, we will not accepting this PR in its current form. If you feel there are additional improvements that could be made in this space, please do not hesitate to open an issue for discussion.

brucearctor:
Makes sense.  And I approx the response!

PR #2231, opened 2026-03-15
Expose raw Failure and Payload protos on SDK error and result types

### LINKED ISSUE #2203: Return raw gRPC responses for Update and Activity polling
For Update and Standalone Activities, the SDK converts the raw proto `Outcome` into an `EncodedValue` or `error`. 

Have public functions that return  the raw `Outcome` proto for update and standalone activities.

**Additional context**

brucearctor:
# Objective

Fixes #2203 by exposing the full underlying proto data for errors and results from update/activities.

SDK users generally expect to be able to extract all raw details about failures and payloads when high-level helper methods do not cover all use cases (e.g., retrieving stack traces from the CLI without parsing error strings, or getting the original payload representation without relying on a data converter).

# Changes

* **Error types**: Renamed `failure()` to `Failure()` on the internal `temporalError` type, which is embedded in other public error types such as `ApplicationError`, `TimeoutError`, and `ActivityError`. This allows SDK users to retrieve the underlying `*failurepb.Failure` object directly. Adjusted the internal `FailureHolder` back to `failureHolder` to avoid method/field conflicts on `NexusOperationError`.
* **Result types**: Added an optional `ValuesPayloads` interface to `converter` and a `GetPayloads` helper. 
  * Why not add `Payloads() *commonpb.Payloads` directly to the `EncodedValue` and `EncodedValues` interfaces? Doing so would be a backwards-incompatible breaking change for any users outside of `sdk-go` that have implemented or mocked these interfaces. 
  * With `GetPayloads(encoded interface{}) *commonpb.Payloads`, existing implementations won't break, but Temporal's internal `EncodedValue` implementations now satisfy the new interface, allowing extraction of the raw `*commonpb.Payloads`.
* Updated the generated testing mocks.


internal/error.go
```diff
@@ -566,6 +566,10 @@ func (e *temporalError) failure() *failurepb.Failure {
 	return e.originalFailure
 }
 
+func (e *temporalError) Failure() *failurepb.Failure {
```
yuandrew, line 597: I would add some additional information on how to use this new field, something like 

```suggestion
  // Failure returns the original proto Failure this error was created from, if one is available.
  //
  // This is intended for advanced callers that need structured failure data such as stack traces,
  // encoded details, or the full cause chain. Application code should generally
  // continue using errors.As with concrete SDK error types such as *ApplicationError.
  //
  // When working with an arbitrary error value, use errors.As with an interface:
  //
  //	type failureProvider interface {
  //		Failure() *failurepb.Failure
  //	}
  //
  //	var fp failureProvider
  //	if errors.As(err, &fp) {
  //		failure := fp.Failure()
  //	}
func (e *temporalError) Failure() *failurepb.Failure {
```

converter/value.go
```diff
@@ -57,3 +57,22 @@ func (v RawValue) MarshalJSON() ([]byte, error) {
 func (v *RawValue) UnmarshalJSON(b []byte) error {
 	return fmt.Errorf("RawValue is not JSON serializable")
 }
+
+// ValuesPayloads is an optional interface that EncodedValue and EncodedValues may implement
+// to expose the underlying commonpb.Payloads without decoding.
+type ValuesPayloads interface {
+	// Payloads gets the underlying commonpb.Payloads
```
yuandrew: I think it's worth adding "without decoding" here, as that's the difference between this API and the `Get(valuePtr interface{}) error` method from the `EncodedValue`/`EncodedValues` interfaces

```suggestion
	// Payloads gets the underlying commonpb.Payloads without decoding
```

yuandrew (commented):
Overall change looks good, I mainly would like some additional tests that verify this adds the value we're looking for:
      - PollWorkflowUpdate success result supports converter.GetPayloads.
      - PollWorkflowUpdate failure error exposes Failure() with stack trace/details.
      - PollActivityResult success/failure does the same.

yuandrew (commented):
Thanks! This is much closer. I’d still like at least one test per real polling surface, because the CLI depends on the actual PollWorkflowUpdate / PollActivityResult outputs, where we're not just newEncodedValue/FailureToError being called to simulate the behavior. 

This could even just be extending existing tests and asserting the new outputs

brucearctor:
@yuandrew : should be good to go now

internal/internal_workflow_client_test.go
```diff
@@ -2713,6 +2714,146 @@ func TestUpdate(t *testing.T) {
...
+		// Use PollWorkflowUpdate directly to access the raw output.
+		output, err := client.PollWorkflowUpdate(
+			context.TODO(),
+			refFromRequest(req),
+		)
+		require.NoError(t, err)
+		require.NotNil(t, output.Result)
+		payloads := converter.GetPayloads(output.Result)
+		require.NotNil(t, payloads)
+		require.Len(t, payloads.GetPayloads(), 1)
```
yuandrew, line 2744: We assert length, but we should also assert this matches `outPayloads`
```suggestion
                require.Equal(t, outPayloads, payloads)
		require.Len(t, payloads.GetPayloads(), 1)
```

yuandrew (approved):

yuandrew:
Added a single missing assert line for one of the tests, to avoid another round of back and forth. 

Thanks for the contribution! 🥳

PR #2245, opened 2026-03-19
Serverless Lambda support

Sushisource:
# Usage
I intend to publish a sample that is meant to be used in conjunction with this. Although the wrapper is easy to use there is, in typical AWS fashion, a bunch of random ancillary crap you need to do to get it to work that was not at all obvious to someone like me who doesn't live and breathe AWS. So, that should help significantly with getting people started.

# Changes summary

- Adds a new contrib/aws/serverlesslambdaworker package that provides an ergonomic
  wrapper for running Temporal workers inside AWS Lambda
- RunWorker manages the full per-invocation lifecycle: dials Temporal, starts a
  worker with Lambda-appropriate defaults, polls until the invocation deadline
  approaches, then gracefully shuts down
- Derives worker identity from the Lambda request context (function name + request
  ID)
- Auto-discovers client connection config (address, namespace, TLS, API key) via
  envconfig, with Lambda-aware config file resolution (TEMPORAL_CONFIG_FILE >
  $LAMBDA_TASK_ROOT/temporal.toml > ./temporal.toml)
- Includes an otel sub-package with ready-made OpenTelemetry/ADOT integration for
  metrics and tracing
- Comprehensive test coverage for configuration, defaults, and the worker run
  lifecycle

### DIFF
--- a/contrib/aws/lambdaworker/README.md
+++ b/contrib/aws/lambdaworker/README.md
@@ -10,17 +10,21 @@ invocation deadline.
 ```go
 package main
 
-import "go.temporal.io/sdk/contrib/aws/lambdaworker"
+import (
+    "go.temporal.io/sdk/contrib/aws/lambdaworker"
+    "go.temporal.io/sdk/worker"
+)
 
 func main() {
-    lambdaworker.RunWorker(
-        func(ctx *lambdaworker.ConfigureWorkerContext) error {
-            ctx.SetTaskQueue("my-task-queue")
-            ctx.RegisterWorkflow(MyWorkflow)
-            ctx.RegisterActivity(MyActivity)
-            return nil
-        },
-    )
+    lambdaworker.RunWorker(worker.WorkerDeploymentVersion{
+        DeploymentName: "my-service",
+        BuildID:        "v1.0",
+    }, func(ctx *lambdaworker.Options) error {
+        ctx.TaskQueue = "my-task-queue"
+        ctx.RegisterWorkflow(MyWorkflow)
+        ctx.RegisterActivity(MyActivity)
+        return nil
+    })
 }
 ```
 
@@ -36,9 +40,9 @@ automatically from a TOML config file and/or environment variables via
 
 The file is optional — if absent, only environment variables are used.
 
-Use `MutateClientOptions` and `MutateWorkerOptions` on the
-`ConfigureWorkerContext` to override any defaults programmatically. User
-overrides are applied last, so they always win.
+Set the public fields on `Options` (e.g. `ClientOptions`, `WorkerOptions`) in
+the configure callback to override any defaults programmatically. User overrides
+are applied last, so they always win.
 
 ## Lambda-tuned worker defaults
 
@@ -51,14 +55,9 @@ Metrics and tracing are opt-in. The `otel` sub-package provides convenience
 helpers for AWS Distro for OpenTelemetry (ADOT):
 
 ```go
-import (
-    "go.temporal.io/sdk/client"
-    "go.temporal.io/sdk/contrib/aws/lambdaworker/otel"
-)
+import "go.temporal.io/sdk/contrib/aws/lambdaworker/otel"
 
-ctx.MutateClientOptions(func(opts *client.Options) error {
-    return otel.ApplyDefaults(ctx, opts, otel.Options{})
-})
+otel.ApplyDefaults(ctx, &ctx.ClientOptions, otel.Options{})
 ```
 
 You can also use `otel.ApplyMetrics` or `otel.ApplyTracing` individually.
--- a/contrib/aws/lambdaworker/doc.go
+++ b/contrib/aws/lambdaworker/doc.go
@@ -53,8 +53,8 @@
 //
 // # Observability
 //
-// Observability (metrics, tracing) is opt-in. Use [Options.MutateClientOptions] to
-// supply your own metrics handler and tracing interceptor. A convenience sub-package at
+// Observability (metrics, tracing) is opt-in. Set [Options.ClientOptions] fields directly
+// or use the helpers in the convenience sub-package at
 // contrib/aws/lambdaworker/otel/ provides ready-made OTel configuration for AWS Distro
 // for OpenTelemetry (ADOT).
 package lambdaworker
--- a/contrib/aws/lambdaworker/otel/otel.go
+++ b/contrib/aws/lambdaworker/otel/otel.go
@@ -28,7 +28,7 @@ import (
 )
 
 // ShutdownRegistrar accepts a function to be called at the end of each Lambda invocation.
-// [lambdaworker.ConfigureWorkerContext] implements this interface.
+// [lambdaworker.Options] implements this interface.
 type ShutdownRegistrar interface {
 	OnShutdown(func(context.Context) error)
 }
--- a/contrib/aws/lambdaworker/options.go
+++ b/contrib/aws/lambdaworker/options.go
@@ -12,7 +12,7 @@ import (
 	"go.temporal.io/sdk/workflow"
 )
 
-// Compile-time check that ConfigureWorkerContext implements worker.Registry.
+// Compile-time check that Options implements worker.Registry.
 var _ worker.Registry = (*Options)(nil)
 
 // Options is passed to the configure callback of [RunWorker]. It implements [worker.Registry] so


contrib/aws/serverlesslambdaworker/doc.go
```diff
@@ -0,0 +1,55 @@
...
+// [RunWorker] handles the Lambda lifecycle: on each invocation it dials the Temporal server,
+// creates and starts a worker with Lambda-appropriate defaults, polls for tasks until the
+// invocation deadline approaches, and then gracefully shuts down the worker and closes the client.
+// This per-invocation lifecycle ensures a clean worker state on every Lambda invocation.
+//
+// # Configuration
+//
+// Client connection options (address, namespace, TLS, API key) are loaded automatically from
+// a TOML config file and environment variables via
+// [go.temporal.io/sdk/contrib/envconfig.LoadClientOptions].
```
02strich: That reference is kinda hard to understand as there is no readme or the like explaining what the environment variables are.

contrib/aws/lambdaworker/doc.go
```diff
@@ -0,0 +1,55 @@
...
+// Client connection options (address, namespace, TLS, API key) are loaded automatically from
+// a TOML config file and environment variables via
+// [go.temporal.io/sdk/contrib/envconfig.LoadClientOptions].
+//
+// Because AWS Lambda does not have a standard config file directory, it resolves the config file
+// path in the following order:
+//
+//  1. TEMPORAL_CONFIG_FILE environment variable, if set.
+//  2. temporal.toml in the Lambda code root ($LAMBDA_TASK_ROOT), which is typically /var/task.
+//  3. temporal.toml in the current working directory as a final fallback.
```
02strich, line 38: would these ever be different?

contrib/aws/lambdaworker/doc.go
```diff
@@ -0,0 +1,55 @@
...
+//	        return nil
+//	    })
+//	}
+//
+// [RunWorker] handles the Lambda lifecycle: on each invocation it dials the Temporal server,
+// creates and starts a worker with Lambda-appropriate defaults, polls for tasks until the
+// invocation deadline approaches, and then gracefully shuts down the worker and closes the client.
+// This per-invocation lifecycle ensures a clean worker state on every Lambda invocation.
+//
+// # Configuration
```
02strich, line 25: after reading this I am still unclear on how to reference my credentials. Given that neither API key nor TLS client key should be in an environment variable, the mentioned config approaches do not quite match.

General recommendation would be to read these secrets from Secret Manager. Do we already have helpers for that anywhere?

contrib/aws/lambdaworker/runworker.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	var workerOpts worker.Options
+	applyLambdaWorkerDefaults(&workerOpts)
+	if configCtx.mutateWorkerOpts != nil {
+		if err := configCtx.mutateWorkerOpts(&workerOpts); err != nil {
+			return fmt.Errorf("mutating worker options: %w", err)
+		}
+	}
+	deps.setCacheSize(defaultStickyCacheSize)
```
02strich, line 101: does this override settings from the `mutate*` calls?

Sushisource: Yeah good catch that should be higher

contrib/aws/lambdaworker/doc.go
Sushisource: We do not, to my knowledge, but that's a good point we should call out in the docstring

contrib/aws/lambdaworker/doc.go
Sushisource: Probably not unless you've configured a weird custom container layout, but I think it's still a reasonable strategy

contrib/aws/serverlesslambdaworker/doc.go
Sushisource: I'll include a link to our docs site about it

contrib/aws/lambdaworker/doc.go
02strich: ah you mean for container-backed Lambda's :)

Yeah the structure works, was more curious why you made the distinction.

contrib/aws/lambdaworker/doc.go
02strich: We should see about making this nicer as part of this contrib (can be another PR of cause)

contrib/aws/serverlesslambdaworker/go.mod
```diff
@@ -0,0 +1,43 @@
+module go.temporal.io/sdk/contrib/aws/serverlesslambdaworker
```
jaypipes: I'd recommend naming this module `go.temporal.io/sdk/contrib/aws/lambda/worker`. serverless is redundant with Lambda :)

contrib/aws/serverlesslambdaworker/configure.go
```diff
@@ -0,0 +1,129 @@
...
+	"go.temporal.io/sdk/workflow"
+)
+
+// Compile-time check that ConfigureWorkerContext implements worker.Registry.
+var _ worker.Registry = (*ConfigureWorkerContext)(nil)
+
+// ConfigureWorkerContext is passed to the configure callback of [RunWorker]. It implements
+// [worker.Registry] so that workflows, activities, and Nexus services can be registered directly on
+// it.
+type ConfigureWorkerContext struct {
```
jaypipes: other structs in the contrib modules and elsewhere in sdk-go are named `Options`. I think it would be easier to align with other places in sdk-go to rename this file `options.go` and just make the struct's fields exported:

```go
type Options struct {
	ClientOptions *client.Options
	WorkerOptions *worker.Options
	TaskQueue string
	ShutdownDeadlineBuffer time.Duration
}
```

and then validate/set defaults in the supplied `Options` that are passed to a `New()` (or `NewWorker()` function:

(I'd recommend renaming `runworker.go` to `worker.go`)

```go
// lambdaWorker wraps a normal temporal Worker and
// validates that the Lambda-specific configuration is valid.
type lambdaWorker struct {
	worker.Worker
}

// Handler returns the functor supplied to lambda.Start() that will
// start the lambdaWorker.
func (w *lambdaWorker) Handler()  func(context.Context) error {
	return func(ctx context.Context) error {
		deadline, ok := ctx.Deadline()
		if ok {
			workTime := time.Until(deadline) - shutdownBuffer
			if workTime <= time.Second {
				return fmt.Errorf(
					"Lambda timeout leaves almost no time for work "+
						"(workTime %v, shutdownBuffer %v); increase the "+
						"function timeout or decrease the shutdown deadline "+
						"buffer",
					workTime, shutdownBuffer,
				)
			} else if workTime < 5*time.Second {
				if logger != nil {
					logger.Warn(
						"Lambda timeout leaves less than 5s for work after "+
							"shutdown buffer; consider increasing the function "+
							"timeout or decreasing the shutdown deadline buffer",
						"workTime", workTime,
						"shutdownBuffer", shutdownBuffer,
					)
				}
			}
		}

		if err := w.Start(); err != nil {
			return fmt.Errorf("error starting worker: %w", err)
		}

		if !ok {
			<-ctx.Done()
		} else {
			shutdownAt := deadline.Add(-shutdownBuffer)
			waitDuration := time.Until(shutdownAt)
			if waitDuration > 0 {
				timer := time.NewTimer(waitDuration)
				defer timer.Stop()
				select {
				case <-timer.C:
				case <-ctx.Done():
				}
			}
		}

		// Stop the worker before running shutdown hooks so that hooks (e.g. OTLP telemetry flushes)
		// see all spans/metrics emitted during the drain phase.
		w.Stop()

		// Use context.Background because invocationCtx may already be cancelled. No timeout is
		// needed — Lambda hard-kills the process at the deadline regardless.
		for _, fn := range configCtx.shutdownFuncs {
			if err := fn(context.Background()); err != nil {
				fmt.Fprintf(os.Stderr,
					"serverlesslambdaworker: shutdown hook error: %v\n", err)
			}
		}
	}
}

// New returns a new specialized Worker ready to be
// started in a lambda StartHandler.
func New(options Options) (worker.Worker, error) {
	// validate the supplied options after merging with any defaults from
	// the environment...
	defaultOptionsFromEnvironment(&options)
	if err := validateOptions(options); err != nil {
		return nil, err
	}
	clientOpts := envconfig.LoadClientOptions(
		envconfig.LoadClientOptionsRequest{
			ConfigFilePath: lambdaDefaultConfigFilePath(os.Getenv),
		},
	)
	// adjust the client opts as needed for the Lambda env.
	applyLambdaClientDefaults(&clientOpts)
	c, err := client.Dial(clientOpts)
	if err != nil {
		return nil, err
	}
	w, err := worker.New(c, options.TaskQueue, options.WorkerOptions)
	if err != nil {
		return nil, err
	}
	return &lambdaWorker{Worker: w}, nil
}
```

This would match the pattern used in most of the other sdk-go contrib modules like the [opentelemetry](https://github.com/temporalio/sdk-go/blob/master/contrib/opentelemetry/tracing_interceptor.go) and [opentracing](https://github.com/temporalio/sdk-go/blob/master/contrib/opentracing/interceptor.go) modules.

It would also mean you wouldn't need to reimplement all the RegisterXXX methods below.

The main.go example would then look almost exactly like a "normal" Temporal worker but with the expected lambda Start handler:

```go
package main

import (
	"github.com/aws/aws-lambda-go/lambda"
	"github.com/aws/aws-lambda-go/lambdacontext"
    lambdaworker "go.temporal.io/sdk/contrib/aws/lambda/worker"
)

var (
    w worker.Worker
)

func init() {
    var err error
    opts := lambdaworker.Options{
        TaskQueue: "my-task-queue",
    }
    w, err = lambdaworker.New(opts)
    if err != nil {
        fmt.Fprintf(
            os.Stderr,
            "failed to create lambda worker in execution environment init: %s",
            err.Error(),
        )
    } else {
        w.RegisterWorkflow(MyWorkflow)
        w.RegisterActivity(MyActivity)
    }
}

func main() {
    if w != nil {
        lambda.Start(w.Handler())
    }
}
```

contrib/aws/lambdaworker/README.md
```diff
@@ -0,0 +1,71 @@
+# serverlesslambdaworker
+
+A wrapper for running [Temporal](https://temporal.io) workers inside AWS Lambda. A single
+`RunWorker` call handles the full per-invocation lifecycle: dialing the Temporal server, creating a
+worker with Lambda-tuned defaults, polling for tasks, and gracefully shutting down before the
+invocation deadline.
+
+## Quick start
```
yuandrew, line 8: You already mentioned writing a sample in the PR desc, could be nice to link it here (you were probably already going to do that, but just in case)

contrib/aws/serverlesslambdaworker/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+You can also use `otel.ApplyMetrics` or `otel.ApplyTracing` individually.
+
+If you use OTEL, you can use
+[ADOT](https://aws-otel.github.io/docs/getting-started/lambda/lambda-go) (the AWS Distro For
+OpenTelemetry) to automatically integrate with AWS observability functionality. Namely, you will
+want to add the Lambda layer in the aforementioned link. We'll handle setting up the SDK for you.
+
+See more [here](https://aws-observability.github.io/aws-otel-collector/?utm_source=chatgpt.com)
```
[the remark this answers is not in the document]
yuandrew: agree here, seems strange to keep the `?utm_source=chatgpt.com` part of the link. Any reason for keeping this?

contrib/aws/serverlesslambdaworker/go.mod
Sushisource: Good point, was considering that at one point. I'm not sure the / in `lambda/worker` is worth much, though?

contrib/aws/serverlesslambdaworker/configure.go
Sushisource: Yeah, good point on the naming. I was keeping the fields private for flexibility reasons, but, the reality is they're so one-to-one with stuff you _have_ to be able to change that that probably doesn't make much sense. Agreed.

I dunno how I feel about flipping the API around though. I can see the appeal... I'll get some input from other folks on SDK team about it.

contrib/aws/lambdaworker/otel/go.mod
```diff
@@ -0,0 +1,54 @@
+module go.temporal.io/sdk/contrib/aws/serverlesslambdaworker/otel
+
+go 1.25.0
```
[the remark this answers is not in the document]
yuandrew: Can you help me understand this? I saw you thumbsed it up, but it's still at 1.25 when the parent dir has 1.23, and seems like this doesn't have to be 1.25?

contrib/aws/serverlesslambdaworker/configure_test.go
```diff
@@ -0,0 +1,143 @@
...
+		<-interruptCh
+	}
+	return args.Error(0)
+}
+
+func (m *mockWorker) Stop() {
+	m.Called()
+}
+
+func TestConfigureWorkerContext_ImplementsRegistry(t *testing.T) {
```
yuandrew: Don't think this adds any value over the compile-time check itself in `configure.go`

contrib/aws/serverlesslambdaworker/go.mod
yuandrew: I actually like `lambdaworker` more, mainly because it prevents users from having to alias this import, in case they're using this new package and `go.temporalio/sdk/worker`

contrib/aws/lambdaworker/go.mod
```diff
@@ -0,0 +1,43 @@
...
+	golang.org/x/text v0.24.0 // indirect
+	golang.org/x/time v0.3.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20240827150818-7e3bb234dfed // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20240827150818-7e3bb234dfed // indirect
+	google.golang.org/grpc v1.67.1 // indirect
+	google.golang.org/protobuf v1.36.6 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace (
```
yuandrew, line 40: Reminder to remove these before merging, and setting `go.temporal.io/sdk/contrib/envconfig` to a real version

contrib/aws/lambdaworker/otel/go.mod
```diff
@@ -0,0 +1,54 @@
...
+	golang.org/x/text v0.34.0 // indirect
+	golang.org/x/time v0.3.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20260209200024-4cfbd4190f57 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20260209200024-4cfbd4190f57 // indirect
+	google.golang.org/grpc v1.79.2 // indirect
+	google.golang.org/protobuf v1.36.11 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace (
```
yuandrew, line 51: Same here

contrib/aws/serverlesslambdaworker/otel/otel.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	if options.ServiceName == "" {
+		options.ServiceName = "temporal-lambda-worker"
+	}
+	if options.CollectorEndpoint == "" {
+		options.CollectorEndpoint = os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+	}
+
+	var grpcOpts []otlpmetricgrpc.Option
+	grpcOpts = append(grpcOpts, otlpmetricgrpc.WithInsecure())
```
yuandrew: Do we always want to be using `WithInsecure` here? My understanding is this is fine for ADOT, but for other collectors might need TLS support. Given this sub-package is named a general `otel` and not ADOT specific (even though our docs mention these are convenience functions for ADOT), should we support passing TLS or making this configurable?

contrib/aws/lambdaworker/otel/otel.go
```diff
@@ -0,0 +1,194 @@
+// Package otel provides convenience helpers for configuring OpenTelemetry
+// metrics and tracing on a Temporal client running inside AWS Lambda.
+//
+// Use [ApplyDefaults] inside a [serverlesslambdaworker.ConfigureWorkerContext.MutateClientOptions]
+// callback for a batteries-included setup that creates OTLP gRPC exporters and an AWS X-Ray ID
+// generator, suitable for use with the AWS Distro for OpenTelemetry (ADOT) Lambda layer.
+//
+// Use [ApplyDefaultsWithProviders] if you need to supply your own MeterProvider and TracerProvider.
+package otel
```
yuandrew, line 9: Do we want any tests for this file/package?

contrib/aws/lambdaworker/otel/otel.go
```diff
@@ -0,0 +1,194 @@
...
+	traceGrpcOpts = append(traceGrpcOpts, otlptracegrpc.WithInsecure())
+	if options.CollectorEndpoint != "" {
+		grpcOpts = append(grpcOpts, otlpmetricgrpc.WithEndpoint(options.CollectorEndpoint))
+		traceGrpcOpts = append(traceGrpcOpts, otlptracegrpc.WithEndpoint(options.CollectorEndpoint))
+	}
+
+	// Build a shared resource for both providers so that metrics and traces
+	// carry the same service.name, enabling correlation in backends.
+	res, err := resource.Merge(
+		resource.Default(),
```
yuandrew, line 110: Claude warns this can produce high metrics cardinality, since `host.name` returns the sandbox hostname. Which Lambda recycles for warm starts, but not for cold. Should we be concerned with metrics cardinality?

contrib/aws/serverlesslambdaworker/runworker.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+// RunWorker starts a Temporal worker inside an AWS Lambda execution environment. It calls the
+// configure callback to collect registrations and option overrides, then delegates to the Lambda
+// runtime. On each invocation, it dials the Temporal server, starts a worker, polls for tasks
+// until the invocation deadline approaches, and then gracefully shuts down the worker and closes
+// the client. RunWorker does not return under normal operation. On fatal configuration error, it
+// logs to stderr and calls os.Exit(1).
+func RunWorker(configure func(ctx *ConfigureWorkerContext) error) {
```
yuandrew: Since this is the entry point to this all, do we want to make it abundantly clear what the minimum should be for this callback? From what I read, Task Queue and then registering a Workflow or Activity is the bare minimum. Could be helpful for new users only looking to use Lambdas or even AI agents?

contrib/aws/serverlesslambdaworker/configure.go
yuandrew: I'm a fan of how the current design feels simpler for users. Mirroring the Worker API would be familiar/consistent, but I think the simplification for lambda usage can be worth it for the deviation.

contrib/aws/lambdaworker/otel/go.mod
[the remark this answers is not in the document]
Sushisource: It's because go mod tidy automatically changes it, because `go.opentelemetry.io/contrib/propagators/aws v1.42.0` requires 1.25. I could see about rolling back to an older version but it's 7+ months old to get 1.23 support.

contrib/aws/serverlesslambdaworker/otel/otel.go
Sushisource: Yeah, good point. Will pass through.

contrib/aws/lambdaworker/otel/otel.go
Sushisource: Not a whole lot interesting to test I think, would really just be re-stating the defaults

contrib/aws/lambdaworker/otel/otel.go
Sushisource: I don't think this actually does include the host name?

contrib/aws/serverlesslambdaworker/runworker.go
Sushisource: Yeah makes sense

contrib/aws/serverlesslambdaworker/configure.go
jaypipes: > I'm a fan of how the current design feels simpler for users. Mirroring the Worker API would be familiar/consistent, but I think the simplification for lambda usage can be worth it for the deviation.

@yuandrew Note that I'm using the `init()` function to handle the per-Lambda-execution environment setup and not the `main()` function. That's a big difference between @Sushisource's original proposed code as well.

contrib/aws/lambdaworker/go.mod
yuandrew: oops, didn't realize other contrib packages do this, disregard

contrib/aws/serverlesslambdaworker/go.mod
Sushisource: Changed to lambdaworker

contrib/aws/serverlesslambdaworker/configure.go
Sushisource: I updated the Options struct per your suggestions, but I think we'll stick with this flow for now. Appreciate the feedback

contrib/aws/lambdaworker/README.md
Sushisource: Will do later, since we'll need to merge the sample separately

CLAUDE.md
```diff
@@ -0,0 +1,3 @@
+# Important requirements
```
yuandrew: Can we avoid adding this file for now? If we want to add one, I think it should be in its own PR

Sushisource: Yes, didn't mean to add, thanks.

PR #2254, opened 2026-03-25
Add poller group ID tracking across all poller types

tconley1428:
## Summary
- Add `pollerGroupTracker` that distributes poll requests across server-provided poller groups using weighted random selection, prioritizing groups with zero pending polls
- Integrate tracker into workflow, activity, and nexus task pollers: set `PollerGroupId` on poll requests and update groups from poll responses
- Forward `PollerGroupId` from poll responses to `RespondQueryTaskCompleted`, `RespondNexusTaskCompleted`, and `RespondNexusTaskFailed` for proper routing

Depends on API changes from https://github.com/temporalio/api/pull/744

## Test plan
- [x] Unit tests for `pollerGroupTracker` (zero-pending prioritization, weighted distribution, release, group updates)
- [x] Unit tests for `weightedRandom` (single candidate, zero weights, convergence)
- [x] Integration tests verifying `PollerGroupId` is empty on first poll, set on subsequent polls for all three poller types
- [x] Integration tests verifying `PollerGroupId` forwarding on query responses (success and error paths)
- [x] Integration tests verifying `PollerGroupId` forwarding on nexus completion and failure (via full `Execute` path)
- [x] Integration test for GRPC-too-large query response fallback path


internal/internal_task_pollers.go
```diff
@@ -1193,14 +1197,19 @@ func (wtp *workflowTaskPoller) poll(ctx context.Context) (taskForWorker, error)
 		wtp.logger.Debug("workflowTaskPoller::Poll")
 	})
 
+	groupId := wtp.pollerGroupTracker.getNextGroupId()
+	defer wtp.pollerGroupTracker.release(groupId)
```
Sushisource: Any particular reason to do this before `getNextPollRequest()` ?

internal/poller_group_tracker.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// pollerGroupTracker distributes pollers across server-provided poller groups
+// based on weights. Each call to getNextGroupId returns a group ID such that
+// every group has at least one pending (unreleased) request, and beyond that
+// minimum the distribution follows group weights.
+type pollerGroupTracker struct {
+	mu      sync.Mutex
+	groups  []*taskqueuepb.PollerGroupInfo
+	pending map[string]int // number of unreleased requests per group ID
```
Sushisource: IMO it'd make sense to combine these fields into one map that looks like `id -> (weight, pending)`

So it's not possible to have inconsistent views

internal/poller_group_tracker.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	chosen := weightedRandom(candidates)
+	t.pending[chosen]++
+	return chosen
+}
+
+// weightedRandom selects a group ID from candidates randomly based on weights.
+// candidates must be non-empty.
+func weightedRandom(candidates []*taskqueuepb.PollerGroupInfo) string {
```
Sushisource: Classic Go minimalism

Meanwhile https://docs.rs/rand/latest/rand/distr/weighted/struct.WeightedIndex.html

Sushisource (commented):
Generally makes sense, only one structural change I think could simplify things slightly

internal/internal_task_pollers.go
tconley1428: I don't think so.

internal/poller_group_tracker.go
tconley1428: Why give tools when you can have to write it yourself?

Sushisource (approved):

internal/internal_task_pollers.go
```diff
@@ -1141,6 +1145,9 @@ func (wtp *workflowTaskPoller) getNextPollRequest() (request *workflowservice.Po
 		panic("unknown workflow task poller mode")
 	}
 
+	groupId := wtp.pollerGroupTracker.getNextGroupId()
+	defer wtp.pollerGroupTracker.release(groupId)
```
ShahabT: `release` should be called after the poll returns (e.g. in `(wtp *workflowTaskPoller) release`) otherwise tracker cannot track _pending_ polls.

internal/internal_task_pollers.go
```diff
@@ -430,6 +432,7 @@ func (wtp *workflowTaskProcessor) createPoller(mode workflowTaskPollerMode) task
 		eagerActivityExecutor:        wtp.eagerActivityExecutor,
 		numNormalPollerMetric:        wtp.numNormalPollerMetric,
 		numStickyPollerMetric:        wtp.numStickyPollerMetric,
+		pollerGroupTracker:           newPollerGroupTracker(),
```
ShahabT, line 437: Mixed mode needs two separate trackers (for sticky and normal). This is because the polls go to different queues and each need to ensure proper balancing among all poller groups.

ShahabT (changes requested):

internal/internal_task_pollers.go
```diff
@@ -1190,16 +1190,24 @@ func (wtp *workflowTaskPoller) poll(ctx context.Context) (taskForWorker, error)
...
 	})
-
 	request := wtp.getNextPollRequest()
+
+	groupTracker := wtp.pollerGroupTracker
+	if request.TaskQueue.Kind == enumspb.TASK_QUEUE_KIND_STICKY {
+		groupTracker = wtp.stickyPollerGroupTracker
+	}
+	groupId := groupTracker.getNextGroupId()
+	defer groupTracker.release(groupId)
```
ShahabT, line 1200: it feels safer if we more this to `wtp.release` and make it happen inside the `requestLock` block.

internal/internal_task_pollers.go
```diff
@@ -1190,16 +1190,24 @@ func (wtp *workflowTaskPoller) poll(ctx context.Context) (taskForWorker, error)
...
 		wtp.logger.Debug("workflowTaskPoller::Poll")
 	})
-
 	request := wtp.getNextPollRequest()
+
+	groupTracker := wtp.pollerGroupTracker
+	if request.TaskQueue.Kind == enumspb.TASK_QUEUE_KIND_STICKY {
+		groupTracker = wtp.stickyPollerGroupTracker
+	}
+	groupId := groupTracker.getNextGroupId()
```
ShahabT, line 1199: it feels safer if we more this to wtp.getNextPollRequest and make it happen inside the requestLock block. Ideally the tracker state and the counters (`pendingStickyPollCount/pendingReqularPollCount`) should be in sync.

We can even go further and get rid of the counters and ask the group tracker to track overall counts as well.

internal/internal_worker_base.go
```diff
@@ -737,6 +840,30 @@ func newPollScalerReportHandle(options pollScalerReportHandleOptions) *pollScale
 	return psr
 }
 
+func (tw *scalableTaskPoller) ensurePollerGroupCapacity(groupCount int) {
```
ShahabT, line 843: 1. Does this also work for poller auto-tuning turned off?
2. Also in that case (Mixed mode) how are we ensuring at least `groupCount` pollers for sticky and normal separately?

internal/internal_worker_base.go
```diff
@@ -830,6 +840,30 @@ func newPollScalerReportHandle(options pollScalerReportHandleOptions) *pollScale
...
```
ShahabT, line 843: what about the case when the number of groups increases after worker start?

yuandrew:
This is replaced by https://github.com/temporalio/sdk-go/pull/2422

PR #2255, opened 2026-03-27
:boom: Experimental: Payload limit validation

### LINKED ISSUE #2165: SDK should fail workflow task if payloads size it known to be too large - Go

### LINKED ISSUE #2167: Warn if the SDK tried to send a payload above a specific size - Go

jmaeagle99:
## What was changed

- 💥 Update worker to get payload error limits from Temporal server and enforce those limits by default. The SDK will now eagerly fail a task that has too large payloads instead of uploading them to the server and having the server fail the workflow. If the server doesn't report error limits, error limits are not enforced in the worker.
- Add client payload limits options for setting payload size warning thresholds. Warnings are emitted in client, worker, and replayer for any payloads that are above these thresholds.

Note that memo specific limits have been left out of this PR as that likely requires changes in temporalio/api-go to the payload visitor. Will address separately.

This PR is a redo of https://github.com/temporalio/sdk-go/pull/2175 based on the payload visitor.

### 💥 Behavioral change details

Within workers, if a payload exceeds the server limits, the worker will eagerly fail the current task instead of uploading the object with the too large payload. This allows the task to be retried instead of entirely failing the workflow from within the server.

Customers who use gRPC proxies that alter payloads before they are passed to the server (encryption in the proxy, offloading to external storage within proxy) should disable this new behavior on the worker using the new `DisablePayloadErrorLimit` option on the worker. For example:

```go
c, err := client.Dial(...)

w := worker.New(c, "queue-name", worker.Options{
    DisablePayloadErrorLimit: true,
})
```

### Examples

Workflow attempts to start an activity with too large of an input:

```
2026/03/27 13:49:54 WARN  Workflow task postprocess error: [TMPRL1103] Attempted to upload payloads with size that exceeded the error limit. Namespace payload-limits-namespace TaskQueue tq-f8cd2c56-bade-426a-8caa-b3c408b3d2ce-TestPayloadLimitsTestSuite/TestPayloadSizeErrorMultipleArguments WorkerID 24502@Justins-MacBook-Pro.local@ WorkflowType payload-size-error-multi-arg WorkflowID TestPayloadLimitsTestSuite/TestPayloadSizeErrorMultipleArguments RunID 019d310f-8672-7d46-91b2-cfece82893cf Attempt 2 PayloadSize 12298 PayloadSizeLimit 10240
```

Activity attempts to complete with too large of a result:

```
2026/03/27 13:48:26 ERROR Activity task postprocess error: [TMPRL1103] Attempted to upload payloads with size that exceeded the error limit. Namespace payload-limits-namespace TaskQueue tq-892c48d1-10cf-4251-a2ba-b07aae4fb20c-TestPayloadLimitsTestSuite/TestPayloadSizeErrorActivityResult WorkerID 23903@Justins-MacBook-Pro.local@ WorkflowID TestPayloadLimitsTestSuite/TestPayloadSizeErrorActivityResult RunID 019d310e-2bd6-788b-8b90-69a8ca840899 ActivityType large-payload-activity Attempt 2 PayloadSize 11269 PayloadSizeLimit 10240
```

Workflow starts an activity with input size over warning limit but under error limit:

```
2026/03/27 13:51:34 WARN  [TMPRL1103] Attempted to upload payloads with size that exceeded the warning limit. Namespace payload-limits-namespace TaskQueue tq-dc85b48e-4c4c-45aa-b970-6c3d36e63bf6-TestPayloadLimitsTestSuite/TestPayloadSizeWarningActivityInput WorkerID 25216@Justins-MacBook-Pro.local@ PayloadSize 1053 PayloadSizeLimit 512
```

## Why?

Users need to know when payload sizes are approaching or have exceeded size limits. This will help prevent workflow outages and inform users to adjust their workflows to make use of alternate storage methods or to break down their payloads more granularly.

## Checklist

1. Closes #2165
2. Closes #2167
3. How was this tested: Unit and integration tests
4. Any docs updates needed? Yes


client/client.go
```diff
@@ -222,6 +222,9 @@ type (
 	// Credentials are optional credentials that can be specified in ClientOptions.
 	Credentials = internal.Credentials
 
+	// PayloadLimitOptions are optional payload size limits that can be specified in ClientOptions.
```
yuandrew, line 225: Looks like you added `NOTE: Experimental` to the internal type, but missed this one 😀

internal/payload_limits.go
```diff
@@ -0,0 +1,90 @@
...
+	"go.temporal.io/sdk/log"
+)
+
+// Options for when payload sizes exceed limits.
+//
+// NOTE: Experimental
+//
+// Exposed as: [go.temporal.io/sdk/client.PayloadLimitOptions]
+type PayloadLimitOptions struct {
+	// The limit (in bytes) at which a payload size warning is logged.
```
yuandrew, line 24: Worth mentioning what that unspecified/`0` defaults to 512 KiB?

test/payload_limits_test.go
```diff
@@ -0,0 +1,973 @@
...
+	ConfigAndClientSuiteBase
+	server *testsuite.DevServer
+	worker worker.Worker
+}
+
+func TestPayloadLimitsTestSuite(t *testing.T) {
+	suite.Run(t, new(PayloadLimitsTestSuite))
+}
+
+const PAYLOAD_SIZE_ERROR_LIMIT = 10 * 1024  // 10 KiB
```
yuandrew: consts are CamelCase for go

test/payload_limits_test.go
```diff
@@ -0,0 +1,973 @@
...
+	ts.assertLogContains(logger, PAYLOAD_WARNING_MESSAGE)
+}
+
+func (ts *PayloadLimitsTestSuite) TestPayloadSizeWarningActivityInput() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	logger := ilog.NewMemoryLogger()
+	ts.ResetClientAndWorker(func(opts *client.Options) {
+		//opts.Logger = logger
```
yuandrew: commented out? TestPayloadSizeWarningActivityInput is failing in CI because of this commented out code, I think

internal/internal_task_handlers.go
```diff
@@ -2293,6 +2297,10 @@ func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice
 	canCtx, cancel := context.WithCancelCause(rootCtx)
 	defer cancel(nil)
 
+	if err := visitProtoPayloads(canCtx, ath.inboundPayloadVisitor, t); err != nil {
+		return ath.visitorErrorToActivityFailure("Activity task preprocess error:", t, err), nil
```
yuandrew:
```suggestion
		return ath.visitorErrorToActivityFailure("Activity task preprocess error: ", t, err), nil
```

internal/payload_limits.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (e payloadSizeError) Error() string {
+	return e.message
+}
+
+type payloadLimits struct {
+	payloadSize int64
+}
+
+func PayloadLimitOptionsToLimits(options PayloadLimitOptions) (payloadLimits, error) {
```
yuandrew: This is only used within `internal`, shouldn't need to be exported

internal/internal_task_pollers.go
```diff
@@ -461,21 +458,34 @@ func (wtp *workflowTaskProcessor) processWorkflowTask(task *workflowTask) (retEr
...
+
+	var taskErr error
+	if taskErr = visitProtoPayloads(ctx, wtp.inboundPayloadVisitor, task.task); taskErr != nil {
+		keyvals := []any{
+			tagWorkflowType, task.task.WorkflowType.GetName(),
+			tagWorkflowID, task.task.WorkflowExecution.GetWorkflowId(),
+			tagRunID, task.task.WorkflowExecution.GetRunId(),
+			tagAttempt, task.task.Attempt,
+		}
+		if errPayloadSize, isPayloadSizeError := taskErr.(payloadSizeError); isPayloadSizeError {
```
yuandrew: We typically want to use [errors.As](https://pkg.go.dev/errors#As) to match error types, they go through the error chain, in case the taskErr's outermost error isn't `payloadSizeError`

internal/payload_limits.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"errors"
+	"sync/atomic"
+
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/api/proxy"
+	"go.temporal.io/sdk/log"
+)
+
+// Options for when payload sizes exceed limits.
```
yuandrew: Should start with `PayloadLimitOptions`

internal/internal_task_handlers.go
```diff
@@ -2086,6 +2088,8 @@ func newActivityTaskHandlerWithCustomProvider(
 			params.UseBuildIDForVersioning,
 			params.DeploymentOptions.Version,
 		),
+		inboundPayloadVisitor:  params.inboundPayloadVisitor,
+		outboundPayloadVisitor: params.outboundPayloadVisitor,
 	}
 }
```
yuandrew, line 2098: Won't let me select below, but for `func (i *temporalInvoker) Heartbeat`(line 2114), are we planning on eventually checking the payload passed in there? Not sure how common it is to pass in a large payload there, but for completeness seems like we should have the same check there?

internal/internal_worker.go
```diff
@@ -1295,10 +1297,21 @@ func (aw *AggregatedWorker) start() error {
...
 	if err != nil {
 		return err
 	}
+
+	if nsLimits != nil && aw.executionParams.setPayloadErrorLimits != nil {
+		payloadSizeError := int64(0)
+		if nsLimits.BlobSizeLimitError > 0 {
+			payloadSizeError = nsLimits.BlobSizeLimitError
+		}
+		aw.executionParams.setPayloadErrorLimits(&payloadLimits{
```
yuandrew, line 1312: Claude flagged that we don't have a unit test for overwriting this value. The test if gave seems simple and reasonable

```
  func TestPayloadLimitsVisitorErrorLimitsOverwrite(t *testing.T) {
      visitor, setErrorLimits := newPayloadLimitsVisitor(payloadLimits{payloadSize: 100000}, nil)

      // Set initial error limit
      setErrorLimits(&payloadLimits{payloadSize: 1000})
      ctx := &proxy.VisitPayloadsContext{}
      _, err := visitor.Visit(ctx, []*commonpb.Payload{makeTestPayload(2000)})
      require.Error(t, err)

      // Overwrite with higher limit — same payload should now pass
      setErrorLimits(&payloadLimits{payloadSize: 100000})
      _, err = visitor.Visit(ctx, []*commonpb.Payload{makeTestPayload(2000)})
      require.NoError(t, err)
  }
```

internal/internal_task_pollers.go
yuandrew: Won't comment on all instances, a few instances of `err.(payloadSizeError)` as well

internal/internal_workflow_client.go
```diff
@@ -1549,33 +1550,39 @@ func (wc *WorkflowClient) loadCapabilities(ctx context.Context) (*workflowservic
 }
 
 // Get namespace capabilities, lazily fetching from server if not already obtained.
-func (wc *WorkflowClient) loadNamespaceCapabilities(metricsHandler metrics.Handler) (*namespacepb.NamespaceInfo_Capabilities, error) {
+func (wc *WorkflowClient) loadNamespaceData(metricsHandler metrics.Handler) (*namespacepb.NamespaceInfo_Capabilities, *namespacepb.NamespaceInfo_Limits, error) {
```
yuandrew: returning 3 fields is technically okay for now, but if we add any more, we'll probably want to switch to nesting them in a new `namespaceData` struct. Up to you if you want to do that now or defer to when another return type gets added, don't feel strongly here

internal/internal_workflow_client.go
```diff
@@ -1549,33 +1550,39 @@ func (wc *WorkflowClient) loadCapabilities(ctx context.Context) (*workflowservic
...
 	ctx := contextWithNewHeader(context.Background())
-	wc.namespaceCapabilitiesLock.RLock()
+	wc.namespaceDataLock.RLock()
 	capabilities := wc.namespaceCapabilities
-	wc.namespaceCapabilitiesLock.RUnlock()
-	if capabilities != nil {
-		return capabilities, nil
+	limits := wc.namespaceLimits
+	wc.namespaceDataLock.RUnlock()
+	if capabilities != nil && limits != nil {
```
yuandrew: This would mean we send an extra RPC for older server versions with only `capabilities` and no `limits` support instead of pulling from cache. The shared `namespaceData` struct would avoid this issue.

internal/payload_visitor.go
```diff
@@ -12,6 +12,27 @@ type PayloadVisitor interface {
 	Visit(ctx *proxy.VisitPayloadsContext, payloads []*commonpb.Payload) ([]*commonpb.Payload, error)
 }
 
+type compositePayloadVisitor struct {
```
yuandrew, line 20: Do we want to assert `var _ PayloadVisitor = (*compositePayloadVisitor)(nil)` for safety

test/payload_limits_test.go
```diff
@@ -0,0 +1,973 @@
...
+	workerOptions := worker.Options{}
+	if workerOpt != nil {
+		workerOpt(&workerOptions)
+	}
+	ts.worker = worker.New(ts.client, ts.taskQueueName, workerOptions)
+	ts.NoError(ts.worker.Start())
+}
+
+// pollForHistoryEvent polls the workflow history until an event of the specified type is found.
+// Returns the last event of the specified type, or nil if not found within the polling period.
```
yuandrew: Looks like this is hard coded to 10s, would be maybe good to mention this in the doc comment

internal/internal_task_handlers.go
jmaeagle99: Looks like server responds with an activity task failure instead of failing the workflow: https://github.com/temporalio/temporal/blob/main/service/frontend/workflow_handler.go#L1263

But we can probably still validate this from the SDK side, but what would we do in that case? My best guess without trying it is that it gets logged [here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_handlers.go#L2209) and then the error is just discarded [here](https://github.com/temporalio/sdk-go/blob/master/internal/internal_activity.go#L413). Maybe it should be the same behavior. Not sure that panicking is correct since it doesn't affect workflow determinism.

internal/internal_worker.go
jmaeagle99: I'll think about if that's the behavior we want or if it should error / panic instead.

jmaeagle99: I added this test.

internal/internal_task_handlers.go
jmaeagle99: https://github.com/temporalio/sdk-go/issues/2271

internal/internal_task_pollers.go
jmaeagle99: Include error handling and reporting, similar to top of the function.

jmaeagle99, line 819: We should be skipping size validation of failure requests since server doesn't cause a workflow failure.

eamsden (approved):

eamsden (approved):

PR #2257, opened 2026-03-30
Experimental: AWS S3 storage driver

### LINKED ISSUE #2251: S3 External Storage Driver - Go
Parent issue: https://github.com/temporalio/features/issues/783

jmaeagle99:
## What was changed

- Add an AWS S3 integration module that ships separately from the SDK.
  - Driver can consume arbitrary clients via an `s3driver.Client` implementation.
  - Module provide a client implementation for the AWS SDK v2 library via `awssdkv2.NewClient`.
  - Driver allows dynamic setting of the bucket via `Options.Bucket` field. Use `s3driver.StaticBucket` to set a static value.
  - Driver uploads payloads using content addressable storage under a simple format: `v0/d/sha256/<hash>`. More complex key specification is NYI.
  - Driver checks for key existence before attempting to upload large payloads.
  - Driver verifies integrity of externally retrieved payload as modified content would impact workflow determinism.
- Add unit tests that use the `gofakes3` server to simulate an S3 server.

## Why?

Provide a default driver for externally storing large payloads in AWS S3.

## Checklist

1. Closes #2251
2. How was this tested: Unit tests
3. Any docs updates needed? Yes

### DIFF
--- a/contrib/aws/s3driver/awssdkv2/client_test.go
+++ b/contrib/aws/s3driver/awssdkv2/client_test.go
@@ -87,6 +87,36 @@ func TestFakeS3_GetObject_NotFound(t *testing.T) {
 	assert.Contains(t, err.Error(), "NoSuchKey")
 }
 
+func TestFakeS3_PutObject_BucketNotFound(t *testing.T) {
+	client := newFakeS3(t)
+	ctx := context.Background()
+
+	err := client.PutObject(ctx, "no-such-bucket", "my/key", []byte("data"))
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "NoSuchBucket")
+}
+
+func TestFakeS3_ObjectExists_BucketNotFound(t *testing.T) {
+	client := newFakeS3(t)
+	ctx := context.Background()
+
+	// gofakes3 returns a generic NotFound for HeadObject on a missing bucket,
+	// which ObjectExists maps to (false, nil) — the same as a missing key.
+	// Real S3 would return NoSuchBucket here instead.
+	exists, err := client.ObjectExists(ctx, "no-such-bucket", "my/key")
+	assert.NoError(t, err)
+	assert.False(t, exists)
+}
+
+func TestFakeS3_GetObject_BucketNotFound(t *testing.T) {
+	client := newFakeS3(t)
+	ctx := context.Background()
+
+	_, err := client.GetObject(ctx, "no-such-bucket", "my/key")
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "NoSuchBucket")
+}
+
 func TestFakeS3_LargeObject(t *testing.T) {
 	client := newFakeS3(t, "test-bucket")
 	ctx := context.Background()


contrib/aws/s3driver/driver.go
```diff
@@ -0,0 +1,235 @@
...
+	claimKeyKey           = "key"
+	claimKeyHashAlgorithm = "hash_algorithm"
+	claimKeyHashValue     = "hash_value"
+)
+
+// BucketFunc resolves the target S3 bucket for a given payload. Use
+// StaticBucket for a fixed bucket name.
+//
+// NOTE: Experimental
+type BucketFunc func(ctx context.Context, payload *commonpb.Payload) string
```
jmaeagle99: StorageDriverStoreContext

contrib/aws/s3driver/awssdkv2/client.go
```diff
@@ -0,0 +1,72 @@
+// Package awssdkv2 provides an s3driver.Client implementation backed by the
+// AWS SDK for Go v2. Import this package alongside s3driver when using the
+// official AWS SDK v2 client; use a different adapter package for other
+// SDK versions.
+//
+// NOTE: Experimental
+package awssdkv2
```
yuandrew, line 7: I believe this package should have its own go.mod. That way users can use s3driver without pulling in awssdkv2 if they don't want to. And this would also put `gofakes3` solely as a awssdkv2 dep, not for the s3driver itself

contrib/aws/s3driver/driver.go
```diff
@@ -0,0 +1,235 @@
...
+	driverName     string
+	maxPayloadSize int
+}
+
+// NewDriver creates a new S3 StorageDriver with the given options.
+//
+// NOTE: Experimental
+func NewDriver(opts Options) (converter.StorageDriver, error) {
+	if opts.Client == nil {
+		return nil, errors.New("s3driver: Client is required")
```
yuandrew: nit: it's not idiomatic to include `s3driver:` prefix in Go errors, I don't think we do this anywhere else in the repo

contrib/aws/s3driver/awssdkv2/client_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/johannesboyne/gofakes3"
+	"github.com/johannesboyne/gofakes3/backend/s3mem"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// newFakeS3 starts an in-process fake S3 server and returns an
+// s3driver.S3StorageDriverClient backed by a real AWS SDK v2 client pointing
```
yuandrew: `S3StorageDriverClient` doesn't exist, guessing you renamed it to `Client`

contrib/aws/s3driver/driver.go
```diff
@@ -0,0 +1,235 @@
...
+	DriverName string
+
+	// MaxPayloadSize is the maximum serialized payload size in bytes that
+	// the driver will accept. Defaults to 50 MiB.
+	MaxPayloadSize int
+}
+
+// s3StorageDriver implements converter.StorageDriver by storing payloads in
+// Amazon S3 using content-addressable keys based on SHA-256 hashes.
+type s3StorageDriver struct {
```
yuandrew, line 63: Do we want to assert that s3StorageDriver always implements the `converter.StorageDriver` interface? We can do something like 

```
// Compile-time check that s3StorageDriver implements converter.StorageDriver.
var _ converter.StorageDriver = (*s3StorageDriver)(nil)
```

contrib/aws/s3driver/driver.go
```diff
@@ -0,0 +1,235 @@
...
+//
+// Two phases are used to avoid partial S3 uploads when validation fails:
+//  1. Marshal and validate all payloads concurrently.
+//  2. Upload concurrently — only reached if all payloads passed validation.
+func (d *s3StorageDriver) Store(
+	ctx converter.StorageDriverStoreContext,
+	payloads []*commonpb.Payload,
+) ([]converter.StorageDriverClaim, error) {
+	prepared := make([]preparedPayload, len(payloads))
+	eg1, _ := errgroup.WithContext(ctx.Context)
```
yuandrew: We're not actually using the errgroup context here. If we have `eg1, eg1Ctx` and use that for `d.bucketFunc`, if one goroutine fails, the other goroutines would bail sooner and we'd return an err sooner

contrib/aws/s3driver/driver_test.go
```diff
@@ -0,0 +1,421 @@
...
+	buckets := map[string]bool{}
+	for _, c := range claims {
+		buckets[c.ClaimData["bucket"]] = true
+	}
+	assert.True(t, buckets["bucket-a"])
+	assert.True(t, buckets["bucket-b"])
+}
+
+// errClient wraps a memClient and injects errors.
+type errClient struct {
```
yuandrew, line 260: Do we want to add a test for `existence check failed`?

contrib/aws/s3driver/driver.go
```diff
@@ -0,0 +1,235 @@
...
+func (d *s3StorageDriver) Retrieve(
+	ctx converter.StorageDriverRetrieveContext,
+	claims []converter.StorageDriverClaim,
+) ([]*commonpb.Payload, error) {
+	payloads := make([]*commonpb.Payload, len(claims))
+	g, gctx := errgroup.WithContext(ctx.Context)
+
+	for i, c := range claims {
+		g.Go(func() error {
+			bucket := c.ClaimData[claimKeyBucket]
```
yuandrew: Is it possible for `ClaimData` to ever be `nil` here?  Or if `bucket` or `key` are missing, the S3 error returned could be confusing/not clear.

Claude says 
> Retrieve can receive claims that were serialized/deserialized through Temporal's persistence layer, so corruption or version skew is plausible. A quick nil check and validation of required fields would give much better error messages.

contrib/aws/s3driver/driver.go
```diff
@@ -0,0 +1,235 @@
...
+			data, err := d.client.GetObject(gctx, bucket, key)
+			if err != nil {
+				return fmt.Errorf("s3driver: download failed [bucket=%s, key=%s]: %w", bucket, key, err)
+			}
+
+			if algo, ok := c.ClaimData[claimKeyHashAlgorithm]; ok && algo != hashAlgorithm {
+				return fmt.Errorf("s3driver: unsupported hash algorithm %q", algo)
+			}
+
+			if expectedHash, ok := c.ClaimData[claimKeyHashValue]; ok {
```
yuandrew: If `expectedHash` doesn't exist and `ok` is false, we skip this check entirely. Can we ever have `actualHash` be something but `claimKeyHashValue` not exist in `ClaimData`? Do we want to defensively program against this?

yuandrew: I think same for the check right above

yuandrew: Oh, do we intentionally not error here? Looking at `TestRetrieve_NoHashFields_BackwardCompat` for legacy behavior. But what if a claim is corrupted and used to have hash fields, but now doesn't? That would fail to flag any issues, I think

contrib/aws/s3driver/driver.go
```diff
@@ -0,0 +1,235 @@
...
+
+// Retrieve downloads payloads from S3 using the given claims, verifies their
+// integrity via SHA-256, and returns the deserialized payloads. Claims are
+// processed concurrently.
+func (d *s3StorageDriver) Retrieve(
+	ctx converter.StorageDriverRetrieveContext,
+	claims []converter.StorageDriverClaim,
+) ([]*commonpb.Payload, error) {
+	payloads := make([]*commonpb.Payload, len(claims))
+	g, gctx := errgroup.WithContext(ctx.Context)
```
yuandrew, line 182: nit: in `Store` you used the name `eg`, versus here is `g`.  Claude says `g`/`gctx` is more common in Go. And in `Store` using something like `prepareGroup` and `uploadGroup` could be more descriptive than `eg1` and `eg2`

contrib/aws/s3driver/awssdkv2/client_test.go
```diff
@@ -0,0 +1,145 @@
...
+		_, err := client.CreateBucket(ctx, &s3.CreateBucketInput{Bucket: aws.String(b)})
+		require.NoError(t, err)
+	}
+
+	return awssdkv2.NewClient(client), ts.Close
+}
+
+func TestFakeS3_PutGetRoundTrip(t *testing.T) {
+	c, cleanup := newFakeS3(t, "test-bucket")
+	defer cleanup()
```
yuandrew: nit: I think the more idiomatic thing here is to `t.Cleanup(ts.Close)` inside `newFakeS3`, then you wouldn't need to return any callback. And it ensures things are cleaned up even if the test panics.

contrib/aws/s3driver/client.go
```diff
@@ -0,0 +1,21 @@
...
+
+// Client is the interface that the driver uses to interact with S3. It covers
+// the three operations the driver needs: put, existence check, and get.
+// Use [go.temporal.io/sdk/contrib/aws/s3driver/awssdkv2.NewClient] to obtain
+// an implementation backed by the AWS SDK v2, or supply a custom
+// implementation for testing or alternative S3-compatible storage.
+//
+// NOTE: Experimental
+type Client interface {
+	// PutObject uploads data to the given bucket and key.
```
yuandrew: nit: Claude suggested that since these are interfaces that 3rd parties could implement, they could be more descriptive on the contract, like if `GetObject` returns a specific error type on not-found. I don't think this is strictly necessary, since I think most of our interfaces _don't_  do this, but in case you wanted to.

contrib/aws/s3driver/awssdkv2/client_test.go
```diff
@@ -0,0 +1,145 @@
...
+		Credentials: aws.CredentialsProviderFunc(func(_ context.Context) (aws.Credentials, error) {
+			return aws.Credentials{
+				AccessKeyID:     "test",
+				SecretAccessKey: "test",
+			}, nil
+		}),
+	})
+
+	ctx := context.Background()
+	for _, b := range buckets {
```
yuandrew, line 43: Do we want tests to verify what happens if we try to Put or Get a bucket that wasn't pre-created? I don't think there's a test for `Retrieve` either for a non-existent bucket

contrib/aws/s3driver/driver.go
jmaeagle99: I've removed the phase 1 errgroup since it's mostly very quick cpu bound work.

contrib/aws/s3driver/driver.go
jmaeagle99: I've made the driver strict on the fields. They are all required now.

PR #2258, opened 2026-03-30
add sdk plumbing for CreateWorkerDeploymentVersion

jaypipes:
Adds the SDK plumbing to create WorkerDeploymentVersions. The `internal.WorkerDeploymentHandle` interface was modified to add a new `CreateVersion` method that ends up calling the workflowservice's `CreateWorkerDeploymentVersion` RPC API.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -8,12 +8,13 @@ package client
 import (
 	"context"
 	"crypto/tls"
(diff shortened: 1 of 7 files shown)


jaypipes:
Not going this route.

PR #2261, opened 2026-03-31
Drain polled tasks on shutdown

### LINKED ISSUE #1197: Drain polled tasks on shutdown
When a worker is shutting down, if it has polled for tasks, but not started processing them it will drop them and continue to close

When a worker is shutting down, it should stop accepting new tasks, but let any it already has polled try to be processed in the worker stop timeout

**Additional context**
https://github.com/temporalio/sdk-go/pull/1195

yuandrew:
## What was changed
Refactored worker shutdown to use a two-stage approach: pollers shut down first, then the task dispatcher drains any remaining tasks before exiting. This ensures tasks polled during shutdown are processed rather than silently dropped.

Key changes:
- Added pollerWG to baseWorker to track poller goroutines separately from the global stopWG
- A closer goroutine waits for all pollers to finish, then closes taskQueueCh
- runTaskDispatcher now ranges over taskQueueCh
- pollTask always sends to taskQueueCh
- Removed the 5s timeout hack in doPoll from #2253 (no longer needed)

note: Graceful shutdown drain may bypass worker dispatch rate limiting for already-polled tasks, while still respecting slot concurrency. I think this is okay because the tasks are already polled by the worker, thus "owned", and this speeds up draining of tasks, decreasing likelihood of hitting `WorkerStopTimeout`. I don't think users would have any issues with this.

## Why?
PR #2199 changed shutdown to let the server complete in-flight polls, instead of cancelling them. This exposed a pre-existing race when a poller receives a task during shutdown, Go would silently dropping the task. The dispatcher had the same issue — it could exit on stopCh before reading pending tasks from the channel.

This aligns the Go SDK's shutdown with how Core SDK handles it:
1. Set a flag so pollers stop polling after their current attempt
2. Close channels from pollers → task processing
3. Wait for all in-flight tasks to complete & channels to be empty

## Checklist

1. Closes #1197

2. How was this tested:
    - New unit test TestTaskNotDroppedDuringShutdown — verifies a task polled during shutdown is processed, not dropped
    - Existing TestDoPollGracefulShutdown — validates both graceful and legacy poll completion

3. Any docs updates needed?
No

---

> [!NOTE]
> **Medium Risk**
> Changes core worker stop, poll cancellation, and task dispatch semantics; incorrect drain behavior could drop or mishandle tasks, though coverage is extensive.
> 
> **Overview**
> Refactors **worker shutdown** into a two-stage flow: pollers exit first, then the task dispatcher **drains** `taskQueueCh` so tasks received during shutdown are processed instead of dropped (fixes #1197).
> 
> When `workerPollCompleteOnShutdown` is enabled, **`baseWorker`** tracks pollers with `pollerWG`, closes `taskQueueCh` only after all pollers finish, and the dispatcher **ranges** over the channel (blocking sends from pollers, no race with `stopCh`). **`doPoll`** drops the old 5s grace wait and simply waits for poll completion on graceful shutdown. Workflow, activity, and Nexus **`ProcessTask`** paths skip `errStop` while drain mode is active.
> 
> **`AggregatedWorker`** now issues **`ShutdownWorker`** for session creation/activity task queues, sets **`noRepoll`** on session workers, and threads the same shutdown-drain flag into workflow/activity/nexus (and session) workers. Dispatch during drain uses a dedicated limiter context; tests cover rate limiting, legacy vs graceful behavior, and integration shutdown scenarios. Test server build enables **`frontend.enableCancelWorkerPollsOnShutdown`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 810efd4198483740382df1e54ac94cdea436d76e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/internal_worker_base.go
```diff
@@ -639,11 +644,10 @@ func (bw *baseWorker) pollTask(taskWorker scalableTaskPoller, slotPermit *SlotPe
 			taskWorker.pollerAutoscalerReportHandle.handleTask(task)
 		}
 
-		select {
-		case bw.taskQueueCh <- &polledTask{task: task, permit: slotPermit}:
-			didSendTask = true
-		case <-bw.stopCh:
```
Sushisource, line 669: Just checking the stop channel is still used elsewhere since we removed it in two spots

Sushisource (approved):

internal/internal_worker_base.go
yuandrew: yep! still used in plenty of other places

yuandrew:
recheck

yuandrew:
Closing in favor of #2368

PR #2274, opened 2026-04-07
💥 filter nil values from `SearchAttributes` for workflow start and continue-as-new

### LINKED ISSUE #2150: Stop sending null search attribute values on workflow start and continue as new

Null search attribute values can be produced in a few ways, namely because upsert search attributes puts them in workflow info (which arguably it shouldn't, but that's a separate discussion) and we reuse that map for continue as new. We should stop sending null search attributes on start or continue as new. Don't forget to mark this as 💥 backwards incompatible since it does very slightly change what attributes appear in a new workflow in this rare case.

eamsden:
## What was changed
💥 Start-like requests now omit unset search attributes instead of sending `binary/null` payloads in `SearchAttributes`.

Added `sanitizeSearchAttributesForStart` and applied it to:
- workflow start serialization
- continue-as-new requests
- workflow test environment start setup

Also added regression coverage to verify that:
- `nil` / `MetadataEncodingNil` search attributes are removed from start requests
- non-nil payloads that happen to have `Data == nil` (for example `[]byte(nil)` encoded as `binary/plain`) are preserved
- continue-as-new no longer carries over a search attribute after it has been unset

## Why?
Unset search attributes are represented in workflow state as nil payloads so upserts can remove them, but start-like requests should not resend those unset values.

Without this change, a workflow that unset a search attribute and then continued as new could carry that removed attribute forward incorrectly. The sanitization is intentionally keyed to `MetadataEncodingNil` so we only drop true nil payloads and do not accidentally strip other payloads that may have `Data == nil`.

## 💥 Behavior change details
- Workflow start and signal-with-start requests now omit search attributes encoded as `MetadataEncodingNil` (`binary/null`) instead of sending them to the server.
- Continue-as-new now applies the same sanitization, so a search attribute removed in the current run is not carried into the next run.
- The filtering is based on payload encoding, not `payload.Data == nil`, so values like `[]byte(nil)` encoded as `binary/plain` are still sent.
- The workflow test environment start helpers now mirror this behavior so tests match production request construction.

## Checklist

1. Closes #2150 

2. How was this tested:
Added/updated tests in:
- `internal/internal_workflow_client_test.go`
- `test/integration_test.go`
- `test/workflow_test.go`

Ran:
```sh
go test ./internal -run 'TestWorkflowClientSuite/(TestStartWorkflowWithMemoAndSearchAttr|TestSignalWithStartWorkflowWithMemoAndSearchAttr|TestSerializeSearchAttributesOmitsNilValuesOnStart)$'
```

The broader integration suite was not run in this session.

3. Any docs updates needed?
No.

### DIFF
--- a/internal/internal_search_attributes.go
+++ b/internal/internal_search_attributes.go
@@ -416,6 +416,31 @@ func serializeTypedSearchAttributes(searchAttributes map[SearchAttributeKey]inte
 	return &commonpb.SearchAttributes{IndexedFields: serializedAttr}, nil
 }
 
+// Start-like requests should not carry null or unset search attributes, even
+// though workflow info may still contain them in the current run.
+func sanitizeSearchAttributesForStart(attributes *commonpb.SearchAttributes) *commonpb.SearchAttributes {
+	if attributes == nil || len(attributes.IndexedFields) == 0 {
+		return attributes
+	}
+
+	filteredAttributes := make(map[string]*commonpb.Payload, len(attributes.IndexedFields))
+	filteredAny := false
+	for key, payload := range attributes.IndexedFields {
+		if payload == nil || string(payload.GetMetadata()[converter.MetadataEncoding]) == converter.MetadataEncodingNil {
+			filteredAny = true
+			continue
+		}
+		filteredAttributes[key] = payload
+	}
+	if !filteredAny {
+		return attributes
+	}
+	if len(filteredAttributes) == 0 {
+		return nil
+	}
+	return &commonpb.SearchAttributes{IndexedFields: filteredAttributes}
+}
+
 func serializeSearchAttributes(
 	untypedAttributes map[string]interface{},
 	typedAttributes SearchAttributes,
@@ -435,7 +460,7 @@ func serializeSearchAttributes(
 			return nil, err
 		}
 	}
-	return searchAttr, nil
+	return sanitizeSearchAttributesForStart(searchAttr), nil
 }
 
 func convertToTypedSearchAttributes(logger log.Logger, attributes map[string]*commonpb.Payload) SearchAttributes {
--- a/internal/internal_search_attributes.go
+++ b/internal/internal_search_attributes.go
@@ -426,7 +426,7 @@ func sanitizeSearchAttributesForStart(attributes *commonpb.SearchAttributes) *co
 	filteredAttributes := make(map[string]*commonpb.Payload, len(attributes.IndexedFields))
 	filteredAny := false
 	for key, payload := range attributes.IndexedFields {
-		if payload == nil || string(payload.GetMetadata()[converter.MetadataEncoding]) == converter.MetadataEncodingNil {
+		if string(payload.GetMetadata()[converter.MetadataEncoding]) == converter.MetadataEncodingNil {
 			filteredAny = true
 			continue
 		}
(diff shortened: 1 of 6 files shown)


jmaeagle99 (approved):

eamsden:
Please do not merge until @yuandrew can review

internal/internal_search_attributes.go
```diff
@@ -416,6 +416,31 @@ func serializeTypedSearchAttributes(searchAttributes map[SearchAttributeKey]inte
 	return &commonpb.SearchAttributes{IndexedFields: serializedAttr}, nil
 }
 
+// Start-like requests should not carry null or unset search attributes, even
+// though workflow info may still contain them in the current run.
```
yuandrew: Go doc style is to have the function name be the first word in the doc string

https://go.dev/blog/godoc

yuandrew (approved):

internal/internal_search_attributes.go
```diff
@@ -416,6 +416,31 @@ func serializeTypedSearchAttributes(searchAttributes map[SearchAttributeKey]inte
...
+func sanitizeSearchAttributesForStart(attributes *commonpb.SearchAttributes) *commonpb.SearchAttributes {
+	if attributes == nil || len(attributes.IndexedFields) == 0 {
```
alexshtin:
```suggestion
	if len(attributes.GetIndexedFields()) == 0 {
```

internal/internal_search_attributes.go
```diff
@@ -416,6 +416,31 @@ func serializeTypedSearchAttributes(searchAttributes map[SearchAttributeKey]inte
...
+		return attributes
+	}
+
+	filteredAttributes := make(map[string]*commonpb.Payload, len(attributes.IndexedFields))
+	filteredAny := false
+	for key, payload := range attributes.IndexedFields {
+		if payload == nil || string(payload.GetMetadata()[converter.MetadataEncoding]) == converter.MetadataEncodingNil {
```
alexshtin: Reads on `nil` maps are fine:

```suggestion
		if string(payload.GetMetadata()[converter.MetadataEncoding]) == converter.MetadataEncodingNil {
```

internal/internal_search_attributes.go
```diff
@@ -416,6 +416,31 @@ func serializeTypedSearchAttributes(searchAttributes map[SearchAttributeKey]inte
...
 
+// Start-like requests should not carry null or unset search attributes, even
+// though workflow info may still contain them in the current run.
+func sanitizeSearchAttributesForStart(attributes *commonpb.SearchAttributes) *commonpb.SearchAttributes {
+	if attributes == nil || len(attributes.IndexedFields) == 0 {
+		return attributes
+	}
+
+	filteredAttributes := make(map[string]*commonpb.Payload, len(attributes.IndexedFields))
+	filteredAny := false
```
alexshtin:
```suggestion
	var filteredAttributes map[string]*commonpb.Payload
```

and then use `filteredAttributes == nil` instead of `filteredAny` and move allocation on line 443. In 99% there will be nothing to filter.

alexshtin:
I faced this problem recently. Thanks for fixing it. Left you some minor comments.

internal/internal_search_attributes.go
eamsden: Fixed in 76b6471

PR #2280, opened 2026-04-08
Add support for Stand Alone Nexus Operations

Quinn-With-Two-Ns:
# Add standalone Nexus operation support 
                                                                                                                         
Adds the ability to start, poll, describe, cancel, terminate, list, and count Nexus operations directly from the     
  client, outside of workflow context. This mirrors the standalone activity pattern and is based on the API we already reviewed the cross SDK design for.                            
   
  ## API                                                                                                                    
                                                                                                                       
  - client.NewNexusClient(options) — creates a NexusClient bound to an endpoint+service                                  
  - NexusClient.ExecuteOperation(ctx, operation, input, options) — starts an operation and returns a handle
  - NexusOperationHandle.GetID() / GetRunID() / Get() / Describe() / Cancel() / Terminate() — handle operations          
  - client.GetNexusOperationHandle(options) — creates a handle to an existing operation (no network call)                
  - client.ListNexusOperations(ctx, options) — lists operations via visibility query                                     
  - client.CountNexusOperations(ctx, options) — counts operations via visibility query                                   
                                                                                                                         
  ## Interceptor support                                                                                                    
                                                                                                                       
  Adds 6 methods to ClientOutboundInterceptor: ExecuteNexusOperation, GetNexusOperationHandle, CancelNexusOperation,     
  TerminateNexusOperation, DescribeNexusOperation, PollNexusOperationResult. List and Count bypass the interceptor,    
  consistent with standalone activities.                                                                                 
                                                                                                                                                                                                                                                                                                                      
  ## Testing                                

  - Unit tests for interceptor header propagation, client validation, and operation name resolution                      
  - Integration tests covering execute+get, describe, get handle, cancel, terminate, list, count, and validation





<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Adds new experimental client APIs and interceptor hooks for starting/polling/canceling Nexus operations, which touches core client surface area and gRPC request handling. Risk is mitigated by new unit/integration coverage but may impact compatibility for custom interceptors/mocks.
> 
> **Overview**
> Adds **experimental standalone Nexus operation support** to the public `client` API, including `NewNexusClient`, `NexusClient.ExecuteOperation`, `NexusOperationHandle` (get/describe/cancel/terminate), plus `ListNexusOperations` and `CountNexusOperations` visibility queries.
> 
> Implements the feature end-to-end in `internal/internal_nexus_client.go` with new gRPC calls, typed options/metadata structs, operation-name resolution, and long-poll result retrieval; extends `ClientOutboundInterceptor` with Nexus-related methods and updates mocks/tests accordingly.
> 
> Updates integration/unit tests to cover the new standalone flow (including endpoint provisioning, handle behaviors, cancel/terminate, list/count) and adjusts CI/dev-server config to use a Nexus-capable dev server and to optionally disable the new tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8408739bd0308bcf9479163acccc9c847b0f3597. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


internal/internal_nexus_client.go
```diff
@@ -0,0 +1,864 @@
...
+	// ClientStartNexusOperationOptions contains configuration parameters for starting a Nexus operation execution.
+	//
+	// NOTE: Experimental
+	//
+	// Exposed as: [go.temporal.io/sdk/client.StartNexusOperationOptions]
+	ClientStartNexusOperationOptions struct {
+		// ID - The business identifier of the operation.
+		//
+		// Mandatory: No default.
+		ID string
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: This is to be consistent with Stand alone activities naming approach

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,864 @@
...
+	//
+	// Methods may be added to this interface; implementing it directly is discouraged.
+	//
+	// NOTE: Experimental
+	//
+	// Exposed as: [go.temporal.io/sdk/client.NexusClient]
+	ClientNexusClient interface {
+		// ExecuteOperation starts a Nexus operation and returns a handle to it.
+		//
+		// NOTE: Experimental
```
jmaeagle99, line 292: nit: Any reason why this particular func is called out as experimental when the others are not?

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,864 @@
...
+		// Describe returns detailed information about current state of the operation execution.
+		Describe(ctx context.Context, options ClientDescribeNexusOperationOptions) (*ClientNexusOperationExecutionDescription, error)
+		// Cancel requests cancellation of the operation.
+		Cancel(ctx context.Context, options ClientCancelNexusOperationOptions) error
+		// Terminate terminates the operation.
+		Terminate(ctx context.Context, options ClientTerminateNexusOperationOptions) error
+	}
+
+	// nexusClientImpl is the default implementation of ClientNexusClient.
+	nexusClientImpl struct {
```
jmaeagle99, line 334: nit: Do we want to compile-time assert that this and `clientNexusOperationHandleImpl` implement `ClientNexusClient` and `ClientNexusOperationHandle` respectively?

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,864 @@
...
+		result *ClientPollNexusOperationResultOutput
+	}
+)
+
+// GetSummary returns summary of the operation. See ClientStartNexusOperationOptions.Summary.
+// Returns empty string if there is no summary.
+// Uses the data converter of the client used to make the Describe call. Returns error if data conversion fails.
+//
+// NOTE: Experimental
+func (d *ClientNexusOperationExecutionDescription) GetSummary() (string, error) {
```
jmaeagle99, line 359: Because of the introduction of external storage, there is now payload visiting that takes a context. Do we want to consider plumbing through a context so that the call can be canceled if it hangs for some reason? I don't think this needs to be solved now, but something for us to think about.

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,864 @@
...
+// converter of the client used to make the Describe call. Returns nil if there was no failure.
+//
+// NOTE: Experimental
+func (d *ClientNexusOperationExecutionDescription) GetLastAttemptFailure() error {
+	failure := d.LastAttemptFailure
+	if failure == nil {
+		return nil
+	}
+	if err := visitProtoPayloads(context.Background(), d.inboundPayloadVisitor, failure); err != nil {
+		return err
```
jmaeagle99, line 386: This is the error of the payload visitor, not the result of `GetLastAttemptFailure`. There's no way for the caller to distinguish them. Changing the return type to `(error, error)` _might_ be more correct, even if it does look strange.

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,864 @@
...
+	}
+	return d.failureConverter.FailureToError(failure)
+}
+
+// GetLastAttemptFailure returns the last attempt failure of the cancellation info.
+// Returns nil if there was no failure.
+//
+// NOTE: Experimental
+func (c *ClientNexusOperationCancellationInfo) GetLastAttemptFailure() error {
```
jmaeagle99, line 395: Ditto https://github.com/temporalio/sdk-go/pull/2280/changes#r3066738258

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,864 @@
...
+		RunId:       in.RunID,
+		WaitStage:   enumspb.NEXUS_OPERATION_WAIT_STAGE_CLOSED,
+	}
+
+	var resp *workflowservice.PollNexusOperationExecutionResponse
+	for resp.GetOutcome() == nil {
+		grpcCtx, cancel := newGRPCContext(ctx, grpcLongPoll(true), grpcTimeout(pollNexusOperationTimeout), defaultGrpcRetryParameters(ctx))
+		var err error
+		resp, err = w.client.WorkflowService().PollNexusOperationExecution(grpcCtx, request)
+		cancel()
```
jmaeagle99, line 748: Should this be moved to it's own function and use `defer cancel()` in case a panic occurs?

mocks/Client.go
```diff
@@ -1294,6 +1294,104 @@ func (_m *Client) CountActivities(ctx context.Context, options client.CountActiv
 	return r0, r1
 }
 
+// NewNexusClient provides a mock function with given fields: options
+func (_m *Client) NewNexusClient(options client.NexusClientOptions) (client.NexusClient, error) {
+	ret := _m.Called(options)
+
+	if len(ret) == 0 {
+		panic("no return value specified for NexusClient")
```
jmaeagle99:
```suggestion
		panic("no return value specified for NewNexusClient")
```

client/client.go
```diff
@@ -1499,6 +1585,30 @@ type (
...
+		// For Nexus operations within workflows, use workflow.NexusClient instead.
+		//
+		// NOTE: Experimental
+		NewNexusClient(options NexusClientOptions) (NexusClient, error)
+
+		// GetNexusOperationHandle creates a handle to the referenced Nexus operation.
+		// No network call is made. The handle can be used to poll, describe, cancel, or terminate.
+		//
+		// NOTE: Experimental
+		GetNexusOperationHandle(options GetNexusOperationHandleOptions) NexusOperationHandle
```
jmaeagle99, line 1599: How is this able to return a handle to a Nexus operation if it doesn't have endpoint or service information?

jmaeagle99: I understand that calling this method doesn't incur a network call. But user code is something like:

```go
client := client.Dial(...)
handle := client.GetNexusOperationHandle(client.GetNexusOperationHandleOptions{Id: "some-id"})
desc, err := handle.Describe(ctx, client.DescribeNexusOperationOptions{})
```

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,864 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	groups := make([]ClientCountNexusOperationsAggregationGroup, len(resp.Groups))
+	for i, group := range resp.Groups {
+		groupValues := make([]any, len(group.GroupValues))
+		for j, groupValue := range group.GroupValues {
+			// should never fail, and if it does, leaving nil behind
+			_ = converter.GetDefaultDataConverter().FromPayload(groupValue, &groupValues[j])
```
jmaeagle99, line 634: Should it be documented on the public surface that `CountNexusOperations` only uses the default data converter?

client/client.go
```diff
@@ -1499,6 +1585,30 @@ type (
 		// NOTE: Experimental
 		CountActivities(ctx context.Context, options CountActivitiesOptions) (*CountActivitiesResult, error)
 
+		// NexusClient creates a new Nexus client bound to the given endpoint and service.
```
jmaeagle99:
```suggestion
		// NewNexusClient creates a new Nexus client bound to the given endpoint and service.
```

internal/internal_nexus_client_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	_, err = client.NewNexusClient(ClientNexusClientOptions{Endpoint: "ep"})
+	require.ErrorContains(t, err, "service is required")
+
+	nc, err := client.NewNexusClient(ClientNexusClientOptions{Endpoint: "ep", Service: "svc"})
+	require.NoError(t, err)
+	require.NotNil(t, nc)
+}
+
+func TestResolveNexusOperationName(t *testing.T) {
```
jmaeagle99, line 91: Add cases for typed operation (success) and nearly correct type (failure).

test/integration_test.go
```diff
@@ -9095,3 +9099,194 @@ func (ts *IntegrationTestSuite) TestPanicWithDeferredYield() {
...
+func (ts *IntegrationTestSuite) TestExecuteNexusOperationSuite() {
+	if os.Getenv("DISABLE_STANDALONE_NEXUS_TESTS") != "" {
+		ts.T().SkipNow()
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), ctxTimeout)
+	defer cancel()
+
+	// Create a Nexus endpoint targeting our task queue.
+	endpoint := "sdk-go-nexus-standalone-test-ep-" + uuid.NewString()
```
jmaeagle99, line 9112: Should the endpoint be deleted at the end of the test? Maybe a deferred deletion?

client/client.go
Quinn-With-Two-Ns: You don't need a service or endpoint to get a handler to a Nexus operation, all you need is the ID. Yes your code sample is correct.

mocks/Client.go
Quinn-With-Two-Ns: Ah I renamed this method and missed a few places thanks!

test/integration_test.go
```diff
@@ -9095,3 +9099,194 @@ func (ts *IntegrationTestSuite) TestPanicWithDeferredYield() {
...
+			ID:                     operationID,
+			ScheduleToCloseTimeout: 10 * time.Second,
+		})
+
+		// Wait for operation to complete.
+		err := handle.Get(ctx, nil)
+		ts.NoError(err)
+
+		// Get a handle to the same operation.
+		handle2 := ts.client.GetNexusOperationHandle(client.GetNexusOperationHandleOptions{
```
jmaeagle99, line 9212: Consider adding a test where the run ID is not specified.

internal/internal_nexus_client.go
Quinn-With-Two-Ns: Summary is a fixed very small size so unlikely it would use external storage also `GetSummary()` on other APIs doesn't take a context here so I don't think we should be inconsistent

internal/internal_nexus_client.go
Quinn-With-Two-Ns: Yes in general we never distinguish for these cases anywhere in the SDK this is not unique. When you get a workflow result and the deserialization fails that is returned as the error same as if the workflow itself failed

internal/internal_nexus_client.go
Quinn-With-Two-Ns: I don't see why we would? this is generally consistent with how we poll in other places in the Go SDK https://github.com/temporalio/sdk-go/blob/81673515d90aa5ab849cee3e31f091fb8d4f0053/internal/internal_activity_client.go#L668

internal/internal_nexus_client.go
Quinn-With-Two-Ns: No harm in doing that, but `clientNexusOperationHandleImpl` is used as a `ClientNexusClient` so the SDK already wouldn't compile if it didn't

internal/internal_nexus_client.go
Quinn-With-Two-Ns: No, other methods should be called out as experimental to good catch

internal/internal_nexus_client.go
Quinn-With-Two-Ns: another example is `func (d *ClientActivityExecutionDescription) GetLastFailure() error`

internal/internal_nexus_client.go
Quinn-With-Two-Ns: It's a payload generated by the server so I am not sure how that information would be relevant to a user?

internal/internal_nexus_client.go
Quinn-With-Two-Ns: Generally you would distinguish by the type of the error

internal/internal_nexus_client.go
Quinn-With-Two-Ns: Done!

Quinn-With-Two-Ns:
Note this PR is missing support for otel I was planning on addressing that in a separate PR to keep the scope down

.github/workflows/ci.yml
```diff
@@ -147,6 +147,7 @@ jobs:
           DISABLE_SERVER_1_27_TESTS: "1"
           DISABLE_PRIORITY_TESTS: "1"
           DISABLE_STANDALONE_ACTIVITY_TESTS: "1"
+          DISABLE_STANDALONE_NEXUS_TESTS: "1"
```
VegetarianOrc, line 150: This is following the existing pattern, but should we be defaulting to exclude all these tests in CI? Or maybe I am misunderstanding the setup.

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,897 @@
...
+	if in.Options.ID == "" {
+		return nil, errors.New("operation ID is required")
+	}
+	if in.Options.ScheduleToCloseTimeout < 0 {
+		return nil, errors.New("ScheduleToCloseTimeout must not be negative")
+	}
+
+	// Encode input as a single Payload (not Payloads)
+	var inputPayload *commonpb.Payload
+	if in.Input != nil {
```
jmaeagle99: Should the input be converter unconditionally so that it creates a "null" payload instead of a literal nil value?

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,897 @@
...
+		payloads := &commonpb.Payloads{Payloads: []*commonpb.Payload{v.Result}}
+		return &ClientPollNexusOperationResultOutput{Result: newEncodedValue(payloads, w.client.dataConverter)}, nil
+	case *workflowservice.PollNexusOperationExecutionResponse_Failure:
+		return &ClientPollNexusOperationResultOutput{Error: w.client.failureConverter.FailureToError(v.Failure)}, nil
+	default:
+		return nil, fmt.Errorf("unexpected nexus operation outcome type: %T", v)
+	}
+}
+
+func (w *workflowClientInterceptor) DescribeNexusOperation(
```
jmaeagle99, line 782: Should cancel, describe, and terminate check that the operation ID is not empty?

jmaeagle99 (approved):

test/nexus_test.go
```diff
@@ -395,7 +395,7 @@ func TestNexusSyncOperation(t *testing.T) {
 	t.Run("timeout", func(t *testing.T) {
 		_, err := nexusclient.ExecuteOperation(ctx, nc, syncOp, "timeout", nexus.StartOperationOptions{
 			// Force shorter timeout to speed up the test and get a response back.
-			Header: nexus.Header{nexus.HeaderRequestTimeout: "300ms"},
+			Header: nexus.Header{nexus.HeaderRequestTimeout: "5s"},
```
VegetarianOrc, line 398: Is this change needed/related? Was 300ms too aggressive and causing flakes?

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,897 @@
...
+		id     string
+		runID  string
+		result *ClientPollNexusOperationResultOutput
+	}
+)
+
+var (
+	_ ClientNexusClient          = &nexusClientImpl{}
+	_ ClientNexusOperationHandle = &clientNexusOperationHandleImpl{}
+)
```
VegetarianOrc, line 362: Nit:
```suggestion
var (
	_ ClientNexusClient          = (*nexusClientImpl)(nil)
	_ ClientNexusOperationHandle = (*clientNexusOperationHandleImpl)(nil)
)
```

VegetarianOrc (approved):
Couple of questions, but looks good!

.github/workflows/ci.yml
yuandrew: We should definitely revisit how we do this, last time I checked there was definitely some that could be removed at this point. I can add to TODO for myself

client/client.go
```diff
@@ -990,6 +990,92 @@ type (
...
+	StartNexusOperationOptions = internal.ClientStartNexusOperationOptions
+
+	// NexusClientOptions contains options for creating a NexusClient.
+	//
+	// NOTE: Experimental
+	NexusClientOptions = internal.ClientNexusClientOptions
+
+	// NexusClient is the client for starting Nexus operations bound to a specific endpoint and service.
+	// This is for standalone Nexus operations outside of workflow context.
+	// For Nexus operations within workflows, use workflow.NexusClient.
```
yuandrew, line 1005: Do we want to also mention in `workflow.NexusClient` that we have `client.NexusClient` for standalone operations?

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,897 @@
...
+type (
+	// ClientStartNexusOperationOptions contains configuration parameters for starting a Nexus operation execution.
+	//
+	// NOTE: Experimental
+	//
+	// Exposed as: [go.temporal.io/sdk/client.StartNexusOperationOptions]
+	ClientStartNexusOperationOptions struct {
+		// ID - The business identifier of the operation.
+		//
+		// Mandatory: No default.
```
yuandrew: nit: sometimes we use `Mandatory`, sometimes we use `Required`. We should probably align on a single term. I lean towards `Required`, and it's used more commonly in the repo today.

internal/internal_nexus_client.go
```diff
@@ -0,0 +1,897 @@
...
+	}
+	if in.Options.ScheduleToCloseTimeout < 0 {
+		return nil, errors.New("ScheduleToCloseTimeout must not be negative")
+	}
+
+	// Encode input as a single Payload (not Payloads)
+	var inputPayload *commonpb.Payload
+	if in.Input != nil {
+		var err error
+		inputPayload, err = dataConverter.ToPayload(in.Input)
```
yuandrew: I was going to ask if we want to add serialization context here, but I realize I didn't add it to non-standalone nexus operations yet

yuandrew (approved):

.github/workflows/ci.yml
Quinn-With-Two-Ns: The logic here just excludes when we are using docker to test since the docker tests don't use the pre-release CLI

test/nexus_test.go
Quinn-With-Two-Ns: Yeah, specifically the endpoint registry is eventually consistent right now so under load it can take too long for the endpoint to become visible

PR #2293, opened 2026-04-13
Port system Nexus envelope encoding rewrite

tconley1428:
## Summary
- port the Python SDK system Nexus envelope handling to the Go SDK
- keep the outer SignalWithStartWorkflowExecution system envelope unencoded while rewriting nested user payloads through codec and external storage
- add unit and integration coverage for the new rewrite path

## Testing
- GOCACHE=/tmp/temporal-sdk-go-cache go test ./internal -run TestSystemNexusOutboundPayloadVisitor_RewritesNestedPayloadsOnly -count=1
- GOCACHE=/tmp/temporal-sdk-go-cache go test ./internal -run TestDoesNotExist
- GOCACHE=/tmp/temporal-sdk-go-cache go test ./temporalnexus/system -run TestDoesNotExist
- GOCACHE=/tmp/temporal-sdk-go-cache go test . -run TestSystemNexusDefersOuterEnvelopeEncoding -count=1


internal/workflow.go
```diff
@@ -3061,7 +2982,12 @@ func (wc *workflowEnvironmentInterceptor) prepareNexusOperationParams(ctx Contex
 		return executeNexusOperationParams{}, fmt.Errorf("invalid 'operation' parameter, must be an OperationReference or a string")
 	}
 
-	payload, err := dc.ToPayload(input.Input)
+	payloadConverter := dc
```
tconley1428, line 3076: It isn't really a payload converter in all cases

PR #2294, opened 2026-04-13
Enable external storage concurrency

jmaeagle99:
## What was changed

Allow worker external storage visiting to be concurrent by default with a default value of 3 max concurrent visits per workflow task. Client calls and activity tasks do not concurrently visit across their proto graphs as there is not much gain to do so.

## Why?

With external storage being primarily I/O bound, a small amount of default concurrency should be enabled for workflow tasks so that uploads/downloads can process faster.

## Checklist

1. How was this tested: New unit tests on payload visitor
2. Any docs updates needed? Yes

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -43,7 +43,7 @@ const (
 
 	defaultDefaultHeartbeatThrottleInterval               = 30 * time.Second
 	defaultMaxHeartbeatThrottleInterval                   = 60 * time.Second
-	defualtMaxConcurrentWorkflowTaskExternalStorageVisits = 3
+	defaultMaxConcurrentWorkflowTaskExternalStorageVisits = 3
 )
 
 var (
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -41,8 +41,9 @@ const (
 
 	noRetryBackoff = time.Duration(-1)
 
-	defaultDefaultHeartbeatThrottleInterval = 30 * time.Second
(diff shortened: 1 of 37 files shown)


internal/internal_task_handlers.go
```diff
@@ -41,8 +41,9 @@ const (
 
 	noRetryBackoff = time.Duration(-1)
 
-	defaultDefaultHeartbeatThrottleInterval = 30 * time.Second
-	defaultMaxHeartbeatThrottleInterval     = 60 * time.Second
+	defaultDefaultHeartbeatThrottleInterval               = 30 * time.Second
+	defaultMaxHeartbeatThrottleInterval                   = 60 * time.Second
+	defualtMaxConcurrentWorkflowTaskExternalStorageVisits = 3
```
jmaeagle99: default

internal/internal_worker.go
```diff
@@ -2631,6 +2634,9 @@ func setWorkerOptionsDefaults(options *WorkerOptions) {
 	if options.MaxHeartbeatThrottleInterval == 0 {
 		options.MaxHeartbeatThrottleInterval = defaultMaxHeartbeatThrottleInterval
 	}
+	if options.MaxConcurrentWorkflowTaskExternalStorageVisits == 0 {
```
jmaeagle99, line 2641: Validate lower bound

jmaeagle99: Add tests for this as well.

eamsden (approved):

PR #2295, opened 2026-04-14
Experimental: Payload HTTP Handler for Codecs and External Storage

jmaeagle99:
## What was changed

- Move all external storage types, implementations, and algorithms to `internal/extstore` package. Re-export public types via the `converter` package. Necessary so the implementation and algorithms can be shared internally with `converter` package.
- Add `NewPayloadHTTPHandler` and `PayloadHTTPHandlerOptions` to allow Web UI and CLI to encode, decode, and download payloads when external storage is enabled. The handler is fully backward compatible for Web UI and CLI use, but should not be used as the target of a remote codec without careful consideration to the fact that it does more than applying a single list of codecs as the existing codec HTTP handlers do.

## Why?

The ordering of payload operations is inconsistent under different customer configurations when external storage is enabled. The introduction of the new HTTP handler allows the Web UI and CLI to encode, decode, and optionally download externally stored payloads without having to understand the order of operations itself.

Examples of customer scenarios (showing only the encode path):
- Customer has encryption and external storage configured in worker: `Codecs (in DataConverter) -> External Storage`
- Customer has external storage configured in worker and encryption in proxy: `External Storage -> Codecs (in proxy)`
- Customer has compression and external storage configured in worker and encryption in proxy: `Codecs (in DataConverter) -> External Storage -> Codecs (in proxy)`

## Checklist

1. How was this tested: Tests
1. Any docs updates needed? Yes

### DIFF
--- a/converter/ui_payload_handler.go
+++ b/converter/ui_payload_handler.go
@@ -1,0 +1,210 @@
+// Package converter provides an HTTP handler for a Temporal codec server that
+// understands the two-layer codec architecture used with external payload storage.
+//
+// The handler exposes two routes:
+//
+//   - POST /ui/decode — decodes payloads through the post-storage then pre-storage
+//     codec chains. If a payload is a storage reference after post-storage
+//     decoding it is returned as-is so the caller can resolve it via /ui/download.
+//
+//   - POST /ui/download — accepts storage reference payloads, retrieves the
+//     original payloads from the configured storage drivers, then decodes
+//     them through the pre-storage codec chain.
+//
+// The wire format for both routes is identical to the existing
+// [NewPayloadCodecHTTPHandler]: a JSON-encoded [commonpb.Payloads] request body
+// and response body, with Content-Type application/json.
+
+package converter
+
+import (
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"strings"
+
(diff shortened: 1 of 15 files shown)


converter/ui_payload_handler.go
```diff
@@ -0,0 +1,210 @@
...
+// /ui/download routes for Temporal Web UI using the provided options.
+//
+// NOTE: Experimental
+func NewUIPayloadHTTPHandler(options UIPayloadHTTPHandlerOptions) (http.Handler, error) {
+	h := &uiPayloadHTTPHandler{
+		postStorageCodecs: options.PostStorageCodecs,
+		preStorageCodecs:  options.PreStorageCodecs,
+	}
+	if len(options.StorageDrivers) > 0 {
+		params, err := extstore.ExternalStorageToParams(ExternalStorage{Drivers: options.StorageDrivers})
```
jmaeagle99: Add a selector that no-ops to satisfy the conditions in ExternalStorageToParams. The selector isn't really needed for the download path.

jmaeagle99: Add a test that has multiple drivers

eamsden (approved):

eamsden (approved):

PR #2298, opened 2026-04-18
Internal: fix already-ready future chaining to properly complete dest…

kritibehl:
## What was changed
Updated futureImpl.Chain so that when the source future is already ready, the destination future is completed via Set(...) instead of manually copying value, err and ready=true.

Specifically, this changes the already-ready fast path from direct field assignment to:

val, err := ch.GetValueAndError()
f.Set(val, err)

Also added regression tests covering:

chaining from an already-ready future and then calling Get()
propagation through an already-chained dependent future

## Why?
The previous already-ready fast path left the destination future in a partially completed state.

It set ready=true, but it did not close the destination future’s internal channel and did not propagate to already-chained dependents. As a result:

IsReady() could return true while Get() still blocked
chained dependents were skipped

Routing this path through Set(...) keeps completion semantics consistent with other future completion paths by closing the channel and propagating to chained futures.

## Checklist

1. Closes #2279

2. How was this tested:

Added TestChainAlreadyReadyFuture
Added TestChainAlreadyReadyFuturePropagates
Ran:
go run . unit-test -run TestChainAlreadyReadyFuture
go run . unit-test -run Chain

3. Any docs updates needed?

No docs updates needed. This is an internal bug fix with regression coverage.


Sushisource (approved):
Thank you!

PR #2299, opened 2026-04-21
Workflowcheck enforce anonymous func isn't used as local activity

### LINKED ISSUE #1341: WorkflowChecker should consider anonymous functions in local activities deterministic, but maybe warn about them anyway

Quinn-With-Two-Ns:
## What was changed

Change `workflowcheck` tool to check any `ExecuteLocalActivity` in a workflow to make sure a user isn't trying to use an anonymous function as a local activity. Check is essentially best effort 

## Why?

Anonymous function names are not deterministic so they cannot be used as local activities in workflows.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/1341

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Adds new static-analysis rules and AST flow heuristics that may introduce false positives/negatives when determining whether a variable holds an anonymous function at a call site.
> 
> **Overview**
> `workflowcheck` now treats passing an anonymous function to `workflow.ExecuteLocalActivity` as non-deterministic by adding a configurable rule (`RejectsAnonymousFuncArgs`) to the determinism analyzer and emitting a new `ReasonAnonymousFunc` diagnostic.
> 
> The analysis includes best-effort tracking of whether an identifier argument could contain a func literal via prior assignments/decls (and conservatively through branches), and expands ignore handling to cover ignored RHS assignment call expressions. Testdata is extended with many `ExecuteLocalActivity` scenarios plus stub `ExecuteLocalActivity` definitions in workflow/internal packages to compile the new cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 356a149601f84ccc04d503135401b52e12031994. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/contrib/tools/workflowcheck/determinism/checker.go
+++ b/contrib/tools/workflowcheck/determinism/checker.go
@@ -516,6 +516,11 @@ func containsAnonAssignToVar(node ast.Node, obj types.Object, callPos token.Pos,
 		if found || n == nil || n.Pos() >= callPos {
 			return false
 		}
+		// Don't descend into function literals — assignments inside closures
+		// that are only defined (not invoked) should not count.
+		if _, ok := n.(*ast.FuncLit); ok {
+			return false
+		}
 		if assign, ok := n.(*ast.AssignStmt); ok {
 			for i, lhs := range assign.Lhs {
 				if lhsIdent, ok := lhs.(*ast.Ident); ok && info.ObjectOf(lhsIdent) == obj && i < len(assign.Rhs) {
--- a/contrib/tools/workflowcheck/determinism/checker.go
+++ b/contrib/tools/workflowcheck/determinism/checker.go
@@ -31,6 +31,10 @@ type Config struct {
 	EnableObjectFacts bool
 	// Map `package -> function names` with functions making any argument deterministic
 	AcceptsNonDeterministicParameters map[string][]string
+	// Map of fully-qualified function name to the argument index that must not
+	// be an anonymous function (function literal). If an anonymous function is
+	// passed at the specified index, it is flagged as non-deterministic.
+	RejectsAnonymousFuncArgs map[string]int
 }
 
 // Checker is a checker that can run analysis passes to check for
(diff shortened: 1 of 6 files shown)


contrib/tools/workflowcheck/determinism/checker.go
```diff
@@ -430,6 +445,96 @@ func (c *collector) applyFuncNonDeterminisms(f *funcInfo, p PackageNonDeterminis
...
+}
+
+func containsAnonAssignToVar(node ast.Node, obj types.Object, callPos token.Pos, info *types.Info) bool {
+	found := false
+	ast.Inspect(node, func(n ast.Node) bool {
+		if found || n == nil || n.Pos() >= callPos {
+			return false
+		}
+		// Don't descend into function literals — assignments inside closures
+		// that are only defined (not invoked) should not count.
```
yuandrew, line 520: This comment isn't entirely accurate, right? The function literal could, in fact, be invoked down the line? In the tests below, we're missing this scenario, which I think would fail due to us unconditionally ignoring function literals

```
  f := myLocalActivity
  g := func() {
      f = func(ctx context.Context) error { return nil }
  }
  g()
  workflow.ExecuteLocalActivity(ctx, f)
```

contrib/tools/workflowcheck/workflow/testdata/src/a/workflow.go
```diff
@@ -68,6 +68,144 @@ func NotWorkflow(ctx context.Context) {
...
+	return nil
+}
+
+func WorkflowLocalActivityBranchOverridden(ctx workflow.Context) error {
+	if true {
+		f := func(ctx context.Context) error { return nil }
+		_ = f
+	}
+	// f here is a different variable; this uses myLocalActivity directly
+	workflow.ExecuteLocalActivity(ctx, myLocalActivity)
```
yuandrew, line 162: I think this test isn't doing what was intended, it seems like you wanted to test shadowing f, but `ExecuteLocalActivity` is calling myLocalActivity directly, not an outer `f`
```suggestion
    f := myLocalActivity
	if true {
		f := func(ctx context.Context) error { return nil }
		_ = f
	}
	// f here is a different variable; this uses myLocalActivity directly
	workflow.ExecuteLocalActivity(ctx, f)
```

contrib/tools/workflowcheck/determinism/checker.go
Quinn-With-Two-Ns: Yeah this is true, but this would add significant complexity to the check since we would need to check if the closure is ever invoked and that can get very complex.

yuandrew: yeah, that's valid. But I would recommend making this comment a little more clear that we're consciously choosing not to evaluate inside of closures, regardless of if they're invoked or not, due to complexity

PR #2302, opened 2026-04-23
feat(otel): add otel tracing for standalone activities.

### LINKED ISSUE #2286: Standalone Activities - tracing interceptor doesn't intercept standalone activities client calls
`tracingClientOutboundInterceptor` does not currently implement the new interceptor methods related to standalone activities. This results in not creating the client-side spans when executing activities.

cconstable:
## What changed

- Add tracing support for standalone activities (activities started directly via the client, outside a workflow). The `tracingClientOutboundInterceptor` now instruments `ExecuteActivity`
- Fix small issue with doclink tool

## Why

- Fixes https://github.com/temporalio/sdk-go/issues/2286

## Checklist

- Adds new `TestStandaloneActivitySpanCreation` to `tracing_interceptor_test.go`

### DIFF
--- a/contrib/opentelemetry/tracing_interceptor_test.go
+++ b/contrib/opentelemetry/tracing_interceptor_test.go
@@ -14,10 +14,8 @@ import (
 	"go.opentelemetry.io/otel/sdk/trace/tracetest"
 	"go.opentelemetry.io/otel/trace"
 
-	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/contrib/opentelemetry"
 	"go.temporal.io/sdk/interceptor"
-	"go.temporal.io/sdk/internal"
 	"go.temporal.io/sdk/internal/interceptortest"
 	"go.temporal.io/sdk/temporal"
 	"go.temporal.io/sdk/testsuite"
@@ -35,34 +33,6 @@ func TestSpanPropagation(t *testing.T) {
 	testTracer := &testTracer{Tracer: tracer, rec: &rec}
 	interceptortest.RunTestWorkflow(t, testTracer)
 	interceptortest.AssertSpanPropagation(t, testTracer)
-}
-
-func TestStandaloneActivitySpanCreation(t *testing.T) {
-	rec := tracetest.NewSpanRecorder()
-	tracer, err := opentelemetry.NewTracer(opentelemetry.TracerOptions{
-		Tracer: sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(rec)).Tracer(""),
-	})
-	require.NoError(t, err)
-
-	ctx := internal.NewHeaderContext(context.Background())
-	outbound := interceptor.NewTracingInterceptor(tracer).InterceptClient(&testNoopClientOutbound{})
-	_, _ = outbound.ExecuteActivity(ctx, &interceptor.ClientExecuteActivityInput{
-		ActivityType: "test-saa",
-		Options:      &client.StartActivityOptions{ID: "test-saa-123"},
-	})
-
-	spans := rec.Ended()
-	require.Len(t, spans, 1)
-	assert.Equal(t, "StartActivity:test-saa", spans[0].Name())
-
-	var foundActivityID bool
-	for _, attr := range spans[0].Attributes() {
-		if string(attr.Key) == "temporalActivityID" {
-			foundActivityID = true
-			assert.Equal(t, "test-saa-123", attr.Value.AsString())
-		}
-	}
-	require.True(t, foundActivityID, "expected activity ID span attribute")
 }
 
 type testTracer struct {
@@ -264,12 +234,4 @@ func TestSpanFromWorkflowContextNoOpSpan(t *testing.T) {
 
 	require.True(t, env.IsWorkflowCompleted())
 	require.NoError(t, env.GetWorkflowError())
-}
-
-type testNoopClientOutbound struct {
-	interceptor.ClientOutboundInterceptorBase
-}
-
-func (n *testNoopClientOutbound) ExecuteActivity(_ context.Context, _ *interceptor.ClientExecuteActivityInput) (client.ActivityHandle, error) {
-	return nil, nil
 }
--- a/contrib/opentelemetry/tracing_interceptor_test.go
+++ b/contrib/opentelemetry/tracing_interceptor_test.go
@@ -14,8 +14,10 @@ import (
 	"go.opentelemetry.io/otel/sdk/trace/tracetest"
 	"go.opentelemetry.io/otel/trace"
 
+	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/contrib/opentelemetry"
 	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/internal"
 	"go.temporal.io/sdk/internal/interceptortest"
 	"go.temporal.io/sdk/temporal"
 	"go.temporal.io/sdk/testsuite"
@@ -33,6 +35,34 @@ func TestSpanPropagation(t *testing.T) {
 	testTracer := &testTracer{Tracer: tracer, rec: &rec}
 	interceptortest.RunTestWorkflow(t, testTracer)
 	interceptortest.AssertSpanPropagation(t, testTracer)
+}
+
+func TestStandaloneActivitySpanCreation(t *testing.T) {
+	rec := tracetest.NewSpanRecorder()
+	tracer, err := opentelemetry.NewTracer(opentelemetry.TracerOptions{
+		Tracer: sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(rec)).Tracer(""),
+	})
+	require.NoError(t, err)
+
+	ctx := internal.NewHeaderContext(context.Background())
+	outbound := interceptor.NewTracingInterceptor(tracer).InterceptClient(&testNoopClientOutbound{})
+	_, _ = outbound.ExecuteActivity(ctx, &interceptor.ClientExecuteActivityInput{
+		ActivityType: "test-saa",
+		Options:      &client.StartActivityOptions{ID: "test-saa-123"},
+	})
+
+	spans := rec.Ended()
+	require.Len(t, spans, 1)
+	assert.Equal(t, "StartActivity:test-saa", spans[0].Name())
+
+	var foundActivityID bool
+	for _, attr := range spans[0].Attributes() {
+		if string(attr.Key) == "temporalActivityID" {
+			foundActivityID = true
+			assert.Equal(t, "test-saa-123", attr.Value.AsString())
+		}
+	}
+	require.True(t, foundActivityID, "expected activity ID span attribute")
 }
 
 type testTracer struct {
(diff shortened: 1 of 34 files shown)


internal/interceptor.go
```diff
@@ -419,6 +419,7 @@ type ClientOutboundInterceptor interface {
 	DescribeWorkflow(context.Context, *ClientDescribeWorkflowInput) (*ClientDescribeWorkflowOutput, error)
 
 	// ExecuteActivity intercepts client.Client.ExecuteActivity.
+	// interceptor.Header will return a non-nil map for this context.
```
cconstable, line 429: consistent with other interceptor methods

internal/client_testsuite.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// TestClientEnvironment is the environment used to test client-side
+	// operations. It simulates the client interceptor chain without requiring a
+	// live Temporal server.ew3
+	//
+	// Exposed as: [go.temporal.io/sdk/testsuite.TestClientEnvironment]
+	TestClientEnvironment struct {
+		interceptor ClientOutboundInterceptor
+	}
+)
```
cconstable: I followed the same pattern as `workflow_testsuite.go`

internal/cmd/tools/doclink/doclink.go
```diff
@@ -390,7 +390,25 @@ func processInternal(cfg config, file *os.File, pairs map[string]map[string]stri
...
+			funcSpaces = -1
+			inFunc = false
+		}
+		if strings.HasSuffix(trimmedLine, "interface {") {
+			interfaceSpaces = indentSize
+			inInterface = true
+		} else if inInterface && trimmedLine == "}" && interfaceSpaces == indentSize {
+			interfaceSpaces = -1
+			inInterface = false
+		}
```
cconstable, line 414: My initial commits failed the doclink check because doclink was erroneously interpreting the first line of a top-level function as a top-level line.

Without the fix:
```
go run ./internal/cmd/tools/doclink/doclink.go
Missing doc in internal/client_testsuite.go for internal:ClientOutboundInterceptor to interceptor:ClientOutboundInterceptor
2026/04/24 17:28:05 Changes needed, see previous stdout for which objects. Re-run command with -fix to auto-generate new docs.

func (s *ClientTestSuite) NewTestClientEnvironment(interceptors ...ClientInterceptor) *TestClientEnvironment {
	var chain ClientOutboundInterceptor = &noopClientOutboundInterceptor{}
              ^^^^^^^^^^^^^^^^^^^^^^^^^
```
and running `go run ./internal/cmd/tools/doclink/doclink.go -fix` would add this:
```
func (s *ClientTestSuite) NewTestClientEnvironment(interceptors ...ClientInterceptor) *TestClientEnvironment {
	//
	// Exposed as: [go.temporal.io/sdk/interceptor.ClientOutboundInterceptor]
	var chain ClientOutboundInterceptor = &noopClientOutboundInterceptor{}
```
which is not correct. It read `var chain ClientOutboundInterceptor` as a top-level line.

testsuite/testsuite.go
```diff
@@ -21,6 +21,12 @@ type (
 
 	// TestUpdateCallback is a basic implementation of the UpdateCallbacks interface for testing purposes.
 	TestUpdateCallback = internal.TestUpdateCallback
+
+	// ClientTestSuite is the test suite for testing client-side operations.
+	ClientTestSuite = internal.ClientTestSuite
+
+	// TestClientEnvironment is the environment used to test client-side operations.
+	TestClientEnvironment = internal.TestClientEnvironment
```
Quinn-With-Two-Ns: Why are we adding a new test suite? That seems unrelated to adding otel support and IMO not something we want to add to our public API

interceptor/tracing_interceptor.go
```diff
@@ -394,6 +394,85 @@ func (t *tracingClientOutboundInterceptor) UpdateWithStartWorkflow(
...
+	}
+	var finishOpts TracerFinishSpanOptions
+	defer span.Finish(&finishOpts)
+
+	err = t.Next.CancelActivity(ctx, in)
+	finishOpts.Error = err
+	return err
+}
+
+func (t *tracingClientOutboundInterceptor) TerminateActivity(ctx context.Context, in *ClientTerminateActivityInput) error {
```
Quinn-With-Two-Ns: We generally don't add spans for `Terminate` or `Describe`. We are generally just spans for calls calls that `Client -> Temporal -> Worker` or `Worker -> Temporal -> Worker` . You'll notice other calls like `DescribeWorkflow` are not here and I think we should be consistent here

testsuite/testsuite.go
Quinn-With-Two-Ns: I don't think any other SDK has a `ClientEnvironment` so if it something we want to add it should have cross SDK design discussion.

contrib/opentelemetry/tracing_interceptor_test.go
```diff
@@ -35,6 +35,18 @@ func TestSpanPropagation(t *testing.T) {
 	interceptortest.AssertSpanPropagation(t, testTracer)
 }
 
+func TestClientSAActivitySpanPropagation(t *testing.T) {
```
Quinn-With-Two-Ns: I would recommend following the pattern of other methods, `SAA` doesn't need it own unique pattern for testing

Quinn-With-Two-Ns: See `TestSpanKind` and the real integration tests in `TestOpenTelemetryTracing`

cconstable: I might be missing something obvious but `TestSpanKind` just tests span kinds (e.g. client, server) which doesn't seem relevant to this work and the `TestOpenTelemetryTracing` test is doing e2e testing which I don't want. I just want to test that the OTel interceptor is creating a span for outbound client execute activity operations. Additionally, the `TestOpenTelemetryTracing` is being skipped because it is flakey. I didn't think using either of those was appropriate. 

```
func (ts *IntegrationTestSuite) TestOpenTelemetryTracing() {
	ts.T().Skip("issue-1650: Otel Tracing intergation tests are flaky")
```

interceptor/tracing_interceptor.go
cconstable: I previously added tracing for all SAA operations (except "list activities") to the other SDKs I made PRs for:
- https://github.com/temporalio/sdk-python/pull/1471
- https://github.com/temporalio/sdk-dotnet/pull/647

Should those be removed?

In the future, where can I look to see what should or should not be included? fwiw, I can imagine scenarios where `Terminate` and `Describe` could be useful.

contrib/opentelemetry/tracing_interceptor_test.go
```diff
@@ -35,6 +37,34 @@ func TestSpanPropagation(t *testing.T) {
 	interceptortest.AssertSpanPropagation(t, testTracer)
 }
 
+func TestStandaloneActivitySpanCreation(t *testing.T) {
```
cconstable: I originally named this `TestStandaloneActivitySpanPropagation` but I realized there is another test, `TestSpanPropagation`, that actually tests the propagation. We just care about the span creation here: an SAA should create a span with the correct name and `temporalActivityID`.

contrib/opentelemetry/tracing_interceptor_test.go
```diff
@@ -35,6 +37,34 @@ func TestSpanPropagation(t *testing.T) {
...
+	rec := tracetest.NewSpanRecorder()
+	tracer, err := opentelemetry.NewTracer(opentelemetry.TracerOptions{
+		Tracer: sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(rec)).Tracer(""),
+	})
+	require.NoError(t, err)
+
+	ctx := internal.NewHeaderContext(context.Background())
```
cconstable: The interceptor below writes to a header map expected to be on the context. If the map is missing it panics. To work around this, I went through a number of options:

1. Add internal.NewHeaderContext wrapper around contextWithNewHeader. Additive change, reachable from the contrib package. This is what I did. I think it's simple.
2. Rename contextWithNewHeader to ContextWithNewHeader. Requires touching every internal call site and didn't feel right.
3. Promote it to "public" api as interceptor.HeaderContext(ctx). Don't like exposing test scaffolding like this.
4. Make writeSpanToHeader a noop for null headers. Don't like silently failing. Seems bad.
5. Spin up a real client. Too heavy, I just want to unit test span creation.

testsuite/testsuite.go
cconstable: I mostly did this to hide some internal go-specific things (e.g. panicking if we attempt to write to null context header) and chose that name to stay in line with how workflows were providing a testing environment for their tests. I've swapped approaches and removed the client test suite.

interceptor/tracing_interceptor.go
cconstable: We had some discussion about this and I'll be pushing updates to make sure only start activity creates spans for now across the sdks to keep things consistent.

contrib/opentelemetry/tracing_interceptor_test.go
Quinn-With-Two-Ns: I think you do want to do an e2e test, that is more realistic and closer to what real users would do. The `TestOpenTelemetryTracing` are flaky, but I believe that is just because of they use `query` so a new test that focus on SAA shouldn't have any issues.

internal/cmd/tools/doclink/doclink.go
yuandrew: Thanks for the fix here

internal/interceptor_header.go
```diff
@@ -23,6 +23,10 @@ func contextWithNewHeader(ctx context.Context) context.Context {
 	return context.WithValue(ctx, headerKey{}, map[string]*commonpb.Payload{})
 }
 
+func NewHeaderContext(ctx context.Context) context.Context {
```
yuandrew: Could this live in internal/interceptortest instead? The only caller appears to be the OTel unit test, and `tracing_interceptor_test.go` already imports interceptortest, so this avoids exporting a test-only helper from the broad internal package.

interceptor/tracing_interceptor.go
yuandrew: Looks like these were removed from this PR, and discussion is happening offline with the other SDKs

yuandrew (commented):
overall looks good, but I'd still like us to have some E2E integration test coverage. I think the easiest way would be to either write a specific new test that mirrors `TestInterceptorStandaloneActivity`, or add otel specific asserts to `TestInterceptorStandaloneActivity` itself. In either case, you'll want to add the test name to https://github.com/temporalio/sdk-go/blob/ad514adf9cc39b7304403c894cb3526e605a6d8d/test/integration_test.go#L152-L154 to record spans for the test

test/integration_test.go
```diff
@@ -3053,6 +3054,52 @@ func (ts *IntegrationTestSuite) TestInterceptorStandaloneActivity() {
...
+	spans := ts.openTelemetrySpanRecorder.Ended()
+
+	var startActivitySpan sdktrace.ReadOnlySpan
+	for _, s := range spans {
+		if s.Name() == "StartActivity:tracingTestStandaloneActivity" {
+			startActivitySpan = s
+			break
+		}
+	}
+	ts.Require().NotNil(startActivitySpan, "expected a StartActivity span for the standalone activity")
```
yuandrew, line 3102: Can we also add an assert for the worker-side activity span? This will help catch regressions with request header propogation. 

Something like 

```
ts.Equal("RunActivity:tracingTestStandaloneActivity", runActivitySpan.Name())
ts.Equal(startActivitySpan.SpanContext().SpanID(), runActivitySpan.Parent().SpanID())
ts.Equal(startActivitySpan.SpanContext().TraceID(), runActivitySpan.SpanContext().TraceID())
```

yuandrew: And for assertion robustness, could we also assert the span name for `startActivitySpan`? `ts.Equal("StartActivity:tracingTestStandaloneActivity", startActivitySpan.Name())`

yuandrew (approved):

PR #2309, opened 2026-04-26
feat: expose `ExecuteNexusOperationParams` for external SDKs

xepozz:
## Why?

- Exposed internal apis for nexus and [roadrunner-temporal](https://github.com/temporalio/roadrunner-temporal) integration 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/internal_worker_base.go
```diff
@@ -65,13 +65,37 @@ type (
...
 	}
+)
+
+// NewExecuteNexusOperationParams builds a parameters struct for
+// WorkflowEnvironment.ExecuteNexusOperation. Exposed so that non-Go SDKs
+// (e.g. roadrunner-temporal proxying for PHP) can populate the struct from
+// outside the `internal` package — the fields themselves stay unexported to
+// keep the SDK free to evolve them.
+func NewExecuteNexusOperationParams(
+	client NexusClient,
```
yuandrew: I think we want these in a `NewExecuteNexusOperationParamsOptions` struct, that way if we ever need to add a new field, that would be a breaking change for all callers.

yuandrew (approved):

PR #2311, opened 2026-04-26
:boom: Use ExternalStorageReference protobuf message for payload references

jmaeagle99:
## What was changed

- Store and retrieve payload references as ExternalStorageReference protobuf message encoded with protojson.
- Allow backward compatible retrieval of prerelease plain json encoding.
- Embed ExternalStorageReference proto message definition in sdk-go until api-go contains the same definition. Generate structure and data such that wire format and data are exactly the same as would be in api-go.

### :boom: Breaking Change

Workflow history that contains an external storage reference that is encoded using the proto-JSON converter will likely not be runnable on workers using prior versions of the SDK due to this format change.

## Why?

- Standardize on protobuf message for encoding storage references.

## Checklist

1. How was this tested: Existing tests, compat tests, and new tests


yuandrew:
For reference, this is the API PR: https://github.com/temporalio/api/pull/772

internal/extstore/internal_extstore.go
```diff
@@ -112,23 +122,46 @@ func StorageTargetFromContext(ctx context.Context) StorageDriverTargetInfo {
...
+const metadataEncodingProtoJSON = "json/protobuf"
+
+// metadataEncodingStorageRefLegacy is the encoding written by earlier prerelease
+// SDK versions. Retained solely for backward-compatible reads.
+const metadataEncodingStorageRefLegacy = "json/external-storage-reference"
 
-type storageReference struct {
+// legacyStorageReference is the old wire format retained for backward compatibility
+// with payloads written by earlier prerelease SDK versions.
+type legacyStorageReference struct {
```
yuandrew, line 138: Why do we want/need to maintain the earlier prerelease SDK version? Since it's still in prerelease, should we just make the breaking change here, so we only need to support one format?

internal/extstore/internal_extstore.go
```diff
@@ -112,23 +122,46 @@ func StorageTargetFromContext(ctx context.Context) StorageDriverTargetInfo {
 // format. Mirrors converter.MetadataEncoding without importing converter package.
 const metadataEncoding = "encoding"
 
-// metadataEncodingStorageRef is the metadata encoding value used to identify
-// payloads that are storage references rather than actual data.
-const metadataEncodingStorageRef = "json/external-storage-reference"
+// metadataMessageType is the key used in payload metadata to identify the proto
+// message type. Mirrors converter.MetadataMessageType without importing converter package.
```
yuandrew, line 125: nit: Do we want to add a mirroring comment to `converter.MetadataMessageType` that it matches this new value? Same question for metadataEncodingProtoJSON

internal/extstore/extstore_test.go
```diff
@@ -851,12 +970,50 @@ func TestClaimDeserialization(t *testing.T) {
...
+	},
+	"data": "eyJjbGFpbURhdGEiOnsiYnVja2V0IjoidGVzdC1idWNrZXQiLCJoYXNoX2FsZ29yaXRobSI6InNoYTI1NiIsImhhc2hfdmFsdWUiOiI2Y2EyMmMzNDU2MGNmMzVhYzI0NDI3ZGM3NjE5YzlhYjQ3MmE4MmNmMThmMjg2ZjI3ODcxNjQ5YTJiNTYwOGM4Iiwia2V5IjoidjAvbnMvZGVmYXVsdC93dC9MYXJnZUlPV29ya2Zsb3cvd2kvZjFkMmE0YWMtZjhjYi00NWQzLTkwOGMtOTNhMGYzM2FiMjQ1L3JpL251bGwvZC9zaGEyNTYvNmNhMjJjMzQ1NjBjZjM1YWMyNDQyN2RjNzYxOWM5YWI0NzJhODJjZjE4ZjI4NmYyNzg3MTY0OWEyYjU2MDhjOCJ9LCJkcml2ZXJOYW1lIjoiYXdzLnMzZHJpdmVyIn0=",
+	"externalPayloads": [
+		{
+			"sizeBytes": "385"
+		}
+	]
+}`
+
+	externalStorageReferenceMessageType = string((*sdkpb.ExternalStorageReference)(nil).ProtoReflect().Descriptor().FullName())
```
yuandrew: Is this line needed? Don't we already set this value in internal_extstore.go?

internal/extstore/internal_extstore.go
```diff
@@ -22,10 +24,18 @@ type StorageParameters struct {
 	payloadSizeThreshold int
 }
 
-// IsStorageReference reports whether p is an external-storage reference payload
-// (i.e. its encoding metadata equals the internal storage-reference marker).
+// IsStorageReference reports whether p is an external-storage reference payload.
+// It recognises both the current protojson format (encoding=json/protobuf,
```
yuandrew: nit: `recognises` is the british form of recognizes? Not sure we use British spelling elsewhere in the SDK. I know we use Canadian forms of words (i.e. Behaviour), but not aware of any british spelling, haha
```suggestion
// It recognizes both the current protojson format (encoding=json/protobuf,
```

internal/extstore/internal_extstore.go
yuandrew: I guess we maybe don't want to completely break previous workflows, but do we plan on deprecating this legacy stuff before we hit public preview/GA?

internal/temporalapi/README.md
```diff
@@ -0,0 +1,39 @@
+# internal/temporalapi
```
yuandrew, line 1: Any reason we need this over waiting for an API release to ingest the changes from there?

internal/extstore/internal_extstore.go
```diff
@@ -112,23 +122,46 @@ func StorageTargetFromContext(ctx context.Context) StorageDriverTargetInfo {
...
-// metadataEncodingStorageRef is the metadata encoding value used to identify
-// payloads that are storage references rather than actual data.
-const metadataEncodingStorageRef = "json/external-storage-reference"
+// metadataMessageType is the key used in payload metadata to identify the proto
+// message type. Mirrors converter.MetadataMessageType without importing converter package.
+const metadataMessageType = "messageType"
+
+// metadataEncodingProtoJSON is the standard protojson encoding value, shared with
+// ProtoJSONPayloadConverter. Mirrors converter.MetadataEncodingProtoJSON.
+const metadataEncodingProtoJSON = "json/protobuf"
```
yuandrew, line 130: technically this is breaking if we use an old SDK to read the new deployment format for a rolling deployment. A safer pattern would be to introduce the new format but continue to use the old, then a version later switch to the new format by default. But I assume we don't have time for this multi-release rollout. I'm fine without it, but not sure if we want to add a 💥 . It's a bit of an edge case

internal/extstore/internal_extstore.go
```diff
@@ -112,23 +122,46 @@ func StorageTargetFromContext(ctx context.Context) StorageDriverTargetInfo {
...
+var (
+	// compile-time assertion that ExternalStorageReference implements proto.Message.
+	_ proto.Message = (*sdkpb.ExternalStorageReference)(nil)
+
+	// externalStorageReferenceMessageType is the fully-qualified proto message name,
+	// derived from the descriptor so it stays in sync with the generated code.
+	externalStorageReferenceMessageType = string((*sdkpb.ExternalStorageReference)(nil).ProtoReflect().Descriptor().FullName())
+
+	protoMarshalOptions   = protojson.MarshalOptions{}
+	protoUnmarshalOptions = protojson.UnmarshalOptions{}
```
yuandrew: Do we want to have `DiscardUnknown: true`? For if an older SDK version tries to read a payload from a new SDK that added a new field to `ExternalStorageReference`

https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#UnmarshalOptions

internal/extstore/extstore_test.go
```diff
@@ -851,12 +970,50 @@ func TestClaimDeserialization(t *testing.T) {
...
+}
+
+// ---------------------------------------------------------------------------
+// Claim Compatibility: current-format fixed claim JSON produced by another SDK
+// ---------------------------------------------------------------------------
+
+// TestClaimDeserialization_OtherSdk_ProtoJson verifies that a full proto-JSON
+// payload produced by another language SDK (e.g. Python) is correctly parsed by
+// the Go SDK's payloadToStorageReference function.
+func TestClaimDeserialization_OtherSdk_ProtoJson(t *testing.T) {
```
yuandrew: JSON in `ProtoJson` should be all caps, same with `PlainJson`

internal/extstore/internal_extstore.go
jmaeagle99: We don't want to immediately break anyone who used the prerelease bits. We'll remove before GA.

internal/extstore/internal_extstore.go
jmaeagle99: Added boom to title

internal/extstore/internal_extstore.go
jmaeagle99: Wouldn't we have had to add that option to the plain JSON converter in the old SDK though to work with the scenario that you outlined?

internal/extstore/internal_extstore.go
jmaeagle99: Would it be better to move these to some internal package and re-export them so we don't have to synchronize? Not sure we want this type of comment to show in public docs.

internal/extstore/internal_extstore.go
yuandrew: It's similar to the scenario I mentioned above, but I was mainly thinking about in the future. i.e. if we add something like `string region = 3;`, a similar forward-compatibility concern of if an older SDK can unmarshal a payload from a newer SDK

internal/temporalapi/README.md
jmaeagle99: Discussed offline, will use this temporary solution and adopt API release when it becomes available.

internal/extstore/internal_extstore.go
yuandrew: Good point, and no it's fine, converter doesn't rely on the `internal` package today, not worth adding a dep just for a string or 2

yuandrew (approved):

PR #2312, opened 2026-04-27
Payload visit activity heartbeat

jmaeagle99:
## What was changed

Updated activity heartbeat to have the outbound payload visitor applied. The activity is failed (proactively sent failure to the server) and the activity context is cancelled.

## Why?

Consistency of application of payload visitor to all protos that have payloads.

## Checklist

1. Closes #2271
1. How was this tested: Existing tests and new tests

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -2097,6 +2097,14 @@ func newActivityTaskHandlerWithCustomProvider(
 	}
 }
 
+// heartbeatVisitorError wraps an outbound payload visitor error from a heartbeat.
+// It is used as the context cancellation cause so Execute() can detect that
+// RespondActivityTaskFailed was already sent proactively and skip sending a second response.
+type heartbeatVisitorError struct{ err error }
+
+func (e heartbeatVisitorError) Error() string { return e.err.Error() }
+func (e heartbeatVisitorError) Unwrap() error { return e.err }
+
 type temporalInvoker struct {
 	sync.Mutex
 	identity       string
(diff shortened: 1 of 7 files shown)


internal/internal_task_handlers.go
```diff
@@ -2183,7 +2193,31 @@ func (i *temporalInvoker) internalHeartBeat(ctx context.Context, details *common
...
+		TaskToken: i.taskToken,
+		Details:   details,
+		Identity:  i.identity,
+		Namespace: i.namespace,
+	}
+	var err error
+	if visitErr := visitProtoPayloads(ctx, i.outboundPayloadVisitor, request, 0); visitErr != nil {
+		// Proactively fail the task so the server can retry immediately rather than
+		// waiting for the heartbeat timeout. Errors are ignored — if the RPC fails the
+		// activity will still be timed out by the server.
```
yuandrew, line 2206: Are there any additional logs/diagnostics we should add if the RPC fails? I wonder if users would left a little directionless if they see a timeout err and not a visit error.

internal/internal_worker.go
```diff
@@ -1147,14 +1147,6 @@ func getDataConverterFromActivityCtx(ctx context.Context) converter.DataConverte
 	return WithContext(ctx, dataConverter)
 }
```
yuandrew, line 1149: nice

internal/internal_task_handlers.go
yuandrew: actually on second thought, i think this is fine

yuandrew (approved):

PR #2315, opened 2026-04-29
Use protojson marshaling for payload HTTP handler response

jmaeagle99:
## What was changed

Update the `payloadHTTPHandler` to write response as Protobuf JSON (according to [codec server spec](https://github.com/temporalio/samples-go/blob/main/codec-server/README.md#codec-server-protocol)) instead of as plain JSON.

## Why?

The [codec server spec](https://github.com/temporalio/samples-go/blob/main/codec-server/README.md#codec-server-protocol) says that requests and responses should be interpreted with Protobuf JSON encoders.

The main observable difference is that plain JSON encoding would use snake_case for field names whereas Protobuf JSON uses camelCase for field names. This hasn't been an issue in the past because `Payloads` and `Payload` messages have property names that were only one word, thus the same in camelCase and snake_case.

With the introduction of the `external_payloads` field on the `Payload` protobuf message and the `size_bytes` field on the `ExternalPayloadDetails` protobuf message, these fields are being written as snake_case when serializing as plain json. The Web UI is expecting camelCase properties.

Brief testing with the Web UI and inspection of the code shows that this works. The Temporal CLI is [already using NewRemotePayloadCodec](https://github.com/temporalio/cli/blob/main/cliext/client.go#L290) from sdk-go, which is using [protojson.Unmarshal](https://github.com/temporalio/sdk-go/blob/main/converter/codec.go#L403) on the response.

## Checklist

1. How was this tested:
- [x] Tested with existing WebUI.
- [x] Inspect Temporal CLI for conformance.


converter/payload_handler.go
```diff
@@ -131,8 +130,13 @@ func (h *payloadHTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	bs, err = protojson.Marshal(&commonpb.Payloads{Payloads: payloads})
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
 	w.Header().Set("Content-Type", "application/json")
-	if err = json.NewEncoder(w).Encode(commonpb.Payloads{Payloads: payloads}); err != nil {
+	if _, err = w.Write(bs); err != nil {
```
yuandrew, line 139: I love writing some BS 😆

yuandrew (approved):

PR #2319, opened 2026-04-30
Add `CONTINUE_AS_NEW_VERSIONING_BEHAVIOR_USE_RAMPING_VERSION`

THardy98:
## What was changed
DWISOTT - adds `UseRampingVersion` as a CaN versioning option

## Why?
support new versioning option

1. Part of https://github.com/temporalio/features/issues/807

2. How was this tested:
Added integration tests

3. Any docs updates needed?
Maybe ?

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -95,9 +95,10 @@ const (
 	ContinueAsNewVersioningBehaviorAutoUpgrade = 1
 
 	// ContinueAsNewVersioningBehaviorUseRampingVersion - Use the Ramping Version of the workflow's task queue at start time,
-	// regardless of the workflow's Target Version (according to f(workflow_id, ramp_percentage)). After the first workflow
-	// task completes, the workflow will use whatever Versioning Behavior it is annotated with. If there is no Ramping
-	// Version by the time that the first workflow task is dispatched, it will be sent to the Current Version.
+	// regardless of the workflow's Target Version.
+	//
+	// After the first workflow task completes, the workflow will use whatever Versioning Behavior it is annotated with. If
+	// there is no Ramping Version by the time that the first workflow task is dispatched, it will be sent to the Current Version.
 	//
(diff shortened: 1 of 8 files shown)


internal/workflow.go
```diff
@@ -93,6 +93,23 @@ const (
 	//
 	// Exposed as: [go.temporal.io/sdk/workflow.ContinueAsNewVersioningBehaviorAutoUpgrade]
 	ContinueAsNewVersioningBehaviorAutoUpgrade = 1
+
+	// ContinueAsNewVersioningBehaviorUseRampingVersion - Use the Ramping Version of the workflow's task queue at start time,
+	// regardless of the workflow's Target Version (according to f(workflow_id, ramp_percentage)). After the first workflow
```
carlydf: instead of f(workflow_id, ramp_percentage), should we say something along the lines of:

```
  // The Target Version is chosen with the default formula:
  //   if calcRampThreshold(workflow_id) <= ramp_percentage:
  //     target=ramping_version
  //   else:
  //     target=current_version
```

I feel like `f(workflow_id, ramp_percentage)` will invite lots of questions.
I am ok being transparent about this formula -- we won't change it by default after GA because that could cause workflows to "roll back" accidentally, even with same ramp percentage

carlydf: calcRampThreshold is the name of the server function we actually use, so people (or claude) can look it up if they're interested

THardy98: I'm going to remove it entirely, I don't think we need to leak server versioning internals here (I also don't anticipate people finding this particularly necessary)

carlydf: We define Ramping, Current, and Target version here fwiw https://docs.temporal.io/worker-versioning#versioning-definitions fwiw, so people can look those up there

carlydf (approved):

yuandrew (approved):

Sushisource (approved):

PR #2232, opened 2026-03-15
Add missing documentation for exposed structs in internal package

brucearctor:
Fixes #2177

PR #2324, opened 2026-05-05
Add doc comments to new struct fields for serialization context

yuandrew:
## What was changed

Add comments for each field in the new structs I added in https://github.com/temporalio/sdk-go/pull/2225.

## Why?

New changes from the PR mentioned above is causing CI in https://github.com/temporalio/sdk-go/pull/2232 to fail, it's probably easiest if I just create this PR instead of suggesting changes, as these are all with new structs I added recently

---

> [!NOTE]
> **Low Risk**
> Documentation-only updates to exported option structs; no functional or behavioral changes expected.
> 
> **Overview**
> Adds detailed godoc to the activity completion/heartbeat option structs used by `CompleteActivity*WithOptions` and `RecordActivityHeartbeat*WithOptions`, documenting required vs optional fields and how each field affects serialization/external-storage context.
> 
> Also adds prominent warnings in the public `client` type aliases that serialization-context values (e.g., `ActivityType`, `WorkflowType`, `TaskQueue`, `WorkflowID`) are **not validated by the SDK** and incorrect values may break codecs that depend on them.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3a86f94581be346e6058f7e9770e252b8a12022. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/client.go
```diff
@@ -559,106 +559,224 @@ type (
...
+
+		// RunID is the Run ID of the Workflow Execution that scheduled the Activity.
+		// Leave empty to target the latest run for WorkflowID.
+		RunID string
+
+		// ActivityID is the ID of the Activity to complete.
+		// This field is required.
+		ActivityID string
+
+		// Result is the Activity result to report on successful completion.
```
jmaeagle99: nit: The wording makes it sound like "if the activity is successfully completed, then use this result". You might want to update the wording to (1) indicate that setting this field will complete the activity, and (2) is mutually exclusive with the `Err` field.

Whatever is done, do the same for the other `Result` fields on the other completion options structs.

internal/client.go
```diff
@@ -559,106 +559,224 @@ type (
...
+		Result interface{}
+
+		// Err is the Activity error to report on failed completion.
+		Err error
```
jmaeagle99, line 619: Similar to https://github.com/temporalio/sdk-go/pull/2324/changes#r3220298604. Additionally, what happens when both fields are set? Maybe call out that it is undefined behavior to do so if we aren't explicitly handling it in the SDK.

Whatever is done, do the same for the other `Err` fields on the other completion options structs.

internal/client.go
```diff
@@ -559,106 +559,224 @@ type (
...
 	CompleteActivityByIDOptions struct {
-		Namespace  string // required
-		WorkflowID string // required
-		RunID      string
-		ActivityID string // required
-		Result     interface{}
-		Err        error
+		// Namespace is the Temporal Namespace containing the Workflow Execution.
+		// This field is required.
+		Namespace string
```
jmaeagle99, line 597: Why do we allow setting the namespace different from the one to which the client is bound?

internal/client.go
```diff
@@ -559,106 +559,224 @@ type (
...
 		// for encoding.
-		//
-		// These values are not validated by the SDK. Providing incorrect values
-		// may cause serialization/deserialization mismatches if your codec uses
-		// them (e.g., as encryption keys or signature input).
-		Namespace    string
-		WorkflowID   string
+
+		// Namespace is the optional Temporal Namespace to include in the Activity
+		// serialization context. If unset, the Client's Namespace is used.
```
jmaeagle99: > If unset, the Client's Namespace is used.

Is this also true of the other `Namespace` fields on the other completion option structs?

internal/client.go
```diff
@@ -559,106 +559,224 @@ type (
...
+
+		// RunID is the Run ID of the Workflow Execution that scheduled the Activity.
+		// Leave empty to target the latest run for WorkflowID.
+		RunID string
+
+		// ActivityID is the ID of the Activity to heartbeat.
+		// This field is required.
+		ActivityID string
+
+		// Details are the Activity heartbeat details to report.
```
jmaeagle99, line 667:
```suggestion
		// Details are the Activity heartbeat details to report.
		// These values are encoded with the Client's DataConverter.
```
to be consistent with the other `Details` fields. Also, if we are calling this out specifically, should we call out that they may be externally stored via ExternalStorage as well?

internal/client.go
yuandrew: Good question, had to dig into this

It looks like only for `CompleteActivityOptions` should the namespace match the client namespace, since that is task-token based, which always uses the client namespace, whereas all of the others can target other namespaces.

internal/client.go
yuandrew: yes, made them all consistent and added blurb that external storage can be used

internal/client.go
yuandrew: https://github.com/temporalio/sdk-go/blob/main/internal/internal_task_pollers.go#L1597

Looks like if `Err` exists, it takes precedence over `Result`. Will update the docs to say the same

internal/client.go
yuandrew: only RecordActivityHeartbeatOptions.Namespace and RecordActivityHeartbeatByIDOptions.Namespace follow this, CompleteActivityByIDOptions.Namespace and CompleteActivityByActivityIDOptions.Namespace both have these as required fields

internal/client.go
jmaeagle99: >  whereas all of the others can target other namespaces.

I recall that server deprecated cross-namespace workflow commands (see https://github.com/temporalio/sdk-go/pull/2267). I know that this is change is on the client, so it's different. Does enable some hypothetical client that is not attached to a specific namespace to send async completions? If a client is always attached to one namespace, I don't think we should add these namespace fields.

jmaeagle99 (approved):

internal/client.go
yuandrew: Updated comments to not encourage cross-namespace things

PR #2325, opened 2026-05-06
Add memo update support for schedules

chaptersix:
## Summary
- Add `Memo` field to `ScheduleUpdate` to allow updating schedule-level memo via `UpdateScheduleRequest`
- Uses pointer-to-map semantics: `nil` = don't update, non-nil empty map = clear, non-nil with entries = replace
- Only supported on CHASM-backed schedules; workflow-backed schedules return a `FailedPrecondition` error
- Enable `frontend.allowedExperiments` config in dev server and CI so integration tests can create CHASM schedules via the `temporal-experiment` gRPC header

Depends on API change: https://github.com/temporalio/api/pull/748


internal/internal_schedule_client.go
```diff
@@ -305,6 +321,7 @@ func (scheduleHandle *scheduleHandleImpl) Update(ctx context.Context, options Sc
 		Identity:         scheduleHandle.client.identity,
 		RequestId:        uuid.NewString(),
 		SearchAttributes: newSA,
+		Memo:             newMemo,
```
yuandrew, line 324: Should schedule-level memo update support the same `*commonpb.Payload` pass-through behavior as [`ScheduleWorkflowAction.Memo` updates](https://github.com/temporalio/sdk-go/blob/8b9d2c2589cd0466c73089eb9e71c2e8491ff2e9/internal/internal_schedule_client.go#L922-L924)?

This is a full-replacement field, but users may build that replacement from `input.Description.Memo.Fields` to preserve existing keys while changing one value. Those existing values are already `*commonpb.Payload`s. `encodeScheduleWorkflowMemo` preserves those payloads during workflow-action memo updates, but this new path uses `getWorkflowMemo`, which re-encodes them as Go protobuf objects.

If we intend for `ScheduleUpdate.Memo` to only accept raw decoded Go values, then this may just need documentation. Otherwise, it would be worth sharing the existing pass-through helper behavior here too

internal/schedule_client.go
```diff
@@ -540,6 +540,14 @@ type (
 		// attributes present: replace any and all pre-existing assigned search attributes with the defined search
 		//                     attributes, i.e. upsert
 		TypedSearchAttributes *SearchAttributes
+
+		// Memo - Non-indexed user supplied information to replace on the schedule.
+		// If set, replaces the entire memo. If nil, leaves the existing memo intact.
+		// An initialized but empty map will clear the memo.
```
yuandrew, line 546: nit: for `TypedSearchAttributes *SearchAttributes` right above, their docs use a three-state doc
> // nil: leave any pre-existing assigned search attributes intact
// empty: remove any and all pre-existing assigned search attributes
// attributes present: replace any and all pre-existing assigned search attributes with the defined search
//                     attributes, i.e. upsert

should we follow the same style here, for consistency?

chaptersix:
will circle back to this

PR #2334, opened 2026-05-13
Generic Temporal Nexus Operation Handler

Quinn-With-Two-Ns:
## What was changed

  Added a generic Nexus operation handler that consolidates the common pattern of "either return a synchronous
   result or start a workflow as an async operation." 

## Why?

  The existing `NewWorkflowRunOperation` /` NewSyncOperation` split forces users to pick the operation shape up front, which is awkward when:  
  - The start handler decides at runtime whether to respond sync or async.
  - The handler needs access to the Temporal client.                          
  - The user wants a custom cancel implementation (e.g. terminate instead of cancel).                                                    
                                                                                                                                           
  The new TemporalOperation unifies these cases behind one API and makes the common path much shorter. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> New experimental Nexus/operation surface starts and cancels workflows from handlers; mistakes affect operation lifecycle and cancel behavior, though changes are additive with tests.
> 
> **Overview**
> Adds an **experimental** generic Nexus handler API in `temporalnexus` so one operation can return either a **sync** result or start a **workflow-backed async** run from a single `Start` callback, instead of choosing up front between `NewSyncOperation` and `NewWorkflowRunOperation`.
> 
> **`NewTemporalOperation` / `MustNewTemporalOperation`** build a `nexus.Operation` with validation (required name/start, reserved `__temporal_` prefix) and a default workflow-run cancel via `CancelWorkflow`. **`TemporalOperationResult`** plus `NewSyncResult` / `NewAsyncResult` map to Nexus sync vs async start responses. **`NexusClient`** scopes the Temporal client per invocation; **`StartWorkflow`** / **`StartUntypedWorkflow`** wire callbacks/links/request ID and enforce **one async start per invocation**.
> 
> **Cancel** decodes tokens with new **`loadTokenType`** in `token.go` before full workflow-token parsing. Unit tests cover result conversion, validation, token parsing, and the double-start guard; integration tests register a Nexus service and exercise sync, typed/untyped async, default cancel, custom terminate cancel, and `GetWorkflowClient` in `Start`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86be12d18ced4eec5b2a9feeb0d04839f2c9aa16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/temporalnexus/temporal_operation.go
+++ b/temporalnexus/temporal_operation.go
@@ -91,8 +91,6 @@ func (r TemporalOperationResult[R]) toHandlerResult() (nexus.HandlerStartOperati
 // NOTE: Experimental
 type NexusClient struct {
 	client                client.Client
-	namespace             string
-	taskQueue             string
 	startOperationOptions nexus.StartOperationOptions
 	asyncStarted          *atomic.Bool
 }
@@ -250,15 +248,12 @@ func (o *temporalOperation[I, O]) Start(
 	// Prevent the test env client from panicking.
 	ctx = context.WithValue(ctx, internal.IsWorkflowRunOpContextKey, true)
 
-	nctx, ok := internal.NexusOperationContextFromGoContext(ctx)
-	if !ok {
+	if _, ok := internal.NexusOperationContextFromGoContext(ctx); !ok {
 		return nil, nexus.NewHandlerErrorf(nexus.HandlerErrorTypeInternal, "internal error")
 	}
 
 	nc := NexusClient{
 		client:                GetClient(ctx),
-		namespace:             nctx.Namespace,
-		taskQueue:             nctx.TaskQueue,
 		startOperationOptions: options,
 		asyncStarted:          &atomic.Bool{},
 	}
--- a/temporalnexus/temporal_operation.go
+++ b/temporalnexus/temporal_operation.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"strings"
+	"sync/atomic"
 
 	"github.com/nexus-rpc/sdk-go/nexus"
 	"go.temporal.io/sdk/client"
@@ -54,7 +55,7 @@ type NexusClient struct {
 	namespace             string
 	taskQueue             string
 	startOperationOptions nexus.StartOperationOptions
-	asyncStarted          *bool
+	asyncStarted          *atomic.Bool
 }
 
 // GetWorkflowClient returns the underlying Temporal client for advanced or fallback use cases.
@@ -93,7 +94,7 @@ func StartUntypedWorkflow[R any](
 	workflow any,
 	args ...any,
 ) (TemporalOperationResult[R], error) {
-	if nc.asyncStarted != nil && *nc.asyncStarted {
+	if nc.asyncStarted != nil && nc.asyncStarted.Load() {
 		return TemporalOperationResult[R]{}, nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "only one async operation can be started per operation invocation")
 	}
 
@@ -106,7 +107,7 @@ func StartUntypedWorkflow[R any](
 	}
 
 	if nc.asyncStarted != nil {
-		*nc.asyncStarted = true
+		nc.asyncStarted.Store(true)
 	}
 	nexus.AddHandlerLinks(ctx, handle.link())
 	return NewAsyncResult[R](handle.token()), nil
@@ -204,13 +205,12 @@ func (o *temporalOperation[I, O]) Start(
 		return nil, nexus.NewHandlerErrorf(nexus.HandlerErrorTypeInternal, "internal error")
 	}
 
-	started := false
 	nc := NexusClient{
 		client:                GetClient(ctx),
 		namespace:             nctx.Namespace,
 		taskQueue:             nctx.TaskQueue,
 		startOperationOptions: options,
-		asyncStarted:          &started,
+		asyncStarted:          &atomic.Bool{},
 	}
 
 	result, err := o.options.Start(ctx, nc, input, options)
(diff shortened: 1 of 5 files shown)


temporalnexus/temporal_operation.go
```diff
@@ -0,0 +1,263 @@
...
+
+	// Prevent the test env client from panicking when we try to use it from an operation.
+	ctx = context.WithValue(ctx, internal.IsWorkflowRunOpContextKey, true)
+
+	handle, err := ExecuteUntypedWorkflow[R](ctx, nc.startOperationOptions, workflowOpts, workflow, args...)
+	if err != nil {
+		return TemporalOperationResult[R]{}, err
+	}
+
+	if nc.asyncStarted != nil {
```
Quinn-With-Two-Ns: Should use a atomic here

temporalnexus/temporal_operation.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// NexusClient wraps the Temporal client.Client with Nexus-aware context.
+// It is scoped to a single operation invocation and should not be stored beyond the Start callback.
+type NexusClient struct {
+	client                client.Client
+	namespace             string
+	taskQueue             string
+	startOperationOptions nexus.StartOperationOptions
+	asyncStarted          *atomic.Bool
```
VegetarianOrc, line 95: nit: this might be better as a non-pointer since we know we'll want it every time. Would avoid nil checks around existence when checking for the started flag.

temporalnexus/temporal_operation.go
```diff
@@ -0,0 +1,263 @@
...
+// Useful for invoking workflows that don't follow the single argument / single return type signature.
+// See [StartWorkflow] for the type-safe variant.
+func StartUntypedWorkflow[R any](
+	ctx context.Context,
+	nc NexusClient,
+	workflowOpts client.StartWorkflowOptions,
+	workflow any,
+	args ...any,
+) (TemporalOperationResult[R], error) {
+	if nc.asyncStarted != nil && nc.asyncStarted.Load() {
```
VegetarianOrc: related to the above comment about making `asyncStarted`, this check doesn't work if it's somehow nil. Should never be possible but 🤷.

Also I think this needs to be a `CompareAndSwap` to reserve the execution, then in the error scenario set back to `false` to allow for more attempts. Without that, the TOCTOU race still exists.

temporalnexus/temporal_operation_test.go
```diff
@@ -0,0 +1,160 @@
...
+	_, err = loadTokenType(invalidJSON)
+	require.ErrorContains(t, err, "failed to unmarshal operation token")
+
+	// Missing type field (t=0)
+	missingType := base64.URLEncoding.WithPadding(base64.NoPadding).EncodeToString([]byte(`{"ns":"ns"}`))
+	_, err = loadTokenType(missingType)
+	require.ErrorContains(t, err, "missing or zero token type")
+}
+
+func TestDoubleStartGuard(t *testing.T) {
```
VegetarianOrc, line 142: This test confirms that if the atomic is set, it prevents another async op but doesn't confirm the race is avoided.

What do you think of making a small `client.Client` implementation that can be used to block during ExecuteWorkflow? Then we could use two goroutines to confirm that one is rejected while the other is still "in-flight". Maybe heavy handed, but thought I'd mention it!

temporalnexus/temporal_operation.go
```diff
@@ -0,0 +1,263 @@
...
+	// Name of the operation. Required.
+	Name string
+	// Start handles the operation start. It receives the operation context, a [NexusClient] scoped
+	// to this invocation, the operation input, and the Nexus start operation options.
+	// Required.
+	Start func(ctx context.Context, nc NexusClient, input I, options nexus.StartOperationOptions) (TemporalOperationResult[O], error)
+	// CancelWorkflowRun handles cancel for workflow-run operation tokens.
+	// It receives a NexusClient and the workflow ID extracted from the token.
+	// If nil, defaults to cancelling the workflow via the Temporal client.
+	CancelWorkflowRun func(ctx context.Context, nc NexusClient, workflowID string, options nexus.CancelOperationOptions) error
```
VegetarianOrc: I'm wondering if this should receive a `NexusClient` or just `client.Client`. The useful attributes of `NexusClient` (attaching callbacks, etc) seem irrelevant to the cancellation path. I think the best parts are the contextual info like namespace & task queue. I'm unsure if those are accessible in another way or not.

VegetarianOrc (commented):
Overall looks good to me after getting the CompareAndSwap in.

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: >I'm unsure if those are accessible in another way or not

Those are accessible on the `context` through the operation window

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: It is a pointer so if we pass the Client around it shares the same flag

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: I think it is a good point that here we should just use the normal client

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: Yeah it should be compare and swap good catch

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: I also adjusted the API to take an options struct so we can add more fields in the future

temporalnexus/temporal_operation_test.go
Quinn-With-Two-Ns: hmm I thought maybe a bit overkill? But if others think it is worth while I can add

VegetarianOrc (approved):
LGTM! Like the change to the options struct on CancelWorkflowRun.

yuandrew (approved):

PR #2337, opened 2026-05-14
feat: add ExecuteStandaloneActivity to TestWorkflowEnvironment

### LINKED ISSUE #2318: Add support for Standalone Activities to TestWorkflowEnvironment
There is no way to test the full standalone activity client flow (client → interceptor chain → worker) without spinning up a real Temporal server. TestActivityEnvironment.ExecuteActivity runs the activity function directly via taskHandler.Execute() (internal_workflow_testsuite.go:796), bypassing the ClientOutboundInterceptor chain entirely. This means user interceptors (e.g., tracing, metrics) are never invoked during tests.

Add ExecuteStandaloneActivity to TestWorkflowEnvironment that goes through the full client interceptor chain, similar to how SignalWorkflow, QueryWorkflow, and CancelWorkflow already work. It should:
  - Set up a header context via contextWithNewHeader (like WorkflowClient.ExecuteActivity
  does)
  - Call through the ClientOutboundInterceptor chain (so user interceptors like tracing are
  invoked)
  - Execute the activity via the test environment's existing activity task handler
  - Return a ClientActivityHandle supporting Get(), Cancel(), and Terminate()

**Additional context**
This is a follow-up to #2158 (standalone activity support). Once this is implemented, PR #2302 can use the standard interceptortest.RunTestWorkflow pattern and extend existing tests like TestSpanKind, TestSpanPropagation, and TestBenignErrorSpanStatus to cover standalone activities — with no new exports from internal/.

brucearctor:
## What

Adds `ExecuteStandaloneActivity` to `TestWorkflowEnvironment` that executes standalone activities through the full `ClientOutboundInterceptor` chain.

Closes #2318

## Why

Currently, `TestActivityEnvironment.ExecuteActivity()` runs activity functions directly via `taskHandler.Execute()`, completely bypassing the `ClientOutboundInterceptor` chain. This prevents testing of interceptor-dependent features like OpenTelemetry tracing (#2302), metrics, and header propagation without spinning up a real server.

This method fills that gap by matching the real `Client.ExecuteActivity` flow through the interceptor chain while keeping execution local to the test environment.

## How

- **`testClientOutboundInterceptor`**: a terminal `ClientOutboundInterceptor` that replaces gRPC calls with the existing test activity task handler execution
- **`testClientActivityHandleImpl`**: implements `ClientActivityHandle` for the test environment
- **`executeStandaloneActivity`**: builds the interceptor chain from registered `WorkerOptions.Interceptors` (type-asserting `WorkerInterceptor` to `ClientInterceptor`) and routes execution through it

The flow is: `contextWithNewHeader(ctx)` → build interceptor chain → `chain.ExecuteActivity()` → user interceptors → `testClientOutboundInterceptor.ExecuteActivity()` → `taskHandler.Execute()`.

## Design Decisions

There were a few decision points during implementation. I went with what seemed most natural given the existing codebase patterns, but am happy to revisit any of these if the maintainers prefer a different approach:

1. **Method lives on `TestWorkflowEnvironment` (not a new `TestClientEnvironment`)** — The issue requested it on `TestWorkflowEnvironment`, and it keeps the API surface minimal. If a separate `TestClientEnvironment` is preferred for cleaner separation, I can refactor.

2. **Interceptor chain built per-call** — Rather than caching the chain, each call to `ExecuteStandaloneActivity` rebuilds it. This matches how the real client initializes its chain and avoids stale state if `SetWorkerOptions` is called between invocations. If caching is preferred for performance, that would be straightforward to add.

3. **Interceptors sourced from `WorkerOptions.Interceptors`** — Since the test environment uses `SetWorkerOptions` to configure interceptors, I type-assert each `WorkerInterceptor` to `ClientInterceptor` when building the chain. This works for the common case where users pass full `Interceptor` implementations (which embed both). If a separate `SetClientInterceptors` option is preferred, I can add that.

4. **Synchronous execution with result on the handle** — Activities execute synchronously (matching existing test suite behavior), so `handle.Get()` returns immediately. `Cancel`, `Terminate`, and `Describe` return errors indicating they are unsupported in the test environment.

## Tests

Added 6 tests covering:
- Basic execution and result retrieval
- Default options (auto-generated ID, default task queue and timeouts)
- Activity failure propagation
- **Interceptor chain traversal** — verifies user interceptor `ExecuteActivity` is called and headers are present
- Unregistered activity (expects panic, matching existing behavior)
- Custom data converter

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -3944,3 +3944,182 @@ func (res *updateResult) post_callbacks(env *testWorkflowEnvironmentImpl) {
 	}
 	res.callbacks = []updateCallbacksWrapper{}
 }
+
+// testClientOutboundInterceptor is the terminal interceptor in the
+// ClientOutboundInterceptor chain for TestWorkflowEnvironment. Instead of
+// making gRPC calls to a real server, it executes activities directly via the
+// test activity task handler.
+type testClientOutboundInterceptor struct {
+	ClientOutboundInterceptorBase
+	env *testWorkflowEnvironmentImpl
+}
+
+func (t *testClientOutboundInterceptor) ExecuteActivity(
+	ctx context.Context,
+	in *ClientExecuteActivityInput,
+) (ClientActivityHandle, error) {
+	if in.Options == nil {
+		return nil, errors.New("options are required")
+	}
+	if in.Options.ID == "" {
+		in.Options.ID = uuid.NewString()
+	}
+	if in.Options.TaskQueue == "" {
+		in.Options.TaskQueue = defaultTestTaskQueue
+	}
+	if in.Options.ScheduleToCloseTimeout == 0 && in.Options.StartToCloseTimeout == 0 {
+		in.Options.StartToCloseTimeout = 10 * time.Minute
+	}
+
+	dc := t.env.GetDataConverter()
+	input, err := encodeArgs(dc, in.Args)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode activity args: %w", err)
+	}
+
+	scheduleTaskAttr := &commandpb.ScheduleActivityTaskCommandAttributes{
+		ActivityId:             in.Options.ID,
+		ActivityType:           &commonpb.ActivityType{Name: in.ActivityType},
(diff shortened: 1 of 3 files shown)


yuandrew:
Thanks for the contribution, but we've decided the issue itself is not worth the effort and not quite what we want and have closed, so closing this contribution as well

PR #2338, opened 2026-05-14
Add missing Describe() method to external storage s3driver docs

Ozoniuss:
## What was changed
Just added `Describe()` to the docs which was missing. I was reading through the driver code and spotted this. Describe is [a newer addition](https://github.com/temporalio/sdk-go/pull/2316) to the original interface.

## Why?
Keep the `Client` interface in the documentation in sync with the actual definition.


yuandrew (approved):
Thanks!

PR #2340, opened 2026-05-15
feat(otel): add UseMonotonicCounters option for correct Prometheus counter type

### LINKED ISSUE #2140: OpenTelemetry incorrect metric type for Counter
## Expected Behavior

## Actual Behavior

```
func main() {
	if err != nil {
	}

	go func() {
		}
	}()

		}),
	})
	if err != nil {
	}
	defer c.Close()

	if err != nil {
	}

	// Shutdown the MeterProvider
	}
}
```

The rest is the same.

## Specific Problem

```
        ...
}
```

Then it works as expected.

## Specifications

apridosimarmata:
## Problem

The OpenTelemetry metrics handler uses `Int64UpDownCounter` for all counters. This causes Prometheus to report Temporal counters (like `workflow_completed`) as gauges instead of counters with the `_total` suffix, since `Int64UpDownCounter` is non-monotonic.

See #2140 for details.

## Solution

Added a `UseMonotonicCounters` option to `MetricsHandlerOptions`. When set to `true`, the handler creates counters using `Int64Counter` (monotonic) instead of `Int64UpDownCounter`.

The default remains `false` for backward compatibility, as discussed in the issue thread.

## Usage

```go
handler := opentelemetry.NewMetricsHandler(opentelemetry.MetricsHandlerOptions{
    Meter:                meterProvider.Meter("temporal-sdk-go"),
    UseMonotonicCounters: true,
})
```

## Testing

Added `TestMonotonicCounterHandler` that verifies counters are reported with `IsMonotonic: true` when the option is enabled. All existing tests continue to pass.

Fixes #2140

### DIFF
--- a/contrib/opentelemetry/handler.go
+++ b/contrib/opentelemetry/handler.go
@@ -16,9 +16,10 @@ var _ client.MetricsHandler = MetricsHandler{}
 // MetricsHandler is an implementation of client.MetricsHandler
 // for open telemetry.
 type MetricsHandler struct {
-	meter      metric.Meter
-	attributes attribute.Set
-	onError    func(error)
+	meter                metric.Meter
+	attributes           attribute.Set
+	onError              func(error)
+	useMonotonicCounters bool
 }
 
 // MetricsHandlerOptions are options provided to NewMetricsHandler.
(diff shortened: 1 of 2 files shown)


yuandrew:
Thank you for the contribution, we would need you to sign the CLA before reviewing this PR

yuandrew:
This issue was completed

PR #2366, opened 2026-05-24
contrib: add Google Cloud Storage driver for external payload storage

### LINKED ISSUE #2364: Proposal: Google Cloud Storage external storage driver in contrib (mirror of aws/s3driver)
There is an existing [Python proposal](https://github.com/temporalio/sdk-python/issues/1502), this is basically the same one but for the go sdk. The high-level approach is similar to the Python proposal (mirror the existing aws implementation, introduced in this **[PR](https://github.com/temporalio/sdk-go/pull/2257)).** Also, I'd be happy to do the implementation PRs and help maintaining once we agree on the high-level approach.

To re-iterate I was thinking:

- Keep a similar wrapper (module) of the actual google sdk like the [aws driver does](https://github.com/temporalio/sdk-go/tree/main/contrib/aws/s3driver/awssdkv2), to avoid pulling in the google storage dependency if one wants to implement a custom storage driver. The interface will be defined like the [AWS one](https://github.com/temporalio/sdk-go/blob/main/contrib/aws/s3driver/client.go), though note that methods may differ due to differences between vendors (I think this is acceptable, but let me know if you think otherwise)
- The actual driver implementation would probably be more or less as the [existing](https://github.com/temporalio/sdk-go/blob/main/contrib/aws/s3driver/driver.go) driver code, since the only thing that really is different is the interaction with the cloud provider

I did start the driver implementation [here](https://github.com/Ozoniuss/sdk-go/tree/ozoniuss.gcp-driver/contrib/gcp/storagedriver) to understand how it would look like and what the differences were compared to the AWS sdk and I also wanted to experiment a bit with integration tests against a real bucket. Would like to clarify some details before moving forward:

- I did use the [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) for a fake GCS implementation though let me know if you have another preference. Honestly, I've never used a package for mocking external storage before and this seemed to be the most popular. But I did write the tests such that they work with a real bucket as well. See [my implementation](https://github.com/Ozoniuss/sdk-go/blob/ozoniuss.gcp-driver/contrib/gcp/storagedriver/gcpsdk/client_test.go) for more details. I kept it since I was going to test against the real thing anyway and realized I could do it with minimal test structure changes. I think it's probably not worth setting up automated tests against a real bucket, but found it to be a useful addition for me while developing locally. Note that unless the integration tests are configured to run, only the tests against the fake will be picked up.
- There will probably be a lot of duplication in the actual driver implementation (for aws, [driver.go](https://github.com/temporalio/sdk-go/blob/main/contrib/aws/s3driver/driver.go)). How would you prefer to handle this? I think for two drivers code duplication is not too bad, and will also be future-proof in case the client interfaces will diverge in the future between vendors. Though at the same time I also think it may end up being harder to maintain if more drivers are added. I was thinking one approach to handle this duplication could be to extend [ExternalStorage](https://github.com/Ozoniuss/sdk-go/blob/ozoniuss.gcp-driver/internal/extstore/extstore.go#L4) with potentially shared functionality -- for example, `extstore.ExternalStorage` may be configurable to use a custom function for building the object key, which is currently fully handled by the driver. Though, I haven't really explored the code much to understand what changes this would require -- can come back with more thoughts later.

There are some minor differences in terms of the GCS sdk compared to the AWS one but I think it's probably best to discuss those on the actual PR?

Thanks!

brucearctor:
## Summary

Add a GCS-backed `StorageDriver` implementation under `contrib/gcp/gcsdriver`, mirroring the existing S3 driver architecture (`contrib/aws/s3driver`).

Resolves #2364

## Two-module structure

| Module | Purpose |
|--------|---------|
| `gcsdriver` | SDK-agnostic driver + `Client` interface (zero GCP dependency) |
| `gcsdriver/gcssdk` | Concrete `Client` backed by `cloud.google.com/go/storage` |

## Key features

- Content-addressable keys (SHA-256) with namespace/workflow/activity scoping
- Concurrent uploads/downloads via `errgroup` (bounded to 10 concurrent operations) with two-phase validation
- Dynamic bucket selection via `BucketFunc`
- Integrity verification on retrieval
- GCS-specific percent-encoding for object name segments per [GCS object naming](https://cloud.google.com/storage/docs/objects#naming) recommendations — encodes control characters, `# [ ] * ? : " < > |`, `/`, and `%`, while leaving readable Unicode intact
- `MaxPayloadSize` enforced on Store
- Object name key length guard: falls back to namespace-scoped digest-only path if the key exceeds GCS's 1024-byte limit
- Encoding handles `.` / `..` (forbidden by GCS)

## Design decisions

- **Check-then-write pattern:** Store calls `ObjectExists` before `PutObject` to skip unnecessary uploads for cache hits, consistent with the S3 driver.
- **ObjectExists returns `(false, nil)` for missing buckets**, consistent with the S3 driver's behavior where `HeadObject` on a missing bucket maps to a generic `NotFound`.
- **Key length fallback preserves namespace:** When the generated key exceeds 1024 bytes, the fallback path retains the `/ns/<namespace>` prefix so customers can use it for object authorization policies.
- **Sorted diagnostic metadata:** `describeClient` sorts map keys for deterministic error messages.
- **Data retention guidance:** README links to the [External Storage Lifecycle](https://docs.temporal.io/external-storage#lifecycle) docs for determining the correct retention window (max workflow run timeout + namespace retention).

## Changes since v1

Addressed all review feedback:

- Renamed `contrib/google` → `contrib/gcp` (all files, module paths, imports)
- Updated `go 1.24.0` → `go 1.25.4` and synced dependency versions with main (`api v1.63.4`, `sync v0.20.0`, `grpc v1.82.1`, etc.)
- Replaced `url.PathEscape` with GCS-specific `encodeObjectNameSegment` using rune-based iteration for correct Unicode handling
- Changed `ObjectExists` to return `(false, nil)` for missing buckets (removed bucket-existence caching)
- Added `CHANGELOG.md` for both modules
- Simplified `PutObject` to use `w.Write(data)` directly instead of `io.Copy`
- Added `errgroup.SetLimit(10)` to bound concurrent GCS operations
- Sorted `describeClient` output for deterministic error messages
- Key length fallback preserves namespace scope for authorization policies
- Added data retention guidance linking to Temporal external storage docs
- Corrected IAM permissions documentation

## Tests

- **49** driver-level unit tests (encoding, key generation, bucket selection, content addressing, round-trips, error paths, key length fallback)
- **9** GCS SDK integration tests using `fake-gcs-server` (put/get, existence, bucket-not-found, large objects, full driver round-trip)
- All tests pass with `go 1.25.4`


contrib/gcp/gcsdriver/README.md
```diff
@@ -0,0 +1,114 @@
+# GCS Storage Driver for Temporal Go SDK
```
jmaeagle99, line 1: Please rename `contrib/google` directory to `contrib/gcp` for all files and update the pkg imports.

contrib/google/gcsdriver/README.md
```diff
@@ -0,0 +1,114 @@
...
+    ObjectExists(ctx context.Context, bucket, key string) (bool, error)
+    GetObject(ctx context.Context, bucket, key string) ([]byte, error)
+}
+```
+
+Pass your implementation as `Options.Client` when calling `NewDriver`.
+
+## ObjectExists Behavior
+
+The GCS driver's `ObjectExists` implementation returns `(false, error)` when the parent bucket does not exist. This differs from the S3 driver, which returns `(false, nil)` in the same scenario due to how the AWS SDK maps `HeadObject` responses. The GCS behavior is a deliberate choice based on the `Client` interface contract: *"a non-nil error only when the existence of the object cannot be determined."* A missing bucket means existence truly cannot be determined, so an error is the more correct response. Bucket existence is cached after the first successful check, so this adds at most one extra RPC per unique bucket.
```
jmaeagle99: I disagree with this interpretation. If we are validating the existence of a stored object, it matters that the object exists in the named bucket that must exist and at the key path within that bucket. Checking for existence shouldn't error if the bucket doesn't exist, that should just return false. If you don't have authorization to inspect the bucket contents or have a network issue, that's what should return an error, because you cannot determine if it exists or not.

I would ask that this just returns false if the bucket is missing. If we feel the need to change the behavior or optimize it further, I think we can do that more holistically for all first-party drivers.

contrib/google/gcsdriver/go.mod
```diff
@@ -0,0 +1,27 @@
+module go.temporal.io/sdk/contrib/google/gcsdriver
+
+go 1.24.0
```
jmaeagle99: The go version and several of the mod versions have been updated since you opened this PR. Please update with the latest from main and adjust them to match appropriately.

contrib/gcp/gcsdriver/driver.go
```diff
@@ -0,0 +1,277 @@
...
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+	return payloads, nil
+}
+
+func objectKey(target converter.StorageDriverTargetInfo, hexDigest string) string {
```
jmaeagle99, line 241: We've [updated the encoding behavior in the S3](https://github.com/temporalio/sdk-go/commit/10fa2bb693b0ade24f176c6f861387a0c50fe45e) driver to be more specific so that it adheres to the guidance given by AWS. Please do the same for the GCS driver using the recommendations from GCP. Something similar was done already for the TypeScript GCS driver at https://github.com/temporalio/sdk-typescript/blob/main/contrib/external-storage-gcs/src/driver.ts#L69

jmaeagle99 (commented):
Thanks for contributing this GCS driver for the Go SDK! Have a couple of comments for you to take a look at. And when you do merge/rebase, there will be a CHANGELOG.md that I would ask for you to put an blurb describing the addition of this driver.

brucearctor:
All review comments addressed — added replies on each thread below.

- **Rename `contrib/google` → `contrib/gcp`**: Done — renamed all files, updated all module paths and imports.
- **ObjectExists behavior**: Agreed, updated to return `(false, nil)` for missing buckets. Removed the bucket-existence caching.
- **go.mod versions**: Rebased and updated — `go 1.25.4`, `api v1.63.4`, `sync v0.20.0`, etc.
- **GCS-specific encoding**: Done — replaced `url.PathEscape` with a GCS-specific encoder following the TypeScript driver's approach. Uses rune iteration to preserve Unicode while encoding GCS-unsafe characters.
- **CHANGELOG.md**: Added.

contrib/gcp/gcsdriver/README.md
```diff
@@ -0,0 +1,123 @@
...
+The Google Cloud credentials used by your Storage client must have the following IAM permissions on the target bucket:
+
+- `storage.objects.create` — required by components that store payloads (typically the Temporal Client and Workers sending Workflow/Activity inputs and results). Because keys are content-addressable, the driver writes unconditionally (idempotent overwrite); no read permission is needed for storing.
+- `storage.objects.get` — required by components that retrieve payloads (typically Workers and Clients reading inputs and results).
+
+The predefined `roles/storage.objectUser` role includes both permissions. Components that only retrieve payloads can use the narrower `roles/storage.objectViewer` role.
+
+## Data Retention
+
+The driver creates objects but never deletes them. Without a cleanup policy the bucket will grow indefinitely and accrue unbounded storage costs. Configure a [GCS Object Lifecycle Management](https://cloud.google.com/storage/docs/lifecycle) rule on the bucket to automatically delete objects after a number of days aligned with your Temporal namespace retention period. For example, the following lifecycle rule deletes objects older than 90 days:
```
jmaeagle99: I think the prescribed determination is not accurate enough; it should be "max workflow run timeout" + "namespace retention". We have this centrally described at https://docs.temporal.io/external-storage#lifecycle. I would suggest linking to the docs.

contrib/gcp/gcsdriver/gcssdk/README.md
```diff
@@ -0,0 +1,29 @@
+# Cloud Storage Client for GCS Storage Driver
```
jmaeagle99, line 1: Please add a CHANGELOG.md for this module.

contrib/gcp/gcsdriver/driver.go
```diff
@@ -0,0 +1,379 @@
...
+		}
+	}
+
+	claims := make([]converter.StorageDriverClaim, len(payloads))
+	g, gctx := errgroup.WithContext(ctx.Context)
+	g.SetLimit(10)
+	for i, pp := range prepared {
+		g.Go(func() error {
+			key := objectKey(ctx.Target, pp.hexDigest)
+			if err := d.client.PutObject(gctx, pp.bucket, key, pp.data); err != nil {
```
jmaeagle99: Please add the `ObjectExists` check back as you originally had it. It's typically cheaper to pay for an existence check than it is to have to upload the blob (or multiple parts) and pay for the network cost and latency for that data for a cache hit.

If we need to adjust this behavior, let's log a GitHub issue and discuss there.

contrib/gcp/gcsdriver/driver.go
```diff
@@ -0,0 +1,379 @@
...
+				if _, ok := d.allowedBuckets[bucket]; !ok {
+					return fmt.Errorf("bucket %q is not in the allowed set", bucket)
+				}
+			}
+
+			data, err := d.client.GetObject(gctx, bucket, key)
+			if err != nil {
+				return fmt.Errorf("download failed [bucket=%s, key=%s%s]: %w", bucket, key, describeClient(d.client), err)
+			}
+			if len(data) > d.maxPayloadSize {
```
jmaeagle99: This actually doesn’t prevent unbounded memory allocation on retrieval. In fact, the code is just retrieving and materializing the entire object into memory and then checking the size. If we need something like this, it would likely have to be implemented through limits on a size limited reader like `io.LimitReader`, it should probably be decoupled from the upload limit, and we should do it across all of the drivers and languages.

I would like to see a GitHub issue that describes the circumstances and scenario that this is trying to prevent before adopting it.

contrib/gcp/gcsdriver/driver.go
```diff
@@ -0,0 +1,379 @@
...
+			if algo != hashAlgorithm {
+				return fmt.Errorf("unsupported hash algorithm %q", algo)
+			}
+
+			expectedHash, ok := c.ClaimData[claimKeyHashValue]
+			if !ok {
+				return fmt.Errorf("claim missing field %q", claimKeyHashValue)
+			}
+			if sha256Hex(data) != expectedHash {
+				return fmt.Errorf(
```
jmaeagle99, line 220: I would add the hashes back to help with better diagnostics. I don’t think this is actionable from a threat perspective because the downloaded objects are large by construction and the whole design is content-addressable by hash value: the hash value is already in the object key so you already had access to it from workflow history.

This one is fairly minor, but would like to see an issue logged that describes why this should be changed (across the board) or should be specific to the GCS driver.

contrib/gcp/gcsdriver/driver.go
```diff
@@ -0,0 +1,379 @@
...
+
+	// MaxPayloadSize is the maximum serialized payload size in bytes that
+	// the driver will accept. Defaults to 50 MiB.
+	MaxPayloadSize int
+
+	// AllowedBuckets is an optional set of bucket names from which the
+	// driver is permitted to retrieve payloads. When non-empty, Retrieve
+	// rejects claims targeting buckets outside this set. Leave empty to
+	// allow any bucket.
+	AllowedBuckets []string
```
jmaeagle99: I like this idea, but it needs more refinement and I would exclude from this change for now, for similar reasons as previously stated. If we adopt something like this, some ideas to consider are deriving `AllowedBuckets` from `StaticBucket` (if that is set), validating that `BucketFunc` produces a value that is in `AllowedBuckets`, etc. Please open a separate GitHub for discussion.

contrib/gcp/gcsdriver/driver.go
```diff
@@ -0,0 +1,379 @@
...
+		key = keyVersion +
+			"/ns/" + encodeObjectNameSegment(t.Namespace) +
+			"/at/" + encodeObjectNameSegment(t.ActivityType) +
+			"/ai/" + encodeObjectNameSegment(t.ActivityID) +
+			"/ri/" + encodeObjectNameSegment(t.RunID) +
+			digestSegment
+	default:
+		return keyVersion + digestSegment
+	}
+	if len(key) > maxGCSObjectNameBytes {
```
jmaeagle99, line 262: I like this, but I would make an adjustment. Please preserve the namespace scope information as this is likely what customers will use for object authorization policies.

contrib/gcp/gcsdriver/driver.go
```diff
@@ -0,0 +1,379 @@
...
+				b.WriteByte(upperHexDigit(c >> 4))
+				b.WriteByte(upperHexDigit(c & 0xf))
+			}
+		} else {
+			b.WriteRune(r)
+		}
+	}
+	encoded := b.String()
+	// Disambiguate from the empty-string sentinel "null".
+	if encoded == nullSegment {
```
jmaeagle99: Not sure this is truly necessary; the injective property isn't truly holdable in a content-address storage scheme and gets undermined with the fallback behaviors. We should address holistically if we feel it needs to be done.

I would ask to remove and open a GitHub issue for it. Maybe we can do a separate refactor to have both (and future) drivers produce similar shapes and guards while they apply their own validity measures.

contrib/gcp/gcsdriver/gcssdk/README.md
brucearctor: Added.

contrib/gcp/gcsdriver/README.md
brucearctor: Updated — replaced the prescriptive 90-day example with the correct determination (max workflow run timeout + namespace retention) and linked to https://docs.temporal.io/external-storage#lifecycle.

contrib/gcp/gcsdriver/driver.go
brucearctor: Reverted — ObjectExists check is back. Filed #2497 for discussing unconditional uploads across drivers.

contrib/gcp/gcsdriver/driver.go
brucearctor: Removed. Filed #2498 for a proper io.LimitReader-based approach across drivers.

contrib/gcp/gcsdriver/driver.go
brucearctor: Reverted — hashes are back in the error message. Filed #2501 for cross-driver discussion.

contrib/gcp/gcsdriver/driver.go
brucearctor: Removed from this PR. Filed #2499 for design discussion.

contrib/gcp/gcsdriver/driver.go
brucearctor: Updated — the key length fallback now preserves the namespace scope: `v0/ns/<namespace>/d/sha256/<hash>`.

contrib/gcp/gcsdriver/driver.go
brucearctor: Removed. Filed #2500 for holistic discussion across drivers.

contrib/gcp/gcsdriver/README.md
```diff
@@ -0,0 +1,114 @@
...
+})
+```
+
+The above example stores payloads in the `large-payloads` bucket if their size is greater than 10 MiB; otherwise, the payloads are stored in the `small-payloads` bucket.
+
+## Required IAM Permissions
+
+The Google Cloud credentials used by your Storage client must have the following IAM permissions on the target bucket:
+
+- `storage.objects.create` — required by components that store payloads (typically the Temporal Client and Workers sending Workflow/Activity inputs and results). Because keys are content-addressable, the driver writes unconditionally (idempotent overwrite); no read permission is needed for storing.
```
jmaeagle99: The last sentence is no longer correct since the driver is not unconditionally storing and store does require read permission.

contrib/gcp/gcsdriver/driver_test.go
```diff
@@ -0,0 +1,718 @@
+package gcsdriver
```
jmaeagle99, line 1: Could use a one or two tests for the concurrent aspects, for example:
- test that the driver actually caps out at 10 concurrent uploads
- test that cancellation will actually cancel the other in-flight stores and retrieves

jmaeagle99 (approved):
Just a few more comments. And if you could, please do not squash/rebase for these last changes (makes it difficult to assess what actually has changed since the last review pass). There shouldn't be any merge conflicts since this is all new.

brucearctor:
Since no discussion on the follow up tickets [ 2497 - 2501 ] that were requested, I'll assume not much concern in which implementation, so will just proceed and send PRs.

jmaeagle99:
> Since no discussion on the follow up tickets [ 2497 - 2501 ] that were requested, I'll assume not much concern in which implementation, so will just proceed and send PRs.  

Lack of engagement on logged issues != consensus on priority, or scheduling, or implementation. I suggest you engage on each issue until there is consensus on the individual issues before proceeding with implementation.

brucearctor:
> > Since no discussion on the follow up tickets [ 2497 - 2501 ] that were requested, I'll assume not much concern in which implementation, so will just proceed and send PRs.
> 
> Lack of engagement on logged issues != consensus on priority, or scheduling, or implementation. I suggest you engage on each issue until there is consensus on the individual issues before proceeding with implementation.

Opening a PR makes it more concrete - to otherwise serve as basis for needed discussion.  Doesn't that help advance things?  Bias for action?

PR #2368, opened 2026-05-26
Shutdown refactor

### LINKED ISSUE #1197: Drain polled tasks on shutdown

yuandrew:
non-fork version of https://github.com/temporalio/sdk-go/pull/2261, so cloud tests can run

## What was changed
Refactored worker shutdown to use a two-stage approach: pollers shut down first, then the task dispatcher drains any remaining tasks before exiting. This ensures tasks polled during shutdown are processed rather than silently dropped.

Key changes:
- Added pollerWG to baseWorker to track poller goroutines separately from the global stopWG
- A closer goroutine waits for all pollers to finish, then closes taskQueueCh
- runTaskDispatcher now ranges over taskQueueCh
- pollTask always sends to taskQueueCh
- Removed the 5s timeout hack in doPoll from #2253 (no longer needed)

## Why?
PR #2199 changed shutdown to let the server complete in-flight polls, instead of cancelling them. This exposed a pre-existing race when a poller receives a task during shutdown, Go would silently dropping the task. The dispatcher had the same issue — it could exit on stopCh before reading pending tasks from the channel.

This aligns the Go SDK's shutdown with how Core SDK handles it:
1. Set a flag so pollers stop polling after their current attempt
2. Close channels from pollers → task processing
3. Wait for all in-flight tasks to complete & channels to be empty

## Checklist

1. Closes #1197

2. How was this tested:
    - New unit test TestTaskNotDroppedDuringShutdown — verifies a task polled during shutdown is processed, not dropped
    - Existing TestDoPollGracefulShutdown — validates both graceful and legacy poll completion

3. Any docs updates needed?
No

---

> [!NOTE]
> **Medium Risk**
> Changes core worker stop, poll, and dispatch paths used on every deploy; incorrect drain or legacy behavior could drop or duplicate task handling, though capability-gated and heavily tested.
> 
> **Overview**
> Refactors **worker shutdown** into a two-stage flow when the server supports **poll-complete-on-shutdown**: stop new polls, let in-flight long polls finish, then **drain** tasks already on `taskQueueCh` instead of dropping them on `stopCh` (fixes #1197).
> 
> **`baseWorker`** tracks pollers with `pollerWG`, closes `taskQueueCh` only after all pollers exit, and runs the dispatcher in drain mode with a separate **`taskLimiterContext`** so in-flight polled tasks still honor dispatch rate limits. **`doPoll`** no longer uses the temporary 5s wait on graceful shutdown.
> 
> **Aggregated shutdown** calls **`stopPolling()`** on workflow, activity, Nexus, and **session** workers; **`ShutdownWorker`** is sent for the main queue plus session creation and session-specific activity queues. Session workers now share the same drain capability instead of forcing legacy cancel behavior.
> 
> CI/dev-server and integration coverage add shutdown-under-load and session shutdown scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit be8f99394cfd59b3d40155db4a1e6fb5a9bbc300. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/internal_worker.go
```diff
@@ -1598,15 +1599,44 @@ func (aw *AggregatedWorker) shutdownWorker() {
...
 
+	aw.sendShutdownWorkerRPCForTaskQueue(grpcCtx, aw.executionParams.TaskQueue, stickyTaskQueue, aw.activeTaskQueueTypes(), heartbeat)
+
+	if util.IsInterfaceNil(aw.sessionWorker) {
+		return
+	}
+
+	aw.sendShutdownWorkerRPCForTaskQueue(
+		grpcCtx,
+		aw.sessionWorker.creationWorker.executionParameters.TaskQueue,
```
jmaeagle99: nit: Consider having the session worker report its task queues via a func so aggregate worker doesn't have to know what the structure of the session worker is.

internal/internal_worker.go
```diff
@@ -1523,10 +1520,14 @@ func (aw *AggregatedWorker) Stop() {
 	if !util.IsInterfaceNil(aw.nexusWorker) {
 		aw.nexusWorker.worker.noRepoll.Store(true)
 	}
+	if !util.IsInterfaceNil(aw.sessionWorker) {
+		aw.sessionWorker.creationWorker.worker.noRepoll.Store(true)
```
jmaeagle99: nit: Add a func to session worker to stop polling so that aggregate worker doesn't have to know about the details of the session worker.

internal/internal_worker.go
jmaeagle99, line 1550: With more work happening in the stops for each worker, consider parallelizing these calls on each of the workers.

internal/internal_worker_test.go
```diff
@@ -1952,7 +1952,7 @@ func createWorkerWithThrottle(
 	setupPollingMocks(namespace, service, activitiesPerSecond)
 
 	service.EXPECT().ShutdownWorker(gomock.Any(), gomock.Any(), gomock.Any()).
-		Return(&workflowservice.ShutdownWorkerResponse{}, nil).Times(1)
+		Return(&workflowservice.ShutdownWorkerResponse{}, nil).Times(3)
```
jmaeagle99: This and the below `Times(3)` feel like magic constants. Maybe move them to a constant and explain why this value was chosen or (if we can) calculate what it should be.

jmaeagle99 (approved):

internal/internal_worker.go
yuandrew: good idea, but if we do this, I think I'd want to do this in a separate PR

internal/internal_worker_test.go
yuandrew: thanks, added!

PR #2371, opened 2026-05-27
update PAT token for temporaliotest docker registry

THardy98:
## What was changed
Omes GH action fails because of a stale PAT token, this refreshes it.

## Why?
fix borked GH action


yuandrew (approved):
LGTM, make sure to remove the temp branch before merging. Also, do you mind removing `master` branch as well? That branch no longer exists and should be safe to remove now

THardy98:
looks like we're axing this anyway - brb gonna have a delete pr 🔨

PR #2373, opened 2026-05-28
Converter: add optional retry policy for PayloadCodec calls

### LINKED ISSUE #2370: Add optional retry policy for PayloadCodec failures

A `PayloadCodec` may call an external service during Encode/Decode — examples include the in-tree `NewRemotePayloadCodec` (HTTP to a sidecar), KMS-backed encryption codecs that call a cloud KMS for data-key operations, and codecs that fetch schemas from a remote registry. Any of these can fail transiently (network blip, throttling, 5xx). Today the SDK has no way to retry a codec call: the first error from any codec immediately aborts the encode/decode and surfaces to user code. Each downstream user has to either fork the codec or wrap retry logic inside their own codec implementation, which duplicates effort and doesn't help the stdlib `RemotePayloadCodec`, which has no retries today (see `converter/codec.go` — `RemotePayloadCodec.encodeOrDecode` does a single `pc.options.Client.Do(req)` and returns the error directly).

A sibling constructor `NewCodecDataConverterWithOptions(parent, codecs, CodecDataConverterOptions)` mirroring the `*WithOptions` pattern used in PR #2225 (Serialization Context). The options struct exposes an optional retry policy applied per codec call:

```go
type CodecDataConverterOptions struct {
    RetryPolicy *CodecRetryPolicy // nil => no retry (default)
    IsRetryable func(error) bool  // nil => retry all errors
    Context     context.Context   // nil => context.Background()
}

type CodecRetryPolicy struct {
    InitialInterval    time.Duration
    BackoffCoefficient float64
    MaximumInterval    time.Duration
    MaximumAttempts    int
    ExpirationInterval time.Duration
}
```

The existing `NewCodecDataConverter(parent, codecs...)` stays unchanged and delegates to the new one with a zero-value options struct, preserving backward compatibility (no retries by default).

Retry boundary: each codec's Encode/Decode is retried independently, so a chain of `[compress, encrypt]` re-encrypts on transient KMS failure without re-compressing (relevant for non-idempotent codecs like authenticated encryption).

**Example usage**

```go
dc := converter.NewCodecDataConverterWithOptions(
    parent,
    []converter.PayloadCodec{kmsCodec},
    converter.CodecDataConverterOptions{
        RetryPolicy: &converter.CodecRetryPolicy{
            InitialInterval:    100 * time.Millisecond,
            BackoffCoefficient: 2.0,
            MaximumAttempts:    4,
        },
        IsRetryable: func(err error) bool {
            // codec-specific predicate, e.g. detect KMS throttling
            return isTransientKMSError(err)
        },
        Context: ctx,
    },
)
```

- Each codec author implements retry internally. Works, but duplicates effort and doesn't fix `RemotePayloadCodec`.
- Exposing `internal/common/backoff` publicly. Larger surface; out of proportion to the need.
- Adding `context.Context` to `PayloadCodec.Encode/Decode`. Interface-breaking; separate discussion. The options-struct Context only governs retry sleeps, not the codec call itself.

**Additional context**

Related prior issues: #2206 (closed by #2228) hit exactly this scenario — a remote codec returned `DeadlineExceeded` during `encodeArgs` and, combined with a session activity cancellation, permanently bricked a workflow. PR #2228 fixed the catastrophic side-effect (no more panic; the workflow task fails gracefully and the server retries), but the underlying transient codec failure still kills the in-flight task and forces a full task retry. This proposal is the missing root-cause complement: codecs that opt in recover from a transient blip without burning a workflow task at all.

I'm happy to implement this and send up a PR, but ideally I want to confirm the API shape with maintainers before doing so. Primary open questions remaining are:

1. The local `CodecRetryPolicy` struct as proposed, or should it thread something else (e.g. `temporal.RetryPolicy`)? A local struct avoids an import cycle from `converter/` back into `temporal/`.
2. Any objection to the per-codec (vs. whole-chain) retry boundary?
3. Default `IsRetryable` semantics — retry-all (proposed) vs. retry-none (force explicit classification)?

whitemanthedj:
Adds NewCodecDataConverterWithOptions and a CodecDataConverterOptions struct exposing an opt-in retry policy applied per individual codec call. Motivated by codecs that call external services (the in-tree RemotePayloadCodec, KMS-backed codecs, schema-registry codecs) where transient failures today abort the whole conversion with no recovery path.

NewCodecDataConverter signature is unchanged; the zero-value options struct preserves current behavior.

Fixes #2370

## What was changed

New public surface, all in `converter/codec.go`:

```go
func NewCodecDataConverterWithOptions(
    parent DataConverter,
    codecs []PayloadCodec,
    options CodecDataConverterOptions,
) DataConverter

type CodecDataConverterOptions struct {
    RetryPolicy *CodecRetryPolicy
    IsRetryable func(error) bool
    Context     context.Context
}

type CodecRetryPolicy struct {
    InitialInterval    time.Duration
    BackoffCoefficient float64
    MaximumInterval    time.Duration
    MaximumAttempts    int
    ExpirationInterval time.Duration
}
```

Behavior:

- Default (no `RetryPolicy` set): each codec is invoked exactly once and the first error aborts the chain. Byte-for-byte identical to current behavior. A `TestCodecDataConverter_Retry_NewCodecDataConverterIsZeroOptions` test asserts equivalence.
- With `RetryPolicy` set: each codec in the chain is wrapped in a retry loop independently. `IsRetryable` (if non-nil) classifies each error; nil means retry all, matching the convention in `internal/common/backoff.Retry` and `temporal.RetryPolicy` (empty `NonRetryableErrorTypes` retries everything). `Context` (defaulting to `context.Background()`) governs cancellation of sleeps between attempts.
- `CodecRetryPolicy` zero-value field defaults: 100ms initial interval, backoff coefficient 2.0, max interval equal to 100 times initial. These mirror the conventions used by `internal/common/retry.Default*` via three unexported package constants declared at the top of `converter/codec.go`.

Retry boundary is per-codec, not per-chain. A chain of `[compress, encrypt]` only retries the failing codec. Re-running a successful predecessor would invalidate its output for non-idempotent codecs such as authenticated encryption and KMS data-key wrapping.

Files modified:

- `converter/codec.go` (+186, -6): new exported types and constructor, three unexported default constants, private `callWithRetry` helper (no `internal/` import), `options` field on `CodecDataConverter`, updated `WithSerializationContext` to propagate options through derived converters. Existing `NewCodecDataConverter` doc comment updated to mention the new sibling.
- `converter/codec_test.go` (+334): `flakyCodec`, `countingCodec`, `retryTestSerializationContext`, and `fastRetry` helpers, plus ten new retry tests.

Zero changes outside `converter/`. No new module dependencies (`go.mod` and `go.sum` are untouched; verified by `go mod tidy` producing no diff). No changes under `internal/`. No changes to the `PayloadCodec` interface. No changes to `NewPayloadCodecHTTPHandler`.

Out of scope, deliberately:

- Retry behavior on `PayloadConverter` and `FailureConverter`. Those have programmer-error failure modes rather than transient I/O.
- Adding `context.Context` to the `PayloadCodec` interface itself. The options-struct `Context` only governs sleeps between retries.
- Retry inside `NewPayloadCodecHTTPHandler`. Server-side retry is the operator's responsibility.
- Promotion of `internal/common/backoff` to a public package.

## Why?

A `PayloadCodec` may call an external service during Encode and Decode. The in-tree `RemotePayloadCodec` is the canonical example (HTTP to a sidecar), and KMS-backed encryption codecs plus schema-registry codecs are common community variants. Any of these can fail transiently from network blips, throttling, or 5xx responses. Today the SDK has no way to retry a codec call: the first error from any codec immediately aborts the encode/decode and surfaces to user code. Each downstream user has to either fork the codec or wrap retry logic inside their own codec implementation, which duplicates effort across the ecosystem and does not help the stdlib `RemotePayloadCodec`, which has no retries today (`pc.options.Client.Do(req)` followed by `return payloads, err`).

This proposal complements PR #2228, which made codec-failure side effects non-catastrophic for session workflows. That fix prevents permanent bricking when a codec failure cascades through a session activity cancellation, but a transient `DeadlineExceeded` from a remote codec still kills the in-flight workflow task and forces the server to retry the entire task. This PR is the root-cause complement: codecs that opt in can absorb the transient blip in place without ever losing the workflow task.

Determinism: codecs run at the SDK serialization boundary, never inside the workflow goroutine's user code. Retrying does not introduce new history events, new commands, or new branches in workflow code. History replay re-invokes the codec the same way it would on the original execution. Replay-aware retries are safe by construction.

Design choices the maintainers may want to weigh in on (carried from issue #2370):

1. Local `CodecRetryPolicy` struct as proposed, or thread something else (`temporal.RetryPolicy` would require breaking an import cycle from `converter/` back into `temporal/`).
2. Per-codec retry boundary as proposed, or would a `RetryWholeChain bool` opt-in be preferred.
3. `IsRetryable nil = retry all` matches `backoff.Retry` and `temporal.RetryPolicy` conventions. Would the safer `IsRetryable nil = retry none` default be preferred instead.

## Checklist

1. Closes #2370

2. How was this tested:

Local verification, all from `internal/cmd/build`:

```bash
go run . check                                   # vet, errcheck, staticcheck, doclink: clean
go run . unit-test -run "TestCodecDataConverter" # 11 tests pass (10 new plus 1 pre-existing)
go run . unit-test                               # full suite passes, zero FAIL lines
go run . integration-test -dev-server            # passes; one pre-existing Nexus flake unrelated to converter
```

Ten new tests in `converter/codec_test.go`:

- `TestCodecDataConverter_Retry_EncodeSucceedsAfterTransientFailures`: flaky codec fails twice and succeeds on the third attempt.
- `TestCodecDataConverter_Retry_EncodeExhaustsAttempts`: always-failing codec exhausts `MaximumAttempts` and returns the last error.
- `TestCodecDataConverter_Retry_DecodeRetried`: symmetric coverage on the Decode path.
- `TestCodecDataConverter_Retry_NonRetryableErrorFailsImmediately`: `IsRetryable` returns false; single attempt.
- `TestCodecDataConverter_Retry_DefaultPolicyPreservesOldBehavior`: backward compatibility confirmed against the pre-existing `errorCodecOnEncode` fake.
- `TestCodecDataConverter_Retry_PerCodecBoundary`: verifies a successful predecessor codec is not re-run when a later codec retries.
- `TestCodecDataConverter_Retry_ContextCancellationStopsRetries`: context cancellation aborts the retry loop and surfaces the codec error, not `context.Canceled`.
- `TestCodecDataConverter_Retry_NewCodecDataConverterIsZeroOptions`: `NewCodecDataConverter` produces output behaviorally identical to `NewCodecDataConverterWithOptions` with a zero-value options struct.
- `TestCodecDataConverter_Retry_OptionsPreservedThroughSerializationContext`: `WithSerializationContext` derivation does not silently drop retry config.
- `TestCodecDataConverter_Retry_NilContextDefaultsToBackground`: `Context: nil` does not panic and falls back to `context.Background()`.

The pre-existing `TestCodecDataConverter_ToPayload_EncodeError` was not modified and continues to pass without changes. The six other pre-existing `TestCodecDataConverter_*` tests (propagation, signing-mismatch, etc.) also pass unchanged.

In addition to in-repo tests, an out-of-tree consumer program that calls `NewCodecDataConverterWithOptions` against a flaky codec returning `kms 503` for the first two attempts succeeds on the third attempt and prints `err=<nil> payload=true`. This confirms the public API shape is usable from outside the SDK module.

All retry tests use sub-millisecond intervals so the new suite runs in well under one second.

3. Any docs updates needed?

No external docs updates are required. The public API additions carry doc comments on every exported symbol matching the density set by `serialization_context.go`. The existing `NewCodecDataConverter` doc comment was updated to point readers at the new sibling constructor for the retry-aware path. No README updates needed; no docs.temporal.io updates needed since this is a Go SDK addition with self-documenting types.

### DIFF
--- a/converter/codec.go
+++ b/converter/codec.go
@@ -3,11 +3,14 @@ package converter
 import (
 	"bytes"
 	"compress/zlib"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"strings"
+	"time"
 
 	commonpb "go.temporal.io/api/common/v1"
 	"google.golang.org/protobuf/encoding/protojson"
@@ -130,12 +133,69 @@ func encodePayloads(payloads []*commonpb.Payload, codecs []PayloadCodec) ([]*com
 	return payloads, nil
 }
 
+// Defaults applied by (*CodecDataConverter).callWithRetry when a
+// CodecRetryPolicy field is zero. Values mirror the conventions in
+// internal/common/retry.Default* so codec retry behavior is consistent with
+// the rest of the SDK's retry surface.
+const (
+	defaultCodecRetryInitialInterval       = 100 * time.Millisecond
+	defaultCodecRetryBackoffCoefficient    = 2.0
+	defaultCodecRetryMaximumIntervalFactor = 100
+)
+
+// CodecRetryPolicy controls retry and backoff for individual PayloadCodec
+// Encode/Decode calls made by a CodecDataConverter. Zero values for each
+// field fall back to the documented defaults so the zero value is usable.
+type CodecRetryPolicy struct {
+	// InitialInterval is the wait before the first retry. Defaults to 100ms.
+	InitialInterval time.Duration
+	// BackoffCoefficient is the multiplier applied to InitialInterval after
+	// each failed attempt. Defaults to 2.0.
+	BackoffCoefficient float64
+	// MaximumInterval caps the per-attempt wait. Defaults to 100 *
+	// InitialInterval.
+	MaximumInterval time.Duration
+	// MaximumAttempts is the total attempt count including the first call.
+	// 0 means unlimited (bounded only by ExpirationInterval and the
+	// retry-loop context).
+	MaximumAttempts int
+	// ExpirationInterval is the absolute wall-clock budget across all
+	// attempts. 0 means no time bound.
+	ExpirationInterval time.Duration
+}
+
+// CodecDataConverterOptions configures a CodecDataConverter built with
+// NewCodecDataConverterWithOptions. The zero value preserves the behavior of
+// NewCodecDataConverter exactly: each codec is invoked once and the first
+// error aborts the chain.
+type CodecDataConverterOptions struct {
+	// RetryPolicy, if non-nil, is applied to each individual PayloadCodec
+	// Encode and Decode call. Each codec in the chain is retried
+	// independently, so a transient failure in one codec does not re-run
+	// codecs that already succeeded — this matters for non-idempotent codecs
+	// such as authenticated encryption.
+	RetryPolicy *CodecRetryPolicy
+
+	// IsRetryable, if non-nil, is consulted on every codec error to decide
+	// whether to retry. Return true to keep retrying (subject to RetryPolicy
+	// limits), false to fail fast. If nil, all errors are treated as
+	// retryable.
+	IsRetryable func(error) bool
+
+	// Context, if non-nil, governs cancellation of waits between retry
+	// attempts. Defaults to context.Background(). The context is only used
+	// to short-circuit retry sleeps; it is not propagated into PayloadCodec
+	// calls, which do not take a context today.
+	Context context.Context
+}
+
 // CodecDataConverter is a DataConverter that wraps an underlying data
 // converter and supports chained encoding of just the payload without regard
 // for serialization to/from actual types.
 type CodecDataConverter struct {
-	parent DataConverter
-	codecs []PayloadCodec
+	parent  DataConverter
+	codecs  []PayloadCodec
+	options CodecDataConverterOptions
 }
 
 // NewCodecDataConverter wraps the given parent DataConverter and performs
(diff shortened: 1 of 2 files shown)


yuandrew:
Codec authors can decide what and when to retry things internally in their codec, since they'll have a better idea than we would with generic "retry any failure" logic, see issue discussion for more details

PR #2381, opened 2026-06-02
Add Standalone Activity support to Temporal Nexus Operation Handler

Quinn-With-Two-Ns:
## What was changed

Add Standalone Activity support to Temporal Nexus Operation Handler

## Why?

Allows starting Standalone Activities from Nexus Operation Handlers

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Touches Nexus async operation lifecycle, stand-alone activity RPC fields, and completion callbacks—important for correctness but well covered by new integration and unit tests; some behavior still depends on server fixes.
> 
> **Overview**
> **Activity-backed async Nexus operations** — `temporalnexus.StartActivity` and `StartUntypedActivity` let `MustNewTemporalOperation` handlers schedule **stand-alone** activities and return an async **activity-execution** operation token (new token type, cancel via default `CancelActivityExecution`). Nexus start options (request ID, completion callbacks, links) are propagated onto `ClientStartActivityOptions`, including **USE_EXISTING** attach behavior for shared backing activities.
> 
> **Client & linking** — Internal-only setters on start-activity options wire callbacks, links, on-conflict attach, and capture the start response **Link**. `Link_Activity` ↔ Nexus link conversion is added and wired into link parsing.
> 
> **Test workflow environment** — Nexus test client implements `ExecuteActivity` (when the workflow-run op context is set), simulates Nexus completion callbacks when the activity finishes, and supports cancel via activity handles.
> 
> **Tests & CI** — Large integration suite for success, cancel, failures, timeouts, retries, shared activity / dual callbacks, and terminated caller; workflow test suite unit test; dev server enables `activity.enableCallbacks`; docker CI sets `DISABLE_ACTIVITY_BACKED_NEXUS_TESTS`. Some link and heartbeat-details assertions are skipped pending server work (NEXUS-400).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 199861ca611468443093a40fd6da70973e871752. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,6 +169,7 @@ jobs:
           DISABLE_PRIORITY_TESTS: "1"
           DISABLE_STANDALONE_ACTIVITY_TESTS: "1"
           DISABLE_STANDALONE_NEXUS_TESTS: "1"
+          DISABLE_ACTIVITY_BACKED_NEXUS_TESTS: "1"
           DISABLE_NEXUS_CALLER_TIMEOUT_TESTS: "1"
           DISABLE_NEW_NEXUS_ERROR_FORMAT_TESTS: "1"
         working-directory: ./internal/cmd/build
--- a/temporalnexus/temporal_operation.go
+++ b/temporalnexus/temporal_operation.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/nexus-rpc/sdk-go/nexus"
-	"go.temporal.io/api/common/v1"
+	commonpb "go.temporal.io/api/common/v1"
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/internal"
 	"go.temporal.io/sdk/workflow"
@@ -20,6 +20,8 @@ var (
 		Cause: errors.New("cannot cancel an UpdateWorkflow operation"),
 	}
 )
+
+const errMultipleAsyncOperationsMsg = "only one async operation can be started per operation invocation"
 
 // StartTemporalOperationOptions are options provided to the Start callback of a Temporal Nexus operation.
 // Mirrors [nexus.StartOperationOptions].
@@ -62,6 +64,18 @@ type CancelTemporalUpdateWorkflowOptions struct {
 	// UpdateID to be cancelled.
 	UpdateID string
 	// RunID of the workflow
+	RunID string
+}
+
+// CancelTemporalActivityExecutionOptions are options provided to the CancelActivityExecution
+// callback of a Temporal Nexus operation.
+//
+// NOTE: Experimental
+type CancelTemporalActivityExecutionOptions struct {
+	// ActivityID extracted from the operation token.
+	ActivityID string
+	// RunID extracted from the operation token. May be empty for tokens generated before the
+	// activity was started (e.g. callback tokens).
 	RunID string
 }
 
@@ -160,7 +174,7 @@ func StartUntypedWorkflow[R any](
 	args ...any,
 ) (TemporalOperationResult[R], error) {
 	if nc.asyncStarted != nil && !nc.asyncStarted.CompareAndSwap(false, true) {
-		return TemporalOperationResult[R]{}, nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "only one async operation can be started per operation invocation")
+		return TemporalOperationResult[R]{}, nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, errMultipleAsyncOperationsMsg)
 	}
 
 	// Prevent the test env client from panicking when we try to use it from an operation.
@@ -219,7 +233,7 @@ func StartUpdateWorkflow[R any](
 	}
 	if !nc.asyncStarted.CompareAndSwap(false, true) {
 		return TemporalOperationResult[R]{}, nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest,
-			"only one async operation can be started per operation invocation")
+			errMultipleAsyncOperationsMsg)
 	}
 	defer func() {
 		if opFailed {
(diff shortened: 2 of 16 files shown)


temporalnexus/temporal_operation.go
```diff
@@ -155,6 +168,143 @@ func StartUntypedWorkflow[R any](
...
+// See [StartActivity] for the type-safe variant.
+func StartUntypedActivity[R any](
+	ctx context.Context,
+	nc NexusClient,
+	activityOpts client.StartActivityOptions,
+	activity any,
+	args ...any,
+) (TemporalOperationResult[R], error) {
+	if nc.asyncStarted != nil && !nc.asyncStarted.CompareAndSwap(false, true) {
+		return TemporalOperationResult[R]{}, nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "only one async operation can be started per operation invocation")
```
VegetarianOrc: Maybe we should consider defining this error as a package var for re-use as we add more functionality to `NexusClient`?

VegetarianOrc (approved):

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: I made the message a var we can reuse, I couldn't do the whole error because it is mutable

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +170,143 @@ func StartUntypedWorkflow[R any](
...
+			},
+		})
+	}
+
+	// Duplicated in links to be compatible with older servers that don't read links from callbacks.
+	internal.SetLinksOnStartActivityOptions(&activityOpts, links)
+	internal.SetOnConflictOptionsOnStartActivityOptions(&activityOpts)
+	responseInfo := internal.SetResponseInfoOnStartActivityOptions(&activityOpts)
+
+	handle, err := c.ExecuteActivity(ctx, activityOpts, activity, args...)
```
yuandrew: Do we want a `ResolveActivityName` that mirrors `ResolveWorkflowName` for NexusOperationContext before we call `ExecuteActivity`?

.github/workflows/ci.yml
```diff
@@ -148,6 +148,7 @@ jobs:
           DISABLE_PRIORITY_TESTS: "1"
```
yuandrew, line 169: Not related to this file, but do we want Activity support for `testSuiteClientForNexusOperations`? Codex pointed out that we have workflow impl, but no activity impl. So users could theoretically write activity-backed Nexus operations that work against a real server but panic in the SDK test environment

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +170,143 @@ func StartUntypedWorkflow[R any](
 	return NewAsyncResult[R](handle.token()), nil
 }
 
+// StartActivity schedules a stand-alone activity execution for a Nexus operation and returns an
+// async result with an activity-execution operation token.
+//
+// The activity parameter must have the signature func(context.Context, I) (O, error). For
+// activities that don't follow this signature, use [StartUntypedActivity].
+//
+// activityOpts must specify an ID (required — auto-generation would produce a non-deterministic
```
yuandrew: Where does this get checked? I couldn't find any `activityOpts.ID` check

yuandrew:
oh yeah, and a CHANGELOG.md entry would be good

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: Yes, the SDK should already check this , here we call `WorkflowClient.ExecuteActivity` that should do the check

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: `ExecuteActivity ` should resolve the activity name, this means we should resolve the name the same as an normal client would

.github/workflows/ci.yml
Quinn-With-Two-Ns: Sure, like workflow support it won't be 100% but we can at least match the support we have for workflows.

tekkaya:
While working on implementing the canaries for the functionality introduced in this PR I have faced an issue when the activity is registered with a custom name using `RegisterActivityWithOptions(fn, {Name: "my.activity"})` passed to `StartActivity`. I was able to run the canaries successfully with this change in the sdk-go code: https://github.com/temporalio/sdk-go/commit/654385bf0eb38646bfad812e18c7b4b0398249a0

test/integration_test.go
```diff
@@ -10017,25 +10017,527 @@ func (ts *IntegrationTestSuite) TestTemporalOperationSuite() {
...
 			ts.Equal(tc.expected, result)
+			if tc.verify != nil {
+				tc.verify(run)
+			}
 		})
 	}
 }
+
+// TestStandaloneActivityHeartbeatDetailsRegression demonstrates that the
+// missing-LastHeartbeatDetails bug on stand-alone activity heartbeat timeouts is NOT Nexus
```
yuandrew, line 10094: Is there an issue tracking this bug?

test/integration_test.go
```diff
@@ -10017,25 +10017,527 @@ func (ts *IntegrationTestSuite) TestTemporalOperationSuite() {
...
+// specific. The activity is scheduled directly via client.ExecuteActivity (no Nexus
+// involvement), heartbeats once with deterministic details, then stalls until the server
+// fires a HEARTBEAT-typed timeout. The caller polls via ClientActivityHandle.Get and asserts
+// that TimeoutError exposes LastHeartbeatDetails — the same invariant the legacy
+// workflow-backed activity timeout path upholds at
+// service/history/timer_queue_active_task_executor.go:346.
+//
+// **Currently expected to FAIL on NEXUS-400**: chasm/lib/activity/statemachine.go's
```
yuandrew, line 10102: Not the biggest fan of referencing internal Jira tickets in SDK comments, is there a public facing issue we can link here?

yuandrew (approved):
nothing blocking

test/integration_test.go
Quinn-With-Two-Ns: This is actually resolved, just not released

test/integration_test.go
Quinn-With-Two-Ns: Same as https://github.com/temporalio/sdk-go/pull/2381#discussion_r3625496712 basically just pending a release

PR #2382, opened 2026-06-02
Support Standalone Activity start delay

maciejdudko:
## What was changed

Added support for start delay when starting a Standalone Activity.

## Checklist

1. How was this tested:
New test added.

### DIFF
--- a/internal/internal_activity_client.go
+++ b/internal/internal_activity_client.go
@@ -110,6 +110,8 @@ type (
 		//
 		// WARNING: Task queue priority is currently experimental.
 		Priority Priority
+		// StartDelay - Time to wait before dispatching the activity. This delay is not applied to retry attempts.
+		StartDelay time.Duration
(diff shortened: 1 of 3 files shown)


internal/internal_activity_client.go
```diff
@@ -110,6 +110,8 @@ type (
 		//
 		// WARNING: Task queue priority is currently experimental.
 		Priority Priority
+		// StartDelay - Time to wait before dispatching the activity. This delay is not applied to retry attempts.
+		StartDelay time.Duration
```
yuandrew, line 114: Do we need to mark this as experimental?

yuandrew (approved):

internal/internal_activity_client.go
maciejdudko: The whole SAA is experimental, so no need to mark this individually.

PR #2383, opened 2026-06-02
Fix activity alias collision when functions share a short name across packages

### LINKED ISSUE #2379: Activity invocation with duplicate function names is silently redirected to aliased registration
## Expected Behavior
The workflow should invoke the activity registered with the requested name, not an aliased activity with the same underlying function name.

The worker registers two activities that have the same function name but different package implementations: one without an alias and one with an alias.

```
w.RegisterActivity(acvitity_v1.Activity)
w.RegisterActivityWithOptions(
    acvitity_v2.Activity,
    activity.RegisterOptions{
       Name: "Activity_v2",
    },
)

```
The workflow invokes the non-aliased activity either by name or by function:

```

err := workflow.ExecuteActivity(ctx, "Activity", name).Get(ctx, &result) or
err := workflow.ExecuteActivity(ctx, acvitity_v1.Activity, name).Get(ctx, &result)

```

However, the workflow ends up invoking activity_v2.Activity.

I think the issue is that the function [getActivityFunctionName](https://github.com/temporalio/sdk-go/blob/a11780227b4a4a2ef87375360772a16f573445a9/internal/internal_worker.go#L2084-L2090) finds an alias registered for the given function name and resolves with the function attached to the alias

## Actual Behavior
The worker should invoke the activity registered under the requested name, rather than silently resolving the invocation to an aliased activity with the same function name.

Attached you can find the reproduction 

[30967.zip](https://github.com/user-attachments/files/28370890/30967.zip)

amit-writes-code:
Fix activity alias collision when functions share a short name across packages

When two activity functions from different packages share the same short name
(e.g. activities_v1.Activity and activities_v2.Activity), registering one with
an alias caused that alias to incorrectly match the other at dispatch time.

The root cause: activityAliasMap was keyed by the short function name returned
by getFunctionName ("Activity"), which is identical for both. A new helper
getFullFunctionName keys the alias map by the fully qualified runtime name
instead, making each entry unique regardless of short-name collisions.

Fixes #2379.

## What was changed

- Added `getFullFunctionName` helper in `internal/internal_worker.go` that
  returns the fully qualified runtime function name
  (e.g. `github.com/.../activities_v2.Activity`) before the package path is
  stripped, unlike `getFunctionName` which returns only the short name
  (`Activity`).
- `RegisterActivityWithOptions` now keys `activityAliasMap` by the full
  function name instead of the short name, so aliases from different packages
  can never collide.
- `getActivityFunctionName` now looks up aliases by full name, ensuring the
  alias written for `activities_v2.Activity` is never applied to
  `activities_v1.Activity` at dispatch time.
- Updated `TestAliasUnqualifiedNameClash` and `TestAliasStringNameClash` whose
  expectations were pinning the previously-buggy behavior — both now assert the
  correct result.
- Added `TestAliasDifferentPackageSameFunctionName` covering the exact scenario
  from #2379: two functions from different packages with the same short name,
  one aliased, dispatched by function reference.

## Why?

`activityAliasMap` used the short function name as its key. Two functions from
different packages can share the same short name — `activities_v1.Activity` and
`activities_v2.Activity` both truncate to `"Activity"`. This meant registering
`activities_v2.Activity` with an alias wrote `activityAliasMap["Activity"] =
"Activity_v2"`, which then silently redirected any dispatch of
`activities_v1.Activity` to `Activity_v2` instead. The fix keys the alias map
by the fully qualified name, which is unique per function regardless of
short-name collisions.

## Checklist

1. Closes #2379
2. How was this tested:
   - Existing `TestAliasStringNameClash` and `TestAliasUnqualifiedNameClash`
     updated and passing — these previously asserted buggy behavior, now assert
     correct behavior.
   - New `TestAliasDifferentPackageSameFunctionName` added covering the exact
     scenario from the issue: two functions with the same short name from
     different packages, one aliased, dispatched by function reference.
   - Run with: `go test ./internal/ -run "TestAlias" -v`
3. Any docs updates needed? No — this is an internal bug fix with no API or
   behavior change visible to users registering activities normally.

### DIFF
--- a/internal/internal_search_attributes.go
+++ b/internal/internal_search_attributes.go
@@ -416,6 +416,38 @@ func serializeTypedSearchAttributes(searchAttributes map[SearchAttributeKey]inte
 	return &commonpb.SearchAttributes{IndexedFields: serializedAttr}, nil
 }
 
+// sanitizeSearchAttributesForStart removes null or unset search attributes from
+// Start-like requests, even though workflow info may still contain them in the
+// current run.
+func sanitizeSearchAttributesForStart(attributes *commonpb.SearchAttributes) *commonpb.SearchAttributes {
+	indexedFields := attributes.GetIndexedFields()
+	if len(indexedFields) == 0 {
+		return attributes
+	}
+
+	var filteredAttributes map[string]*commonpb.Payload
+	for _, payload := range indexedFields {
+		if string(payload.GetMetadata()[converter.MetadataEncoding]) == converter.MetadataEncodingNil {
+			filteredAttributes = make(map[string]*commonpb.Payload, len(indexedFields)-1)
+			break
+		}
+	}
+	if filteredAttributes == nil {
+		return attributes
+	}
+
+	for key, payload := range indexedFields {
+		if string(payload.GetMetadata()[converter.MetadataEncoding]) == converter.MetadataEncodingNil {
+			continue
+		}
+		filteredAttributes[key] = payload
+	}
+	if len(filteredAttributes) == 0 {
+		return nil
+	}
+	return &commonpb.SearchAttributes{IndexedFields: filteredAttributes}
+}
+
 func serializeSearchAttributes(
 	untypedAttributes map[string]interface{},
 	typedAttributes SearchAttributes,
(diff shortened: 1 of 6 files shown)


amit-writes-code:
another approach could be to check activityFuncMap before activityAliasMap but this approach is bit more messier

amit-writes-code:
duplicate PR https://github.com/temporalio/sdk-go/pull/2385

PR #2384, opened 2026-06-02
Warn users when task aliases cause ambiguities

### LINKED ISSUE #2379: Activity invocation with duplicate function names is silently redirected to aliased registration

dplyukhin:
## What was changed
Fix logic to disambiguate activities with aliases

## Why?
Bug

## Checklist

1. Closes #2379

2. How was this tested: Added a test to internal_worker_test.go

3. Any docs updates needed? Nope

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -700,6 +700,21 @@ func (r *registry) RegisterWorkflowWithOptions(
 	if len(alias) > 0 && r.workflowAliasMap != nil {
 		r.workflowAliasMap[fnName] = alias
 	}
+
+	if r.workflowAliasMap != nil {
+		// If activityAliasMap[f] = a and activityFuncMap[f] exists, invoking "f" will actually trigger activityFuncMap[a].
+		// Unfortunately this is expected behavior when the user disobeys us by not turning on DisableRegistrationAliasing;
+		// see TestAliasStringNameClash, TestAliasUnqualifiedNameClash, and TestAliasAntialiasing. At least we can warn them.
+		a, ok1 := r.workflowAliasMap[fnName]
+		_, ok2 := r.workflowFuncMap[fnName]
+		if ok1 && ok2 {
+			fmt.Printf("WARNING: Workflow alias collision detected: invoking workflow \"%v\" will actually trigger \"%v\". Consider turning on 'WorkerOptions.DisableRegistrationAliasing'.\n", fnName, a)
+		}
+		a, ok3 := r.workflowAliasMap[registerName]
+		if ok3 {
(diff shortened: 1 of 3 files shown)


internal/internal_worker_test.go
```diff
@@ -3384,6 +3384,43 @@ func TestAliasUnqualifiedNameClash(t *testing.T) {
...
+		ctx = WithActivityOptions(ctx, ActivityOptions{ScheduleToCloseTimeout: 5 * time.Second})
+		var str1 string
+		if err := ExecuteActivity(ctx, "MyActivity").Get(ctx, &str1); err != nil {
+			return "", err
+		}
+		var str2 string
+		if err := ExecuteActivity(ctx, "MyActivity_v2").Get(ctx, &str2); err != nil {
+			return "", err
+		}
+		return str1 + "-" + str2, nil
```
yuandrew, line 3403: I'm not sure this solution is right, I also tried adding the following line into the test, and it's returning running the `aliasReproV2` activity and returning func2

```suggestion
        var str3 string
		if err := ExecuteActivity(ctx, aliasReproV1{}.MyActivity).Get(ctx, &str3); err != nil {
			return "", err
		}
		
		return str1 + "-" + str2 + "-" + str3, nil
```

internal/internal_worker.go
```diff
@@ -2625,6 +2625,9 @@ func getFunctionName(i interface{}) (name string, isMethod bool) {
 }
 
 func getActivityFunctionName(r *registry, i interface{}) string {
+	if name, ok := i.(string); ok {
+		return name
+	}
 	result, _ := getFunctionName(i)
```
yuandrew, line 2658: hmm, `getFunctionName` does this same logic at the beginning of the function, why do we need this at this level?

PR #2386, opened 2026-06-03
Add workflowstreams contrib package

brianstrauch:
## What

Adds a new `contrib/workflowstreams` package: a durable publish/subscribe log hosted inside a Temporal workflow.

External code (activities, starters, other workflows) publishes messages to named topics via signals; subscribers long-poll for new items via updates; a query exposes the current offset. The stream is backed by Temporal's durable execution, giving exactly-once, ordered, cross-language delivery with client-side batching, publisher dedup, continue-as-new survival, truncation, and response paging.

The handler names, JSON envelope field names, and per-item payload encoding match the Python (`temporalio.contrib.workflow_streams`) and TypeScript (`@temporalio/workflow-streams`) packages, so a Go publisher/subscriber interoperates with a Python/TypeScript workflow and vice versa.

## Supporting changes

- Adds a reserved-name prefix exception allowing the `__temporal_workflow_stream` prefix for signal, update, and query handlers (`internal/internal_utils.go`, `internal/workflow.go`), which `workflowstreams` relies on for its internal handler names.
- Adds `internal/reserved_name_exception_test.go` covering the exception.
- Adds a CODEOWNERS entry for `/contrib/workflowstreams/` (`@temporalio/sdk @temporalio/ai-sdk`).


contrib/workflowstreams/client.go
```diff
@@ -0,0 +1,242 @@
...
+func (c *Client) followContinueAsNew(ctx context.Context) bool {
+	if !c.followCAN {
+		return false
+	}
+	desc, err := c.c.DescribeWorkflowExecution(ctx, c.workflowID, "")
+	if err != nil {
+		return false
+	}
+	return desc.GetWorkflowExecutionInfo().GetStatus() == enumspb.WORKFLOW_EXECUTION_STATUS_CONTINUED_AS_NEW
+}
```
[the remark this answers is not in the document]
brianstrauch: This bug also exists for Python and TypeScript. Fixing across all languages...

contrib/workflowstreams/README.md
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	// Optionally publish from workflow code:
+	if err := stream.Topic("events").Publish("hello from the workflow"); err != nil {
+		return err
+	}
+
+	// Run your workflow; the stream serves external publishers and subscribers
+	// for as long as the workflow is running.
+	return workflow.Await(ctx, func() bool { return done })
```
yuandrew, line 41: Where does `done` get defined?

contrib/workflowstreams/README.md
```diff
@@ -0,0 +1,128 @@
...
+}
+```
+
+For workflows that support continue-as-new, thread a
```
yuandrew: Is "support" here misleading? Should we say "use" here?

contrib/workflowstreams/README.md
```diff
@@ -0,0 +1,128 @@
...
+`*workflowstreams.WorkflowStreamState` field through your workflow input and
+pass it to `NewStream` — it is `nil` on a fresh start and carries accumulated
+state across continue-as-new boundaries:
+
+```go
+return stream.ContinueAsNew(ctx, MyWorkflow, func(state *workflowstreams.WorkflowStreamState) []any {
+	return []any{state}
```
yuandrew: When would someone want to add different logic into this callback?

contrib/workflowstreams/README.md
```diff
@@ -0,0 +1,128 @@
...
+	return workflow.Await(ctx, func() bool { return done })
+}
+```
+
+For workflows that support continue-as-new, thread a
+`*workflowstreams.WorkflowStreamState` field through your workflow input and
+pass it to `NewStream` — it is `nil` on a fresh start and carries accumulated
+state across continue-as-new boundaries:
+
+```go
```
yuandrew: It took me a bit to connect what this code sample was saying based off of the paragraph above it. I think it could use more clarity/explanation. The paragraph above talks about `NewStream`, but then the code sample itself shows the call for CaN. I can see others struggle to make the connection

contrib/workflowstreams/README.md
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return err
+	}
+	defer c.Close(ctx)
+
+	topic := c.Topic("events")
+	for i := range 100 {
+		topic.Publish(fmt.Sprintf("item %d", i), false)
+	}
+	return nil // Close flushes the remaining buffer
```
yuandrew: `// Close flushes the remaining buffer`

This comment seems more appropriate on the `defer c.Close(ctx)` line itself?

contrib/workflowstreams/README.md
```diff
@@ -0,0 +1,128 @@
...
+c.Topic("events").Publish("from outside", true /* forceFlush */)
+```
+
+Items are buffered and flushed automatically every `BatchInterval` (default 2s),
+when the buffer reaches `MaxBatchSize`, on `forceFlush`, on an explicit
+`Flush`, or on `Close`.
+
+## Subscribing
+
+`Subscribe` returns a range-over-func iterator (Go 1.23+):
```
yuandrew: I don't think we need to mention the Go version for this, `go.mod` already shows the required Go version for this package

internal/internal_utils.go
```diff
@@ -33,6 +34,10 @@ const (
 	temporalPrefixError = "__temporal_ is a reserved prefix"
 )
 
+func isReservedNamePrefixException(name string) bool {
```
yuandrew: If this function is specifically to check for workflow streaming reserved names, the function name should be more specific. The current name makes me think it's checking `__temporal_` prefix

contrib/workflowstreams/workflow.go
```diff
@@ -0,0 +1,304 @@
...
+	publisherLastSeen map[string]float64
+	draining          bool
+
+	topicHandles map[string]*WorkflowTopicHandle
+}
+
+// NewStream constructs a WorkflowStream and registers its signal, update, and
+// query handlers on the current workflow. Pass priorState (which may be nil) to
+// restore state carried across a continue-as-new boundary.
+func NewStream(ctx workflow.Context, priorState *WorkflowStreamState) (*WorkflowStream, error) {
```
yuandrew: This should be `NewWorkflowStream`, since it's creating a `WorkflowStream`

contrib/workflowstreams/types.go
```diff
@@ -0,0 +1,107 @@
...
+	// update when the requested offset has already been truncated.
+	ErrTypeTruncatedOffset = "TruncatedOffset"
+	// ErrTypeTruncateOutOfRange is the ApplicationError type returned by
+	// Truncate when the requested offset is past the end of the log.
+	ErrTypeTruncateOutOfRange = "TruncateOutOfRange"
+	// ErrTypeStreamDraining is the ApplicationError type the poll update's
+	// validator returns while the stream is detaching for continue-as-new. It
+	// tells a subscriber the rollover is in progress so it retries (rather than
+	// surfacing an error) until the poll lands on the successor run.
+	ErrTypeStreamDraining = "StreamDraining"
```
yuandrew, line 36: Codex gave me feedback that this is a Go only error, and that cross-lang behaviour may not work as intended, 
>   - Go workflow + Go subscriber: works
>  - Go workflow + Python/TS subscriber: likely surfaces StreamDraining as an unexpected update failure
>  - Python/TS workflow + Go subscriber: Python/TS reject draining polls with generic errors, not StreamDraining, so Go likely surfaces the error instead of retrying

contrib/workflowstreams/publisher.go
```diff
@@ -0,0 +1,302 @@
...
+	return nil
+}
+
+func (p *publisher) encodeBuffer(items []bufItem) ([]PublishEntry, error) {
+	out := make([]PublishEntry, len(items))
+	for i, it := range items {
+		payload, ok := it.value.(*commonpb.Payload)
+		if !ok {
+			var err error
+			payload, err = p.dc.ToPayload(it.value)
```
yuandrew, line 209: In Go, DataConverter.ToPayload may include payload codecs (i.e. [converter.NewCodecDataConverter](https://github.com/temporalio/sdk-go/blob/main/converter/codec.go#L146)). But the docs say

> // The data converter codec chain (encryption, compression) runs once on the
> // signal/update envelope that carries each batch — not per item — so payloads
> // are never double-encoded.

If a caller passes a codec DC in workflowstreams.Options, each item would get double-encoded. 

Since Go `converter.DataConverter` may include codecs, accepting it here makes it easy for callers to pass a codec-enabled converter and accidentally run codecs per item. Is there a way this package can use only the payload conversion portion for item construction, matching Python/TypeScript?

contrib/workflowstreams/workflow.go
```diff
@@ -0,0 +1,304 @@
...
+// Name returns the topic name.
+func (h *WorkflowTopicHandle) Name() string { return h.name }
+
+// Publish appends value to the stream on this topic. value goes through the
+// default data converter; a pre-built *commonpb.Payload bypasses conversion.
+func (h *WorkflowTopicHandle) Publish(value any) error {
+	payload, ok := value.(*commonpb.Payload)
+	if !ok {
+		var err error
+		payload, err = converter.GetDefaultDataConverter().ToPayload(value)
```
yuandrew: This uses the default data converter, but if a workflow uses a custom data converter via `workflow.WithDataConverter`, this stream
  publish path would ignore it. Can we use the workflow context’s configured converter when serializing workflow-published values here?

contrib/workflowstreams/workflow.go
```diff
@@ -0,0 +1,304 @@
...
+type internalEntry struct {
+	topic   string
+	payload *commonpb.Payload
+}
+
+// WorkflowStream is the workflow-side stream object: an append-only log served
+// to external publishers (via signal), subscribers (via update), and offset
+// queries (via query).
+//
+// Construct it once at the start of your workflow function with New. The
```
yuandrew:
```suggestion
// Construct it once at the start of your workflow function with NewWorkflowStream. The
```

contrib/workflowstreams/workflow.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// DetachPollers unblocks all waiting poll handlers and rejects new polls. Used
+// before continue-as-new.
+func (s *WorkflowStream) DetachPollers() {
+	s.draining = true
+}
+
+// GetState returns a serializable snapshot of stream state for continue-as-new.
+// Publisher dedup entries idle longer than publisherTTL are pruned.
```
yuandrew: nit: This sentence feels hard to understand, something like 
```
  // GetState returns a serializable snapshot of stream state for continue-as-new.
  // It drops per-publisher sequence tracking for publishers that have not sent a
  // batch within publisherTTL.
```

could be more clear

contrib/workflowstreams/workflow.go
```diff
@@ -0,0 +1,304 @@
...
+// Returning the result from your workflow function ends the current run:
+//
+//	return stream.ContinueAsNew(ctx, MyWorkflow, func(state *workflowstreams.WorkflowStreamState) []any {
+//		return []any{state}
+//	})
+//
+// State is captured with the default 15-minute publisher TTL. For a custom TTL,
+// use the manual recipe: DetachPollers, Await(AllHandlersFinished), GetState,
+// then workflow.NewContinueAsNewError.
+func (s *WorkflowStream) ContinueAsNew(ctx workflow.Context, wfn any, buildArgs func(state *WorkflowStreamState) []any) error {
```
yuandrew: Should this be named `NewContinueAsNewError` to mirror `workflow.NewContinueAsNewError`? 

The current `ContinueAsNew` matches Python/TS, but the Go method returns an error that callers must return. On the other hand, it does more than construct an error because it detaches pollers and waits for handlers, so `New...` may understate the side effects.

contrib/workflowstreams/README.md
brianstrauch: It's supposed to be a generic exit condition, set by a signal for example. Let me add an explanation to the comment.

contrib/workflowstreams/README.md
brianstrauch: The idea is to give the user a place to combine the workflow streams CAN state with their own CAN state. Typically you would just combine these into a list of two, but you might want something more complex like agent history compaction.

brianstrauch: Updating the text to make that more clear...

contrib/workflowstreams/README.md
brianstrauch: Agreed, broke this out into two steps for clarity.

contrib/workflowstreams/workflow.go
brianstrauch: Went ahead with the rename. Cross-SDK compatibility is less important than maintaining existing patterns from what I can tell. Most developers are just using one Temporal SDK.

contrib/workflowstreams/types.go
brianstrauch: Nice find, I have some PRs for this here:
* https://github.com/temporalio/sdk-python/pull/1581
* https://github.com/temporalio/sdk-typescript/pull/2102

yuandrew (approved):

PR #2388, opened 2026-06-04
Fix upload/download stats to use wall-clock duration

### LINKED ISSUE #2378: [Bug] External storage transfer stats reports summation of durations instead of wall-clock duration
### What are you really trying to do?

Get more accurate transfer statistics for external storage uploads and downloads.

### Describe the bug

External storage transfer stats are summing the durations of each driver invocation instead of tracking overlapping intervals. The stats should be reporting wall-clock time rather than duration summation. See https://github.com/temporalio/sdk-go/blob/v1.44.1/internal/internal_task_pollers.go#L883. For example, if 3 drivers are concurrently invoked during retrieval and each takes about 100 ms, the reported duration will be 300 ms instead of the expected value that is much closer to 100 ms. 

### Possible solution

Use edge detection to track and record durations.

amit-writes-code:
Use edge detection (Start/Stop) to bracket visitProtoPayloads rather than
summing individual PayloadBatchCompleted durations

Fixes #2378

## What was changed

**`internal/internal_task_pollers.go`**

- Replaced `totalDuration time.Duration` field on `workflowTaskStorageMetrics`
  with `startTime` and `endTime time.Time` fields.
- Added `Start()` — records wall-clock start immediately before
  `visitProtoPayloads` is called (rising edge).
- Added `Stop()` — records wall-clock end immediately after
  `visitProtoPayloads` returns (falling edge), before any subsequent
  workflow execution or network calls.
- Added `WallClockDuration()` — returns `endTime.Sub(startTime)`, a fixed
  interval between two captured timestamps. Replaces `totalDuration` at
  both log sites (`tagPayloadDownloadDuration`, `tagPayloadUploadDuration`).
- `PayloadBatchCompleted` no longer accumulates `duration` — it only tracks
  `payloadCount`, `totalSize`, and `driverNames`.

**`internal/internal_task_pollers_test.go`**

- `TestWorkflowTaskStorageMetrics_WallClockDuration` — verifies the
  Start/Stop/WallClockDuration lifecycle: returns 0 before `Start()`,
  returns 0 before `Stop()`, returns a positive duration after `Stop()`,
  does not shrink on a second `Stop()` call.
- `TestWorkflowTaskStorageMetrics_PayloadBatchCompleted` — simulates two
  concurrent callbacks and asserts that `payloadCount` and `totalSize`
  accumulate correctly, driver names are collected from both callbacks,
  and wall-clock duration is less than the sum of individual durations
  (proving edge detection rather than summation).

## Why?

`PayloadBatchCompleted` was called once per `Visit` invocation with a
`duration` argument representing the wall-clock time for that Visit.
When `payloadVisitorConcurrency > 1`, multiple Visit calls run
concurrently. Summing their individual durations overcounts proportionally
to the number of concurrent visits:

```
Visit 1: t=0ms  → t=100ms   duration=100ms
Visit 2: t=10ms → t=120ms   duration=110ms   (runs concurrently with Visit 1)

Old: totalDuration = 100 + 110 = 210ms  ← wrong, they overlapped
New: Stop() − Start() = 120ms − 0ms = 120ms  ← correct wall-clock
```

## Checklist

1. Closes #2378
2. How was this tested:
   - `TestWorkflowTaskStorageMetrics_WallClockDuration` — unit test for
     the Start/Stop/WallClockDuration lifecycle.
   - `TestWorkflowTaskStorageMetrics_PayloadBatchCompleted` — concurrent
     callbacks assert wall-clock < sum of individual durations.
   - Run: `go test ./internal/ -run "TestWorkflowTaskStorageMetrics" -v`
3. Any docs updates needed? No — internal observability change only, no
   public API or user-visible behavior change.

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -459,6 +459,7 @@ func (wtp *workflowTaskProcessor) processWorkflowTask(task *workflowTask) (retEr
 	defer close(doneCh)
 
 	downloadPayloadMetrics := &workflowTaskStorageMetrics{}
+	downloadPayloadMetrics.Start() // start wall-clock before visits begin
 	ctx := extstore.WithStorageOperationCallback(context.Background(), downloadPayloadMetrics)
 
 	var taskErr error
@@ -466,6 +467,7 @@ func (wtp *workflowTaskProcessor) processWorkflowTask(task *workflowTask) (retEr
 		wtp.handleInboundVisitorError(task.task, taskErr)
 		return nil
 	}
+	downloadPayloadMetrics.Stop() // capture end time before any other work
 
 	wfctx, err := wtp.contextManager.GetOrCreateWorkflowContext(task.task, task.historyIterator)
 	if err != nil {
@@ -589,6 +591,7 @@ func (wtp *workflowTaskProcessor) RespondTaskCompletedWithMetrics(
 	}
 
 	uploadPayloadMetrics := &workflowTaskStorageMetrics{}
+	uploadPayloadMetrics.Start() // start wall-clock before visits begin
 	ctx := extstore.WithStorageOperationCallback(context.Background(), uploadPayloadMetrics)
 	ctx = extstore.WithStorageTarget(ctx, extstore.StorageDriverWorkflowInfo{
 		Namespace:    wtp.namespace,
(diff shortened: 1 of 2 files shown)


amit-writes-code:
@jmaeagle99 would love to know if there is any scope of optimzations here, in terms of accuracy

jmaeagle99:
Hi @amit-writes-code, I apologize for not getting back to you in a timely manner.

This isn't quite the shape of the fix that we would like because it forces the holder of the `workflowTaskStorageMetrics` to start/stop the metric counting. It also makes it difficult to use the same instance for recording subsequent invocations of external storage. Finally, we might want to extend this further to capture the wall-clock durations of individual drivers and record them via this mechanism, which using a public stop/stop invocation would not be able to accurately capture.

I've opened a new PR which makes the duration measurement self-contained: #2658

Issue #2191, opened 2026-02-13
Expose whether a task queue is sticky to slot supplier.


billrich2001:
**Is your feature request related to a problem? Please describe.**
A custom slot supplier cannot determine if a TQ is sticky when reserving slots

**Describe the solution you'd like**
Expose stickiness to slot supplier

**Additional context**
All other Temporal SDKs expose this to SlotSupplier

PR #2389, opened 2026-06-05
Refactor poller autoscaling

yuandrew:
## What was changed

Dynamic polling behavior was moved out into its own logic. Fundamentally, we no longer create a set number of goroutines (the `maximum` for autoscaling) on worker start. We instead create a `pollerAutoscaler` manager than spins up goroutines after reserving a worker slot for each poll RPC.

With this refactor, `pollerSemaphore` is no longer needed, autoscaling is now controlled by `dynamicScalableTaskPollerRunner.active` vs `pollerAutoscaler.target`.

Was also able to now specify task queue type slot reservation with autoscaling, although simplemax is still unspecified because they are mixed at reservation time.

NOTE: The `PollerStartCounter` metric was removed for autoscaling, it's not really a metric that makes sense for autoscaling, and there wasn't a good way to maintain its pre-refactor behavior.

## Why?
Poller autoscaling was originally designed to mirror the existing polling structure of creating a set number of goroutines (the `maximum` for autoscaling) on worker start.

We're now aligning with Rust on creating new goroutines on each poll RPC

## Checklist

1. Closes 
Partially addresses https://github.com/temporalio/sdk-go/issues/2191 for the autoscaling case, SimpleMax still doesn't specify TQ type

2. How was this tested:
  - Updates autoscaling tests to assert active dynamic poll count instead of semaphore permits.
  - Adds coverage that autoscaling uses the dynamic runner while SimpleMaximum does not.
  - Adds shutdown regression coverage to ensure tasks returned during shutdown are still processed.
  - Adds balancer regression coverage to ensure autoscaling does not hold worker slots while blocked waiting for another
    poller type.

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes core worker polling concurrency and shutdown/drain paths; mistakes could starve tasks or leak slots, though coverage targets balancer blocking and shutdown processing.
> 
> **Overview**
> Refactors **poller autoscaling** so workers no longer start a fixed pool of goroutines up to the autoscaling maximum. A single manager loop (`runAutoscalingPoller`) waits on an `autoscalingTaskPollerRunner` target, reserves a worker slot, then spawns short-lived goroutines per poll RPC—replacing `pollerSemaphore` and the old `pollScalerReportHandle` / `scaleCallback` permit model (`pollerAutoscaler` + `targetChangedCallback`).
> 
> **Slot reservation** now carries `taskQueueKind` on `SlotReservationInfo` and per-poller overrides (`taskQueueKindForPoller`): autoscaling workflow pollers split normal vs sticky; activity/nexus/local activity use normal; SimpleMaximum workflow pollers stay **unspecified** because queue choice happens after reserve. Workers set default kinds in `slotReservationData` where applicable.
> 
> **Behavior fixes:** balancer runs before slot reserve on the autoscaling path so pollers do not hold slots while blocked on capacity or other poller types; shared `reserveSlotAsync` extracts slot reservation. **PollerStartCounter** is no longer incremented for autoscaling pollers. Tests were renamed/expanded around active poll counts, task queue kind propagation, shutdown drain, and slot-holding regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c06ea6f793bf62c07b364a0082ea245c92a19ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	commonpb "go.temporal.io/api/common/v1"
+	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
 	"golang.org/x/time/rate"
 	"google.golang.org/grpc/codes"
@@ -187,10 +188,10 @@ type (
 		taskPollerType string
 		// pollerCount is the number of pollers tasks to start. There may be less than this
 		// due to limited slots, rate limiting, or poller autoscaling.
-		pollerCount      int
-		taskPoller       taskPoller
-		pollerAutoscaler *pollerAutoscaler
-		dynamicRunner    *dynamicScalableTaskPollerRunner
+		pollerCount       int
+		taskPoller        taskPoller
+		pollerAutoscaler  *pollerAutoscaler
+		autoscalingRunner *autoscalingTaskPollerRunner
 	}
 
 	// baseWorkerOptions options to configure base worker.
@@ -289,7 +290,7 @@ type (
 
 	barrier chan struct{}
 
-	dynamicScalableTaskPollerRunner struct {
+	autoscalingTaskPollerRunner struct {
 		autoscaler *pollerAutoscaler
 		wakeCh     chan struct{}
 		activeMu   sync.Mutex
@@ -422,7 +423,7 @@ func (bw *baseWorker) Start() {
 			bw.pollerBalancer.registerPollerType(taskWorker.taskPollerType)
 		}
 
-		if taskWorker.dynamicRunner != nil {
+		if taskWorker.autoscalingRunner != nil {
 			bw.stopWG.Add(1)
 			bw.pollerWG.Add(1)
 			go bw.runAutoscalingPoller(taskWorker)
@@ -499,7 +500,7 @@ func (bw *baseWorker) runPoller(taskWorker scalableTaskPoller) {
 			}
 		}
 
-		bw.reserveSlotAsync(ctx, reserveChan)
+		bw.reserveSlotAsync(ctx, reserveChan, taskWorker)
 
 		select {
 		case <-bw.stopCh:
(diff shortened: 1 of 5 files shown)


Sushisource (commented):
Overall, looking good to me. Let's get a nice long run going with Thomas' tool to make sure there are no regressions.

As part of this change, can we also include the sticky/non-sticky info for the slot supplier reservation call? That's ask (there's an issue for it somewhere) we solve with this.

internal/internal_worker_base.go
```diff
@@ -622,6 +628,30 @@ func (bw *baseWorker) reserveSlotAsync(ctx context.Context, reserveChan chan<- *
...
+func taskQueueKindForPoller(taskWorker scalableTaskPoller) (enumspb.TaskQueueKind, bool) {
+	switch taskWorker.taskPollerType {
+	case metrics.PollerTypeWorkflowStickyTask:
+		return enumspb.TASK_QUEUE_KIND_STICKY, true
+	case metrics.PollerTypeWorkflowTask:
+		// Autoscaling workflow pollers are split by queue kind. SimpleMaximum
+		// workflow pollers are mixed, so their kind is not known at reservation time.
+		if taskWorker.autoscalingRunner != nil {
+			return enumspb.TASK_QUEUE_KIND_NORMAL, true
+		}
```
Sushisource, line 651: Hrm... I was hoping the refactor would make it easier in both cases but I suppose it makes sense that it didn't. 🤔

internal/tuning.go
```diff
@@ -46,6 +47,10 @@ type SlotReservationInfo interface {
 	// TaskQueue returns the task queue for which a slot is being reserved. In the case of local
 	// activities, this is the same as the workflow's task queue.
 	TaskQueue() string
+	// TaskQueueKind returns the kind of task queue for which a slot is being reserved.
+	// Workflow task reservations return unspecified for SimpleMaximum pollers because
+	// they may choose between normal and sticky queues after slot reservation.
```
Sushisource, line 52: Comment is wrong - we always return normal not unspecified

yuandrew: WFT reservations for SimpleMax currently only return unspecified, do you want to change that to always return normal? Unspecified seems more accurate here (unable to differentiate) than always saying normal, even if it sometimes is sticky

```
taskQueueKind: enumspb.TASK_QUEUE_KIND_UNSPECIFIED,
```

Sushisource: Ah yeah I read the if on my other comment backwards. In this case yeah I think it makes sense to keep it how it is then.

Sushisource (approved):

PR #2390, opened 2026-06-05
Add gzip compression on by default

Sushisource:
## What was changed
Add and default enable transport-level gzip compression

## Why?
Saves wire space, being done in all SDKs

## Checklist

1. Closes 

2. How was this tested:
Simple test that headers are enabled. Core has more comprehensive tests confirming server actually responds with compressed values when such headers are set. We could replicate those here too if we feel it's necessary.

3. Any docs updates needed?

### DIFF
--- a/internal/grpc_dialer.go
+++ b/internal/grpc_dialer.go
@@ -103,9 +103,9 @@ func dial(params dialParameters) (*grpc.ClientConn, error) {
 	opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxPayloadSize)))
 
 	switch compression := params.UserConnectionOptions.GrpcCompression.(type) {
-	case nil, GrpcCompressionGzip:
+	case nil, *GrpcCompressionGzip:
 		opts = append(opts, grpc.WithDefaultCallOptions(grpc.UseCompressor(grpcgzip.Name)))
-	case GrpcCompressionNone:
+	case *GrpcCompressionNone:
 	default:
 		return nil, fmt.Errorf("unsupported gRPC compression option %T", compression)
 	}
--- a/internal/grpc_dialer.go
+++ b/internal/grpc_dialer.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/grpc_dialer.go
```diff
@@ -100,6 +102,14 @@ func dial(params dialParameters) (*grpc.ClientConn, error) {
 	opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallSendMsgSize(maxPayloadSize)))
 	opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxPayloadSize)))
 
+	switch compression := params.UserConnectionOptions.GrpcCompression.(type) {
```
yuandrew, line 105: For user-supplied closed-set config options that are later type-switched by the SDK, I think the safer pattern is pointer-only implementations, like [versioningOverride](https://github.com/temporalio/sdk-go/blob/3edbf0f52827e4055b50ac39b60da4c5fe5ab8fe/internal/internal_workflow_execution_options.go#L133) or [scheduleAction](https://github.com/temporalio/sdk-go/blob/3edbf0f52827e4055b50ac39b60da4c5fe5ab8fe/internal/internal_schedule_client.go#L606C19-L606C33)

Currently, `GrpcCompression: &client.GrpcCompressionNone{}`, compiles but would fail at runtime, since it would go through the `default` path.

dnr (commented):
I mentioned some of this in DM but ftr:

I'm a little concerned about the speed and cpu overhead of gzip, and I really wish a more modern faster algorithm was available here (lz4, snappy, or zstd's negative levels, are commonly considered more appropriate for transport compression in a datacenter setting).

Can we do a load test with a lot of schedules (v1) for example?

Sushisource:
> I mentioned some of this in DM but ftr:
> 
> I'm a little concerned about the speed and cpu overhead of gzip, and I really wish a more modern faster algorithm was available here (lz4, snappy, or zstd's negative levels, are commonly considered more appropriate for transport compression in a datacenter setting).
> 
> Can we do a load test with a lot of schedules (v1) for example?

I did a run with the core version and the change in CPU usage was negligible, SDK side. Are you thinking of tests on the server end?

dnr:
> I did a run with the core version and the change in CPU usage was negligible, SDK side. Are you thinking of tests on the server end?

I thought schedules v1 would be a good test because the server worker (using go sdk) is doing a lot of work, plus the server itself. It's definitely worth trying at a larger scale but maybe we don't need a specific load test right now. We should look at the cpu metrics in the release validation pipelines after this gets integrated.

Sushisource:
@dnr Traffic team took a look at dashboards after running w/ this change and it appears the impact is ~2% CPU increase on the frontend

internal/grpc_dialer.go
Sushisource: Done!

yuandrew (approved):

PR #2392, opened 2026-06-08
Test poller target halves on ResourceExhausted

veeral-patel:
## What

Adds `TestAutoscalingHalvesOnResourceExhausted` to `ScalableTaskPollerSuite` in `internal/internal_worker_base_test.go`.

The test drives `pollScalerReportHandle.handleError` (built through the production `newScalableTaskPoller` wiring) and asserts the **live `pollerSemaphore` max permits** follow the halving sequence:

| Error | Target | maxPermits |
|---|---|---|
| `ResourceExhausted` | 16 → 8 | 8 |
| `ResourceExhausted` | 8 → 4 | 4 |
| other (`Internal`) | 4 → 3 | 3 |
| `ResourceExhausted` | 3 → 1 | 1 (clamped at min) |
| `ResourceExhausted` | 1 → 1 | 1 |

## Why

Asserting the exact step size — `16→8→4` for `ResourceExhausted` versus `4→3` for a non-`ResourceExhausted` error — is what distinguishes **halving (`t/2`)** from a plain `-1` decrement. A regression in either direction (RE stops halving, or halving leaks to other error types) fails the test.

This complements the existing `PollScalerReportHandleSuite.TestErrorScaleDown`, which asserts the same arithmetic against a **mock `scaleCallback`**. The new test additionally verifies the halved value reaches the **real `pollerSemaphore`** through the `scaleCallback → updatePermits` wiring that `newScalableTaskPoller` sets up.

## Notes

- Deterministic and fast; no Temporal server required.
- Reuses the suite's existing `semaphoreProbeTaskPoller` and `readSemaphoreState` helpers; no production code changes.

## Test

```
go test ./internal/ -run 'TestScalableTaskPollerSuite/TestAutoscalingHalvesOnResourceExhausted' -count=1
```

### DIFF
--- a/internal/internal_worker_base_test.go
+++ b/internal/internal_worker_base_test.go
@@ -222,6 +222,60 @@ func (s *ScalableTaskPollerSuite) TestAutoscalingScalesDownToMinimum() {
 	}, 200*time.Millisecond, 10*time.Millisecond, "should not scale below minimum")
 }
 
+// TestAutoscalingHalvesOnResourceExhausted verifies that a ResourceExhausted poll
+// error HALVES the poller target (t/2) - as opposed to the -1 decrement applied to
+// other errors - and that the halved value reaches the live pollerSemaphore via the
+// scaleCallback wiring. Asserting the exact step sequence (16 -> 8 -> 4, then 4 -> 3
+// for a non-RE error, then 3 -> 1) is what distinguishes halving from a plain
+// scale-down: a -1 decrement would yield 15, 7, etc., failing these assertions.
+//
+// PollScalerReportHandleSuite.TestErrorScaleDown covers the same arithmetic against
+// a mock scaleCallback; this test additionally proves the real semaphore is resized.
+func (s *ScalableTaskPollerSuite) TestAutoscalingHalvesOnResourceExhausted() {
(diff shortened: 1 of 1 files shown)


internal/internal_worker_base_test.go
```diff
@@ -222,6 +222,60 @@ func (s *ScalableTaskPollerSuite) TestAutoscalingScalesDownToMinimum() {
...
+		minimumNumberOfPollers: 1,
+	}
+
+	// The server advertising poller autoscaling is what enables error-driven
+	// scale-down (the everSawScalingDecision || serverSupportsAutoscaling guard
+	// in handleError). Simulate that here.
+	serverSupportsAutoscaling := &atomic.Bool{}
+	serverSupportsAutoscaling.Store(true)
+
+	poller := newScalableTaskPoller(&semaphoreProbeTaskPoller{}, ilog.NewNopLogger(), behavior, serverSupportsAutoscaling)
```
yuandrew, line 248: Can we use `newSemaphoreProbeTaskPoller` here? All of the other tests use this helper to create the poller

yuandrew (commented):
looks like `go run . check` is failing

PR #2393, opened 2026-06-08
Add temporal_poller_target gauge metric

veeral-patel:
## What

Adds a `temporal_poller_target` gauge that reports the **poller-autoscaling target** (the desired number of pollers the autoscaler has settled on), tagged by `poller_type` — the production-observability counterpart to the existing `temporal_num_pollers` (the *actual* number of pollers currently inside a poll RPC).

## Why

Today operators can only see `temporal_num_pollers`, which reflects how many pollers are mid-poll at an instant — it collapses under poll backoff and idle long-polls and does **not** reflect the autoscaling decision. 

There's no way to see whether autoscaling is scaling up, scaling down, or pinned at min/max. 

`temporal_poller_target` makes the decision observable, and pairs with `num_pollers` for a target-vs-actual dashboard.

## Tags

`poller_type` ∈ {`workflow_task`, `workflow_sticky_task`, `activity_task`, `nexus_task`} — same convention as `num_pollers`.


veeral-patel:
Closing, won't be able to merge due to difficulty in adding new worker side metrics.

Created this related PR to add poller_scale_decision metric on the server side: https://github.com/temporalio/temporal/pull/11084

PR #2394, opened 2026-06-08
Implement nexus-delivered activity cancels

Sushisource:
## What was changed
Implement activity cancellations being delivered on the new Nexus command channel for workers

## Why?
More reliable activity cancellation, parity.

## Checklist

1. Closes 

2. How was this tested:
Added unit / integration tests

3. Any docs updates needed?


internal/internal_worker.go
```diff
@@ -228,6 +228,9 @@ type (
 
 		workerInstanceKey string
 
+		workerControlTaskQueue        string
+		activityCancellationCallbacks *activityCancellationCallbacks
```
yuandrew, line 233: nit: all struct params are separated by a blank line

internal/internal_worker.go
```diff
@@ -2266,6 +2273,12 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 	// All worker systems that depend on the capabilities to process workflow/activity tasks
 	// should take a pointer to this struct and wait for it to be populated when the worker is run.
 	var capabilities workflowservice.GetSystemInfoResponse_Capabilities
+	var activityCancellationCallbacks *activityCancellationCallbacks
+	if client.heartbeatManager != nil {
+		activityCancellationCallbacks = client.heartbeatManager.
+			sharedNamespaceWorkerFor(client.namespace).
+			activityCancellationCallbacks
```
yuandrew, line 2284: I hate that Go keeps the `.` at the end of each line...

internal/internal_worker.go
```diff
@@ -2422,7 +2437,6 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 			Name: sessionCompletionActivityName,
 		})
 	}
-
```
yuandrew, line 2428: nit: This line seemed fine

internal/internal_worker_heartbeat.go
```diff
@@ -39,6 +47,43 @@ func newHeartbeatManager(client *WorkflowClient, interval time.Duration, logger
...
 }
 
+func (m *heartbeatManager) sharedNamespaceWorkerFor(namespace string) *sharedNamespaceWorker {
+	m.workersMutex.Lock()
+	defer m.workersMutex.Unlock()
+	return m.sharedNamespaceWorkerForLocked(namespace)
+}
+
+func (m *heartbeatManager) sharedNamespaceWorkerForLocked(namespace string) *sharedNamespaceWorker {
+	hw := m.workers[namespace]
```
yuandrew, line 57: seems like we can keep the `m.workersMutex.Lock()`, inside `sharedNamespaceWorkerForLocked`?

internal/internal_worker_heartbeat.go
```diff
@@ -156,7 +204,9 @@ func (hw *sharedNamespaceWorker) sendHeartbeats() error {
 	hw.callbacksMutex.RLock()
 	callbacks := make([]func() *workerpb.WorkerHeartbeat, 0, len(hw.callbacks))
 	for _, cb := range hw.callbacks {
-		callbacks = append(callbacks, cb)
+		if cb != nil {
```
yuandrew, line 207: when would `cb` be nil? Can we instead ensure we don't register nil callbacks?

internal/internal_worker_heartbeat.go
```diff
@@ -186,6 +236,125 @@ func (hw *sharedNamespaceWorker) sendHeartbeats() error {
 	return nil
 }
 
+func (hw *sharedNamespaceWorker) runWorkerCommands() {
+	for {
+		select {
+		case <-hw.heartbeatCtx.Done():
```
yuandrew: now that we're using this for worker commands ctx, should we rename it more generally? Like `workerCtx` or something?

internal/internal_worker_heartbeat.go
```diff
@@ -186,6 +236,125 @@ func (hw *sharedNamespaceWorker) sendHeartbeats() error {
...
+			}
+			hw.logger.Warn("Failed polling worker command task", "Error", err)
+			select {
+			case <-time.After(time.Second):
+			case <-hw.heartbeatCtx.Done():
+				return
+			}
+			continue
+		}
+		if task == nil || len(task.TaskToken) == 0 || task.GetRequest() == nil {
```
yuandrew: I get that the first 2 in this chain mirror nexus polling, but why do we silently continue on `task.GetRequest() == nil`? Wouldn't that be a malformed poll response? Should we log or something?

yuandrew: Codex says Rust logs "Worker command nexus task missing request"

internal/internal_worker_heartbeat.go
```diff
@@ -186,6 +236,125 @@ func (hw *sharedNamespaceWorker) sendHeartbeats() error {
...
+		}
+	}
+
+	responsePayload, err := proto.Marshal(&workerservicepb.ExecuteCommandsResponse{Results: results})
+	if err != nil {
+		return fmt.Errorf("encode ExecuteCommandsResponse: %w", err)
+	}
+
+	grpcCtx, cancel := newGRPCContext(
+		context.Background(),
```
yuandrew, line 335: How come we use context.Background() here, where above we us `hw.heartbeatCtx`?

internal/internal_worker.go
Sushisource: Me too lol

internal/internal_worker_heartbeat.go
Sushisource: No - the point of the locked version is that you're expected to already hold the lock when it's called

internal/internal_worker_heartbeat.go
Sushisource: If worker heartbeating is disabled in general seems to be the only case. But, as a bit of defensive programming I think it's reasonable.

internal/internal_worker_heartbeat.go
Sushisource: It would allow any in-progress completion to finish instead of cancelling it during worker shutdown.

internal/internal_worker_heartbeat.go
yuandrew: right. I was thinking too simplicatically that this just needed to be held while we use `m.workers`. But in reality, we need to hold the lock across the full namespace-worker lifecycle update so registration can’t race with unregister stopping and deleting the shared worker.

I wonder if we need to add a docstring to m.workersMutex or a comment somewhere to make this more obvious

yuandrew (approved):
LGTM, up to you if you want to add the comment, I might just add it in a follow up PR if not here

Sushisource:
>I was thinking too simplicatically

I too often think simplicatically :laughing: 

I'll let you add the comment though. I'm not sure exactly what you have in mind. I'd just be adding a comment like "hold the lock while modifying things in workers" but that feels obvious

PR #2397, opened 2026-06-11
Adding links to Nexus signals

Evanthx:
This adds links to Nexus signals - the receiving workflow has a link to the workflow sending the signal. and the sending workflow has a link to the receiving workflow.

### DIFF
--- a/internal/link_converter.go
+++ b/internal/link_converter.go
@@ -1,0 +1,228 @@
+// The WorkflowEvent <-> Nexus link conversion logic in this file is duplicated in
+// temporalio/temporal/components/nexusoperations/link_converter.go. Any changes here or there must
+// be replicated. This is temporary until the temporal repo updates to the most recent SDK version.
+
+package internal
+
+import (
+	"fmt"
+	"net/url"
+	"regexp"
+	"strconv"
+
+	"github.com/nexus-rpc/sdk-go/nexus"
+	commonpb "go.temporal.io/api/common/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+	nexuspb "go.temporal.io/api/nexus/v1"
+)
+
+const (
+	urlSchemeTemporalKey = "temporal"
+	urlPathNamespaceKey  = "namespace"
+	urlPathWorkflowIDKey = "workflowID"
+	urlPathRunIDKey      = "runID"
+	urlPathTemplate      = "/namespaces/%s/workflows/%s/%s/history"
+
+	linkWorkflowEventReferenceTypeKey = "referenceType"
+	linkEventIDKey                    = "eventID"
+	linkEventTypeKey                  = "eventType"
+	linkRequestIDKey                  = "requestID"
+)
+
+var (
+	rePatternNamespace  = fmt.Sprintf(`(?P<%s>[^/]+)`, urlPathNamespaceKey)
+	rePatternWorkflowID = fmt.Sprintf(`(?P<%s>[^/]+)`, urlPathWorkflowIDKey)
+	rePatternRunID      = fmt.Sprintf(`(?P<%s>[^/]+)`, urlPathRunIDKey)
+	urlPathRE           = regexp.MustCompile(fmt.Sprintf(
+		`^/namespaces/%s/workflows/%s/%s/history$`,
+		rePatternNamespace,
+		rePatternWorkflowID,
+		rePatternRunID,
+	))
+	eventReferenceType     = string((&commonpb.Link_WorkflowEvent_EventReference{}).ProtoReflect().Descriptor().Name())
+	requestIDReferenceType = string((&commonpb.Link_WorkflowEvent_RequestIdReference{}).ProtoReflect().Descriptor().Name())
+)
+
+// ConvertLinkWorkflowEventToNexusLink converts a Link_WorkflowEvent type to Nexus Link.
+//
+// NOTE: Experimental
+func ConvertLinkWorkflowEventToNexusLink(we *commonpb.Link_WorkflowEvent) nexus.Link {
+	u := &url.URL{
+		Scheme: urlSchemeTemporalKey,
+		Path:   fmt.Sprintf(urlPathTemplate, we.GetNamespace(), we.GetWorkflowId(), we.GetRunId()),
+		RawPath: fmt.Sprintf(
+			urlPathTemplate,
+			url.PathEscape(we.GetNamespace()),
+			url.PathEscape(we.GetWorkflowId()),
+			url.PathEscape(we.GetRunId()),
+		),
+	}
+
+	switch ref := we.GetReference().(type) {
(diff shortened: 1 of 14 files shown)


openapitools.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "./node_modules/@openapitools/openapi-generator-cli/config.schema.json",
```
chrsmith: Is this something we should check into the repo? Or did it just happen to be part of your personal dev flow?

I don't have strong opinions, but if this isn't something we expect to use for regular edits to the Golang SDK then maybe this should be removed? And/or added to the `.gitignore`?

internal/nexus_operations.go
```diff
@@ -175,6 +209,33 @@ type nexusOperationLinksKeyType struct{}
...
+// WorkflowEvent variant. Registered by the temporalnexus package via
+// SetNexusLinkToWorkflowEventLinkConverter to avoid an import cycle. The task handler uses it to
+// forward inbound Nexus task links onto the RPCs (e.g. signal, signalWithStart) the handler issues.
+// It returns (nil, false) when the link is not a parseable WorkflowEvent link.
+var nexusLinkToWorkflowEventLink func(*nexuspb.Link) (*commonpb.Link, bool)
+
+// SetNexusLinkToWorkflowEventLinkConverter registers the converter used to turn inbound Nexus task
+// links into common.v1.Links. Called once from the temporalnexus package's init.
+func SetNexusLinkToWorkflowEventLinkConverter(fn func(*nexuspb.Link) (*commonpb.Link, bool)) {
+	nexusLinkToWorkflowEventLink = fn
```
chrsmith: Nit: This is totally fine as-is, but maaaaaaybe we might want to use `sync.Once` and/or panic if we ever see the `SetNexusLinkToWorkflowEventLinkConverter` function is called twice?

openapitools.json
Evanthx: Good catch, added to .gitignore and removed!

internal/nexus_operations.go
Evanthx: It is just called from the init method so won't be called twice - and this is consistent with the other set methods in the package. So I see your point, but unless you disagree I will leave it as is.

temporalnexus/operation.go
```diff
@@ -19,18 +19,56 @@ package temporalnexus
...
+	internal.SetNexusLinkToWorkflowEventLinkConverter(func(link *nexuspb.Link) (*common.Link, bool) {
+		nexusLink := nexus.Link{Type: link.GetType()}
+		if link.GetUrl() != "" {
+			u, err := url.Parse(link.GetUrl())
+			if err != nil {
+				return nil, false
+			}
+			nexusLink.URL = u
+		}
+		we, err := ConvertNexusLinkToLinkWorkflowEvent(nexusLink)
```
yuandrew: We need to add a nil check in `ConvertNexusLinkToLinkWorkflowEvent` for when `link` is nil, specifically this line would panic if `link.URL` is `nil`

`if link.URL.Scheme != urlSchemeTemporalKey {`

and adding a test for this would also be good

temporalnexus/operation.go
```diff
@@ -19,18 +19,56 @@ package temporalnexus
...
 	"go.temporal.io/api/enums/v1"
+	nexuspb "go.temporal.io/api/nexus/v1"
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/internal"
 	"go.temporal.io/sdk/internal/common/metrics"
 	"go.temporal.io/sdk/log"
 	"go.temporal.io/sdk/workflow"
 )
 
+func init() {
```
yuandrew: If I'm understanding this correctly, users would need to import the `temporalnexus` package to get links in Nexus signals support? Would users expect to have this support when they use nexus w/o the `temporalnexus` package?

temporalnexus/operation.go
Evanthx: Check and test added!

temporalnexus/operation.go
Evanthx: I think you are correct - without Nexus, this would just do not do anything. But if I understand this right, then issuing a signal or signalWithStart you have to go through the temporalNexus package to get the nexus client, so that means things would be set up.

yuandrew: > issuing a signal or signalWithStart you have to go through the temporalNexus package to get the nexus client

not necessarily, I had Codex spin up a quick sample to show how signalling can work w/o `temporalnexus` https://github.com/temporalio/samples-go/pull/491 (also for my own learning to prove this was possible haha)

Evanthx: That is a pretty cool test! And it seems like you're right, so I learned something! 

So digging - this seems to only be an issue with Go, introduced by an attempt to avoid an import cycle. The link converters are in temporalnexus, which then eventually allows this to happen. (TemporalNexus imports internal, so I couldn't use it in internal/nexus_operations and it goes from there.)

So any fix gets weird, but it is a hole and seems like a good idea to fix. So first check out link_converter.go - it says:
// The WorkflowEvent <-> Nexus link conversion logic in this file is duplicated in
// temporalio/temporal/components/nexusoperations/link_converter.go. Any changes here or there must
// be replicated. This is temporary until the temporal repo updates to the most recent SDK version.

That note is from a few years ago, so I have a suspicion it is time to remove that temporary fix. But if I follow this right, the Temporal server includes the go SDK, so it seems like this is a change in the Temporal Server as well if we address that note? The server does have a link_converter.go but it's not exactly the same, so that will be fun.

So I basically moved the link converters to internal, which then allows this all to work like it does in the other SDKs. That note from two years ago then shows up in my PR, but it is just from moving code around.

So in brief. I think I fixed this hole. I think a follow-up PR would be to remove components/nexusoperations/link_converter.go from the server and thus unify these - but I don't want to mix that PR with this PR. And -- do you agree with this fix?

Evanthx: I created a PR to unify the link converters - https://temporalio.atlassian.net/browse/NEXUS-469

internal/link_converter.go
```diff
@@ -0,0 +1,228 @@
+// The WorkflowEvent <-> Nexus link conversion logic in this file is duplicated in
+// temporalio/temporal/components/nexusoperations/link_converter.go. Any changes here or there must
+// be replicated. This is temporary until the temporal repo updates to the most recent SDK version.
```
yuandrew, line 3: Can we make sure this is addressed? Like you mentioned, this can be in a separate PR

yuandrew (approved):

internal/link_converter.go
Evanthx: Already created! https://temporalio.atlassian.net/browse/NEXUS-469
I will be out for two weeks starting tomorrow so it'll be a little bit - but this has been in here for years it looks like, so that is probably OK. :-)

PR #2406, opened 2026-06-16
Grant explicit actions:read to features reusable-workflow caller

cconstable:
## What 
Adds a job-level permissions block granting `contents: read` and `actions: read` to the job that calls the temporalio/features workflow. 

## Why
The org default GITHUB_TOKEN permissions are now restrictive (`actions: none`), so without this grant the reusable workflow fails validation at parse time and the check never starts.


yuandrew (approved):
Thanks for fixing this!

PR #2410, opened 2026-06-17
Fix more test flakes, and autoscaling startup race

yuandrew:
## What was changed

Fixed a few tests flakes:
  - Worker poller startup/shutdown: fixed a non-test race in worker startup where autoscaling pollers could begin running before every poller type was registered with the balancer. We now register all poller types before starting any poller goroutines, so one poller type cannot temporarily consume slots before the balancer knows about them.

  - TestAutoscalingScalesDownToMinimum: The flake was a race between the test assertion and the asynchronous autoscaling control loop. Now we only assert after we actually observe/apply the minimum state, instead of a potential transient state of 0 failing our tests, even though the effective poller count never goes below 1

  - TestLocalActivityTaskTimeoutHeartbeat: The flake was stopping the worker the LA was reliably blocked on shutdown, and we were also waiting on `GetWorkflowHistory` indefinitely, potentially up to `15m`. Now we send an explicit signal after LA has started, and added a bounded ctx, and stopped long polling for WF history.

  - TestStartDevServer_SpecificVersion: stopped pinning the obsolete Temporal CLI v0.3.0, whose embedded server can crash during startup with a tchannel-go concurrent map read/write panic. The test now uses CLI v1.6.1, which is still an explicit older version and verifies the specific-version download path without relying on a crash-prone release.

  - TestTaskNotProcessedDuringLegacyShutdown: replaced an unbounded wait for the fake poller to start with a 5s timeout. If the poller never starts, the test now fails locally with a useful error instead of hanging until the whole package times out.

  - TestAsyncOperationFromWorkflow_CancellationTypes/Abandon: increased the context timeout for the Abandon case from 10s to 30s. CI hit the test context deadline before run.Get could observe the workflow cancellation result.

  - TestPinnedBehaviorThreeWorkers: The test now verifies routing config through the SDK client’s normalized Describe response instead of reading a single raw proto field. This matches the public API behavior and avoids an always-false wait when the server represents the same routing version through the fallback wire field.

  - TestLongRunningActivityWithHBAndGrpcRetries: reduced induced heartbeat RPC failures from 3 to 2 and raised the heartbeat timeout from 2s to 3s. This still verifies heartbeat retry metrics, but avoids Windows CI timing out the activity while the test is intentionally delaying heartbeat calls.

## Why?

More reliable CI

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Low Risk**
> The only runtime change is ordering in worker startup; remaining edits are test-only timing, assertions, and client API usage for waits.
> 
> **Overview**
> Fixes a **worker startup race** where autoscaling pollers could begin polling before every poller type was registered with the balancer. `baseWorker.Start` now **registers all poller types first**, then starts poller goroutines, so one type cannot consume slots before the balancer knows about the others.
> 
> The rest of the PR **hardens flaky tests** without changing product behavior: autoscaling minimum-concurrency assertions wait on stable state; legacy shutdown tests use bounded waits for poller startup; local-activity shutdown timing waits for the LA to start, uses a bounded context, and avoids long-polling workflow history; heartbeat/grpc-retry integration tests use fewer induced failures and a slightly longer heartbeat timeout; Nexus abandon cancellation gets a longer test timeout; worker deployment routing waits use the SDK `Describe` API and version-aware “running” checks; dev server version pin moves from crash-prone CLI **v0.3.0** to **v1.6.1**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d220f5e6175ad742c6a165f1b689e56db5746e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/internal_worker_base_test.go
+++ b/internal/internal_worker_base_test.go
@@ -817,7 +817,7 @@ func TestTaskNotProcessedDuringLegacyShutdown(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			taskProcessed := make(chan struct{}, 1)
-			pollStarted := make(chan struct{})
+			pollStarted := make(chan struct{}, 1)
 
 			// This poller simulates a poll returning a task after shutdown has
 			// already started. Legacy shutdown should not dispatch that task.
--- a/internal/internal_worker_base_test.go
+++ b/internal/internal_worker_base_test.go
@@ -842,7 +842,11 @@ func TestTaskNotProcessedDuringLegacyShutdown(t *testing.T) {
 			})
 
 			bw.Start()
-			<-pollStarted
+			select {
+			case <-pollStarted:
+			case <-time.After(5 * time.Second):
+				t.Fatal("poller did not start in time")
+			}
 
 			// AggregatedWorker.Stop sets noRepoll before stopping base workers.
 			bw.noRepoll.Store(true)
--- a/internal/internal_worker_base_test.go
+++ b/internal/internal_worker_base_test.go
@@ -306,10 +306,11 @@ func (s *ScalableTaskPollerSuite) TestAutoscalingScalesDownToMinimum() {
 
 	eventuallyAutoscalingPollerState(s.T(), poller.autoscalingRunner, 1, "expected concurrency to reduce to minimum")
 
-	require.Never(s.T(), func() bool {
-		blockingPoller.Allow(readAutoscalingPollerState(poller.autoscalingRunner))
-		return readAutoscalingPollerState(poller.autoscalingRunner) == 0
-	}, 200*time.Millisecond, 10*time.Millisecond, "should not scale below minimum")
+	for range 5 {
+		assert.Equal(s.T(), int64(1), poller.pollerAutoscaler.target.Load(), "should not scale target below minimum")
+		blockingPoller.Allow(1)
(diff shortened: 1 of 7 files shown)


internal/internal_worker_base_test.go
```diff
@@ -306,10 +306,11 @@ func (s *ScalableTaskPollerSuite) TestAutoscalingScalesDownToMinimum() {
 
 	eventuallyAutoscalingPollerState(s.T(), poller.autoscalingRunner, 1, "expected concurrency to reduce to minimum")
 
-	require.Never(s.T(), func() bool {
-		blockingPoller.Allow(readAutoscalingPollerState(poller.autoscalingRunner))
-		return readAutoscalingPollerState(poller.autoscalingRunner) == 0
-	}, 200*time.Millisecond, 10*time.Millisecond, "should not scale below minimum")
+	for range 5 {
```
Sushisource, line 309: Why does this become 5?

Sushisource (approved):
Nice to see flake paydown! Only one non-blocking question

Shivs11 (approved):
reviewed the worker deployment test only, which looks great to me

internal/internal_worker_base_test.go
yuandrew: Instead of being time-based, we're now checking a fixed number of poll cycles. 5 is kinda arbitrary, but gives us enough of a signal that several empty polls have been allowed through after scaling down, and each leaves the target clamped at a minimum

PR #2412, opened 2026-06-20
Clear worker slot metrics on stop

### LINKED ISSUE #873: Ensure slots available metric is updated on worker stop
## Expected Behavior

Slots available added on start are removed on stop

## Actual Behavior

Glancing at the code, these may not be removed properly

kritibehl:
## What was changed

This PR clears worker slot gauges when a worker stops.

Specifically, it adds metric cleanup to the tracking slot supplier and calls it from `baseWorker.Stop()` after worker shutdown/drain. The cleanup is idempotent and guarded so late slot releases after shutdown do not republish stale slot availability.

A regression test was added to verify that worker slot gauges are reset to zero on stop and remain zero after a late slot release.

## Why?

Stopped workers should not continue reporting available task capacity through `worker_task_slots_available`.

Without cleanup, slot metrics can remain at their last published value after worker shutdown, which may leave stale capacity signals in metrics reporting.

## Checklist

1. Closes #873

2. How was this tested:

* `go test ./internal -run 'TestScalableTaskPollerSuite/TestTrackingSlotSupplierStopsSlotMetrics' -count=1`
* `go test ./internal -run 'TestScalableTaskPollerSuite/TestTrackingSlotSupplier' -count=1`
* `go test ./internal -race -run 'TestScalableTaskPollerSuite/TestTrackingSlotSupplierStopsSlotMetrics' -count=1`
* `go test ./internal -count=1`

3. Any docs updates needed?

No docs updates needed.

### DIFF
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -218,12 +218,12 @@ type (
 
 	// baseWorker that wraps worker activities.
 	baseWorker struct {
-		options              baseWorkerOptions
-		isWorkerStarted      bool
+		options         baseWorkerOptions
+		isWorkerStarted bool
 		// stopCh is created by newBaseWorker and closed by baseWorker.Stop().
 		// It is internal to baseWorker and stops its poller, dispatcher, autoscaler,
 		// and throttling/backoff loops.
-		stopCh chan struct{}
+		stopCh               chan struct{}
 		stopWG               sync.WaitGroup // The WaitGroup for stopping existing routines.
 		pollLimiter          *rate.Limiter
 		taskLimiter          *rate.Limiter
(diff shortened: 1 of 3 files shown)


yuandrew:
Thanks for working on this. After looking more closely, clearing these gauges on shutdown is not just a local cleanup. It defines how stopped workers should appear in metrics, affects workers sharing the same metric labels, and may hide tasks still running after the stop timeout. Java and Core-based SDKs currently retain the last reported values, so we should align on the intended cross-SDK behavior before changing Go.

Is this related to an issue you're hitting? If it is, more information on that issue would help us prioritize. But given the broader design work involved and the relatively low priority of the original issue, I'm not sure we'd plan to pursue this right now otherwise.

kritibehl:
Thanks for the context, that makes sense. This isn’t related to a production issue I’m hitting; I picked it up based on #873 and interpreted it as requiring the worker’s slot metrics to stop reporting stale capacity after shutdown.

I agree that this needs a broader definition of the shutdown semantics before changing Go independently, especially with shared metric labels, work that may outlive the stop timeout, and the current Java/Core behavior.

I’m happy to investigate the corresponding behavior across the SDKs and summarize the options on #873 if that would be useful. Otherwise, given the priority, I’m also fine leaving this here until there’s agreement on the intended cross-SDK behavior.

PR #2416, opened 2026-06-23
Add workflow.SignalWithStartWorkflow and System Nexus generator script

dplyukhin:
# What was changed

- Add `workflow-service.wit`, an IDL file that defines the shape of a `workflow.SignalWithStartWorkflow` API.
- Add `internal/cmd/gen-system-nexus`, a command that uses checked-in `.wit` files and a `temporalio/nexgen` binary to implement workflow APIs.
- Add `SignalWithStartWorkflow` API inside the `workflow` package. It's implemented as a Nexus operation with the internal `__temporal_system` endpoint.

## Why?

So we can use `SignalWithStartWorkflowExecution` inside a workflow, not just in a client.

## Checklist

1. Closes: N/A

2. How was this tested:

   - `go build ./...`
   - `go test ./internal/...`
   - Added integration tests for request encoding and payload processing.
   - Added unit tests for the system endpoint and service exemption.

3. Any docs updates needed?

   No.


internal/nexussystem/Makefile
```diff
@@ -0,0 +1,33 @@
+# Regenerates the system Nexus client bindings (api.go, model_overrides.go) and
```
tconley1428: I think we just want it as part of the root make file.

internal/internal_utils.go
```diff
@@ -32,10 +31,21 @@ const (
...
 	temporalPrefix      = "__temporal_"
 	temporalPrefixError = "__temporal_ is a reserved prefix"
+
+	// systemNexusEndpoint and systemNexusService identify the built-in "system
+	// Nexus" endpoint/service pair. These intentionally use the otherwise
+	// reserved __temporal_ prefix because they are Temporal's own internal use
+	// of that prefix (e.g. workflow.SignalWithStartWorkflow). NewNexusClient
+	// permits this specific pair despite the reserved-prefix guard.
+	systemNexusEndpoint = "__temporal_system"
+	systemNexusService  = "temporal.api.workflowservice.v1.WorkflowService"
```
tconley1428: Should be gone with change to endpoint only.

internal/workflow.go
```diff
@@ -3059,11 +3031,15 @@ func NewNexusClient(endpoint, service string) NexusClient {
...
 		panic("service must not be empty")
 	}
-	if strings.HasPrefix(endpoint, temporalPrefix) {
-		panic("endpoint cannot use reserved __temporal_ prefix")
-	}
-	if strings.HasPrefix(service, temporalPrefix) {
-		panic("service cannot use reserved __temporal_ prefix")
+	// The built-in system Nexus endpoint/service pair is Temporal's own use of
+	// the otherwise reserved __temporal_ prefix and is therefore permitted.
+	if !isSystemNexusClient(endpoint, service) {
```
tconley1428: Changes with the endpoint only change

internal/cmd/nexussystemgen/main.go
```diff
@@ -0,0 +1,302 @@
...
+				Path: &ast.BasicLit{Kind: token.STRING, Value: strconv.Quote(internalImport)},
+			})
+			break
+		}
+	}
+}
+
+// workflowForwarder returns the generated public re-export that surfaces the
+// system Nexus operations on the workflow package.
+func workflowForwarder() string {
```
tconley1428: Bad boy claude

.github/workflows/ci.yml
```diff
@@ -11,6 +11,24 @@ permissions:
...
+  check-generated:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          submodules: recursive
+
+      - name: Setup Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
```
yuandrew: btw, i think this is now stale

internal/cmd/nexussystemgen/main.go
```diff
@@ -0,0 +1,213 @@
...
+
+// copyFile copies a file from src to dst, creating dst if necessary.
+func copyFile(dst, src string) error {
+	in, err := os.Open(src)
+	if err != nil {
+		return fmt.Errorf("opening %s: %w", src, err)
+	}
+	defer func() { _ = in.Close() }()
+
+	out, err := os.Create(dst)
```
yuandrew: nit: if copy fails halfway through, there's a partial generated file. I think it would be safer to create a temp file, then rename it to the final file name

internal/cmd/nexussystemgen/main.go
```diff
@@ -0,0 +1,213 @@
...
+		_ = out.Close()
+		return fmt.Errorf("copying %s to %s: %w", src, dst, err)
+	}
+	if err := out.Close(); err != nil {
+		return fmt.Errorf("closing %s: %w", dst, err)
+	}
+	return nil
+}
+
+func gofmt(path string) error {
```
yuandrew: Could we use go.format.Source instead of spawning gofmt? that should let us get rid of gofmt() and most of copyFile

internal/cmd/nexussystemgen/main.go
```diff
@@ -0,0 +1,213 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("running nex-gen: %w", err)
+	}
+
+	if err := gofmt(serviceTmp); err != nil {
+		return err
+	}
+	if err := copyFile(serviceDst, serviceTmp); err != nil {
```
yuandrew: Small optimization: we could compare the formatted output with the existing workflowservice.go and only write if the files differ

internal/cmd/nexussystemgen/main.go
```diff
@@ -0,0 +1,213 @@
...
+	repoRoot, err := repoRoot()
+	if err != nil {
+		return err
+	}
+
+	witMain := filepath.Join(repoRoot, "internal", "nexussystem", "wit", "workflow-service.wit")
+	witDeps := filepath.Join(filepath.Dir(witMain), "deps")
+	outPkgDir := filepath.Join(repoRoot, "workflow")
+
+	tmpDir, err := os.MkdirTemp("", "nexussystemgen-")
```
yuandrew: One small debugging tip that Codex left, we could add a -keep-temp flag or log/preserve the dir on failure, so malformed generator output can be inspected. Up to you on if you think that's helpful and worth implementing

workflow/workflow.go
```diff
@@ -857,6 +857,12 @@ func IsContinueAsNewError(err error) bool {
 	return errors.As(err, &continueAsNewErr)
 }
 
+// GetDataConverter returns the data converter associated with the workflow
+// context, with workflow serialization context applied.
+func GetDataConverter(ctx Context) converter.DataConverter {
```
yuandrew: Is this still needed? Not seeing where it's used, guessing it was from a previous iteration?

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
+// Code generated by nex-gen. DO NOT EDIT.
```
yuandrew: nit: can we have a blank line after this comment? Usually comments right above package declarations are package documentation

internal/workflow.go
```diff
@@ -3131,7 +3131,15 @@ func (wc *workflowEnvironmentInterceptor) prepareNexusOperationParams(ctx Contex
 		return ExecuteNexusOperationParams{}, fmt.Errorf("invalid 'operation' parameter, must be an OperationReference or a string")
 	}
 
-	payload, err := dc.ToPayload(input.Input)
+	var payload *commonpb.Payload
+	var err error
+	if input.Client.Endpoint() == systemNexusEndpoint {
+		// System Nexus envelopes are encoded as proto-binary because they're
+		// consumed by Temporal, not by user code.
+		payload, err = converter.NewProtoPayloadConverter().ToPayload(input.Input)
```
yuandrew, line 3146: `ToPayload` could return `(nil, nil)` for unsupported values. Do we want to add a nil payload check here and return an actionable error about system Nexus input needing to be a protobuf message?

internal/workflow.go
```diff
@@ -3083,7 +3083,7 @@ func NewNexusClient(endpoint, service string) NexusClient {
 	if service == "" {
 		panic("service must not be empty")
 	}
-	if strings.HasPrefix(endpoint, temporalPrefix) {
+	if strings.HasPrefix(endpoint, temporalPrefix) && endpoint != systemNexusEndpoint  {
```
yuandrew: Do we want to expose system nexus construction through the public `workflow.NewNexusClient` API? 

We could have a separate `NewSystemNexusClient` in `internal/workflow.go`, which workflowservice.go could access via `workflow.newSystemNexusClient`

internal/workflow.go
yuandrew, line 3146: Codex traced this, 

Here we explicitly bypass a user dc for the outer system request, but server returns a binary-protobuf `SignalWithStartWorkflowExecutionResponse`, which calls `fut.Get()` eventually does 

```
dataConverter := getDataConverterFromWorkflowContext(ctx)
dataConverter.FromPayloads(payloads, valuePtr)
```

which would fail with a user dc that doesn't support protobuf

workflow/nexus_system_support.go
```diff
@@ -0,0 +1,198 @@
...
+				Pinned: &workflowpb.VersioningOverride_PinnedOverride{
+					Behavior: workflowpb.VersioningOverride_PINNED_OVERRIDE_BEHAVIOR_PINNED,
+					Version: &deploymentpb.WorkerDeploymentVersion{
+						DeploymentName: v.Version.DeploymentName,
+						BuildId:        v.Version.BuildID,
+					},
+				},
+			},
+		}, nil
+	case *client.AutoUpgradeVersioningOverride:
```
yuandrew: `OneTimeVersioningOverride` was added recently, I think we should support here

workflow/nexus_system_support.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		fields[key] = payload
+	}
+	return &common.Memo{Fields: fields}, nil
+}
+
+// --- SearchAttributes (temporal.api.common.v1.SearchAttributes) ---
+
+func searchAttributesToProto(_ Context, searchAttributes *temporal.SearchAttributes) (*common.SearchAttributes, error) {
+	if searchAttributes == nil {
```
yuandrew, line 92: The generated caller always passes &m.SearchAttributes, so the nil check here never omits the zero value. This produces a present SearchAttributes message with an empty map, whereas the existing Signal-With-Start serialization omits the field when Size() == 0. Could this check searchAttributes == nil || searchAttributes.Size() == 0 so the WIT option<search-attributes> semantics are preserved?

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+		IdReusePolicy:      idReusePolicy,
+		IdConflictPolicy:   idConflictPolicy,
+		RetryPolicy:        opts.RetryPolicy,
+		CronSchedule:       cronSchedule,
+		Memo:               opts.Memo,
+		SearchAttributes:   opts.SearchAttributes,
+		Priority:           opts.Priority,
+		VersioningOverride: opts.VersioningOverride,
+		StartDelay:         startDelay,
+		UserMetadata:       &opts.UserMetadata,
```
yuandrew, line 414: `buildUserMetadata` omits the message if both fields are empty and omits each empty field individually, can we mirror that here? Right now we're always passing a non-nil pointer

workflow/nexus_system_support.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, err
+	}
+	return value, nil
+}
+
+func payloadsToProto(ctx Context, values []any) (*common.Payloads, error) {
+	return getWorkflowDataConverter(ctx).ToPayloads(values...)
+}
+
+func getWorkflowDataConverter(ctx Context) converter.DataConverter {
```
yuandrew: Should we be using `internalGetDataConverterFromWorkflowContext` here instead of using reflaction and hard-coded `wfEnvOptions` here?

internal/internal_workflow.go
```diff
@@ -1620,6 +1620,14 @@ func getDataConverterFromWorkflowContext(ctx Context) converter.DataConverter {
 	return WithWorkflowContext(ctx, dataConverter)
 }
 
+// GetDataConverterFromWorkflowContext returns the data converter associated
+// with the workflow context, with workflow serialization context applied.
+//
+// Exposed as: [go.temporal.io/sdk/workflow.GetDataConverter]
+func GetDataConverterFromWorkflowContext(ctx Context) converter.DataConverter {
```
yuandrew: I think we can keep this and use it from workflow/nexus_system_support.go, but remove the `// Exposed as:` comment

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+	return result
+}
+
+// --- Operations (public API) ---
+
+type UserMetadata struct {
+	// Optional.
+	// Single-line fixed summary for the workflow execution that may appear in UI and CLI.
+	// This can be in single-line Temporal Markdown format.
+	StaticSummary string
```
yuandrew, line 195: Should we follow the existing godoc strings? 
```
// StaticSummary - Single-line fixed summary for this workflow execution that will appear in UI/CLI. This can be
		// in single-line Temporal markdown format.
		//
		// Optional: defaults to none/empty.
```

godoc convention starts with the field name as the first word https://go.dev/blog/godoc

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+	if opts.CronSchedule != "" {
+		cronSchedule = &opts.CronSchedule
+	}
+	var startDelay *time.Duration
+	if opts.StartDelay != 0 {
+		startDelay = &opts.StartDelay
+	}
+	workflowName := ""
+	{
+		switch rv := reflect.ValueOf(workflow); rv.Kind() {
```
yuandrew, line 382: Should we use the same resolution as `ExecuteChildWorkflow`/`getWorkflowFunctionName`? This currently ignores custom names useres set with `RegisterWorkflowWithOptions`

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+	{
+		converted := proto.GetStarted()
+		value.Started = &converted
+	}
+	return value, nil
+}
+
+type SignalWithStartWorkflowOptions struct {
+	// Required. Unique identifier for the workflow execution.
+	Id string
```
yuandrew, line 282: Everywhere else we use `ID`, same with other ones below, like `RequestID`

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+// Input workflow: Workflow function identifying the workflow to start.
+// Input args: Arguments for the workflow.
+//
+// Returns: A workflow handle to the started workflow.
+func SignalWithStartWorkflowTyped[WorkflowArg any, WorkflowResult any](
+	ctx Context,
+	opts SignalWithStartWorkflowOptions,
+	signal string,
+	signalArg any,
+	workflow func(Context, WorkflowArg) WorkflowResult,
```
yuandrew, line 431: I forget if we talked about this in our last call going over this, but how it is today, this only supports workflows where `WorkflowResult` is `err`, since workflow functions must return either `err` or `(result, err)`

workflow/workflowservice.go
yuandrew: This comment applies to all go files in this PR

workflow/workflowservice.go
Quinn-With-Two-Ns: Yeah should always be `ID` in Go

workflow/nexus_system_support.go
```diff
@@ -0,0 +1,198 @@
+package workflow
```
Quinn-With-Two-Ns, line 1: Are these methods used anywhere? Or am I missing something?

Quinn-With-Two-Ns: Ah NVM

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+	ExecutionTimeout time.Duration
+	// Optional.
+	// Timeout of a single workflow run.
+	RunTimeout time.Duration
+	// Optional.
+	// Timeout of a single workflow task.
+	TaskTimeout time.Duration
+	// Optional.
+	// Request ID used to deduplicate workflow start requests.
+	RequestId string
```
Quinn-With-Two-Ns, line 296: Did we mean to expose this to users? We don't expose this in the client API

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+		value.Started = &converted
+	}
+	return value, nil
+}
+
+type SignalWithStartWorkflowOptions struct {
+	// Required. Unique identifier for the workflow execution.
+	Id string
+	// Required. Task queue to run the workflow on.
+	TaskQueue string
```
Quinn-With-Two-Ns, line 284: I would still argue we should use the workflows task queue if the user doesn't set one, but I won't die on that hill for pre-release

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+	// Optional.
+	// Total workflow execution timeout, including retries and continue-as-new.
+	ExecutionTimeout time.Duration
```
Quinn-With-Two-Ns, line 287: Did we intentionally use different names then what these are called in `StartWorkflowOptions`?

workflow/workflowservice.go
```diff
@@ -0,0 +1,496 @@
...
+	Priority *temporal.Priority
+	// Optional.
+	// Override for workflow versioning behavior.
+	VersioningOverride client.VersioningOverride
+	// Optional.
+	// Amount of time to wait before starting the workflow. This does not work with
+	// cron-schedule.
+	StartDelay time.Duration
+	// Optional.
+	UserMetadata
```
Quinn-With-Two-Ns, line 329: Why is this option embedded?

internal/nexussystem/wit/workflow-service.wit
```diff
@@ -0,0 +1,130 @@
...
+    versioning-override: option<versioning-override>,
+    /// @nexus.doc "Amount of time to wait before starting the workflow. This does not work with cron-schedule."
+    /// @nexus.proto-field "workflow_start_delay"
+    start-delay: option<duration>,
+    user-metadata: option<user-metadata>,
+    /// @nexus.source python="workflow_namespace()" typescript="workflowNamespace()" go="GetInfo(ctx).Namespace" dotnet="TemporalWorkflowContext.WorkflowNamespace()"
+    namespace: string,
+    /// @nexus.omit
+    control: placeholder,
+    /// @nexus.omit
```
Quinn-With-Two-Ns, line 94: If we omit headers, how is tracing supposed to work? How would a user setup otel tracing on this API?

internal/nexussystem/wit/workflow-service.wit
```diff
@@ -0,0 +1,130 @@
...
+    time-skipping-config: placeholder,
+  }
+
+  /// @nexus.experimental
+  /// @nexus.proto "temporal.api.workflowservice.v1.SignalWithStartWorkflowExecutionResponse" typescript-import="@temporalio/proto"
+  record signal-with-start-workflow-response {
+    run-id: option<string>,
+    started: option<bool>,
+    /// @nexus.omit
+    first-execution-run-id: placeholder,
```
Quinn-With-Two-Ns, line 108: Why are we omitting this field? it is useful to help target future messages.

workflow/workflowservice.go
tconley1428: I see, it's private. In Python it is public but commented as experimental, which is kinda weird.

internal/nexussystem/wit/workflow-service.wit
tconley1428: This will be changing, but good call.

internal/nexussystem/wit/workflow-service.wit
tconley1428: It didn't exist when the effort started, so it was omitted. It's not currently used in the actual result type either, so it won't do anything to keep it currently.

PR #2417, opened 2026-06-23
NEXUS-474: Support UpdateWorkflow as a Nexus Operation

mavemuri:
## What was changed

Adds support for UpdateWorkflow as a Nexus operation

## Why?

Part of effort to expose all Temporal primitives via Nexus operations 

## Checklist

1. Closes NEXUS-474

2. How was this tested:

- [x] added integration tests - run as `go run . integration-test -dev-server -run UpdateWorkflowOperation`

3. Any docs updates needed?

- [ ] TBD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ to docs, or any other relevant information.
 # Changelog
 
 ## [Unreleased]
+- Add support for Workflow Updates as Nexus Operations 
 
 ### Added
 
--- a/internal/client.go
+++ b/internal/client.go
@@ -1841,32 +1841,19 @@ func (e *WorkflowUpdateServiceTimeoutOrCanceledError) Error() string {
 
 func (e *WorkflowUpdateServiceTimeoutOrCanceledError) Unwrap() error { return e.cause }
 
-// SetRequestIDOnStartWorkflowOptions is an internal only method for setting a requestID on StartWorkflowOptions.
-// RequestID is purposefully not exposed to users for the time being.
-func SetRequestIDOnStartWorkflowOptions(opts *StartWorkflowOptions, requestID string) {
-	opts.requestID = requestID
-}
-
-// SetCallbacksOnStartWorkflowOptions is an internal only method for setting callbacks on StartWorkflowOptions.
-// Callbacks are purposefully not exposed to users for the time being.
-func SetCallbacksOnStartWorkflowOptions(opts *StartWorkflowOptions, callbacks []*commonpb.Callback) {
-	opts.callbacks = callbacks
-}
-
-// SetLinksOnStartWorkflowOptions is an internal only method for setting links on StartWorkflowOptions.
-// Links are purposefully not exposed to users for the time being.
-func SetLinksOnStartWorkflowOptions(opts *StartWorkflowOptions, links []*commonpb.Link) {
-	opts.links = links
-}
-
 // interface utility wrapper to allow setting links and callbacks
-// on temporal primitive operation options (UpdateWorkflowOptions, etc)
-// draft-review: set it on StartWorkflowOptions above as well
+// on temporal primitive operation options (UpdateWorkflowOptions, StartWorkflowOptions)
 type nexusTemporalOperationOptions interface {
 	setRequestID(requestID string)
 	setLinks(links []*commonpb.Link)
 	setCallbacks(callbacks []*commonpb.Callback)
 }
+
+// nexusTemporalOperationOptions conforming interfaces
+var (
+	_ nexusTemporalOperationOptions = (*UpdateWorkflowOptions)(nil)
+	_ nexusTemporalOperationOptions = (*StartWorkflowOptions)(nil)
+)
 
 // Set links on any [nexusTemporalOperationOptions] interface via the setLinks API.
 //
--- a/internal/client.go
+++ b/internal/client.go
@@ -1862,17 +1862,38 @@ func SetLinksOnStartWorkflowOptions(opts *StartWorkflowOptions, links []*commonp
 // interface utility wrapper to allow setting links and callbacks
 // on temporal primitive operation options (UpdateWorkflowOptions, etc)
 // draft-review: set it on StartWorkflowOptions above as well
-type nexusOperationOptions interface {
+type nexusTemporalOperationOptions interface {
+	setRequestID(requestID string)
 	setLinks(links []*commonpb.Link)
 	setCallbacks(callbacks []*commonpb.Callback)
 }
 
-func SetLinksOnNexusOperation(opts nexusOperationOptions, links []*commonpb.Link) {
+// Set links on any [nexusTemporalOperationOptions] interface via the setLinks API.
+//
+// Intended to be used only internally as a consistent way of setting
+// links on all Nexus Operations
+func SetLinksOnNexusOperation(opts nexusTemporalOperationOptions, links []*commonpb.Link) {
 	opts.setLinks(links)
 }
 
-func SetCallbacksOnNexusOperation(opts nexusOperationOptions, callbacks []*commonpb.Callback) {
+// Set callbacks on any [nexusTemporalOperationOptions] interface via the setCallbacks API.
+//
+// Intended to be used only internally as a consistent way of setting
+// callbacks on all Nexus Operations
+func SetCallbacksOnNexusOperation(opts nexusTemporalOperationOptions, callbacks []*commonpb.Callback) {
 	opts.setCallbacks(callbacks)
+}
+
+// Set non-empty requestID on any [nexusTemporalOperationOptions] interface via the setRequestID API.
+// Used for deduping requests server-side
+//
+// Intended to be used only internally as a consistent way of setting
+// requestIDs on all Nexus Operations
+func SetRequestIDOnNexusOperation(opts nexusTemporalOperationOptions, requestID string) {
+	if requestID == "" {
+		return
+	}
+	opts.setRequestID(requestID)
 }
 
 // SetOnConflictOptionsOnStartWorkflowOptions is an internal only method for setting conflict
--- a/internal/client.go
+++ b/internal/client.go
@@ -1859,6 +1859,22 @@ func SetLinksOnStartWorkflowOptions(opts *StartWorkflowOptions, links []*commonp
 	opts.links = links
 }
 
+// interface utility wrapper to allow setting links and callbacks
+// on temporal primitive operation options (UpdateWorkflowOptions, etc)
+// draft-review: set it on StartWorkflowOptions above as well
+type nexusOperationOptions interface {
+	setLinks(links []*commonpb.Link)
+	setCallbacks(callbacks []*commonpb.Callback)
+}
+
+func SetLinksOnNexusOperation(opts nexusOperationOptions, links []*commonpb.Link) {
+	opts.setLinks(links)
+}
+
+func SetCallbacksOnNexusOperation(opts nexusOperationOptions, callbacks []*commonpb.Callback) {
+	opts.setCallbacks(callbacks)
+}
+
 // SetOnConflictOptionsOnStartWorkflowOptions is an internal only method for setting conflict
 // options on StartWorkflowOptions.
 // OnConflictOptions are purposefully not exposed to users for the time being.
@@ -1876,6 +1892,16 @@ func SetOnConflictOptionsOnStartWorkflowOptions(opts *StartWorkflowOptions) {
 func SetResponseInfoOnStartWorkflowOptions(opts *StartWorkflowOptions) *startWorkflowResponseInfo {
 	if opts.responseInfo == nil {
 		opts.responseInfo = &startWorkflowResponseInfo{}
+	}
+	return opts.responseInfo
+}
+
+// SetResponseInfoOnUpdateWorkflowOptions is an internal only method to set and return a
+// responseInfo pointer. This is done to capture links from the response RPC to be used
+// for nexus forward links on UpdateWorkflow Nexus Operations
+func SetResponseInfoOnUpdateWorkflowOptions(opts *UpdateWorkflowOptions) *updateWorkflowResponseInfo {
+	if opts.responseInfo == nil {
+		opts.responseInfo = &updateWorkflowResponseInfo{}
 	}
 	return opts.responseInfo
 }
--- a/internal/cmd/build/main.go
+++ b/internal/cmd/build/main.go
@@ -165,9 +165,10 @@ func (b *builder) integrationTest() error {
 				"--dynamic-config-value", `component.nexusoperations.useSystemCallbackURL=false`,
 				"--dynamic-config-value", `component.nexusoperations.callback.endpoint.template="http://localhost:7243/namespaces/{{.NamespaceName}}/nexus/callback"`,
 				"--dynamic-config-value", "nexusoperation.enableStandalone=true",
-				"--dynamic-config-value", "history.enableChasmCallbacks=true",
+				"--dynamic-config-value", "history.enableCHASMCallbacks=true",
 				"--dynamic-config-value", "frontend.ListWorkersEnabled=true",
 				"--dynamic-config-value", "activity.startDelayEnabled=true",
+				"--dynamic-config-value", "history.enableUpdateCallbacks=true",
 			},
 		})
 		if err != nil {
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -511,9 +511,11 @@ type ClientUpdateWorkflowInput struct {
 	// WaitForStage is the stage to wait for.
 	WaitForStage        WorkflowUpdateStage
 
-	// links. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+	// request ID for server de-duplication. Only settable by the SDK - e.g. [temporalnexus.updateWorkflowOperation].
+	requestID string
+	// links. Only settable by the SDK - e.g. [temporalnexus.updateWorkflowOperation].
 	links     []*commonpb.Link
-	// for backward links from the target namespace sent via operation options. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+	// callbacks. Only settable by the SDK - e.g. [temporalnexus.updateWorkflowOperation].
 	callbacks []*commonpb.Callback
 	// gRPC request response trap for nexus forward links
 	responseInfo *updateWorkflowResponseInfo
--- a/internal/interceptor.go
+++ b/internal/interceptor.go
@@ -510,6 +510,13 @@ type ClientUpdateWorkflowInput struct {
 	FirstExecutionRunID string
 	// WaitForStage is the stage to wait for.
 	WaitForStage        WorkflowUpdateStage
+
+	// links. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+	links     []*commonpb.Link
+	// for backward links from the target namespace sent via operation options. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+	callbacks []*commonpb.Callback
+	// gRPC request response trap for nexus forward links
+	responseInfo *updateWorkflowResponseInfo
 }
 
 // Exposed as: [go.temporal.io/sdk/interceptor.ClientUpdateWithStartWorkflowInput]
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -30,6 +30,12 @@ const (
 	WorkflowUpdateStageCompleted
 )
 
+// workflow is async iff update stage is WorkflowUpdateStageAccepted-
+// as WorkflowUpdateStageCompleted blocks on UpdateWorkflow itself
+func (w WorkflowUpdateStage) IsAsyncUpdateWorkflow() bool {
+	return w == WorkflowUpdateStageAccepted
+}
+
 const (
 	updateStateNew              updateState = "New"
 	updateStateRequestInitiated updateState = "RequestScheduled"
--- a/internal/internal_update.go
+++ b/internal/internal_update.go
@@ -30,12 +30,6 @@ const (
 	WorkflowUpdateStageCompleted
 )
 
-// workflow is async iff update stage is WorkflowUpdateStageAccepted-
-// as WorkflowUpdateStageCompleted blocks on UpdateWorkflow itself
-func (w WorkflowUpdateStage) IsAsyncUpdateWorkflow() bool {
-	return w == WorkflowUpdateStageAccepted
-}
-
 const (
 	updateStateNew              updateState = "New"
 	updateStateRequestInitiated updateState = "RequestScheduled"
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -995,6 +995,26 @@ type UpdateWorkflowOptions struct {
 	// then the server will reject the update request with an error.
 	// Note that it is incompatible with UpdateWithStartWorkflowOperation.
 	FirstExecutionRunID string
+
+	// workflow completion callback. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+	callbacks []*commonpb.Callback
+	// for backward links from the target namespace sent via operation options. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+	links []*commonpb.Link
+	// gRPC request response trap for nexus forward links
+	responseInfo *updateWorkflowResponseInfo
+}
+
+type updateWorkflowResponseInfo struct {
+	// Link to the workflow event.
+	Link *commonpb.Link
+}
+
+func (u *UpdateWorkflowOptions) setLinks(links []*commonpb.Link) {
+	u.links = links
+}
+
+func (u *UpdateWorkflowOptions) setCallbacks(callbacks []*commonpb.Callback) {
+	u.callbacks = callbacks
 }
 
 // UpdateWithStartWorkflowOptions encapsulates the parameters used by UpdateWithStartWorkflow.
@@ -2846,6 +2866,10 @@ func (w *workflowClientInterceptor) UpdateWorkflow(
 		return nil, err
 	}
 
+	if responseInfo := in.responseInfo; responseInfo != nil {
+		responseInfo.Link = resp.GetLink()
+	}
+
 	// Here we know the update is at least accepted
 	desiredLifecycleStage := updateLifeCycleStageToProto(in.WaitForStage)
 	return w.updateHandleFromResponse(ctx, desiredLifecycleStage, resp)
@@ -2873,6 +2897,12 @@ func createUpdateWorkflowInput(options *UpdateWorkflowOptions) (*ClientUpdateWor
 		return nil, errors.New("WaitForStage WorkflowUpdateStageAdmitted is not supported")
 	}
 
+	// draft-review: check if this could be valid for some edge cases 
+	if options.WaitForStage == WorkflowUpdateStageCompleted && len(options.callbacks) > 0 {
+		return nil, errors.New("WaitForStage WorkflowUpdateStageCompleted does not support callbacks " +
+			"as it is already a synchronous operation")
+	}
+
 	return &ClientUpdateWorkflowInput{
 		UpdateID:            updateID,
 		WorkflowID:          options.WorkflowID,
@@ -2881,6 +2911,9 @@ func createUpdateWorkflowInput(options *UpdateWorkflowOptions) (*ClientUpdateWor
 		RunID:               options.RunID,
 		FirstExecutionRunID: options.FirstExecutionRunID,
 		WaitForStage:        options.WaitForStage,
+		links:               options.links,
+		callbacks:           options.callbacks,
+		responseInfo:        options.responseInfo,
 	}, nil
 }
 
@@ -2915,6 +2948,9 @@ func (w *workflowClientInterceptor) createUpdateWorkflowRequest(
 		},
 		FirstExecutionRunId: in.FirstExecutionRunID,
 		Request: &updatepb.Request{
+			RequestId:           in.UpdateID,
+			CompletionCallbacks: in.callbacks,
+			Links:               in.links,
 			Meta: &updatepb.Meta{
 				UpdateId: in.UpdateID,
 				Identity: w.client.identity,
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -1023,6 +1023,18 @@ func (u *UpdateWorkflowOptions) setRequestID(requestID string) {
 	u.requestID = requestID
 }
 
+func (s *StartWorkflowOptions) setLinks(links []*commonpb.Link) {
+	s.links = links
+}
+
+func (s *StartWorkflowOptions) setCallbacks(callbacks []*commonpb.Callback) {
+	s.callbacks = callbacks
+}
+
+func (s *StartWorkflowOptions) setRequestID(requestID string) {
+	s.requestID = requestID
+}
+
 // UpdateWithStartWorkflowOptions encapsulates the parameters used by UpdateWithStartWorkflow.
 // See UpdateWithStartWorkflow and NewWithStartWorkflowOperation.
 type UpdateWithStartWorkflowOptions struct {
@@ -1086,7 +1098,6 @@ type lazyUpdateHandle struct {
 // completed. Used for Nexus operations that back into operations at a later stage
 // in a non-retriable manner. Eg. UpdateWorkflow could fail at Accepted(failed validation)
 // but its still Admitted and isnt captured in the rpc errors and keeps getting retried
-// draft-review: is there a better way to do it instead?
 func IsUpdateWorkflowCompleted(handle WorkflowUpdateHandle) bool {
 	_, ok := handle.(*completedUpdateHandle)
 	return ok
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -996,9 +996,11 @@ type UpdateWorkflowOptions struct {
 	// Note that it is incompatible with UpdateWithStartWorkflowOperation.
 	FirstExecutionRunID string
 
-	// workflow completion callback. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+	// request ID for de-duplication during server processing. Only settable by the SDK - e.g. [temporalnexus.updateWorkflowOperation].
+	requestID string
+	// callbacks. Only settable by the SDK - e.g. [temporalnexus.updateWorkflowOperation].
 	callbacks []*commonpb.Callback
-	// for backward links from the target namespace sent via operation options. Only settable by the SDK - e.g. [temporalnexus.workflowRunOperation].
+	// for backward links from the target namespace sent via operation options. Only settable by the SDK - e.g. [temporalnexus.updateWorkflowOperation].
 	links []*commonpb.Link
 	// gRPC request response trap for nexus forward links
 	responseInfo *updateWorkflowResponseInfo
@@ -1015,6 +1017,10 @@ func (u *UpdateWorkflowOptions) setLinks(links []*commonpb.Link) {
 
 func (u *UpdateWorkflowOptions) setCallbacks(callbacks []*commonpb.Callback) {
 	u.callbacks = callbacks
+}
+
+func (u *UpdateWorkflowOptions) setRequestID(requestID string) {
+	u.requestID = requestID
 }
 
 // UpdateWithStartWorkflowOptions encapsulates the parameters used by UpdateWithStartWorkflow.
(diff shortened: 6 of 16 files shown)


contrib/aws/lambdaworker/go.mod
```diff
@@ -1,6 +1,6 @@
 module go.temporal.io/sdk/contrib/aws/lambdaworker
 
-go 1.24.0
```
Quinn-With-Two-Ns: I'd avoid bumping the required Go Version in this PR since this declares the min. Go Version our users need.

internal/client.go
```diff
@@ -1855,6 +1855,22 @@ func SetLinksOnStartWorkflowOptions(opts *StartWorkflowOptions, links []*commonp
...
 
+// interface utility wrapper to allow setting links and callbacks
+// on temporal primitive operation options (UpdateWorkflowOptions, etc)
+// draft-review: set it on StartWorkflowOptions above as well
+type nexusOperationOptions interface {
+	setLinks(links []*commonpb.Link)
+	setCallbacks(callbacks []*commonpb.Callback)
+}
+
+func SetLinksOnNexusOperation(opts nexusOperationOptions, links []*commonpb.Link) {
```
Quinn-With-Two-Ns: Nit: GoDocs

internal/internal_update.go
```diff
@@ -30,6 +30,12 @@ const (
 	WorkflowUpdateStageCompleted
 )
 
+// workflow is async iff update stage is WorkflowUpdateStageAccepted-
+// as WorkflowUpdateStageCompleted blocks on UpdateWorkflow itself
+func (w WorkflowUpdateStage) IsAsyncUpdateWorkflow() bool {
```
Quinn-With-Two-Ns: `WorkflowUpdateStage` is a type we expose to users, don't think we want to expose this helper to users so I would at least make it private or inline it. I would personally lean towards inlining it since it is so simple

temporalnexus/temporal_operation.go
```diff
@@ -299,6 +396,11 @@ func (o *temporalOperation[I, O]) Cancel(ctx context.Context, token string, opti
 		return o.options.CancelWorkflowRun(ctx, GetClient(ctx), CancelTemporalWorkflowRunOptions{
 			WorkflowID: wfToken.WorkflowID,
 		}, options)
+	case operationTokenTypeUpdateWorkflow:
+		return &nexus.HandlerError{
```
Quinn-With-Two-Ns: The default may be to fail, but we should still allow users to customize the cancellation behaviour, so here we should call `o.options.CancelWorkflowUpdate` and then the default value for that lambda would be to error

temporalnexus/temporal_operation.go
```diff
@@ -299,6 +396,11 @@ func (o *temporalOperation[I, O]) Cancel(ctx context.Context, token string, opti
...
+			Type:  nexus.HandlerErrorTypeBadRequest,
```
Quinn-With-Two-Ns: I think `nexus.HandlerErrorTypeNotImplemented` fits this situation better, what do you think?

temporalnexus/token.go
```diff
@@ -10,31 +10,63 @@ import (
 type operationTokenType int
 
 const (
-	operationTokenTypeWorkflowRun = operationTokenType(1)
+	operationTokenTypeReserved operationTokenType = iota
+	operationTokenTypeWorkflowRun
+	operationTokenTypeUpdateWorkflow
```
Quinn-With-Two-Ns, line 16: `operationTokenTypeUpdateWorkflow` should be `3` per https://app.notion.com/p/temporalio/Nexus-Operation-Token-Format-3738fc56773880d483dcd02931d6f90a?source=copy_link

temporalnexus/token.go
```diff
@@ -10,31 +10,63 @@ import (
...
+// workflowRunOperationToken is the decoded form of the workflow run operation token.
+type workflowRunOperationToken struct {
+	operationToken
+	NamespaceName string `json:"ns"`
+	WorkflowID    string `json:"wid"`
+}
+
+// updateWorkflow contains only meta - because it cannot be cancelled
+type updateWorkflowOperationToken struct {
+	operationToken
```
Quinn-With-Two-Ns, line 54: Missing Workflow ID and Update ID , the expected token format is defined here https://app.notion.com/p/temporalio/Nexus-Operation-Token-Format-3738fc56773880d483dcd02931d6f90a?source=copy_link

test/nexus_update_op_test.go
```diff
@@ -0,0 +1,153 @@
...
+		return 0, err
+	}
+
+	workflow.GetSignalChannel(ctx, doneSignal).Receive(ctx, nil)
+	workflow.GetLogger(ctx).Info("finished workflow, exiting now...", "final counter", counter)
+	return counter, nil
+}
+
+func getAddOperation(isAsync bool) (nexus.Operation[updateAddInput, updateAddOutput], error) {
+	updateStage := client.WorkflowUpdateStageCompleted
```
Quinn-With-Two-Ns: I think for now we should only support the `client.WorkflowUpdateStageAccepted` , `client.WorkflowUpdateStageCompleted` is a big footgun for users since it required the entire update to complete in under 10s

test/nexus_update_op_test.go
```diff
@@ -0,0 +1,153 @@
...
+func TestAsyncUpdateWorkflowOperation(t *testing.T) {
+	testUpdateWorkflowOperation(t, true)
+}
+
+func TestSyncUpdateWorkflowOperation(t *testing.T) {
+	testUpdateWorkflowOperation(t, false)
+}
+
+// run both via "go run . integration-test -dev-server -run UpdateWorkflowOperation"
+func testUpdateWorkflowOperation(t *testing.T, isAsync bool) {
```
Quinn-With-Two-Ns: Please also assert the forward and backlinks are present

Quinn-With-Two-Ns: I would test a few other scenarios to make sure the SDK handles these scenarios logically:
* Test what happens if the user tried an invalid update handler (ie not registered)
* Test what happens if the update handler has a validator that rejects the update
* Test what happens if the update handler returns an error
* Test what happens if the update handler is sync (ie it returns immediately and doesn't wait)
* Test what happens if Nexus operation targets the same, runing, workflow update (same workflow ID same update ID)
* Test what happens if a Nexus operation targets a already completed workflow update
* Test what happens if the user targets a invalid workflow (ie workflow already completed)

Let me know if you want to talk through any of these scenarios with me or what the expected behaviour should be

contrib/aws/lambdaworker/go.mod
mavemuri: version bump seems to be from here - https://github.com/temporalio/api-go/commit/d1423d20079f1c319da740179ca1d2b6e2375e44 (in-turn [here](https://github.com/nexus-rpc/nexus-proto-annotations/blob/main/go.mod#L3))- went in v1.62.13+

We could maybe fix it all the way back but not sure as there might be others already using it- is there another way? Or should we flag this for further discussion?

temporalnexus/temporal_operation.go
mavemuri: Could it be confusing for users to trigger a cancel UpdateWorkflow on async? I was thinking that because there isnt a way currently to cancel an UpdateWorkflow, this also remains consistent with that - please correct/add if needed 

maybe for another discussion- but could there be some compensating ops/events for such events?

temporalnexus/token.go
mavemuri: thanks! missed this doc

test/nexus_update_op_test.go
mavemuri: Was thinking it should be 1:1 with primitives because there could be edge use-cases - like a best effort update that still uses the validation logic and has the provenance in caller etc so they dont use signal nexus op 

Eg pickup-driver request matching service that needs to be done in 10s or it backoff/retries with a different payload - validation may not be enough as it checks some other subsystem(nexus op chaining) later on etc

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: > I was thinking that because there isnt a way currently to cancel an UpdateWorkflow

There isn't a built in way to cancel an `UpdateWorkflow` yes, but a user could build their own way to cancel a WorkflowUpdate

contrib/aws/lambdaworker/go.mod
Quinn-With-Two-Ns: AH okay it came from the API bump I understand, bumping to `1.25.x` is fine then since it is within the min Go Version we support

test/nexus_update_op_test.go
Quinn-With-Two-Ns: So I don't think we want to be 1:1 with the primitive in the Nexus Handler because the Nexus handler has certain limits that normal callers do not have. A nexus handler only has 10s to run so if a user combines `client.WorkflowUpdateStageCompleted` with a long running workflow update then they could get themselves into a bad state where their handler just keeps timing out over and over. From a performance perspective I am not sure there is any value of supporting `client.WorkflowUpdateStageCompleted` for Nexus since the operation will always wait for the operation to complete

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +161,151 @@ func StartUntypedWorkflow[R any](
...
+
+	handle, err := GetClient(ctx).UpdateWorkflow(ctx, updateWorkflowOptions)
+	if err != nil {
+		return TemporalOperationResult[R]{}, err
+	}
+
+	if internal.IsUpdateWorkflowCompleted(handle) {
+		// if workflow handle is completed and it has an error => its an unretriable error
+		// like validation failing on the update handler
+		if err := handle.Get(ctx, nil); err != nil {
```
Quinn-With-Two-Ns: later we do
```
	if err := handle.Get(ctx, &result); err != nil {
		return TemporalOperationResult[R]{}, err
	}
```
I would unify these two code paths, if the `handle` is complete we can just return the result or error

Quinn-With-Two-Ns: We may need to move the link code though

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +161,151 @@ func StartUntypedWorkflow[R any](
...
+	}
+	return NewSyncResult(result), nil
+}
+
+// validations to be performed specifically for UpdateWorkflow as a Nexus Operation.
+// Acts as a no-retriable fast fail for cases where invalid configurations were
+// submitted that would have (100%) failed later on required because there isnt a
+// way for handler to otherwise tell that something has failed in a non-recoverable way.
+// draft-review: should there be some kind of error from handler that can inform better
+func validateUpdateWorkflowNexusOperation(u client.UpdateWorkflowOptions) error {
```
Quinn-With-Two-Ns, line 314: The SDK should already do this validation when you call `client.UpdateWorkflow`

test/nexus_update_op_test.go
```diff
@@ -0,0 +1,350 @@
...
+	}, counterWorkflow)
+	require.NoError(t, err)
+
+	stopCounterWorkflow := func() {
+		require.NoError(t, tc.client.SignalWorkflow(ctx, counterWorkflowID, "", doneSignal, nil))
+		require.NoError(t, counterWorkflowRun.Get(ctx, nil))
+	}
+
+	// fails validation, shouldnt be retried
+	invalidUpdateWorkflowRun, err := tc.client.ExecuteWorkflow(ctx, client.StartWorkflowOptions{
```
Quinn-With-Two-Ns: Can you break these up into different sub tests

test/nexus_update_op_test.go
```diff
@@ -0,0 +1,350 @@
+package test_test
```
Quinn-With-Two-Ns, line 1: The convention would be to make these part of `IntegrationTestSuite`

test/nexus_update_op_test.go
```diff
@@ -0,0 +1,350 @@
...
+	require.NoError(t, err)
+	require.Error(t, counterUpdateWorkflowRun.Get(ctx, nil))
+}
+
+func counterWorkflow(ctx workflow.Context) (int, error) {
+	counter := 0
+
+	updateHandler := func(ctx workflow.Context, amount int) (updateAddOutput, error) {
+		counter += amount
+		_ = workflow.Sleep(ctx, time.Second)
```
Quinn-With-Two-Ns: A couple this we are missing is if this sleep is removed the update handler should return immediately with the result , we should also test if an error is returned here (both immediately and after sleeping)

test/nexus_update_op_test.go
```diff
@@ -0,0 +1,350 @@
...
+		WorkflowId: counterUpdateWorkflowRun.GetID(),
+		RunId:      counterUpdateWorkflowRun.GetRunID(),
+		Reference: &common.Link_WorkflowEvent_EventRef{EventRef: &common.Link_WorkflowEvent_EventReference{
+			EventId:   getEventIDByType(ctx, tc.client, counterUpdateWorkflowRun, enums.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED),
+			EventType: enums.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED,
+		}},
+	}}}
+	counterWorkflowLinks := getEventLinks(ctx, tc.client, counterWorkflowRun, eventsFilter)
+	callerWorkflowLinks := getEventLinks(ctx, tc.client, counterUpdateWorkflowRun, eventsFilter)
+	// draft-review: check why the links arent promoted on a sync request - seems server related
```
Quinn-With-Two-Ns: I think you need to update the link handling code `temporalnexus/link_converter.go` to handle the type of links you get from sync updates. Specifically `Workflow` links https://github.com/temporalio/api/commit/6487d668338e799e958e68ac56166dc42c67d3fb#diff-de60adc512f72d87dc1e38c0f30b41a16dede9243dfaf6209b4af9b5447fa551R260

test/nexus_update_op_test.go
Quinn-With-Two-Ns: One other tests I realized we are missing is, if their is no worker listing on the task queue for the workflow update then we should make sure the Nexus Operation doesn't fail

mavemuri: >update handler should return immediately with the result

It will still be an async response right? Because we are now limiting the sdk to only support async, server will always return on Accepted so itll always have to be completed via callbacks even if it can be completed immediately

temporalnexus/temporal_operation.go
mavemuri: This is for the case we were talking about where if the caller op params are invalid, handler shouldnt keep retrying. Without this, bad config push/etc on the operation would cause it to keep retrying - only guarding those cases

temporalnexus/temporal_operation.go
mavemuri: The link is empty if the op failed too - for UpdateWorkflow if it fails validation - if we move it, then the error becomes untraceable so kept it in normal flow
multiple Get on handler itself is safe so kept it this way for ease of reading

test/nexus_update_op_test.go
Quinn-With-Two-Ns: I believe the server will return the result immediately to the SDK if the update completes immediately OR if the update is already completed (if the user is reusing an Update ID)

Quinn-With-Two-Ns: We should test both of these IMO

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +171,158 @@ func StartUntypedWorkflow[R any](
...
+	}
+	internal.SetLinksOnNexusOperation(&updateWorkflowOptions, links)
+	internal.SetRequestIDOnNexusOperation(&updateWorkflowOptions, nc.startOperationOptions.RequestID)
+	header := nc.startOperationOptions.CallbackHeader
+	if header == nil {
+		header = make(nexus.Header)
+	}
+	// This field is expected to be populated by servers older than 1.27.0.
+	// draft-review: can this check be removed now? at least for UpdateWorkflow
+	header.Set(nexusHeaderOperationId, encodedToken)
```
Quinn-With-Two-Ns: I think we can drop this, it was needed for old servers, but they don't support Workflow Update Callbacks anyway

test/nexus_update_op_test.go
mavemuri: > if the update is already completed (if the user is reusing an Update ID)

this one is covered here 

> if the update completes immediately

but the async handle just returns immediately after update is accepted right? 
IIUC only thing we can do is maybe a quick check on Get (small context cancel/etc) immediately after handle returns - but not sure if its worth checking as its an async handle anyway and user can just do what they want

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: > The link is empty if the op failed too - for UpdateWorkflow if it fails validation

Oh really? That might be a bug then I think it should have a link. We should check with @long-nt-tran

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: I think per https://temporaltechnologies.slack.com/archives/C067YB0ET6Y/p1782406501894759?thread_ts=1782404316.749189&cid=C067YB0ET6Y @bergundy preference was to fail retryabilty for now. I am okay with that personally and then we should revist before GA

temporalnexus/temporal_operation.go
long-nt-tran: I think the link is attached in validation rejection here: https://github.com/temporalio/temporal/blob/main/service/history/api/updateworkflow/api.go#L283-L294

mavemuri: Thanks for the pointers - I incorrectly assumed it'd be empty because theres no history event in such cases 

This may need a quick discussion - because this is non-retriable failure/error, links wont get attached to it 

We may have to add a Links field to [StartOperationResponse_Failure](https://github.com/temporalio/api-go/blob/c027d896bbe85fa88a6ba230a886d4c321779916/nexus/v1/message.pb.go#L690) so it can get added in the [start op hande](https://github.com/temporalio/sdk-go/blob/3b7af44114d63653a8abf820f925bc5fe6d44b6b/internal/internal_nexus_task_handler.go#L287)

Based on the server code above- and what users may expect- if update fails validation, need to link to the workflow itself. Its not clear what user could do by following link because there wont be any history - but at least it should be preserved

temporalnexus/temporal_operation.go
```diff
@@ -43,6 +45,16 @@ type CancelTemporalWorkflowRunOptions struct {
...
 }
 
+// CancelTemporalUpdateWorkflowOptions are options provided to CancelWorkflowUpdate callback
+// of a Temporal Nexus Operation
+//
+// NOTE: Experimental
+type CancelTemporalUpdateWorkflowOptions struct {
+	// WorkflowID extracted from the operation token.
+	WorkflowID string
+	UpdateID   string
```
Quinn-With-Two-Ns: Nit: Go Doc

temporalnexus/temporal_operation.go
```diff
@@ -43,6 +45,16 @@ type CancelTemporalWorkflowRunOptions struct {
 	WorkflowID string
 }
 
+// CancelTemporalUpdateWorkflowOptions are options provided to CancelWorkflowUpdate callback
+// of a Temporal Nexus Operation
+//
+// NOTE: Experimental
+type CancelTemporalUpdateWorkflowOptions struct {
+	// WorkflowID extracted from the operation token.
+	WorkflowID string
```
Quinn-With-Two-Ns, line 61: We can have run ID here I think

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +167,149 @@ func StartUntypedWorkflow[R any](
...
+			// Failure case: If workflow handle is completed and it has an error => its
+			// an unretriable error like validation failing on the update handler.
+			return TemporalOperationResult[R]{}, &nexus.OperationError{
+				State:   nexus.OperationStateFailed,
+				Message: err.Error(),
+				Cause:   err,
+			}
+		}
+		// Success case: Required for correctness on retried UpdateWorkflow rpc with same updateID.
+		// Also handles case where multiple update workflows with same updateID are pending-
```
Quinn-With-Two-Ns: I don't think this comment is correct. There can only ever be one workflow update per update ID, when multiple handlers send requests with the same update ID the server just attaches multiple callbacks to the same workflow update

temporalnexus/temporal_operation.go
```diff
@@ -189,6 +344,10 @@ type TemporalOperationOptions[I, O any] struct {
 	// It receives the Temporal client, the workflow-run options extracted from the token, and the
 	// Nexus cancel options. If nil, defaults to cancelling the workflow via the Temporal client.
 	CancelWorkflowRun func(ctx context.Context, c client.Client, options CancelTemporalWorkflowRunOptions, cancelOptions nexus.CancelOperationOptions) error
+	// CancelWorkflowUpdate handles cancel for UpdateWorkflow operation tokens.
+	// It receives the Temporal client, the update-workflow options extracted from the token, and the
+	// Nexus cancel options. If nil, will error out as there is no default behavior for cancelling an UpdateWorkflow.
```
Quinn-With-Two-Ns: nit: call out what error is returned.

test/nexus_update_op_test.go
Quinn-With-Two-Ns: > async handle just returns immediately after update is accepted right

Depends if the update handle complete immediately the result will be immediately returned , the server will return completed.

temporalnexus/temporal_operation.go
mavemuri: run ID is not in the op token [here](https://app.notion.com/p/temporalio/Nexus-Operation-Token-Format-3738fc56773880d483dcd02931d6f90a) - should we add it there for this? I agree with you- think it makes sense to add - user might want to cancel an update on a specific runID and if thats not the current runID, they'll expect/handle the error etc

temporalnexus/temporal_operation.go
mavemuri: agree - removed it

test/nexus_update_op_test.go
mavemuri: Confused on this - theeres a TC here that tests this exact case - update "can" complete imediately but still returns async - this one "Verify immediate returns are still async" 
Im probably missing something - will sync up offline

Quinn-With-Two-Ns (approved):

CHANGELOG.md
```diff
@@ -51,3 +51,4 @@ to docs, or any other relevant information.
 - Support propagating standalone Nexus operation links.
 - OpenTelemetry tracing support for standalone activities started from the client.
 - Doclink now links interfaces when they're re-exported from `private` to a public package.
+- Add support for Workflow Updates as Nexus Operations
```
dplyukhin: 1.46.0 is already out, so this should go under the [Unreleased] heading

internal/client.go
```diff
@@ -1859,6 +1859,43 @@ func SetLinksOnStartWorkflowOptions(opts *StartWorkflowOptions, links []*commonp
 	opts.links = links
 }
 
+// interface utility wrapper to allow setting links and callbacks
+// on temporal primitive operation options (UpdateWorkflowOptions, etc)
+// draft-review: set it on StartWorkflowOptions above as well
```
yuandrew: `// draft-review: set it on StartWorkflowOptions above as well`

Do we still want this line?

internal/interceptor.go
```diff
@@ -510,6 +510,15 @@ type ClientUpdateWorkflowInput struct {
 	FirstExecutionRunID string
 	// WaitForStage is the stage to wait for.
 	WaitForStage        WorkflowUpdateStage
+
+	// request ID for server de-duplication. Only settable by the SDK - e.g. [temporalnexus.updateWorkflowOperation].
+	requestID string
```
yuandrew, line 515: I think including these unexported fields here can be a problem. If an external interceptor tries to reconstruct the input here, it could unknowingly drop these private fields, which could cause silent failures. Why do we need this when we've added these to `UpdateWorkflowOptions`?

internal/internal_workflow_client.go
```diff
@@ -1056,6 +1082,16 @@ type lazyUpdateHandle struct {
 	client *WorkflowClient
 }
 
+// IsUpdateWorkflowCompleted is a utility to detect if an operation has immediately
+// completed. Used for Nexus operations that back into operations at a later stage
+// in a non-retriable manner. Eg. UpdateWorkflow could fail at Accepted(failed validation)
+// but its still Admitted and isnt captured in the rpc errors and keeps getting retried
+// draft-review: is there a better way to do it instead?
```
yuandrew: another comment that should be addressed or removed

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +177,161 @@ func StartUntypedWorkflow[R any](
...
+// NOTE: Experimental
+func StartUpdateWorkflow[R any](
+	ctx context.Context,
+	nc NexusClient,
+	updateWorkflowOptions client.UpdateWorkflowOptions,
+) (TemporalOperationResult[R], error) {
+	if updateWorkflowOptions.UpdateID == "" {
+		// If an Update ID is not provided, use the RequestID. This is done to protect
+		// against cases where an unset UpdateID request with the same RequestID is
+		// retried due to n/w failure and gets new IDs in createUpdateWorkflowInput
```
yuandrew: For my own learning, what is a n/w failure?

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +177,161 @@ func StartUntypedWorkflow[R any](
...
+// NOTE: UpdateID will never be empty now that RequestID for nexus op is generated if empty
+// TBD: confirm one more time if retry-forever-on-invalid-config is ok before merge
+// and/or set WaitForStage after receiving the options (not done now to avoid silently
+// overwriting user provided options)
+func validateUpdateWorkflowNexusOperation(u client.UpdateWorkflowOptions) error {
+	switch {
+	// case u.WorkflowID == "":
+	// 	return errors.New("workflow ID cannot be empty")
+	// case u.UpdateName == "":
+	// 	return errors.New("update name cannot be empty")
```
yuandrew: commented code? Uncomment or remove

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +177,161 @@ func StartUntypedWorkflow[R any](
...
+	return NewAsyncResult[R](token), nil
+}
+
+// Validations to be performed specifically for UpdateWorkflow as a Nexus Operation.
+// Currently only guards against a sync Update being passed to the server. Invalid
+// updates - missing workflowID/updateName/etc - are expected to be retried forever.
+// NOTE: UpdateID will never be empty now that RequestID for nexus op is generated if empty
+// TBD: confirm one more time if retry-forever-on-invalid-config is ok before merge
+// and/or set WaitForStage after receiving the options (not done now to avoid silently
+// overwriting user provided options)
```
yuandrew: `TBD:` comment should also be removed before merging

internal/client.go
```diff
@@ -1859,6 +1859,43 @@ func SetLinksOnStartWorkflowOptions(opts *StartWorkflowOptions, links []*commonp
...
+type nexusTemporalOperationOptions interface {
```
yuandrew, line 1846: I think it would be good to add a compile-time assertion for each type that we implement this interface for

`  var _ nexusTemporalOperationOptions = (*UpdateWorkflowOptions)(nil)` and once you add StartWorkflowOptions, `_ nexusTemporalOperationOptions = (*StartWorkflowOptions)(nil)`

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +177,161 @@ func StartUntypedWorkflow[R any](
...
+				Cause:   err,
+			}
+		}
+		// Success case: Required for correctness on retried UpdateWorkflow rpc with same updateID.
+		return NewSyncResult(result), nil
+	}
+	opFailed = false
+	// regenerate token so it has the actual run ID that update is running on
+	// previous generation is only to handle completion before handle is returned
+	token, err = generateUpdateOperationToken(nctx.Namespace,
```
yuandrew, line 298: Do we have a test that exercises the pre-start response token?

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +177,161 @@ func StartUntypedWorkflow[R any](
 	return NewAsyncResult[R](handle.token()), nil
 }
 
+// StartUpdateWorkflow starts a type-safe workflow update run for a Nexus operation
+// linking the execution chain to the Nexus operation (callbacks, links, request ID).
+// It returns either an async result with a workflow-run operation token for pending
```
yuandrew: Should this be "workflow-run", or "workflow-update"? and can we be more explicit about the two-token behaviourin the GoDoc here?

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +177,161 @@ func StartUntypedWorkflow[R any](
...
+		return TemporalOperationResult[R]{}, &nexus.HandlerError{
+			Type:  nexus.HandlerErrorTypeInternal,
+			Cause: errors.New("unexpected error retrieving links from UpdateWorkflow response"),
+		}
+	}
+
+	nexus.AddHandlerLinks(ctx, ConvertCommonLinkToNexusLink(responseInfo.Link))
+
+	if internal.IsUpdateWorkflowCompleted(handle) {
+		// If the update workflow handle is completed already, return a sync response.
```
yuandrew, line 281: Can we assume a completed handle means this is a retry of an already-completed update? If a new update completes before `UpdateWorkflow` returns, could this produce both a synchronous result and a callback? Do we need an explicit signal to distinguish these cases?

temporalnexus/temporal_operation.go
yuandrew: Why would we want to retry forever for invalid configs, and not give a fatal error?

temporalnexus/link_converter.go
```diff
@@ -34,3 +34,23 @@ func ConvertLinkNexusOperationToNexusLink(no *commonpb.Link_NexusOperation) nexu
...
+	return internal.ConvertWorkflowlinkToNexusLink(workflowLink)
+}
+
+// ConvertCommonLinkToNexusLink converts a Common Link to a Nexus Link. Will be used
+// to safely point to non-workflow event links for cases where Nexus operations fail.
+// Eg. UpdateWorkflow fails validation -> point to Workflow instead of WorkflowEvent
+// as there will not be a history event.
+//
+// NOTE: Experimental
+func ConvertCommonLinkToNexusLink(commonLink *commonpb.Link) nexus.Link {
```
yuandrew, line 52: Since this is exported, we should be more explicit on what happens if someone passes in a non-workflow Link (Nexus, activity). Today this would silently convert into a bogus workflow link, I think

test/nexus_update_op_test.go
```diff
@@ -0,0 +1,466 @@
...
+	serviceName  = "counter-update-service"
+	addOperation = "addOperation"
+	addUpdate    = "addUpdate"
+	doneSignal   = "done"
+
+	taskQueue = "nexusOpUpdateWorkflowTQ"
+)
+
+func (ts *IntegrationTestSuite) TestNexusUpdateWorkflowOperation() {
+	ctx, cancel := context.WithTimeout(context.TODO(), defaultNexusTestTimeout)
```
yuandrew: This 10-second context is shared across all of the sequential subtests, which can cause timeout issues (I think we're seeing this on the oldstable Ubuntu job)

internal/cmd/build/main.go
```diff
@@ -165,9 +165,10 @@ func (b *builder) integrationTest() error {
 				"--dynamic-config-value", `component.nexusoperations.useSystemCallbackURL=false`,
 				"--dynamic-config-value", `component.nexusoperations.callback.endpoint.template="http://localhost:7243/namespaces/{{.NamespaceName}}/nexus/callback"`,
 				"--dynamic-config-value", "nexusoperation.enableStandalone=true",
-				"--dynamic-config-value", "history.enableChasmCallbacks=true",
+				"--dynamic-config-value", "history.enableCHASMCallbacks=true",
 				"--dynamic-config-value", "frontend.ListWorkersEnabled=true",
 				"--dynamic-config-value", "activity.startDelayEnabled=true",
+				"--dynamic-config-value", "history.enableUpdateCallbacks=true",
```
yuandrew, line 171: Should we also enable this in `dynamic-config-custom.yaml`? The current Docker Compose job in CI is failing, i think due to missing this DC

internal/link_converter.go
```diff
@@ -80,6 +80,22 @@ func ConvertLinkWorkflowEventToNexusLink(we *commonpb.Link_WorkflowEvent) nexus.
 	}
 }
 
+func ConvertWorkflowlinkToNexusLink(wl *commonpb.Link_Workflow) nexus.Link {
```
yuandrew: missed capitalizing the first "link"
```suggestion
func ConvertWorkflowLinkToNexusLink(wl *commonpb.Link_Workflow) nexus.Link {
```

internal/client.go
mavemuri: removed the comment, added it for StartWorkflowOptions as well 
cc @Quinn-With-Two-Ns

internal/interceptor.go
mavemuri: ClientUpdateWorkflowInput needs those fields because it copies the fields from UpdateWorkflowOptions and finally uses them when creating the createUpdateWorkflowRequest. IIUC interceptors should just pass along the input and modify what they need instead of recreating completely

Only way to fully avoid is maybe passing these values(requestID, callbacks, responseInfo, links) inside context - that wya, we can just read from context when creating the createUpdateWorkflowRequest 

But the current way in repo - from StartWorkflow- was to add these in the options/input

Please lmk if we need further discussion on this

internal/internal_workflow_client.go
mavemuri: removed comment

temporalnexus/temporal_operation.go
mavemuri: Oh just used as shorthand for "network" - fixed it

temporalnexus/temporal_operation.go
mavemuri: Re these two - there has been some discussion about letting the user retry invalid config 
Im not sure on this - i think these specific ones we can immediately give a fatal error - because these are not cases where it could "eventually" become a valid config either 
like client supplying WID that is not yet started - WID empty will always fail 
Will uncomment this section unless there are other objections 
cc @Quinn-With-Two-Ns can you please comment on this too?

temporalnexus/temporal_operation.go
mavemuri: Current version exists as "Verify immediate returns are still async"
Will modify after https://temporalio.atlassian.net/browse/NEXUS-489

temporalnexus/temporal_operation.go
mavemuri: missed it, updated to "workflow-update"
> can we be more explicit about the two-token behaviourin the GoDoc here

not sure if its worth calling out in the go doc- the first token is never given/available to the user - its an internal-only thing 
But can add it to a more appropriate place if needed

temporalnexus/temporal_operation.go
mavemuri: Yes, completed handle without any error is only a retried update currently. If theres an error, we just return that error 
> If a new update completes before UpdateWorkflow returns, could this produce both a synchronous result and a callback

There seems to be an edge case where even if update completed before handle returns, it still comes as async - tracking it through this https://temporalio.atlassian.net/browse/NEXUS-489 - still need to investigate that but its not high priority as updates are mandated to be waitForStage Accepted for Nexus Op and could anytime get async even if it can be completed 

But even if the above is identified/fixed, the code here would just fall in line and return a sync response. In that case, sync response would be for both retried updates and accepted->completed updates

temporalnexus/link_converter.go
mavemuri: agree, good point, will fix

test/nexus_update_op_test.go
mavemuri: didnt know that- will bump up, thanks!

internal/cmd/build/main.go
mavemuri: Will do!

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +177,158 @@ func StartUntypedWorkflow[R any](
...
+			"unexpected error reconstructing token %w",
+			err,
+		)
+	}
+	return NewAsyncResult[R](token), nil
+}
+
+// Validations to be performed specifically for UpdateWorkflow as a Nexus Operation.
+// Currently only guards against a sync Update being passed to the server. Invalid
+// updates - missing workflowID/updateName/etc - are expected to be retried forever.
```
yuandrew: this is stale now? If I'm understanding https://github.com/temporalio/sdk-go/pull/2417#discussion_r3583190391 correctly, implying these failures are no longer always being retried

yuandrew (approved):

temporalnexus/temporal_operation.go
mavemuri: Yes, thanks for catching! Missed updating this comment, fixed it

temporalnexus/temporal_operation.go
Quinn-With-Two-Ns: We need to retryso the handler author can fix it

temporalnexus/temporal_operation.go
```diff
@@ -155,6 +177,156 @@ func StartUntypedWorkflow[R any](
...
+) (TemporalOperationResult[R], error) {
+	if updateWorkflowOptions.UpdateID == "" {
+		// If an Update ID is not provided, use the RequestID. This is done to protect
+		// against cases where an unset UpdateID request with the same RequestID is
+		// retried due to network glitch and gets new IDs in createUpdateWorkflowInput
+		updateWorkflowOptions.UpdateID = nc.startOperationOptions.RequestID
+	}
+
+	if err := validateUpdateWorkflowNexusOperation(updateWorkflowOptions); err != nil {
+		return TemporalOperationResult[R]{}, &nexus.OperationError{
```
Quinn-With-Two-Ns: This should be a retryable error

Quinn-With-Two-Ns (changes requested):

Quinn-With-Two-Ns (approved):

PR #2425, opened 2026-07-01
Always respect SDK flags regardless of if GetSystemInfo works & enable per-RPC compression downgrade

Sushisource:
## What was changed
Title

## Why?
If `GetSystemInfo` returns `UNIMPLEMENTED` for whatever reason (recent ex: gzip compression not working specifically for this endpoint), then we will fail to respect flags already written into history. Fix that.

Additionally allow downgrading to not using compression if a specific RPC fails when trying to use it.

## Checklist

1. Closes 

2. How was this tested:
Upgrade existing tests

3. Any docs updates needed?

### DIFF
--- a/internal/grpc_dialer.go
+++ b/internal/grpc_dialer.go
@@ -156,7 +156,8 @@ func requiredInterceptors(
 	case nil, *GrpcCompressionGzip:
 		interceptors = append(interceptors, newGzipDowngradeInterceptor())
 	}
-	interceptors = append(interceptors,
+	interceptors = append(
+		interceptors,
 		// Performs retries *IF* retry options are set for the call.
 		grpc_retry.UnaryClientInterceptor(),
 		// Prevents retrying grpc message too large errors, while allowing retries of other resource exhausted errors.
--- a/internal/grpc_dialer.go
+++ b/internal/grpc_dialer.go
@@ -195,7 +195,6 @@ func newGzipDowngradeInterceptor() grpc.UnaryClientInterceptor {
 		}
 		msg := strings.ToLower(err.Error())
 		if status.Code(err) != codes.Unimplemented ||
-			!strings.Contains(msg, grpcgzip.Name) ||
 			(!strings.Contains(msg, "decompress") &&
 				!strings.Contains(msg, "grpc-encoding") &&
 				!strings.Contains(msg, "compressor")) {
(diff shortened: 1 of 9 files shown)


Sushisource:
@claude Review

internal/internal_workflow_client.go
```diff
@@ -1609,6 +1609,12 @@ func (wc *WorkflowClient) loadCapabilities(ctx context.Context) (*workflowservic
 	return capabilities, nil
 }
 
+func isUnknownMethodUnimplemented(err error) bool {
+	var unimplemented *serviceerror.Unimplemented
+	return (errors.As(err, &unimplemented) || status.Code(err) == codes.Unimplemented) &&
+		strings.Contains(strings.ToLower(err.Error()), "unknown method")
+}
```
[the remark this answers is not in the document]
Sushisource: Not exactly the most useful, but, sure.

internal/grpc_dialer.go
```diff
@@ -147,13 +151,19 @@ func requiredInterceptors(
 		// By default the grpc retry interceptor *is disabled*, preventing accidental use of retries.
 		// We add call options for retry configuration based on the values present in the context.
 		retry.NewRetryOptionsInterceptor(excludeInternalFromRetry),
+	}
+	switch clientOptions.ConnectionOptions.GrpcCompression.(type) {
+	case nil, *GrpcCompressionGzip:
+		interceptors = append(interceptors, newGzipDowngradeInterceptor())
+	}
+	interceptors = append(interceptors,
```
yuandrew:
```suggestion
	interceptors = append(
	    interceptors,
```

internal/internal_flags.go
```diff
@@ -113,17 +113,16 @@ func newSDKFlagSet(capabilities *workflowservice.GetSystemInfoResponse_Capabilit
 	}
 }
 
-// tryUse returns true if this flag may currently be used. If record is true, always returns
-// true and records the flag as being used.
+// tryUse returns true if this flag may currently be used.
```
yuandrew, line 116: Why remove this part of the docstring?

yuandrew (approved):

internal/internal_flags.go
Sushisource: Because the "records flag" part doesn't always happen now

PR #2439, opened 2026-07-07
[AI-222] Add ADK Go (v2) Plugin

DABH:
## What

Adds `contrib/googleadk`: an integration that makes [Google ADK](https://google.github.io/adk-docs/) (`adk-go`) agents **durable and replay-safe under Temporal** without rewriting the agent. The agent's orchestration loop runs inside a Temporal Workflow; the model call runs as a durable Activity, and I/O tools opt into Activities.

Targets **ADK-Go v2** (`google.golang.org/adk/v2`). The seams this relies on (the `platform` TaskRunner/time/UUID providers, `tool/toolutils.PackTool`, and the `model.Register`/`NewLLM` registry) merged upstream after the v2.0.0 release, so `go.mod` pins `adk/v2` to a `main`-branch pseudo-version until a release ships with them.

## Design

You build the agent the native ADK way (`llmagent.New` + `runner.New`) and change two things:

1. **Model** — set `googleadk.NewModel("<model-name>")` as the agent's `Model`. It's a `model.LLM` whose `GenerateContent` dispatches the `InvokeModel` Activity; the real model is reconstructed **worker-side** by a `ModelFactory` (credentials never enter the workflow). Model calls are always Activities.
2. **Context** — pass `googleadk.NewContext(ctx)` to `Runner.Run`. It installs Temporal-deterministic `time` / `uuid` / task-fan-out providers on ADK's `platform` seams.

**Tools run in-workflow by default** — the idiomatic Temporal model: the workflow is deterministic, and anything doing I/O opts into an Activity. Opt in with `googleadk.ActivityAsTool` (exposes an existing `func(ctx, TArgs) (TResults, error)` Temporal activity as a tool) or `googleadk.NewMCPToolset` (workflow-side proxy; the live MCP session runs worker-side). Worker-side model/MCP handlers live in the registry built by `googleadk.NewActivities`.

This mirrors the Python `temporalio.contrib.google_adk_agents` integration: model interception via a `TemporalModel`, tools deterministic-by-default with an explicit activity opt-in — rather than an ADK plugin constructed inside the workflow.

## Features

- Single- and multi-agent (`SubAgents`) trees; in-workflow function tools; `ActivityAsTool`; stateless MCP; Gemini built-in tools.
- **Human-in-the-loop tool confirmation** over ADK's `adk_request_confirmation` flow (`PendingConfirmations` / `ConfirmationResponse`), enabled naturally because tools run in-workflow.
- **Continue-as-new** session snapshot/restore (`SessionSnapshot` / `ExportSession` / `ImportSession`) for long conversations.
- SSE streaming to external consumers via `workflowstreams` (`NewModel(..., WithStreaming(topic, 0))`).
- Typed, classifiable failures (`ApplicationError` tagged `googleadk.ModelError`/`.ToolError`/`.McpError`; `IsNonRetryable`).
- `FakeModel` / `FakeMCPServer` test helpers for LLM-free unit tests.

## Notes for reviewers

- Deterministic-by-default tools fix a correctness gap in the earlier revision of this branch: when every tool was forced through an Activity, a tool's session-state mutations (`ctx.State().Set`, `ctx.Actions()`) were silently dropped (read-only snapshot + throwaway `EventActions`). In-workflow tools now see the real, mutable context; there's a regression test.
- The package directory/module is `contrib/googleadk` (lowercasenospaces, matching the other contrib packages; the import needs no alias).
- The runnable example moves to a samples-go PR rather than living in-repo.

## Testing

`gofmt` / `go vet` / `go build ./...` clean and `go test ./...` green (unit tests via `testsuite.TestWorkflowEnvironment`, plus dev-server integration and offline-stubbed gemini e2e tests). New tests cover in-workflow tool state mutation, the HITL confirmation round-trip, and continue-as-new round-trip.


contrib/google_adk_agents/plugin.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// InWorkflowToolNames lists tool names whose handlers run inside the workflow
+	// coroutine instead of being dispatched to the CallTool Activity. Use it only
+	// for pure, deterministic, replay-safe tools — no network, clock, randomness,
+	// or goroutines — because their Run executes in the workflow and is re-executed
+	// on replay, with no Activity-level retry or timeout isolation. This is for
+	// plain function tools; ActivityAsTool and MCP tools are inherently
+	// activity-bound and will error if named here. ADK's built-in control tools
+	// (transfer_to_agent, exit_loop) always run in-workflow regardless of this list.
+	InWorkflowToolNames []string
```
brianstrauch: This is the opposite of our typical approach. Tools should default to in-workflow (deterministic), and then users can opt-in to running them as activities, but only if they need to.

contrib/google_adk_agents/examples/weatheragent/main.go
```diff
@@ -0,0 +1,181 @@
+// Copyright 2026 Google LLC, Temporal Technologies Inc.
```
brianstrauch: Move to samples-go repo?

contrib/google_adk_agents/README.md
```diff
@@ -0,0 +1,241 @@
...
+1. the ADK `*plugin.Plugin` from `googleadk.Plugin(...)` on your `runner.Config`; and
+2. the bridged context from `googleadk.NewContext(workflowCtx)` passed to `r.Run`.
+
+The real model and tool handlers run **worker-side**, behind the Activity
+boundary, from a registry you build with `googleadk.NewActivities(...)`.
+
+## Install
+
+```go
+import googleadk "go.temporal.io/sdk/contrib/google_adk_agents"
```
brianstrauch: Did you mean to put a "go get" command here?

contrib/google_adk_agents/README.md
```diff
@@ -0,0 +1,241 @@
...
+	"google.golang.org/adk/agent"
+	"google.golang.org/adk/agent/llmagent"
+	"google.golang.org/adk/model"
+	"google.golang.org/adk/model/gemini"
+	"google.golang.org/adk/plugin"
+	"google.golang.org/adk/runner"
+	"google.golang.org/adk/session"
+	"google.golang.org/genai"
+
+	googleadk "go.temporal.io/sdk/contrib/google_adk_agents"
```
brianstrauch: Name the directory after the package (googleadk) so users don't have to alias all over the place? All of our other contrib packages follow the lowercasenospaces convention

contrib/google_adk_agents/README.md
```diff
@@ -0,0 +1,241 @@
...
+
+	// NewContext bridges the workflow.Context into the context ADK reads its
+	// determinism/executor seams from. Pass it straight to Run.
+	adkCtx := googleadk.NewContext(ctx)
+	msg := genai.NewContentFromText(question, genai.RoleUser)
+
+	var answer string
+	for ev, rerr := range r.Run(adkCtx, "user-1", "session-1", msg, agent.RunConfig{}) {
+		if rerr != nil {
+			return "", rerr
```
brianstrauch: More idiomatic this way, IMO
```suggestion
	for ev, err := range r.Run(adkCtx, "user-1", "session-1", msg, agent.RunConfig{}) {
		if err != nil {
			return "", err
```

contrib/google_adk_agents/README.md
```diff
@@ -0,0 +1,241 @@
...
+
+const taskQueue = "adk"
+
+// AgentWorkflow runs a native ADK agent. Every model/tool call inside r.Run is
+// short-circuited into a Temporal Activity by the plugin.
+func AgentWorkflow(ctx workflow.Context, question string) (string, error) {
+	pl, err := googleadk.Plugin(googleadk.Options{TaskQueue: taskQueue})
+	if err != nil {
+		return "", err
+	}
```
brianstrauch: Creating the plugin in the Workflow doesn't make sense because we don't care about durability/replay, and diverges substantially from our other plugins. I get that we want to pass it to the ADK runner config, but could we split out the things that the runner needs from the plugin?

contrib/google_adk_agents/README.md
```diff
@@ -0,0 +1,241 @@
...
+ADK emits its own OpenTelemetry spans; register your tracing interceptor on the
+worker as usual.
+
+## Determinism & limits
+
+- **Supported:** single- and multi-agent (`SubAgents`) trees, `functiontool`,
+  stateless MCP, Gemini built-in tools (executed server-side inside `InvokeModel`),
+  pure control tools, the in-memory session service, and SSE streaming.
+- **Not in v1:** `RunLive` (bidi), agent-as-tool / sub-agent-as-child-workflow,
+  live memory/artifact tools, HITL tool confirmation, continue-as-new state carry,
```
brianstrauch: HITL tool confirmation and continue-as-new state carry seem important enough to include in a v1

contrib/google_adk_agents/GETTING_STARTED.md
```diff
@@ -0,0 +1,91 @@
+# Getting started
```
brianstrauch: I've never seen a GETTING_STARTED.md file before. Maybe the README.md is a better place for this? I wouldn't know to look for it.

DABH: Yeah, this was in there to help while we were pointing at my fork of adk-go and doing janky things like that.  Now that everything is cleaned up and we're merging to `main` here, this file should be able to go away, and any tips that are still relevant can be folded into README.md.

contrib/google_adk_agents/examples/weatheragent/main.go
DABH: Done, moved to draft temporalio/samples-go PR

contrib/google_adk_agents/README.md
DABH: Yes, good catch; added `go get go.temporal.io/sdk/contrib/googleadk` to the Install section (the import alone wasn't enough).

contrib/google_adk_agents/README.md
DABH: Done

contrib/google_adk_agents/README.md
DABH: Incorporated this change

contrib/google_adk_agents/README.md
DABH: Agreed. There's no longer any plugin constructed in the workflow. Model interception is now a TemporalModel (a model.LLM you set as the agent's Model, à la Python's TemporalModel), and tools handle their own dispatch, so googleadk.Plugin/Options and the BeforeModel/BeforeTool callback seam are gone entirely. The only workflow-side wiring left is NewContext(ctx) for the determinism providers.

contrib/google_adk_agents/README.md
DABH: Included both in cd226e2. HITL rides ADK's native adk_request_confirmation flow - PendingConfirmations(events) surfaces a pause and ConfirmationResponse(...) resumes it (this became natural once tools run in-workflow, since RequestConfirmation can mutate the workflow's EventActions). Continue-as-new gets ExportSession/ImportSession over a JSON-serializable SessionSnapshot, plus a documented pattern in the README. Tests: TestHITLConfirmationResumesTool, TestContinueAsNewRoundTrip.  Lmk if you spot any issues with these new additions.

contrib/google_adk_agents/plugin.go
DABH: Good call.  I redid this.  Ordinary function tools now run in-workflow on the deterministic dispatcher, and I/O tools opt into an Activity via ActivityAsTool (MCP stays activity-backed). This matches the Python google_adk_agents integration.

This actually also fixes a correctness bug in the old default: a tool forced through the CallTool Activity got a read-only state snapshot and a throwaway EventActions, so any tool doing ctx.State().Set(...) silently lost the mutation; only transfer_to_agent/exit_loop were exempted. In-workflow tools now see the real, mutable context (regression test: TestInWorkflowToolMutatesState).

CallTool/InWorkflowToolNames are removed now ofc.

brianstrauch (approved):

contrib/googleadk/README.md
```diff
@@ -0,0 +1,314 @@
...
+tagged version once a release ships that includes them.
+
+## Module versioning
+
+The Google ADK integration is released as a separate Go module from the core
+Temporal Go SDK. See [CHANGELOG.md](CHANGELOG.md) for release notes.
+
+## Hello world
+
+Two halves: the **worker** registers the real model handler as an Activity; the
```
yuandrew, line 57: Do we also want to point to samples-go? I guess the link would be a dead link until we merge the samples PR, but I'd be okay if we merge in what we expect the link to be. Or we can do a follow up PR once the samples PR goes in

contrib/googleadk/context.go
```diff
@@ -0,0 +1,161 @@
...
+		case len(tasks) == 0:
+			return
+		case sequential || len(tasks) == 1:
+			// Tasks run on the calling coroutine, so they dispatch on whatever
+			// workflow.Context runnerCtx already carries (the root context).
+			for _, t := range tasks {
+				t(runnerCtx)
+			}
+		default:
+			done := workflow.NewChannel(ctx)
```
yuandrew: Could we derive the active `workflow.Context` from `runnerCtx` here instead of using the captured root ctx? During nested tool fan-out, the root coroutine is already blocked on the outer join, so `done.Receive(ctx, ...)` can panic with “trying to block on coroutine which is already blocked.” Please use the per-invocation context for `NewChannel`, `workflow.Go`, and `Receive`, and add nested fan-out coverage.

contrib/googleadk/mcp.go
```diff
@@ -0,0 +1,241 @@
...
+		"tool %q not found in MCP toolset %q", in.Tool, in.Toolset)
+}
+
+func (a *Activities) mcpToolset(ctx context.Context, name string) (tool.Toolset, error) {
+	f, ok := a.mcp[name]
+	if !ok {
+		return nil, newApplicationError(ErrorTypeMCP, false, nil,
+			"no MCP toolset %q registered; add it to Config.MCPToolsets", name)
+	}
+	ts, err := f(ctx)
```
yuandrew, line 293: How is the toolset returned by this factory cleaned up? `mcptoolset.Tools` lazily opens and retains a `ClientSession`, and the MCP SDK says the client must close that session. For `CommandTransport`, closing the session is also what terminates the subprocess. Since both activities construct and then discard the toolset without a cleanup path, this appears to leak the session/process. Should `MCPFactory` include an explicit lifecycle contract or return a cleanup function?

contrib/googleadk/activities.go
```diff
@@ -0,0 +1,110 @@
...
+
+// registerGeminiOnce registers a Gemini factory in ADK's model registry the
+// first time Activities are set up. In ADK v2 model/gemini no longer
+// self-registers, yet InvokeModel resolves unmapped model names via
+// model.NewLLM; without this, "gemini-*" names would not resolve unless the
+// caller supplied an explicit Config.Models factory. It runs once because
+// model.NewLLM treats two patterns matching the same name as ambiguous.
+var registerGeminiOnce sync.Once
+
+func registerGemini() {
```
yuandrew: Codex is claiming because we unconditionally register the same process-global Gemini regex that upstream ADK tells applications to register, an existing identical registration leads to a panic, and that we should avoid global mutation and resolve Gemini locally, or make registration caller-owned.

contrib/googleadk/model.go
```diff
@@ -0,0 +1,391 @@
...
+		if !ok {
+			yield(nil, errMissingContext)
+			return
+		}
+		ao := m.activityOptions
+		if ao.StartToCloseTimeout == 0 {
+			ao.StartToCloseTimeout = defaultModelTimeout
+		}
+		if m.streamingTopic != "" && ao.HeartbeatTimeout == 0 {
+			ao.HeartbeatTimeout = defaultStreamHeartbeat
```
yuandrew, line 116: we install a 30s timeout here, but `model.go:229` heartbeats only after receiving a chunk. A delay over 30s retries an otherwise healthy 2 minute (start to close timeout) request, potentially duplicating paid model calls. Maybe we heartbeat periodically while waiting or remove the implicit timeout?

contrib/googleadk/continueasnew.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil, fmt.Errorf("googleadk: export session %q: not found", sessionID)
+	}
+	s := resp.Session
+	snap := &SessionSnapshot{
+		AppName:   s.AppName(),
+		UserID:    s.UserID(),
+		SessionID: s.ID(),
+		State:     map[string]any{},
+	}
+	for k, v := range s.State().All() {
```
yuandrew, line 54: We remove `app:`, `user:`, and `temp:` from `snap.State`, but we then copy all `s.Events().All()` immediately after. Codex claims these events can still contain entries like `event.Actions.StateDelta["app:setting"] = oldValue`. And during import, line 87 re-appends every event and ADK's session service reapplies each event's state delta, including `app:` and `user:` entries. 

Codex claims the fix is to deep-copy exported events and remove non-session state deltas from those copies. A test should use all state scopes and verify that only session-scoped state survives export/import.

contrib/googleadk/tool.go
```diff
@@ -0,0 +1,340 @@
...
+// mock: a tool that reads them across the Activity boundary should see "unset".
+func (c *activityToolContext) IsolationScope() string          { return "" }
+func (c *activityToolContext) ResumedInput(string) (any, bool) { return nil, false }
+
+func (c *activityToolContext) SearchMemory(context.Context, string) (*memory.SearchResponse, error) {
+	return nil, newApplicationError(ErrorTypeTool, false, nil,
+		"SearchMemory is not available inside a Temporal Activity")
+}
+
+func (c *activityToolContext) ToolConfirmation() *toolconfirmation.ToolConfirmation { return nil }
```
yuandrew: Codex is claiming ADK's built-in MCP confirmation flow can't work through this proxy. Because we always return `nil` from `ToolConfirmation`, we return a non-retryable error from `RequestConfirmation`.

`mcp.go:118` doesn't  handle confirmation before dispatching the activity, so an MCP tool using ADK's standard confirmation option fails rather than producing a pending confirmation.

contrib/googleadk/tool.go
```diff
@@ -0,0 +1,340 @@
...
+// produce well-formed arguments — and recurses into nested structs and slices.
+func schemaFromStruct(t reflect.Type) *genai.Schema {
+	s := &genai.Schema{Type: genai.TypeObject, Properties: map[string]*genai.Schema{}}
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if !f.IsExported() {
+			continue
+		}
+		name := f.Name
+		if tag := f.Tag.Get("json"); tag != "" && tag != "-" {
```
yuandrew:
```
  type Args struct {
      Query  string `json:"query"`
      Secret string `json:"-"`
  }
```

I think the current impl would include `Secret`, even though JSON conversion excludes it.

contrib/googleadk/tool.go
```diff
@@ -0,0 +1,340 @@
...
+// error); the tool's parameter schema is inferred from TArgs. Register the same
+// activity on the worker as usual — the tool dispatches it as a Temporal
+// Activity when the agent calls it.
+func ActivityAsTool(activityFn any, opts ActivityToolOptions) (tool.Tool, error) {
+	v := reflect.ValueOf(activityFn)
+	if v.Kind() != reflect.Func {
+		return nil, fmt.Errorf("googleadk: ActivityAsTool requires a function, got %T", activityFn)
+	}
+	ft := v.Type()
+	if ft.NumIn() < 2 {
```
yuandrew: We're only validating the number, but the error says we've validated the 2 args are context.Context and TArgs. Can we make this check more specific?

contrib/googleadk/model.go
```diff
@@ -0,0 +1,391 @@
...
+	var topic *workflowstreams.TopicHandle
+	if wsc, cerr := workflowstreams.NewClientFromActivity(ctx, opts); cerr != nil {
+		log.Warn("streaming client unavailable; not publishing chunks", "error", cerr)
+	} else {
+		defer func() { _ = wsc.Close(ctx) }()
+		topic = wsc.Topic(in.StreamingTopic)
+	}
+
+	var agg *model.LLMResponse
+	for resp, gerr := range llm.GenerateContent(ctx, in.Request, true) {
```
yuandrew, line 276: Codex left this comment, 

> model.go:223 folds both Gemini’s partial chunks and its already-aggregated terminal response. `contrib/googleadk/model.go:266` also ignores later non-text parts once content exists. Real SSE responses can therefore duplicate answers or drop completed FunctionCall arguments. Treat the final non-partial response as authoritative or implement part-aware aggregation, with tests matching Gemini’s actual iterator shape.

was having trouble verifying the legitimacy myself, so had codex write a test and proposed fix for this issue, https://github.com/temporalio/sdk-go/pull/2466

yuandrew (commented):
overall looks good, a few of the comments left were from Codex, they sounded legit and the consequence if it was right seemed bad enough that I've left them here

contrib/googleadk/README.md
DABH: Added a `## Samples` section linking https://github.com/temporalio/samples-go/tree/main/googleadk — the link goes live when the samples PR merges, per the plan. (01e3360)

contrib/googleadk/context.go
DABH: Good catch. Blocking calls bind to the coroutine stored **in the Context**, so `done.Receive(rootCtx)` from a task coroutine panics once the root is parked on the outer join; in a real agent tree ADK's node runtime swallows the panic and the workflow silently wedges until timeout, which is arguably worse. `newWorkflowTaskRunner` now recovers the invoking coroutine's `workflow.Context` from `runnerCtx` (captured root only as fallback) and uses it for `NewChannel`, `workflow.Go`, and `Receive`.

Nested coverage in `nested_fanout_test.go`: a direct nested `platform.RunTasks` repro (failed pre-fix with exactly the "trying to block on coroutine which is already blocked" panic) and a real `agenttool` tree whose sub-agent turn fans out again (wedged pre-fix). (01e3360)

contrib/googleadk/mcp.go
DABH: Confirmed the leak: `mcptoolset`'s connection refresher retains the lazily-opened `ClientSession` and exposes no Close at the pinned version (the session field is unexported, so even the factory author can't reach it), the MCP SDK requires the client to close the session, and for `CommandTransport` that close is what terminates the subprocess — so per-call construction leaked one session per Activity execution (~once per turn via `ListMcpTools`).

Since upstream gives us nothing to close per call, the viable lifecycle contract is caching: the factory now runs **at most once per toolset name per `Activities` value** (construction errors aren't cached, so Temporal retries reconstruct), and a new `Activities.Close()` closes cached toolsets that implement `Close() error` at worker shutdown (idempotent and terminal). The contract is documented on `MCPFactory`. Tests: `mcp_lifecycle_test.go` (factory-runs-once across both activities — was 6 calls pre-fix; Close closes exactly once; failed construction not cached) plus `mcp_closed_internal_test.go`. (01e3360)

contrib/googleadk/activities.go
DABH: Codex is right. `model.Register` panics on a duplicate pattern string, and upstream's registry docs tell applications to register exactly `^(?i)gemini-.*` themselves — I reproduced the panic in both orders (app-first panics `NewActivities` at worker start; package-first panics the app's own `model.Register`), plus the runtime ambiguity failure with an overlapping app pattern like `^gemini-`.

Fixed by never touching the global registry: resolution is now `Config.Models` factory → application-owned registry (`model.NewLLM`) → built-in zero-config Gemini fallback, with the fallback gated on the registry's no-match error specifically, so a failing app-registered gemini factory or an app-caused ambiguity surfaces unchanged instead of being masked. Tests in `registry_internal_test.go`, including registering the upstream-documented pattern after `NewActivities` (panicked pre-fix). (01e3360)

contrib/googleadk/model.go
DABH: Right — the heartbeat timeout counts from Activity **start**, heartbeat timeouts are retryable, and a non-heartbeating attempt only learns of its death through heartbeat responses it isn't sending — so attempt 1 keeps streaming paid tokens while attempt 2 duplicates the call.

Went with heartbeating periodically: `InvokeModel` now runs a background heartbeater for the whole call (interval = `HeartbeatTimeout/3`; the SDK's 0.8×timeout throttle keeps it polite), covering model resolution, the pre-first-chunk window, and the non-streaming branch (which previously never heartbeated at all — a guaranteed false timeout for any user-set `HeartbeatTimeout`); the per-chunk heartbeat is gone. Kept the 30s streaming default since it preserves stuck-worker detection and pre-first-chunk cancellation delivery — the periodic heartbeater just removes the false-timeout mode. Also wired the same helper into `ListMcpTools`/`CallMcpTool` (no-op unless a heartbeat timeout is set). Tests: `heartbeat_test.go` — both failed pre-fix with TIMEOUT_TYPE_HEARTBEAT. (01e3360)

contrib/googleadk/continueasnew.go
DABH: Confirmed — and it's worse than stated: `AppendEvent` applies `app:`/`user:` deltas to **service-level** state, so import resurrected stale values not just on the imported session but on every session of that app/user in the new service (reproduced against the pinned ADK).

`ExportSession` now sanitizes each exported event: shallow copy with a fresh `Actions.StateDelta` stripped of `app:`/`user:`/`temp:` keys (`temp:` defensively — the in-memory service already trims it, but `ExportSession` accepts any `session.Service`). The copies matter: `Get` returns event pointers shared with the service's stored history, so stripping in place would corrupt the live session. `TestExportSessionStripsNonSessionScopedEventDeltas` covers all four scopes in both `snap.State` and event deltas, asserts the original events are unmutated after export, and probes an unrelated second session in the destination service for contamination. (01e3360)

contrib/googleadk/tool.go
DABH: Codex is right, on both passes: `mcptoolset` enforces `RequireConfirmation` **inside the tool's `Run`** — worker-side, across our Activity boundary — where `ToolConfirmation()` returned nil and `RequestConfirmation` errored non-retryably; and on resume the approval couldn't reach the tool since `contextSnapshot` carried no confirmation.

Fixed by tunneling ADK's standard protocol through the boundary in both directions: `contextSnapshot` now carries `ToolConfirmation()` out to the worker; worker-side `RequestConfirmation` records into local `EventActions` exactly like ADK's own context; `CallMcpTool` translates `tool.ErrConfirmationRequired` into a *successful* `mcpCallOutput{Confirmation: …}` (denial → non-retryable error); and the workflow-side proxy re-records the request via `ctx.RequestConfirmation` and returns ADK's sentinel — so the runner emits `adk_request_confirmation` and `PendingConfirmations`/`ConfirmationResponse` work unchanged for MCP tools. Tests: `mcp_hitl_test.go` — pause → approve → exactly-one-execution, and denial-never-executes; both failed pre-fix at the "must pause" assertion.

One caveat found on the way, documented on `ConfirmationResponse`: upstream ADK re-queues a **batch** of approved calls in Go map order (not replay-stable), so multi-decision resumes of Activity-backed tools can reorder Activity commands across replay — resume one decision per Run pass for now; upstream fix being submitted (ADK already fixed the same hazard in `generateRequestConfirmationEvent` but missed the resume processor). (01e3360)

contrib/googleadk/tool.go
DABH: Right — `json:"-"` fields were advertised under their Go field name while the JSON data converter can never deliver them (and anything the model fabricated for `Secret` would be durably recorded in workflow history). Now skipped: the check runs on the raw tag before comma-splitting, so `json:"-,"` still means a field literally named `-` (matching `encoding/json`), and the skip applies recursively to nested structs. Test: `TestActivityAsToolSchemaExcludesJSONDashFields`. (01e3360)

contrib/googleadk/tool.go
DABH: Done — and the loose check was hiding worse than a misleading message: `func(A, B) error` was accepted with the tool schema inferred from the *wrong* struct (`In(1)`), and since the data converter zero-fills mismatched arguments silently, nothing ever errored. Now enforced at construction: exactly 2 inputs, first implements `context.Context` (which automatically rejects `workflow.Context`), and 1–2 returns with the last implementing `error` — mirroring Temporal's own activity contract so error-only activities that already worked keep working — each violation with a precise error. Tests: `tool_validation_test.go` (8 cases; 5 shapes were accepted with `err=nil` pre-fix). (01e3360)

contrib/googleadk/model.go
DABH: Verified the claim against the gemini streaming implementation and #2466's fixture: the stream aggregator forces `Partial=true` on every raw chunk and yields `aggregator.Close()` last as a non-partial response already carrying the full text + function calls — so folding it duplicated the text, and function-call parts never survived `appendText` (a functionCall-only chunk has empty `concatText`), meaning the tool was never invoked from the aggregate the workflow sees.

Adopted #2466's core insight (thanks for the repro — its SSE fixture is reused, credited, in `aggregate_terminal_internal_test.go`) with two amendments: (1) the non-partial response is authoritative for **content** (deep-copied replace) but metadata still folds — Gemini's terminal response empirically carries no `ModelVersion`, which the whole-struct replace in #2466 would drop; (2) a stream carrying several non-partial responses keeps all parts (matching how ADK's flow fully processes each non-partial event) rather than last-wins, protecting custom `ModelFactory` LLMs that never set `Partial`. Tests cover the Gemini-shape aggregate, the ModelVersion guard, deep-copy on the replace path, and the multi-authoritative case. (01e3360)

contrib/googleadk/tool.go
DABH: Upstream fix for the batch-resume ordering: https://github.com/google/adk-go/pull/1169

contrib/googleadk/aggregate_terminal_internal_test.go
```diff
@@ -0,0 +1,175 @@
...
+// chunk Partial and ends with a NON-partial response that already carries the
+// fully aggregated text and function-call parts; folding that terminal
+// response's text into the partial-accumulated aggregate duplicated the
+// message and dropped the function calls (the tool was never invoked
+// in-workflow). A non-partial response must therefore REPLACE
+// partial-accumulated content, while metadata keeps folding — Gemini's
+// terminal response omits fields (e.g. ModelVersion) that earlier chunks
+// carried — and a stream carrying several non-partial responses must keep
+// every part, mirroring how ADK's flow fully processes each non-partial event.
+// The SSE reproduction is adopted from PR #2466.
```
yuandrew: nit: I don't think we need this last line

yuandrew: Same comment for all of the other references of `#2466` here

contrib/googleadk/tool.go
```diff
@@ -108,17 +114,34 @@ type activityTool struct {
...
+	// silently at dispatch (the data converter zero-fills missing arguments
+	// rather than erroring). Outputs mirror Temporal's own activity contract —
+	// one or two return values, the last implementing error.
+	if ft.NumIn() != 2 {
+		return nil, fmt.Errorf("googleadk: ActivityAsTool function must take exactly (context.Context, TArgs), got %d input arguments", ft.NumIn())
+	}
+	// Implements (not type equality) accepts custom context.Context
+	// implementations while still rejecting workflow.Context, whose Done()
+	// returns a workflow.Channel.
+	if !ft.In(0).Implements(reflect.TypeOf((*context.Context)(nil)).Elem()) {
```
yuandrew: I think we want to be even more specific on this check, this can technically pass with any parameter type that implements `context.Context`, since the activity takes in `context.Context` specifically. (comment above would also need to be updated)

```suggestion
	contextType := reflect.TypeOf((*context.Context)(nil)).Elem()
        if ft.In(0) != contextType {
```

contrib/googleadk/tool.go
DABH: As commented above, we're just resuming one decision per Run pass for now, but once upstream merges/ships that PR, I will do an update to our integration and bump the version

contrib/googleadk/aggregate_terminal_internal_test.go
DABH: Removed both `#2466` references from the test comments. (359a08c)

contrib/googleadk/tool.go
DABH: Right — a narrower type (an interface embedding `context.Context`, or a concrete implementation) passes the `Implements` check but fails at dispatch anyway, since the Activity is invoked with a plain `context.Context` value that reflection can't assign to the narrower parameter. Adopted the exact-type check as suggested and updated the comment to explain why equality is required. Added `TestActivityAsToolRejectsNarrowerContextArg` (an interface embedding `context.Context` plus an extra method), which constructed successfully with err=nil under the old check and is now rejected. (359a08c)

yuandrew (approved):

PR #2442, opened 2026-07-09
Auto-enroll pollers into autoscaling on PollerAutoscalingAutoEnroll (Go SDK)

veeral-patel:
## What changed

Wires the server's new `PollerAutoscalingAutoEnroll` namespace capability (see [temporalio/temporal#10688](https://github.com/temporalio/temporal/pull/10688)) into the SDK.

When a namespace advertises the capability, a worker automatically switches a poller type to **poller autoscaling** — but only for poller types the user left at the default (they set neither a fixed count via `MaxConcurrent<Type>TaskPollers` nor a `<Type>TaskPollerBehavior`). Explicitly configured pollers are never touched.

Applies to **workflow, activity, and nexus** pollers.

## Testing

- Unit test for `baseWorker.setTaskPollers` (balancer creation).
- Integration tests: default → autoscaling for workflow/activity/nexus + `serverSupportsAutoscaling` on; capability off → stays fixed at 2; explicit counts → untouched.
- Full `internal` package suite passes; new tests pass under `-race`.


internal/internal_worker.go
```diff
@@ -493,6 +465,60 @@ func (ww *workflowWorker) Stop() {
...
+				params.serverSupportsAutoscaling,
+			),
+		}
+	}
+}
+
+// applyPollerBehavior rebuilds the workflow worker's task pollers to use the
+// given behavior. It must only be called before the worker is started, while
+// the poller list is still dormant.
+func (ww *workflowWorker) applyPollerBehavior(behavior PollerBehavior) {
```
Sushisource, line 513: IMO rather than re-building the pollers in `start` it'd probably make more sense at this point to just delay poller creation until then, but I don't feel super strongly about that.

internal/internal_worker.go
```diff
@@ -217,6 +217,14 @@ type (
 		// NexusTaskPollerBehavior defines the behavior of the nexus task poller.
 		NexusTaskPollerBehavior PollerBehavior
 
+		// The following flags record that the corresponding poller behavior was
+		// left at its default (the user set neither a fixed poller count nor a
+		// poller behavior). They gate poller-autoscaling auto-enrollment when the
+		// namespace advertises the PollerAutoscalingAutoEnroll capability.
+		workflowTaskPollerBehaviorDefaulted bool
+		activityTaskPollerBehaviorDefaulted bool
+		nexusTaskPollerBehaviorDefaulted    bool
```
Sushisource: I think it'd make sense to wrap these up in a struct since most places they're used are passing through all three values at the same time.

Sushisource (commented):
Overall perfectly sensible IMO, just a few comments but @yuandrew should do the main review.

internal/worker.go
```diff
@@ -433,6 +433,14 @@ type (
 		//
 		// NOTE: Experimental
 		DisablePayloadErrorLimit bool
+
```
yuandrew: I'm not sure this is the right place to store these fields, `WorkerOptions` is used by users to configure workers. There are no other non-exported fields. This would actually break some users that specifying WorkerOptions with unkeyed, positional inputs.

Can we put this in workerExecutionParameters or AggregatedWorker?

internal/internal_worker.go
```diff
@@ -1381,6 +1423,30 @@ func (aw *AggregatedWorker) start() error {
...
 
+	// If the namespace opts workers into poller autoscaling, auto-enroll any
+	// poller type that was left at its default (the user set neither a fixed
+	// poller count nor a poller behavior). Auto-enroll implies full autoscaling
+	// support, including scaling down, so it also enables serverSupportsAutoscaling.
+	// The workers are constructed but dormant at this point (no baseWorker.Start
+	// has run yet), so swapping their poller lists in place is safe. The nexus
+	// worker is built further below from aw.executionParams, so mutating the
+	// parameter is sufficient for it.
+	if nsData.capabilities.GetPollerAutoscalingAutoEnroll() {
```
yuandrew, line 1434: `sessionWorker.activityWorker` is skipped in this conversion. `sessionworker.creationWorker` should not be converted though

yuandrew (commented):
We need to update the docs to include this possibility, like the docstring of `MaxConcurrentActivityTaskPollers`.

Also need a CHANGELOG.md entry for this change.

internal/internal_worker.go
yuandrew: I also prefer delaying poller creation so we're not rebuilding and just picking on creation, but I don't think it's block-worthy. 

If you do feel like taking on this change, we need to make sure we have proper guards preventing us from starting workers with uninitialized pollers, and thorough testing.

internal/internal_worker_test.go
```diff
@@ -1792,6 +1792,103 @@ func (s *internalWorkerTestSuite) TestCreateWorkerWithDataConverter() {
...
+	s.IsType(&pollerBehaviorAutoscaling{}, worker.executionParams.ActivityTaskPollerBehavior)
+	s.IsType(&pollerBehaviorAutoscaling{}, worker.executionParams.NexusTaskPollerBehavior)
+
+	// The actual running pollers reflect the autoscaling structure.
+	for _, p := range worker.workflowWorker.worker.options.taskPollers {
+		s.NotNil(p.autoscalingRunner)
+	}
+	for _, p := range worker.activityWorker.worker.options.taskPollers {
+		s.NotNil(p.autoscalingRunner)
+	}
```
yuandrew, line 1853: can we also mock and start a nexus worker to ensure nexus worker also reflects the autoscaling structure?

internal/internal_worker.go
veeral-patel: Done, created a `pollerBehaviorDefaulted` wrapper struct

internal/internal_worker_test.go
veeral-patel: Done, modified TestPollerAutoscalingAutoEnrollWithDefaults test so it mocks and starts a nexus worker as well

internal/worker.go
veeral-patel: Done, moved these fields to pollerBehaviorDefaulted which I added under workerExecutionParameters. WorkerOptions is unchanged now

internal/internal_worker.go
veeral-patel: Done in this commit: https://github.com/temporalio/sdk-go/pull/2442/changes/090f546f85cf9a9265fe5eff2c4ad8771fcc675d

veeral-patel:
> We need to update the docs to include this possibility, like the docstring of MaxConcurrentActivityTaskPollers.
> Also need a CHANGELOG.md entry for this change.

I updated the docstrings for MaxConcurrentActivityTaskPollers, MaxConcurrentWorkflowTaskPollers, and MaxConcurrentNexusTaskPollers and added a CHANGELOG.md entry.

internal/internal_worker.go
```diff
@@ -217,6 +217,12 @@ type (
 		// NexusTaskPollerBehavior defines the behavior of the nexus task poller.
 		NexusTaskPollerBehavior PollerBehavior
 
+		// pollerBehaviorDefaulted records, per poller type, whether the poller
+		// behavior was left at its default. It gates poller-autoscaling
+		// auto-enrollment when the namespace advertises the
+		// PollerAutoscalingAutoEnroll capability.
+		pollerBehaviorDefaulted pollerBehaviorDefaulted
```
yuandrew: I think this name is a little confusing, this is documenting the effective behaviour value, but this field more-so tracks the eligibility for autoscaling enrollment based on Namespace. Do you think something like `pollerAutoEnrollEligibility autoEnrollEligibility` would be more clear?

internal/internal_worker_test.go
```diff
@@ -1889,6 +1907,34 @@ func (s *internalWorkerTestSuite) TestPollerAutoscalingAutoEnrollRespectsExplici
 	s.Equal(3, worker.activityWorker.worker.options.taskPollers[0].pollerCount)
 }
 
+func (s *internalWorkerTestSuite) TestPollerAutoscalingAutoEnrollSessionWorker() {
+	worker := s.newWorkerWithNamespaceCapabilities(
+		&namespacepb.NamespaceInfo_Capabilities{PollerAutoscalingAutoEnroll: true},
+		WorkerOptions{EnableSessionWorker: true},
+	)
+	// A registered activity lets the session worker start.
```
yuandrew: This comment is not needed, can we remove?

internal/internal_worker_test.go
```diff
@@ -1889,6 +1907,34 @@ func (s *internalWorkerTestSuite) TestPollerAutoscalingAutoEnrollRespectsExplici
...
+	worker.RegisterActivity(testActivityNoResult)
+
+	require.NoError(s.T(), worker.Start())
+	defer worker.Stop()
+
+	require.NotNil(s.T(), worker.sessionWorker)
+
+	// The session activity worker polls a normal task queue and is enrolled into
+	// autoscaling like any other activity worker.
```
yuandrew: same with this one, not adding any value beyond the code itself

internal/internal_worker_test.go
```diff
@@ -1889,6 +1907,34 @@ func (s *internalWorkerTestSuite) TestPollerAutoscalingAutoEnrollRespectsExplici
...
+	require.NotEmpty(s.T(), worker.sessionWorker.activityWorker.worker.options.taskPollers)
+	for _, p := range worker.sessionWorker.activityWorker.worker.options.taskPollers {
+		s.NotNil(p.autoscalingRunner)
+	}
+
+	// The session creation worker is deliberately pinned to a single fixed
+	// poller and must not be converted.
```
yuandrew: and this

internal/internal_worker_test.go
```diff
@@ -1792,6 +1793,148 @@ func (s *internalWorkerTestSuite) TestCreateWorkerWithDataConverter() {
 	worker.Stop()
 }
 
+// newWorkerWithNamespaceCapabilities builds an AggregatedWorker whose namespace
+// DescribeNamespace response reports the given capabilities, with permissive
+// polling/shutdown mocks so the worker can be started and stopped.
+func (s *internalWorkerTestSuite) newWorkerWithNamespaceCapabilities(
+	caps *namespacepb.NamespaceInfo_Capabilities, options WorkerOptions,
```
yuandrew: nit: can each arg be on its own line? since the function signature is already multi-lined

internal/internal_worker_test.go
```diff
@@ -1792,6 +1793,148 @@ func (s *internalWorkerTestSuite) TestCreateWorkerWithDataConverter() {
...
+	s.Nil(wfPoller.autoscalingRunner)
+	s.Equal(defaultConcurrentPollRoutineSize, wfPoller.pollerCount)
+
+	require.Len(s.T(), worker.activityWorker.worker.options.taskPollers, 1)
+	actPoller := worker.activityWorker.worker.options.taskPollers[0]
+	s.Nil(actPoller.autoscalingRunner)
+	s.Equal(defaultConcurrentPollRoutineSize, actPoller.pollerCount)
+}
+
+func (s *internalWorkerTestSuite) TestPollerAutoscalingAutoEnrollRespectsExplicitConfig() {
```
yuandrew, line 1887: Can we extend this test to cover explicit *TaskPollerBehavior options and Nexus configuration as well? Right now we're only testing explicit MaxConcurrent*TaskPollers configurations, we should make sure we verify both configuration forms protect the new contract we're adding

internal/worker.go
```diff
@@ -152,7 +152,12 @@ type (
 		// temporal-server to retrieve activity tasks. Changing this value will affect the
 		// rate at which the worker is able to consume tasks from a task queue.
 		//
-		// NOTE: This option is mutually exclusive with WorkflowTaskPollerBehavior.
+		// NOTE: This option is mutually exclusive with ActivityTaskPollerBehavior.
```
yuandrew, line 202: nice :) thanks for this fix

internal/worker.go
```diff
@@ -152,7 +152,12 @@ type (
...
+		//
+		// NOTE: If this is left at its default (i.e. neither this nor
+		// ActivityTaskPollerBehavior is set) and the worker's namespace is configured
+		// to auto-enroll workers into poller autoscaling, the worker will automatically
+		// use poller autoscaling for activity tasks instead of a fixed number of pollers.
```
yuandrew: Should we also document the auto-enrollment behavior on the three `*TaskPollerBehavior` fields? Since auto-enrollment occurs only when both this field and the corresponding `MaxConcurrent*TaskPollers` field are unset, we should document both to make this behaviour more discoverable

internal/internal_worker.go
veeral-patel: Done, renamed to autoEnrollEligibility (type) / pollerAutoEnrollEligibility (field).

internal/internal_worker_test.go
veeral-patel: Done, removed

internal/internal_worker_test.go
veeral-patel: Done, removed

internal/internal_worker_test.go
veeral-patel: Done, removed

internal/internal_worker_test.go
veeral-patel: Done

internal/internal_worker_test.go
veeral-patel: Done, split into MaxConcurrentTaskPollers and TaskPollerBehavior subtests, both now covering Nexus too.

internal/worker.go
```diff
@@ -391,6 +391,11 @@ type (
 		//
 		// NOTE: This option is mutually exclusive with MaxConcurrentWorkflowTaskPollers.
 		//
+		// NOTE: If this is left unset (i.e. neither this nor MaxConcurrentWorkflowTaskPollers
```
yuandrew: > If this is left unset

This is misleading, it should just be "If neither this nor MaxConcurrentWorkflowTaskPollers is set"

internal/worker.go
veeral-patel: Done, added the auto-enroll note to all three *TaskPollerBehavior fields

internal/worker.go
yuandrew: same applies with MaxConcurrent*TaskPollers comments

yuandrew (approved):

internal/worker.go
veeral-patel: Reworded all 6!

veeral-patel:
Tested this change manually successfully.

Setup: Built temporal server from my branch, 1 worker, 2 namespaces

Phase 1 — frontend.pollerAutoscalingAutoEnroll unset (both namespaces AutoEnroll=false):

```
| ns | activity_task | workflow_task | workflow_sticky_task |
|---|---|---|---|
| autoscale-ns | 2 | 1 | 1 | ← fixed SimpleMaximum(2), stable |
| default | 2 | 1 | 1 | ← fixed, stable |
```

Phase 2 — flag enabled for autoscale-ns only (hot-reloaded; probe confirmed autoscale-ns=true, default=false):

```
| ns | activity_task | workflow_task | workflow_sticky_task |
|---|---|---|---|
| autoscale-ns | 5 | 5 | 5 | ← autoscaling (initial=5) |
| default | 2 | 1 | 1 | ← still fixed |
```

PR #2444, opened 2026-07-09
fix: Avoid blocking eager TryReserveSlot during resource tuner ramp wait

### LINKED ISSUE #2443: `ResourceBasedSlotSupplier`: eager `TryReserveSlot` can block for up to `RampThrottle`
## Summary

When resource-based tuning is enabled with a non-zero `RampThrottle` (the default for activities is 50ms), the non-blocking `TryReserveSlot` -- used for **eager** activity/workflow dispatch -- can block for up to `RampThrottle` while another goroutine is waiting out the ramp in `ReserveSlot`. Eager dispatch exists to start a task immediately and skip a poll round-trip, so this adds latency in exactly the place the SDK is trying to remove it.

## Expected Behavior

`TryReserveSlot` returns promptly -- a permit if one is available, otherwise `nil` -- regardless of whether other goroutines are currently waiting out the ramp throttle inside `ReserveSlot`.

## Actual Behavior

`TryReserveSlot` acquires `lastIssuedMu`, and `ReserveSlot` holds that **same** mutex across its `time.After(mustWaitFor)` ramp-throttle wait (up to `RampThrottle`). So while any poller is in the ramp wait, a concurrent eager `TryReserveSlot` blocks on the mutex for up to `RampThrottle`.

```go
// internal/resource_tuner.go -- ReserveSlot
if r.options.RampThrottle > 0 {
    r.lastIssuedMu.Lock()
    mustWaitFor := r.options.RampThrottle - time.Since(r.lastSlotIssuedAt)
    if mustWaitFor > 0 {
        select {
        case <-time.After(mustWaitFor): // lock held for up to RampThrottle
        case <-ctx.Done():
            r.lastIssuedMu.Unlock()
            return nil, ctx.Err()
        }
    }
    r.lastIssuedMu.Unlock()
}
```

1. Configure resource-based tuning with `RampThrottle > 0` (e.g. `NewResourceBasedTuner` with `ActivityRampThrottle`, or just the activity default of 50ms).
2. While one goroutine is in `ReserveSlot`'s ramp wait, call `TryReserveSlot` from another goroutine.
3. Observe that `TryReserveSlot` blocks for ~`RampThrottle` instead of returning immediately. (A focused test with `RampThrottle = 2s` measures a ~2s block.)

## Why this matters

- Eager dispatch is a latency optimization; blocking its `Try*` path defeats the purpose.
- Under multi-poller contention this serializes eager reservations behind ramp waits.
- The ramp throttle does **not** depend on holding the lock during the wait: `TryReserveSlot` already enforces spacing via its own `time.Since(r.lastSlotIssuedAt) > r.options.RampThrottle` check under the lock. So the wait can happen outside the critical section with no behavior change.

## Suggested fix

Release `lastIssuedMu` before the wait -- snapshot under the lock, unlock, then wait:

```go
if r.options.RampThrottle > 0 {
    r.lastIssuedMu.Lock()
    mustWaitFor := r.options.RampThrottle - time.Since(r.lastSlotIssuedAt)
    r.lastIssuedMu.Unlock()
    if mustWaitFor > 0 {
        select {
        case <-time.After(mustWaitFor):
        case <-ctx.Done():
            return nil, ctx.Err()
        }
    }
}
```

Happy to open a PR (with a regression test and a test that confirms the ramp throttle still bounds issuance).

## Specifications

- Version: reproduced on `main` (also present in v1.46.0)
- Platform: any
- Area: `internal/resource_tuner.go` (`worker.ResourceBasedSlotSupplier`)

ridwanakf:
### What

`ResourceBasedSlotSupplier.ReserveSlot` held `lastIssuedMu` across its `time.After(mustWaitFor)` ramp-throttle wait. Because the non-blocking `TryReserveSlot` (used for eager task dispatch) locks the same mutex, a concurrent `TryReserveSlot` could block for up to `RampThrottle` (50ms by default for activities, higher if configured). This releases the lock before the wait.

### Why

Fixes #2443.

Eager dispatch exists to start a task immediately and avoid a poll round-trip; blocking its `Try*` path for up to `RampThrottle` works against that. Holding the lock during the wait is unnecessary: the ramp throttle is independently enforced by `TryReserveSlot`'s own `time.Since(lastSlotIssuedAt) > RampThrottle` check under the lock, so slot issuance stays bounded to ~one per `RampThrottle` regardless of how many `ReserveSlot` goroutines wait concurrently.

### Changes

- `internal/resource_tuner.go`: snapshot `lastSlotIssuedAt` under `lastIssuedMu`, unlock, then wait instead of holding the lock across the wait.
- `internal/resource_tuner_eager_test.go`:
  - `TestTryReserveSlotDoesNotBlockDuringRampWait` -- regression test; fails before the change (eager `TryReserveSlot` blocks ~`RampThrottle`), passes after.
  - `TestRampThrottleStillBoundsIssuance` -- confirms the throttle still bounds issuance to ~one slot per `RampThrottle` after the concurrency change.
- `CHANGELOG.md`: entry under `## [Unreleased]` → `### Fixed`.

### Testing

- `cd internal/cmd/build && go run . check` -- clean (`vet` / `errcheck` / `staticcheck` / `doclink`).
- `go run . unit-test -run 'TestTryReserveSlotDoesNotBlockDuringRampWait|TestRampThrottleStillBoundsIssuance|TestPidDecisions|TestPidDecisionEmitsUsageMetrics'` -- pass (also verified locally under `-race`; regression test verified red→green).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ to docs, or any other relevant information.
   declining slots. Previously the retry loop observed the context only while the ramp throttle was making
   the caller wait, so a poller goroutine could outlive worker shutdown, keeping the worker's stop
   `WaitGroup` from draining and continuing to sample system resources for the life of the process.
+- Resource-based tuner: `TryReserveSlot` (used for eager task dispatch) no longer blocks for up to
+  `RampThrottle` while a concurrent `ReserveSlot` waits out the ramp throttle. The throttle behavior
+  is unchanged; only the unnecessary lock contention on the eager path is removed.
 
 ## [1.46.0] - 2026-07-07
 
--- a/internal/resource_tuner.go
+++ b/internal/resource_tuner.go
@@ -193,15 +193,17 @@ func (r *ResourceBasedSlotSupplier) ReserveSlot(ctx context.Context, info SlotRe
 		if r.options.RampThrottle > 0 {
 			r.lastIssuedMu.Lock()
 			mustWaitFor := r.options.RampThrottle - time.Since(r.lastSlotIssuedAt)
+			// Release before waiting: holding lastIssuedMu across the wait would stall the
+			// non-blocking TryReserveSlot used for eager dispatch. The throttle is still
+			// enforced by TryReserveSlot's own lastSlotIssuedAt check.
+			r.lastIssuedMu.Unlock()
 			if mustWaitFor > 0 {
 				select {
 				case <-time.After(mustWaitFor):
 				case <-ctx.Done():
-					r.lastIssuedMu.Unlock()
 					return nil, ctx.Err()
 				}
 			}
-			r.lastIssuedMu.Unlock()
 		}
 
 		maybePermit := r.TryReserveSlot(info)
--- a/internal/resource_tuner_eager_test.go
+++ b/internal/resource_tuner_eager_test.go
@@ -2,6 +2,8 @@ package internal
 
 import (
 	"context"
+	"sort"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -13,8 +15,8 @@ import (
 	"go.temporal.io/sdk/log"
 )
 
-// countingReserveInfo is a minimal SlotReservationInfo that records how many times
-// NumIssuedSlots is called, so a test can observe that ReserveSlot has begun without a fixed sleep.
+// countingReserveInfo counts NumIssuedSlots calls so a test can observe that ReserveSlot has begun
+// without resorting to a fixed sleep.
 type countingReserveInfo struct {
 	issued         int
 	numIssuedCalls atomic.Int32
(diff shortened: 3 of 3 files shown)


Sushisource (commented):
This makes sense as far as unblocking `TryReserve` goes, but, I'm thinking this could possibly cause some issue with multiple concurrent `ReserveSlot` calls.

EX:

1. A call has finished long enough ago that `mustWaitFor` would be calculated <= 0 
2. Two simultaneous calls enter the critical section at the same time, and both thus decide they do not need to wait
3. Thusly, `RampThrottle` is not appropriately applied to one of the calls when it should have been

Doesn't look like there's any UT that properly captures that idea, hence no test failure, but I think it just means we need a separate lock for this purpose.

ridwanakf:
Hi @Sushisource , Thanks for the feedback! I went to reproduce this and ended up somewhere I didn't expect, so let me lay it out, though you may have context here that I don't.

On current `main` branch, `ReserveSlot` already drops `lastIssuedMu` before anything is decided ([resource_tuner.go#L186-L200](https://github.com/temporalio/sdk-go/blob/8e2c89c7c9d8d41f633bf039063422dd10c1fec5/internal/resource_tuner.go#L186-L200)):

```go
if r.options.RampThrottle > 0 {
	r.lastIssuedMu.Lock()
	mustWaitFor := r.options.RampThrottle - time.Since(r.lastSlotIssuedAt)
	if mustWaitFor > 0 { /* time.After / ctx.Done */ }
	r.lastIssuedMu.Unlock()
}

maybePermit := r.TryReserveSlot(info) // takes the same mutex again
```

The wait happens under the lock, but the lock is released before the call at the bottom, and `TryReserveSlot` takes it again. So two callers that both compute `mustWaitFor <= 0` can already both fall through today, which would mean that part of your example doesn't depend on this PR. CMIIW.

What stops them both getting a slot, as far as I can tell, is `TryReserveSlot` itself ([resource_tuner.go#L208-L226](https://github.com/temporalio/sdk-go/blob/8e2c89c7c9d8d41f633bf039063422dd10c1fec5/internal/resource_tuner.go#L208-L226)). It checks the throttle and stamps `lastSlotIssuedAt` inside a single `lastIssuedMu` section:

```go
r.lastIssuedMu.Lock()
defer r.lastIssuedMu.Unlock()

numIssued := info.NumIssuedSlots()
if numIssued < r.options.MinSlots || (numIssued < r.options.MaxSlots &&
	time.Since(r.lastSlotIssuedAt) > r.options.RampThrottle) {
	decision, err := r.controller.pidDecision(...)
	...
	if decision {
		r.lastSlotIssuedAt = time.Now()
		return &SlotPermit{}
	}
}
```

Whichever caller arrives second sees the first one's stamp, fails the check, gets `nil` back, and goes around the loop again. Since the check and the stamp happen under the same lock, I read that as the throttle being enforced here rather than by the lock spanning the wait, but correct me if I've misread it.

Rather than just assert that, I try creating a test on a different branch off main branch with nothing else in it ([74a2e5d](https://github.com/ridwanakf/sdk-go/commit/74a2e5dcc5dfc7d9edf2e90172965411acf0515d)). This test tries to replicate four goroutines released together into `ReserveSlot`, `RampThrottle` at 100ms, `lastSlotIssuedAt` left at zero so they all skip the wait, which is your steps 1 and 2. It passes on plain `main`, slots coming out ~100ms apart, and passes identically on this PR. Move the `lastSlotIssuedAt` stamp outside `TryReserveSlot`'s lock and it fails straight away at ~8µs, so it isn't passing by accident.

You're right that nothing covered this. `TestRampThrottleStillBoundsIssuance` is sequential and would never have caught a concurrent violation. Glad to fold the new test into this PR wherever you'd like it to live.

Where I do think you're onto something: this PR changes behaviour, just not the bound. On current `main` branch, one `ReserveSlot` goroutine sits in the `time.After` while the rest queue on the mutex; now they all wake together, one wins, and the others drop into the 10ms poll and wait again. Same spacing between slots, more simultaneous wakeups. The separate lock you suggested handles that well: `lastIssuedMu` keeps guarding `lastSlotIssuedAt`, and a new `rampWaitMu` spans the wait, so waiters stay staggered and `TryReserveSlot` stays non-blocking. If I've followed it correctly that would be bounding the wakeups rather than fixing the throttle, though CMIIW there too.

Happy to go whichever way you think is right for the repo: land this with the test, redo it with the separate lock, or close it if the eager-path latency isn't worth the churn. You know how this gets used in practice far better than I do. Let me know and I'll push.

Sushisource:
@ridwanakf Interesting! Thanks for the detailed investigation.

>On current main branch, ReserveSlot already drops lastIssuedMu before anything is decided

Before any permit is reserved, yes certainly, but importantly not before the waiting has happened. Then as you say in TryReserve we check again. That aside though, yeah, I think you're right that the fact that we check again in TryReserve is sufficient.

I'm happy for you to add the test to this PR and I think we're good. Thanks!

ridwanakf:
Hi @Sushisource,  I've added the test in d1a33677. Also on the wording: you're right that the lock does span the waiting, my "before anything is decided" was loose. Thanks for the correction! Would you mind take another look now and see if everything looks good?

Sushisource (approved):

PR #2445, opened 2026-07-09
Report poller scale decisions in worker heartbeat

veeral-patel:
## What changed

The poller autoscaler now counts its scale decisions, keyed by reason (`server_up`, `server_up_suppressed`, `server_down`, `empty_poll_down`, `re_halve`, `error_down`), and each worker heartbeat reports the **per-interval** breakdown per poller type via `WorkerPollerInfo.scale_decisions`.

- `internal_worker_base.go` — autoscaler records decisions into a pre-populated `map[string]*atomic.Int64` (lock-free increment); `drainDecisions()` reads and resets the counts, so each read yields the counts since the previous heartbeat.
- `internal_worker.go` — `collectScaleDecisions()` drains and sums each autoscaler's counts across a worker's pollers by poller type, wired into the heartbeat callback.
- `internal_worker_heartbeat_metrics.go` — sets `WorkerPollerInfo.ScaleDecisions` from that per-interval map.

## Why

Surfaces poller autoscaling behavior — notably server-requested scale-ups suppressed by the throughput gate — to the server via heartbeat, without requiring a metrics pipeline.

## Testing

- `TestPollerAutoscalerDrainDecisions` — drives suppressed scale-ups and a re-halve, then asserts drain returns per-interval counts (first drain has the accumulated counts, a quiet interval is empty, a later decision is a delta of 1 rather than a cumulative total).
- `TestScaleDecisionsHeartbeat` — asserts the per-poller-type breakdown lands on the heartbeat's poller info.

## Dependency

⚠️ Draft: depends on the `scale_decisions` proto field from **temporalio/api#823**. Needs that released and `go.temporal.io/api` bumped in go.mod before CI will build. Verified locally against the generated proto.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -2516,6 +2516,9 @@ func NewAggregatedWorker(client *WorkflowClient, taskQueue string, options Worke
 			if aw.nexusWorker != nil {
 				populateOpts.nexusSlotSupplierKind = aw.nexusWorker.worker.slotSupplier.GetSlotSupplierKind()
 			}
+			// Drain each autoscaler's scale decisions since the last heartbeat, grouped by
+			// poller type.
+			populateOpts.scaleDecisionsByPollerType = aw.collectScaleDecisions()
 			heartbeatTime := time.Now()
 			elapsedSinceLastHeartbeat := heartbeatTime.Sub(previousHeartbeatTime)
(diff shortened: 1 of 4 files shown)


veeral-patel:
Closing in favor of a opt in server side metric - https://github.com/temporalio/temporal/pull/11084

Rather than in worker heartbeat

PR #2447, opened 2026-07-09
Implement preferred version provider callback

### LINKED ISSUE #2446: Add preferred version provider
See https://github.com/temporalio/features/issues/854

Sushisource:
## What was changed
Per a customer request, added a way to specify a "preferred" version to use when making a get version call which may not necessarily be the `maxVersion` which would normally be used when a `GetVersion` call is encountered for the first time.

This is accomplished by adding a worker option: `PreferredVersionProvider` which accepts a callback that can be used to determine the version for a given call.

## Why?
Users would like to have a way to perform a rolling deployment of a worker change that includes an introduction of a `GetVersion` call to a workflow. Without this capability, such a rolling deployment means that if a worker with the new code picks up a task, runs the call, and then the workflow replays on the old code, a WFT failure occurs because the old code is missing the call. (The other direction is acceptable). 

This makes rolling deployments noisy, especially if they proceed slowly.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/2446

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/internal/worker.go
+++ b/internal/worker.go
@@ -94,6 +94,52 @@ type (
 		DefaultVersioningBehavior VersioningBehavior
 	}
 
+	// PreferredVersionProvider provides the version to record for a GetVersion call when the
+	// version marker is created for the first time.
+	//
+	// It is called only during non-replay workflow task execution, before the SDK records a
+	// version marker for the change ID. It is not called during replay or after a version was
+	// already recorded for the same change ID. Returning nil uses the default of MaxSupported.
+	//
+	// The provider runs on a workflow goroutine. It can read a configuration file or another
+	// nondeterministic source because the selected version is recorded in workflow history.
+	// A panic from the provider fails the current workflow task, which Temporal retries.
+	//
+	// NOTE: Experimental
+	//
+	// Exposed as: [go.temporal.io/sdk/worker.PreferredVersionProvider]
+	PreferredVersionProvider func(PreferredVersionProviderInput) *VersionPreference
+
+	// PreferredVersionProviderInput is the input passed to a PreferredVersionProvider.
+	//
+	// NOTE: Experimental
+	//
+	// Exposed as: [go.temporal.io/sdk/worker.PreferredVersionProviderInput]
+	PreferredVersionProviderInput struct {
+		// WorkflowInfo is information about the workflow execution handling the GetVersion call.
+		WorkflowInfo *WorkflowInfo
+		// ChangeID is the change ID passed to GetVersion.
+		ChangeID string
+		// MinSupported is the minimum version supported by the GetVersion call.
+		MinSupported Version
+		// MaxSupported is the maximum version supported by the GetVersion call.
+		MaxSupported Version
+	}
+
+	// VersionPreference is the version preference returned by a PreferredVersionProvider.
+	//
+	// NOTE: Experimental
+	//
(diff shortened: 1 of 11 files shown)


internal/worker.go
```diff
@@ -94,6 +94,52 @@ type (
 		DefaultVersioningBehavior VersioningBehavior
 	}
 
+	// PreferredVersionProvider provides the version to record for a GetVersion call when the
+	// version marker is created for the first time.
+	//
+	// It is called only during non-replay workflow task execution, before the SDK records a
+	// version marker for the change ID. It is not called during replay or after a version was
+	// already recorded for the same change ID. Returning nil uses the default of MaxSupported.
```
dplyukhin:
```suggestion
	// already recorded for the same change ID. Returning nil is equivalent to choosing
	// [PreferredVersionProviderInput.MaxSupported].
```

internal/worker.go
```diff
@@ -353,6 +399,15 @@ type (
 		// for more.
 		DeploymentOptions WorkerDeploymentOptions
 
+		// Optional: Provides the version to record the first time a non-replay
+		// [workflow.GetVersion] call is encountered for a change ID. If unset, or if the provider
+		// returns nil, the SDK records maxSupported. This can support a rolling deployment that
+		// introduces a GetVersion call: initially return [workflow.DefaultVersion] so workers with
```
dplyukhin: Links didn't work because workflow is shadowed.
```suggestion
		// workflow.GetVersion call is encountered for a change ID. If unset, or if the provider
		// returns nil, the SDK records GetVersion's maxSupported argument. This can support a rolling deployment that
		// introduces a GetVersion call: initially return workflow.DefaultVersion so workers with
```

worker/worker.go
```diff
@@ -208,6 +208,21 @@ type (
 	// WorkerDeploymentVersion represents a specific version of a worker in a deployment.
 	WorkerDeploymentVersion = internal.WorkerDeploymentVersion
 
+	// PreferredVersionProvider provides the version to record when a GetVersion marker is first created.
```
dplyukhin:
```suggestion
	// PreferredVersionProvider provides the version to record when a [workflow.GetVersion] marker is first created.
```

workflow/workflow.go
```diff
@@ -497,8 +497,9 @@ const DefaultVersion Version = internal.DefaultVersion
 // It is not allowed to update workflow code while there are workflows running as it is going to break
 // determinism. The solution is to have both old code that is used to replay existing workflows
 // as well as the new one that is used when it is executed for the first time.
-// GetVersion returns maxSupported version when is executed for the first time. This version is recorded into the
-// workflow history as a marker event. Even if maxSupported version is changed the version that was recorded is
+// GetVersion returns maxSupported version when it is executed for the first time, unless the worker configures a
+// [worker.Options.PreferredVersionProvider] that selects a different supported version. This version is recorded into
```
dplyukhin:
```suggestion
// GetVersion returns its maxSupported argument when executed for the first time, unless the worker configures a
// [go.temporal.io/sdk/worker.Options.PreferredVersionProvider]. The returned version is recorded into
```

dplyukhin (approved):
Looks good, just a few broken links and clarifications for the docs.

PR #2449, opened 2026-07-10
Honor WorkflowDefinitionFactory for dynamic workflows

### LINKED ISSUE #2448: RegisterDynamicWorkflow accepts a WorkflowDefinitionFactory but panics on execution
### Expected Behavior

A dynamic (catch-all) workflow registered with a `WorkflowDefinitionFactory` should execute via `NewWorkflowDefinition()`, exactly like a statically-registered factory. `worker.RegisterDynamicWorkflow` already special-cases and stores a factory, so it should work end-to-end (both on real workers and in `testsuite`).

### Actual Behavior

Registration succeeds, but the first workflow task for a dynamically-dispatched type panics inside the SDK:

```
reflect: call of reflect.Value.Call on ptr Value      // pointer factory (e.g. roadrunner)
reflect: call of reflect.Value.Call on struct Value   // value factory (see repro below)
```

On a real worker this surfaces as a `WorkflowTaskFailed` (cause `WORKFLOW_TASK_FAILED_CAUSE_WORKFLOW_WORKER_UNHANDLED_FAILURE`) and the execution never progresses; in `testsuite` it fails the workflow.

### Root Cause

`registry.RegisterDynamicWorkflow` stores a factory as-is (`internal/internal_worker.go`, ~L709):

```go
if factory, ok := wf.(WorkflowDefinitionFactory); ok {
    r.dynamicWorkflow = factory
    r.dynamicWorkflowOptions = options
    return
}
```

…but **both** dispatch paths resolve the `"dynamic"` sentinel to the stored factory and then wrap it in a reflection-based `workflowExecutor` without re-checking the `WorkflowDefinitionFactory` interface:

1. **Production** — `registry.getWorkflowDefinition` (`internal/internal_worker.go`, ~L957):
   ```go
   var dynamic bool
   if d, ok := wf.(string); ok && d == "dynamic" {
       wf = r.dynamicWorkflow          // the factory…
       dynamic = true
   }
   executor := &workflowExecutor{workflowType: lookup, fn: wf, ...}  // …treated as a func → panic
   return newSyncWorkflowDefinition(executor), nil
   ```
2. **Test suite** — `testWorkflowEnvironmentImpl.getWorkflowDefinition` (`internal/internal_workflow_testsuite.go`, ~L681): the same pattern with `workflowExecutorWrapper`.

The named-workflow path a few lines above *does* honor the factory (`return wdf.NewWorkflowDefinition(), nil`); the dynamic path does not. So `RegisterDynamicWorkflow` and both `getWorkflowDefinition`s disagree.

### Steps to Reproduce the Problem

Runnable with public packages only (no `internal` access):

```go
package dynamicrepro

import (
	"testing"
	"time"

	"go.temporal.io/sdk/internalbindings"
	"go.temporal.io/sdk/testsuite"
	"go.temporal.io/sdk/workflow"
)

type dynDef struct{ env internalbindings.WorkflowEnvironment }

func (d *dynDef) Execute(env internalbindings.WorkflowEnvironment, _ *commonpb.Header, _ *commonpb.Payloads) {
	d.env = env
}
func (d *dynDef) OnWorkflowTaskStarted(_ time.Duration) {
	name := d.env.WorkflowInfo().WorkflowType.Name
	result, err := converter.GetDefaultDataConverter().ToPayloads(name)
	d.env.Complete(result, err)
}
func (d *dynDef) StackTrace() string { return "" }
func (d *dynDef) Close()             {}

type dynFactory struct{}

func (dynFactory) NewWorkflowDefinition() internalbindings.WorkflowDefinition { return &dynDef{} }

func TestDynamicWorkflowFactory(t *testing.T) {
	var s testsuite.WorkflowTestSuite
	env.RegisterDynamicWorkflow(dynFactory{}, workflow.DynamicRegisterOptions{})

	env.ExecuteWorkflow("pipeline-blog-publish") // any unregistered type

	require.True(t, env.IsWorkflowCompleted())
	require.NoError(t, env.GetWorkflowError())
	var got string
	require.NoError(t, env.GetWorkflowResult(&got))
	require.Equal(t, "pipeline-blog-publish", got)
}
```

Today this fails with `reflect: call of reflect.Value.Call on struct Value`. This also reproduces in production through `roadrunner-temporal` (the PHP SDK host), which registers every workflow as one shared `WorkflowDefinitionFactory` proxy.

### Proposed Fix

In both `getWorkflowDefinition` implementations, after resolving the dynamic sentinel to the stored factory, re-check the factory interface (mirroring the named path):

```go
if wdf, ok := wf.(WorkflowDefinitionFactory); ok {
    return wdf.NewWorkflowDefinition(), nil
}
```

### Specifications

- **Version:** `go.temporal.io/sdk` v1.44.1 (present on `main` at the referenced lines)
- **Platform:** SDK-internal; platform-independent (repro'd on darwin/arm64)

### Why it matters

This is the one blocker to **Dynamic Workflow support in the PHP SDK** — PHP and TypeScript are the only SDKs without it. `roadrunner-temporal` can register its shared factory dynamically, but every dynamic execution panics until this is fixed.

datashaman:
## What was changed

`RegisterDynamicWorkflow` accepts and stores a `WorkflowDefinitionFactory`, but neither `getWorkflowDefinition` honored it for the dynamic case — the factory was wrapped in a reflection-based `workflowExecutor` and called as a function, panicking with `reflect: call of reflect.Value.Call on {ptr,struct} Value`. Both dispatch paths now re-check the factory interface after resolving the dynamic workflow, matching the existing named-workflow behavior.

- **`internal/internal_worker.go`** — `registry.getWorkflowDefinition` (production path)
- **`internal/internal_workflow_testsuite.go`** — `testWorkflowEnvironmentImpl.getWorkflowDefinition` (test-suite path)

```go
// after resolving the "dynamic" sentinel to the stored factory:
if wdf, ok := wf.(WorkflowDefinitionFactory); ok {
    return wdf.NewWorkflowDefinition(), nil
}
```

## Why

`RegisterDynamicWorkflow` already special-cases and stores a `WorkflowDefinitionFactory`, but the execution paths only honored factories for **named** workflows — so a factory-registered dynamic workflow registered fine and then panicked on the first task, on both real workers and in `testsuite`. This unblocks host processes that serve workflows through a single shared factory — notably [`roadrunner-temporal`](https://github.com/temporalio/roadrunner-temporal) / the PHP SDK — enabling Dynamic Workflow support there (PHP and TypeScript are currently the only SDKs without it).

## How it was tested

`internal/dynamic_workflow_factory_test.go` adds a test per fixed site (both fail before / pass after):

- `TestGetWorkflowDefinition_DynamicWorkflowFactory` — asserts `registry.getWorkflowDefinition` returns the factory's own definition for an unregistered type, not the reflection wrapper.
- `TestDynamicWorkflowFactoryInTestEnvironment` — registers a factory-backed dynamic workflow in `WorkflowTestSuite` and runs an unregistered type end-to-end.

Also verified end-to-end against `roadrunner-temporal` + the PHP SDK: an unregistered workflow type dispatched to a single shared factory completes and reports its real type name.

## Backwards compatibility

No change for func-registered dynamic workflows or any named workflow. The new branch only triggers when the dynamic workflow implements `WorkflowDefinitionFactory`, which previously always panicked.

Fixes #2448


datashaman:
### Testing

Each of the two fixed dispatch paths has a dedicated regression test in `internal/dynamic_workflow_factory_test.go`; both **fail before** the change (panic: `reflect: call of reflect.Value.Call on ...`) and **pass after**:

- `TestGetWorkflowDefinition_DynamicWorkflowFactory` — production `registry.getWorkflowDefinition`: asserts an unregistered type resolves to the factory's own `NewWorkflowDefinition()` rather than the reflection wrapper.
- `TestDynamicWorkflowFactoryInTestEnvironment` — `testWorkflowEnvironmentImpl`: registers a factory-backed dynamic workflow in `WorkflowTestSuite` and runs an unregistered type end-to-end.

Run through the project's build tool:

```
$ cd internal/cmd/build
$ go run . check                 # vet + errcheck + staticcheck + doclink → clean
$ go run . unit-test -run 'TestGetWorkflowDefinition_DynamicWorkflowFactory|TestDynamicWorkflowFactoryInTestEnvironment'
--- PASS: TestGetWorkflowDefinition_DynamicWorkflowFactory (0.00s)
--- PASS: TestDynamicWorkflowFactoryInTestEnvironment (0.00s)
```

(`unit-test` enforces `-race -count 1`.)

Beyond the unit tests, this was validated end-to-end against [`roadrunner-temporal`](https://github.com/temporalio/roadrunner-temporal) + the PHP SDK — the real motivation (#2448): with `roadrunner-temporal` registering its single shared `WorkflowDefinitionFactory` proxy via `RegisterDynamicWorkflow`, starting an **unregistered** workflow type (`pipeline-blog-publish`) is caught by the dynamic handler, which reads the real type name off `Workflow::getInfo()` and completes — returning `{"dispatchedType":"pipeline-blog-publish"}`. Before this fix the same run failed with a `WorkflowTaskFailed` carrying the reflect panic.

internal/internal_workflow_testsuite.go
```diff
@@ -683,6 +683,13 @@ func (env *testWorkflowEnvironmentImpl) getWorkflowDefinition(wt WorkflowType) (
 		wf = env.registry.dynamicWorkflow
 		dynamic = true
 	}
+	// A dynamic workflow may itself be a WorkflowDefinitionFactory (e.g. the
+	// RoadRunner PHP host process registers one shared factory). Honor it rather
+	// than treating it as a plain function, which would panic when the executor
+	// reflects on the factory value.
+	if wdf, ok := wf.(WorkflowDefinitionFactory); ok {
+		return wdf.NewWorkflowDefinition(), nil
+	}
```
yuandrew, line 701: Skipping `workflowExecutorWrapper` is problematic, I think a test using a dynamically dispatched child factory will fail. We may need a new `testWorkflowDefinitionWrapper` wrapper that does bookkeeping similar to `workflowExecutorWrapper`

Something like 
```
type testWorkflowDefinitionWrapper struct {
	WorkflowDefinition
	env     *testWorkflowEnvironmentImpl
	started bool
}

func (w *testWorkflowDefinitionWrapper) OnWorkflowTaskStarted(timeout time.Duration) {
	if !w.started {
		w.started = true
		w.env.workflowFunctionExecuting = true

		if w.env.isChildWorkflow() {
			// Balances ExecuteChildWorkflow's increment, including delayed starts.
			w.env.runningCount--

			if handler := w.env.startedHandler; handler != nil {
				execution := w.env.workflowInfo.WorkflowExecution
				w.env.parentEnv.postCallback(func() {
					handler(execution, nil)
				}, true)
			}
		}
	}

	w.WorkflowDefinition.OnWorkflowTaskStarted(timeout)
}
```

I think 2 tests worth adding would be 
1. A named parent workflow that starts an unknown child handled by the dynamic factory and waits on GetChildWorkflowExecution().
2. A factory definition that calls DrainUnhandledUpdates() and verifies queued updates are drained.

yuandrew (commented):
Thanks for finding this bug and the contribution! One concern with the testsuite fix

internal/internal_workflow_testsuite.go
datashaman: Good catch, thanks. I added a `testWorkflowDefinitionWrapper` around factory-created dynamic workflow definitions in the test environment. On the first workflow task it restores the bookkeeping previously provided by `workflowExecutorWrapper`: marking the workflow function as executing, balancing the child workflow `runningCount`, and scheduling the child-start callback before delegating to the factory definition.

I also added both suggested regression tests:

1. A named parent starts an unknown child handled by the dynamic factory and waits on `GetChildWorkflowExecution()`.
2. A factory definition queues an update, calls `DrainUnhandledUpdates()`, and verifies that the queued update is drained.

The original factory-dispatch tests still pass as well. I rebased onto the latest `main` and ran the focused tests with `-race` plus `go run . check`; everything passes.

yuandrew (approved):
Thanks for the contribution!

PR #2452, opened 2026-07-13
contrib/opentelemetry: opt-in monotonic counters

### LINKED ISSUE #2140: OpenTelemetry incorrect metric type for Counter

yuandrew:
## What was changed
Add in option for monotonic counters

## Why?
See more details in linked issue. Adding this correction as opt-in, to not cause a wider-scoped breaking change. 

This decision can be revisited before `contrib/opentelemetry` reaches version 1.0.0, ideally alongside an explicit policy for handling negative counter increments, such as rejecting them.

Although `MetricsCounter` is documented as ever-increasing, `Inc` accepts an `int64`, so the type system permits negative increments. The existing OpenTelemetry handler uses `Int64UpDownCounter`, and custom user metrics may currently rely on that behavior.

Changing the default to `Int64Counter` would affect all counters, including custom metrics, and negative increments could produce invalid or backend-dependent metric data. The opt-in allows correctly implemented counters to receive the proper monotonic classification without changing existing behavior.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/2140

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Low Risk**
> Scoped to contrib/opentelemetry metrics mapping behind a default-off flag; no auth or core SDK runtime changes.
> 
> **Overview**
> Adds **`UseMonotonicCounters`** on `MetricsHandlerOptions` so OpenTelemetry **`Counter`** can emit **`Int64Counter`** (monotonic sums) instead of the existing **`Int64UpDownCounter`**. Default stays **false**, so metric types and behavior for current users are unchanged.
> 
> When the option is on, tagged handlers inherit the setting via **`WithTags`**, and a new test asserts exported sums have **`IsMonotonic: true`**. The changelog documents the option and warns that decrements/negative **`Inc`** values are unsafe with monotonic counters.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e3283299e5241d5cfadbcacd9af083cf9f943de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/contrib/opentelemetry/handler.go
+++ b/contrib/opentelemetry/handler.go
@@ -109,17 +109,13 @@ func (m MetricsHandler) WithTags(tags map[string]string) client.MetricsHandler {
 }
 
 func (m MetricsHandler) Counter(name string) client.MetricsCounter {
+	var c interface { Add(context.Context, int64, ...metric.AddOption) }
+	var err error
 	if m.useMonotonicCounters {
-		c, err := m.meter.Int64Counter(name)
-		if err != nil {
-			m.onError(err)
-			return client.MetricsNopHandler.Counter(name)
-		}
-		return metrics.CounterFunc(func(d int64) {
-			c.Add(context.Background(), d, metric.WithAttributeSet(m.attributes))
-		})
+	    c, err = m.meter.Int64Counter(name)
+	} else {
+	    c, err = m.meter.Int64UpDownCounter(name)
 	}
-	c, err := m.meter.Int64UpDownCounter(name)
 	if err != nil {
 		m.onError(err)
 		return client.MetricsNopHandler.Counter(name)
--- a/contrib/opentelemetry/handler.go
+++ b/contrib/opentelemetry/handler.go
@@ -35,12 +35,12 @@ type MetricsHandlerOptions struct {
 	//
 	// Optional: Defaults to panicking on any error.
 	OnError func(error)
-	// UseMonotonicCounters causes Counter to use OpenTelemetry's Int64Counter
+	// UseMonotonicCounters causes [MetricsHandler.Counter] to use OpenTelemetry's Int64Counter
 	// instead of Int64UpDownCounter. This allows exporters to identify SDK
 	// counters as monotonic sums.
 	//
-	// MetricsCounter is documented as ever-increasing, so values passed to Inc
-	// must be non-negative. Negative values may produce invalid or
+	// [client.MetricsCounter] is documented as ever-increasing, so values passed to
+	// [client.MetricsCounter.Inc] must be non-negative. Negative values may produce invalid or
 	// backend-dependent metric data.
 	//
 	// Optional: Defaults to false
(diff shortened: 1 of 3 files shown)


contrib/opentelemetry/handler.go
```diff
@@ -34,6 +35,16 @@ type MetricsHandlerOptions struct {
 	//
 	// Optional: Defaults to panicking on any error.
 	OnError func(error)
+	// UseMonotonicCounters causes Counter to use OpenTelemetry's Int64Counter
+	// instead of Int64UpDownCounter. This allows exporters to identify SDK
+	// counters as monotonic sums.
+	//
+	// MetricsCounter is documented as ever-increasing, so values passed to Inc
+	// must be non-negative. Negative values may produce invalid or
+	// backend-dependent metric data.
```
dplyukhin: Looking at the PR with no context, I wasn't sure what Inc was referring to so I added a link. But this might be obvious to everyone already using the API.
```suggestion
	// UseMonotonicCounters causes [MetricsHandler.Counter] to use OpenTelemetry's Int64Counter
	// instead of Int64UpDownCounter. This allows exporters to identify SDK
	// counters as monotonic sums.
	//
	// [client.MetricsCounter] is documented as ever-increasing, so values passed to
	// [client.MetricsCounter.Inc] must be non-negative. Negative values may produce invalid or
	// backend-dependent metric data.
```

contrib/opentelemetry/handler.go
```diff
@@ -89,13 +101,24 @@ func (m MetricsHandler) WithTags(tags map[string]string) client.MetricsHandler {
...
+		c, err := m.meter.Int64Counter(name)
+		if err != nil {
+			m.onError(err)
+			return client.MetricsNopHandler.Counter(name)
+		}
+		return metrics.CounterFunc(func(d int64) {
+			c.Add(context.Background(), d, metric.WithAttributeSet(m.attributes))
+		})
+	}
 	c, err := m.meter.Int64UpDownCounter(name)
```
dplyukhin: Maybe make it a little DRYer?
```suggestion
	var c interface { Add(context.Context, int64, ...metric.AddOption) }
	var err error
	if m.useMonotonicCounters {
	    c, err = m.meter.Int64Counter(name)
	} else {
	    c, err = m.meter.Int64UpDownCounter(name)
	}
```

dplyukhin (approved):

contrib/opentelemetry/handler.go
yuandrew: neat suggestion! I didn't think to def `c` as an interface like this

PR #2459, opened 2026-07-14
Allow query results to use external storage

### LINKED ISSUE #2456: Queries are blocked from external storage due to hardcoded limit
## Expected Behavior

Offloads of queries are ran before the size is judged, allowing for queries to benefit from external storage.

## Actual Behavior

The hard-coded queryResultSizeLimit = 2,000,000 check ([internal_event_handlers.go:1470](https://github.com/temporalio/sdk-go/blob/c9ae98f38ef5745527f40cd7c04178cb09c77846/internal/internal_event_handlers.go#L1470-L1475)) runs first and fails the query if the raw serialized result is over 2 MB -> the offload visitor only runs after that, so it never gets the chance. Query results are still capped at 2 MB even with External Storage on.

kritibehl:
## What was changed

This PR removes the early hardcoded query result size check in `ProcessQuery`.

Query results now continue through the normal outbound payload visitor path, allowing external storage to offload oversized query payloads before payload size enforcement.

The existing external storage query integration test was updated to use a query result larger than 2 MB, verifying that large query results can be stored externally and retrieved by the client.

## Why?

Previously, query results were checked against a hardcoded 2 MB limit immediately after the query handler returned.

That meant external storage never had a chance to replace the large payload with a small external-storage reference, even when external storage was configured.

This made query results behave differently from other outbound payloads that already pass through the external storage visitor before payload size enforcement.

## Checklist

1. Closes #2456

2. How was this tested:

- `go test ./internal -run 'TestPayloadLimits' -count=1`
- `cd test && go test . -run 'TestExternalStorageSuite/TestQuery' -count=3`
- `cd test && go test . -run 'TestExternalStorageSuite' -count=1`

3. Any docs updates needed?

No docs updates needed. This fixes query result handling to use the existing external storage path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ to docs, or any other relevant information.
 
 ### Fixed
 
+- Allow query results to use external storage before payload size enforcement.
 - Correct schedule catch-up window documentation to state that an unset value is omitted and the
   server applies its one-year default.
 - Resource-based tuner: `ReserveSlot` now honors context cancellation while the resource controller is
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -28,7 +28,6 @@ import (
 )
 
 const (
-	queryResultSizeLimit             = 2000000 // 2MB
 	changeVersionSearchAttrSizeLimit = 2048
 )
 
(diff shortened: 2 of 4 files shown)


jmaeagle99 (commented):
Thank you for contributing a fix for the issue! Would like to ask for one adjustment to fix a test. Please update the `TestLargeQueryResultError` integration test to check for "Blob data size exceeds limit." instead of "query result size (3000036) exceeds limit (2000000)".

kritibehl:
Thanks! Updated `TestLargeQueryResultError` to expect `Blob data size exceeds limit.` now that query results flow through the normal payload limit handling path.

jmaeagle99 (approved):

yuandrew:
Can we add a CHANGELOG.md entry for this?

kritibehl:
Added a CHANGELOG.md entry under Unreleased / Fixed. Thanks!

PR #2462, opened 2026-07-15
fix: honor worker stop while a session poller waits for a token

### LINKED ISSUE #2461: session worker Stop() hangs + leaks poller when at max session capacity
## Expected Behavior

`worker.Stop()` returns promptly and does not leak goroutines, including for a session worker that is at its `MaxConcurrentSessionExecutionSize` when it is stopped.

## Actual Behavior

Stop a session worker while every session slot is in use and it blocks for the full `WorkerStopTimeout`, logs `"Worker graceful stop timed out"`, and leaves the session creation poller running.

To be precise about severity: the leaked poller (and the worker slot it holds) is unconditional, but the blocking is bounded by `WorkerStopTimeout`, which defaults to 0. At the default, `Stop()` returns immediately; it still logs the spurious timeout and still leaks the poller. The multi-second block appears once a stop timeout is configured, which workers do so in-flight activities can drain.

The creation poller reserves a slot and then waits for a free session token before it polls. That wait is a `sync.Cond` ([`waitForAvailableToken`, `internal/session.go`](https://github.com/temporalio/sdk-go/blob/52a0ffa3d2f302fc86ba781ee1d0a6d8b8c2d1f3/internal/session.go#L508-L514)):

```go
func (t *sessionTokenBucket) waitForAvailableToken() {
    t.L.Lock()
    defer t.L.Unlock()
    for t.availableToken == 0 {
        t.Wait()
    }
}
```

The Cond is only signalled when a session finishes ([`addToken`](https://github.com/temporalio/sdk-go/blob/52a0ffa3d2f302fc86ba781ee1d0a6d8b8c2d1f3/internal/session.go#L516-L520)). At capacity `availableToken` is 0, so the poller parks in `t.Wait()`. [`Stop()`](https://github.com/temporalio/sdk-go/blob/52a0ffa3d2f302fc86ba781ee1d0a6d8b8c2d1f3/internal/internal_worker_base.go#L888) closes the poller's stop channel but never touches this Cond, and the poller goroutine is exactly the one `Stop()` is waiting on (it holds a `stopWG.Done`), so `awaitWaitGroup` runs out its timeout and the goroutine lingers until a session completes, which during shutdown it will not.

Drain the token bucket so the creation poller parks in the token wait, then time how long `Stop()` takes. On `main` it blocks for the whole `stopTimeout`; a poller that honored the stop lets `Stop()` return at once.

The test drops into `internal/internal_worker_base_test.go` and reuses the helpers already there (`newBaseWorker`, `newBlockingProbeTaskPoller`, `noopTaskProcessor`); the only new piece is a slot supplier that signals once the poller holds a permit and is committed to the token wait.

<details>
<summary>test</summary>

```go
// signals once the poller holds a permit and is about to wait for a session token
type signalOnReserveSlotSupplier struct{ reserved chan struct{} }

func (s *signalOnReserveSlotSupplier) ReserveSlot(ctx context.Context, _ SlotReservationInfo) (*SlotPermit, error) {
	select {
	case <-ctx.Done():
		return nil, ctx.Err()
	default:
	}
	select {
	case s.reserved <- struct{}{}:
	default:
	}
	return &SlotPermit{}, nil
}

func (s *signalOnReserveSlotSupplier) TryReserveSlot(SlotReservationInfo) *SlotPermit {
	return &SlotPermit{}
}

func (s *signalOnReserveSlotSupplier) MarkSlotUsed(SlotMarkUsedInfo) {}

func (s *signalOnReserveSlotSupplier) ReleaseSlot(SlotReleaseInfo) {}

func (s *signalOnReserveSlotSupplier) MaxSlots() int { return 0 }

func TestSessionTokenWaitHonorsStop(t *testing.T) {
	bucket := newSessionTokenBucket(1)
	require.True(t, bucket.getToken(), "drain the only token so the poller parks in the token wait")

	reserved := make(chan struct{}, 1)
	tp := newBlockingProbeTaskPoller()
	defer tp.Close()

	bw := newBaseWorker(baseWorkerOptions{
		slotSupplier:     &signalOnReserveSlotSupplier{reserved: reserved},
		maxTaskPerSecond: 1000,
		taskPollers: []scalableTaskPoller{
			{taskPollerType: "test", pollerCount: 1, taskPoller: tp},
		},
		taskProcessor:      noopTaskProcessor{},
		workerType:         "SessionTokenStopTest",
		logger:             ilog.NewNopLogger(),
		stopTimeout:        5 * time.Second,
		metricsHandler:     metrics.NopHandler,
		sessionTokenBucket: bucket,
	})

	bw.Start()
	<-reserved // the poller holds a permit and is committed to the token wait

	start := time.Now()
	bw.Stop()
	require.Less(
		t,
		time.Since(start),
		time.Second,
		"Stop() should wake a poller waiting for a session token, not block for the full stopTimeout",
	)
}
```

</details>

On `main` this fails every time: `Stop()` takes the full 5s stop timeout instead of returning at once.

## Question

I came across this reading the poller code, not from a production incident, so I would like a sanity check before turning it into a PR. Is the session creation poller supposed to ignore `Stop()` while it waits for a token, or is this just an oversight? I assumed the latter, since the neighbouring waits on this path (the resource tuner, and `FixedSizeSlotSupplier` via the semaphore) all take a context and bail out on cancel.

If it is worth fixing, the change I have in mind is small: give the token bucket a closed state that `Stop()` sets, have `waitForAvailableToken` wake and return once it is closed instead of parking forever, so the poller exits. A `sync.Cond` wait can only be broken by broadcasting and re-checking the predicate, so the stop state has to live where the waiter reads it under the lock. It is additive: the flag defaults off, the token accounting (`getToken`/`addToken`) does not read it, and the bucket is per worker, so nothing else changes. That is what the test above exercises, and it passes with that change. Happy to send it, or to drop this if I have misread how sessions shut down.

## Specifications

- Version: `main` (post-1.46.0). The relevant code looks unchanged for a long time.
- Platform: macOS (arm64), go1.25. Nothing platform-specific about it.

ridwanakf:
Opening this proactively with a fix, on the assumption that this is an unintended shutdown hang rather than designed behavior. If sessions are meant to drain before the worker stops, or I have misread the shutdown path, I am happy to close it.

### What

Stopping a session worker while it is at `MaxConcurrentSessionExecutionSize` blocks for the full `WorkerStopTimeout` and leaks the session creation poller. (`WorkerStopTimeout` defaults to 0, so at the default `Stop()` returns at once and only leaks; the block is real once a stop timeout is configured.)

The creation poller reserves a slot and then waits for a free session token before polling. That wait is a `sync.Cond` with no stop awareness ([`waitForAvailableToken`, `internal/session.go`](https://github.com/temporalio/sdk-go/blob/52a0ffa3d2f302fc86ba781ee1d0a6d8b8c2d1f3/internal/session.go#L508-L514)):

```go
func (t *sessionTokenBucket) waitForAvailableToken() {
    t.L.Lock()
    defer t.L.Unlock()
    for t.availableToken == 0 {
        t.Wait()
    }
}
```

The Cond is only signalled when a session finishes ([`addToken`](https://github.com/temporalio/sdk-go/blob/52a0ffa3d2f302fc86ba781ee1d0a6d8b8c2d1f3/internal/session.go#L516-L520)). At capacity `availableToken` is 0, so the poller parks in `t.Wait()`. [`Stop()`](https://github.com/temporalio/sdk-go/blob/52a0ffa3d2f302fc86ba781ee1d0a6d8b8c2d1f3/internal/internal_worker_base.go#L888) closes the poller's stop channel but never touches this Cond, and that goroutine is the one `Stop()` waits on, so `awaitWaitGroup` times out (`"Worker graceful stop timed out"`) and the poller lingers until a session completes, which during shutdown it will not.

### Why

Fixes #2461.

The other waits on this path already honor the stop signal (`FixedSizeSlotSupplier` via the semaphore's context, and the resource tuner's `ReserveSlot`), so the session token wait looks like an oversight rather than an intentional block.

On the new `closed` field: a `sync.Cond` wait can only be interrupted by broadcasting and having the waiter re-check its predicate; there is no channel or context to select on. So the stop state has to be readable by the waiter under the same lock, and a bool on the bucket is the minimal place for it. It is additive and does not change existing behavior: it defaults to `false`; `getToken` and `addToken` (the accounting behind `CreateSession` and `CompleteSession`) never read it; and the bucket is built once per session worker and is never reused across a restart (a stopped worker cannot be started again), so `close()` only affects a poller currently parked in the wait.

### Changes

- `internal/session.go`: the bucket gains a `closed` flag that its own predicate honors, so `waitForAvailableToken` returns `true` when a token is free and `false` once the bucket is closed. A new `close` method sets the flag and broadcasts the Cond. The stop condition lives in the bucket, not in a channel threaded through the wait, and the loop stays the standard Cond form.
- `internal/internal_worker_base.go`: the two pollers that wait for a token return when it reports the bucket closed; `Stop()` calls `bucket.close()` after closing the worker stop channel.
- `internal/internal_worker_base_test.go`: a regression test (`TestStopReturnsPromptlyWithPollerAwaitingSessionToken`) next to the existing worker-shutdown tests. It drains the token bucket so the creation poller parks in the token wait, then asserts `Stop()` returns well under the stop timeout. Fails on `main` (Stop blocks the full timeout), passes with this change.
- `CHANGELOG.md`: entry under `## [Unreleased]` → `### Fixed`.

### Testing

- `cd internal/cmd/build && go run . check` clean.
- Regression test verified red on `main`, green with the change, `-race`. Worker, session, and poller suites still pass under `-race`.

### DIFF
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -532,8 +532,8 @@ func (bw *baseWorker) runPoller(taskWorker scalableTaskPoller) {
 				}
 				continue
 			}
-			if bw.sessionTokenBucket != nil {
-				bw.sessionTokenBucket.waitForAvailableToken()
+			if bw.sessionTokenBucket != nil && !bw.sessionTokenBucket.waitForAvailableToken() {
+				return
 			}
 			if bw.pollerBalancer != nil {
 				bw.pollerBalancer.incrementPoller(taskWorker.taskPollerType)
@@ -597,8 +597,8 @@ func (bw *baseWorker) runAutoscalingPoller(taskWorker scalableTaskPoller) {
 			continue
 		}
 
-		if bw.sessionTokenBucket != nil {
-			bw.sessionTokenBucket.waitForAvailableToken()
+		if bw.sessionTokenBucket != nil && !bw.sessionTokenBucket.waitForAvailableToken() {
+			return
 		}
 		if bw.pollerBalancer != nil {
 			bw.pollerBalancer.incrementPoller(taskWorker.taskPollerType)
@@ -905,6 +905,9 @@ func (bw *baseWorker) Stop() {
 	}
 	close(bw.stopCh)
 	bw.limiterContextCancel()
+	if bw.sessionTokenBucket != nil {
+		bw.sessionTokenBucket.close()
+	}
 
 	// Wait for pollers, task dispatch, and task processing to complete, or until stopTimeout elapses.
 	// The task dispatcher drains taskQueueCh after the closer goroutine
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -533,6 +533,7 @@ func (bw *baseWorker) runPoller(taskWorker scalableTaskPoller) {
 				continue
 			}
 			if bw.sessionTokenBucket != nil && !bw.sessionTokenBucket.waitForAvailableToken() {
+				bw.releaseSlot(permit, SlotReleaseReasonUnused)
 				return
 			}
 			if bw.pollerBalancer != nil {
@@ -598,6 +599,8 @@ func (bw *baseWorker) runAutoscalingPoller(taskWorker scalableTaskPoller) {
 		}
 
 		if bw.sessionTokenBucket != nil && !bw.sessionTokenBucket.waitForAvailableToken() {
+			bw.releaseSlot(permit, SlotReleaseReasonUnused)
+			releaseActive()
 			return
 		}
 		if bw.pollerBalancer != nil {
--- a/internal/internal_worker_base_test.go
+++ b/internal/internal_worker_base_test.go
@@ -1228,3 +1228,60 @@ func (s *ScalableTaskPollerSuite) TestNewScalableTaskPollerAllTypes() {
 		})
 	}
 }
+
+// signalOnReserveSlotSupplier grants a permit and signals once, so a test can tell the poller has a
+// permit and is committed to the session-token wait before it calls Stop().
+type signalOnReserveSlotSupplier struct{ reserved chan struct{} }
+
+func (s *signalOnReserveSlotSupplier) ReserveSlot(ctx context.Context, _ SlotReservationInfo) (*SlotPermit, error) {
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+	select {
+	case s.reserved <- struct{}{}:
+	default:
+	}
+	return &SlotPermit{}, nil
+}
+func (s *signalOnReserveSlotSupplier) TryReserveSlot(SlotReservationInfo) *SlotPermit {
+	return &SlotPermit{}
+}
+func (s *signalOnReserveSlotSupplier) MarkSlotUsed(SlotMarkUsedInfo) {}
+func (s *signalOnReserveSlotSupplier) ReleaseSlot(SlotReleaseInfo)   {}
+func (s *signalOnReserveSlotSupplier) MaxSlots() int                 { return 0 }
+
+// TestStopReturnsPromptlyWithPollerAwaitingSessionToken covers a session worker at its maximum session
+// count: the creation poller parks in sessionTokenBucket.waitForAvailableToken. Stop() must wake it,
+// rather than block for the whole stopTimeout while the poller goroutine leaks.
+func TestStopReturnsPromptlyWithPollerAwaitingSessionToken(t *testing.T) {
+	bucket := newSessionTokenBucket(1)
+	require.True(t, bucket.getToken(), "drain the only token so the poller parks in the token wait")
+
+	reserved := make(chan struct{}, 1)
+	tp := newBlockingProbeTaskPoller()
+	defer tp.Close()
+
+	bw := newBaseWorker(baseWorkerOptions{
+		slotSupplier:     &signalOnReserveSlotSupplier{reserved: reserved},
+		maxTaskPerSecond: 1000,
+		taskPollers: []scalableTaskPoller{
+			{taskPollerType: "test", pollerCount: 1, taskPoller: tp},
+		},
+		taskProcessor:      noopTaskProcessor{},
+		workerType:         "SessionTokenStopTest",
+		logger:             ilog.NewNopLogger(),
+		stopTimeout:        5 * time.Second,
+		metricsHandler:     metrics.NopHandler,
+		sessionTokenBucket: bucket,
+	})
+
+	bw.Start()
+	<-reserved // the poller holds a permit and is committed to the token wait
+
+	start := time.Now()
+	bw.Stop()
+	require.Less(t, time.Since(start), time.Second,
+		"Stop() should wake a poller waiting for a session token, not block for the full stopTimeout")
+}
--- a/internal/internal_worker_base_test.go
+++ b/internal/internal_worker_base_test.go
@@ -1229,42 +1229,59 @@ func (s *ScalableTaskPollerSuite) TestNewScalableTaskPollerAllTypes() {
 	}
 }
 
-// signalOnReserveSlotSupplier grants a permit and signals once, so a test can tell the poller has a
-// permit and is committed to the session-token wait before it calls Stop().
-type signalOnReserveSlotSupplier struct{ reserved chan struct{} }
+// sessionTokenWaitProbe is a token bucket lock that reports the first entry into
+// waitForAvailableToken. Entry is the signal a test needs: the poller holds this lock from there
+// until Cond.Wait parks it, and close() takes the same lock, so a Stop() beginning after the signal
+// cannot overtake the waiter.
+type sessionTokenWaitProbe struct {
+	sync.Mutex
+	entered chan struct{}
+	once    sync.Once
+}
 
-func (s *signalOnReserveSlotSupplier) ReserveSlot(ctx context.Context, _ SlotReservationInfo) (*SlotPermit, error) {
+func (p *sessionTokenWaitProbe) Lock() {
+	p.Mutex.Lock()
+	p.once.Do(func() { close(p.entered) })
+}
+
+// releaseRecordingSlotSupplier records the reason each permit is released with.
+type releaseRecordingSlotSupplier struct{ released chan SlotReleaseReason }
+
+func (s *releaseRecordingSlotSupplier) ReserveSlot(ctx context.Context, _ SlotReservationInfo) (*SlotPermit, error) {
 	select {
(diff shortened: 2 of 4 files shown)


internal/internal_worker_base.go
```diff
@@ -518,8 +518,8 @@ func (bw *baseWorker) runPoller(taskWorker scalableTaskPoller) {
 				}
 				continue
 			}
-			if bw.sessionTokenBucket != nil {
-				bw.sessionTokenBucket.waitForAvailableToken()
+			if bw.sessionTokenBucket != nil && !bw.sessionTokenBucket.waitForAvailableToken() {
+				return
```
yuandrew, line 537: Receiving `permit` transfers ownership from `reserveSlotAsync` to this poller. Normally `pollTask` then releases it or transfers it with a dispatched task, but this new return happens before that handoff. Please release the permit with `SlotReleaseReasonUnused` before returning.

internal/internal_worker_base.go
```diff
@@ -583,8 +583,8 @@ func (bw *baseWorker) runAutoscalingPoller(taskWorker scalableTaskPoller) {
 			continue
 		}
 
-		if bw.sessionTokenBucket != nil {
-			bw.sessionTokenBucket.waitForAvailableToken()
+		if bw.sessionTokenBucket != nil && !bw.sessionTokenBucket.waitForAvailableToken() {
+			return
```
yuandrew, line 604: This return occurs after both `releaseActive` and `permit` have been acquired, but before the poll goroutine installs their cleanup defers. Please release the unused slot and call `releaseActive()` before returning.

internal/internal_worker_base_test.go
```diff
@@ -1199,3 +1199,60 @@ func (s *ScalableTaskPollerSuite) TestNewScalableTaskPollerAllTypes() {
...
+		taskProcessor:      noopTaskProcessor{},
+		workerType:         "SessionTokenStopTest",
+		logger:             ilog.NewNopLogger(),
+		stopTimeout:        5 * time.Second,
+		metricsHandler:     metrics.NopHandler,
+		sessionTokenBucket: bucket,
+	})
+
+	bw.Start()
+	<-reserved // the poller holds a permit and is committed to the token wait
```
yuandrew: Receiving `reserved` does not prove that the permit reached `runPoller` or that the poller entered `waitForAvailableToken`, because the signal is sent before `ReserveSlot` returns and before the unbuffered permit handoff. `Stop()` can therefore take the existing stop-channel path and let this test pass without exercising the bug. Please synchronize on entry into the actual token wait; that would also be a good place to assert the permit is released.

yuandrew (commented):
Thanks for finding this bug, a couple of comments for missing cleanup

ridwanakf:
> Thanks for finding this bug, a couple of comments for missing cleanup

thanks, will take a look at the comment and resolved it

internal/internal_worker_base.go
ridwanakf: Good catch, thanks. Fixed in c31b0cb9.

I'd missed that [`pollTask`'s defer](https://github.com/temporalio/sdk-go/blob/e47efb228346419536937b3ca78bf07bb67e9e03/internal/internal_worker_base.go#L801-L805) is the thing that actually hands the permit back, so my early return was the one path where nobody owned it. It releases with `SlotReleaseReasonUnused` before returning now, same as [what `reserveSlotAsync` does](https://github.com/temporalio/sdk-go/blob/e47efb228346419536937b3ca78bf07bb67e9e03/internal/internal_worker_base.go#L645) when its handoff gets cancelled.

The test asserts the release and the reason now, and it fails if I pull just that one line back out.

internal/internal_worker_base.go
ridwanakf: Yep, same thing here. Fixed in c31b0cb9: permit released with `SlotReleaseReasonUnused`, then `releaseActive()`, before the return. Lines up with the two exits already in that loop, the [stop channel one](https://github.com/temporalio/sdk-go/blob/e47efb228346419536937b3ca78bf07bb67e9e03/internal/internal_worker_base.go#L587) and the [nil permit one](https://github.com/temporalio/sdk-go/blob/e47efb228346419536937b3ca78bf07bb67e9e03/internal/internal_worker_base.go#L592).

One thing I should flag: sessions can't actually reach this path today, they run the simple maximum poller behaviour rather than the autoscaling one (CMIIW). I changed it anyway so the two loops dont drift apart, but happy to drop it if you'd rather keep the diff to the path that's reachable.

internal/internal_worker_base_test.go
ridwanakf: You're right, and it's worse than the test just being unproven: it passes on unfixed `main`.

I stuck a 50ms sleep between the `reserved` signal and the permit handoff, standing in for the scheduler simply not getting to the `reserveSlotAsync` goroutine yet, and ran the old test against unfixed `main`. 10 runs, 10 passes. The poller took the `<-bw.stopCh` branch every time and never went near the token wait. My comment on that helper claimed the signal proved the poller had a permit and was committed to the wait, and it doesn't prove either one.

Rewritten in c31b0cb9 to sync on the wait itself. The bucket's `sync.Cond` now sits on a lock that reports the first time it gets taken, and `waitForAvailableToken` takes that lock as its first statement, so the signal comes from the poller's own goroutine, after the permit handoff, with no branch left between the signal and the park. `close()` needs that same lock, so a `Stop()` starting after the signal can't get ahead of the waiter: it blocks until `Cond.Wait` drops the lock, and by then the waiter is queued so the `Broadcast` is guaranteed to reach it.

So it's deterministic now rather than probabilistic. Unfixed `main`: fails 5/5, each run burning the whole 5.00s stop timeout. With the fix: 20/20 under `-race`.

I also added the permit assertion you suggested to the same test. If I remove only the `releaseSlot` line it fails on that assertion in 0.00s while the promptness one still passes, so they're not covering for each other.

ridwanakf:
Hi @yuandrew, I've resolved your comments and feedbacks under c31b0cb9. Had rebased to `main` branch as well.

All three are addressed: the permit gets released on the new stop path in both pollers, `releaseActive()` goes with it in the autoscaling loop, and the test syncs on entry into the token wait now instead of on slot reservation.

The test one is the one I'm glad you caught. The old version passed 10 out of 10 against unfixed `main` once I delayed the permit handoff, so it wasn't reliably exercising the bug at all. It fails 5 out of 5 there now, at the full stop timeout each run, and checks the permit release on the way out.

Please take another look! Thank you!

yuandrew (approved):

yuandrew:
Thanks for the contribution!

PR #2475, opened 2026-07-22
feat(opentelemetry): add support for OTEL v2

### LINKED ISSUE #2463: OpenTelemetry v2 plugin
Implement an OpenTelemetry v2 plugin for the Go SDK, matching the plugin conventions established in the Python and TypeScript SDKs. New opt-in package, not a replacement for the existing `contrib/opentelemetry` interceptor package.

Scope:
- Ships as a `temporal.SimplePlugin` passed to client/worker creation, rather than manually wiring up `NewTracingInterceptor`.
- Span IDs are deterministic across workflow replay.
- Span durations reflect real workflow time, not replay time.
- Users can add their own spans for application logic in workflow code, activity code, and client code, with correct duration and parenting.

Histories will be incompatible for replay with the v1 interceptor package, so this must live as a separate package.

Reference implementations:
- Python: https://github.com/temporalio/sdk-python/pull/1314
- TypeScript: https://github.com/temporalio/sdk-typescript/pull/2142

patbeqo:
## What was changed
Add an opt-in `contrib/opentelemetry-v2` package that provides an OpenTelemetry integration as a `temporal.SimplePlugin`, with a v2 tracing interceptor in `interceptor/tracing`.

Key changes:
- Ships as a plugin (`NewPlugin`/`PluginOptions`) passed to client/worker creation.
- Span IDs are deterministic across workflow replay, derived from the Workflow tracing state
- Deterministic Workflow spans are not re-emitted or re-timed on replay
- Users can create their own spans in Workflow code via `Tracer` and its `Start` method, correctly parented and replay-safe.

The first commit copies the v1 tracing interceptor and the `contrib/opentelemetry` metrics handler into the v2 locations verbatim (the handler under its v2 name, `metrics.go`), so the rest of the PR reads as a diff against v1.

## Why?
Shipping as an opt-in contrib/opentelemetry-v2 module (separate from v1) lets users adopt it deliberately rather than having their existing traces change underneath them.

## Checklist

1. Closes #2463

2. How was this tested:

Unit tests for:
- Metrics
- Tracers
- Logging
- Codec
- Plugin wiring

Integration tests assert the full span tree with the plugin on and with signal/query/update tracing disabled.

The API is marked `NOTE: Experimental`.

3. Any docs updates needed?
Added a README and CHANGELOG for `contrib/opentelemetry-v2` and a root CHANGELOG entry.

---

> [!NOTE]
> **Medium Risk**
> Large new surface area around distributed tracing and workflow replay semantics; v1/v2 incompatibility can break traces if mixed, but adoption is opt-in and v1 is unchanged.
> 
> **Overview**
> Adds an **opt-in** `contrib/opentelemetry-v2` module so OpenTelemetry tracing can be adopted without changing existing v1 traces. v1 and v2 are **not span-compatible** (different parenting and span kinds).
> 
> The new package wires tracing through `NewPlugin` / `PluginOptions` on the Temporal client. It includes `NewTracerProvider` and workflow `Tracer` for **replay-safe** user spans (stable IDs/times via `workflow.GetRandomStream`, no duplicate export on replay), optional metrics via the v1-style handler, and baggage/header propagation.
> 
> In the SDK, **`interceptor/tracing`** gains a reworked tracing interceptor (`tracing_interceptor_v2.go`) with pluggable `Tracer` / `WorkflowTracer` factories, inbound/outbound span directions for correct OTel span kinds, and coverage for client/worker paths including Nexus, updates, and continue-as-new.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91d38aa228d10a1cbde9c516c3237b5e02f9b9de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


test/integration_test.go
```diff
@@ -3538,6 +3574,240 @@ func (ts *IntegrationTestSuite) addOpenTelemetryChildren(
 	}
 }
 
+// TestOpenTelemetryV2QueryActive verifies query-handler tracing when the query
```
xumaple: I would strongly prefer testing in one bigger chunk to capture the entire workflow trace, instead of these small snippets which very easily let different niche span parenting bugs slip through

contrib/opentelemetry-v2/README.md
```diff
@@ -0,0 +1,150 @@
...
+The recommended setup is the plugin, which installs the tracing interceptor on
+the client and on workers created from that client:
+
+```go
+provider := temporalotel.NewTracerProvider(nil, sdktrace.WithBatcher(exporter))
+defer provider.Shutdown(ctx)
+
+plugin, err := temporalotel.NewPlugin(temporalotel.PluginOptions{
+	TracerOptions: temporalotel.TracerOptions{TracerProvider: provider},
+})
```
xumaple: can we tie these together? user should be required to use the temporal replay safe tracerprovider if they're using the plugin? this integration seems like it's providing quite a complicated API for the user

xumaple: I'd say try to use `SimplePlugin`

interceptor/tracing/idempotency.go
```diff
@@ -0,0 +1,54 @@
+package tracing
```
xumaple: I don't think this package needs to go into the main SDK? can just live in contrib/, any reason you didn't put it in contrib/?

contrib/opentelemetry-v2/id_generator.go
```diff
@@ -0,0 +1,153 @@
+package opentelemetry
```
xumaple: I don't think handrolling this is the right way to go, to be honest. But I did some digging with claude help and I don't know if there is anything better at the current moment. Might be worth a conversation with SDK folks to see if they can provide us something so we don't have to handroll this nonsense

contrib/opentelemetry-v2/id_generator.go
```diff
@@ -0,0 +1,153 @@
...
+//
+// A nil fallback defaults to an equivalent of the OpenTelemetry SDK's random
+// generator. Pass a fallback when another generator must produce the
+// non-deterministic IDs, e.g. xray.NewIDGenerator() when exporting to AWS
+// X-Ray.
+//
+// NOTE: Experimental
+func NewDeterministicIDGenerator(fallback sdktrace.IDGenerator) sdktrace.IDGenerator {
+	if fallback == nil {
+		fallback = &randomIDGenerator{}
```
xumaple: we definitely don't want to fallback to nondeterminism ever in temporal workflows

contrib/opentelemetry-v2/id_generator_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const testIDKey = "WorkflowInboundInterceptor:default:wf-id:run-id:1"
```
xumaple: I think the format you had going was `tracing.sdk:...` right?

internal/workflow_testsuite.go
```diff
@@ -396,6 +396,12 @@ func (e *TestWorkflowEnvironment) SetContinuedExecutionRunID(rid string) {
 	e.impl.setContinuedExecutionRunID(rid)
 }
 
+// SetWorkflowRunID sets the value that is returned from
+// GetInfo(ctx).WorkflowExecution.RunID
+func (e *TestWorkflowEnvironment) SetWorkflowRunID(runID string) {
```
xumaple: is this really needed?

contrib/opentelemetry-v2/tracer.go
```diff
@@ -0,0 +1,414 @@
...
+// themselves.
+//
+// A nil TracerOptions.TracerProvider resolves to the OpenTelemetry global
+// provider, so pass the same provider you gave the interceptor/plugin (or
+// register it with otel.SetTracerProvider); otherwise spans are created but
+// silently never reach your exporter.
+//
+// NOTE: Experimental
+func NewWorkflowTracer(ctx workflow.Context, options TracerOptions) WorkflowTracer {
+	return newWorkflowTracer(ctx, options)
```
xumaple: I believe in NewWorkflowTracer, since your idempotency key is calculated purely off of workflow context, when you have multiple instances of your workflow tracer in the same workflow then the idempotency keys collide. I'm surprised your tests don't exercise this path ever?

contrib/opentelemetry-v2/tracer.go
```diff
@@ -0,0 +1,414 @@
...
+		HeaderKey:               t.options.HeaderKey,
+		DisableSignalTracing:    t.options.DisableSignalTracing,
+		DisableQueryTracing:     t.options.DisableQueryTracing,
+		DisableUpdateTracing:    t.options.DisableUpdateTracing,
+		AllowInvalidParentSpans: t.options.AllowInvalidParentSpans,
+	}
+}
+
+func (t *tracer) UnmarshalSpan(m map[string]string) (tracing.TracerSpanRef, error) {
+	if _, ok := m["traceparent"]; !ok {
```
xumaple: claude found this bug: 

```
traceparent hardcoding breaks non-W3C propagators. example:

provider := temporalotel.NewTracerProvider(nil, sdktrace.WithBatcher(zipkinExporter))
plugin, err := temporalotel.NewPlugin(temporalotel.PluginOptions{
      TracerOptions: temporalotel.TracerOptions{
              TracerProvider:    provider,
              TextMapPropagator: b3.New(), // injects "b3" header, never "traceparent"
      },
})

in this case, m["traceparent"] finds nothing.
```

contrib/opentelemetry-v2/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Deterministic span IDs
+
+Use `NewTracerProvider` (or attach `NewDeterministicIDGenerator` to your own
+provider) so workflow spans keep stable IDs across retries and replays. Pass a
+fallback ID generator if you need a specific scheme (for example
+`xray.NewIDGenerator()` for AWS X-Ray), or `nil` for OpenTelemetry's default
+random generator.
+
+The plugin does not enable this for you — pass a provider built this way.
```
xumaple: this is not a good default to have. we should probably not even have it as an option. deterministic only

contrib/opentelemetry-v2/README.md
```diff
@@ -0,0 +1,150 @@
...
+  (`"tracing.sdk"`); the tracing interceptor reserves it for its own
+  SDK-managed spans.
+- A nil `TracerProvider` resolves to the OpenTelemetry global provider, so
+  pass the same provider you gave the interceptor/plugin (or register it with
+  `otel.SetTracerProvider`). If that global provider wasn't built with
+  `NewTracerProvider`/`NewDeterministicIDGenerator`, workflow spans get random
+  IDs instead of stable ones, breaking idempotency across retries and
+  replays.
+
+### Manual interceptor wiring
```
xumaple: Probably don't need to mention this in the readme at all, as we really just want people to use the plugin.

contrib/opentelemetry-v2/go.mod
```diff
@@ -0,0 +1,48 @@
+module go.temporal.io/sdk/contrib/opentelemetry-v2
```
xumaple, line 1:
```suggestion
module opentelemetry.OtelPlugin
```

This is the standard we go with now. `company.Plugin`

xumaple (commented):
I see some stuff in the PR description talking about "supporting both major version of OTel" - maybe debt from copying over from the TS integration? also "no longer risks non-determinism on replay" is suspicious, I don't think that was an issue in Go otel v1 integration?

It's a bit hard to trace how exactly the span parenting is being done without a more comprehensive test to exercise all the paths, so look forward to seeing that.

interceptor/tracing/idempotency.go
patbeqo: I've moved that logic to `contrib/opentelemetry-v2/provider.go`

internal/workflow_testsuite.go
patbeqo: Removed.

contrib/opentelemetry-v2/tracer.go
patbeqo: Good catch. I've refactored so the id stream is stored in `workflow.Context` and can be shared across multiple tracers, and spans created from those tracers, within the same workflow.

contrib/opentelemetry-v2/README.md
patbeqo: Updated.

contrib/opentelemetry-v2/README.md
patbeqo: Updated.

contrib/opentelemetry-v2/id_generator.go
patbeqo: Query handlers and Update validators are not re-executed during replay. Therefore, spans created within them must not consume IDs from the deterministic ID stream, as doing so would cause the stream’s state to diverge during replay.

For the user-facing API, I’ve exposed `StartUnsequenced` for these cases. However, this requires users to select the correct method when creating spans inside Query handlers and Update validators, which feels error-prone. Ideally, `Start` would detect these contexts and avoid advancing the deterministic ID stream automatically.

The Go SDK tracks these execution states internally, but I could not find a public API for detecting them from a `workflow.Context`. I’m open to suggestions.

contrib/opentelemetry-v2/id_generator_test.go
patbeqo: Refactored these tests away so this no longer applies.

contrib/opentelemetry-v2/id_generator.go
patbeqo: Moved this into `provider.go` and cut some of the hand-rolled stuff (dropped the copied random-ID generator, now just uuid). Still some hand-rolling left for the deterministic key derivation. 

Agreed it's not ideal, happy to swap it out for something better.

contrib/opentelemetry-v2/tracer.go
patbeqo: Fixed and added a test case.

patbeqo:
> I see some stuff in the PR description talking about "supporting both major version of OTel" - maybe debt from copying over from the TS integration? also "no longer risks non-determinism on replay" is suspicious, I don't think that was an issue in Go otel v1 integration?
> 
> It's a bit hard to trace how exactly the span parenting is being done without a more comprehensive test to exercise all the paths, so look forward to seeing that.

Updated the PR description to better reflect the changes. Also added more comprehensive tests.

contrib/opentelemetry-v2/id_generator.go
patbeqo: Dropped `StartUnsequenced`. The inbound interceptor now attaches workflow tracing state and marks query / update-validation as unsequenced, so `Start` skips the deterministic ID stream automatically in those contexts.

contrib/opentelemetry-v2/id_generator.go
patbeqo: Resolving in https://github.com/temporalio/sdk-go/issues/2517

contrib/opentelemetry-v2/codec.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+
+	"go.opentelemetry.io/otel/baggage"
+	"go.opentelemetry.io/otel/trace"
+
+	"go.temporal.io/sdk/interceptor/tracing"
+	"go.temporal.io/sdk/log"
+)
+
+type textMapCarrier map[string]string
```
brianstrauch, line 12: Can we use https://pkg.go.dev/maps#Keys here?

contrib/opentelemetry-v2/suite_test.go
```diff
@@ -0,0 +1,150 @@
+package opentelemetry
```
brianstrauch, line 1: Minor: It's more correct to name this file something ending with _test.go so it doesn't get pulled into the built package

contrib/opentelemetry-v2/codec.go
```diff
@@ -0,0 +1,78 @@
...
+	if !c.options.DisableBaggage && tSpan.Baggage.Len() > 0 {
+		ctx = baggage.ContextWithBaggage(ctx, tSpan.Baggage)
+	}
+
+	data := textMapCarrier{}
+	c.options.TextMapPropagator.Inject(ctx, data)
+	return data, nil
+}
+
+func (c *spanCodec) UnmarshalSpan(m map[string]string) (tracing.TracerSpanRef, error) {
```
brianstrauch, line 43: Is it intentional that we're not returning any errors from this function? v1 returned: fmt.Errorf("failed extracting OpenTelemetry span from map")

patbeqo: Yes, intentional. OpenTelemetry propagation doesn't distinguish absent trace data from malformed trace data, and baggage can be valid without a span context. v1 discarded that baggage, whereas v2 follows OpenTelemetry’s behavior by preserving it and returning an empty span context.

[Baggage without a span unit test](https://github.com/temporalio/sdk-go/pull/2475/changes#diff-6615c8866e328cfc28333bef7a30263bdab5b4396465f57900e112afe5232da7R153)

brianstrauch (approved):

contrib/opentelemetry-v2/CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
...
+Loosely based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+Add user-facing changes below under the appropriate heading (create the heading
+if it does not yet exist): Added, Changed, Deprecated, Breaking Changes, Fixed,
+or Security.
+-->
+
+# Changelog
+
+## [Unreleased]
```
yuandrew, line 12: we can technically just make this 0.1.0 immediately, since we're gonna release this right after merging

contrib/opentelemetry-v2/README.md
```diff
@@ -0,0 +1,66 @@
+# OpenTelemetry v2 integration for the Temporal Go SDK
+
+Package `go.temporal.io/sdk/contrib/opentelemetry-v2` provides OpenTelemetry
+tracing and metrics for Temporal. Experimental: APIs may change between releases.
+
+For the stable integration, use
```
yuandrew: What do you mean by "stable integration" here? otel v1 is also still "Experimental", since it's still not 1.0.0 version

contrib/opentelemetry-v2/README.md
```diff
@@ -0,0 +1,66 @@
+# OpenTelemetry v2 integration for the Temporal Go SDK
+
+Package `go.temporal.io/sdk/contrib/opentelemetry-v2` provides OpenTelemetry
+tracing and metrics for Temporal. Experimental: APIs may change between releases.
```
yuandrew, line 4: Don't know if we need to explicitly call out "Experimental" here, it's typical that Go packages feel more at liberty to make breaking changes prior to their 1.0.0 release. So as long as we keep the major version 0, i think most Go users will consider it "Experimental"

contrib/opentelemetry-v2/options.go
```diff
@@ -0,0 +1,33 @@
+package opentelemetry
```
xumaple, line 1: In python, we have this:

```
provider = get_tracer_provider()
  if not isinstance(provider, ReplaySafeTracerProvider):
      raise ValueError("When using OpenTelemetryPlugin, the global trace provider must be a ReplaySafeTracerProvider...")
```

Is it possible to have similar here where we fail fast when tracer is not being properly used?

contrib/opentelemetry-v2/interceptor_tracer_test.go
```diff
@@ -0,0 +1,380 @@
...
+		s.Require().Equal(fullTree, s.formatSpanTree(spans))
+		s.requireUniqueSpanIDs(spans)
+		s.requireUpdateIDs(spans)
+		s.requireContinueAsNewErrorNotRecorded(spans)
+	})
+
+	s.Run("all-disabled", func() {
+		spans := s.runScenario(PluginOptions{
+			TracerOptions: tracing.TracerOptions{
+				DisableSignalTracing: true,
```
xumaple: this is not the way that our other integrations limit which traces are emitted, we simply have `addTemporalSpans: bool`. Curious why you chose to do this?

contrib/opentelemetry-v2/interceptor_tracer.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// interceptorWorkflowSpan suppresses Finish calls during replay.
+type interceptorWorkflowSpan struct {
+	*tracerSpan
+	ctx workflow.Context
+}
+
+func (s *interceptorWorkflowSpan) Finish(opts *tracing.TracerFinishSpanOptions) {
+	if workflow.IsReplaying(s.ctx) {
```
xumaple: I think you need to not call Finish if the workflow task failed here?

xumaple (changes requested):
Overall pretty good but a few small things

contrib/opentelemetry-v2/README.md
yuandrew: I guess the other SDKs mention this

interceptor/tracing/tracing_interceptor_v2.go
```diff
@@ -0,0 +1,963 @@
...
+	SpanFromContext(workflow.Context) TracerSpan
+	// ContextWithSpan creates a workflow context with the given span set.
+	ContextWithSpan(workflow.Context, TracerSpan) workflow.Context
+
+	// CreateSpan starts and returns a span with the given options.
+	CreateSpan(workflow.Context, *TracerStartSpanOptions) TracerSpan
+}
+
+// BaseTracer is a default implementation of Tracer meant for embedding.
+type BaseTracer struct{}
```
yuandrew, line 82: Do we want to assert this always implements `Tracer`? `var _ Tracer = BaseTracer{}`

interceptor/tracing/tracing_interceptor_v2.go
```diff
@@ -0,0 +1,963 @@
...
+}
+
+// TracerFinishSpanOptions are options for TracerSpan.Finish.
+type TracerFinishSpanOptions struct {
+	// Error is present if there was an error in the code traced by this specific
+	// span.
+	Error error
+}
+
+type TracerFactory func() Tracer
```
yuandrew: Can we document when these factories are called, whether they need to be thread-safe, whether they must return a fresh instance? from what I can tell, the client tracer is created once and resued across concurrent client calls

interceptor/tracing/tracing_interceptor_v2.go
```diff
@@ -0,0 +1,963 @@
...
+			workflowIDTagKey: info.WorkflowExecution.ID,
+			runIDTagKey:      info.WorkflowExecution.RunID,
+			activityIDTagKey: info.ActivityID,
+		},
+		Time: info.StartedTime,
+	}, t.root.headerReader(t.tracer, ctx))
+	if err != nil {
+		return nil, err
+	}
+	defer endSpan(&err)
```
yuandrew: `activity.ErrResultPending` is a successfull async handoff, but `Finish` records the activity span as an error

interceptor/tracing/tracing_interceptor_v2.go
```diff
@@ -0,0 +1,963 @@
...
+func (t *tracingWorkerInterceptor) InterceptNexusOperation(
+	ctx context.Context,
+	next interceptor.NexusOperationInboundInterceptor,
+) interceptor.NexusOperationInboundInterceptor {
+	i := &tracingNexusOperationInboundInterceptor{root: t.root, tracer: t.root.newTracer()}
+	i.Next = next
+	return i
+}
+
+type tracingClientOutboundInterceptor struct {
```
yuandrew, line 230: should we handle client-started Nexus operations here too? right now this falls through to the base interceptor, so it doesn't create an outbound span or add anything to `NexusHeader`. the handler then starts a separate trace.

interceptor/tracing/tracing_interceptor_v2.go
```diff
@@ -0,0 +1,963 @@
...
+	var spanErr error
+	defer endSpan(&spanErr)
+	err = t.Next.NewContinueAsNewError(ctx, wfn, args...)
+	if !isContinueAsNewError(err) {
+		spanErr = err
+	}
+	return err
+}
+
+type tracingNexusOperationInboundInterceptor struct {
```
yuandrew, line 751: Should we wrap the Nexus outbound interceptor here too? Otherwise `temporalnexus.GetLogger(ctx)` skips `Tracer.GetLogger`, so Nexus handler logs don't ge tthe trace/span IDs

contrib/opentelemetry-v2/codec.go
```diff
@@ -0,0 +1,78 @@
...
+	"go.opentelemetry.io/otel/baggage"
+	"go.opentelemetry.io/otel/trace"
+
+	"go.temporal.io/sdk/interceptor/tracing"
+	"go.temporal.io/sdk/log"
+)
+
+type textMapCarrier map[string]string
+
+func (t textMapCarrier) Get(key string) string        { return t[key] }
```
yuandrew: should this lookup be case insensitive? Nexus lowercases keys in Header.Set, so a propagator with mixed-case fields can inject successfully but fail to extract in the handler

yuandrew:
bugbot run

contrib/opentelemetry-v2/options.go
patbeqo: Good call out. I've added a similar check and renamed to match Python.

contrib/opentelemetry-v2/interceptor_tracer_test.go
patbeqo: The plugin itself is the opt-in boundary. These flags selectively disable signal, query, and update tracing, preserving the behavior from v1. https://github.com/temporalio/sdk-go/blob/023b14700f1a323d47fb6d2eeb56ce6ef5a92ddb/interceptor/tracing_interceptor.go#L90-L97

contrib/opentelemetry-v2/interceptor_tracer.go
patbeqo: A WFT can fail after the span has already ended.

contrib/opentelemetry-v2/interceptor_tracer_test.go
patbeqo: I've added the `AddTemporalSpans` option with tests.

contrib/opentelemetry-v2/README.md
patbeqo: Removed.

contrib/opentelemetry-v2/README.md
patbeqo: Yes I've kept it to match the other SDKs. Open to removing it though.

contrib/opentelemetry-v2/interceptor_tracer_test.go
patbeqo: I'm open to removing the v1 options.

interceptor/tracing/tracing_interceptor_v2.go
patbeqo: Not for `BaseTracer`, but I've added asserts to other structs that implement the shared interfaces.

contrib/opentelemetry-v2/codec.go
patbeqo: Nice catch. Updated `Get` to be case insensitive.

interceptor/tracing/tracing_interceptor_v2.go
patbeqo: The factory implementation was from a stale change. With `GetRandomStreamI` it is no longer needed. The `Tracer` instance can be passed in directly. I've added a comment mentioning the concurrency requirement.

The underlying Otel tracer is safe for concurrent use https://opentelemetry.io/docs/specs/otel/trace/api/#concurrency-requirements

interceptor/tracing/tracing_interceptor_v2.go
patbeqo: Nice catch, both of these are handled now. I went through the Python and TypeScript v2
interceptors and added every method they trace. There are still a few small inconsistencies
across the three SDKs in span names and in which methods each one traces. I plan to reconcile
those in future PRs before this goes GA.

interceptor/tracing/tracing_interceptor_v2.go
patbeqo: See my other reply.

interceptor/tracing/tracing_interceptor_v2.go
patbeqo: Updated and added a test.

contrib/opentelemetry-v2/README.md
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"go.opentelemetry.io/otel"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.temporal.io/sdk/client"
+	temporalotel "go.temporal.io/sdk/contrib/opentelemetry-v2"
+	"go.temporal.io/sdk/worker"
+)
+
+// Create a replay-safe tracer provider and set it as the OpenTelemetry global
+provider := temporalotel.NewReplaySafeTracerProvider(sdktrace.WithBatcher(exporter))
```
yuandrew, line 34: can we show shutting down the provider here? this uses `WithBatcher`, so without `Shutdown` or `ForceFlush`, users who copy the example can lose buffered spans on exit.

contrib/opentelemetry-v2/provider.go
```diff
@@ -0,0 +1,88 @@
...
+// io.Reader attached to the span-start context (workflow.GetRandom for
+// sequenced workflow spans, or crypto/rand.Reader otherwise). Any ID generator
+// supplied in opts is intentionally overridden. Install the result with
+// otel.SetTracerProvider so NewPlugin interceptors and Tracer share it.
+// Direct otel.Tracer calls are not workflow-safe; workflow code must use
+// Tracer. The caller owns the provider and must shut it down after its
+// clients and workers stop.
+//
+// NOTE: Experimental
+func NewReplaySafeTracerProvider(opts ...sdktrace.TracerProviderOption) *replaySafeTracerProvider {
```
yuandrew: This returns an unexported type, should we export `replaySafeTracerProvider` so users can name it in a struct or return it from a setup helper?

interceptor/tracing/tracing_interceptor_v2.go
```diff
@@ -0,0 +1,1108 @@
...
+	}, t.root.headerWriter(t.root.tracer, ctx), false)
+	if err != nil {
+		return nil, err
+	}
+	defer endSpan(&err)
+
+	return t.Next.ExecuteActivity(ctx, in)
+}
+
+func (t *tracingClientOutboundInterceptor) ExecuteNexusOperation(
```
yuandrew, line 367: can we add a client-started Nexus operation to the integration test too? this fixes the earlier issue, but the current Nexus tests only exercise workflow.NewNexusClient, so this path can regress without failing.

contrib/opentelemetry-v2/spans.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// workflowSpan suppresses End calls during replay.
+type workflowSpan struct {
+	*tracerSpan
+	ctx workflow.Context
+}
+
+func (s *workflowSpan) End(options ...trace.SpanEndOption) {
+	if workflow.IsReplaying(s.ctx) {
```
yuandrew, line 59: what happens to `SpanProcessor` lifecycle here? `Start` has already called the sampler and `OnStart`, but this skips the underlying `End` during replay, so replay-only spans never get a matching `OnEnd`. could we cover this with a counting processor?

patbeqo: This is intentional. Replay-only spans skip `End`, and therefore `OnEnd`, to avoid duplicate exports. Existing replay tests cover this behavior. This matches Python; TypeScript prevents duplicate replay exports through its replay-suppressed sink instead.

contrib/opentelemetry-v2/interceptor_tracer_test.go
xumaple: I don't think we want the v1 options - I'd rather us unify all our OTEL integrations across languages.

interceptor/tracing/tracing_interceptor_v2.go
```diff
@@ -0,0 +1,1108 @@
...
+	if createSpan {
+		options.Direction = SpanDirectionOutbound
+		options.Parent = curr
+		curr = t.CreateSpan(ctx, options)
+		ctx = t.ContextWithSpan(ctx, curr)
+	}
+
+	if headerWriter != nil {
+		if err := headerWriter(curr); err != nil {
+			return ctx, nil, err
```
yuandrew: should we finish the span before returning this error? if header encoding fails after CreateSpan succeeds, this returns without ever calling Finish. the workflow version below has the same issue.

yuandrew (approved):
One comment I meant to leave in the last round of review, but LGTM otherwise!

interceptor/tracing/tracing_interceptor_v2.go
patbeqo: Yes, nice catch. Fixed and added test.

xumaple (approved):
LGTM

PR #2478, opened 2026-07-22
googleadk: add NewPlugin worker-plugin wiring (v0.2.0)

DABH:
## What

Adds `googleadk.NewPlugin`, packaging the worker side of the integration as a worker plugin built on `temporal.SimplePlugin`, and makes it the standard wiring:

```go
adkPlugin, err := googleadk.NewPlugin(googleadk.Config{Models: ..., MCPToolsets: ...})
w := worker.New(c, taskQueue, worker.Options{Plugins: []worker.Plugin{adkPlugin}})
```

- **Worker start** (`RunContextBefore`): registers the `InvokeModel` / `ListMcpTools` / `CallMcpTool` Activities.
- **Worker stop** (`RunContextAfter`): calls `Activities.Close()`, so cached MCP toolsets (sessions / CommandTransport subprocesses) shut down with the worker automatically instead of relying on a user-remembered `defer acts.Close()`.
- **Replay-safe sharing**: run-context hooks also fire once per workflow replay; replays are tracked by instance key (counted — concurrent replays share a key) so a plugin shared with a replayer never closes the worker's toolsets, which would be terminal.
- `NewActivities` + `Register` remain the path for the test environments (which construct no real worker, so plugins do not run there) and manual wiring; `Register` now delegates to a shared helper also used by the plugin.

This mirrors sdk-python's `GoogleAdkPlugin`, which packages the same integration as a `SimplePlugin`.

## Testing plan

- `TestNewPluginRejectsInvalidConfig` — config validation surfaces at construction.
- `TestPluginRegistersActivitiesOnWorkerStart` — exactly the three Activities register through the plugin registry.
- `TestPluginClosesMCPToolsetsOnWorkerStop` — end-to-end through the test env: MCP toolset constructed and cached by a workflow run, closed exactly once on `StopWorker`, second stop is a no-op.
- `TestPluginReplayDoesNotCloseToolsets` — a replay pass neither closes nor poisons the shared Activities; verified the test fails when the replay guard is stubbed to an unconditional Close.
- Full package suite green including `-race`; gofmt/vet clean.

## Release

`CHANGELOG.md` carries this under `## [0.2.0] - 2026-07-22` for today's release cut.

### DIFF
--- a/contrib/googleadk/README.md
+++ b/contrib/googleadk/README.md
@@ -20,7 +20,8 @@ Tools run **in-workflow by default** — the idiomatic Temporal model: your work
 is deterministic, and anything that touches the network, clock, or disk goes
 through an Activity. Opt a tool into an Activity with `googleadk.ActivityAsTool`,
 or use `googleadk.NewMCPToolset` for MCP. The real model and any activity/MCP tool
-handlers live worker-side in the registry built by `googleadk.NewActivities(...)`.
+handlers live worker-side in the registry declared by `googleadk.Config` and
+wired onto the worker by `googleadk.NewPlugin(...)`.
 
 ## Add to your project
 
@@ -134,12 +135,11 @@ func main() {
 	}
 	defer c.Close()
 
-	w := worker.New(c, taskQueue, worker.Options{})
-	w.RegisterWorkflow(AgentWorkflow)
-
-	// Worker-side registry: the real model lives here. API keys are captured in
-	// the factory closure and never cross the Activity boundary.
-	acts, err := googleadk.NewActivities(googleadk.Config{
+	// Worker-side registry, wired as a worker plugin: the real model lives here.
+	// API keys are captured in the factory closure and never cross the Activity
+	// boundary. The plugin registers the Activities at worker start and closes
+	// cached MCP toolsets at worker stop.
+	adkPlugin, err := googleadk.NewPlugin(googleadk.Config{
 		Models: map[string]googleadk.ModelFactory{
 			"gemini-2.0-flash": func(ctx context.Context, name string) (model.LLM, error) {
 				// nil config reads GEMINI_API_KEY / GOOGLE_API_KEY from the env, worker-side.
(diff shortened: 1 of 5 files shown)


contrib/googleadk/README.md
```diff
@@ -201,7 +204,8 @@ the model SDK's own retries (see below), or override the fallbacks.
   a transient failure twice over. Let Temporal own retries.
 - **Test without a live LLM.** `testing.go` ships `FakeModel`, `FakeMCPServer`, and
   `TextResponse` / `FunctionCallResponse` so you can unit-test workflows with no
-  network.
+  network. The test environments do not run plugins; register the Activities
+  directly with `NewActivities` + `Register` (as this repo's own tests do).
```
brianstrauch, line 208: Should the test environments run plugins? We may also want to deprecate NewActivities if it provides no benefit over NewPlugin.

DABH:
CI note: the two failures (`integ-test` / `integ-test-cache`, windows-stable) are pre-existing flakes, not from this change:

- Both failing tests live in the root module's `test/` suite, which doesn't link `contrib/googleadk` (separate module) — this PR's code isn't in those binaries.
- The dominant failure, `TestNexusUpdateWorkflowOperation/Verify_concurrent_Updates_with_same_UpdateID_are_idempotent_and_finish`, fails on `main` across four other platform/version combos in runs without this change (details + run links in #2479; it was added in #2417 and has flaked since).
- `main` hasn't had a fully green CI run since ~07-17, with rotating integ-test flakes (`TestResourceBasedSlotSupplierManyActs`, `TestAsyncOperationFromWorkflow_CancellationTypes/Abandon`, …).

Re-ran the failed jobs.

contrib/googleadk/README.md
DABH: Re: test envs: Ideally yes; filed #2480 proposing it. Today they can't: the mock test environments never construct a real worker, so plugin hooks (worker-lifecycle concepts) have no anchor point. sdk-python doesn't have this gap because its test env runs a real worker, which is exactly why GoogleAdkPlugin composes into Python tests while Go needs the manual path.

Similarly, re: NewActivities: Blocked on #2480: the test environments are currently why the manual path exists (this package's own tests register through it), plus it's the escape hatch for Close error handling and multi-worker sharing. Once #2480 lands, deprecation becomes viable.

brianstrauch (approved):

PR #2485, opened 2026-07-23
fix(nexus): clear operation callbacks on 'abandon' cancellation.

### LINKED ISSUE #2495: [Bug] nexus start/complete callbacks not cleared on abandon
nexus start/complete callbacks not cleared on cancel with abandon

cconstable:
Was seeing a flakey test (`TestAsyncOperationFromWorkflow_CancellationTypes/Abandon`) and traced it back to a bug.

## What was changed

- Cancelling with`NexusOperationCancellationTypeAbandon` now correctly clears nexus callbacks.
- Added a small helper for getting the nexus operation command.

## Why?

When you cancel a workflow running a nexus operation with `NexusOperationCancellationTypeAbandon`, it operations futures directly but doesn't clear the `startedCallback` or `completedCallback`. Then, when a NexusOperationStarted (or completed) event is received, the stale callbacks fires and attempts to set an already resolved future which panics, which WFT fails.

Fixes #2495

## Checklist
- Added tests to verify that we clear callbacks.

### DIFF
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -1215,12 +1215,16 @@ func (h *commandsHelper) handleNexusOperationScheduled(event *historypb.HistoryE
 	command.handleInitiatedEvent()
 }
 
+func (h *commandsHelper) getNexusOperationCommand(seq int64) commandStateMachine {
+	return h.getCommand(makeCommandID(commandTypeNexusOperation, strconv.FormatInt(seq, 10)))
+}
+
 func (h *commandsHelper) handleNexusOperationStarted(scheduledEventID int64) commandStateMachine {
 	seq, ok := h.scheduledEventIDToNexusSeq[scheduledEventID]
 	if !ok {
 		panicIllegalState(fmt.Sprintf("[TMPRL1100] unable to find nexus operation state machine for event ID: %v", scheduledEventID))
 	}
-	command := h.getCommand(makeCommandID(commandTypeNexusOperation, strconv.FormatInt(seq, 10)))
+	command := h.getNexusOperationCommand(seq)
 	command.handleStartedEvent()
 	return command
 }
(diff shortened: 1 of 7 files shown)


internal/internal_command_state_machine.go
```diff
@@ -1215,12 +1215,16 @@ func (h *commandsHelper) handleNexusOperationScheduled(event *historypb.HistoryE
 	command.handleInitiatedEvent()
 }
 
+func (h *commandsHelper) getNexusOperationCommand(seq int64) commandStateMachine {
+	return h.getCommand(makeCommandID(commandTypeNexusOperation, strconv.FormatInt(seq, 10)))
```
cconstable, line 1219: added a small helper since this manual formatting was repeated in a number of places.

internal/workflow.go
```diff
@@ -3188,6 +3202,7 @@ func (wc *workflowEnvironmentInterceptor) ExecuteNexusOperation(ctx Context, inp
 					if !executionFuture.IsReady() {
 						executionSettable.Set(nil, ErrCanceled)
 					}
+					wc.env.AbandonNexusOperation(seq)
```
cconstable, line 3191: the fix. clears out the callbacks so we don't try to resolve an already resolved future.

dplyukhin (approved):

Quinn-With-Two-Ns:
We have the same problem for Child Workflows that already does handle this, is there a reason you solved this in a different way then Child Workflows? 

https://github.com/temporalio/sdk-go/blob/9a42facc9a2d8f8ba3f0c231fe81364b203eab6b/internal/internal_event_handlers.go#L1846

cconstable:
@Quinn-With-Two-Ns I matched the existing nexus pattern of nilling out the callbacks:

```go
func (wc *workflowEnvironmentImpl) RequestCancelNexusOperation(seq int64) {
    ...
    data.startedCallback = nil
}
```

The `handled` flag in child workflows looks like it means we hit a terminal state and we can skip any extra completions. I don't think this maps to nexus directly since nexus operations can complete without a started event. So I think the options are:

1. follow the existing nexus pattern of nilling out the callbacks (what is in the PR)
2. add a `handled` flag and some new logic to all the nexus lifecycle handlers to nil out the started callbacks after firing it (seems like too much scope for this PR) 
3. add a new `abandoned` flag and use it for this case to skip completion callbacks but now this nexus lifecycle handler will diverge from what the others are doing

I actually prefer the idea of having flags/enums for the state here but I feel like that should be it's own PR and would probably need more discussion. I could open another to make that change.

PR #2487, opened 2026-07-24
fix spans' durations underflowing

faucct:
## What was changed
Started using Worker's clock not only for Spans' finishes, but also for starts.

## Why?
I've noticed that using the Temporal Server clock makes the durations underflow sometime.

## Checklist

2. How was this tested:
Unit testing

3. Any docs updates needed?
Not really

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ to docs, or any other relevant information.
 ### Fixed
 
 - Allow query results to use external storage before payload size enforcement.
+- Use the worker clock for `RunWorkflow` and `RunActivity` tracing span start times to avoid negative
+  durations when the server clock is ahead of the worker clock.
 - Correct schedule catch-up window documentation to state that an unset value is omitted and the
   server applies its one-year default.
 - Resource-based tuner: `ReserveSlot` now honors context cancellation while the resource controller is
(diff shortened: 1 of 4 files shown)


yuandrew:
Thanks for the contribution, is this related to a problem you're hitting? TracerStartSpanOptions.Time is public and currently documented to use WorkflowStartTime and StartedTime, so changing it affects both built-in and custom tracer implementations. I’d like to understand the scope before changing this behavior for downstream users.

faucct:
Yes, I’ve noticed it in our installation of temporal.

yuandrew:
We generally don't want to change this behavior without tying it to a concrete user-visible problem, so it would help to have the tracer being used and an example of the invalid timestamps.

Even so, we can't take the current solution as written. `RunWorkflow` is a logical span for the entire Workflow Execution, and it can be recreated when the workflow is replayed after cache eviction or on another worker. `WorkflowStartTime` is stable across those replays, and the deterministic `IdempotencyKey` allows tracer implementations to treat them as the same span. Replacing that with `time.Now()` means the same logical span can receive a different start time on every replay, potentially after child spans have already ocurred, and its duration no longer represents the workflow's lifetime. The added test only exercises a single-task workflow, so it does not catch this.

We'd need a solution that addresses the clock-skew problem while preserving a replay-stable `RunWorkflow` start time. `RunActivity` may also need to be considered separately rather than changing both operations together.

Please start with opening a GitHub issue to track the user-visible problem, and we can iterate on a solution.

Issue #2064, opened 2025-10-02
Allow heartbeats after Activity cancellation


ntaylorkhan:
**Is your feature request related to a problem? Please describe.**
You cannot ignore cancellation on Activities in the Go SDK. Other SDK's do not have this issue and you can already do this in Workflow via `worklow.GetDisconnectedContext`

**Describe the solution you'd like**
Allow heartbeats after cancellation.

**Describe alternatives you've considered**
None.

**Additional context**
n/a

cretz:
Note, this is similar to #1574 but this isn't even a new context, this is a desire to be able to heartbeat even if server has requested cancellation, which we should support. (cancellation is a request, it does not have to be honored)

PR #2492, opened 2026-07-24
Use last non-canceled context for batched heartbeats

### LINKED ISSUE #1574: Batched heartbeat got canceled due to caller context cancellation
## Expected Behavior
Say I have the following activity code
```
func (a *Activities) MyActivity(ctx) (out, error)(
  g, gctx := errgroup.WithContext(ctx)
  g.SetLimit(4)
  for _, batch := range batches {
    batch := batch
    g.Go(func() error {
      activity.RecordHeartbeat(gctx)
      // Goroutine logic
    })
  }
  if err := g.Wait(); err != nil {
    return nil, err
  }

  for _, item := range array {
    activity.RecordHeartbeat(ctx)
    // Additional logic
  }
}
```
I expect heartbeat to get sent out to the server and my activity to NOT time out

## Actual Behavior
Activity timed out with `activity Heartbeat timeout`

This is because `g.Wait()` cancels `gctx` but `gctx` is still referenced in SDK `Heartbeat` function's goroutine https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_handlers.go#L2031. Due to the batch heartbeat logic, `gctx` was used after `g.Wait()` had finished in my code and caused a `context canceled` error when making the call to the server

I could change `activity.RecordHeartbeat(gctx)` to `activity.RecordHeartbeat(ctx)` but wonder if SDK can avoid this pitfall by using context.Background()

  1.
  1.
  1.

## Specifications

  - Version: go.temporal.io/sdk v1.27.0
  - Platform: Linux

imkp1:
## What was changed

`temporalInvoker` now tracks the most recent context passed to `Heartbeat` that
was not already canceled, and the batching goroutine uses that context to flush
buffered heartbeat details instead of the context that opened the batch window.

Also documents the cancellation behavior on `activity.RecordHeartbeat` and
`internal.RecordActivityHeartbeat`.

## Why?

Heartbeats are throttled. The first call in a window sends an RPC and starts a
timer; later calls only record details. When the timer fires, the goroutine sent
the buffered details using the context captured when the window opened. If the
caller had canceled that context by then, the send failed with `context
canceled` and the activity timed out with `activity Heartbeat timeout` even
though it was alive and heartbeating with a live context.

The reported case passes an `errgroup` context to `RecordHeartbeat` from the
group's goroutines. `g.Wait()` cancels it, but the activity keeps running and
keeps heartbeating with its own context. The buffered send still used the dead
`errgroup` context.

This follows the direction in the issue thread: clarify the documentation, and
use the last non-canceled context when `RecordHeartbeat` is called with more
than one.

### Design notes

- The new `lastHeartbeatCtx` field is guarded by the struct's existing embedded
  mutex, in the same critical section that already reads `lastDetailsToReport`.
  No new lock and no new lock ordering.
- Only the `ServiceInvoker.Heartbeat` entry point records a context. The
  batching goroutine calls an unexported `heartbeat` that shares the body but
  does not record, so a context derived for a flush is never itself recorded.
- The acknowledged lock-release race TODO in the batching goroutine is left
  as-is. The new field is read in the same critical section as
  `lastDetailsToReport`, so that race is neither widened nor narrowed.

### One decision worth a maintainer call

If every context the caller supplied has been canceled by the time the window
closes, the flush uses `context.WithoutCancel` of the most recent one rather
than giving up. This is not `context.Background()`: it keeps the activity
context values the heartbeat path needs (activity logger, data converter,
payload visitor, propagated headers) and only drops the cancellation.
`internalHeartBeat` still bounds the RPC with its own timeout, and the batching
goroutine already returns on `closeCh`, so a heartbeat is only ever sent while
the activity is still running. Cancellation of a context a caller passed says
nothing about whether the activity is alive, which is the whole point of the
reported bug.

If you would rather the SDK drop the heartbeat in that case, the `default` arm
of `reportContextLocked` is the only thing to change and the primary behavior
stays intact.

## Checklist

1. Closes #1574

2. How was this tested:

Two new cases in `internal/activity_test.go`, both following the existing
`activityTestSuite` and gomock pattern with no fixed sleeps. They close the
batch window through `workerStopChannel` rather than waiting on the throttle
timer, and assert on `ctx.Err()` seen by the flush RPC.

- `TestActivityHeartbeat_BatchFlushUsesLastNonCanceledContext` reproduces the
  issue: open the window with a child context, buffer through it, cancel it,
  then heartbeat with the still-live parent.
- `TestActivityHeartbeat_BatchFlushWithAllContextsCanceled` covers the fallback.

Both fail on `main` with `context canceled` and pass with this change. Also ran
`go run . check` and the full `go.temporal.io/sdk/internal` package with `-race`.

3. Any docs updates needed?

Included. The doc comments on `activity.RecordHeartbeat` and
`internal.RecordActivityHeartbeat` now state that heartbeats are throttled and
what happens when the calling context is canceled. `CHANGELOG.md` has an entry
under `### Fixed`.

---
🤖 Opened with [Superhuman](https://github.com/gaurav0107/superhuman), an open-source contribution agent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,11 @@ to docs, or any other relevant information.
 
 ### Fixed
 
+- `activity.RecordHeartbeat` no longer loses a throttled heartbeat when the context it was called
+  with is canceled before the throttling window closes. The deferred send now uses the most recent
+  context passed to `RecordHeartbeat` that was not canceled. This fixes activities that heartbeat
+  from a short-lived context, such as one from `errgroup.WithContext`, timing out with
+  `activity Heartbeat timeout`.
 - Allow query results to use external storage before payload size enforcement.
 - Correct schedule catch-up window documentation to state that an unset value is omitted and the
   server applies its one-year default.
--- a/activity/activity.go
+++ b/activity/activity.go
@@ -72,6 +72,14 @@ func GetMetricsHandler(ctx context.Context) metrics.Handler {
 // details - The details that you provide here can be seen in the workflow when it receives TimeoutError. You
 // can check error with TimeoutType()/Details().
 //
+// Heartbeats are throttled, so a call does not always send a request to the server. A call made
+// inside a throttling window records the details and the SDK sends them when the window closes.
+// Cancelling ctx does not cancel a heartbeat that is already recorded: the deferred send uses the
+// most recent context passed to RecordHeartbeat that was not cancelled, so passing a short-lived
+// context, such as a context from [golang.org/x/sync/errgroup.WithContext], does not drop progress
+// recorded through it. If every context passed has been cancelled by the time the window closes,
+// the SDK still sends the heartbeat using the values from the most recent one.
+//
 // Note: If using asynchronous activity completion,
 // after returning [ErrResultPending] users should heartbeat with [go.temporal.io/sdk/client.Client.RecordActivityHeartbeat]
 func RecordHeartbeat(ctx context.Context, details ...interface{}) {
--- a/activity/activity.go
+++ b/activity/activity.go
@@ -75,9 +75,9 @@ func GetMetricsHandler(ctx context.Context) metrics.Handler {
 // Heartbeats are throttled, so a call does not always send a request to the server. A call made
 // inside a throttling window records the details and the SDK sends them when the window closes.
 // Cancelling ctx does not cancel a heartbeat that is already recorded: the deferred send uses the
-// most recent context passed to RecordHeartbeat that was not cancelled, so passing a short-lived
+// most recent context passed to RecordHeartbeat that was not canceled, so passing a short-lived
 // context, such as a context from [golang.org/x/sync/errgroup.WithContext], does not drop progress
-// recorded through it. If every context passed has been cancelled by the time the window closes,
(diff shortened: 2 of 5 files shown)


imkp1:
One design point worth surfacing, since it is the part of this change not directly covered by the guidance in the issue.

When the batch window closes, this uses the most recent context that was not canceled. If every context passed to `RecordHeartbeat` has been canceled by then, it falls back to `context.WithoutCancel` of the most recent one rather than dropping the heartbeat. That keeps the logger, data converter, and headers, and `internalHeartBeat` still bounds the RPC with its own timeout. The batching goroutine already returns on `closeCh`, so a flush only happens while the activity is alive.

This seemed consistent with the behavior documented for #2064 in temporalio/documentation#3970, which states the heartbeat is still sent even when the context is canceled. Cancellation of a context the caller passed does not by itself mean the activity has stopped.

It does not change what callers see. The direct path still passes the caller's context to `internalHeartBeat`, so a heartbeat on a canceled context still returns `context canceled`. Only the deferred batched send uses the derived context, and its error was already discarded.

Happy to drop the fallback and skip the flush instead if you would rather not send on a fully canceled context. It is isolated in `reportContextLocked` and reverting is a two line change.

CI has not run yet. The workflows need maintainer approval since this is my first contribution here.

PR #2496, opened 2026-07-27
feat(extstore): Add support for Nexus task handling

jmaeagle99:
## What was changed

- Add external storage support to Nexus task handling.
- Disable payload limit check for Nexus task completions but only warn for Nexus task failures. Parity with server and core-based SDKs.

## Why?

Round out the support for external storage for all task handlers.

## Checklist

1. Closes 

2. How was this tested:  Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 - Add support for Workflow Updates as Nexus Operations 
-- Add support for external storage to Nexus task handling
 
 ### Changed
 
@@ -41,6 +40,8 @@ to docs, or any other relevant information.
 - Added `worker.Options.PreferredVersionProvider`, which can select the version recorded by a
   newly encountered `workflow.GetVersion` call. This supports gradual rollout of a new
   `GetVersion` call before activating its new behavior.
+
+- Add support for external storage to Nexus task handling.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 - Add support for Workflow Updates as Nexus Operations 
+- Add support for external storage to Nexus task handling
 
 ### Changed
 
--- a/internal/internal_nexus_task_poller.go
+++ b/internal/internal_nexus_task_poller.go
@@ -15,16 +15,16 @@ import (
 
 type nexusTaskPoller struct {
 	basePoller
-	namespace       string
-	taskQueueName   string
-	identity        string
-	service         workflowservice.WorkflowServiceClient
-	taskHandler     *nexusTaskHandler
-	logger          log.Logger
-	numPollerMetric *numPollerMetric
-
+	namespace              string
+	taskQueueName          string
+	identity               string
+	service                workflowservice.WorkflowServiceClient
+	taskHandler            *nexusTaskHandler
+	logger                 log.Logger
+	numPollerMetric        *numPollerMetric
 	inboundPayloadVisitor  PayloadVisitor
 	outboundPayloadVisitor PayloadVisitor
+	backgroundContext      context.Context
 }
 
 type nexusTask struct {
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 - Add support for Workflow Updates as Nexus Operations 
+- Add support for external storage to Nexus task handling
```
yuandrew: Can we add this to the `### Added section`? Added a fix for the one already in the CHANGELOG, but it's still waiting on review, https://github.com/temporalio/sdk-go/pull/2490

internal/internal_nexus_task_poller.go
```diff
@@ -140,6 +146,14 @@ func (ntp *nexusTaskPoller) ProcessTask(task interface{}) error {
 		return err
 	}
 
+	// Retrieve any operation input payloads that were offloaded to external storage
+	// before the handler consumes them. A Nexus task carries at most one input and
+	// one result payload, so payload visits are unconcurrent (the configurable
+	// visit concurrency is scoped to workflow tasks).
+	if visitErr := visitProtoPayloads(context.Background(), ntp.inboundPayloadVisitor, response, 0); visitErr != nil {
```
yuandrew: Instead of using `context.Background()`, I think we need to create a task-lifecycle context before retrieval, handler execution, outbound storage, and reporting. Otherwise a blocked storage driver can out-live cancellation or shutdown. 

same for the uses of context.Background() below

internal/internal_nexus_task_poller.go
```diff
@@ -203,6 +217,17 @@ func (ntp *nexusTaskPoller) ProcessTask(task interface{}) error {
 		return err
 	}
 
+	// Offload any large result or failure payloads to external storage before reporting.
+	if res != nil {
+		if visitErr := visitProtoPayloads(context.Background(), ntp.outboundPayloadVisitor, res, 0); visitErr != nil {
```
yuandrew: Can you help me understand when we call visitProtoPayloads with `extstore.WithStorageTarget`? Looks like sometimes we do and sometimes we don't. 

Looks like we use `context.Background()` primarily in activity client, nexus client, nexus poller, and then `retrievingHistoryIterator.GetNextPage()`, and `WorkflowReplayer.replayWorkflowHistoryRoot`

test/external_storage_nexus_test.go
```diff
@@ -0,0 +1,221 @@
...
+	return res, err
+}
+
+func extStoreNexusBigResultCaller(ctx workflow.Context, endpoint string) (string, error) {
+	c := workflow.NewNexusClient(endpoint, extStoreNexusServiceName)
+	var res string
+	err := c.ExecuteOperation(ctx, extStoreNexusBigResultOp, "small", workflow.NexusOperationOptions{
+		ScheduleToCloseTimeout: 20 * time.Second,
+	}).Get(ctx, &res)
+	return res, err
```
yuandrew: Codex is saying this can mistake a workflow completion retry for a Nexus task retry. And recommends instead to validate the large Nexus result, but return something small

internal/internal_nexus_task_poller.go
jmaeagle99: The target is not strictly necessary for external storage. It just helps the driver _attempt_ to store data in a way that allows for lifecycle management, but it is a suggestion at best. That being said, we should be providing target information for every external store operation as best as possible.

There are a few gaps that should be updated (in a separate PR):
- worker-side activity heartbeat (should store in workflow or activity target depending on if its a workflow activity or standalone activity)
- client-side activity heartbeat (should store in workflow or activity target depending on the available information, fallback to workflow target with minimal information)
- gRPC message too large (should store in workflow target)

The Nexus calls _should_ have a target, but we don't have a good sense of what information should be provided to them. Ideally, it's contextual information about the consumer of the payload, but that is complicated because Nexus hides a lot of that information (intentionally). So we've deferred on that for now.

Logged issue at https://github.com/temporalio/sdk-go/issues/2502

yuandrew (approved):

PR #2508, opened 2026-07-29
NEXUS-550: Nexus Operation Query support

mavemuri:
## What was changed
Adds support for QueryWorkflow as a Nexus operation

## Why?
Part of effort to expose all Temporal primitives via Nexus operations

## Checklist

1. Closes NEXUS-550

2. How was this tested:
- [x] added integration tests 

3. Any docs updates needed?
- [ ] ~TBD~


temporalnexus/temporal_operation.go
```diff
@@ -338,6 +339,61 @@ func validateUpdateWorkflowNexusOperation(u client.UpdateWorkflowOptions) error
...
 }
 
+// StartQueryWorkflow starts a type-safe workflow query for a Nexus operation
+// linking the execution chain to the Nexus operation (callbacks, links, request ID).
+// It always returns a sync response akin to Temporal's primitive Query operation
+//
+// These are free functions because Go does not allow generic methods on non-generic structs.
+//
+// NOTE: Experimental
+func StartQueryWorkflow[R any](
```
Quinn-With-Two-Ns: We agreeded we would just use the regular clients query method, we don't need to add these

internal/interceptor.go
```diff
@@ -639,6 +639,8 @@ type ClientQueryWorkflowInput struct {
 	Args                 []interface{}
 	// QueryRejectCondition is the query reject condition.
 	QueryRejectCondition enumspb.QueryRejectCondition
+	// gRPC request response trap for nexus forward links
+	responseInfo *queryWorkflowResponseInfo
```
Quinn-With-Two-Ns: This is overkill for queries , in the query handler we can just do something like what we did for signal eg:
```
	// If this signal was issued from inside a Nexus operation handler, capture the server's response
	// link (pointing at the WorkflowExecutionSignaled event) so the task handler can attach it to the
	// StartOperationResponse, linking the caller workflow's history event to the callee. Servers
	// without history.enableCHASMSignalBacklinks leave the link unset; AddResponseLink ignores
	// nil.
	if nctx, ok := NexusOperationContextFromGoContext(ctx); ok {
		nctx.AddResponseLink(response.GetLink())
	}
```

temporalnexus/link_converter.go
```diff
@@ -38,7 +38,10 @@ func ConvertNexusLinkToLinkNexusOperation(link nexus.Link) (*commonpb.Link_Nexus
 //
 // NOTE: Experimental
 func ConvertWorkflowLinkToNexusLink(workflowLink *commonpb.Link_Workflow) nexus.Link {
-	return apinexus.ConvertWorkflowLinkToNexusLink(workflowLink)
+	// TODO: when bumping up api-go, also fix "ConvertWorkflowLinkToNexusLink" to use urlPathWorkflowTemplate
```
Quinn-With-Two-Ns: If there is a bug here we should fix it before updating the api

test/nexus_query_op_test.go
```diff
@@ -0,0 +1,167 @@
...
+	queryDoneSignal = "done"
+
+	queryTaskQueue = "nexusOpUpdateWorkflowTQ"
+)
+
+type queryInput struct {
+	WorkflowID string
+}
+
+func (ts *IntegrationTestSuite) TestNexusQueryWorkflowOperation() {
```
Quinn-With-Two-Ns, line 45: Could we also test a few other scenarios:
* Query fails, because the user returned an error from the query, that should fail the Nexus operation if the user just returns the error IMO
* Query a workflow that doesn't exist should fail the nexus operation by default
* Query when no worker is running should fail retry-ably

Quinn-With-Two-Ns (approved):

PR #2513, opened 2026-07-31
Fix Pending Activity cancellation is not removed after completion when using a custom Activity ID

### LINKED ISSUE #2512: Pending Activity cancellation is not removed after completion when using a custom Activity ID
## Expected Behavior

 When an Activity completes before a pending cancellation command is sent, the Go SDK should remove the cancellation command regardless of whether the Activity uses a custom Activity ID. The Workflow Task should complete without sending RequestCancelActivityTask for an already-completed Activity.

  ## Actual Behavior

  When an Activity uses a custom Activity ID, the SDK can fail to remove its pending cancellation command.  The cancellation state machine is keyed by the Activity’s scheduled Event ID, but Activity completion attempts to remove it using the custom Activity ID. The SDK consequently sends a stale RequestCancelActivityTask command, and the Server rejects the Workflow Task:

  `BadRequestCancelActivityAttributes: invalid history builder state for action:
  add-activitytask-cancel-requested-event, ScheduledEventID: 12`

  This can cause repeated Workflow Task failures.

  1. Schedule an Activity with a custom Activity ID.
  2. Request Workflow cancellation, causing the SDK to queue an Activity cancellation.
  3. Complete the Activity after the Workflow Task is scheduled but before it starts.
  4. Observe that the SDK sends RequestCancelActivityTask for the completed Activity.
  5. Observe the Workflow Task fail with BAD_REQUEST_CANCEL_ACTIVITY_ATTRIBUTES.

## Specifications

  - Version:
  - Platform:

Quinn-With-Two-Ns:
## What was changed

When an Activity completes before a pending cancellation command is sent, the Go SDK should remove the cancellation command regardless of whether the Activity uses a custom Activity ID. The Workflow Task should complete without sending RequestCancelActivityTask for an already-completed Activity.

## Why?

The Server reject the cancel command causing the workflow to get stuck.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/2512

2. How was this tested:
Unit tests, and replay tests

3. Any docs updates needed?
No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 
+### Fixed
+
+- Prevent workflow task failures when an activity with a custom ID completes while its cancellation
+  command is pending.
 
 ## [1.46.0] - 2026-07-28
 
--- a/internal/internal_command_state_machine.go
+++ b/internal/internal_command_state_machine.go
@@ -1137,7 +1137,7 @@ func (h *commandsHelper) handleActivityTaskClosed(activityID string, scheduledEv
 	command := h.getCommand(makeCommandID(commandTypeActivity, activityID))
 	// If, for whatever reason, we were going to send an activity cancel request, don't do that anymore
 	// since we already know the activity is resolved.
-	possibleCancelID := makeCommandID(commandTypeRequestCancelActivityTask, activityID)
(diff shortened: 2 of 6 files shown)


Sushisource (approved):
Huh, interesting one. So we weren't reading the actual event ID from history?

Quinn-With-Two-Ns:
When we generate the cancel command we use the schedule event ID from history not the activity ID

yuandrew (approved):

PR #2514, opened 2026-07-31
[AI-383] contrib/googleadk: add replay-safe OpenTelemetry provider wrappers

DABH:
## What was changed

`contrib/googleadk` gains replay-safe OpenTelemetry provider wrappers — `NewReplaySafeTracerProvider`, `NewReplaySafeLoggerProvider`, `NewReplaySafeMeterProvider`. Each wraps a real provider and drops emissions from **replaying** workflow code; everything else (first executions, activities, non-workflow contexts) passes through unchanged. The gate recovers the `workflow.Context` the adapter already stashes in the `context.Context` it hands ADK, and checks `workflow.IsReplaying`.

- **Misconfiguration warning**: `NewPlugin` logs one best-effort warning per raw, unwrapped OTel SDK provider found in the globals, at worker start and replayer creation. Wrapped, no-op, never-set-proxy, and custom providers stay silent — a false positive is worse than a miss.
- **Tests**: regression (1 real run + 5 replays → counts identical with wrappers and nonzero; delta-based control shows each replay adds one full copy without them; per-gate coverage of all 8 sync instrument kinds plus `Logger.Emit`/`Enabled`), plus unit tests for passthrough, provider classification, warning behavior, and the OTel meter error contract.
- **README**: new "Telemetry and replay" section — install snippet (wrappers must be the *first* global providers set; ADK captures the OTel proxies at package init) and exact semantics: point telemetry is first-execution-only / **at-least-once** (a workflow task retry re-records, same as `workflow.GetMetricsHandler`); spans are **at-most-once** across restarts (evicted spans export once, truncated; spans open at crash/shutdown are lost).
- **CHANGELOG** entries; **go.mod**: OTel API modules become direct requirements, OTel SDK modules added (used by detection and tests).

## Why

ADK emits its telemetry (spans carrying `gen_ai.usage.*` token attributes, `gen_ai.*` log events) through the process-global OTel API from code this adapter runs workflow-side — so every replay (cache eviction, worker restart, redeploy) re-emits all of it. Measured: 1 real execution + 5 replays → `generate_content` spans 2→12, summed input tokens 250→1500, `gen_ai.choice` events 2→12, while real model calls stayed at 2. Dashboards silently inflate by a nondeterministic multiplier.

Temporal's replay-safe telemetry (`workflow.GetMetricsHandler`, the tracing interceptor) can't help — ADK bypasses it via the globals. Hence wrappers at the provider boundary, living in this package because the gate needs the adapter's internal context bridge.

adk-go emits no OTel *metrics* yet (upstream `google/adk-go` TODO #479); the meter wrapper covers user-recorded workflow-side metrics today and ADK's metrics when they arrive.

## Checklist

1. Tested: `go vet` + `go test -race -count=1 ./...` (including the dev-server tests), repo-pinned `staticcheck`/`errcheck`, and `go run . check` — all green.
2. Docs: README updated here; a docs-site determinism/observability note is planned separately.


contrib/googleadk/telemetry.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (p replaySafeTracerProvider) Tracer(name string, opts ...trace.TracerOption) trace.Tracer {
+	return replaySafeTracer{Tracer: p.TracerProvider.Tracer(name, opts...)}
+}
+
+type replaySafeTracer struct{ trace.Tracer }
+
+func (t replaySafeTracer) Start(ctx context.Context, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	if replaySuppressed(ctx) {
+		return suppressedTracer.Start(ctx, spanName)
```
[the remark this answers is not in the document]
DABH: Good catch — fixed in 95c1cfb2. The suppressed path uses a noop tracer that ignores span start options, so behavior is unchanged, but passing `opts...` through keeps the wrapper aligned with the `trace.Tracer` contract and future-proofs the suppressed path.

contrib/googleadk/internal/pluginwarn/pluginwarn_test.go
```diff
@@ -0,0 +1,75 @@
+// Package pluginwarn_test proves NewPlugin wires the raw-global-provider
+// warning into both of its hooks: worker start and workflow replayer creation.
+// It is a separate test-only package because it must install a raw OTel SDK
+// provider as a process global, and the OTel global proxy binds its delegate
+// on the first Set*Provider call permanently — doing that in the main
+// googleadk test binary would poison the switchable globals its
+// replay-telemetry tests install.
+package pluginwarn_test
```
brianstrauch, line 11: This feels like it was generated by an agent for debugging since it's in a separate package?

contrib/googleadk/CHANGELOG.md
```diff
@@ -11,6 +11,21 @@ or Security.
...
+
+- `NewReplaySafeTracerProvider`, `NewReplaySafeLoggerProvider`, and
+  `NewReplaySafeMeterProvider` wrap OpenTelemetry providers so history replays re-emit
+  no workflow-side telemetry. ADK records spans (including `gen_ai.usage.*` token
+  attributes) and `gen_ai.*` log events through the OTel process globals from inside
+  the workflow, so previously every replay re-emitted one full copy of all of it.
+  Install the wrappers as the first global providers set in the process; non-workflow
+  telemetry passes through unchanged. Spans still open when their workflow leaves the
+  worker are truncated (sticky-cache eviction) or lost (worker shutdown/crash) rather
+  than duplicated — see "Telemetry and replay" in the README for the exact contract.
```
brianstrauch: This feels more verbose than a normal changelog entry

contrib/googleadk/telemetry.go
```diff
@@ -0,0 +1,325 @@
...
+// installed once and never wrapped, the realistic misconfiguration, is
+// recognized. NewPlugin runs this at worker start and workflow replayer
+// creation against the OTel process globals.
+func warnOnNonReplaySafeTelemetryProviders(logger log.Logger, tracerProvider, loggerProvider, meterProvider any) {
+	warn := func(global, wrapper string, p any) {
+		if !isRawOTelSDKProvider(p) {
+			return
+		}
+		logger.Warn("The global OpenTelemetry "+global+" is not replay-safe: ADK emits telemetry through it from workflow code, and every history replay will re-emit one full copy. Wrap it with googleadk."+wrapper+" and install the wrapper as the first global provider set in the process; see \"Telemetry and replay\" in the contrib/googleadk README.",
+			"provider", fmt.Sprintf("%T", p))
```
brianstrauch: Put the whole thing in a `fmt.Sprintf()`?

contrib/googleadk/CHANGELOG.md
DABH: Agreed — condensed in b802f3ef: the two entries are now four lines total, with the semantics detail living in the README's "Telemetry and replay" section instead.

contrib/googleadk/telemetry.go
DABH: Done in b802f3ef — the message is now a single `fmt.Sprintf` with `%s` placeholders.

contrib/googleadk/internal/pluginwarn/pluginwarn_test.go
DABH: Good question - reworded the comment. The reason it's a separate package is that `go test` runs each package's tests in its own process, and these tests must install a raw OTel SDK provider as a process global to prove the plugin's warning fires. The OTel global proxy binds its delegate permanently on the first Set*Provider call, so a fresh test process is the only clean way to exercise the real global-install path. Happy to drop the wiring test instead if you'd rather not carry the extra package (not the most important test imo).

brianstrauch (approved):

PR #2516, opened 2026-08-02
Honor equals function in test env MutableSideEffect

### LINKED ISSUE #2109: MutableSideEffect in test framework ignores cmp function
In this line cmp is ignored
https://github.com/temporalio/sdk-go/blob/master/internal/internal_workflow_testsuite.go#L2875

## Expected Behavior
In test framework MutableSideEffect with cmp `func(a, b any) bool { return true }` should not update value, even if value is different by any means.

## Actual Behavior
In test framework MutableSideEffect with cmp `func(a, b any) bool { return true }` updates value

run
```
func WF(ctx workflow.Context) error {
	var out1 bool
	err := workflow.MutableSideEffect(ctx, "key",
		func(workflow.Context) any { return true }, 
		nil,
	).Get(&out1)
	if err != nil {
		panic(err)
	}

	var out2 bool
	err = workflow.MutableSideEffect(ctx, "key",
		func(workflow.Context) any { return false },
		func(a, b any) bool { return true },
	).Get(&out1)
	if err != nil {
		panic(err)
	}
	
	if out1 != !out2 {
		return fmt.Errorf("side effect values are inconsistent: %v vs %v", out1, out2)
	}

	return nil
}
```
in test framework, error happens

## Specifications

  - Version: 1.26.0

PardhuTI:
## What

`TestWorkflowEnvironment.MutableSideEffect` ignored the user-supplied `equals`
function and kept no per-id state, so it returned `f()` on every call. A real
worker only records/returns a new value when `equals(new, old)` reports a
change; otherwise it returns the previously recorded value. As a result,
workflows that use `MutableSideEffect` (for example, to read dynamically
changing config) behave differently under test than in production.

Fixes #2109 (reproduced on current `main`; reported against v1.26.0).

## Change

- Track the last recorded value per id in the test environment and honor
  `equals`: the first value is recorded, and on a later call the new value is
  only recorded/returned when `equals` reports a change.
- Extract the comparison into a shared helper `isEqualMutableSideEffectValue`;
  the production `isEqualValue` now delegates to it, so the test environment and
  the real worker use identical semantics and cannot drift again.
- The mock path (`OnMutableSideEffect`) is unchanged.

## Testing

- Added `Test_MutableSideEffect_HonorsEqualsFunc`: equals-true pins the value to
  the first result, equals-false updates on each call, and distinct ids are
  independent.
- `go test ./internal -run TestUnitTestSuite` passes; production
  `Test_MutatingFunctionsInMutableSideEffect` and `...InSideEffect` pass;
  `go vet ./internal` and `gofmt` are clean.

### DIFF
--- a/internal/internal_workflow_testsuite_test.go
+++ b/internal/internal_workflow_testsuite_test.go
@@ -826,8 +826,14 @@ func (s *WorkflowTestSuiteUnitTest) Test_MutableSideEffect_HonorsEqualsFunc() {
 			func(ctx Context) interface{} { return 10 }, neverEqual))
 		changedSecond := get(MutableSideEffect(ctx, "changed",
 			func(ctx Context) interface{} { return 20 }, neverEqual))
+		// The previous call must have stored 20 as the new baseline. With equals
+		// true, this returns that stored baseline (20), not the newly computed 30.
+		// An implementation that returns the changed value but forgets to store it
+		// would keep a stale baseline (10) and return 10 here.
+		changedThird := get(MutableSideEffect(ctx, "changed",
+			func(ctx Context) interface{} { return 30 }, alwaysEqual))
 
-		return []int{pinnedFirst, pinnedSecond, changedFirst, changedSecond}, nil
+		return []int{pinnedFirst, pinnedSecond, changedFirst, changedSecond, changedThird}, nil
 	}
 
 	env := s.NewTestWorkflowEnvironment()
@@ -837,7 +843,7 @@ func (s *WorkflowTestSuiteUnitTest) Test_MutableSideEffect_HonorsEqualsFunc() {
 	s.NoError(env.GetWorkflowError())
 	var result []int
 	s.NoError(env.GetWorkflowResult(&result))
-	s.Equal([]int{1, 1, 10, 20}, result)
+	s.Equal([]int{1, 1, 10, 20, 20}, result)
 }
 
 func (s *WorkflowTestSuiteUnitTest) Test_LongRunningSideEffect() {
--- a/internal/internal_workflow_testsuite_test.go
+++ b/internal/internal_workflow_testsuite_test.go
@@ -801,6 +801,45 @@ func (s *WorkflowTestSuiteUnitTest) Test_OnMutableSideEffect() {
 	env.AssertExpectations(s.T())
 }
 
+func (s *WorkflowTestSuiteUnitTest) Test_MutableSideEffect_HonorsEqualsFunc() {
+	// The test environment must honor the user-supplied equals function the same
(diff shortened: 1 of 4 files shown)


internal/internal_workflow_testsuite_test.go
```diff
@@ -801,6 +801,45 @@ func (s *WorkflowTestSuiteUnitTest) Test_OnMutableSideEffect() {
...
+		pinnedFirst := get(MutableSideEffect(ctx, "pinned",
+			func(ctx Context) interface{} { return 1 }, alwaysEqual))
+		pinnedSecond := get(MutableSideEffect(ctx, "pinned",
+			func(ctx Context) interface{} { return 2 }, alwaysEqual))
+
+		// equals always false: value updates on every call.
+		changedFirst := get(MutableSideEffect(ctx, "changed",
+			func(ctx Context) interface{} { return 10 }, neverEqual))
+		changedSecond := get(MutableSideEffect(ctx, "changed",
+			func(ctx Context) interface{} { return 20 }, neverEqual))
```
yuandrew, line 828: Could we add a third call for the "changed" ID that depends on the newly stored baseline? This is unlikely, by technically an implementation that returns the cahnged value but forgets to save it would still pass these 2 calls and use stale state later.

yuandrew (commented):
Overall LGTM, but also needs a CHANGELOG.md entry, as this affects users's use of TestWorkflowEnvironment

PardhuTI:
Thanks for the review. Added a CHANGELOG entry under Unreleased > Fixed in a7a8caa.

yuandrew (approved):
Thanks for the contribution!

internal/internal_workflow_testsuite_test.go
PardhuTI: Good point. Added a third call on the "changed" id with equals=true; it reads back the stored baseline (returns 20, not the freshly computed 30), so an implementation that returns the changed value without storing it would fail here (it would return the stale 10). Done in c5dda6f.

PardhuTI:
The failing check `integ-test (macos-intel, stable)` is unrelated to this change. It's `TestPayloadLimitsTestSuite` (test/payload_limits_test.go:73) failing with a transient `503 Service Unavailable / 502 Bad Gateway` while fetching info, failing in 0.25s. This PR only touches the test-env `MutableSideEffect` and a shared equals helper, which are covered by the unit-test jobs (green across the whole matrix). Could you re-run that job when you get a chance?

Both review points are addressed:
- Added the CHANGELOG.md entry under Unreleased > Fixed.
- Added a third call on the "changed" id that reads back the newly stored baseline (returns the stored 20, not the freshly computed 30), so an implementation that returns the changed value but forgets to store it would fail here.

PR #2518, opened 2026-08-03
contrib: remove ObjectExists check from Store in S3 and GCS drivers

### LINKED ISSUE #2497: External storage drivers: consider unconditional (check-free) uploads for content-addressable keys
Currently the external storage drivers (S3, GCS) use a check-then-write pattern: `ObjectExists` followed by `PutObject`. Since keys are content-addressable (SHA-256), an overwrite produces identical content, making the existence check unnecessary in most cases.

Removing the check would halve GCS/S3 latency per payload in the common case (new object) since the existence check almost always returns false (keys include the RunID). For GCS specifically, conditional uploads (`storage.Conditions{DoesNotExist: true}`) could make the write explicitly idempotent while avoiding unnecessary data transfer for cache hits.

Worth discussing whether this optimization should be adopted across all drivers.

Related: #2366

brucearctor:
## Summary

Both external storage drivers used a check-then-write pattern (`ObjectExists` followed by `PutObject`) that added an unnecessary round-trip per payload. Since keys are content-addressable (SHA-256) and include the RunID, overwrites produce identical content, making the existence check redundant in virtually all cases.

This change removes the `ObjectExists` call from `Store` in both drivers, halving per-payload latency in the common case (new object). The `Client` interface retains `ObjectExists` for testing and potential future use.

## Changes

**`contrib/gcp/gcsdriver/driver.go`** and **`contrib/aws/s3driver/driver.go`**:
- Remove `ObjectExists` → conditional `PutObject` pattern from `Store`
- Replace with direct `PutObject` call (idempotent overwrite)
- Update `Store` docstrings

**`contrib/gcp/gcsdriver/README.md`** and **`contrib/aws/s3driver/README.md`**:
- Update IAM/permissions docs: storing no longer requires read permissions (`storage.objects.get` / `s3:GetObject`)
- Store-only components can now use narrower permission sets

**`contrib/gcp/gcsdriver/driver_test.go`** and **`contrib/aws/s3driver/driver_test.go`**:
- Update `TestStore_Deduplication`: now expects 2 puts (idempotent overwrite), verifies identical claims
- Remove `TestStore_ObjectExistsError`: no longer applicable since `Store` does not call `ObjectExists`

## Rationale

- The existence check almost always returns `false` because keys include the RunID — the only way to get a cache hit is storing the exact same serialized bytes within the same Workflow execution
- Removing the check halves GCS/S3 latency per payload in the common case
- Content-addressable keys make overwrites idempotent by definition — the bytes are identical
- Marginal increase in PUT/write request costs for the rare duplicate is negligible

Closes #2497


brucearctor:
Also, this is for after https://github.com/temporalio/sdk-go/pull/2366 gets merged.

brucearctor:
@jmaeagle99  since related to the other PR you are reviewing [ https://github.com/temporalio/sdk-go/pull/2366 ].  Made concrete one option for addressing https://github.com/temporalio/sdk-go/issues/2497

Please have a look at your convenience.

PR #2519, opened 2026-08-04
contrib: bound memory on Retrieve with size-limited reader

### LINKED ISSUE #2498: External storage drivers: bound memory on Retrieve with size-limited reader
The external storage drivers (S3, GCS) download payloads via `io.ReadAll()` with no size bound on the Retrieve path. Since claim data is user-controllable (injected via workflow starts/signals), a crafted claim could point to an arbitrarily large object, causing unbounded memory allocation.

A post-download size check doesn't fully address this since the data is already materialized in memory. A proper fix would use `io.LimitReader` on the download stream, decoupled from the upload limit, and should be adopted across all drivers and language SDKs.

Related: #2366

brucearctor:
## Problem

Both S3 and GCS storage drivers use `io.ReadAll()` with no size bound on the `Retrieve` path. Since claim data is user-controllable (injected via workflow starts/signals), a crafted claim could point to an arbitrarily large object, causing unbounded memory allocation.

`MaxPayloadSize` is enforced on the upload (`Store`) path but the download (`Retrieve`) path is completely unbounded.

### 1. `Client.GetObject` returns `io.ReadCloser` instead of `[]byte`

The root cause is that `GetObject` returned `[]byte`, forcing full materialization inside the client adapter. The client now returns a stream — the driver controls the read lifecycle.

**This is a breaking change** to the (`Experimental`) `Client` interface. Migration: update your `GetObject` to return an `io.ReadCloser`. Use `io.NopCloser(bytes.NewReader(data))` to wrap in-memory bytes.

### 2. `io.LimitReader` in the driver's `Retrieve` method

Each driver wraps the stream with `io.LimitReader(reader, maxRetrieveSize+1)` and checks the read length. Objects exceeding the limit are rejected with the new `ErrPayloadTooLarge` sentinel error before being fully read into memory.

### 3. New `MaxRetrieveSize` option

Defaults to `max(MaxPayloadSize, 50 MiB)`. This decouples the retrieve bound from the store bound — operators can reduce `MaxPayloadSize` for new writes without breaking retrieval of older, larger payloads.

### 4. `StorageDriver.Retrieve` contract documented

The interface contract now documents that implementations must bound total bytes read per payload, since the SDK cannot enforce this after the driver returns.

## Test coverage

14 new tests across both drivers:
- Constructor validation: negative, default, `max()` behavior, explicit override
- Retrieve behavior: oversized rejected, exactly at limit succeeds, one byte over rejected

All existing tests (including fake-server integration tests) pass.

Closes #2498


brucearctor:
Might not be the desired implementation.  Since there wasn't much discussion on #2498 , I figured just put something out there.

PR #2521, opened 2026-08-04
feat: add workflow.GetRandomStream and workflow.IsReadOnly

### LINKED ISSUE #2517: Deterministic randomness API
## Summary

Add a deterministic, replay-stable randomness API.

The motivating consumer is plugins, but the API is general purpose and usable from workflow code. Values are derived from workflow identity rather than recorded in history, so a caller gets the same sequence on every replay at no history cost.

## Motivation

Plugins need entropy inside a workflow (span and trace IDs) that regenerates identically on replay. `SideEffect` can produce it, but records each value in history, so adding or removing a plugin changes what a running workflow replays. It is also unavailable in query handlers and update validators.

## Requirements

- Named deterministic random sources: independent per name, so one caller cannot shift what another observes.
- Nothing written to history; values regenerate from workflow identity on replay.
- Stable across reset (a run reset to an earlier workflow task reproduces the values it drew originally).
- Independent per continue-as-new run.
- Usable from query handlers and update validators without advancing the workflow's deterministic sequence (values there need not be replayable).
- Experimental public API under `workflow`.

## Scope

**In:** the API, with defined behavior across replay, reset, continue-as-new, and read-only paths.

**Out:** changing existing randomness guidance in `workflow/doc.go`.

patbeqo:
## What was changed

Added `workflow.GetRandomStream`, a deterministic pseudorandom source private to a caller-supplied name. Returns `*rand.ChaCha8` (not a `*rand.Rand` wrapper) so both the concrete source and a `rand.New` wrapper stay reachable. Seeded from workflow identity, so values reproduce on replay without touching history.

Also added `workflow.IsReadOnly`, which reports whether the workflow context is running where state cannot be mutated. Added so callers can gate draws for OpenTelemetry tracing and other cases where a seeded random must not be used in a read-only context.

## Why?

#2517

## Checklist

1. Closes #2517

2. How was this tested:

- Unit tests and integration tests created

3. Any docs updates needed?

`CHANGELOG.md` entry and `GetRandomStream` / `IsReadOnly` doc comments added.

---

> [!NOTE]
> **High Risk**
> Introduces replay-sensitive pseudorandom state not recorded in history and pins seed derivation via golden tests; incorrect use or future seed changes would break workflow replay. Reset reseeding and read-only gating touch core dispatcher and query paths.
>
> **Overview**
> Adds **experimental** `workflow.GetRandomStream(ctx, name)` returning a memoized `*rand.ChaCha8` seeded from the current run ID and name (SHA256 with versioned salt). Repeated calls with the same name continue the same stream; draws are **not** written to history, so replay must consume randomness in the same order. On workflow **reset**, existing sources are re-seeded in place when the run ID changes so pre-reset draws replay identically and post-reset draws diverge; **continue-as-new** gets a new run ID and thus a new sequence.
>
> Adds **experimental** `workflow.IsReadOnly(ctx)` so shared code can avoid mutating workflow state (e.g. skipping `GetRandomStream` draws) in query handlers, update validators, and side-effect callbacks. Query execution now toggles the dispatcher read-only flag around `HandleQuery`, aligning with existing read-only enforcement for validators and side effects.
>
> Worker, test environment, and changelog are updated; unit golden tests pin seed/output for replay compatibility, plus integration tests for replay queries, reset, and continue-as-new.
>
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1b6050373ce699fa3e937906304fe5030cc1b2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workflow_random.go
```diff
@@ -0,0 +1,36 @@
...
+func deriveSeed(runID, name string) [32]byte {
+	// The separators keep ("ab", "c") from colliding with ("a", "bc")
+	return sha256.Sum256([]byte(seedVersion + "\x00" + runID + "\x00" + name))
+}
+
+func getRandomSeed() [32]byte {
+	var seed [32]byte
+	_, _ = cryptorand.Read(seed[:])
+	return seed
+}
```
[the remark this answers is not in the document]
patbeqo: https://pkg.go.dev/crypto/rand#Read
> Read fills b with cryptographically secure random bytes. It never returns an error, and always fills b entirely.

internal/internal_workflow_testsuite.go
```diff
@@ -741,6 +744,10 @@ func (env *testWorkflowEnvironmentImpl) GenerateSequence() int64 {
 	return env.nextID()
 }
 
+func (env *testWorkflowEnvironmentImpl) Random(name string) *rand.Rand {
+	return getRandom(env.randoms, env.workflowInfo.OriginalRunID, name)
+}
```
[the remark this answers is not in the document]
patbeqo: Updated `testWorkflowEnvironmentImpl` to set `OriginalRunID`

yuandrew:
bugbot run

internal/internal_workflow_testsuite.go
```diff
@@ -741,6 +744,10 @@ func (env *testWorkflowEnvironmentImpl) GenerateSequence() int64 {
 	return env.nextID()
 }
 
+func (env *testWorkflowEnvironmentImpl) GetRandomStream(name string) *rand.ChaCha8 {
+	return getRandomStream(env.randoms, env.workflowInfo.currentRunID, name)
```
yuandrew, line 747: I think this makes child reruns in `TestWorkflowEnvironment` restart the same random stream. Each child run gets `<workflowID>_RunID`, and continue-as-new/retry/cron reruns reuse the same workflow ID, so `currentRunID`, and therefore the seed, never changes. In production, each run gets a new Run ID. Can we give each simulated child run a unique deterministic Run ID and add a child continue-as-new test?

workflow/workflow.go
```diff
@@ -465,6 +466,37 @@ func SideEffectWithOptions(ctx Context, options SideEffectOptions, f func(ctx Co
...
+//	buf := make([]byte, 32)
+//	c := workflow.GetRandomStream(ctx, "example.com/myplugin/bytes")
+//	c.Read(buf)
+//
+// A reset replays the same values up to the reset point, so decisions the
+// workflow already made still hold. Past that point the source starts a fresh
+// sequence rather than handing back the values the abandoned run drew.
+//
+// Each continue-as-new run gets its own sequence and never repeats the values
+// drawn by the run before it.
```
yuandrew: “Never repeats” feels stronger than what we guarantee here. Each continue-as-new run is independently seeded, but two pseudorandom streams can still collide. Can we say “Each continue-as-new run gets an independently seeded sequence” instead, so callers don’t read this as a uniqueness guarantee?

internal/internal_event_handlers.go
```diff
@@ -762,6 +767,10 @@ func (wc *workflowEnvironmentImpl) GenerateSequence() int64 {
 	return wc.commandsHelper.getNextID()
 }
 
+func (wc *workflowEnvironmentImpl) GetRandomStream(name string) *rand.ChaCha8 {
```
yuandrew: I agree that keeping `Read` available is useful, but I don’t think that requires exposing `*rand.ChaCha8` directly. The SDK’s need to reseed on reset is an implementation detail; it can retain and reseed the concrete source behind a small SDK-owned interface:

```
type RandomStream interface {
    rand.Source
    io.Reader
}
```

Backed by an unexported wrapper, this still lets callers use Read directly or pass the stream to rand.New. This also mirrors the TS shape, where getRandomStream returns a WorkflowRandomStream rather than exposing the underlying PRNG.

Since streams are memoized by name, exposing `Seed` and `UnmarshalBinary` means any caller holding that named stream can reset or rewind its shared state. That can repeat generated IDs, bypass the SDK’s seed derivation and reset behavior, and cause replay divergence if the mutation is not reproduced identically. Keeping those operations internal makes the SDK the only component that can reseed the stream.

It also avoids putting ChaCha8 in the public signature and gives us the option to define stable Read/Uint64 interleaving semantics instead of inheriting ChaCha8’s undefined ordering. How does that sound?

internal/internal_workflow_testsuite.go
patbeqo: I've updated `runID` to use `nextID`, similar to how `workflowID` is set. I've also added a unit test for the child continue-as-new case.

internal/internal_event_handlers.go
patbeqo: That makes sense to me. I’ve updated the implementation to use an unexported wrapper with a public interface. 

I added a TODO for stable `Read` and `Uint64` interleaving which I propose addressing in a follow-up PR.

internal/workflow_random.go
```diff
@@ -12,23 +13,50 @@ func deriveSeed(runID, name string) [32]byte {
...
+// Exposed as: [go.temporal.io/sdk/workflow.WorkflowRandomStream]
+type WorkflowRandomStream interface {
+	rand.Source
+	io.Reader
+}
+
+// workflowRandomStream wraps ChaCha8 instead of embedding it so callers cannot
+// mutate a shared named stream through Seed or UnmarshalBinary.
+//
+// TODO: Define stable interleaving semantics. See [ChaCha8.Read].
```
yuandrew: Can we track this as a GH issue? Just so this TODO doesn't get lost

yuandrew: Maybe in the issue itself mention that `WorkflowRandomStream` should remain experimental until this is ironed out

yuandrew (approved):
Thanks for implementing this! 😄

internal/workflow_random.go
patbeqo: Created a separate issue and linked the TODO to it.

yuandrew (approved):

PR #2522, opened 2026-08-04
Decode LastHeartbeatDetails for reset-workflow failures

mittalpk:
## What was changed
`FailureToError`'s `ResetWorkflowFailureInfo` branch passed the raw `*commonpb.Payloads` proto straight through as the `ApplicationError`'s details, instead of wrapping it with `newEncodedValues` like every sibling branch (Timeout, Canceled, Activity, etc.) does. Calling `Details()` on the resulting error failed with `too many arguments` instead of decoding the payload. Fixed to match the sibling pattern.

## Why?
Every other failure-info branch in `FailureToError` wraps its payload the same way before handing it to the caller; this one didn't, so `Details()` broke specifically for reset-workflow failures — an easy case to miss since it's not the common path, but a genuine inconsistency with the rest of the function.

## Checklist
1. Closes: no linked issue — found via code audit, not filed as a bug report first
2. How was this tested: added a regression test constructing a `Failure` with `ResetWorkflowFailureInfo` directly and asserting `Details()` decodes correctly; confirmed it fails on `main` with `too many arguments` and passes after the fix. `go run . check` and the full `internal` package test suite pass with no regressions.
3. Any docs updates needed?: no


dplyukhin:
Thanks for the contrib. Does this close an existing issue? Please use the PR template we provided and review the contributing guide: https://github.com/temporalio/sdk-go/blob/main/CONTRIBUTING.md

mittalpk:
No linked issue — this is a small, scoped bug fix (root cause + fix + regression test in the description above), which your CONTRIBUTING.md says can go straight to a PR. I don't see a PR template file in .github either, just ISSUE_TEMPLATE. Happy to add anything specific if I'm missing something.

yuandrew:
The template is org-wide, which is why I believe there is no file in `.github`. But when you open a PR, the template loads
```
<!--- Note to EXTERNAL Contributors -->
<!-- Thanks for opening a PR! 
If it is a significant code change, please **make sure there is an open issue** for this. 
We work best with you when we have accepted the idea first before you code. -->

<!--- For ALL Contributors 👇 -->

## What was changed
<!-- Describe what has changed in this PR -->

## Why?
<!-- Tell your future self why have you made these changes -->

## Checklist
<!--- add/delete as needed --->

1. Closes <!-- add issue number here -->

2. How was this tested:
<!--- Please describe how you tested your changes/how we can test them -->

3. Any docs updates needed?
<!--- update README if applicable
      or point out where to update docs.temporal.io -->

```

mittalpk:
PR updated as per template.

internal/error_test.go
```diff
@@ -1253,6 +1253,34 @@ func Test_convertFailureToError_ApplicationFailure(t *testing.T) {
...
+	}
+
+	err = fc.FailureToError(f)
+	var applicationErr *ApplicationError
+	require.True(errors.As(err, &applicationErr))
+	require.Equal(true, applicationErr.NonRetryable())
+
+	// Before the fix, the raw *commonpb.Payloads proto was passed straight through as the
+	// error's details instead of being wrapped by newEncodedValues, so Details() panicked
+	// with a reflect type mismatch instead of decoding the payload.
```
yuandrew: I think it would return `ErrTooManyArg` instead of decoding it, 
```suggestion
	// Before the fix, the raw *commonpb.Payloads proto was treated as a single
	// detail value, so Details() returned ErrTooManyArg instead of decoding it.
```

yuandrew (commented):
Thanks for the contrib, can we update this branch and ensure the CHANGELOG.md entry is in the right spot? We just released 1.48, so I think the entry needs to be moved to a new section

mittalpk:
Rebased onto main and updated the comment per your suggestion. Also moved the CHANGELOG entry -- 1.48.0 had already been cut and released while this was open, so it was sitting in an already-shipped section instead of Unreleased.

yuandrew (approved):
Thanks for the contribution!

PR #2523, opened 2026-08-04
Add script to prepare new releases

dplyukhin:
## What was changed
  Added `internal/cmd/prepare-release`, a command that that opens a draft PR and draft release for the given SDK version. It:

  - Validates the requested version, changelog state, SDK version, and API dependency.
  - Updates the changelog.
  - Updates `internal.SDKVersion`.
  - Commits and pushes a release branch.
  - Opens a draft PR and draft GitHub release.
  - Supports `--stop-before-push` for local validation without remote writes.
  - Preserves failed results in a worktree, along with a record of commands executed and worktree cleanup instructions.

The script is based on [prepare_release.py](https://github.com/temporalio/sdk-python/blob/main/scripts/prepare_release.py) in the Python SDK.

## Why?

To simplify the Go release process

## Checklist

1. Closes 

2. How was this tested:

  Added unit tests covering:

  - Version and changelog validation.
  - Changelog and SDK version updates.
  - Failure during preparation, e.g. if github is down.
  - Worktree creation and cleanup.
  - `--stop-before-push`.

3. Any docs updates needed?


dplyukhin:
~~Should the script also programmatically create the GitHub release? That part will need to take in the merged commit hash as input.~~ EDIT: I updated the script to also create a draft PR.

I'm adding the skip-changelog label since this doesn't need to go in user-facing changelog imho

internal/cmd/prepare-release/main.go
```diff
@@ -0,0 +1,374 @@
...
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var (
+	// Matches release versions such as "1.48.0".
+	versionRE = regexp.MustCompile(`^[0-9]+\.[0-9]+\.[0-9]+$`)
```
yuandrew: nit: Should we validate the canonical SemVer? I think technically 01.48.0 would be okay with the current regex rule

internal/cmd/prepare-release/main.go
```diff
@@ -0,0 +1,374 @@
...
+	// Validate git state and create a release branch
+
+	if err := ensureCleanWorktree(eff, root); err != nil {
+		return "", "", err
+	}
+	if _, err := eff.runCommand(root, "git", "fetch", "origin", "main"); err != nil {
+		return "", "", err
+	}
+	branch := "chore/release-" + version
+	if _, err := eff.runCommand(root, "git", "switch", "--create", branch, "origin/main"); err != nil {
```
yuandrew: I think currently if any of these steps fail, it puts us in a weird state that we can't simply re-run the script, due to failed branch creation or the clean worktree check.

Do we want to roll back changes on failure, or maybe add logic to skip steps like branch creation if branch is already created?

internal/cmd/prepare-release/main.go
```diff
@@ -0,0 +1,374 @@
...
+
+// replaceSDKVersion updates the sole SDKVersion declaration in version.go.
+func replaceSDKVersion(text, version string) (string, error) {
+	if _, err := validateVersion(version); err != nil {
+		return "", err
+	}
+	if len(sdkVersionRE.FindAllStringIndex(text, -1)) != 1 {
+		return "", errors.New("could not find exactly one SDKVersion declaration in internal/version.go")
+	}
+	return sdkVersionRE.ReplaceAllString(text, "${1}"+version+"${2}"), nil
```
yuandrew: Do we want to add logic to validate that the version number is strictly increasing? Either by major, minor, or patch #? Seems like a nice safeguard we get with a script, versus manually doing

internal/cmd/prepare-release/main.go
```diff
@@ -0,0 +1,374 @@
...
+	// Commit and push changes, then open a PR
+
+	commitArgs := append([]string{"commit", "-m", "Prepare release " + version, "--"}, releaseFiles...)
+	if _, err := eff.runCommand(root, "git", commitArgs...); err != nil {
+		return "", "", err
+	}
+	if _, err := eff.runCommand(root, "git", "push", "--set-upstream", "origin", branch); err != nil {
+		return "", "", err
+	}
+	prURL, err := eff.runCommand(root, "gh", "pr", "create", "--base", "main", "--head", branch,
```
yuandrew: I like the idea of creating a PR in draft mode, where the releaser would have to manually mark "Ready for review", that way they've hopefully verified the changes. Thinking similarly from how AI generated code is untrusted and should be reviewed before posting/marking for review

What do you think?

internal/cmd/prepare-release/main.go
dplyukhin: I changed this to ensure the leading digit is 1, not sure if it's worth doing more than that

internal/cmd/prepare-release/main.go
dplyukhin: Good point! Went kinda crazy and did a huge refactor. Now it creates a worktree, which you can inspect and manually clean up in case of failure. I figured if branch already exists, some deeper issue is probably at play, so I keep failing there.

internal/cmd/prepare-release/main.go
dplyukhin: Done. I hard-coded that the major release is 1 because in practice we really wouldn't want to bump that number

internal/cmd/prepare-release/main.go
dplyukhin: Good idea. Now everything it creates is in draft mode.

yuandrew (approved):

PR #2525, opened 2026-08-05
Fix double sending link on SAA

Quinn-With-Two-Ns:
## What was changed

Fix double sending link on SAA + Nexus. Previously if a Standalone Activity was started by a Nexus Operation handler we attached the link to the callback and the request

## Why?

Previously I based the SAA code on the Workflow code that has some legacy behaviour for old server that we don't need to handle for SAA.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes Nexus–activity linking and start-request shaping on the client `ExecuteActivity` path, which affects correlation and idempotent conflict attachment but not auth or data handling.
> 
> **Overview**
> Fixes **duplicate link attachment** when a Nexus operation handler starts a **standalone activity** via `client.ExecuteActivity`. Caller inbound links are now added on the **start request only when there is no completion callback** (async activity-backed Nexus ops keep links on the callback). **OnConflictOptions** for attaching request ID, callbacks, and links is set only inside a Nexus operation context when those fields are actually present.
> 
> **Response link propagation** moves into `ExecuteActivity`: the server’s activity start link is recorded on the Nexus operation context via `AddResponseLink`, replacing the removed `ClientStartActivityOptions` / `temporalnexus` helpers that duplicated links on both the request and callback paths.
> 
> Adds **`TestStandaloneActivityStartLinks`** (bidirectional Nexus↔activity links) and a unit test that a linkless Nexus `ExecuteActivity` does not send **OnConflictOptions**. Integration tests drop stubbed NEXUS-400 link checks and assert **workflow-event links on shared-activity callbacks** instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f2298c341a56d2c55c5847f7edddd7c019741f64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/internal_activity_client.go
+++ b/internal/internal_activity_client.go
@@ -599,7 +599,8 @@ func (w *workflowClientInterceptor) ExecuteActivity(
 			request.Links = links
 		}
 	}
-	if _, ok := NexusOperationContextFromGoContext(ctx); ok {
+	if _, ok := NexusOperationContextFromGoContext(ctx); ok &&
+		(len(request.GetCompletionCallbacks()) > 0 || len(request.GetLinks()) > 0) {
 		request.OnConflictOptions = &commonpb.OnConflictOptions{
 			AttachRequestId:           request.GetRequestId() != "",
 			AttachCompletionCallbacks: len(request.GetCompletionCallbacks()) > 0,
--- a/internal/internal_activity_client.go
+++ b/internal/internal_activity_client.go
@@ -113,27 +113,12 @@ type (
 		// StartDelay - Time to wait before dispatching the activity. This delay is not applied to retry attempts.
 		StartDelay time.Duration
 
-		// responseInfo holds the response Link populated by the server when the activity is started.
-		// Only settable by the SDK - e.g. [temporalnexus.temporalOperation].
-		responseInfo *startActivityResponseInfo
 		// requestID is the request ID used to dedup retried starts.
 		// Only settable by the SDK - e.g. [temporalnexus.temporalOperation].
 		requestID string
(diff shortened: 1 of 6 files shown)


VegetarianOrc (approved):

internal/internal_activity_client.go
```diff
@@ -605,6 +590,22 @@ func (w *workflowClientInterceptor) ExecuteActivity(
...
+		if links, ok := ctx.Value(NexusOperationRequestLinksKey).([]*commonpb.Link); ok {
+			request.Links = links
+		}
+	}
+	if _, ok := NexusOperationContextFromGoContext(ctx); ok {
+		request.OnConflictOptions = &commonpb.OnConflictOptions{
+			AttachRequestId:           request.GetRequestId() != "",
+			AttachCompletionCallbacks: len(request.GetCompletionCallbacks()) > 0,
+			AttachLinks:               len(request.GetLinks()) > 0,
+		}
```
yuandrew: Codex found a potential bug, 

`ExecuteActivity` always generates a request ID, so this sends `AttachRequestId: true` even when the Nexus request has no inbound links and the activity has no completion callback. Nexus links are optional, and the server rejects this combination because `attach_request_id` requires at least one callback or link. Could we omit `OnConflictOptions` when both collections are empty and add coverage for a linkless Nexus request?

yuandrew (approved):

Issue #2507, opened 2026-07-29
Add optional retry options to NewRemotePayloadCodec


yuandrew:
**Is your feature request related to a problem? Please describe.**
`NewRemotePayloadCodec` makes one HTTP request for each encode or decode operation. Transient transport errors, throttling, and server failures are returned immediately.

Generic retry behavior does not belong in `CodecDataConverter`, but `RemotePayloadCodec` is SDK-owned and cannot otherwise be configured to retry.

**Describe the solution you'd like**
Add optional, bounded retry settings to `RemotePayloadCodecOptions`, such as:
- Maximum attempts
- Exponential-backoff intervals
- optional expiration
- retryable error/status classification

Retries should be disabled by default for backward compatibility.

## Acceptance criteria

- Existing callers still make one request
- Encode and decode can recover from transient failures
- Non-retryable failures make one attempt
- Documentation warns that retries consume the Workflow Task timeout budget

**Additional context**

Related: #2370 #2506

PR #2529, opened 2026-08-06
Add PayloadCodec WorkflowTaskFailureError marker

### LINKED ISSUE #2506: Allow PayloadCodec to request Workflow Task failure
A returned `PayloadCodec` error during workflow-side decoding can fail the Workflow Execution instead of the current Workflow Task.

This affects workflow inputs and errors returned from `Future.Get`. Codec authors can force Workflow Task retry by panicking, but that is undocumented and produces misleading panic logs and metrics.

Add an opt-in error marker, tentatively `WorkflowTaskFailureError`, that a codec can return to request Workflow Task failure.

The marker should:
- Fail the current WFT while keeping the Workflow Execution open
- Be independent of `WorkflowPanicPolicy`
- Preserve the underlying error through `errors.Is`/`errors.As`
- Not be classified or logged as a workflow panic
- Behave as an ordinary error in Activity-worker and client contexts

## Acceptance criteria

- Workflow-input decoding fails once, retries the task, then completes in the same Workflow Execution
- Activity-result decoding fails once; the Activity executes once and decoding succeeds on task replay
- Both `BlockWorkflow` and `FailWorkflow` honor the marker identically
- Unmarked errors preserve existing behavior
- Adjacent workflow-side codec call sites are audited and their behavior documented

**Additional context**

Other SDKs route ordinary workflow-side codec exceptions to Workflow Task failure. Go cannot change its default behavior compatibly, so this adds an explicit opt-in marker for the same outcome.

Related: #2370

whitemanthedj:
## What was changed

Adds an opt-in error marker, `converter.WorkflowTaskFailureError`, that a `PayloadCodec` can return from `Encode` or `Decode` to request that the SDK fail the current Workflow Task instead of the Workflow Execution.

- New `converter.WorkflowTaskFailureError` type plus a `converter.NewWorkflowTaskFailureError(cause)` constructor. It wraps a cause and exposes it through `errors.Is` / `errors.As`.
- `applyWorkflowPanicPolicy` detects the marker on a returned workflow-side codec error and returns it so it becomes a `WorkflowTaskFailed`. The server then retries the task while the Workflow Execution stays open.

The routing honors the constraints from the issue: it is checked before the `WorkflowPanicPolicy` switch (so `BlockWorkflow` and `FailWorkflow` behave identically), it is not converted into a `workflowPanicError` and is not logged or classified as a workflow panic, it preserves the underlying cause via `errors.Is` / `errors.As`, and in activity-worker and client contexts it behaves as an ordinary returned error.

## Why?

A returned `PayloadCodec` error during workflow-side decoding fails the Workflow Execution, not just the current task. This affects workflow inputs and errors returned from `Future.Get`: `workflowExecutor.Execute` returns the decode error, it is stored as an ordinary workflow result error, and `applyWorkflowPanicPolicy` only promotes `*workflowPanicError` into the task-failure branch, so a plain returned error falls through to a `FailWorkflowExecution`.

Today the only way for a codec author to get task-retry behavior on that path is to `panic` from inside the codec and rely on the coroutine recover turning it into a `workflowPanicError`. That works, but it is undocumented, it only works in workflow context (activity and client-side codec calls get nothing), and it surfaces as a workflow panic in logs and metrics. This marker gives Go codec authors a supported way to request the same workflow-side outcome that codec exceptions produce in other SDKs, without the panic idiom. In-codec retry remains the primary recovery mechanism since codec calls consume the Workflow Task timeout budget; the marker is a backstop after local retries are exhausted.

This came out of the discussion in #2370, which was split into #2506 (this change) and #2507 (retry options on `NewRemotePayloadCodec`, not covered here).

### Adjacent workflow-side sites (audit)

Per the acceptance criteria, where a codec error reaches the marker routing today and where it does not:

- Honored: workflow input decode, workflow result encode, activity result decode via `Future.Get`, dynamic workflow args (when the workflow propagates the error, the normal pattern).
- Not honored, by design: queries. A query handler error becomes a query failure returned to the caller, not a Workflow Task failure. Queries are read-only and replayable, so failing the task on a query would wrongly block the workflow.
- Not honored, documented limitations: signals swallow decode errors (`assignValue` logs "Corrupted signal" and drops the signal); update argument decode routes a codec error to an update Rejection; update result encode already fails the task, but via `panic` into `workflowPanicError` (the idiom this change replaces). Honoring the marker in the signal receive loop and the update decode path would be reasonable follow-ups; left out to keep this focused.

## Checklist

1. Closes #2506 

2. How was this tested:

Unit tests:
   - `converter`: marker unwraps and matches via `errors.Is` / `errors.As`, including after being re-wrapped with `%w` the way the decode path wraps codec errors; nil-cause behavior.
   - `internal` task handler: a marker returned during workflow input decode fails the Workflow Task and produces no completion request, asserted under both `BlockWorkflow` and `FailWorkflow`; a marker during activity result decode does the same; an unmarked decode error preserves existing behavior and still fails the Workflow Execution.

   Integration tests (embedded dev server):
   - Input decode: a codec returns the marker once, the task retries, the retry decodes cleanly, and the Workflow Execution completes.
   - Activity result decode: same fail-once behavior, and the activity runs exactly once because the retry replays its cached result from history.

   Ran via the build tool: `go run . check`, `go run . unit-test`, and `go run . integration-test -dev-server -run "TestIntegrationSuite/TestWorkflowTaskFailureMarker"`.

3. Any docs updates needed?

   No external docs.temporal.io changes. The public `WorkflowTaskFailureError` type and constructor carry doc comments describing the contract and the "in-codec retry first, marker as backstop" guidance, and there is a `CHANGELOG.md` entry under `## [Unreleased]`. Happy to add more or pivot in another direction if the team wants.

---

> [!NOTE]
> **Medium Risk**
> Changes workflow task failure routing for codec errors on core payload paths; mis-tagged or overly broad use of the marker could cause extra task retries instead of failing fast, but only opt-in **`WorkflowTaskFailureError`** from codecs is affected.
> 
> **Overview**
> Adds **`converter.WorkflowTaskFailureError`** (via **`NewWorkflowTaskFailureError`**) so a **`PayloadCodec`** can opt into failing the **current Workflow Task** instead of the **Workflow Execution** when encode/decode fails on selected workflow-side paths—letting the server retry the task after transient codec issues (e.g. KMS throttling).
> 
> **`CodecDataConverter`** tags only codec-boundary failures with an internal **`codecerror`** marker so workflow code cannot spoof the behavior. The worker’s **`applyWorkflowPanicPolicy`** recognizes tagged errors **before** **`WorkflowPanicPolicy`**, returns them as task failures (not **`FailWorkflowExecution`**), and avoids workflow-panic logging; encode-side markers are unwrapped from **`workflowPanicError`**. Side effect / mutable side effect summary encoding preserves the marker by re-panicking instead of replacing it with a generic string panic.
> 
> Honored paths include workflow input decode, activity/child result decode, activity/child arg encode, side-effect summary encode, and local-activity marker encode; unmarked codec errors and markers returned directly from workflow code keep prior execution-failure behavior. Signal/update/query paths are unchanged. Unit and integration tests cover retry-then-complete scenarios; **CHANGELOG** documents the feature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79874fdd9640aa3ddb7a4e2207a2e16dd22a02c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/internal_task_handlers.go
```diff
@@ -1306,6 +1306,17 @@ func (w *workflowExecutionContextImpl) applyWorkflowPanicPolicy(workflowTask *wo
...
 
 	if workflowError == nil && w.err != nil {
+		// A PayloadCodec running workflow-side code (e.g. decoding workflow input
+		// or an activity result) may return a *converter.WorkflowTaskFailureError
+		// to request that the current Workflow Task fail while the Workflow
+		// Execution stays open. Honor it here, before the WorkflowPanicPolicy
+		// switch below, so the behavior is independent of that policy and is not
+		// classified or logged as a workflow panic. Returning the error converts
+		// it into a WorkflowTaskFailed which the server retries.
+		var codecTaskErr *converter.WorkflowTaskFailureError
```
yuandrew: I think there's a gap in the `Encode` behavior here. The new API says marked `Encode` and `Decode` errors are handled independently of `WorkflowPanicPolicy`, but workflow-side encoded errors, such as activity argument encoding, ar panicked and wrapped in a `workflowPanicError`.

Can we recognize a marker contained inside `workflowPanicError` and add coverage for marked encode failures under both policies? I think users will expect the marker's behavior to be consistent across `Encode` and `Decode`, rather than depending on how a particular SDK call site propagates the codec error.

converter/errors.go
```diff
@@ -28,3 +28,51 @@ var (
...
+// retry should remain the primary recovery mechanism.
+//
+// The marker is honored independently of the worker's WorkflowPanicPolicy and
+// is not classified or logged as a workflow panic. When returned from a codec
+// running in an Activity worker or on the client, it behaves as an ordinary
+// returned error and does not trigger any special routing.
+//
+// The wrapped Cause is preserved through errors.Is and errors.As so failure
+// converters and observability tooling still see the original error.
+type WorkflowTaskFailureError struct {
```
yuandrew, line 48: Can we shorten this documentation and make the supported scope more precise? It currently reads as though any marked `Encode` or `Decode` error during WFT processing will fail the task, but that isn't true for every codec call. 

For example, signal decoding still treats the marker as a corrupted signal and drops it, while update argument decoding rejects the update. Since the codec explicitly marked the error as transient and requested a WFT retry, I think those paths should honor it rather than permanently dropping or rejecting the input. Queries can remain query failures because they are read-only requests with their own failure response.

Once the behavior is consistent, I think the public contract only needs to say:
- it requests WFT failure during workflow-side codec processing
- the WF execution remains open
- it is independent of `WorkflowPanicPolicy`
- outside WFT processing it is an ordinary error
- persistent failures may cause repeated WFT failures

I think this can fit in 2 short paragraphs, with the bounded-retry recommendation on the constructor.

internal/internal_task_handlers_test.go
```diff
@@ -1158,6 +1158,143 @@ func (t *TaskHandlersTestSuite) TestWorkflowTask_WorkflowPanics() {
...
+			ActivityId:   "0",
+			ActivityType: &commonpb.ActivityType{Name: "Greeter_Activity"},
+			TaskQueue:    &taskqueuepb.TaskQueue{Name: taskQueue},
+		}),
+		createTestEventActivityTaskStarted(6, &historypb.ActivityTaskStartedEventAttributes{}),
+		createTestEventActivityTaskCompleted(7, &historypb.ActivityTaskCompletedEventAttributes{ScheduledEventId: 5, Result: activityResult}),
+		createTestEventWorkflowTaskStarted(8),
+	}
+	params := t.getTestWorkerExecutionParams()
+	params.WorkflowPanicPolicy = BlockWorkflow
```
yuandrew: Can we run this activity-result test under both BlockWorkflow and FailWorkflow, like the workflow-input test does?

The feature’s contract says the marker behaves identically under both policies, and this protects the important Future.Get path from accidentally falling through to the normal panic-policy handling.

converter/errors.go
```diff
@@ -28,3 +28,51 @@ var (
...
+	// Cause is the underlying, transient error that should be treated as a
+	// task-level failure. It must not be nil.
+	Cause error
```
yuandrew: Could we keep Cause private? Unwrap already exposes it through the standard errors.Is, errors.As, and errors.Unwrap APIs.

Leaving it exported permanently commits us to this field name and type, and it lets callers construct or mutate invalid state. A constructor plus Unwrap seems like enough API for codec authors.

Also, the comment says Cause must not be nil, but the implementation and test intentionally support a nil cause. We should make that contract consistent one way or the other.

internal/internal_task_handlers.go
yuandrew: Can we instead tag errors at the PayloadCodec boundary and only honor that codec-originated form? Checking the final workflow error makes this a general WFT-failure API. Direct workflow returns should remain ordinary errors.

whitemanthedj: Good catch, fixed. Encode-side markers (e.g. activity-arg or child-workflow-arg encoding) get panicked and wrapped in workflowPanicError, so recognizing them needs an unwrap. The marker is now tagged at the codec boundary, so by the time it panics it's already the codec-originated form; the honoring path unwraps workflowPanicError and recognizes it, routes it as a WFT failure, and does not log or classify it as a panic. So Encode and Decode behave the same

converter/errors.go
whitemanthedj: Split this into two.

Docs: shortened the type doc to two short paragraphs with the contract bullets you listed (requests WFT failure during workflow-side codec processing, execution stays open, independent of WorkflowPanicPolicy, ordinary error outside WFT processing, persistent failures repeat the WFT failure), and moved the bounded-retry recommendation onto the constructor.

Signals and updates: agreed that if a codec explicitly marked the error transient, dropping the signal or rejecting the update is the wrong outcome. That's a behavior change in the signal/update delivery paths though (routing them into a WFT failure instead of the current swallow/reject), and it's determinism-sensitive enough that it wants its own replay coverage, so I'd rather not fold it into this PR. I'll take it as a focused follow-up right after this lands, keeping queries as query failures like you said. Does landing this one for the encode/decode paths and handling signals/updates in the follow-up work for you?

internal/internal_task_handlers_test.go
whitemanthedj: Done. Pulled the activity-result-decode setup into a helper and it now runs under both BlockWorkflow and FailWorkflow, matching the input-decode test. The new encode-side and child-workflow tests do the same.

converter/errors.go
whitemanthedj: Done. Cause is now unexported, with the constructor plus Unwrap as the only way in and out. I kept nil support (a nil cause is valid) and made the contract consistent, the doc, impl, and test now agree instead of the doc claiming it must be non-nil. errors.Is/As/Unwrap still reach the cause.

internal/internal_task_handlers.go
whitemanthedj: Reworked to do exactly this. Instead of inspecting the final workflow error, the marker is now tagged at the PayloadCodec boundary: CodecDataConverter.encode/decode wrap a returned WorkflowTaskFailureError in an internal codec-originated form, and the worker only honors that. A WorkflowTaskFailureError returned directly by workflow code never passes through the codec, so it stays an ordinary error and fails the execution, added a regression test for that. Because it's one boundary that every payload op flows through, it covers input, result, activity args/result, and child-workflow args/result consistently rather than a hand-picked set of call sites.

On API surface (re: your Cause note on the other thread): the tag is an unexported wrapper in the converter package. The worker detects it via a method on that unexported type, so nothing new shows up in the public API, and direct-return vs codec-origin stays cleanly separated.

internal/internal_task_handlers.go
```diff
@@ -1306,16 +1306,16 @@ func (w *workflowExecutionContextImpl) applyWorkflowPanicPolicy(workflowTask *wo
...
-			return nil, w.err
+		// A PayloadCodec running workflow-side code (encoding activity arguments,
+		// decoding workflow input or an activity result, etc.) may return a
+		// *converter.WorkflowTaskFailureError to request that the current Workflow
+		// Task fail while the Workflow Execution stays open. It is tagged at the
+		// codec boundary so only codec-originated markers are honored here, before
+		// the WorkflowPanicPolicy switch, making the behavior identical under both
+		// policies and keeping it out of the workflow-panic logging path. Returning
+		// the error becomes a WorkflowTaskFailed which the server retries.
+		if codecErr, ok := codecWorkflowTaskFailureFrom(w.err); ok {
```
yuandrew, line 1317: When `SideEffect` or `MutableSideEffect` summary encoding fails, the marker is lost. I believe we need to update `(wc *workflowEnvironmentImpl) SideEffect` to check for the new marker and panic(err) instead, that way existing behavior is preserved, but this new behavior is consistent.

Something like 
```
if _, ok := codecWorkflowTaskFailureFrom(err); ok {
    panic(err)
}
```
above the existing 
```
if err != nil {
	panic(fmt.Sprintf("failed to build user metadata for side effect: %v", err))
}
```

converter/errors.go
yuandrew: Yeah, I agree this should probably deserve its own PR/change. Can we update tehe docs to reflect the current scope? "Workflow-side payload processing" overstates the behavior because signal, update, and query decoding retain their existing handling.

This will likely include cross-SDK design and replay compatibility analysis, like you mentioned. Would you be okay creating this issue to track?

internal/internal_task_handlers.go
yuandrew: Even though the wrapper is unexported, Go interfaces are structural. Any error implementing `RequestsWorkflowTaskFailure() bool` satisfies this interface, so `errors.As` does not prove codec origin. The Boolean is also never checked, so even false matches.

Hmm, what if we use a private concrete tag in `internal/codecerror`? For example:
```
package codecerror

import "errors"

type tagged struct {
    err error
}

func (e *tagged) Error() string {
    return e.err.Error()
}

func (e *tagged) Unwrap() error {
    return e.err
}

func Tag(err error) error {
    return &tagged{err: err}
}

func Extract(err error) (error, bool) {
    var taggedErr *tagged
    if !errors.As(err, &taggedErr) {
	        return nil, false
	    }

    return taggedErr.err, true
}
```
The converter would call Tag, and the worker would call Extract. This enforces codec origin without adding public API. LMK what you think

internal/internal_task_handlers.go
whitemanthedj: Good catch, fixed. Added the check at both encode sites: SideEffect and MutableSideEffect (via recordMutableSideEffect). If the buildUserMetadata error carries the marker we panic(err) with the raw error before the existing formatted panic, so it flows through workflowPanicError and the honoring path recognizes it as a WFT failure. Non-marker errors keep the old formatted panic, so existing behavior is preserved.

Added unit tests for both, each under BlockWorkflow and FailWorkflow. They use a content-scoped codec that fails only the summary payload, so the side effect's own value encode still passes and the summary path is what's exercised (without the fix the string panic drops the marker and the assertions fail).

internal/internal_task_handlers.go
whitemanthedj: Agreed on both points; I went with your codecerror approach. New internal/codecerror package with a private concrete tag type plus Tag/Extract. The CodecDataConverter encode/decode paths call Tag at the boundary and the worker calls Extract. Removed the old interface and the RequestsWorkflowTaskFailure method as well.

Also added a codecerror test asserting an error that implements the old RequestsWorkflowTaskFailure() bool is not extracted, and the direct-return regression test still holds (a marker returned straight from workflow code never passes through the codec, so it stays an ordinary error). No import cycle: codecerror imports nothing but errors, and both converter and internal import it

converter/errors.go
whitemanthedj: Added on the docs and dropped the "workflow-side payload processing" framing that overstated it. Now the narrowed scope includes: workflow-input decode, activity/child-workflow result decode via a Future, activity/child-workflow arg encode, and side-effect summary encode (with an explicit note that signal decoding, update-argument decoding, and query decoding keep their existing behavior)

Agreed for 2 it deserves a tracked follow-up. Given it pulls in cross-SDK design and replay compatibility questions, that feels like a call for the team to scope and prioritize? I'd feel more comfortable if the team could scope it out more thoroughly following this, does that sound ok?

CHANGELOG.md
```diff
@@ -20,6 +20,24 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 
+### Added
+
+- `converter.WorkflowTaskFailureError`: an opt-in error a `PayloadCodec` can return
```
yuandrew, line 25: Can we reduce this to one or two sentences and leave the detailed scope, exclusions, and retry guidance in the API docs? Changelog entries should summarize the user-facing change rather than duplicate its full contract.

For example, 
> - Added `converter.WorkflowTaskFailureError`, allowing a `PayloadCodec` to request a Workflow Task retry for supported workflow-side encoding or decoding failures without failing the Workflow Execution.

internal/codecerror/codecerror.go
```diff
@@ -0,0 +1,39 @@
...
+func Tag(err error) error {
+	return &tag{err: err}
+}
+
+// Extract reports whether err carries the codec origin tag and, when it does,
+// returns the tagged error so the caller can surface the underlying cause.
+func Extract(err error) (error, bool) {
+	var t *tag
+	if errors.As(err, &t) {
+		return t.err, true
```
yuandrew: Can we preserve the outer error for non-panic paths and unwrap only `workflowPanicError`? Extract currently drops SDK context added after the codec boundary, so workflow-input failures report only the codec cause instead of identifying the failed operation and workflow type. This makes repeated WFT failures harder to diagnose from history and logs.

whitemanthedj: Replaced Extract with a presence check Has(err) bool and left the error in place. Decode paths now return the outer error unchanged so the SDK context (operation + workflow type) survives; only the workflowPanicError wrapper is unwrapped, for encode-side panics. errors.Is/errors.As still reach the marker through the tag

CHANGELOG.md
whitemanthedj: Done. Cut it to two sentences and pointed at the WorkflowTaskFailureError docs

internal/internal_task_handlers.go
```diff
@@ -1306,6 +1306,17 @@ func (w *workflowExecutionContextImpl) applyWorkflowPanicPolicy(workflowTask *wo
...
 	if workflowError == nil && w.err != nil {
+		// A PayloadCodec running workflow-side code (encoding activity arguments,
+		// decoding workflow input or an activity result, etc.) may return a
+		// *converter.WorkflowTaskFailureError to request that the current Workflow
+		// Task fail while the Workflow Execution stays open. It is tagged at the
+		// codec boundary so only codec-originated markers are honored here, before
+		// the WorkflowPanicPolicy switch, making the behavior identical under both
+		// policies and keeping it out of the workflow-panic logging path. Returning
+		// the error becomes a WorkflowTaskFailed which the server retries.
+		if codecErr, ok := codecWorkflowTaskFailureFrom(w.err); ok {
```
yuandrew, line 1317: This only checks w.err, but ProcessLocalActivityResult passes marker-data encoding errors as workflowError. A tagged codec failure there bypasses this branch; under FailWorkflow it permanently fails the execution and is logged as a workflow panic. Can we check workflowError for the marker before applying the panic policy and cover local-activity marker encoding under both policies?

internal/internal_task_handlers_test.go
```diff
@@ -1158,6 +1223,324 @@ func (t *TaskHandlersTestSuite) TestWorkflowTask_WorkflowPanics() {
...
+		request, err := t.processInputDecodeFailure(policy, converter.NewWorkflowTaskFailureError(cause))
+
+		// The Workflow Task fails (the returned error is turned into a
+		// WorkflowTaskFailed the server retries) and no completion request is
+		// produced, so the Workflow Execution stays open.
+		t.Error(err, "policy %v", policy)
+		t.Nil(request, "policy %v", policy)
+		var marker *converter.WorkflowTaskFailureError
+		t.True(errors.As(err, &marker), "policy %v", policy)
+		t.True(errors.Is(err, cause), "policy %v", policy)
```
yuandrew, line 1327: Can we assert the returned error contains the workflow-input decode context and workflow type? The current errors.As/Is checks passed even before the diagnostic wrapper fix.

internal/internal_task_handlers.go
whitemanthedj: Yea true. ProcessLocalActivityResult returns the marker-encode error as workflowError, not w.err, so it skipped the guard and hit the panic policy, permanently failing the execution under FailWorkflow. Fixed by also checking workflowError before the panic switch:

```
if codecErr, ok := codecWorkflowTaskFailureFrom(workflowError); ok {
     return nil, codecErr
 }
```

Safe on nil and won't match a non-codec panic. Added TestWorkflowTask_CodecWorkflowTaskFailureError_LocalActivityMarkerEncode (drives the real LA poller, both policies); confirmed it fails without the guard.

internal/internal_task_handlers_test.go
whitemanthedj: Agreed, errors.As/Is passed even before the round-3 wrapper fix. Tightened InputDecode to pin the wrapped context:

t.ErrorContains(err, "unable to decode the workflow function input payload", "policy %v", policy)
t.ErrorContains(err, "HelloWorld_Workflow", "policy %v", policy)

yuandrew:
bugbot run

internal/internal_task_handlers.go
```diff
@@ -1306,11 +1306,28 @@ func (w *workflowExecutionContextImpl) applyWorkflowPanicPolicy(workflowTask *wo
...
```
yuandrew, line 1317: This only recognizes the marker if it remains the workflow’s final error. If workflow code handles a marked Future.Get error, it never reaches this check and the Workflow Task can complete. Can we detect the marker and call panic(err) inside Get, then cover handled Get errors under both policies?

whitemanthedj: Yes good catch, updated to identify that error and then panic

PR #2539, opened 2026-08-11
Add envconfig support to integration test harness

THardy98:
## What was changed

Add `-envconfig` flag to integration test suite, which enables connection through `contrib/envconfig`.

In the existing harness path, `Config` remains the source for clients. Existing configuration callbacks work as before. In `envconfig` mode, the loaded `client.Options` are the sole client source, so API keys, TLS settings, gRPC metadata, and authority are preserved; `Config` is only a lifecycle projection used by existing test setup.

The existing two-test Temporal Cloud job now uses this path with the standard `envconfig` certificate-data variables.

## Why?

Add `envconfig` backed harness to the integration test suite, allowing tests to run against arbitrary server environments.

## Checklist

1. Part of temporalio/features#851.

2. How was this tested:

- `go test -run '^TestConfigAndClientSuiteBase' -count=1 .`
- `go test -run '^$' ./...`
- `go run . integration-test -envconfig -packages . -run '^TestConfigAndClientSuiteBaseFromEnvConfig$'`
- `go run . integration-test -dev-server -p 1 -packages . -run 'TestIntegrationSuite/TestBasic$'`
- Verified that combining `-dev-server` and `-envconfig` fails with a clear error.

3. Any docs updates needed?

No. This changes internal SDK integration-test infrastructure and its existing Cloud CI job.

### DIFF
--- a/test/test_utils_test.go
+++ b/test/test_utils_test.go
@@ -38,25 +38,15 @@ type (
 		TLS                     *tls.Config
 	}
 	// context.WithValue need this type instead of basic type string to avoid lint error
-	contextKey string
-	// testEnvironment configures one test server target. Its invariants are:
-	//   - When envConfigClientOptions is nil, config is the source for clients.
-	//   - When envConfigClientOptions is non-nil, it is the only source for clients.
-	//     config is a read-only projection used by test setup and lifecycle logic;
-	//     the two representations are not synchronized after construction.
-	//   - Per-client configuration is applied to a value copy of the selected
-	//     source; callbacks must not mutate reference-valued members inherited
-	//     from it.
-	testEnvironment struct {
-		config                 Config
-		envConfigClientOptions *client.Options
-	}
-	// ConfigAndClientSuiteBase owns a resolved test environment and any shared
-	// client created from that environment.
+	contextKey               string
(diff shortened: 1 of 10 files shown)


test/test_utils_test.go
```diff
@@ -248,22 +318,25 @@ func (ts *ConfigAndClientSuiteBase) InitClient(clientOpts ...ConfigureClientOpti
...
-		WorkerHeartbeatInterval: -1,
+	return client.Dial(ts.newClientOptions(clientOpts...))
+}
+
+func (ts *ConfigAndClientSuiteBase) newClientOptions(clientOpts ...ConfigureClientOptions) client.Options {
+	var options client.Options
+	if ts.envConfigClientOptions == nil {
+		options = ts.config.toClientOptions()
+	} else {
+		options = *ts.envConfigClientOptions
```
yuandrew, line 353: Should all test clients start from `envConfigClientOptions?` Some still dial from `ts.config` or `NewConfig()`, so `-envconfig` loses credentials, headers, authority, and TLSDisabled. #2544 fixes most of these, but not the standalone Nexus tests

yuandrew (approved):

PR #2544, opened 2026-08-11
Refactor integration test client creation

dplyukhin:
## What was changed

Replaced `client.Dial` in integration tests with calls to `ConfigAndClientSuiteBase.newDefaultClient` and `ConfigAndClientSuiteBase.newIntegrationTestClient`.

## Why?

Ensure all clients use the configured connection info instead of local defaults. Useful for cloud CI.

## Checklist

1. Closes  nothing

2. How was this tested:

Re-ran existing tests and confirmed they still work.

3. Any docs updates needed?

No.

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -175,22 +175,18 @@ func (ts *IntegrationTestSuite) SetupTest() {
 
 	var err error
 	trafficController := test.NewSimpleTrafficController()
-	ts.client, err = client.Dial(client.Options{
-		HostPort:  ts.config.ServiceAddr,
-		Namespace: ts.config.Namespace,
-		Logger:    ilog.NewDefaultLogger(),
-		ContextPropagators: []workflow.ContextPropagator{
+	ts.client, err = ts.newDefaultClient(func(options *client.Options) {
+		options.WorkerHeartbeatInterval = -1
+		options.ContextPropagators = []workflow.ContextPropagator{
 			NewKeysPropagator([]string{testContextKey1}),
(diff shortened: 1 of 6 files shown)


test/integration_test.go
```diff
@@ -8830,12 +8803,16 @@ func (ts *IntegrationTestSuite) TestClientPlugin() {
...
-	_, err := client.Dial(client.Options{Namespace: ts.config.Namespace, Plugins: []client.Plugin{plugin}})
+	_, err := ts.newDefaultClient(func(options *client.Options) {
+		options.Plugins = []client.Plugin{plugin}
+	})
 	ts.ErrorContains(err, "intentional client error")
 
 	// Now succeed dial and run simple workflow
 	plugin = &clientPluginForTest{ts: ts}
-	cl, err := client.Dial(client.Options{Namespace: ts.config.Namespace, Plugins: []client.Plugin{plugin}})
+	cl, err := ts.newDefaultClient(func(options *client.Options) {
```
yuandrew, line 8819: This plugin is specifically meant to supply `HostPort` and `TLS`, but `newDefaultClient` populates both from `ts.config`. This test now passes even if those plugin assignments regress, including in cloud CI where they were previously exercised. Should we still directly call `client.Dial` for this test, or clear these fields before installing the plugin so the plugin remains responsible for them?

dplyukhin: Nice catch!! Modified it so we write garbage values with `newDefaultClient` and then check if there's a change.

yuandrew (approved):

PR #2549, opened 2026-08-12
Propagate data converter errors from Nexus input deserialization

bergundy:
The Nexus SDK no longer replaces input deserialization errors with a
generic BAD_REQUEST handler error, so the error the data converter
returns now reaches the caller.

Classify it in payloadSerializer.Deserialize: an ApplicationError or a
HandlerError already carries an intent for the caller and is passed
through to the regular error handling path, anything else means the
input could not be read and becomes a BAD_REQUEST HandlerError that
keeps the original error as its cause.

NOTE: Unlike the [Java change](https://github.com/temporalio/sdk-java/pull/3002), this isn't breaking because the default behavior was to already return a BAD_REQUEST.

⚠️ DO NOT MERGE ⚠️ until `go.mod` references a released version of nexus-rpc/sdk-go.


internal/internal_nexus_task_handler.go
```diff
@@ -610,7 +610,26 @@ type payloadSerializer struct {
...
+	}
+	// The Nexus SDK propagates serializer errors as-is, so errors that already carry an intent for the caller are
+	// passed through and converted by the regular error handling path. Anything else means the input could not be
+	// read and is reported as a bad request.
+	// Not using errors.As to be consistent ApplicationError checking with the rest of the SDK.
+	if _, ok := err.(*ApplicationError); ok {
+		return err
+	}
+	var handlerErr *nexus.HandlerError
+	if errors.As(err, &handlerErr) {
```
Quinn-With-Two-Ns, line 639: Why do `errors.As` for HandlerError?

internal/internal_nexus_task_handler.go
```diff
@@ -610,7 +610,26 @@ type payloadSerializer struct {
...
+		return err
+	}
+	return &nexus.HandlerError{
+		Type:    nexus.HandlerErrorTypeBadRequest,
+		Message: fmt.Sprintf("cannot deserialize operation input: %v", err),
```
Quinn-With-Two-Ns: Aren't we basically including the error twice here?

bergundy: Ah true

internal/internal_nexus_task_handler.go
bergundy: It's consistent with what we do in other places in this file

internal/internal_nexus_task_handler.go
bergundy: Fixed.

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns (approved):

PR #2552, opened 2026-08-13
Fix activity/workflow alias collision for anonymous functions (#2141)

### LINKED ISSUE #2141: Activity Alias Collision in Test Environment When Using Anonymous Functions
## Summary

The test environment's activity alias registry uses short function names (e.g., `func1`) as keys, which causes collisions when multiple packages register activities using anonymous functions. This results in local activities being incorrectly routed to unrelated registered activity mocks, causing panics due to signature mismatches.

## Affected Version

- `go.temporal.io/sdk` v1.38.0 (and likely earlier versions)

### Minimal Test Case Demonstrating the Bug

```go
package alias_collision_test

import (
	"context"
	"testing"
	"time"

	"github.com/stretchr/testify/suite"
	"go.temporal.io/sdk/activity"
	"go.temporal.io/sdk/testsuite"
	"go.temporal.io/sdk/workflow"
)

type AliasCollisionSuite struct {
	suite.Suite
	testsuite.WorkflowTestSuite
}

func TestAliasCollisionSuite(t *testing.T) {
	suite.Run(t, new(AliasCollisionSuite))
}

// SomeActivityRequest is a request type for a registered activity
type SomeActivityRequest struct {
	Value string
}

// SomeActivityResponse is a response type for a registered activity
type SomeActivityResponse struct {
	Result string
}

// TestAliasCollision_Fails demonstrates the bug where anonymous functions
// with the same short name collide in the activity alias registry.
func (s *AliasCollisionSuite) TestAliasCollision_Fails() {

	// Register an activity using an anonymous function.
	// This anonymous function gets a short name like "func1" in the registry.
	env.RegisterActivityWithOptions(
		func(ctx context.Context, req *SomeActivityRequest) (*SomeActivityResponse, error) {
			return &SomeActivityResponse{Result: "from registered activity"}, nil
		},
		activity.RegisterOptions{Name: "SomeActivity"},
	)

	// This workflow executes a LOCAL activity using an INLINE anonymous function.
	// The inline function also gets a short name like "func1".
	// Due to the collision, the test environment incorrectly tries to call
	// the registered "SomeActivity" with the wrong arguments.
	testWorkflow := func(ctx workflow.Context) (string, error) {
		lao := workflow.LocalActivityOptions{
			ScheduleToCloseTimeout: 10 * time.Second,
		}
		ctx = workflow.WithLocalActivityOptions(ctx, lao)

		var result string
		// This inline anonymous function takes only context and returns (string, error)
		// but the registry thinks it should call SomeActivity which takes
		// (context.Context, *SomeActivityRequest) -> (*SomeActivityResponse, error)
		err := workflow.ExecuteLocalActivity(ctx, func(ctx context.Context) (string, error) {
			return "from inline local activity", nil
		}).Get(ctx, &result)

		return result, err
	}

	env.ExecuteWorkflow(testWorkflow)

	require.True(s.T(), env.IsWorkflowCompleted())
	require.NoError(s.T(), env.GetWorkflowError())

	var result string
	require.NoError(s.T(), env.GetWorkflowResult(&result))
	require.Equal(s.T(), "from inline local activity", result)
}

// TestAliasCollision_Workaround demonstrates the workaround: use named functions
// instead of anonymous functions when registering activities.
func (s *AliasCollisionSuite) TestAliasCollision_Workaround() {

	// WORKAROUND: Use a named function instead of an anonymous function.
	// Named functions have unique full names that won't collide.
	env.RegisterActivityWithOptions(
		namedActivityMock, // This gets a unique name like "alias_collision_test.namedActivityMock"
		activity.RegisterOptions{Name: "SomeActivity"},
	)

	testWorkflow := func(ctx workflow.Context) (string, error) {
		lao := workflow.LocalActivityOptions{
			ScheduleToCloseTimeout: 10 * time.Second,
		}
		ctx = workflow.WithLocalActivityOptions(ctx, lao)

		var result string
		err := workflow.ExecuteLocalActivity(ctx, func(ctx context.Context) (string, error) {
			return "from inline local activity", nil
		}).Get(ctx, &result)

		return result, err
	}

	env.ExecuteWorkflow(testWorkflow)

	require.True(s.T(), env.IsWorkflowCompleted())
	require.NoError(s.T(), env.GetWorkflowError())

	var result string
	require.NoError(s.T(), env.GetWorkflowResult(&result))
	require.Equal(s.T(), "from inline local activity", result)
}

// namedActivityMock is a named function used as a workaround for the alias collision bug.
func namedActivityMock(ctx context.Context, req *SomeActivityRequest) (*SomeActivityResponse, error) {
	return &SomeActivityResponse{Result: "from registered activity"}, nil
}
```

### Expected Behavior

Both tests should pass. The inline local activity function should execute independently of the registered activity mock, as they are completely separate functions with different purposes.

### Actual Behavior

`TestAliasCollision_Fails` panics with:
```
reflect: Call with too few input arguments
```

The test environment incorrectly routes the inline local activity to the registered "SomeActivity" mock because both anonymous functions have the same short name (`func1`).

### 1. Short function names used as alias keys

In `internal/internal_worker.go`, the `getFunctionName` function returns only the short name:

```go
func getFunctionName(i interface{}) (name string, isMethod bool) {
    fullName := runtime.FuncForPC(reflect.ValueOf(i).Pointer()).Name()
    elements := strings.Split(fullName, ".")
    shortName := elements[len(elements)-1]  // Returns "func1", "func2", etc.
    return strings.TrimSuffix(shortName, "-fm"), isMethod
}
```

### 2. Alias stored using short name

In `internal/internal_worker.go` line ~730:
```go
if len(alias) > 0 && r.activityAliasMap != nil {
    r.activityAliasMap[fnName] = alias  // fnName is "func1"
}
```

### 3. Test environment looks up alias for local activities

In `internal/internal_workflow_testsuite.go` line ~1559:
```go
func (env *testWorkflowEnvironmentImpl) ExecuteLocalActivity(...) {
    ae := &activityExecutor{name: getActivityFunctionName(env.registry, params.ActivityFn), fn: params.ActivityFn}
    if at, _ := getValidatedActivityFunction(params.ActivityFn, params.InputArgs, env.registry); at != nil {
        // This finds the wrong alias because both functions have short name "func1"
        ae.name = at.Name
    }
    // ...
}
```

### 4. Mock lookup uses the wrong name

In `internal/internal_workflow_testsuite.go` line ~1817:
```go
m := &mockWrapper{env: a.env, name: a.name, fn: a.fn, isWorkflow: false}
if mockRet := m.getActivityMockReturnWithActualArgs(ctx, inputArgs); mockRet != nil {
    // Finds the mock for "SomeActivity" when it shouldn't
}
```

### Option 1: Use full function name as alias key (Recommended)

Change the alias map to use full function names instead of short names:

```go
func (r *registry) RegisterActivityWithOptions(a interface{}, options RegisterActivityOptions) {
    // ... existing code ...
    
    // Use full function name as alias key
    fullName := runtime.FuncForPC(reflect.ValueOf(af).Pointer()).Name()
    
    if len(alias) > 0 && r.activityAliasMap != nil {
        r.activityAliasMap[fullName] = alias
    }
}

func (r *registry) getActivityAlias(fnName string) (string, bool) {
    // fnName should also be the full function name
    alias, ok := r.activityAliasMap[fnName]
    return alias, ok
}
```

This requires updating all callers of `getActivityAlias` to pass the full function name.

### Option 2: Skip alias lookup for anonymous functions in test environment

In `ExecuteLocalActivity`, detect anonymous functions and skip the alias lookup:

```go
func (env *testWorkflowEnvironmentImpl) ExecuteLocalActivity(...) {
    ae := &activityExecutor{name: getActivityFunctionName(env.registry, params.ActivityFn), fn: params.ActivityFn}
    
    // Only look up alias if this is not an anonymous function
    fullName := runtime.FuncForPC(reflect.ValueOf(params.ActivityFn).Pointer()).Name()
    if !isAnonymousFunction(fullName) {
        if at, _ := getValidatedActivityFunction(params.ActivityFn, params.InputArgs, env.registry); at != nil {
            ae.name = at.Name
        }
    }
    // ...
}

func isAnonymousFunction(fullName string) bool {
    // Anonymous functions have names ending in ".funcN" where N is a number
    parts := strings.Split(fullName, ".")
    if len(parts) == 0 {
        return false
    }
    lastPart := parts[len(parts)-1]
    return strings.HasPrefix(lastPart, "func") && len(lastPart) > 4
}
```

## Real-World Impact

This bug affects users of [protoc-gen-go-temporal](https://github.com/cludden/protoc-gen-go-temporal), which generates code that uses anonymous functions for expression evaluation in workflow ID generation:

```go
// Generated code in lifecycle_temporal.pb.go
func (o *ServerDrainChildOptions) Build(ctx workflow.Context, req protoreflect.Message) (workflow.ChildWorkflowOptions, error) {
    // ...
    if err := workflow.ExecuteLocalActivity(ctx, func(ctx context.Context) (string, error) {
        id, err := expression.EvalExpression(ServerDrainIdexpression, req)
        // ...
    }).Get(ctx, &opts.WorkflowID); err != nil {
        // ...
    }
    // ...
}
```

When users register activities with anonymous functions in their tests, these collide with the generated expression evaluation functions.

## Workaround

Until this is fixed, users should use **named functions** instead of anonymous functions when calling `RegisterActivityWithOptions` or generated `Register*Activity` functions:

```go
// Instead of this (FAILS):
env.RegisterActivityWithOptions(
    func(ctx context.Context, req *MyRequest) (*MyResponse, error) {
        return &MyResponse{}, nil
    },
    activity.RegisterOptions{Name: "MyActivity"},
)

// Do this (WORKS):
func myActivityMock(ctx context.Context, req *MyRequest) (*MyResponse, error) {
    return &MyResponse{}, nil
}

env.RegisterActivityWithOptions(myActivityMock, activity.RegisterOptions{Name: "MyActivity"})
```

## Environment

- Go version: 1.25.5
- OS: macOS (also affects Linux)
- Temporal SDK version: v1.38.0

jaideeppyne:
## What was changed

`getFunctionName` collapses a function's runtime name to its last dot-separated element. For a closure the Go compiler emits `<enclosing>.funcN`, and that numbering restarts per package, so **every package's first closure is `func1`**. The bare `func1` is then used as the key in the test environment's activity and workflow **alias maps**.

The result: an activity registered in one package with an anonymous function and an explicit `Name` stores `aliasMap["func1"] = "SomeActivity"`. A *different* closure in another package (e.g. an inline local activity) also resolves to `func1`, matches that alias, and gets routed to the wrong registration — panicking on the signature mismatch. This is the collision reported in #2141.

The fix keeps the fully-qualified name for closures (detected via the compiler's `.funcN` marker), giving each closure a stable, unique identity across packages. Named functions and methods are unaffected — they still resolve to their short names (`Foo`, `RunActivity`, etc.).

## Why

Closes #2141. The alias maps need a key that uniquely identifies a function; the short name does not for closures.

## Checklist

1. Added `TestGetFunctionName_ClosuresKeepQualifiedName` in `internal/internal_worker_test.go`. It fails before the change (both closures resolve to `func1`) and passes after (each retains its enclosing identity), and asserts named functions/methods are unchanged.
2. `go test ./internal/` passes (full package, no regressions); `gofmt` and `go vet ./internal/` are clean.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"os"
 	"reflect"
+	"regexp"
 	"runtime"
 	"sort"
 	"strconv"
@@ -2724,6 +2725,10 @@ func isError(inType reflect.Type) bool {
 	return inType != nil && inType.Implements(errorElem)
 }
 
+// closureNamePattern matches the compiler-generated name of a Go closure, e.g.
+// "pkg.Outer.func1" or a nested "pkg.Outer.func1.2".
+var closureNamePattern = regexp.MustCompile(`\.func\d`)
+
 func getFunctionName(i interface{}) (name string, isMethod bool) {
 	if fullName, ok := i.(string); ok {
 		return fullName, false
@@ -2732,6 +2737,14 @@ func getFunctionName(i interface{}) (name string, isMethod bool) {
 	// Full function name that has a struct pointer receiver has the following format
 	// <prefix>.(*<type>).<function>
 	isMethod = strings.ContainsAny(fullName, "*")
+	// Closures are named "<enclosing>.funcN" by the Go runtime, so their short
+	// name collapses every closure in a scope to "func1", "func2", ... These
+	// collide in the registry's alias maps and can misroute one activity to
+	// another's registration (see temporalio/sdk-go#2141). Keep the
+	// fully-qualified name so each closure has a stable, unique identity.
+	if closureNamePattern.MatchString(fullName) {
+		return fullName, isMethod
+	}
 	elements := strings.Split(fullName, ".")
 	shortName := elements[len(elements)-1]
 	// This allows to call activities by method pointer
--- a/internal/internal_worker_test.go
+++ b/internal/internal_worker_test.go
@@ -3665,3 +3665,44 @@ func (s *internalWorkerTestSuite) TestRegisterMultipleDynamicWorkflow() {
 	require.Error(s.T(), err)
 	require.Contains(s.T(), err.Error(), "dynamic activity already registered")
 }
+
+// firstClosureA and firstClosureB each return their enclosing function's first
+// closure. The Go runtime names both "<pkg>.firstClosureX.func1", so their short
+// names collide as "func1" — the collision behind temporalio/sdk-go#2141.
+func firstClosureA() func(context.Context) error {
+	return func(context.Context) error { return nil }
+}
+
+func firstClosureB() func(context.Context) error {
+	return func(context.Context) error { return nil }
+}
+
+type getFunctionNameActivities struct{}
+
+func (*getFunctionNameActivities) NamedActivity(context.Context) error { return nil }
+
+func TestGetFunctionName_ClosuresKeepQualifiedName(t *testing.T) {
+	// The Go compiler numbers closures per package ("func1", "func2", ...), so
+	// each package restarts at "func1" and the bare short name collides across
+	// packages. That misroutes activities in the registry's alias maps
+	// (temporalio/sdk-go#2141). A closure must therefore keep its fully-qualified
+	// name, which retains the enclosing identity and is unique across packages.
+	nameA, isMethod := getFunctionName(firstClosureA())
+	require.False(t, isMethod)
+	require.Contains(t, nameA, "firstClosureA",
+		"closure name must retain its enclosing identity, got %q", nameA)
+
+	nameB, _ := getFunctionName(firstClosureB())
+	require.Contains(t, nameB, "firstClosureB", "got %q", nameB)
+	require.NotEqual(t, nameA, nameB)
+
+	// Named functions keep their short name.
+	namedFn, _ := getFunctionName(testWorkflowHello)
+	require.Equal(t, "testWorkflowHello", namedFn)
+
+	// Methods keep their short name and are reported as methods.
+	a := &getFunctionNameActivities{}
+	method, isMethod := getFunctionName(a.NamedActivity)
+	require.True(t, isMethod)
+	require.Equal(t, "NamedActivity", method)
+}


dplyukhin:
Thanks for the contrib, Jaideep! I think this is quite a tricky issue, related to a deeper refactoring of aliases I'm trying to do in the Go SDK as part of #2379 . As Andrew mentioned in the issue you cited, changing `getFunctionName` directly will break old histories - correct me if I'm wrong. I'm working on a solution that brings Go in line with the other SDKs without introducing NDEs.

jaideeppyne:
Thanks @dplyukhin, and you're right — I should've caught that. `getFunctionName` also produces the registered activity/workflow *type name*, so qualifying closures would rename anonymous activities/workflows registered without an explicit `Name`, and replaying a history that recorded the old `funcN` would NDE.

One thing that might help with scoping: the actual misrouting in #2141 surfaces through the test workflow environment's alias maps, not the name that reaches history — so a narrower fix that gives those maps a collision-free key while leaving `getFunctionName`'s output untouched would kill the test-only bug with no NDE risk. That said, if #2379 is already realigning aliasing with the other SDKs, that's clearly the better home for it.

Happy to either narrow this to the alias maps as a stopgap or close it in favor of #2379 — whichever is least work for you. Thanks for the context on the refactor.

PR #2562, opened 2026-08-14
Apply activity serialization context to local activity results

xepozz:
## What changed?
Serialize local activity results with the activity serialization context (`IsLocal=true`) on both ends, like regular activities.

## Why?
They were encoded plain but decoded with the workflow context — context-aware converters got a mismatch.

## Checklist

1. Closes 

2. How was this tested:

https://github.com/temporalio/features/pull/868

3. Any docs updates needed?


internal/workflow.go
```diff
@@ -1288,6 +1288,12 @@ func (wc *workflowEnvironmentInterceptor) ExecuteLocalActivity(ctx Context, type
 		Attempt:                     1, // Attempts always start at one
 	}
 
+	// Decode the local activity result with the same activity serialization
+	// context it was encoded with, instead of the workflow context.
+	if df, ok := future.(*decodeFutureImpl); ok {
+		df.dataConverter = params.DataConverter
```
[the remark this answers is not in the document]
yuandrew: Looks like this is legit.  I think we will need to use an SDK flag here to avoid breaking users' replay, if they hit this edge case. When introducing an SDK flag, we need to keep it disabled by default for at least a release or 2 before we can enable it by default, so this fix will be a little delayed before it's in and enabled by default

internal/internal_task_pollers.go
```diff
@@ -980,7 +978,7 @@ func (lath *localActivityTaskHandler) executeLocalActivityTask(task *localActivi
 		)
 	})
 	ctx, err := WithLocalActivityTask(lath.backgroundContext, task, lath.logger, lath.metricsHandler,
-		lath.dataConverter, lath.interceptors, lath.client, lath.workerStopChannel)
+		task.params.DataConverter, lath.interceptors, lath.client, lath.workerStopChannel)
```
yuandrew: I think this may stop TestActivityEnvironment.ExecuteLocalActivity from using the converter configured through SetDataConverter. This path doesn’t set params.DataConverter, and we’re removing the handler-level fallback to env.dataConverter, so it looks like execution would fall back to the default converter.

Can we preserve the fallback or set the converter on the params here, and add a test using a custom converter?

internal/workflow.go
yuandrew, line 1291: I think this exposes a mismatch for static local-activity mocks. OnActivity(...).Return(value, nil) is encoded using env.GetDataConverter(), which has the workflow context, but the future will now decode it using the activity context here.

Can we update the static mock path to encode with the converter from the activity context and add coverage for this case?

internal/internal_task_pollers.go
yuandrew: Could this pass a workflow-bound converter into the local-activity goroutine? task.params.DataConverter may already be the value returned by WithWorkflowContext, and that returned converter isn’t required to still implement ContextAware, so we may not be able to rebind it to the activity context.

Would it be safer to apply the activity serialization context to the worker-configured converter here and keep task.params.DataConverter for decoding on the workflow side?

test/replaytests/replay_test.go
```diff
@@ -17,6 +19,47 @@ import (
...
+type replayTestSigningCodec struct {
+	signature string
+}
+
+func (c *replayTestSigningCodec) WithSerializationContext(ctx converter.SerializationContext) converter.PayloadCodec {
+	switch sc := ctx.(type) {
+	case converter.WorkflowSerializationContext:
+		return &replayTestSigningCodec{signature: sc.WorkflowID}
+	case converter.ActivitySerializationContext:
+		return &replayTestSigningCodec{signature: sc.WorkflowID + ":" + sc.ActivityType}
```
yuandrew: Could we include sc.IsLocal in this signature? Right now a regular activity and a local activity with the same workflow and activity IDs produce the same signature, so I don’t think this test verifies that the result is using IsLocal=true.

internal/workflow.go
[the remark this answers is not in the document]
yuandrew: actually, after talking with Quinn, i think it's okay we take this fix without an SDK flag, but we will definitely want to call out in the CHANGELOG of this potential breaking behavior for users

internal/internal_task_pollers.go
xepozz: Real issue, but applying the context on the worker side only would split the ends: the worker would encode with the activity context while the workflow side still decodes with the workflow one. Instead the environment now keeps the worker-configured converters unwrapped, and activity/local activity/child workflow contexts are applied to those. Same as Java, which stores both separately:

https://github.com/temporalio/sdk-java/blob/e947cc23/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java#L82-L84

Not local-activity specific — regular activities never got ActivitySerializationContext either for such a converter. Tests cover both, changelog updated.

xepozz:
@yuandrew please have a look again

internal/internal_workflow.go
```diff
@@ -1635,6 +1637,33 @@ func getDataConverterFromWorkflowContext(ctx Context) converter.DataConverter {
 	return WithWorkflowContext(ctx, dataConverter)
 }
 
+func getRootDataConverterFromWorkflowContext(ctx Context) converter.DataConverter {
+	options := getWorkflowEnvOptions(ctx)
+	if options == nil || options.RootDataConverter == nil {
+		return getDataConverterFromWorkflowContext(ctx)
+	}
+	return WithWorkflowContext(ctx, options.RootDataConverter)
```
yuandrew: The root converter is still passed through `WithWorkflowContext` here before the activity/child serialization is applied. `workflow.ContextAware.WithWorkflowContext` may return a plain `converter.DataConverter`, then subsequent `WithDataConverterSerializationContext calls are no-ops. 

Can we preserve the raw worker converter until the target context is applied and add a regression test where `WithWorkflowContext` returna non-context-aware converter?

internal/internal_workflow_testsuite.go
```diff
@@ -897,6 +904,7 @@ func (env *testWorkflowEnvironmentImpl) executeLocalActivity(
 		backgroundContext:  env.workerOptions.BackgroundActivityContext,
 		metricsHandler:     env.metricsHandler,
 		logger:             env.logger,
+		dataConverter:      env.dataConverter,
```
yuandrew, line 922: I think `TestActivityEnvironment.ExecuteLocalActivity` should match the real worker. This path doesn’t set `ActivityType` or a context-bound `DataConverter`, so the fallback uses the raw configured converter without applying `ActivitySerializationContext{IsLocal: true}`.

The new test verifies that the configured converter is used, but because its codec doesn’t inspect the serialization context, it won’t catch this difference. Can we construct the same local-activity serialization context in this path and update the test to assert `IsLocal` and the activity type?

yuandrew (approved):
LGTM! Thanks for the fix

PR #2566, opened 2026-08-15
fix: awaitWithOptions yields once before checking an already-true condition

### LINKED ISSUE #2537: Non-determinism error after upgrading from v1.39 to v1.46
## Expected Behavior

Workflows continue to work normally

## Actual Behavior

```
TMPRL1100] During replay, a matching Timer command was expected in history event position 35. However, the replayed code did not produce that. Possible causes are nondeterministic workflow definition code, or an incompatible change in the workflow definition.
```

stack trace

```
process event for export [panic]:
go.temporal.io/sdk/internal.panicIllegalState(...)
	go.temporal.io/sdk@v1.46.0/internal/internal_command_state_machine.go:501
go.temporal.io/sdk/internal.(*commandsHelper).getCommand(0x40015cf801?, {0x4, {0x4000f1ea58, 0x2}})
	go.temporal.io/sdk@v1.46.0/internal/internal_command_state_machine.go:1075 +0x110
go.temporal.io/sdk/internal.(*commandsHelper).handleTimerStarted(0x1ea6326?, {0x4000f1ea58?, 0x400007d008?})
	go.temporal.io/sdk@v1.46.0/internal/internal_command_state_machine.go:1626 +0x2c
go.temporal.io/sdk/internal.(*workflowExecutionEventHandlerImpl).ProcessEvent(0x4000a579b0, 0x4000d51290, 0x3?, 0x0)
	go.temporal.io/sdk@v1.46.0/internal/internal_event_handlers.go:1294 +0x4b0
go.temporal.io/sdk/internal.(*workflowExecutionContextImpl).ProcessWorkflowTask(0x40018cd220, 0x400163ede0)
	go.temporal.io/sdk@v1.46.0/internal/internal_task_handlers.go:1225 +0x1464
go.temporal.io/sdk/internal.(*workflowTaskHandlerImpl).ProcessWorkflowTask(0x4000718120, 0x400163ede0, 0x40018cd220, 0x400144f9f0)
	go.temporal.io/sdk@v1.46.0/internal/internal_task_handlers.go:962 +0x450
go.temporal.io/sdk/internal.(*workflowTaskProcessor).processWorkflowTask(0x40000fc680, 0x400163ede0)
	go.temporal.io/sdk@v1.46.0/internal/internal_task_pollers.go:504 +0x3bc
go.temporal.io/sdk/internal.(*workflowTaskProcessor).ProcessTask(0x40000fc680, {0x1abac80?, 0x400163ede0?})
	go.temporal.io/sdk@v1.46.0/internal/internal_task_pollers.go:444 +0x27c
go.temporal.io/sdk/internal.(*baseWorker).processTaskAsync.func1()
	go.temporal.io/sdk@v1.46.0/internal/internal_worker_base.go:703 +0xf4
created by go.temporal.io/sdk/internal.(*baseWorker).processTaskAsync in goroutine 201
	go.temporal.io/sdk@v1.46.0/internal/internal_worker_base.go:682 +0x84
```

### Possible Root cause from Claude

`awaitWithOptions` changed from check-then-yield to yield-then-check.

Before — [v1.39.0 `internal/workflow.go#L662-L680`](https://github.com/temporalio/sdk-go/blob/v1.39.0/internal/workflow.go#L662-L680):

```go
timer := NewTimerWithOptions(ctx, options.Timeout, options.TimerOptions)
for !condition() {
    ...
    if timer.IsReady() { return false, nil }
    state.yield(functionName)
}
return true, nil
```

After — [v1.41.0 `internal/workflow.go#L713-L750`](https://github.com/temporalio/sdk-go/blob/v1.41.0/internal/workflow.go#L713-L750),
unchanged in [v1.46.0#L763-L800](https://github.com/temporalio/sdk-go/blob/v1.46.0/internal/workflow.go#L763-L800):

```go
cancelTimerOnCondition := wc.env.TryUse(SDKFlagCancelAwaitTimerOnCondition)
if cancelTimerOnCondition && condition() { return true, nil }
...
timer := NewTimerWithOptions(timerCtx, options.Timeout, options.TimerOptions)
for {
    ...
    if timer.IsReady() { return false, nil }
    state.yield(functionName)   // now runs unconditionally
    if condition() { break }
}
```

When the condition is already true at call time, ≤1.39 returns without yielding.
≥1.41 yields once first. That yield hands control to the dispatcher, so any other
runnable coroutine emits its commands before the calling coroutine emits its next one.
For workflows that interleave `Await` with `workflow.Go`, the command order changes and
replay diverges.

The reachable path is [`channelImpl.ReceiveWithTimeout`](https://github.com/temporalio/sdk-go/blob/v1.46.0/internal/internal_workflow.go#L841-L854),
which delegates to `AwaitWithTimeout`.

Note that plain [`Await`](https://github.com/temporalio/sdk-go/blob/v1.46.0/internal/workflow.go#L746-L761)
still uses `for !condition()`, so the two functions now disagree on yield semantics.

Uploaded repro + scrubbed history.json
[repro.zip](https://github.com/user-attachments/files/30913442/repro.zip)

mittalpk:
Fixes #2537.

`awaitWithOptions` only checked "condition already satisfied, return without waiting" when `cancelTimerOnCondition` (gated by `SDKFlagCancelAwaitTimerOnCondition`) was true. For a workflow execution that hasn't adopted that flag, the function always creates a timer and enters its wait loop, which yields once before its first `condition()` check — unlike `Await`, and unlike this function's own pre-1.41 loop, both of which check first.

For an in-flight workflow whose history was recorded entirely under the flag-disabled path, replaying it under a newer SDK now produces an extra yield the first time `AwaitWithTimeout`/`AwaitWithOptions`/`ReceiveWithTimeout` is called with an already-true condition. That yield hands control to the dispatcher, so any other runnable coroutine emits its commands first — reordering history and tripping `TMPRL1100` on replay, matching the reported panic exactly.

Moved the already-satisfied check above the flag lookup so it always runs regardless of `cancelTimerOnCondition`. The flag still gates everything it's meant to — proactively cancelling the timer if the condition becomes true mid-wait.

**Verification:** added `TestAwaitWithTimeoutDoesNotYieldWhenConditionAlreadyTrue`, asserting a coroutine spawned before an already-satisfied `AwaitWithTimeout` call doesn't run before the await returns. Confirmed it fails against unpatched code via `git stash` (`other coroutine` runs before `await returned`, exactly the reported reordering) and passes after. Also ran the existing `TestAwaitWithTimeoutNoTimerCancel` replay test (which already covers the flag-disabled path via a real recorded history, just not this specific already-true-at-entry case) — still passes unchanged. Full `go test ./internal -race`: passing. Full `go test ./test/replaytests/...`: passing. `gofmt`/`go vet` clean on the changed lines (this repo's `internal/workflow.go` has pre-existing gofmt drift elsewhere, confirmed unrelated via `git stash` comparison). Added a `CHANGELOG.md` entry under `[Unreleased]`.


internal/workflow.go
```diff
@@ -750,11 +750,17 @@ func (wc *workflowEnvironmentInterceptor) awaitWithOptions(ctx Context, options
 	state := getState(ctx)
 	defer state.unblocked()
 
-	cancelTimerOnCondition := wc.env.TryUse(SDKFlagCancelAwaitTimerOnCondition)
-	if cancelTimerOnCondition && condition() {
+	// Matches Await's contract: an already-satisfied condition returns
```
Quinn-With-Two-Ns, line 753: Please reword this comment so it make sense on its own without the context of this issue or this PR

internal/workflow.go
```diff
@@ -750,11 +750,17 @@ func (wc *workflowEnvironmentInterceptor) awaitWithOptions(ctx Context, options
...
+	// without yielding. Must NOT be gated by SDKFlagCancelAwaitTimerOnCondition
+	// below -- that flag only governs cancelling the timer mid-wait. Coupling
+	// them here yielded once unconditionally on every pre-flag workflow before
+	// its first condition() check, an extra command that breaks replay (#2537).
+	if condition() {
 		return true, nil
 	}
 
+	cancelTimerOnCondition := wc.env.TryUse(SDKFlagCancelAwaitTimerOnCondition)
```
Quinn-With-Two-Ns, line 768: What validation was done to ensure this change is backwards compatible with old histories? We need to validate this change is safe for workflow histories generated before and after this flag was set in history.

mittalpk:
Great catch, and you were right to push on this. I verified end-to-end against a real embedded dev server rather than just reasoning about it: ran the pre-fix code live to capture a real history with an orphaned timer followed by genuine subsequent activity, confirmed that history genuinely fails to replay under my first fix with the exact TMPRL1100 error you'd expect, then fixed it properly.

The fix now gates the already-satisfied short-circuit behind a new SDK flag (SDKFlagSkipAwaitTimerWhenConditionMet, disabled by default this release), the same way SDKFlagCancelAwaitTimerOnCondition already gates timer cancellation. An in-flight workflow whose history predates the flag keeps reproducing its own orphaned timer on replay; a new workflow (or this same one, going forward once live) can adopt the fix. Confirmed with the flag force-enabled that #2537 is genuinely fixed (no orphaned timer), and confirmed the old history still replays correctly even with the flag globally enabled, since TryUse never grants an unrecorded flag during replay.

Added the real captured pre-fix history as a permanent replay fixture (confirmed it fails against the previous commit, passes against this one) plus a unit test locking in the default-disabled behavior alongside the existing fixed-behavior one.

PR #2567, opened 2026-08-16
Fix query payload limits with external storage

danielmapar:
## What was changed

- Move query answer payload-limit validation from the `ContextHook` to the payload `Visit` path, after earlier visitors can externalize the payload.
- Preserve the existing behavior that translates oversized inline answers into failed query responses.
- Keep query failure details exempt from payload-limit enforcement.
- Strengthen the external-storage query integration test so its store and retrieval assertions can only be satisfied by the query result.

## Why?

#2459 removed the early 2 MB check in `ProcessQuery` so external storage could run before payload-size enforcement. The worker payload-limit visitor still performs an equivalent query-result check from its `ContextHook`, which `proxy.VisitPayloads` invokes before traversing the result payload.

When the namespace advertises `BlobSizeLimitError`, that hook emits `TMPRL1103`, clears the query result, and marks it failed before the preceding external-storage visitor can replace the payload with a reference. Applications using read-write external storage must therefore disable the worker-wide payload error limit to return oversized queries.

This change performs the specialized query check on the payloads produced by earlier visitors, while retaining the server-compatible failure translation when the post-visit payload is still too large.

Follow-up to #2456 and #2459.

## How was this tested?

From `internal/cmd/build`:

- `go run . check`
- `go run . unit-test`
- `go run . integration-test -dev-server -run "TestExternalStorageSuite/TestQuery"`

The unit and integration test commands run with the race detector through the repository build tool.

## Docs

No API documentation changes are needed. The user-facing fix is included in `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ to docs, or any other relevant information.
 
 ### Fixed
 
+- Fixed worker payload-limit validation running before external storage for query results.
 - Prevent workflow task failures when an activity with a custom ID completes while its cancellation
   command is pending.
 - `TestWorkflowEnvironment.MutableSideEffect` now honors the provided equals function and only
--- a/internal/payload_limits.go
+++ b/internal/payload_limits.go
@@ -127,11 +127,33 @@ func (v *payloadLimitsVisitorImpl) Visit(ctx *proxy.VisitPayloadsContext, payloa
 		size = int64((&commonpb.Payloads{Payloads: payloads}).Size())
 	}
 
-	err := v.checkPayloadSize(size, getPayloadLimitChecks(ctx.Context))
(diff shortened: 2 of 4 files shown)


yuandrew:
Please sign the CLA before we can review this code

danielmapar:
Signed

> Please sign the CLA before we can review this code

Signed

yuandrew:
@danielmapar You may need to add `dparreira@netflix.com` to your GH, or to amend the commit and change the `user.email` to match your GH account in order for the CLA check to pass

danielmapar:
Superseded by #2641, which implements this fix with expanded coverage and has now merged. Closing this PR as no longer needed. Thanks!

PR #2573, opened 2026-08-17
contrib/googleadk: harden replay-safe telemetry wrappers

DABH:
Follow-up hardening to #2514, from cross-checking the sibling Python PR (temporalio/sdk-python#1710) against this SDK.

## What changed

- **Gate sync instruments' `Enabled(ctx)` during replay**: OTel 1.43 instruments expose `Enabled`, which passed through ungated and reported true while the matching `Add`/`Record` would drop; it now reports false while suppressed, matching the logger wrapper.
- **Surface-drift guard**: reflection test failing on any wrapped-interface method neither declared nor allowlisted with a rationale — embedded interfaces delegate new OTel methods ungated (the inverse of Python's ABC no-op drift), so dependency bumps must audit them. Verified to fail on all 8 instruments with the `Enabled` fix reverted.
- **Query/validator telemetry always records**: the gate composes `workflow.IsReplaying` with `!workflow.IsReadOnly` (#2521, Experimental; suggested by @patbeqo) — the Go analog of Python's `is_replaying_history_events`; queries and validators never re-execute from history, so once-per-request recordings survive catch-up replays. Probes regression-test both trailing-history shapes; `go.temporal. Requires `go.temporal.io/sdk` >= v1.48.0 (ships `workflow.IsReadOnly` and `workflow.GetRandomStream`).
- **README**: note `go.opentelemetry.io/otel/log` is pre-1.0 (v0.19).

## Testing

`gofmt`, `go vet`, and `go test -race -count=1 ./...` green in `contrib/googleadk`, dev-server probes included; flipped probes also pass `-count=2`.


contrib/googleadk/README.md
```diff
@@ -405,11 +405,28 @@ re-creation non-recording:
...
-their callbacks never run under a workflow context), covering both your own
-workflow-side recordings through the global meter today and ADK's metrics once
-adk-go#479 lands.
+`NewReplaySafeMeterProvider` is forward-looking: it gates synchronous
+instrument recordings and reports their `Enabled` false while suppressed
+(observable instruments pass through — their callbacks never run under a
+workflow context), covering both your own workflow-side recordings through
+the global meter today and ADK's metrics once adk-go#479 lands.
+
+**Query handlers** are the one place the gate can drop live telemetry. The
```
patbeqo: Would using the new `workflow.IsReadOnly` in `replaySuppressed` fix this? 

`return !workflow.IsReadOnly(wfCtx) && workflow.IsReplaying(wfCtx)`

DABH: Nice, didn't know this landed - thanks for pointing it out.  Updated to use this!

DABH:
Note: will wait till at least tomorrow (when next sdk-go release ships) before merging to remove the pinning issue present now

contrib/googleadk/CHANGELOG.md
```diff
@@ -15,8 +15,12 @@ or Security.
 
 - `NewReplaySafeTracerProvider`, `NewReplaySafeLoggerProvider`, and
   `NewReplaySafeMeterProvider`: wrap the global OpenTelemetry providers so ADK
-  telemetry emitted from workflow code is not re-emitted on history replay. See
-  "Telemetry and replay" in the README.
+  telemetry emitted from workflow code is not re-emitted on history replay;
+  while suppressed, the logger's and sync instruments' `Enabled` report false.
+  The gate composes `workflow.IsReplaying` with `!workflow.IsReadOnly`
+  (Experimental, requires `go.temporal.io/sdk` >= v1.48.0), so telemetry from
```
patbeqo: nit: Can this be excluded now that we've released 1.48.0?
```suggestion
  (Experimental), so telemetry from
```

contrib/googleadk/README.md
```diff
@@ -39,9 +39,10 @@ This package depends on the deterministic ADK `platform` seams
 (`WithTimeProvider`, `WithUUIDProvider`, `WithTaskRunner`), `tool/toolutils.PackTool`,
 and the `model.NewLLM` registry lookup from upstream `google.golang.org/adk/v2`
 (the registry itself stays application-owned; this package never registers into it).
-These merged after the latest tagged ADK release (v2.0.0), so `go.mod` pins
-`adk/v2` to a `main`-branch pseudo-version for now; it reverts to an ordinary
-tagged version once a release ships that includes them.
+These merged after the latest tagged ADK release (v2.0.0), so `go.mod` pins a
+main-branch pseudo-version of `google.golang.org/adk/v2`; bump it to a tagged
```
patbeqo, line 43: Should we bump to latest version 2.2.0 ? https://pkg.go.dev/google.golang.org/adk/v2

contrib/googleadk/README.md
```diff
@@ -39,9 +39,10 @@ This package depends on the deterministic ADK `platform` seams
...
+version once a release ships that includes them. The telemetry gate composes
+`workflow.IsReadOnly`, which requires `go.temporal.io/sdk` >= v1.48.0.
```
patbeqo: nit
```suggestion
`workflow.IsReadOnly`.
```

contrib/googleadk/CHANGELOG.md
DABH: Good call — dropped it in 03c4ae04; go.mod already requires v1.48.0, so the note was redundant.

contrib/googleadk/README.md
DABH: Not in this PR — the v2.2.0 bump is #2538, which is more than a version move: it also pulls in the Go 1.26.5 and genai v1.66.0 floors and the multi-decision confirmation-resume change in hitl.go. Folding it in here would duplicate that PR and drag unrelated floor/behavior changes into a telemetry change. Once #2538 lands, this branch inherits v2.2.0 on rebase and this paragraph's adk/v2 pseudo-version note goes away with it.

contrib/googleadk/README.md
DABH: Done in 03c4ae04.

patbeqo (approved):

PR #2576, opened 2026-08-18
test: run categorized integration tests against Cloud

THardy98:
## What was changed

Cloud CI job now provisions an isolated Temporal Cloud namespace for each matrix row, configures the SDK through envconfig, and deletes the namespace after testing.

The existing oldstable/FIPS rows continue running the race-enabled Cloud smoke tests. Stable/non-FIPS runs the full race-enabled integration suite.

Tests that cannot run against Cloud use `skipOnCloud` with a categorized reason. Unmarked tests run against Cloud by default. Exclusions can be inspected with:

  ```bash
  rg -n '^\s*skipOnCloud\(' test --glob '*_test.go'
  ```
  
  An improved ability to inspect cloud skipped tests may be added in the future for better tracking.

## Why?

This lets us run the Go SDK integration suite against Cloud without maintaining an allowlist or sharing a long-lived test namespace. It also records beside each excluded test why it is not yet Cloud-compatible.

  ## Checklist

  1. Part of temporalio/features#851.

  2. How was this tested:

  - Cloud matrix execution against isolated namespaces.
  - Canonical `go run . check`.
  - Namespace helper and builder tests.
  - `actionlint` and focused Cloud-skip tests.

  3. Any docs updates needed?

  No


.github/scripts/cloud-namespace/go.mod
```diff
@@ -0,0 +1,19 @@
+module go.temporal.io/sdk/.github/scripts/cloud-namespace
+
+go 1.25.4
```
yuandrew: 1.26.0 please

.github/scripts/cloud-namespace/main.go
```diff
@@ -0,0 +1,152 @@
...
+	"go.temporal.io/cloud-sdk/cloudclient"
+)
+
+const (
+	cloudNamespaceRegion = "aws-ca-central-1"
+	operationTimeout     = 10 * time.Minute
+)
+
+func main() {
+	if err := run(context.Background(), os.Args[1:]); err != nil {
```
yuandrew: Instead of `ctx, cancel := context.WithTimeout(ctx, operationTimeout)` down below, can we have this at the timeout at the top level? Today if this stalls it can hang the GH job and prevent cleanup til the global timeout. Could also be good to set a 12 minute GH job level timeout 
```
  - name: Create Cloud namespace
    timeout-minutes: 12
```

.github/scripts/cloud-test-inventory.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
yuandrew: What is this script used for? And if it's useful, can we write it in Go instead?

yuandrew (commented):
Asked this offline as well, but this increases our CI times from ~13 min to 19 min, is there any way we can keep it under the current times?

.github/scripts/cloud-test-inventory.sh
THardy98: Small shell script that basically runs a `rg` command that gets tests according to their cloud runnability. We could write it in Go, but I would suspect it's simpler in bash

yuandrew: why is it needed if it's not getting called anywhere?

THardy98: It's intended for us (devs) to use for tracking, so we can keep track of which tests still need adaptation for cloud. The test suite itself doesn't use it because they are skipped if ineligible for cloud.

That being said - I'm going to remove it. It's just an `rg` command. I thought it'd be easier to keep it as a command that we can call in the repo, but i'll just write the command in the description of the tracking issue.

THardy98: I do think it'd be nice to have as a follow-up (nice, formatted output of cloud testing inventory that makes it easily human-readable)

yuandrew (approved):

PR #2577, opened 2026-08-18
fix(extstore): update GCS driver metadata to be consistent with GCS docs.

### LINKED ISSUE #2572: Update GCS driver object name metadata to be consistent with GCS docs
GCS drivers store the object's name as `key` in the metadata when `object_name` is the idiomatic term.

Replace `key` with `object_name` in the metadata and add a backwards compat fallback.

cconstable:
## What was changed
- claim metadata field `key` changed to `object_name`

## Why?
- consistency with gcs language and other temporal sdks
- Fixes https://github.com/temporalio/sdk-go/issues/2572

## Checklist
- updated tests
- updated readme
- updated changelog

### DIFF
--- a/contrib/gcp/gcsdriver/driver.go
+++ b/contrib/gcp/gcsdriver/driver.go
@@ -20,14 +20,19 @@ const (
 	driverType            = "gcp.gcsdriver"
 	defaultDriverName     = "gcp.gcsdriver"
 	hashAlgorithm         = "sha256"
-	keyVersion            = "v0"
+	objectNameVersion     = "v0"
 	nullSegment           = "null"
 	maxGCSObjectNameBytes = 1024
 
 	claimKeyBucket        = "bucket"
-	claimKeyKey           = "key"
+	claimKeyObjectName    = "object_name"
 	claimKeyHashAlgorithm = "hash_algorithm"
 	claimKeyHashValue     = "hash_value"
+
+	// TODO: v0.1.0 stored the object name under "key". Retrieve still accepts it so
+	// claims already written to history remain readable. Remove once the driver
+	// reaches GA.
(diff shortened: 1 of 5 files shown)


contrib/gcp/gcsdriver/driver.go
```diff
@@ -150,20 +156,20 @@ func (d *gcsStorageDriver) Store(
...
 				}
 			}
 			claims[i] = converter.StorageDriverClaim{
 				ClaimData: map[string]string{
 					claimKeyBucket:        pp.bucket,
-					claimKeyKey:           key,
+					claimKeyObjectName:    name,
 					claimKeyHashAlgorithm: hashAlgorithm,
 					claimKeyHashValue:     pp.hexDigest,
 				},
```
dplyukhin, line 176: Could we have a mixed fleet deployment scenario where (1) a new worker writes a claim with only the `"object-name"` field, (2) an old worker reads that claim, and (3) old worker errors out because of the missing `key` field?

dplyukhin (commented):
Should `contrib/gcp/gcsdriver/client.go` also change its use of "key"?

yuandrew:
Can we link the GH this PR addresses?

cconstable:
@yuandrew linked

contrib/gcp/gcsdriver/driver.go
cconstable: Yes, however it's only if newer SDK drivers store payloads and older drivers try to read (not the other way). We could have the newer drivers write both keys into the metadata. It's also still an experimental feature.

cconstable: Just pushed an update to include the old key to avoid this issue.

dplyukhin (approved):

PR #2581, opened 2026-08-19
contrib/googleadk: recreate spans during replay with deterministic IDs

DABH:
## What was changed

Stacked on #2573 (targets its branch). Adopts the `contrib/opentelemetry-v2` span model for the googleadk tracer wrapper, and moves the deterministic UUID seam onto the same workflow random-stream mechanism, per review feedback on #2573:

- Spans are now created for real during replay (with `workflow.Now` start timestamps) and their `End` is suppressed while replaying, instead of suppressing `Start`. `NewReplaySafeTracerProvider` now builds and owns its `sdktrace.TracerProvider` (`func(opts ...sdktrace.TracerProviderOption) *ReplaySafeTracerProvider`, promoting `Shutdown`/`ForceFlush`) and force-installs a deterministic span-ID generator fed per-tracer from `workflow.GetRandomStream` (crypto/rand fallback) that can never be omitted, so replayed spans are byte-identical to their originals — nothing exports twice and catch-up exports stitch into their original traces.
- Fixes the documented straddle loss: a span open across a sticky-cache eviction or worker restart now exports exactly once, complete — with its `gen_ai.usage.*` token attributes, original IDs, and parent linkage — instead of once truncated (evictions suppress ADK's deferred force-End because task teardown marks the handler replaying before running defers).
- The workflow span's `TracerProvider()` returns the owned replay-safe provider rather than the raw inner provider, so tracers derived from a span (`span.TracerProvider().Tracer(...)`) keep the End gate and the stream-fed IDs. The span wrapper joined the surface-drift guard, with its six state mutators (plus `IsRecording`/`SpanContext`) allowlisted as intentionally delegated — replay re-creations must re-accumulate the state a later live `End` exports.
- **Deterministic UUIDs now come from the same random stream.** `NewContext`'s `platform.WithUUIDProvider` seam (ADK's invocation / tool / event IDs) draws from `workflow.GetRandomStream` via `uuid.NewRandomFromReader` (v4) instead of a `workflow.SideEffect`-seeded v5 counter. This drops a per-workflow marker event, unifies the determinism mechanism with the span-ID generator, and gives read-only contexts (query handlers, update validators, side-effect functions) ordinary random IDs — they must not consume the deterministic stream. **Compatibility:** this changes the generated IDs and the command history, so a workflow that recorded the SideEffect marker under v0.1.0 or v0.2.0 must finish on that release; drain in-flight runs before upgrading. Documented under `## [Unreleased] → Changed` in the module CHANGELOG.
- Read-only contexts (queries/validators/side effects) keep real span ends, wall clocks, and random IDs; meter/logger wrappers and the #2573 gating are unchanged.
- Five dev-server probes: a verbatim legacy-design control, a deterministic-ID fan-out test (parallel tools, 5 replays, identical ID tuples in identical order), straddle and task-retry exports, and a derived-tracer end-to-end probe. The UUID seam's cross-replay stability is asserted directly (re-executing the same run redraws identical IDs) and exercised end-to-end through the same `GetRandomStream` mechanism by the span probes.

## Why?

ADK's `generate_content` spans are open precisely across the model Activity await — the window most exposed to eviction/restart — and previously lost their token attributes there. Deterministic IDs plus End-suppression recover them without re-inflating span counts (per-straddle export ledger: before 1 truncated, after 1 complete; verified by the in-file legacy control). Owning the generator inside the provider means a replay-safe tracer can never ship without it. Moving the UUID seam onto `workflow.GetRandomStream` retires a history marker, aligns it with the span-ID mechanism, and keeps query/validator IDs live rather than replay-suppressed — at the cost of a one-time in-flight determinism break, called out in the CHANGELOG.

## Checklist

1. Closes N/A (review follow-up to #2573)

2. How was this tested: full `contrib/googleadk` suite plus the five probes under `-race`, two consecutive runs (dev-server tests running, not skipped) plus the probe file and context tests at `-count=2`, and the replay/eviction probes under forced eviction; `go run . check` clean across all modules. The UUID seam has a direct cross-replay determinism assertion (fixed test run ID → identical redraw). The README hello-world and telemetry snippets compile verbatim and run end-to-end from a scratch module (`replace` → this module) against released `go.temporal.io/sdk` v1.48.0 — a minimal FakeModel workflow exports each workflow span once live with its token attributes and re-creates identical span IDs on replay, exporting nothing.

3. Any docs updates needed? README telemetry section rewritten in this PR — the install snippet collapses to the one-step `googleadk.NewReplaySafeTracerProvider(sdktrace.WithBatcher(exporter))` wiring (`otel.SetTracerProvider(tp)`; `defer tp.Shutdown(ctx)`) and compiles as written; the package godoc and the "Deterministic by construction" README bullet describe the stream-backed UUID generator; CHANGELOG updated with the owned-provider constructor and the UUID-provider change plus its in-flight compatibility note.


contrib/googleadk/telemetry.go
```diff
@@ -62,28 +127,69 @@ var suppressedTracer = noop.NewTracerProvider().Tracer("googleadk.replay-suppres
...
-// bypasses this wrapper permanently. Replays add no spans, but a span still
-// open when its workflow leaves the worker is truncated (sticky-cache
-// eviction) or lost (worker shutdown, crash): the replay re-creation on
-// resume is non-recording. See "Telemetry and replay" in the README for the
-// span-lifetime contract.
+// bypasses this wrapper permanently. See "Telemetry and replay" in the README
+// for the span-lifetime contract.
 func NewReplaySafeTracerProvider(inner trace.TracerProvider) trace.TracerProvider {
 	return replaySafeTracerProvider{TracerProvider: inner}
 }
```
patbeqo: Can we enforce the use of `NewWorkflowSpanIDGenerator`?

contrib/googleadk/telemetry.go
```diff
@@ -48,12 +52,73 @@ func replaySuppressed(ctx context.Context) bool {
...
+	return sid
+}
+
+func readSpanRandom(ctx context.Context, p []byte) {
+	r, _ := ctx.Value(otelRandomKey{}).(io.Reader)
+	if r == nil {
+		r = cryptorand.Reader
+	}
+	_, _ = io.ReadFull(r, p)
+}
```
patbeqo, line 96: Can we import this directly from the otel v2 package? Not sure if adding it as a dependency would cause any issues.

contrib/googleadk/telemetry.go
```diff
@@ -62,28 +127,69 @@ var suppressedTracer = noop.NewTracerProvider().Tracer("googleadk.replay-suppres
...
+	s.Span.End(opts...)
+}
+
+// TracerProvider returns the replay-safe wrapper around the inner span's
+// provider rather than the inner provider itself: deriving a tracer from a
+// span is the documented way to create more spans on the span's pipeline, and
+// an unwrapped tracer reached that way from workflow code would bypass the
+// End gate and the stream-fed span IDs.
+func (s *workflowSpan) TracerProvider() trace.TracerProvider {
+	return NewReplaySafeTracerProvider(s.Span.TracerProvider())
```
patbeqo: Could we store the replay-safe provider after verifying it, and then return that here instead of wrapping an arbitrary provider?

patbeqo:
I think we can also update `newDeterministicUUIDProvider` to use the new APIs.

```go
func newDeterministicUUIDProvider(ctx workflow.Context) platform.UUIDProvider {
	stream := workflow.GetRandomStream(ctx, "go.temporal.io/sdk/contrib/googleadk/uuid")
	return func() string {
		if workflow.IsReadOnly(ctx) {
			return uuid.New().String()
		}
		id, _ := uuid.NewRandomFromReader(stream)
		return id.String()
	}
}
```

contrib/googleadk/telemetry.go
DABH: Yes — done. Enforcing it turned out to require owning construction: sdktrace exposes no getter/setter for the IDGenerator (`WithIDGenerator` is applied only at `NewTracerProvider` and the field is private with no accessor), so we can neither retrofit the generator onto an arbitrary provider nor detect whether one is present. So `NewReplaySafeTracerProvider` now matches otel-v2's shape — `NewReplaySafeTracerProvider(opts ...sdktrace.TracerProviderOption) *ReplaySafeTracerProvider` — and builds the provider itself, force-appending the generator. It's now impossible to get a replay-safe tracer without deterministic IDs, and the generator became unexported as a result. (`25401238`)

contrib/googleadk/telemetry.go
DABH: I looked into this and don't think a direct import works, for a few independent reasons: otel-v2's generator type and its random-source context key are both unexported with no exported constructor, so external code can't name the generator or feed it our `workflow.GetRandomStream`; otel-v2 attaches the reader only inside its `WorkflowTracer.Start(workflow.Context)` path, whereas ADK captures the OTel global at package init and calls `Start(context.Context, …)` — on that path otel-v2's generator falls back to crypto/rand, i.e. non-deterministic IDs; and `contrib/opentelemetry-v2` is an untagged experimental module, so a released module depending on it would force a pseudo-version or a replace directive plus a release-ordering constraint. That's a lot of coupling for ~15 near-identical lines. And with the owned-provider change above the generator is now unexported here too, so there's no public duplication left to reconcile.

contrib/googleadk/telemetry.go
DABH: Done — this falls out of the owned-provider change. The constructor now owns the `*ReplaySafeTracerProvider`, the tracer and span carry a pointer to it, and `TracerProvider()` returns that canonical instance directly instead of re-wrapping `s.Span.TracerProvider()`. It's guaranteed to be the generator-bearing, End-gating provider, and it also removes the latent `==` panic risk in the derived-tracer path (comparing a concrete pointer now rather than a value struct over a possibly-non-comparable provider). (`25401238`)

DABH:
Good call — done in `7533258b`. Now that we require SDK v1.48.0 anyway, the provider draws from `workflow.GetRandomStream` and generates v4 UUIDs (with `uuid.New()` on the `IsReadOnly` path for queries/validators), which drops the per-workflow `SideEffect` marker and simplifies the code — essentially your snippet.

One thing worth calling out: since the UUID provider shipped in v0.2.0, removing the `SideEffect` is a determinism-breaking change for workflows already in-flight across the upgrade (their history recorded the marker; the new code won't reproduce it on replay). Given the package is experimental and this lands bundled with the rest of the telemetry rework, I think that's acceptable — noted it in the CHANGELOG under Changed. Shout if you'd rather gate it behind a worker version or hold it for a cleaner break.

DABH:
Small correction on the compat note above: the earlier SideEffect-based provider shipped in both v0.1.0 and v0.2.0, so any in-flight workflow that recorded the marker under either release (not just v0.2.0) needs to drain before upgrading. Fixed the CHANGELOG scope accordingly.

patbeqo (approved):

contrib/googleadk/CHANGELOG.md
```diff
@@ -19,10 +19,32 @@ or Security.
...
+
+- The deterministic UUID provider (`platform.WithUUIDProvider`, used for ADK's
+  invocation/tool IDs) now draws from `workflow.GetRandomStream` instead of a
+  `workflow.SideEffect`-seeded counter, dropping a per-workflow marker event and
+  giving query/update-validator contexts ordinary random IDs. This changes the
+  generated IDs and the command history: a workflow that recorded the SideEffect
+  marker under an earlier release (v0.1.0 or v0.2.0) must finish on that release,
+  because replaying it on the new provider is non-deterministic. Drain in-flight
+  runs before upgrading.
+- `NewReplaySafeTracerProvider` builds and owns its `sdktrace.TracerProvider`
```
brianstrauch, line 41: Could `NewTemporalTracerProvider` work here too?

contrib/googleadk/telemetry.go
```diff
@@ -2,15 +2,16 @@ package googleadk
 
 import (
 	"context"
+	cryptorand "crypto/rand"
```
brianstrauch: No naming conflict
```suggestion
	"crypto/rand"
```

contrib/googleadk/CHANGELOG.md
patbeqo: `NewReplaySafeTracerProvider` aligns with Python’s `ReplaySafeTracerProvider` type.

DABH: This one deliberately matches `contrib/opentelemetry-v2`'s `NewReplaySafeTracerProvider` — mirroring that module was the whole point of this change, and the tracer/logger/meter constructors form a symmetric `NewReplaySafe*` trio. `ReplaySafe` also names the behavior directly, where `Temporal` is a little redundant inside the temporal contrib (the only `NewTemporal*` in the repo is `temporalnexus.NewTemporalOperation`, where it means "backed by Temporal"). If we want a `NewTemporal*` convention for these providers I'd suggest landing it in opentelemetry-v2 first so googleadk doesn't diverge — happy to follow it here if so; otherwise I'd keep `NewReplaySafeTracerProvider`.

contrib/googleadk/telemetry.go
DABH: Done in 69d305e2 — dropped the alias.

PR #2583, opened 2026-08-19
Fix sticky workflow cache retention after worker shutdown

### LINKED ISSUE #2582: [performance] Stopped workers retain sticky workflow cache state until explicitly purged
## Summary

  Stopping the final Go SDK worker does not release workflow executions held in the process-wide sticky cache.

  Cached workflow state remains reachable after `Worker.Stop()`, even after application references are dropped and garbage collection completes. Calling `worker.PurgeStickyWorkflowCache()` releases the state, but normal worker shutdown should release its cache ownership deterministically.

  ## Practical scope

  This primarily affects applications that stop workers while keeping the same Go process alive.

  It is unlikely to materially affect the common container or Kubernetes deployment model where `Worker.Stop()` is followed by process termination. In that case, the operating system shortly reclaims the entire process heap. Rolling deployments and pod-based autoscaling similarly replace the worker process.

  Potentially affected applications include:
  - Long-lived processes that dynamically create, stop, or rotate workers
  - Multi-tenant worker hosts that remove workers for individual customers or task queues
  - Applications that replace workers in-process after configuration or credential changes
  - Programs that stop one worker and later create another without exiting
  - Test processes that repeatedly create and stop workers

  The measured memory and GC improvements describe this affected in-process lifecycle. They are not estimates of improvement for typical container deployments.

  ## Expected improvement for affected applications
  After the final worker stops, the fix should reclaim essentially all workflow state still held by the sticky cache.

  Using explicit purge as an approximation of the expected post-fix state:

  **32 workflows with 1 MiB state each**
  - Reclaimed **33.0 MiB**
  - Reduced workload-related heap remaining after shutdown by approximately **94%**
  - Reduced controlled GC CPU per cycle by approximately **24%**
  - Removed all 32 retained workflow coroutines

  **16 workflows with 512 KiB state each**
  - Reclaimed **8.9 MiB**
  - Reduced workload-related heap remaining after shutdown by approximately **82%**
  - Reduced controlled GC CPU per cycle by approximately **17%**
  - Removed all 16 retained workflow coroutines

  This fix would not reduce memory while a worker is running. The sticky cache is intentional during normal worker operation.

  The retained amount depends on workflow state. In this synthetic experiment, it was approximately the workflow state size plus 30–60 KiB of SDK/runtime overhead per cached workflow. This should not be extrapolated directly to arbitrary production workflows.

  ## Evidence

  The experiment ran three scenarios in isolated Go processes:
  1. Sticky cache disabled
  2. Sticky cache enabled and explicitly purged after shutdown
  3. Sticky cache enabled and left unpurged after shutdown

  Each scenario was repeated in three fresh-process trials.

  The unpurged scenario retained substantially more memory in every trial:
  - **32 workflows × 1 MiB:** 34,620,616 additional heap-object bytes, or 1,081,894 bytes per cached workflow
  - **16 workflows × 512 KiB:** 9,329,725 additional heap-object bytes, or 583,108 bytes per cached workflow

  Explicit purge returned memory and workflow-coroutine counts to approximately the cache-disabled control.

  Heap-profile diffs attributed the dominant retained memory to workflow state under the SDK workflow-coroutine execution path. In the larger experiment, 32,770 KiB was attributed directly to the workflow allocation.

  ## Suspected cause

  The current ownership graph is effectively:
  ```text
  global sharedWorkerCachePtr
    -> workflowCache
    -> workflowExecutionContextImpl
    -> workflowTaskHandlerImpl.cache
    -> WorkerCache
    -> sharedWorkerCachePtr
  ```

  `NewWorkerCache` increments the shared worker reference count and installs a finalizer responsible for decrementing it.

  Cached workflow task handlers reference that same `WorkerCache`. Because the graph begins at a global root, the finalizable `WorkerCache` remains reachable through the cache it is responsible for releasing.

  Normal worker shutdown does not explicitly release this ownership, so finalizer-based cleanup may never become eligible while the cache contains workflow executions.

  ## Expected behavior

  After the final worker using the shared sticky cache stops:

  - Its cache ownership should be released synchronously.
  - Cached workflow execution state should be cleared.
  - Cleanup should not require `worker.PurgeStickyWorkflowCache()`.
  - Cleanup should not depend on finalizer timing.

  Stopping one of several live workers must preserve cache state still shared by the remaining workers.

  ## Proposed direction

  Separate the shared cache handle from a per-worker ownership lease:
  - Workflow task handlers and cached values may retain the cache handle.
  - Cached values must not retain the ownership lease.
  - `Worker.Stop()` should release its lease exactly once.
  - Releasing one of several leases should preserve the cache.
  - Releasing the final lease should detach and clear the cache.
  - A finalizer or cleanup callback may remain as a fallback, but normal lifecycle correctness should not depend on it.

  The regression test should validate this lifecycle synchronously. It should not depend on heap-size thresholds, `runtime.GC`, or finalizer timing.

  ## Full investigation
  The complete methodology, measurements, profiles, commands, limitations, and proposed design are available in the [full investigation report](https://github.com/temporalio/sdk-go/blob/e0d367e626a0e345bd16cb68e083ee5ab6045f66/sticky-cache-retention-results.md).

  Additional resources:
  - [Profiling artifacts and command manifest](https://github.com/temporalio/sdk-go/blob/
  e0d367e626a0e345bd16cb68e083ee5ab6045f66/.sticky-cache-retention-artifacts/manifest.md)
  - [Investigation commit](https://github.com/temporalio/sdk-go/commit/e0d367e626a0e345bd16cb68e083ee5ab6045f66)
  - [Investigation branch](https://github.com/temporalio/sdk-go/tree/investigate-sticky-cache-retention)

  The experiment was performed against SDK commit:

  `0feb5a6813ab94b6e790260521a94914bb5e3267`

yuandrew:
## What was changed

Fix sticky workflow-cache retention after workers stop.

Workers now hold an explicit cache lease and release it during `Stop()`, with the finalizer retained as a fallback. Live workers share one cache generation; the final owner clears it.

NOTE: The previous cache-retention behavior masked an integration-test isolation bug. Once deterministic cache cleanup was added, tests that disabled caching left the process-global setting at zero for later tests. Test setup now restores the configured cache size before each test.

## Why

Final cache release previously depended on GC finalization. Cached workflow state could keep the ownership graph reachable, retaining workflow state and coroutines after all workers stopped.

With 32 incomplete workflows holding 1 MiB each, local profiling reduced retained heap from approximately 34.6 MB to 20 KB and retained workflow coroutines from 32 to zero.

This primarily benefits long-lived processes that repeatedly create and stop workers. Processes that exit immediately after stopping workers will see little benefit.

Full investigation: [e0d367e6](https://github.com/temporalio/sdk-go/commit/e0d367e6)

## Impact and risk

The final worker now clears cached workflows during `Stop()`. Cache traversal and callback scheduling are synchronous; workflow cleanup remains asynchronous. Clearing 10,000 entries took approximately 3–4 ms locally, though large caches may cause a short-lived CPU and goroutine spike.

The next worker starts with a cold cache and replays workflow histories as tasks arrive. Shutdown may also produce a `sticky_cache_total_forced_eviction` spike under the existing metric semantics.

Tests cover shared ownership, final release, failure paths, concurrent release, fresh cache generations, and processing after worker recreation.

  ## Checklist

  1. Closes #2582

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ to docs, or any other relevant information.
 
 ### Fixed
 
+- Stopped workers now release sticky workflow cache ownership immediately. When the final worker
+  stops, cached workflow state is cleared without waiting for garbage collection.
+
 ### Security
 
 ## [1.48.0] - 2026-08-18
--- a/internal/internal_task_handlers_interfaces_test.go
+++ b/internal/internal_task_handlers_interfaces_test.go
@@ -26,7 +26,9 @@ type (
 )
 
 // Sample Workflow task handler
-type sampleWorkflowTaskHandler struct{}
+type sampleWorkflowTaskHandler struct {
+	cache *WorkerCache
+}
 
 func (wth sampleWorkflowTaskHandler) ProcessWorkflowTask(
 	workflowTask *workflowTask,
@@ -46,7 +48,7 @@ func (wth sampleWorkflowTaskHandler) GetOrCreateWorkflowContext(
 	retme := &workflowExecutionContextImpl{
 		mutex: sync.Mutex{},
 		wth: &workflowTaskHandlerImpl{
-			cache: NewWorkerCache(),
+			cache: wth.cache,
 		},
 	}
 	// The mutex is expected to already be locked in situations where unlock on the execution
@@ -55,8 +57,8 @@ func (wth sampleWorkflowTaskHandler) GetOrCreateWorkflowContext(
 	return retme, nil
 }
 
-func newSampleWorkflowTaskHandler() *sampleWorkflowTaskHandler {
-	return &sampleWorkflowTaskHandler{}
+func newSampleWorkflowTaskHandler(t testing.TB) *sampleWorkflowTaskHandler {
+	return &sampleWorkflowTaskHandler{cache: newTestWorkerCache(t)}
 }
 
 // Sample ActivityTaskHandler
@@ -109,7 +111,7 @@ func (s *PollLayerInterfacesTestSuite) TestProcessWorkflowTaskInterface() {
 	s.NoError(err)
 
 	// Process task and respond to the service.
-	taskHandler := newSampleWorkflowTaskHandler()
+	taskHandler := newSampleWorkflowTaskHandler(s.T())
 	request, err := taskHandler.ProcessWorkflowTask(&workflowTask{task: response}, nil, nil)
 	completionRequest := request.rawRequest.(*workflowservice.RespondWorkflowTaskCompletedRequest)
 	s.NoError(err)
--- a/internal/internal_task_handlers_test.go
+++ b/internal/internal_task_handlers_test.go
@@ -125,9 +125,7 @@ func (t *TaskHandlersTestSuite) SetupSuite() {
 }
 
 func (t *TaskHandlersTestSuite) TearDownTest() {
-	if cache := *sharedWorkerCachePtr.workflowCache; cache != nil {
-		cache.Clear()
-	}
+	PurgeStickyWorkflowCache()
 }
 
 func TestTaskHandlersTestSuite(t *testing.T) {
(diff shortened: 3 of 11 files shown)


yuandrew:
This is on my archived fork, closing in favor of a non-fork PR

PR #2586, opened 2026-08-19
Add payload validation error factory

bergundy:
## Summary

- add `temporal.NewPayloadValidationError(details any)`
- construct the existing non-retryable `*temporal.ApplicationError` with type `PayloadValidationError`
- preserve `details` as the single application error detail
- replace positive manual payload-validation error construction in Nexus tests while retaining the retryable compatibility case

## Testing

- `go run . unit-test -run 'TestNewPayloadValidationError|TestPayloadSerializerDeserializeErrors'`
- `go test -race ./temporal ./internal -run 'TestNewPayloadValidationError|TestPayloadSerializerDeserializeErrors' -count=1`
- `go run . integration-test -dev-server -run '^TestNexusOperationInputDeserializationError$'`
- `go run . check`
- `gofmt` and `git diff --check`

### DIFF
--- a/temporal/error.go
+++ b/temporal/error.go
@@ -188,6 +188,12 @@ func NewNonRetryableApplicationError(message, errType string, cause error, detai
 	)
 }
 
+// NewPayloadValidationError creates a non-retryable ApplicationError for payloads that fail validation.
+// The error has type PayloadValidationError and carries details as its single detail.
+func NewPayloadValidationError(details any) error {
+	return internal.NewPayloadValidationError(details)
+}
+
 // CanceledErrorOptions should be used to set all the desired attributes of a new CanceledError
 // To get a new instance use CanceledErrorAttributes function.
 type CanceledErrorOptions = internal.CanceledErrorOptions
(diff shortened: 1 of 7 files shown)


bergundy (commented):
Initially wanted to expose this in `go.temporal.io/sdk/converter` but that doesn't work due to circular dependencies.

temporal/error.go
```diff
@@ -188,6 +188,12 @@ func NewNonRetryableApplicationError(message, errType string, cause error, detai
 	)
 }
 
+// NewPayloadValidationError creates a non-retryable ApplicationError for payloads that fail validation.
+// The error has type PayloadValidationError and carries details as its single detail.
+func NewPayloadValidationError(details any) error {
```
Quinn-With-Two-Ns, line 193: Generally for details for a regular application error we take a vararg for details

bergundy: Yup, but this is restricting it to a single detail. I don't think we need more than a single one for this error.

Quinn-With-Two-Ns (approved):

PR #2596, opened 2026-08-21
Isolate the workflow replayer cache

### LINKED ISSUE #2595: Workflow replayer caches execution state it cannot reuse
## Expected Behavior

`WorkflowReplayer` is a one-shot replay facility. Each call processes its supplied history once, so replay state that cannot be reused should not be stored in the process-wide sticky cache used by live workers.

## Actual Behavior

`replayWorkflowHistoryRoot` creates its cache with `NewWorkerCache`, so the replayer shares `sharedWorkerCachePtr` with every live worker in the process.

The cached replay context cannot provide a cache hit:

- Each replay call constructs one synthetic workflow task containing the supplied history and creates a new workflow task handler.
- Sticky-cache reuse requires the same task handler and a later history segment beginning immediately after the previously processed workflow task.
- A subsequent replay provides neither. Normal history replays generate a new run ID; replays that retain the original run ID still use a new task handler and supply history from `WorkflowExecutionStarted`, so the cached context is rejected and the workflow is reconstructed from the beginning.

When the supplied history is incomplete, that unusable workflow context remains in the shared cache after the replay returns. Repeated replays can therefore displace live workflow state and force avoidable live-history replays without ever benefiting replay itself.

1. Create a `WorkerCache` and record the size of its workflow cache.
2. Replay an incomplete history containing `WorkflowExecutionStarted`, `WorkflowTaskScheduled`, and `WorkflowTaskStarted`.
3. Observe that the process-wide workflow cache grows by one entry after the replay returns.
4. Replay the history again and observe that it is reconstructed from the beginning rather than resumed from the retained context.

The replayer can instead use a private zero-capacity `WorkerCache`: it still processes the supplied history, but it neither reads nor retains state in the live worker cache.

## Specifications

- Version: current `main` (`4fac81895b0e5cee77d7e3f180ae06893df13b5a`); also reproduced with v1.39.0
- Platform: independent of operating system and server persistence

ali-khokhar-nvidia:
Fixes #2595.

## What changed?

Workflow replays now use a private zero-capacity worker cache instead of the process-wide cache shared by live workers. The incomplete-history replay test verifies that replaying does not change the live sticky-cache size.

## Why?

A replay call processes its supplied history once and cannot receive a later workflow task. Retaining an incomplete replay context in the shared sticky cache consumes capacity without providing a cache hit, and can displace live workflow state.

## How did you test it?

- [ ] built
- [ ] run locally and tested manually
- [ ] covered by existing tests
- [x] added new unit test(s)
- [ ] added new functional test(s)

```text
go test ./internal -run '^TestInternalWorkerTestSuite$/^TestReplayWorkflowHistory_IncompleteWorkflowExecution$' -count=1
```

## Potential risks

Replayer calls no longer retain workflow state between invocations. Live workers still use the same process-wide sticky cache and are otherwise unchanged.


Quinn-With-Two-Ns (approved):

ali-khokhar-nvidia:
Rebasing and the test fail is a flake.

ali-khokhar-nvidia:
Hello @Quinn-With-Two-Ns,
Looks like 2 flakes.

PR #2598, opened 2026-08-21
Omit nil payload validation details

bergundy:


internal/error.go
```diff
@@ -433,6 +433,9 @@ func NewApplicationError(msg string, errType string, nonRetryable bool, cause er
 //
 // Exposed as: [go.temporal.io/sdk/temporal.NewPayloadValidationError]
 func NewPayloadValidationError(details any) error {
+	if details == nil {
```
Quinn-With-Two-Ns, line 436: I'll note if we had just matched `NewApplicationError` this wouldn't have been an issue

Quinn-With-Two-Ns (approved):

internal/error.go
bergundy: Agree

PR #2599, opened 2026-08-21
Add Google Cloud Run CloudRunIDPlugin

seanbollin:
## What

Adds an experimental Google Cloud Run worker-identity plugin — the Cloud Run counterpart to the SDK's AWS Lambda worker support (`contrib/aws/lambdaworker`) — for both Cloud Run worker pools and services. Cloud Run runs a long-lived container, so this is a small client plugin rather than a worker wrapper. It ships as its own module, `go.temporal.io/sdk/contrib/gcp/cloudrun/id`.

> **Naming:** the plugin type is `CloudRunIDPlugin` in package `id`, so usage reads `id.CloudRunIDPlugin` (Go initialism convention capitalizes `ID`, consistent with `WorkflowID`/`RunID` across the SDK). It sets the client identity — which the workers then inherit — and the name carries the Cloud Run context without relying on the full import path.

## What it does

Register the plugin on your client:

```go
    Plugins: []client.Plugin{id.NewCloudRunIDPlugin()},
})
```

When the client connects, the plugin reads the current Cloud Run instance's metadata — the revision (`CLOUD_RUN_REVISION` / `K_REVISION`), the worker pool or service name (`CLOUD_RUN_WORKER_POOL` / `K_SERVICE`), and the unique instance id from the metadata server (`http://metadata.google.internal/computeMetadata/v1/instance/id`) — and sets the client `Identity` to `<instanceID>@<revision>` when one isn't already set. Workers created from that client inherit it.

`id.FetchMetadata` / `Metadata.Identity()` are exposed for reading the values directly. Experimental.


contrib/gcp/cloudrun/id/doc.go
```diff
@@ -0,0 +1,84 @@
...
+//
+// The worker pool (or service) name and revision come from environment variables that Cloud Run
+// injects into every container instance: CLOUD_RUN_WORKER_POOL and CLOUD_RUN_REVISION on worker
+// pools, or K_SERVICE and K_REVISION on services. The unique instance ID is only available from the GCP
+// metadata server, which [FetchMetadata] queries over HTTP; the metadata server is available on
+// both worker pools and services.
+//
+// Because [FetchMetadata] performs a network request, call it at worker startup and never from
+// workflow code: the SDK's workflowcheck analyzer flags net/http usage inside workflows because it
+// is non-deterministic.
```
Sushisource:
```suggestion
```
I don't think anyone would even really attempt this. No need to call this out any more than we'd call it out for any normal client method that shouldn't be used in workflow code.

contrib/gcp/cloudrun/id/doc.go
```diff
@@ -0,0 +1,84 @@
...
+package workerid
+
+import (
+	"log"
+
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/worker"
+)
+
+// Example shows how to configure a normal, long-lived Temporal worker on Cloud Run with the plugin.
```
Sushisource:
```suggestion
// Example shows how to configure Temporal worker on Cloud Run with the plugin.
```

Normal? As opposed to abnormal? Lol

contrib/gcp/cloudrun/id/metadata.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// FetchMetadata reads the current Cloud Run instance's metadata. It reads the worker pool (or
+// service) name and revision from environment variables — CLOUD_RUN_WORKER_POOL and
+// CLOUD_RUN_REVISION on Cloud Run worker pools, or K_SERVICE and K_REVISION on Cloud Run services —
+// and fetches the unique instance ID from the GCP metadata server.
+//
+// FetchMetadata performs a network request and must be called at worker startup, never from
+// workflow code: the SDK's workflowcheck analyzer flags net/http usage inside workflows because it
+// is non-deterministic.
```
Sushisource: More of this
```suggestion
// FetchMetadata performs a network request.
```

Better instead to say why, which I think some of the other PRs do (needs to call that one api for whatever info)

contrib/gcp/cloudrun/workerid/plugin.go
```diff
@@ -0,0 +1,128 @@
...
+	"sync"
+
+	"go.temporal.io/sdk/client"
+)
+
+// pluginName is the name reported by [Plugin.Name].
+const pluginName = "temporal-cloudrun-worker-id"
+
+// PluginOptions configures [NewPlugin]. The zero value is valid and is the normal choice on Cloud
+// Run. The remaining fields are for tests and advanced use.
```
Sushisource: Do these fields actually need to be public? At least `Metadata` itself seemingly does not need to be.

Sushisource (approved):
Approving since nothing blocking but a few more LLM-isms and one possible encapsulation fix.

PR #2602, opened 2026-08-23
Fix structured logging for malformed Nexus links

### LINKED ISSUE #2587: Nexus link parse failure logs a literal %s, uses the URL as an attribute key, and drops the error
## Expected Behavior

When `url.Parse` fails on a Nexus link, the logged entry should name the offending URL and carry the parse error under the `Error` key, with the message rendered rather than containing a literal `%s`.

## Actual Behavior

[`internal/internal_nexus_task_handler.go:202`](https://github.com/temporalio/sdk-go/blob/master/internal/internal_nexus_task_handler.go#L202) calls the logger as if it were a formatting function:

```go
nctx.log.Error("Failed to parse link url: %s", link.GetUrl(), tagError, err)
```

`log.Logger.Error` has the signature `Error(msg string, keyvals ...interface{})` — there is no formatting. So:

- The `%s` in the message is never substituted.
- The remaining three arguments are consumed as key/value pairs, but there are an odd number of them, so they misalign: `link.GetUrl()` lands in a **key** position, the string `"Error"` becomes its **value**, and `err` — the one thing worth logging — is left dangling without a value.

With `log.NewStructuredLogger` (or any `slog`-backed logger), the dangling argument becomes slog's `!BADKEY`:

```json
{
  "level": "ERROR",
  "msg": "Failed to parse link url: %s",
  "https://example.com/%zz": "Error",
  "!BADKEY": "parse \"https://example.com/%zz\": invalid URL escape \"%zz\""
}
```

Using an attacker- or server-supplied URL as an attribute *key* is also undesirable in its own right, since it makes the key space unbounded and can break downstream log indexing.

Suggested fix:

```go
nctx.log.Error("Failed to parse link url", "LinkURL", link.GetUrl(), tagError, err)
```

  1. Configure a client with `Logger: log.NewStructuredLogger(slog.New(slog.NewJSONHandler(os.Stdout, nil)))`.
  1. Have a Nexus task arrive whose `links[].url` is not parseable by `net/url.Parse` (e.g. `https://example.com/%zz`).
  1. Observe the emitted entry: the message still contains `%s`, the URL is used as the attribute key, and the parse error is logged under `!BADKEY`.

This is reachable by inspection alone — the argument count at the call site is odd, which is unambiguously a misuse of the variadic key/value API regardless of input.

## Specifications

  - Version: reproduced on v1.39.0 (line 159) and v1.47.0 / master (line 202); present in the latest release, v1.48.0
  - Platform: any (platform-independent)

AbuBakkarBinAkmal:
## What was changed

- Log malformed Nexus link URLs and parse errors using stable structured fields.
- Add a regression test that drives an invalid link through the real Nexus task handler and the SDK's `slog` adapter.
- Document the fix in the unreleased changelog.

Before:

```json
{
  "msg": "Failed to parse link url: %s",
  "https://example.com/%zz": "Error",
  "!BADKEY": "parse \"https://example.com/%zz\": invalid URL escape \"%zz\""
}
```

After:

```json
{
  "msg": "Failed to parse link url",
  "LinkURL": "https://example.com/%zz",
  "Error": "parse \"https://example.com/%zz\": invalid URL escape \"%zz\""
}
```

## Why?

`log.Logger.Error` accepts a message followed by key/value pairs; it does not apply `printf` formatting. The previous call left `%s` in the message, used the server-supplied URL as an attribute key, and left the parse error dangling under `!BADKEY` in structured logs.

Using explicit `LinkURL` and `Error` keys keeps the log schema stable and preserves the useful parse error.

## Checklist

1. Closes #2587

2. How was this tested:

- `go test ./internal -run '^TestMalformedRequestLinkLogsStructuredFields$' -count=1 -v`
- `cd internal/cmd/build && go run . unit-test -run '^TestMalformedRequestLinkLogsStructuredFields$'`
- `cd internal/cmd/build && go run . check`
- `cd internal/cmd/build && go run . integration-test -dev-server -run '^TestNexusSyncOperation$'`

3. Any docs updates needed?

No documentation changes are needed. `CHANGELOG.md` includes the fix under `Unreleased`.


internal/internal_nexus_task_handler.go
```diff
@@ -199,7 +199,7 @@ func (h *nexusTaskHandler) handleStartOperation(
 		}
 		linkURL, err := url.Parse(link.GetUrl())
 		if err != nil {
-			nctx.log.Error("Failed to parse link url: %s", link.GetUrl(), tagError, err)
+			nctx.log.Error("Failed to parse link url", "LinkURL", link.GetUrl(), tagError, err)
```
Quinn-With-Two-Ns:
```suggestion
			nctx.log.Error("Failed to parse link url", tagLinkURL, link.GetUrl(), tagError, err)
```
The tags should be well defined

AbuBakkarBinAkmal: This has been addressed in the latest commit. Please have a look @Quinn-With-Two-Ns 
Thank you for the feedback

Quinn-With-Two-Ns (approved):

PR #2608, opened 2026-08-24
Preserve child context cancellation order

### LINKED ISSUE #1208: Canceling a context with multiple child contexts can be non deterministic

yuandrew:
## What was changed
Preserve child-context creation order during cancellation propagation.

## Why?
Determinism good

## Checklist

1. Closes #1208 

2. How was this tested:
Added a regression test that fails before the fix.

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes core workflow context cancellation and replay determinism, but default-off flagging and map-based fallback preserve existing histories until flag 9 is enabled.
> 
> **Overview**
> Adds **SDK flag 9** (`SDKFlagOrderedChildCancel`), **off by default**, so existing workflows and replays keep today’s behavior until a later default flip.
> 
> Workflow `cancelCtx` no longer stores children only in a map. It uses a **`childList`** (map plus doubly-linked list) so children stay in **creation order** while add/remove stay O(1). When a parent cancels **more than one** child and the flag is active via `TryUse`, cancellation walks the list in creation order; otherwise it still iterates the map for legacy replay compatibility. The flag is not recorded when there is at most one child, since order cannot matter.
> 
> **CHANGELOG**, unit tests for `childList` and cancel ordering, and replay fixtures (`ordered-child-cancel.json` / `legacy-child-cancel.json`) with `OrderedChildCancelWorkflow` document and lock in deterministic vs legacy timer-cancel ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81c5cf7803cf3f338a25c2fe5bf50ecd19f6ed2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Quinn-With-Two-Ns:
This needs an SDK flag no?

yuandrew:
oops, haha. thanks for calling out

internal/context.go
```diff
@@ -272,10 +275,12 @@ type cancelCtx struct {
 
 	done Channel // closed by the first cancel call.
 
-	children     map[canceler]bool // set to nil by the first cancel call
-	childrenLock sync.Mutex
-	err          error // set to non-nil by the first cancel call
-	errLock      sync.RWMutex
+	// Keep both orders because unflagged histories require map traversal.
+	children        map[canceler]bool
+	orderedChildren []canceler
```
dplyukhin: Is it important that this comment might not make sense out of context? I'd write:
```suggestion
	// An unordered set of cancelable child contexts.
	// The field is set to nil by the first cancel call.
	children        map[canceler]bool
	// An ordered version of the children field, used to cancel children
	// in a deterministic order. See [SDKFlagOrderedChildCancel].
	orderedChildren []canceler
```

test/replaytests/replay_test.go
```diff
@@ -435,13 +435,39 @@ func (s *replayTestSuite) TestCancelOrder() {
 	replayer := worker.NewWorkflowReplayer()
 	replayer.RegisterWorkflow(CancelOrderSelectWorkflow)
 
+	// These histories predate ordered child cancellation.
```
dplyukhin:
```suggestion
	// These histories predate [internal.SDKFlagOrderedChildCancel].
```

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ to docs, or any other relevant information.
 - Corrected stand-alone activity API documentation to use activity terminology, document that
   `GetActivityHandleOptions.RunID` may be empty to target the latest run, and describe
   `TerminateActivityOptions.Reason` as a termination reason.
+- Added disabled-by-default SDK flag 9 for deterministic workflow child-context cancellation.
```
dplyukhin, line 79: Do users know what SDK flags are? I wouldn't mention this until we actually enable the feature, unless it's a suggestion that you should update your SDK now if you want this bugfix in the future.

Btw, how do you remember to enable flags in the next release?

dplyukhin (approved):

CHANGELOG.md
yuandrew: Not sure how much users know/care, but we have generally mentioned SDK flags to users. On the small off-chance that someone wants to try/take the fix immediately, they can enable it with an env variable 

> Btw, how do you remember to enable flags in the next release?

Before I'd just remember mentally, but now-a-days I mark it "blocked" on Jira so it's on my board til I'm ready to enable it

internal/context.go
```diff
@@ -219,9 +219,18 @@ func propagateCancel(parent Context, child canceler) {
...
+			}
+			if _, ok := p.children[child]; !ok {
+				node := &childNode{child: child, prev: p.lastChild}
+				if p.lastChild == nil {
+					p.firstChild = node
+				} else {
+					p.lastChild.next = node
+				}
+				p.lastChild = node
+				p.children[child] = node
```
dplyukhin: To L33T for me personally - I think this is clearer, but no pressure:
```suggestion
				if p.firstChild == nil {
					p.firstChild = node
  				    p.lastChild = node
				} else {
					p.lastChild.next = node
  				    p.lastChild = node
				}
  				p.children[child] = node
```

internal/context.go
```diff
@@ -252,8 +261,21 @@ func removeChild(parent Context, child canceler) {
 		return
 	}
 	p.childrenLock.Lock()
```
dplyukhin, line 251: `defer` the childrenLock here?

internal/context.go
```diff
@@ -265,14 +287,25 @@ type canceler interface {
...
+
 // A cancelCtx can be canceled.  When canceled, it also cancels any children
 // that implement canceler.
 type cancelCtx struct {
 	Context
 
 	done Channel // closed by the first cancel call.
 
-	children     map[canceler]bool // set to nil by the first cancel call
+	// Keep the map for unflagged replay and the list for ordered cancellation.
```
dplyukhin: The map does serve a function for ordered cancellation - finding the node. So you can remove that part of the comment

PR #2609, opened 2026-08-24
fix(metrics): label payload limit failures as PayloadsTooLarge

jmaeagle99:
## What was changed

- Plumb activity failure causing errors from the activity task handler to the activity task poller.
- Associate `PayloadsTooLarge` failure reason with `payloadSizeError` error type.

## Why?

Better observability of failure reason for payload size errors.

## Checklist

1. How was this tested: Updated integration test.

3. Any docs updates needed? Yes

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -2357,7 +2357,7 @@ func newServiceInvoker(
 }
 
 // Execute executes an implementation of the activity.
-func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice.PollActivityTaskQueueResponse) (result any, err error) {
+func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice.PollActivityTaskQueueResponse) (result any, failureErr error, err error) {
 	traceLog(func() {
 		if t.WorkflowExecution.GetWorkflowId() == "" {
 			ath.logger.Debug("Processing new standalone activity task",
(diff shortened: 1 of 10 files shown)


internal/internal_task_handlers.go
```diff
@@ -2357,7 +2357,7 @@ func newServiceInvoker(
 }
 
 // Execute executes an implementation of the activity.
-func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice.PollActivityTaskQueueResponse) (result any, err error) {
+func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice.PollActivityTaskQueueResponse) (result any, failureErr error, err error) {
```
yuandrew: Don't love the `(any, error, error)` return types in any of these func's. Can we instead have `(activityTaskResult, error)`?

internal/internal_task_handlers.go
```diff
@@ -2469,7 +2469,7 @@ func (ath *activityTaskHandlerImpl) Execute(taskQueue string, t *workflowservice
 	// skip sending another response regardless of what the activity returned.
 	var hbVisitorErr heartbeatVisitorError
 	if errors.As(context.Cause(canCtx), &hbVisitorErr) {
-		return nil, nil
+		return nil, nil, nil
```
yuandrew: oversized heartbeat details bypasses this metric, remind me, was that intentional?

yuandrew (commented):
missing a changelog entry?

internal/internal_task_handlers.go
jmaeagle99: The missing metric application was not intentional. But this particular part of the code is, so that we avoid double emitting the activity task failure.

In this current pass, I'm only fixing the locations that are already emitting the failure metrics. Locations that neglected to emit metrics entirely will be fixed in a separate follow up change.

yuandrew (approved):

PR #2615, opened 2026-08-25
Group unit tests by module

yuandrew:
## What was changed

Run every non-integration module through two unit-test workers, including `internal/cmd/build`. Each module uses `-p 1`. Buffer each worker's output and merge complete worker logs after both finish.

Key unit-test caches with every module's `go.sum`. Keep that key exclusive to unit jobs so check and integration jobs cannot save partial contents first.

## Why?

One process per package repeats build and process overhead. Sequential modules leave a long contrib tail on macOS and Windows. Two workers overlap modules without interleaving package logs.

The first two-worker CI run still downloaded 136 dependencies on macOS Intel despite an exact cache hit. The scoped key tests whether complete unit caches remove that overhead.

## CI experiment

The first run populated the new unit-specific caches. A rerun of the same commit restored them with zero dependency downloads.

| Job | Before | Warm cache |
| --- | ---: | ---: |
| macOS Intel stable | 13:55 | 4:53 |
| macOS Intel oldstable | 10:37 | 6:23 |
| Windows stable | 10:47 | 5:27 |
| Windows oldstable | 7:50 | 5:34 |

The warm macOS ARM oldstable row hit an unrelated 5 ms timing-tolerance flake in `TestWaitServerReady_respectsTimeout`.

## Checklist

1. Closes: N/A

2. How was this tested:

- `go test ./...`
- `go run . unit-test`
- Regression coverage for build-module discovery
- Focused coverage and full-console runs
- Parsed the merged JSON and checked package output remains contiguous
- Workflow diff and whitespace validation
- `go run . check` passes through testifylint; doclink still hits the deferred Go 1.26 generic-method issue

3. Any docs updates needed?

No.

---

> [!NOTE]
> **Medium Risk**
> Changes how all unit tests are discovered and executed (including `internal/cmd/build`), which can affect CI timing, coverage layout, and log ordering; CI cache key behavior also changes for unit jobs.
> 
> **Overview**
> **Unit tests now run one `go test ./...` per Go module** (excluding the integration `test/` tree) instead of spawning a separate test run for every directory that contains a `*_test.go` file. Two worker goroutines process modules in parallel; each module uses **`-p 1`** so package output stays sequential within a module. Per-worker logs are written to temp files and **merged** into the main test log/JSON paths; failure reports and “complete logs” sections use those **final** paths.
> 
> Adds **`findUnitModuleDirs`** (with a unit test) and coverage profiles keyed by **module name** rather than per-directory paths. GitHub failure summary writes are guarded with a **mutex** when workers report failures concurrently.
> 
> **CI:** `setup-go` for unit jobs sets **`cache-dependency-path: '**/go.sum'`** so module caches are populated from a full unit run and not partially warmed by other jobs.
> 
> **Docs:** README wording for `-console-output full` changes from “stream” to “show”.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15e539402679478dc2fed277405a6ef1ea8f3437. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/cmd/build/main.go
+++ b/internal/cmd/build/main.go
@@ -38,6 +38,7 @@ func main() {
 
 const coverageDir = ".build/coverage"
 const defaultTestLogDir = ".build/test-logs"
+const integrationTestModuleDir = "test"
 const unitTestPackageConcurrency = "1"
 const unitTestWorkers = 2
 
@@ -216,6 +217,23 @@ func findModuleDirs(root fs.FS) ([]string, error) {
 	}
 	sort.Strings(moduleDirs)
 	return moduleDirs, nil
+}
+
+func findUnitModuleDirs(root fs.FS) ([]string, error) {
+	moduleDirs, err := findModuleDirs(root)
+	if err != nil {
+		return nil, err
+	}
+
+	// Integration tests have their own command and server setup.
+	unitModuleDirs := make([]string, 0, len(moduleDirs))
+	for _, moduleDir := range moduleDirs {
+		if moduleDir == integrationTestModuleDir || strings.HasPrefix(moduleDir, integrationTestModuleDir+"/") {
+			continue
+		}
+		unitModuleDirs = append(unitModuleDirs, moduleDir)
+	}
+	return unitModuleDirs, nil
 }
 
 func (b *builder) integrationTest() error {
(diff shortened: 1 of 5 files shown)


internal/cmd/build/main.go
```diff
@@ -466,30 +487,166 @@ func (b *builder) unitTest() error {
...
+		if err := appendFile(pair[0], pair[1]); err != nil {
+			return err
+		}
+	}
+	if _, err := worker.stdout.WriteTo(output.stdout); err != nil {
+		return fmt.Errorf("failed writing worker stdout: %w", err)
+	}
+	if _, err := worker.stderr.WriteTo(output.stderr); err != nil {
+		return fmt.Errorf("failed writing worker stderr: %w", err)
+	}
```
dplyukhin, line 628: The Readme says `-console-output full` will stream to stdout instead of the log dump we're doing now. I think we can keep this behavior and just change the wording of the Readme

dplyukhin (approved):

PR #2617, opened 2026-08-25
feat: workflow command pagination

jmaeagle99:
## What was changed

- Add support for paginating workflow task completions when the namespace advertises support.
- Retry pagination with exponential backoff when server report buffer loss.
- Proactively fail workflow task completion on first attempt if commands size is known to be larger than configured namespace command buffer size limit.

## Why?

Enable submitting large batches of workflow commands that would otherwise fail the gRPC data size limit.

## Checklist

1. How was this tested: Unit tests, integration tests.

### DIFF
--- a/internal/common/retry/interceptor.go
+++ b/internal/common/retry/interceptor.go
@@ -11,6 +11,7 @@ import (
 	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/retry"
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/util/backoffutils"
 	errordetailspb "go.temporal.io/api/errordetails/v1"
+	"go.temporal.io/api/serviceerror"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -173,8 +174,13 @@ func IsRetryable(err error, excludeInternalFromRetry *atomic.Bool) bool {
 // IsWorkflowTaskCompletionBufferLost reports whether err carries the server's
 // WorkflowTaskCompletionBufferLostFailure detail, meaning it dropped the buffered pages of a
 // paginated RespondWorkflowTaskCompleted and they must be resent from page 0.
+//
+// serviceerror.ToStatus (not status.Convert) is required because the client's errorInterceptor has
+// already converted the error into a *serviceerror.WorkflowTaskCompletionBufferLost by the time it
+// reaches the resend loop; that type carries the detail on its Status() but not via gRPC's
+// GRPCStatus(), so status.Convert would drop it and report codes.Unknown.
 func IsWorkflowTaskCompletionBufferLost(err error) bool {
-	grpcStatus := status.Convert(err)
+	grpcStatus := serviceerror.ToStatus(err)
 	if grpcStatus == nil {
 		return false
 	}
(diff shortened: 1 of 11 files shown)


internal/common/retry/interceptor.go
```diff
@@ -164,6 +170,22 @@ func IsRetryable(err error, excludeInternalFromRetry *atomic.Bool) bool {
 	return false
 }
 
+// IsWorkflowTaskCompletionBufferLost reports whether err carries the server's
+// WorkflowTaskCompletionBufferLostFailure detail, meaning it dropped the buffered pages of a
+// paginated RespondWorkflowTaskCompleted and they must be resent from page 0.
+func IsWorkflowTaskCompletionBufferLost(err error) bool {
+	grpcStatus := status.Convert(err)
```
yuandrew: I think we want to use `serviceerror.ToConvert` here.  The SDK's `errorInteceptor` changes the server error into a Temporal service error (`*serviceerror.WorkflowTaskCompletionBufferLost`), so this would always produce `codes.Unknown` without the details, preventing the page resend. `serviceerror.ToStatus` handles both forms (it falls back on `status.Convert(err)`)

internal/internal_task_pollers.go
```diff
@@ -631,6 +635,32 @@ func (wtp *workflowTaskProcessor) RespondTaskCompletedWithMetrics(
 		taskCompletion = &workflowTaskCompletion{rawRequest: wtp.errorToFailWorkflowTask(task.TaskToken, taskErr)}
 	}
 
+	// A completion whose recombined command bytes exceed the namespace limit would be rejected and
+	// the workflow terminated by the server, so fail it proactively rather than sending doomed
+	// pages. Only buffered command bytes count toward that limit, not messages or metadata.
+	if req, ok := taskCompletion.rawRequest.(*workflowservice.RespondWorkflowTaskCompletedRequest); ok && wtp.workflowTaskCompletionPagination != nil {
+		if limit := wtp.workflowTaskCompletionPagination.sizeLimit.Load(); limit > 0 {
```
yuandrew, line 648: Rust checks whether pagination is needed before enforcing the command-buffer limit. Go here checks the limit first, which means if the server sets a lower-than-needed limit, we reject the request. For this edge case, it seems the 2 languages differ and something we should align on, behavior wise. 

I could see arguments either way, if we want to respect server advertised limits, even if it would actually be fine without pagination, or the argument that we should just ignore server if it sends us a unnecessarily low limit

jmaeagle99: I'm going with Rust's behavior here. The default limit is ~40 MiB, far above the 4 MiB gRPC limit. It would be an operator error if they lowered the buffer size below the gRPC limit.

internal/common/retry/interceptor.go
```diff
@@ -173,8 +174,13 @@ func IsRetryable(err error, excludeInternalFromRetry *atomic.Bool) bool {
 // IsWorkflowTaskCompletionBufferLost reports whether err carries the server's
 // WorkflowTaskCompletionBufferLostFailure detail, meaning it dropped the buffered pages of a
 // paginated RespondWorkflowTaskCompleted and they must be resent from page 0.
+//
+// serviceerror.ToStatus (not status.Convert) is required because the client's errorInterceptor has
```
yuandrew: hmm, should this be part of the functions docstring or just a comment above the `serviceerror.ToStatus` call itself?

yuandrew (approved):

PR #2621, opened 2026-08-26
Route Workflow polls by per-cell sticky backlog

yuandrew:
## Summary

This adds per-cell sticky-backlog-aware scheduling for floating Workflow polls. Required MCN coverage remains authoritative; backlog only decides normal versus sticky after every group has its required normal and sticky poll.

## Scheduling policy

1. Fill missing normal and sticky coverage for every poller group.
2. For floating capacity, select a group using the latest server weights.
3. Poll sticky when that selected group has positive `backlog_count_hint`; otherwise poll normal.

Normal and sticky keep separate `workflowTaskPoller` RPC objects and separate autoscaler feedback. Their manager loops share one admission coordinator whose budget is:

```text
aggregate target = effective normal target + effective sticky target
```

The component targets contribute to the shared budget rather than cap floating capacity by queue kind. Before poller groups are known, the existing per-kind targets remain in force. `pollerBalancer` still runs before Workflow admission, preserving cross-type balancing without holding a slot while the coordinator waits.

Sticky backlog is stored per group and attributed using the response group ID. A zero hint clears it, and group removal/re-addition creates a fresh group incarnation so stale backlog and pending leases cannot leak across it. A shared pending decision prevents one manager goroutine from discarding or rerolling a decision intended for the other queue kind.

## Key tests

- `TestPollerGroupManagerWorkflowCoverageOverridesStickyBacklog`: missing normal coverage wins even when sticky has excess capacity and backlog.
- `TestPollerGroupManagerFloatingPollSelectsGroupBeforeQueueKind`: group selection happens first and the decision is retained for the matching runner.
- `TestPollerGroupManagerZeroBacklogChoosesNormalFloatingPoll`: zero clears a previous sticky hint.
- `TestPollerGroupManagerRemovedAndReaddedGroupLosesBacklog`: re-added groups do not inherit stale backlog or lease state.
- `TestPollerGroupManagerWeightUpdateReconsidersPendingDecision`: weight changes invalidate an unclaimed floating decision.
- `TestWorkflowStickyBacklogUsesResponsePollerGroupID`: hints use the response group when it differs from the request group.
- `TestAutoscalingWorkflowCoordinatorUsesAggregateCapacityForSelectedKind`: separate targets combine into a shared budget and floating capacity can move to the selected kind.
- `TestAutoscalingWorkflowCoordinatorBacklogUpdateWakesSelectedRunner`: backlog changes wake and reconsider blocked admission.
- `TestAutoscalingWorkflowCoordinatorTargetReductionWaitsForCapacity`: reduced targets do not admit excess polls.
- `TestAutoscalingWorkflowCoordinatorSlotFailureAndShutdownReleaseAdmission`: slot failure and shutdown release both admission and group coverage.

## Validation

- `go run . unit-test -run 'Test(PollerGroupManager|WorkflowStickyBacklog)|TestScalableTaskPollerSuite/TestAutoscalingWorkflowCoordinator|TestWorkerTestSuite/TestWorkerAutoEnrollsDefaultPollersWhenGroupsBecomeKnown'` from `internal/cmd/build` (race enabled)
- `go test ./internal -count=1` (1,186 tests)
- `go run . check` from `internal/cmd/build`

## Current limitation

The repository dev-server fixtures cannot emit MCN poller-group updates or per-cell sticky backlog hints, so the routing behavior is covered with deterministic mocked poll responses and coordinator tests rather than a dev-server integration test. The hint is treated as a binary latest observation for the selected cell; positive chooses sticky and zero chooses normal.

### DIFF
--- a/internal/internal_poller_groups.go
+++ b/internal/internal_poller_groups.go
@@ -33,13 +33,18 @@ type (
 
 	// pollerGroupSnapshot is immutable after publication.
 	pollerGroupSnapshot struct {
-		weights    map[string]float32
-		version    int64
-		versionSet bool
+		weights map[string]float32
+		// generations identifies the current incarnation of each group ID. A
+		// group keeps its generation across weight updates, but receives a new one
+		// when removed and re-added.
+		generations map[string]int64
+		version     int64
+		versionSet  bool
 	}
 
-	// pollerGroupManager is owned by a polling path and combines shared routing
-	// state with its local coverage tracker.
+	// pollerGroupManager combines shared routing state with local poll coverage.
+	// For Workflow polling, it also selects whether normal or sticky may poll
+	// next; workflowPollAdmission enforces their shared admission budget.
 	pollerGroupManager struct {
 		groupInfos *pollerGroupInfoStore
 		tracker    *pollerGroupTracker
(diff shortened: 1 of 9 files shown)


internal/internal_poller_groups.go
```diff
@@ -319,16 +391,58 @@ func (t *pollerGroupTracker) release(group *pollerGroupState, queueKind enumspb.
...
+	snapshot pollerGroupSnapshot,
+	groupID string,
+	backlog int64,
+) bool {
+	if t == nil || t.mode == pollerGroupModeNonWorkflow || groupID == "" {
+		return false
+	}
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	wakeWaiters := t.syncWorkflowSnapshot(snapshot)
```
yuandrew: TODO: can remove?

PR #2626, opened 2026-08-26
feat: Expose workflow cancellation reason

xepozz:
## What was changed

Workflow cancellation now carries a reason:

- `workflow.GetCancellationReason(ctx)` returns the `cause` of the `WorkflowExecutionCancelRequested` event.
- `workflow.RequestCancelExternalWorkflowWithOptions` sends a reason when cancelling an external workflow.
- `internalbindings.WorkflowEnvironment` exposes the reason (`GetCancellationReason`, `reason` parameters on `RequestCancelExternalWorkflow` / `RequestCancelChildWorkflow`) for external SDKs built on top of the Go SDK.
- Test environment: `CancelWorkflowWithReason`, `CancelWorkflowByIDWithReason`, `OnRequestCancelExternalWorkflowWithOptions`.
- The replay matcher for `RequestCancelExternalWorkflowExecution` commands now compares `Reason`.

## Why?

The server records the cancellation reason in history, and `client.CancelWorkflowWithOptions` already sends one, but workflow code had no way to read it or to set one when cancelling another workflow. Closes the workflow-side half of temporalio/features#596; also needed by the PHP SDK, which runs on these bindings through RoadRunner.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,14 @@ to docs, or any other relevant information.
 - The `temporal_activity_execution_failed` and `temporal_local_activity_execution_failed` worker
   metrics now carry a `failure_reason` attribute. Each is now split into one time series per
   reason, which may affect existing dashboards.
+- Workflow cancellation now carries a reason. A workflow reads why it was asked to cancel with
+  `workflow.GetCancellationReason(ctx)` and sets one when cancelling another workflow with
+  `workflow.RequestCancelExternalWorkflowWithOptions`. In the test environment,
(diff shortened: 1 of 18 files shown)


Quinn-With-Two-Ns:
Hi, can we please open a issue to discuss this and what the API should look like before we start any implementation.

xepozz:
Hi, it's already opened: https://github.com/temporalio/features/issues/596

yuandrew:
This is the sdk-go side ticket for the features issue https://github.com/temporalio/sdk-go/issues/1817

Quinn-With-Two-Ns:
The current linked Go or Feature issue do not have any proposed API, that is something we need to align on.

PR #2628, opened 2026-08-27
test: reproducer for terminated workflows retained by the sticky cache

xepozz:
Externally terminated runs keep their full cached state and a parked goroutine until LRU pressure evicts them — the server never dispatches a workflow task on termination, so the worker cannot learn the run is closed. 
300 terminated runs × 1 MB → heap 6.2 → 311 MB, +302 goroutines. 
Includes a probe: WorkflowExecutionTerminated is in history, nothing delivers it to the worker. 

Related: [temporalio/features#573](https://github.com/temporalio/features/issues/573), [temporalio/sdk-rust#1135](https://github.com/temporalio/sdk-rust/issues/1135), [temporalio/sdk-php#635](https://github.com/temporalio/sdk-php/issues/635).

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/test/reproducer/terminatedcache/README.md
+++ b/test/reproducer/terminatedcache/README.md
@@ -1,0 +1,53 @@
+# Reproducer: terminated workflows are retained by the sticky cache until LRU capacity
+
+Externally terminated workflow executions are never evicted from the worker's
+sticky cache. The worker keeps the full cached execution context, including the
+parked workflow goroutine, until `defaultStickyCacheSize` (10000) newer runs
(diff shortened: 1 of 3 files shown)


Quinn-With-Two-Ns:
Yes terminated workflows remain in the cache until evicted that is the expected behaviour, what is the purpose of this PR?

Quinn-With-Two-Ns:
Closing as discussed this is expected behaviour

PR #2637, opened 2026-08-28
💥 feat(extstore): pass a dedicated select context to the driver selector

jmaeagle99:
## What was changed

Pass a dedicated `StorageDriverSelectContext` type to the external storage driver selector instead of reusing `StorageDriverStoreContext`.

## Why?

Allow selector and store contexts to evolve independently.


yuandrew (approved):
💣

PR #2639, opened 2026-08-29
Fix legacy query task failure reporting

### LINKED ISSUE #2616: Legacy query task failures are reported with the query task token on RespondWorkflowTaskFailed
# Legacy query task failures are reported with the query task token on RespondWorkflowTaskFailed

## Expected Behavior

A failure while processing a [legacy query](https://github.com/temporalio/api/blob/3ebdff42a9f07ac484b415fe8ff0b483b4ce3340/temporal/api/workflowservice/v1/request_response.proto#L347) task should be reported on the query
endpoint: `RespondQueryTaskCompleted` with `CompletedType: QUERY_RESULT_TYPE_FAILED`,
echoing the query task token. The caller then receives the real error instead of
waiting out its timeout, and no token is sent to an endpoint that cannot resolve it.

A legacy query task is not a workflow task, so there is no workflow task to fail.

The SDK already does exactly this in one place. `reportGrpcMessageTooLarge`
(`internal/internal_task_pollers.go:777`) type-switches on the pending request and
builds a `RespondQueryTaskCompletedRequest` when the task was a query. The other three
sites that build a failure from the poll-response token should behave the same way.

## Actual Behaviour

The SDK calls `RespondWorkflowTaskFailed` and passes the **query** task token.

`Task` and `QueryTask` are wire-compatible
(`temporal/server/api/token/v1/message.proto`):

```
message Task {                        message QueryTask {
  string namespace_id = 1;              string namespace_id = 1;
  string workflow_id  = 2;  <────────    string task_queue   = 2;
  string run_id       = 3;  <────────    string task_id      = 3;
```

Serialized tokens carry no type tag — the service infers the type from the endpoint
that received them. A `QueryTask` therefore decodes cleanly as a `Task`, with
`task_queue` read as `workflow_id` and `task_id` as `run_id`. `namespace_id` aligns
correctly in both, which is why the lookup reaches the right namespace and returns a
plausible `NotFound` rather than a malformed-token error:

```
rpc error: code = NotFound desc = workflow execution not found
  for workflow ID "/_sys/PHASE_QUEUE/3" and run ID "af23e24e-4a4c-4bfb-9bb5-69cfebba9139"
```

Neither value identifies a workflow. `PHASE_QUEUE` is the task queue;
`/_sys/<queue>/<n>` is the internal name of a non-root partition of it, which no
application can create; the UUID is a query task ID, not a run ID.

**Where it originates.** `internal/internal_task_pollers.go` builds a workflow-task
failure from the raw poll-response token at four sites. Three do not check whether the
task was a query:

| Site | Function | Checks task type |
|---|---|---|
| `:589` | `RespondTaskCompletedWithMetrics`, on a `ProcessWorkflowTask` error | no |
| `:631` | `RespondTaskCompletedWithMetrics`, on an outbound payload visitor error | no |
| `:777` | `reportGrpcMessageTooLarge` | yes |
| `:821` | `handleInboundVisitorError` | no |

`:589` is the site the reproduction below exercises. `:631` and `:821` reach the same
reporting path from payload visitor failures, so codec and external-storage faults
arrive here too.

The query-aware failure response in `completeWorkflow` (`internal_task_handlers.go`,
the `if task.Query != nil` branch) is never reached on this path:
`applyWorkflowPanicPolicy` returns a raw error first under the default `BlockWorkflow`
policy (`internal_task_handlers.go:1342`), which short-circuits `completeWorkflow`
entirely.

**Effect.**

  - The query is never answered. The caller receives no result and times out.
  - The failure is never recorded service-side. History is unchanged — after repeated
    failed query replays the execution's history remains at four events with no
    `WorkflowTaskFailed` — and there is no visibility record. The rejected request
    carries the full failure payload, so a real error with a real stack trace is
    discarded; it survives only in the worker's own logs.
  - The workflow execution is otherwise unaffected. Query tasks are not persisted, so
    nothing is written and no state changes.

**Scope.** Only query tasks are affected.

  1. Start a dev server: `temporal server start-dev`.
  1. Run the program below. It disables the sticky cache so the query arrives as a
     legacy query task, starts a workflow that panics whenever it is replayed, waits
     until no workflow task is outstanding, then queries it once. The panic stands in
     for any replay-time failure — a nil dereference in workflow code, a data converter
     or codec error, a payload that no longer deserializes.
  1. Watch the worker's own log. The panic is reported, and the report is rejected:

     ```
     ERROR Workflow panic ... Error failure while replaying history to serve a query
     WARN  Failed to process workflow task. ... Error failure while replaying history to serve a query
     INFO  Task processing failed with error ... Error workflow execution not found
           for workflow ID "PHASE_QUEUE" and run ID "118f47ff-89cd-48d7-b35b-428ad245f0ad"
     query returned: context deadline exceeded
     ```

  1. Confirm nothing was recorded: `temporal workflow show --workflow-id
     query-token-repro` still shows four events, with no `WorkflowTaskFailed`.

The quoted name varies per query, because the matching client rewrites each query to a
randomly picked write partition and `RpcName()` returns the bare queue name for
partition 0 and `/_sys/<queue>/<n>` for the rest. Repeat the query a few times and both
forms appear; a run of four queries here produced `/_sys/PHASE_QUEUE/3` twice,
`/_sys/PHASE_QUEUE/1` once and `PHASE_QUEUE` once. Both forms are the same defect.

```go
package main

import (
	"context"
	"log"
	"time"

	enumspb "go.temporal.io/api/enums/v1"
	"go.temporal.io/sdk/client"
	"go.temporal.io/sdk/worker"
	"go.temporal.io/sdk/workflow"
)

const taskQueue = "PHASE_QUEUE"

// StuckWorkflow stays open so it can be queried, and fails whenever it is replayed.
// Replay is how a legacy query is answered, so this fails on the query path only.
func StuckWorkflow(ctx workflow.Context) error {
	if err := workflow.SetQueryHandler(ctx, "status", func() (string, error) {
		return "never reached", nil
	}); err != nil {
		return err
	}
	if workflow.IsReplaying(ctx) {
		panic("failure while replaying history to serve a query")
	}
	return workflow.Await(ctx, func() bool { return false })
}

func main() {
	// Size 0 disables the sticky cache, so the query arrives as a legacy query task.
	worker.SetStickyWorkflowCacheSize(0)

	c, err := client.Dial(client.Options{HostPort: client.DefaultHostPort})
	if err != nil {
		log.Fatalln(err)
	}
	defer c.Close()

	// Enough pollers to cover every partition of the task queue (4 by default), so the
	// query sync matches locally and the token usually carries a partition name.
	w := worker.New(c, taskQueue, worker.Options{MaxConcurrentWorkflowTaskPollers: 12})
	w.RegisterWorkflow(StuckWorkflow)
	if err := w.Start(); err != nil {
		log.Fatalln(err)
	}
	defer w.Stop()

	ctx := context.Background()
	we, err := c.ExecuteWorkflow(ctx, client.StartWorkflowOptions{
		ID:                       "query-token-repro",
		TaskQueue:                taskQueue,
		WorkflowIDConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING,
		WorkflowExecutionTimeout: 5 * time.Minute,
	}, StuckWorkflow)
	if err != nil {
		log.Fatalln(err)
	}

	// Wait until no workflow task is outstanding. Otherwise the service buffers the
	// query onto the next real workflow task, where it carries a genuine Task token.
	for {
		desc, err := c.DescribeWorkflowExecution(ctx, we.GetID(), we.GetRunID())
		if err != nil {
			log.Fatalln(err)
		}
		if desc.GetPendingWorkflowTask() == nil {
			break
		}
		time.Sleep(100 * time.Millisecond)
	}

	qctx, cancel := context.WithTimeout(ctx, 10*time.Second)
	defer cancel()
	_, err = c.QueryWorkflow(qctx, we.GetID(), we.GetRunID(), "status")
	log.Printf("query returned: %v", err)
}
```

## Specifications

  - Version: `go.temporal.io/sdk` **v1.46.0** as reported from production, reproduced on
    **v1.48.0**. In the affected file the two releases are identical: the same four call
    sites at lines 589, 631, 777 and 821, in the same functions. The unchecked site in
    `RespondTaskCompletedWithMetrics` is present in every release inspected — v1.30.0,
    v1.31.0, v1.34.0, v1.38.0, v1.42.0, v1.45.0, v1.46.0, v1.48.0 — and over that range
    the number of sites building a workflow-task failure from the poll-response token
    grew from one to four. Source references are pinned to `temporalio/sdk-go`
    `909724e8` (2026-08-22, equal to v1.48.0) and `temporalio/temporal` `6805caea5`
    (2026-08-21).
  - Platform: reproduced on darwin/arm64, Go 1.26, against Temporal CLI 1.8.2 (Server
    1.31.2) with the default four write partitions. Originally reported from Linux
    workers that are recycled frequently, so their sticky cache is routinely cold —
    which is what makes queries arrive on the legacy path in production.

kritibehl:
## What was changed

Legacy query task failures are now reported through `RespondQueryTaskCompleted`
with `QUERY_RESULT_TYPE_FAILED` instead of being sent through
`RespondWorkflowTaskFailed`.

The same query-aware failure handling is applied to:
- workflow task processing failures
- outbound payload visitor failures
- inbound payload visitor failures

Normal workflow tasks continue to use `RespondWorkflowTaskFailed`.

Workflow task failure-cause classification was also separated from request
construction so query failures do not unnecessarily construct a workflow task
failure first.

Regression tests cover all three affected legacy-query failure paths.

## Why?

Legacy query task tokens are query task tokens, not workflow task tokens.

When a legacy query fails during processing, sending its token to
`RespondWorkflowTaskFailed` causes the service to interpret it as a workflow
task token. The query is then left unanswered and eventually times out instead
of returning the actual processing failure.

Reporting the error through `RespondQueryTaskCompleted` allows the query caller
to receive the failure directly.

## Checklist

1. Closes #2616

2. How was this tested:

- Added regression tests for legacy query processing failures.
- Added regression tests for outbound payload visitor failures.
- Added regression tests for inbound payload visitor failures.
- Verified existing workflow task failure-cause behavior.
- Focused tests pass with the race detector.
- `go run . check` passes.
- `go run . unit-test` passes with the race detector.
- Verified against a local Temporal dev server using this SDK checkout. The
  reproduction query now returns `failure while replaying history to serve a
  query` instead of timing out.
- Added a server-backed integration test that fails over to a fresh worker to force replay and verifies the legacy-query replay failure is returned to the caller instead of timing out.

3. Any docs updates needed?

No documentation updates are required. Added a changelog entry under
`Unreleased` / `Fixed`.

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -590,12 +590,11 @@ func (wtp *workflowTaskProcessor) RespondTaskCompletedWithMetrics(
 			tagAttempt, task.Attempt,
 			tagError, taskErr)
 		emitFailMetric = true
-		failWorkflowTask := wtp.errorToFailWorkflowTask(task.TaskToken, taskErr)
 		failureReason = metrics.FailureReasonWorkflowError
-		if failWorkflowTask.Cause == enumspb.WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR {
+		if workflowTaskFailureCause(taskErr) == enumspb.WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR {
 			failureReason = metrics.FailureReasonNonDeterminismError
 		}
-		taskCompletion = &workflowTaskCompletion{rawRequest: failWorkflowTask}
+		taskCompletion = wtp.taskFailureCompletion(task, taskErr)
 	}
 
 	uploadPayloadMetrics := &workflowTaskStorageMetrics{}
@@ -612,7 +611,7 @@ func (wtp *workflowTaskProcessor) RespondTaskCompletedWithMetrics(
 	}
 	if taskErr = visitProtoPayloads(ctx, outboundPayloadVisitor, taskCompletion.rawRequest, wtp.payloadVisitorConcurrency); taskErr != nil {
 		// The outbound visitor failed (e.g. storage driver error or panic). We
-		// cannot send the original response, so fall back to an explicit WFT
+		// cannot send the original response, so fall back to an explicit
 		// failure so the server records the error immediately.
 		keyvals := []any{
 			tagWorkflowType, task.WorkflowType.GetName(),
@@ -632,7 +631,7 @@ func (wtp *workflowTaskProcessor) RespondTaskCompletedWithMetrics(
 		if errors.As(taskErr, new(payloadSizeError)) {
 			failureReason = metrics.FailureReasonPayloadsTooLarge
 		}
-		taskCompletion = &workflowTaskCompletion{rawRequest: wtp.errorToFailWorkflowTask(task.TaskToken, taskErr)}
+		taskCompletion = wtp.taskFailureCompletion(task, taskErr)
 	}
 
 	// The namespace limit governs the server's recombined page buffer, so it only applies to a
@@ -940,18 +939,43 @@ func (wtp *workflowTaskProcessor) handleInboundVisitorError(task *workflowservic
 			tagPayloadSizeLimit, errPayloadSize.limit)
 	}
 	wtp.logger.Warn("Workflow task preprocess error: "+visitErr.Error(), keyvals...)
-	// Submit an explicit WFT failure so the server records the error immediately
+	// Submit an explicit failure so the server records the error immediately
 	// rather than waiting for the task to time out.
-	failReq := wtp.errorToFailWorkflowTask(task.TaskToken, visitErr)
-	if _, submitErr := wtp.sendTaskCompletedRequest(&workflowTaskCompletion{rawRequest: failReq}, task); submitErr != nil {
-		wtp.logger.Warn("Failed to submit WFT failure after inbound visitor error.", tagError, submitErr)
+	failCompletion := wtp.taskFailureCompletion(task, visitErr)
+	if _, submitErr := wtp.sendTaskCompletedRequest(failCompletion, task); submitErr != nil {
+		wtp.logger.Warn("Failed to submit failure after inbound visitor error.", tagError, submitErr)
+	}
+}
+
+func (wtp *workflowTaskProcessor) taskFailureCompletion(
+	task *workflowservice.PollWorkflowTaskQueueResponse,
+	err error,
+) *workflowTaskCompletion {
+	if task.Query != nil {
+		return &workflowTaskCompletion{
+			rawRequest: &workflowservice.RespondQueryTaskCompletedRequest{
+				TaskToken:     task.TaskToken,
+				CompletedType: enumspb.QUERY_RESULT_TYPE_FAILED,
+				ErrorMessage:  err.Error(),
+				Namespace:     wtp.namespace,
+				Failure:       wtp.failureConverter.ErrorToFailure(err),
+			},
+		}
+	}
+
+	return &workflowTaskCompletion{
+		rawRequest: wtp.errorToFailWorkflowTask(task.TaskToken, err),
 	}
 }
 
 func (wtp *workflowTaskProcessor) errorToFailWorkflowTask(taskToken []byte, err error) *workflowservice.RespondWorkflowTaskFailedRequest {
+	return wtp.errorToFailWorkflowTaskWithCause(taskToken, err, workflowTaskFailureCause(err))
+}
+
+func workflowTaskFailureCause(err error) enumspb.WorkflowTaskFailedCause {
 	cause := enumspb.WORKFLOW_TASK_FAILED_CAUSE_WORKFLOW_WORKER_UNHANDLED_FAILURE
 	// If it was a panic due to a bad state machine or if it was a history
-	// mismatch error, mark as non-deterministic
+	// mismatch error, mark as non-deterministic.
 	if panicErr, _ := err.(*workflowPanicError); panicErr != nil {
 		if _, badStateMachine := panicErr.value.(stateMachineIllegalStatePanic); badStateMachine {
 			cause = enumspb.WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR
@@ -963,8 +987,7 @@ func (wtp *workflowTaskProcessor) errorToFailWorkflowTask(taskToken []byte, err 
 	} else if errors.As(err, new(payloadSizeError)) {
 		cause = enumspb.WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE
 	}
-
-	return wtp.errorToFailWorkflowTaskWithCause(taskToken, err, cause)
+	return cause
 }
 
 func (wtp *workflowTaskProcessor) errorToFailWorkflowTaskWithCause(taskToken []byte, err error, cause enumspb.WorkflowTaskFailedCause) *workflowservice.RespondWorkflowTaskFailedRequest {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,8 @@ to docs, or any other relevant information.
 - `temporal.IsWorkflowExecutionAlreadyStartedError` now detects wrapped
   `serviceerror.WorkflowExecutionAlreadyStarted` errors.
 - Malformed Nexus link errors now log the link URL and parse error under stable structured fields.
+- Legacy query task processing failures are now reported through `RespondQueryTaskCompleted` instead of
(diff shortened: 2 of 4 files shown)


internal/internal_task_pollers.go
```diff
@@ -816,18 +815,43 @@ func (wtp *workflowTaskProcessor) handleInboundVisitorError(task *workflowservic
...
+	err error,
+) *workflowTaskCompletion {
+	if task.Query != nil {
+		return &workflowTaskCompletion{
+			rawRequest: &workflowservice.RespondQueryTaskCompletedRequest{
+				TaskToken:     task.TaskToken,
+				CompletedType: enumspb.QUERY_RESULT_TYPE_FAILED,
+				ErrorMessage:  err.Error(),
+				Namespace:     wtp.namespace,
+				Failure:       wtp.failureConverter.ErrorToFailure(err),
```
yuandrew, line 961: Can we set `Cause: workflowTaskFailureCause(err)` here? This helper handles replay and visitor failures, so otherwise nondeterminism and payload-size classifications are dropped.

internal/internal_task_pollers.go
```diff
@@ -816,18 +815,43 @@ func (wtp *workflowTaskProcessor) handleInboundVisitorError(task *workflowservic
...
+	}
+}
+
+func (wtp *workflowTaskProcessor) errorToFailWorkflowOrQueryTask(
+	task *workflowservice.PollWorkflowTaskQueueResponse,
+	err error,
+) *workflowTaskCompletion {
+	if task.Query != nil {
+		return &workflowTaskCompletion{
+			rawRequest: &workflowservice.RespondQueryTaskCompletedRequest{
```
jmaeagle99, line 956: Let's add an integration test similar to what is in #2616. I think that'll truly demonstrate the fix and the interaction with server.

jmaeagle99 (commented):
Thanks for your contribution! Just one piece of feedback from me.

internal/internal_task_pollers.go
```diff
@@ -816,18 +815,43 @@ func (wtp *workflowTaskProcessor) handleInboundVisitorError(task *workflowservic
...
-	failReq := wtp.errorToFailWorkflowTask(task.TaskToken, visitErr)
-	if _, submitErr := wtp.sendTaskCompletedRequest(&workflowTaskCompletion{rawRequest: failReq}, task); submitErr != nil {
-		wtp.logger.Warn("Failed to submit WFT failure after inbound visitor error.", tagError, submitErr)
+	failCompletion := wtp.errorToFailWorkflowOrQueryTask(task, visitErr)
+	if _, submitErr := wtp.sendTaskCompletedRequest(failCompletion, task); submitErr != nil {
+		wtp.logger.Warn("Failed to submit failure after inbound visitor error.", tagError, submitErr)
+	}
+}
+
+func (wtp *workflowTaskProcessor) errorToFailWorkflowOrQueryTask(
```
yuandrew: nit:Can we call this `taskFailureCompletion`? The current name is long and exposes the workflow/query branching instead of the returned abstraction

internal/internal_task_pollers.go
yuandrew: This is technically a part of a larger https://github.com/temporalio/sdk-go/issues/2648, but this PR adds a query-failure path for errors we already classify. Setting Cause here would preserve that classification. I’m also fine handling all paths together in #2648 since it doesn’t affect the timeout fix.

internal/internal_task_pollers.go
kritibehl: Added a server-backed integration test based on #2616. It explicitly fails the workflow over to a fresh worker, forces replay and verifies that the legacy query replay failure is returned to the caller instead of timing out.

internal/internal_task_pollers.go
kritibehl: Done, renamed the helper to taskFailureCompletion.

internal/internal_task_pollers.go
kritibehl: Thanks, I’ll keep Cause handling in #2648 so all failed legacy-query paths are updated consistently. This PR stays focused on fixing the incorrect RPC/timeout behavior.

yuandrew (approved):

kritibehl:
The remaining CI failure is TestOpenTelemetryTracingWithoutMessages, which is unrelated to the legacy-query changes in this PR. I reran that test locally with WORKFLOW_CACHE_SIZE=0 six times and all runs passed. I don't have permission to rerun the failed GitHub Actions job — could someone rerun the failed check when convenient?

PR #2641, opened 2026-08-31
fix: offload oversized query results to external storage

### LINKED ISSUE #2640: External Storage (claim check) never offloads Query responses, even with a StorageDriver configured
## Summary

With `client.Options.ExternalStorage` configured (a `StorageDriver` and a `PayloadSizeThreshold`), a Workflow `Update` handler that returns an over-threshold payload offloads correctly — the driver's `Store()` is invoked and the call succeeds. A `Query` handler returning a same-sized payload does **not** offload: `Store()` is never called, and the call fails with the server's raw payload-size-error response (`TMPRL1103`, "Attempted to upload payloads with size that exceeded the error limit"), because the un-offloaded payload is sent inline instead of being replaced with a claim-check reference.

This makes External Storage unusable for any workflow whose Query handlers can return payloads over the configured threshold — which, since a Query is often how state written via Update is read back, largely defeats the purpose of the feature for state that legitimately needs to exceed Temporal's per-payload limit.

## Environment

- `go.temporal.io/sdk`: reproduced on **v1.47.0** and **v1.48.0** (latest at time of filing) — same behavior on both
- `go.temporal.io/api`: v1.63.4
- Server: Temporal CLI dev server (`temporal server start-dev`), CLI 1.7.1, Server 1.31.0 — also observed against this version
- Go: 1.26

## Expected behavior

A Query response payload at or above `PayloadSizeThreshold` should be offloaded to the configured `StorageDriver` and replaced with a claim-check reference in the outbound `RespondQueryTaskCompletedRequest`, the same way a Workflow Task completion's command/message payloads are — and the client should transparently resolve that reference back to the original payload when reading the query result, which the inbound path already does correctly.

## Actual behavior

The Query response is never offloaded. The driver's `Store()` is never invoked. The raw payload is sent inline and rejected by the server with `TMPRL1103` once it exceeds the server's own blob-size-error limit (~2 MiB by default), regardless of `PayloadSizeThreshold` being configured well below that.

## Minimal reproduction

Requires a reachable Temporal server (`temporal server start-dev` is sufficient).

```go
package main

import (
	"context"
	"crypto/sha256"
	"encoding/hex"
	"fmt"
	"strings"
	"sync/atomic"
	"time"

	"go.temporal.io/sdk/client"
	"go.temporal.io/sdk/worker"
	"go.temporal.io/sdk/workflow"
	"google.golang.org/protobuf/proto"
)

type memDriver struct {
	objects    map[string][]byte
	storeCalls int32
}

func (d *memDriver) Name() string { return "verify" }
func (d *memDriver) Type() string { return "mem" }
func (d *memDriver) Store(ctx converter.StorageDriverStoreContext, payloads []*commonpb.Payload) ([]converter.StorageDriverClaim, error) {
	atomic.AddInt32(&d.storeCalls, 1)
	claims := make([]converter.StorageDriverClaim, len(payloads))
	for i, p := range payloads {
		data, _ := proto.Marshal(p)
		sum := sha256.Sum256(data)
		digest := hex.EncodeToString(sum[:])
		d.objects[digest] = data
		claims[i] = converter.StorageDriverClaim{ClaimData: map[string]string{"digest": digest}}
	}
	return claims, nil
}
func (d *memDriver) Retrieve(ctx converter.StorageDriverRetrieveContext, claims []converter.StorageDriverClaim) ([]*commonpb.Payload, error) {
	out := make([]*commonpb.Payload, len(claims))
	for i, c := range claims {
		p := &commonpb.Payload{}
		proto.Unmarshal(d.objects[c.ClaimData["digest"]], p)
		out[i] = p
	}
	return out, nil
}

func main() {
	driver := &memDriver{objects: map[string][]byte{}}
	c, _ := client.Dial(client.Options{
		HostPort: "127.0.0.1:7233",
		ExternalStorage: converter.ExternalStorage{
			Drivers:              []converter.StorageDriver{driver},
			PayloadSizeThreshold: 1024,
		},
	})
	defer c.Close()

	// Flip this to see the failure vs. the (unrelated-knob) workaround below:
	w := worker.New(c, "repro-tq", worker.Options{ /* DisablePayloadErrorLimit: true */ })
	w.RegisterWorkflow(func(ctx workflow.Context) error {
		workflow.SetQueryHandler(ctx, "get_big", func() (string, error) {
			return strings.Repeat("x", 3*1024*1024), nil // 3 MiB, over Temporal's ~2 MiB ceiling
		})
		return workflow.Await(ctx, func() bool { return false })
	})
	go w.Run(worker.InterruptCh())

	wr, _ := c.ExecuteWorkflow(context.Background(), client.StartWorkflowOptions{ID: "repro-wf", TaskQueue: "repro-tq"}, nil)
	time.Sleep(2 * time.Second)
	fmt.Println("started:", wr.GetID())

	_, err := c.QueryWorkflow(context.Background(), "repro-wf", "", "get_big")
	fmt.Println("QueryWorkflow error:", err)                                    // TMPRL1103 without the workaround; nil with it
	fmt.Println("driver.Store() calls:", atomic.LoadInt32(&driver.storeCalls))  // 0 without the workaround; 1 with it
}
```

Output (unmodified):
```
QueryWorkflow error: [TMPRL1103] Attempted to upload payloads with size that exceeded the error limit.
driver.Store() calls: 0
```

For comparison, calling `UpdateWorkflow` against a handler returning the same-sized payload on the same worker succeeds and `driver.Store()` is invoked once, as expected.

## Root cause

`internal/payload_limits.go`, `payloadLimitsVisitorImpl.ContextHook`:

```go
case *workflowservice.RespondQueryTaskCompletedRequest:
    err := v.checkPayloadSize(int64(msg.GetQueryResult().Size()), limitCheckAll)
    // Server translates too large results into failed query results
    if err != nil {
        msg.ErrorMessage = err.Error()
        msg.QueryResult = nil
        msg.CompletedType = enumspb.QUERY_RESULT_TYPE_FAILED
    }
    ctx = withPayloadLimitChecks(ctx, limitCheckNone)
```

This measures the **raw, pre-offload** size of the query result and, if it exceeds the error limit, destructively clears `QueryResult` — before the storage-offload visitor ever runs.

The worker's outbound payload visitor is a composite chain (`internal_worker.go`, ~line 2443):

```go
outboundPayloadVisitor: newCompositePayloadVisitor(
    extstore.NewExternalStorageVisitor(client.storageParams),
    payloadLimitVisitor,
),
```

`payloadLimitsVisitorImpl.ContextHook` runs *before* the composite chain's `Visit()` step, which is where `extstore.NewExternalStorageVisitor` performs the actual offload. `RespondWorkflowTaskCompletedRequest` (what Update/Continue-As-New completions use) has **no equivalent case** in this same `ContextHook` switch, so its payloads flow straight into `Visit()`: offload happens first (shrinking the payload to a small claim-check reference), and only then does the size check run — inside `Visit`, not `ContextHook` — trivially passing on the now-tiny reference.

Confirmed via source instrumentation (debug logging at the `ContextHook` entry, immediately before/after `checkPayloadSize`, and inside `externalStorageVisitor.Visit`): `QueryResult` is non-nil entering `ContextHook`; `externalStorageVisitor.Visit` is never invoked (zero hits); `QueryResult` is nil immediately after `ContextHook` returns.

## Suggested fix

`ContextHook`'s `RespondQueryTaskCompletedRequest` case should not unconditionally reject on raw size when a storage driver is configured. Two options:

1. Skip this case's error-nilling behavior when `ExternalStorage` is configured, deferring the size decision to the same `Visit()`-time flow every other message type already gets.
2. Move the size-limit check for query results to run *after* offload, symmetrically with how `RespondWorkflowTaskCompletedRequest` is handled today.

Either way, the check should presumably remain as-is (fail fast, client-side) when **no** storage driver is configured — the bug is specifically that a configured, working driver never gets a chance to act.

## Workaround (not a fix, noted for completeness)

`worker.Options.DisablePayloadErrorLimit: true` avoids the destructive path (`checkPayloadSize`'s error branch only fires when `v.errorLimits` has been set, which `DisablePayloadErrorLimit` skips), and the reproduction above succeeds with it enabled. This is not scoped to the bug, though — it disables the proactive client-side size-error pre-check for every outbound payload type, not just queries, so a fix that doesn't require reaching for this broader toggle is strongly preferred.

## Additional notes

- The query-result pre-check in `ContextHook` looks like it predates `ExternalStorage` — plausibly written to proactively replicate a server-side rejection and save a round trip back when no offload mechanism existed at all, and not revisited once External Storage shipped. Worth checking blame/PR history on that case versus when `extstore`/`ExternalStorage` was added to confirm.
- Happy to attempt a PR if a maintainer can confirm the preferred direction (option 1 vs. 2 above, or something else) — I've validated the root cause and the workaround above via source instrumentation, but haven't yet implemented or tested an actual fix against this repo's `ExternalStorage` test tier.

ghaskins:
RespondQueryTaskCompletedRequest and WorkflowQueryResult were size-checked in ContextHook, before the external storage visitor got a chance to run. A query result over the server's blob-size error limit was rejected with TMPRL1103 even when a StorageDriver was configured to handle it, while an identically-sized Update result offloaded correctly.

## What was changed

Move the check to Visit, which runs after external storage in the composite chain, so the size checked reflects what is actually sent. Keeps today's degrade-to-failed-query-result behavior and the no-driver fast-fail path unchanged.

## Checklist

1. Closes #2640

2. How was this tested:

- Reproduction test was added first, confirmed fixed by the changes.  Tests are included with the PR

3. Any docs updates needed?

No


internal/payload_limits.go
```diff
@@ -127,13 +149,44 @@ func (v *payloadLimitsVisitorImpl) Visit(ctx *proxy.VisitPayloadsContext, payloa
...
 
+	// Query results degrade to a failed query result instead of erroring (the server
+	// does the same), so this is checked here rather than in ContextHook: Visit runs
+	// after external storage has had a chance to offload the payloads, so an
+	// oversized-before-offload result whose payloads have since shrunk isn't
+	// incorrectly rejected. ctx.Parent identifies the query result message itself
+	// (see the *querypb.WorkflowQueryResult and *workflowservice.RespondQueryTaskCompletedRequest
+	// cases in ContextHook), which distinguishes the Answer/QueryResult field from
+	// payloads nested in a sibling Failure field on the same message.
+	if target := queryResultFromContext(ctx.Context); target != nil && target == ctx.Parent {
```
jmaeagle99: I don't think the `withQueryResult` and `queryResultFromContext` are necessary for this fix. The Visit callback is only called once on the immediate `WorkflowQueryResult` and `RespondQueryTaskCompletedRequest` message types for their `Answer` and `QueryResult` fields respectively; there are no other immediate child fields of these types that are payload-bearing field types. The `Failure` typed fields are [visited](https://github.com/temporalio/api-go/blob/v1.63.5/proxy/interceptor.go#L1616) at a different level within the callbacks.

Let's say a hypothetical change in the future did add more payload-bearing fields to WorkflowQueryResult` and `RespondQueryTaskCompletedRequest` as immediate child; both the current code and the proposed code would not be able to differentiate the existing fields nor the hypothetically added fields.

So I _think_ these context marker and extract methods don't actually help. Just check that `ctx.Parent` is `WorkflowQueryResult` and `RespondQueryTaskCompletedRequest` and validate their respective payload-bearing fields.

internal/payload_limits_test.go
```diff
@@ -478,6 +480,243 @@ func TestPayloadLimitsVisitorSpecializations(t *testing.T) {
...
+		out[i] = d.data[c.ClaimData["key"]]
+	}
+	return out, nil
+}
+
+// decliningSelector is an extstore.StorageDriverSelector that always leaves the
+// payload inline, simulating a driver selector that declines to store it.
+type decliningSelector struct{}
+
+func (decliningSelector) SelectDriver(extstore.StorageDriverStoreContext, *commonpb.Payload) (extstore.StorageDriver, error) {
```
jmaeagle99: The driver selector will need to be updated due to a recent change in https://github.com/temporalio/sdk-go/commit/4a2268540ecdb512c7f7e9e0715c2d87f9d793ab. Changing the name context from `StorageDriverStoreContext` to `StorageDriverSelectContext` should work.

internal/payload_limits.go
```diff
@@ -127,13 +149,44 @@ func (v *payloadLimitsVisitorImpl) Visit(ctx *proxy.VisitPayloadsContext, payloa
...
+		if err := v.checkPayloadSize(size, limitCheckAll); err != nil {
+			failQueryResult(target, err)
```
jmaeagle99: This introduces a mutation within the visitor callback. Previously, the mutation occurred in the context hook callback, which executed synchronously with the proto message graph walk. Now, this mutation is happening concurrently with other invocations of the same Visit function. It's probably okay since it looks like nothing else is mutating these fields, but I think this deserves another test that sets both `Answer` and `Failure` fields on `WorkflowQueryResult` and sets both `QueryResult` and `Failure` fields on `RespondQueryTaskCompletedRequest` while executing the mutating visitor callbacks concurrently by setting [ConcurrencyLimit](https://github.com/temporalio/api-go/blob/v1.63.5/proxy/interceptor.go#L88) to greater than 1.

internal/payload_limits_test.go
```diff
@@ -478,6 +480,243 @@ func TestPayloadLimitsVisitorSpecializations(t *testing.T) {
 	})
 }
 
+// countingStorageDriver is a minimal in-memory extstore.StorageDriver used to
+// confirm that oversized query results are actually offloaded (Store called,
+// result payload becomes an external storage reference) rather than merely not
+// erroring.
+type countingStorageDriver struct {
+	storeCount int
```
jmaeagle99, line 493: Make this an atomic int when concurrency > 1 is added to some of the requested tests.

internal/payload_limits_test.go
```diff
@@ -478,6 +480,243 @@ func TestPayloadLimitsVisitorSpecializations(t *testing.T) {
...
+
+			require.False(t, tc.isFailed(msg), "query result should not degrade once external storage has offloaded it")
+			require.Empty(t, tc.errMessage(msg))
+			require.Equal(t, 1, driver.storeCount, "driver.Store should be called for the oversized result")
+			p := tc.resultPayload(msg)
+			require.NotNil(t, p)
+			require.True(t, extstore.IsStorageReference(p), "result payload should be replaced with a storage reference")
+		})
+
+		t.Run(tc.name+" still offloads when comfortably under the error limit", func(t *testing.T) {
```
jmaeagle99: Not sure this test adds much value. It really isn't testing anything different than the previous test case because both are offloading to external storage before payload limit validation takes hold.

test/external_storage_test.go
```diff
@@ -350,6 +350,82 @@ func (s *ExternalStorageTestSuite) TestQuery() {
...
+		RunID:        run.GetRunID(),
+		UpdateName:   extStoreQueryAndUpdateName,
+		WaitForStage: client.WorkflowUpdateStageCompleted,
+	})
+	s.NoError(err)
+	var updateResult string
+	s.NoError(handle.Get(ctx, &updateResult))
+	s.Equal(expected, updateResult, "Update result should round-trip through external storage")
+
+	// Query leg: this is the leg that fails before the fix (TMPRL1103, with
```
jmaeagle99: I don't think comments should describe previous behavior because it doesn't matter what the old behavior was. Comments should describe, when it's not obvious or easily understood from context, what the current behavior is. Something like "Query leg: returns over-sized value that external storage should offload before payload limits are validated".

jmaeagle99 (changes requested):

ghaskins:
@jmaeagle99 

Addressed all six points, thanks for the thorough review.

- Collapsed to `switch ctx.Parent.(type)`, dropped `failQueryResult` (folded the mutation directly into `Visit`).
- Fixed the selector to `StorageDriverSelectContext`.
- Added `TestPayloadLimitsVisitorQueryResultConcurrentVisit`: both the result field and a payload-bearing `Failure` set, `ConcurrencyLimit = 8`, passes under `-race`.
- Re: the atomic-int ask — I used a `sync.Mutex` instead, since it also needs to guard the driver's in-memory `data` map, which the concurrent-offload subtest genuinely races on.
- Dropped the redundant "comfortably under the limit" subtest.
- Comments now describe current behavior only, in both the integration test and the internal unit tests.

Also added a case for a `StorageDriver.Store` failure on an oversized query result (surfaces as a workflow task error rather than degrading the query result — noted in the CHANGELOG), and a couple of small follow-up cleanups.

jmaeagle99 (approved):

PR #2643, opened 2026-09-01
Fix: Balance sticky and normal workflow slot admission for Autoscaling

yuandrew:
## What was changed
Give the ability for autoscaling pollers to prefer sticky when there's a sticky backlog, when contending for a WFT slot.

Also combined the generic balancer into our sticky/non-sticky admission controller. The previous balancer could coordinate non-autoscaling poller groups, but no such configuration exists today. Supporting one in the future will require another solution or refactor.

## Why?
Fix a part of autoscaling that I missed when refactoring this to match Rust a few months ago. 😢 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Touches core workflow worker polling, slot reservation, and autoscaling concurrency; mis-tuning could starve one queue kind or change throughput under load, though behavior is heavily tested.
> 
> **Overview**
> Replaces the generic multi-poller `pollerBalancer` with a **`workflowAutoscalingBalancer`** that coordinates **normal** and **sticky** workflow task polls when autoscaling uses split pollers and shared workflow task slots.
> 
> Autoscaling workers with sticky cache enabled now wire both pollers to one balancer (fed by worker **max slots**, sticky **backlog hints**, and the sticky poller’s **autoscaler target**). Poll admission runs **before** poll-capacity acquire and slot reserve so blocked queue kinds do not hold slots or consume poller targets.
> 
> **Admission policy:** keep at least one poll of each kind when capacity is tight; when sticky backlog is meaningfully high and sticky is below its scale target, **favor sticky**; once sticky is at target, **normal polls can use spare slots**. Mixed-mode / simple-maximum pollers are unchanged; multi-poller setups without a shared balancer now **panic** at init.
> 
> CHANGELOG documents the autoscaling sticky-backlog behavior fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa49ff99fea897be5690552793957fe07931a0fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/internal_workflow_slot_admission.go
```diff
@@ -0,0 +1,190 @@
...
+// canAdmit applies the queue-kind policy. The caller must hold mu.
+func (a *workflowSlotAdmission) canAdmit(kind enumspb.TaskQueueKind) bool {
+	switch kind {
+	case enumspb.TASK_QUEUE_KIND_NORMAL:
+		// Always allow a first normal poll.
+		if a.normalActive == 0 {
+			return true
+		}
+		// Preserve capacity for the first sticky poll.
+		if a.stickyActive == 0 && a.normalActive+1 >= a.maxSlots {
```
jmaeagle99: I _think_ this condition cannot be achieved under the default worker configuration. The default worker configuration uses a fixed size tuner with a maximum of 1000 slots and the default auto polling behavior has a maximum of 100 pollers per kind. So this will always return false.

The only conditions under which these checks are effective was when a resource-based tuner returns 0 max slots OR to set MaxConcurrentWorkflowTaskExecutionSize to some relatively small number.

internal/internal_workflow_autoscaling_balancer.go
```diff
@@ -0,0 +1,190 @@
...
+		// Always allow a first normal poll.
+		if a.normalActive == 0 {
+			return true
+		}
+		// Preserve capacity for the first sticky poll.
+		if a.stickyActive == 0 && a.normalActive+1 >= a.maxSlots {
+			return false
+		}
+		// Prefer sticky if there is a sticky backlog
+		if a.needsMoreStickyPolls() {
```
jmaeagle99, line 78: This condition ignores capacity. With defaults of 1000 workflow slots and 100 max pollers per kind, and using an example of a sticky backlog of 10 with 5 active sticky polls, this pins normal to a whatever is the current normal poller count (call it N) when there are 1000 - 5 - N slots still free and normal's own autoscaler would allow up to 100. Normal would only resume once the sticky backlog is at most the stick poll count. This withheld capacity is just idling.

internal/internal_workflow_slot_admission.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+	default:
+		return false
+	}
+
+	return a.normalActive+a.stickyActive < a.maxSlots
+}
+
+func (a *workflowSlotAdmission) needsMoreStickyPolls() bool {
+	return a.stickyBacklog >= minMeaningfulStickyBacklog && a.stickyBacklog > int64(a.stickyActive)
```
jmaeagle99: Could this possible starve normal pollers if the backlog doesn't drain down? Does this need to consider the maximum poller count per kind (default 100) in this calculation? Because if sticky active is at 100, then there's nothing that can be done from a count perspective to help drain the backlog faster. Might as well allow normal pollers to potentially be accepted if stick hits the limit.

internal/internal_workflow_slot_admission.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return &workflowSlotAdmission{
+		maxSlots: maxSlots,
+		wakeCh:   make(chan struct{}),
+	}
+}
+
+// attachPollers configures exactly one autoscaling normal poller and one autoscaling sticky poller.
+// It reports whether the input contained that pair.
+func (a *workflowSlotAdmission) attachPollers(taskPollers []scalableTaskPoller) bool {
```
jmaeagle99: Would it be possible to combine this with `buildWorkflowScalableTaskPollers`? That function is the one deriving the `scalableTaskPoller` list while this function tries to reverse engineer it. Why not create slot admission there or fill out some kind of slot admission data to pass directly to this type?

internal/internal_workflow_autoscaling_balancer.go
```diff
@@ -0,0 +1,190 @@
...
+	for {
+		a.mu.Lock()
+		if a.canAdmit(kind) {
+			a.mu.Unlock()
+			return nil
+		}
+		wakeCh := a.wakeCh
+		a.mu.Unlock()
+
+		select {
```
jmaeagle99, line 57: Add a case for `<-bw.stopCh`? Would allow this to exit earlier during graceful drain.

internal/internal_task_pollers.go
```diff
@@ -1239,6 +1240,14 @@ func (wtp *workflowTaskPoller) updateBacklog(taskQueueKind enumspb.TaskQueueKind
 	wtp.requestLock.Lock()
 	wtp.stickyBacklog = backlogCountHint
 	wtp.requestLock.Unlock()
+
+	if wtp.slotAdmission != nil {
+		wtp.slotAdmission.setStickyBacklog(backlogCountHint)
```
jmaeagle99: Any potential for having inconsistent sticky backlog counts since the poller locks its count separate from the slot admission locking its count?

yuandrew: These are technically for 2 different cases (autoscaling, vs. Mixed poller types), so there is no practical inconsistent sticky backlog state. 

It is very confusing... will change naming and add comments to make this more clear

internal/internal_workflow_autoscaling_balancer.go
yuandrew: not needed, we already apss in `limitedContext` into the `bw.workflowAdmission.waitForAdmission(bw.limiterContext, queueKind)` call, and this gets cancelled immediately after

```
  close(bw.stopCh)
  bw.limiterContextCancel()
```

internal/internal_workflow_slot_admission.go
yuandrew: ohh, good catch. I think yeah we want to allow normal to poll in this case. This is also a bug in the rust impl

internal/internal_workflow_autoscaling_balancer.go
yuandrew: I don't think this is an obvious fix to take, this means normal can start polling while sticky polling is ramping up to its max. Agreed that the current code means sticky has to finish scaling up before normal can start polling, but I imagine that isn't as meaningful/beneficial as the risk of having normal polling during that ramp up window and stealing a slot that would have normally gone to sticky.

internal/internal_workflow_autoscaling_balancer.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return a.normalActive+a.stickyActive < a.maxSlots
+}
+
+func (a *workflowAutoscalingBalancer) needsMoreStickyPolls() bool {
+	// Sticky priority only helps while the scaler can start another sticky poll.
+	return a.stickyBacklog >= minMeaningfulStickyBacklog &&
+		a.stickyBacklog > int64(a.stickyActive) &&
+		int64(a.stickyActive) < a.stickyTarget()
```
jmaeagle99: The `stickyTarget()` function is now part of the wait condition for `canAdmit`. Looks like `stickyTarget` is really `stickyScalablePoller.pollerAutoscaler.target.Load`. The value of that field is mutated by `updateTarget`. Not certain the `wakeCh` signaled when that happens though.

yuandrew: good catch, the `pollerAutoscaler` now passes the new target to the balancer, which now stores it and wakes the normaly/sticky pollers blocked in `waitForAdmission`

internal/internal_workflow_autoscaling_balancer.go
```diff
@@ -0,0 +1,170 @@
...
+		case <-wakeCh:
+		}
+	}
+}
+
+// canAdmit applies the queue-kind policy. The caller must hold mu.
+func (a *workflowAutoscalingBalancer) canAdmit(kind enumspb.TaskQueueKind) bool {
+	// Without a finite limit, only prevent either kind from starting a second
+	// poll before the other kind starts its first.
+	if a.maxSlots <= 0 {
```
jmaeagle99: The resource-based slot supplier sets maxSlots to 0, so when it is used, there is no sticky prioritization. At least, not in the way the rest of the algorithm prioritizes sticky.

internal/internal_worker_base.go
```diff
@@ -422,12 +409,24 @@ func (bw *baseWorker) initializeTaskPollers(taskPollers []scalableTaskPoller) {
...
+
+	// Split workflow pollers must use one balancer so queue kinds cannot starve each other.
+	balancer := taskPollers[0].autoscalingBalancer
+	for _, taskWorker := range taskPollers[1:] {
+		if taskWorker.autoscalingBalancer != balancer {
+			panic(inconsistentWorkflowBalancerMessage)
 		}
 	}
+	if balancer == nil {
+		panic(inconsistentWorkflowBalancerMessage)
```
jmaeagle99, line 428: Probably should have a different message like "workers with multiple task pollers require a poll balancer"

jmaeagle99 (approved):

internal/internal_worker_base.go
```diff
@@ -566,18 +547,25 @@ func (bw *baseWorker) runAutoscalingPoller(taskWorker scalableTaskPoller) {
...
 			return
 		}
-		// Call the balancer before reserving a slot so one poller type cannot
-		// hold slots while waiting for another type to start polling.
-		if bw.pollerBalancer != nil {
-			if bw.pollerBalancer.balance(bw.limiterContext, taskWorker.taskPollerType) != nil {
+		// Unknown capacity only enforces queue-kind fairness. Waiting before
+		// acquire keeps blocked queue kinds out of the poller target.
+		if balancer != nil && !balancer.hasFiniteCapacity() {
+			if err := balancer.waitForAdmission(bw.limiterContext, taskWorker.pollKind); err != nil {
```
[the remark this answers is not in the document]
yuandrew: there will always be a small window, we're matching core's behavior here

internal/internal_workflow_autoscaling_balancer.go
```diff
@@ -0,0 +1,157 @@
...
+	"sync"
+
+	enumspb "go.temporal.io/api/enums/v1"
+)
+
+const (
+	// One queued sticky task does not justify suppressing normal polls.
+	minMeaningfulStickyBacklog          int64 = 2
+	invalidAdmissionKindMessage               = "workflow slot admission requires a normal or sticky queue kind"
+	inconsistentWorkflowBalancerMessage       = "workers with multiple task pollers must share one poll balancer"
```
jmaeagle99, line 15: I should have been more clear. This message is now used in two places. The two places where the panic is raised should not have the same explanation.

yuandrew: lol, looks like me working on a saturday does not provide good results

PR #2644, opened 2026-09-01
Add endpoint-aware serialization for Nexus operations

JoshuaFrenchwood:
## What was changed

Added caller side `converter.NexusSerializationContext` support for workflow-scheduled Nexus operations.
This context gives the Nexus Endpoint, Service, and operation name to the data/failure converters

## Why?

This enables codecs to select serialization behavior or encryption keys by Nexus endpoint, service, or operation.

For example, workflows calling two Nexus endpoints can encrypt each endpoint’s payloads with a different key while ensuring that inputs, results, and failures are decoded with the converter selected for the corresponding operation.

## Checklist

1. Closes 

2. How was this tested:

Added unit/functional tests to verify that NexusSerializationContext works as expected.

3. Any docs updates needed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,11 @@ to docs, or any other relevant information.
   platforms), and the operating system and architecture. This is sent
   once per worker with the first heartbeat accepted by the server and can be turned off with
   `client.Options.DisableWorkerEnvironmentInfo`.
+- Added caller-side `converter.NexusSerializationContext` support for Nexus operation inputs,
+  results, and failures. Each scheduled operation retains an isolated contextual data and failure
+  converter keyed by its endpoint, service, and resolved operation name, including across replay.
+  Nexus handlers do not yet receive this context and must use compatible static converter
+  configuration for their endpoint/task queue.
 - Added `temporal.NewPayloadValidationError` to create non-retryable application errors with
   optional structured details for payload validation failures. Passing `nil` omits details.
 - Added Go 1.27+ generic methods on the experimental `temporalnexus.NexusClient` for starting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,10 +29,10 @@ to docs, or any other relevant information.
   once per worker with the first heartbeat accepted by the server and can be turned off with
   `client.Options.DisableWorkerEnvironmentInfo`.
 - Added caller-side `converter.NexusSerializationContext` support for Nexus operation inputs,
-  results, and failures. Each scheduled operation retains an isolated contextual data and failure
-  converter keyed by its endpoint, service, and resolved operation name, including across replay.
-  Nexus handlers do not yet receive this context and must use compatible static converter
-  configuration for their endpoint/task queue.
+  results, and failures. Workflow-scheduled and standalone operations retain isolated contextual
+  data and failure converters keyed by endpoint, service, and resolved operation name, including
+  across workflow replay and standalone operation handles. Nexus handlers do not yet receive this
+  context and must use compatible static converter configuration for their endpoint/task queue.
 - Added `temporal.NewPayloadValidationError` to create non-retryable application errors with
   optional structured details for payload validation failures. Passing `nil` omits details.
 - Added Go 1.27+ generic methods on the experimental `temporalnexus.NexusClient` for starting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,11 +28,12 @@ to docs, or any other relevant information.
   platforms), and the operating system and architecture. This is sent
   once per worker with the first heartbeat accepted by the server and can be turned off with
   `client.Options.DisableWorkerEnvironmentInfo`.
-- Added caller-side `converter.NexusSerializationContext` support for Nexus operation inputs,
-  results, and failures. Workflow-scheduled and standalone operations retain isolated contextual
-  data and failure converters keyed by endpoint, service, and resolved operation name, including
-  across workflow replay and standalone operation handles. Nexus handlers do not yet receive this
-  context and must use compatible static converter configuration for their endpoint/task queue.
+- Added `converter.NexusSerializationContext` support for Nexus operation inputs, results, and
+  failures. Workflow-scheduled and standalone callers retain isolated contextual data and failure
+  converters keyed by endpoint, service, and resolved operation name, including across workflow
+  replay and standalone operation handles. Nexus handlers receive the same context when decoding
+  inputs, encoding synchronous results, and encoding failures produced while handling a Nexus task.
+  The context is not yet propagated to asynchronous operation results.
 - Added `temporal.NewPayloadValidationError` to create non-retryable application errors with
   optional structured details for payload validation failures. Passing `nil` omits details.
 - Added Go 1.27+ generic methods on the experimental `temporalnexus.NexusClient` for starting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,9 +31,11 @@ to docs, or any other relevant information.
 - Added `converter.NexusSerializationContext` support for Nexus operation inputs, results, and
   failures. Workflow-scheduled and standalone callers retain isolated contextual data and failure
   converters keyed by endpoint, service, and resolved operation name, including across workflow
-  replay and standalone operation handles. Nexus handlers receive the same context when decoding
-  inputs, encoding synchronous results, and encoding failures produced while handling a Nexus task.
-  The context is not yet propagated to asynchronous operation results.
+  replay and standalone operation handles returned by `ExecuteOperation`. Nexus handlers receive
+  the same context when decoding inputs, encoding synchronous results, and encoding failures
+  produced while handling a Nexus task. The context is not propagated to asynchronous operation
+  results. Standalone handles use the context of their start request, including when an existing
+  operation is returned, while handles created without starting an operation do not receive it.
 - Added `temporal.NewPayloadValidationError` to create non-retryable application errors with
   optional structured details for payload validation failures. Passing `nil` omits details.
 - Added Go 1.27+ generic methods on the experimental `temporalnexus.NexusClient` for starting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,14 +28,10 @@ to docs, or any other relevant information.
   platforms), and the operating system and architecture. This is sent
   once per worker with the first heartbeat accepted by the server and can be turned off with
   `client.Options.DisableWorkerEnvironmentInfo`.
-- Added `converter.NexusSerializationContext` support for Nexus operation inputs, results, and
-  failures. Workflow-scheduled and standalone callers retain isolated contextual data and failure
-  converters keyed by endpoint, service, and resolved operation name, including across workflow
-  replay and standalone operation handles returned by `ExecuteOperation`. Nexus handlers receive
-  the same context when decoding inputs, encoding synchronous results, and encoding failures
-  produced while handling a Nexus task. The context is not propagated to asynchronous operation
-  results. Standalone handles use the context of their start request, including when an existing
-  operation is returned, while handles created without starting an operation do not receive it.
+- Added `converter.NexusSerializationContext` support for Nexus callers and handlers. Callers use
+  it for inputs, results, and failures; handlers use it for inputs, synchronous results, and
+  failures. Asynchronous handler results and detached standalone handles are not yet supported.
+  Standalone `USE_EXISTING` handles use their start request's context.
 - Added `temporal.NewPayloadValidationError` to create non-retryable application errors with
   optional structured details for payload validation failures. Passing `nil` omits details.
 - Added Go 1.27+ generic methods on the experimental `temporalnexus.NexusClient` for starting
--- a/converter/serialization_context.go
+++ b/converter/serialization_context.go
@@ -2,7 +2,8 @@ package converter
 
 // SerializationContext provides metadata about where serialization is occurring.
 // Implementations include [WorkflowSerializationContext] for workflow-level
-// payloads, and [ActivitySerializationContext] for activity-level payloads.
+// payloads, [ActivitySerializationContext] for activity-level payloads, and
+// [NexusSerializationContext] for caller-side Nexus operation payloads.
 type SerializationContext interface {
 	isSerializationContext()
 }
@@ -33,6 +34,25 @@ type ActivitySerializationContext struct {
 }
 
 func (ActivitySerializationContext) isSerializationContext() {}
+
+// NexusSerializationContext is the serialization context for caller-side Nexus
+// operation input, successful result, and failure decoding. Operation summaries
+// and user metadata use the workflow serialization context instead.
+//
+// Operation is the resolved operation name. This context is not propagated to
+// Nexus handlers, which must use compatible converter configuration.
+//
+// For failures, this context is provided only to the workflow caller's
+// [FailureConverter.FailureToError], not [FailureConverter.ErrorToFailure].
+// Implementations must not assume symmetric failure conversion. Context-dependent
+// encodings should be self-describing and support legacy payloads without context.
+type NexusSerializationContext struct {
+	Endpoint  string
+	Service   string
+	Operation string
+}
+
+func (NexusSerializationContext) isSerializationContext() {}
 
 // DataConverterWithSerializationContext is an optional interface that [DataConverter]
 // implementations can implement to receive serialization context.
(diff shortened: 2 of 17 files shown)


converter/serialization_context.go
```diff
@@ -2,7 +2,8 @@ package converter
 
 // SerializationContext provides metadata about where serialization is occurring.
 // Implementations include [WorkflowSerializationContext] for workflow-level
-// payloads, and [ActivitySerializationContext] for activity-level payloads.
```
chrsmith, line 5: Minor nit: since there are 3x different things, maybe we should use a list rather than a really long sentence? e.g.

```
// SerializationContext provides metadata about where serialization is occurring,
// with the concrete type depending on the context.
// - [WorkflowSerializationContext] for workflow-level payloads
// - [ActivitySerializationContext] for activity-level payloads
// - [NexusSerializationContext] for caller-side Nexus operation payloads
```

Also, what is a "caller-side Nexus operation payload"? Would it be more accurate or clearer to say "Nexus handlers" instead?

converter/serialization_context.go
```diff
@@ -34,6 +35,25 @@ type ActivitySerializationContext struct {
 
 func (ActivitySerializationContext) isSerializationContext() {}
 
+// NexusSerializationContext is the serialization context for caller-side Nexus
+// operation input, successful result, and failure decoding. Operation summaries
+// and user metadata use the workflow serialization context instead.
```
chrsmith: Is the typical reader of this code supposed to know what "Operation summaries" or "user metadata" are referring to? It's probably benign, but if we expect different _fields_ of the same proto to be ran through _different serializers_, that sounds like a bad thing?

internal/internal_event_handlers.go
```diff
@@ -2088,7 +2094,7 @@ func (weh *workflowExecutionEventHandlerImpl) handleNexusOperationCompleted(even
 	state := command.getData().(*scheduledNexusOperation)
```
chrsmith: Nit: Is there a better name we can use here than `state`? We are essentially casting the command's data to the underlying "event type", right? Would something like `event` or `eventSource` be clearer?

internal/internal_workflow_testsuite.go
```diff
@@ -2780,6 +2780,10 @@ func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(
 	callback func(*commonpb.Payload, error),
 	startedHandler func(opID string, e error),
 ) int64 {
+	failureConverter := params.failureConverter
```
chrsmith: Nit: This is perfectly fine as-is, but you can use `cmp.Or(params.failureConverter, env.failureConverter)` to simplify this type of initialization.
https://pkg.go.dev/cmp#Or

internal/workflow.go
```diff
@@ -3136,6 +3134,14 @@ func (wc *workflowEnvironmentInterceptor) prepareNexusOperationParams(ctx Contex
 		return ExecuteNexusOperationParams{}, fmt.Errorf("invalid 'operation' parameter, must be an OperationReference or a string")
 	}
 
+	nsc := converter.NexusSerializationContext{
```
chrsmith, line 3138: Nit: For readability, maybe add a comment to group these things? What do you think of:

```
// Nexus context and data converters.
```

test/serialization_context_test.go
```diff
@@ -133,3 +140,240 @@ func (ts *IntegrationTestSuite) TestSerializationContext_EndToEnd() {
...
+	err    error
+}
+
+type intTestNexusHMACCodec struct {
+	key []byte
+}
+
+const intTestNexusHMACEncoding = "binary/nexus-hmac-test"
+
+func intTestNexusContextKey(endpoint, service, operation string) string {
```
chrsmith: Nit: Is "intTest" short for integration test? If so, maybe its all fine as-is. But IMHO, something like `nexusContextToCodecKey` might be clearer?

test/serialization_context_test.go
```diff
@@ -133,3 +140,240 @@ func (ts *IntegrationTestSuite) TestSerializationContext_EndToEnd() {
...
+	defer callerWorker.Stop()
+
+	run, err := callerClient.ExecuteWorkflow(ctx, client.StartWorkflowOptions{
+		ID:        "nexus-ser-ctx-caller-" + uuid.NewString(),
+		TaskQueue: callerTaskQueue,
+	}, intTestNexusCallerWorkflow, hmacEndpointName, zlibEndpointName)
+	ts.NoError(err)
+	var result string
+	ts.NoError(run.Get(ctx, &result))
+	ts.ElementsMatch([]string{"hmac", "zlib"}, strings.Split(result, "|"))
```
chrsmith: For readability, could we have the workflow expose a proper type rather than a string?

We want to verify that the decoded results match, but It would be great if we could also look at the raw, encoded results between the two codecs and confirm that they do NOT match. (i.e. the hmac and zlib encoders are using two different encodings for the same data.)

Is there any way we can easily verify that? Or would we need to have some sort of state smuggled out of the `intTestNexusCodecSelector`?

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ to docs, or any other relevant information.
 
 ### Added
 
+- Added caller-side `converter.NexusSerializationContext` support for Nexus operation inputs,
+  results, and failures. Each scheduled operation retains an isolated contextual data and failure
```
chrsmith, line 32: How is this supposed to work in the long run?

- This is wired up for Nexus invocations made from workflows, but what about SANOs?
- If the handler-side is supposed to use the same converter, how will this actually work?

I'm just trying to understand the next few steps in getting this all working end-to-end.

chrsmith (commented):
Lots of nitpicks and some questions. But this all looks right. (Though I'd defer approval to someone with a little more background on the SDKs and/or stronger opinions about how any newly exported types/methods should look.)

converter/serialization_context.go
JoshuaFrenchwood: Thanks, I updated this to be more clear.  By caller side, I meant workflows that are calling nexus operations.

converter/serialization_context.go
JoshuaFrenchwood: I removed that last sentence and made this more readable.

internal/internal_event_handlers.go
JoshuaFrenchwood: I renamed it scheduleNexusOperation

internal/internal_workflow_testsuite.go
JoshuaFrenchwood: Nice! used that instead thanks

test/serialization_context_test.go
JoshuaFrenchwood: Yeah intTest is integration test, I renamed it to "intTestNexusCodecKey"

test/serialization_context_test.go
JoshuaFrenchwood: Great point. I updated this workflow to return a struct, and ts.Equal for the final result. Also at the end of the test I look in the History to verify that the encoded data does not match.

CHANGELOG.md
JoshuaFrenchwood: Yes the long term is to have SANO support. This PR starts with Nexus calls from in workflows because the Nexus Context (service, endpoint, operation) are present.  Adding SANO support and handler side will require server changes, because that context is not currently returned when a SANO result or failure is polled, or consistently propagated to Temporal entities started by Nexus handlers.

chrsmith (approved):

converter/serialization_context.go
```diff
@@ -34,6 +37,24 @@ type ActivitySerializationContext struct {
...
+// Implementations must not assume symmetric failure conversion. Context-dependent
+// encodings should be self-describing and support legacy payloads without context.
+type NexusSerializationContext struct {
+	Endpoint  string
+	Service   string
+	Operation string
+}
+
+func (NexusSerializationContext) isSerializationContext() {}
+
```
mavemuri, line 66: minor nit: I see the other 2 also dont have it - but maybe add a `_ SerializationContext = NexusSerializationContext{}` iiuc its the convention in this repo and is helpful whenever things change

internal/internal_event_handlers.go
```diff
@@ -667,13 +668,18 @@ func (wc *workflowEnvironmentImpl) ExecuteNexusOperation(params ExecuteNexusOper
 	}
 
 	command := wc.commandsHelper.scheduleNexusOperation(seq, scheduleTaskAttr, startMetadata)
+	failureConverter := params.failureConverter
+	if failureConverter == nil {
+		failureConverter = wc.failureConverter
```
mavemuri, line 675: Question, not a blocker: should it be a wrapper around some default system converter? eg. convert/trap some known internal failure types before passing onto provided converter. Maybe not specific to this PR but just an overall question

mavemuri (approved):
LGTM!

converter/serialization_context.go
```diff
@@ -1,8 +1,11 @@
...
 
-// SerializationContext provides metadata about where serialization is occurring.
-// Implementations include [WorkflowSerializationContext] for workflow-level
-// payloads, and [ActivitySerializationContext] for activity-level payloads.
+// SerializationContext provides metadata about where serialization is occurring,
+// with the concrete type depending on the context:
+//
+//   - [WorkflowSerializationContext] for workflow-level payloads.
+//   - [ActivitySerializationContext] for activity-level payloads.
+//   - [NexusSerializationContext] for workflows calling Nexus operations.
```
Quinn-With-Two-Ns: I would keep it consistent
```suggestion
//   - [NexusSerializationContext] for nexus-level payloads.
```
since this context would also apply for SANO and on the handler

converter/serialization_context.go
```diff
@@ -34,6 +37,24 @@ type ActivitySerializationContext struct {
...
+// encoding operation inputs and decoding successful results or failures.
+//
+// Operation is the resolved operation name. This context is not propagated to
+// Nexus handlers, which must use compatible converter configuration.
+//
+// For failures, this context is provided only to the workflow caller's
+// [FailureConverter.FailureToError], not [FailureConverter.ErrorToFailure].
+// Implementations must not assume symmetric failure conversion. Context-dependent
+// encodings should be self-describing and support legacy payloads without context.
+type NexusSerializationContext struct {
```
Quinn-With-Two-Ns, line 56: Can we apply this context to SANO operations as well ?

Quinn-With-Two-Ns, line 56: Please mark this as experimental

converter/serialization_context.go
```diff
@@ -34,6 +37,24 @@ type ActivitySerializationContext struct {
...
+	Endpoint  string
+	Service   string
+	Operation string
```
Quinn-With-Two-Ns, line 62: Please add struct comments for these, we try to document all public fields

converter/serialization_context.go
```diff
@@ -34,6 +37,24 @@ type ActivitySerializationContext struct {
 
 func (ActivitySerializationContext) isSerializationContext() {}
 
+// NexusSerializationContext is used by workflows calling Nexus operations when
+// encoding operation inputs and decoding successful results or failures.
+//
+// Operation is the resolved operation name. This context is not propagated to
```
Quinn-With-Two-Ns: So a few things here:
* I think we should pass this to the handler, the issue right now is we can't pass it to async operations the handler starts, that is the real limitation right now
* That being said if we want to push the handler support to another PR I am okay with that

internal/internal_workflow_testsuite.go
```diff
@@ -2780,6 +2781,7 @@ func (env *testWorkflowEnvironmentImpl) ExecuteNexusOperation(
 	callback func(*commonpb.Payload, error),
 	startedHandler func(opID string, e error),
 ) int64 {
+	failureConverter := cmp.Or(params.failureConverter, env.failureConverter)
```
Quinn-With-Two-Ns, line 2784: When can `env.failureConverter` ever be `nil`?

CHANGELOG.md
Quinn-With-Two-Ns: IMO we should add SANO support in this PR, it doesn't require any server side changes. handler support also doesn't require any server side changes so we could do it, but that I would be fine deferring to a separate PR. The thing we can't support is Async operations that will require server work

converter/serialization_context.go
JoshuaFrenchwood: Updated

converter/serialization_context.go
JoshuaFrenchwood: Added SANO support for this

converter/serialization_context.go
JoshuaFrenchwood: Added comments

CHANGELOG.md
Quinn-With-Two-Ns: Technically some fields may not be present when we create the context when getting the SANO result, but I think that is a totally fine tradeoff, since:

1. The main/only use case is really just for the input
2. If the callee did something async the result won't have the context either

converter/serialization_context.go
JoshuaFrenchwood: Thanks, Ill add handler support here

Quinn-With-Two-Ns: Thanks, can you update the code comment here as well

internal/internal_nexus_client.go
```diff
@@ -726,11 +746,17 @@ func (w *workflowClientInterceptor) ExecuteNexusOperation(
 	if err != nil {
 		return nil, err
 	}
+	var nsc *converter.NexusSerializationContext
+	if resp.Started || in.Options.IDConflictPolicy != enumspb.NEXUS_OPERATION_ID_CONFLICT_POLICY_USE_EXISTING {
```
Quinn-With-Two-Ns: What is this check for?

internal/internal_nexus_client.go
```diff
@@ -726,11 +746,17 @@ func (w *workflowClientInterceptor) ExecuteNexusOperation(
...
+	}
 
 	return &clientNexusOperationHandleImpl{
-		client: w.client,
-		id:     in.Options.ID,
-		runID:  resp.RunId,
+		client:                    w.client,
+		id:                        in.Options.ID,
+		runID:                     resp.RunId,
+		resolvedRunID:             resp.RunId,
```
Quinn-With-Two-Ns: Why did we need to add this? Is it related to serialization contexts or is it for something else

internal/internal_workflow_testsuite.go
JoshuaFrenchwood: env.failureConverter should not be nil.  params.failureConverter can be nil when the params are created through NewExecuteNexusOperationParams and passed directly to env.ExecuteNexusOperation(params, completedCallback, startedCallback), so env.failureConverter is used as the fallback.

internal/internal_nexus_client.go
JoshuaFrenchwood: This described whether or not the .Get method must call Describe to retrieve the operation's actual endpoint/service/operation name.  USE_EXISTING policy can return a running operation with the same namespace + operationId, but a differernt nexus endpoint/service/operation name. When USE_EXISTING is set and resp.Started is false (a running operation has been found), the nexus context is unset and Describe is called before decoding.

Quinn-With-Two-Ns: I think adding a describe call on every get is quite heavy when it is unlikely anyone will use this information there, I agree that in theory the endpoint/service/operation name could be different but I don't think it is worth trying to handle here. A similar edge case is possible for the other serilziation contexts as well, but we are not worried about them there so I don't think we need to here.

internal/internal_nexus_client.go
JoshuaFrenchwood: ResolvedRunId ensures the polling targets the same run whos serialization context is was retrieved by Describe.   If the IDReusePolicy is ALLOW_DUPLICATES, a newer run with the same OperationId may exist before polling begins.  Specifying the RunId prevents that polling from selecting the just the latest.

internal/internal_nexus_client.go
JoshuaFrenchwood: I see, Ill remove it. I think we can address this edge case after I make server changes to retrieve NexusSerializationContext.

internal/internal_nexus_client.go
JoshuaFrenchwood: Removed this since we arent using Describe anymore

CHANGELOG.md
JoshuaFrenchwood: Added support for SANO and Handler

Quinn-With-Two-Ns (approved):

PR #2646, opened 2026-09-01
Standalone Activities operator commands and optional payloads

### LINKED ISSUE #2350: Implement operator commands for Standalone Activities
See https://github.com/temporalio/features/issues/822

GregoryTravis:
# Add operator commands for standalone activities

Adds pause, unpause, reset, and update-options to standalone activities, plus the
describe surface needed to observe their effects.

## API

On `client.ActivityHandle`:

```go
handle.Pause(ctx, client.PauseActivityOptions{Reason: "..."})
handle.Unpause(ctx, client.UnpauseActivityOptions{Reason: "...", Jitter: d})
handle.Reset(ctx, client.ResetActivityOptions{
    KeepPaused: bool, Jitter: d, RestoreOriginalOptions: bool, ResetHeartbeat: bool,
})
handle.UpdateOptions(ctx, updates ...client.ActivityOptionsUpdate) (*client.ActivityOptions, error)
handle.RestoreOriginalOptions(ctx) (*client.ActivityOptions, error)
```

### Optional payloads

`client.DescribeActivityOptions` gains four opt-ins, all defaulting to false:
`IncludeInput`, `IncludeOutcome`, `IncludeHeartbeatDetails`, `IncludeLastFailure`.
Payloads are only fetched when asked for, and anything the caller did not request
is stripped client-side in case an older or buggy server returns it anyway.

## Tests

- 15 functional subtests against a real dev server, covering each command, its
  observable effect, and interaction with pause state.
- Unit tests for the parts the server never echoes back and so cannot be observed
  from activity state: field-mask construction, the set/unset/repeated-key
  semantics above, reset flags reaching the wire, jitter being omitted rather than
  sent as an explicit zero, and client-side payload stripping.


yuandrew:
Is there a GH issue for this? Can we link it?

GregoryTravis:
> Is there a GH issue for this? Can we link it?

Linked it under Development

client/client.go
```diff
@@ -1032,6 +1031,35 @@ type (
...
+	// UnpauseActivityOptions contains options for ClientActivityHandle.Unpause call.
+	//
+	// NOTE: Experimental
+	UnpauseActivityOptions = internal.ClientUnpauseActivityOptions
+
+	// ActivityOptions describes the options an activity is currently running with, as returned
+	// by ActivityHandle.UpdateOptions and ActivityHandle.RestoreOriginalOptions.
+	//
+	// NOTE: Experimental
+	ActivityOptions = internal.ClientActivityOptions
```
maciejdudko: The name is too generic and may cause confusion with `workflow.ActivityOptions` and `client.StartActivityOptions`.

internal/internal_activity_client.go
```diff
@@ -228,6 +257,157 @@ type (
...
+	//
+	// Exposed as: [go.temporal.io/sdk/client.ActivityOptionsUpdate]
+	ClientActivityOptionsUpdate struct {
+		// name is the field-mask path being updated.
+		name string
+		// apply writes the value onto the request's ActivityOptions. A nil apply means the
+		// option is being cleared: its path is named in the mask while the field is left absent.
+		apply func(*activitypb.ActivityOptions)
+	}
+)
```
maciejdudko: This API is very unidiomatic for Go. I think it would be better to just have a simple struct

```
ClientActivityOptionsUpdate struct {
	TaskQueue              *string
	ScheduleToCloseTimeout *time.Duration
	ScheduleToStartTimeout *time.Duration
	StartToCloseTimeout    *time.Duration
	HeartbeatTimeout       *time.Duration
	StartDelay             *time.Duration
	RetryPolicy            *RetryPolicy
	Priority               *Priority
}
```

where nil is keep current value, and zero is clearing the value.

internal/internal_activity_client.go
```diff
@@ -269,10 +450,17 @@ type (
 		ClientActivityExecutionInfo
 		// Raw PB message this struct was built from.
 		RawExecutionInfo        *activitypb.ActivityExecutionInfo
+		RawDescription          *workflowservice.DescribeActivityExecutionResponse
```
maciejdudko: 1. Do not call it `RawDescription` - it's the whole server response, it contains things that are not actually part of the description (e.g. long poll key). Rename it to `RawResponse`.
2. Now that we are exposing this field, `RawExecutionInfo` is redundant and can be removed.

internal/internal_activity_client.go
```diff
@@ -317,8 +507,80 @@ func (d *ClientActivityExecutionDescription) GetHeartbeatDetails(valuePtrs ...an
...
+	return d.dataConverter.FromPayloads(outcome.Result, valuePtr)
+}
+
+// GetFailure returns the failure the activity closed with, using the failure converter of the
+// client used to make the Describe call. Returns nil if the activity did not fail, or if the
+// outcome was not requested via ClientDescribeActivityOptions.IncludeOutcome.
+//
+// This is the terminal failure of the execution. It differs from GetLastFailure, which reports
+// the failure of the most recent attempt of an activity that may still be retrying.
+func (d *ClientActivityExecutionDescription) GetFailure() error {
```
maciejdudko: 1. Rename to `GetOutcomeFailure` to differentiate from `GetLastFailure`.
2. Add `HasOutcomeFailure` method.

internal/interceptor.go
```diff
@@ -731,6 +747,67 @@ type ClientTerminateActivityInput struct {
...
+// NOTE: Experimental
+//
+// Exposed as: [go.temporal.io/sdk/interceptor.ClientPauseActivityInput]
+type ClientPauseActivityInput struct {
+	// ActivityID is the ID of the activity.
+	ActivityID string
+	// RunID is the run ID of the activity to pause.
+	RunID string
+	// Reason is the reason for pausing.
+	Reason string
```
maciejdudko: Store the entire Options as a single field instead of copying individual fields. Same for `ClientUnpauseActivityInput` and `ClientDescribeActivityInput`.

maciejdudko (changes requested):

internal/internal_activity_client.go
GregoryTravis: Done, but I added another level of indirection, so we can distinguish between clearing a field and setting it to the zero value, like WorkflowExecutionOptionsChanges.

maciejdudko (approved):

PR #2647, opened 2026-09-01
Report runtime environment information in worker heartbeats

Sushisource:
## What was changed

Worker heartbeats now include an `EnvironmentInfo` message describing the Go runtime version, detected hosting environments (Docker, Kubernetes, AWS Lambda/ECS, Google Cloud Run/App Engine, Azure App Service/Functions/Container Apps), and the OS platform (Linux/macOS/Windows version and architecture).

The information is attached to each worker's heartbeats until the server accepts one, then omitted. Delivery state is tracked in the shared-namespace heartbeat worker and cleared only after a successful `RecordWorkerHeartbeat`.

A new experimental `client.Options.DisableWorkerEnvironmentInfo` turns the reporting off.

## Why?

Port of Core's temporalio/sdk-rust#1457 so all SDKs report the same analytical information about the environments they run in.

## Checklist

1. Closes N/A

2. How was this tested:

New unit tests cover environment detection (runtime, hosting-environment env vars, cgroup parsing), retry-until-accepted delivery, and the disable option. `go vet` and the full `./internal` package pass, and the per-OS platform files cross-compile for linux, darwin, windows, and other.

3. Any docs updates needed?

No. CHANGELOG entry added.

### DIFF
--- a/internal/internal_worker_environment.go
+++ b/internal/internal_worker_environment.go
@@ -3,22 +3,71 @@ package internal
 import (
 	"os"
 	"runtime"
+	"runtime/debug"
 	"strings"
 
 	workerpb "go.temporal.io/api/worker/v1"
 )
 
+const (
+	// roadRunnerModulePrefix is the main module of every RoadRunner binary.
+	roadRunnerModulePrefix = "github.com/roadrunner-server/roadrunner"
+	// roadRunnerTemporalModulePrefix is the RoadRunner plugin through which the PHP SDK uses this SDK.
+	roadRunnerTemporalModulePrefix = "github.com/temporalio/roadrunner-temporal"
+)
+
 // detectEnvironmentInfo collects the runtime, hosting environment, and platform information
 // reported in the first accepted worker heartbeat.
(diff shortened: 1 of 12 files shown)


yuandrew (approved):
Not sure how much we care, but this would technically set PHP to report as Go, right?

Sushisource:
> Not sure how much we care, but this would technically set PHP to report as Go, right?

Well, it is on some level. We just need to _also_ report the roadrunner environment which isn't being done properly so I'll add that.

internal/internal_worker_environment.go
```diff
@@ -3,24 +3,78 @@ package internal
...
 	"runtime"
+	"runtime/debug"
 	"strings"
 
 	workerpb "go.temporal.io/api/worker/v1"
 )
 
+const (
+	// roadRunnerModulePrefix is the main module of every RoadRunner binary, including custom
+	// builds produced by Velox, which reuse RoadRunner's go.mod.
```
yuandrew: Do we need to mention Velox? seems unnecessarily specific

yuandrew (approved):

PR #2659, opened 2026-09-04
Add hints that function literals aren't supported

dplyukhin:
## What was changed

- Update `RegisterWorkflow` and `RegisterActivity` to hint that function literals (closures) aren't supported.
- In case of duplicate registration, use a regex see if the user is registering a function literal. If so, emit a hint.

## Why?

Document an assumption we've been making all along. Go 1.27 changed its naming convention for function literals, making it more likely for users to run into this issue. @stephanos gave an example from the OSS server:

> The test was failing on workflow name "func1" is already registered when using
> ```
> for _, wf := range wfs {
> 	w.RegisterWorkflow(wf)
> }
> ```

## Checklist

1. Closes  nothing

2. How was this tested: Added a test case to see if the error hint is only emitted for functions that match the regex.

3. Any docs updates needed? Updated docstrings for public functions.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"os"
 	"reflect"
+	"regexp"
 	"runtime"
 	"sort"
 	"strconv"
@@ -75,6 +76,8 @@ const (
 	unlimitedDeadlockDetectionTimeout = math.MaxInt64
 
 	testTagsContextKey = "temporal-testTags"
+
+	functionLiteralRegistrationHint = "It looks like you might have registered a function literal (closure), which is not supported."
 )
 
 type (
@@ -753,7 +756,11 @@ func (r *registry) RegisterWorkflowWithOptions(
 
 	if !options.DisableAlreadyRegisteredCheck {
 		if _, ok := r.workflowFuncMap[registerName]; ok {
-			panic(fmt.Sprintf("workflow name \"%v\" is already registered", registerName))
+			message := fmt.Sprintf("workflow name \"%v\" is already registered", registerName)
+			if mightBeFunctionLiteral(wf) {
+				message += ". " + functionLiteralRegistrationHint
+			}
+			panic(message)
 		}
 	}
 	r.workflowFuncMap[registerName] = wf
@@ -835,7 +842,11 @@ func (r *registry) RegisterActivityWithOptions(
 
 	if !options.DisableAlreadyRegisteredCheck {
 		if _, ok := r.activityFuncMap[registerName]; ok {
-			panic(fmt.Sprintf("activity type \"%v\" is already registered", registerName))
+			message := fmt.Sprintf("activity type \"%v\" is already registered", registerName)
+			if mightBeFunctionLiteral(af) {
+				message += ". " + functionLiteralRegistrationHint
+			}
+			panic(message)
 		}
 	}
 	r.activityFuncMap[registerName] = &activityExecutor{name: registerName, fn: af}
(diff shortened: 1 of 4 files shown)


Quinn-With-Two-Ns:
I am not sure we should say function literals are not supported. Heck I think lots of our tests in this repo do use function literals. The issue is just the type name is not deterministic

dplyukhin:
@Quinn-With-Two-Ns Good point! I changed the wording to suggest assigning an alias.

internal/internal_worker.go
```diff
@@ -753,7 +756,11 @@ func (r *registry) RegisterWorkflowWithOptions(
 
 	if !options.DisableAlreadyRegisteredCheck {
 		if _, ok := r.workflowFuncMap[registerName]; ok {
-			panic(fmt.Sprintf("workflow name \"%v\" is already registered", registerName))
+			message := fmt.Sprintf("workflow name \"%v\" is already registered", registerName)
+			if mightBeFunctionLiteral(wf) {
+				message += ". " + functionLiteralRegistrationHint + " Try RegisterWorkflowWithOptions instead."
```
jmaeagle99: I think this is shown when two closures are registered under the same alias, would would be misleading. Maybe only append the hint when `len(alias) == 0`.

internal/internal_worker.go
```diff
@@ -2746,6 +2757,18 @@ func isError(inType reflect.Type) bool {
 	return inType != nil && inType.Implements(errorElem)
 }
 
+// isFunctionLiteral matches strings that look like they could be the names
+// of function literals in Go 1.27+. Specifically, suffixes of the form funcN or
+// funcN.M for some numbers N and M.
+var functionLiteralNamePattern = regexp.MustCompile(`func\d+(?:\.\d+)?$`)
```
jmaeagle99: - False positive: Regex isn't anchored at some kind of boundary, so a function that ends with `func1` trips this for example.
- False negative: Not sure if this will happen, but deeply nested names pass through e.g. `func1.1.1`.

worker/worker.go
```diff
@@ -88,9 +91,10 @@ type (
 	// ActivityRegistry exposes activity registration functions to consumers.
 	ActivityRegistry interface {
 		// RegisterActivity - register an activity function or a pointer to a structure with the worker.
+		// Function literals (closures) are not supported.
```
jmaeagle99: Remove?

worker/worker.go
```diff
@@ -115,6 +119,7 @@ type (
 		// Serialization of all primitive types, structures is supported ... except channels, functions, variadic, unsafe pointer.
 		// This method panics if activityFunc doesn't comply with the expected format or an activity with the same
 		// type name is registered more than once.
+		// To register function literals (closures), use RegisterActivityWithOptions to assign a type name explicitly.
```
jmaeagle99:
```suggestion
		// To register function literals (closures), use [RegisterActivityWithOptions] to assign a type name explicitly.
```

internal/internal_worker.go
jmaeagle99: If we are going to be prescriptive with the guidance, maybe say something like "Register it with RegisterWorkflowWithOptions and set Name to a stable, unique name."

internal/internal_worker.go
```diff
@@ -835,7 +842,11 @@ func (r *registry) RegisterActivityWithOptions(
 
 	if !options.DisableAlreadyRegisteredCheck {
 		if _, ok := r.activityFuncMap[registerName]; ok {
-			panic(fmt.Sprintf("activity type \"%v\" is already registered", registerName))
+			message := fmt.Sprintf("activity type \"%v\" is already registered", registerName)
+			if mightBeFunctionLiteral(af) {
+				message += ". " + functionLiteralRegistrationHint + " Try RegisterActivityWithOptions instead."
```
jmaeagle99: Same as https://github.com/temporalio/sdk-go/pull/2659/changes#r3963337519

internal/internal_worker.go
dplyukhin: Done

internal/internal_worker.go
dplyukhin: The convention for function literals seems to be `funcN` or `funcN.M` with an arbitrary prefix - so you could see a literal with a name like `fizzBuzz.func12.func3.5`. But Go makes no promises that this convention will hold in future versions, so false positives and negatives are unavoidable here. I clarified the docs to account for that

internal/internal_worker.go
dplyukhin: Fixed with len(alias) == 0

jmaeagle99 (approved):

PR #2662, opened 2026-09-04
Replace big LA tunnel with linked list

### LINKED ISSUE #2660: Reduce local activity tunnel allocation
Local activity tunnel size is hard coded to be 100,000. This is taking up more memory than it likely needs. 

```
taskCh: make(chan *localActivityTask, 100000)
```

From this PR https://github.com/temporalio/sdk-go/pull/298. We should find a way to safely decrease this allocation. 

Probably good to reproduce the deadlock issue that old PR was trying to resolve, to ensure we don't regress here. But there should be a way to improve this.

Not clear if an unbuffered, bounded, or dynamically sized design is best.

**Additional context**
This affects OSS, since it internally spins up SDK workers.

Sushisource:
## What was changed
Change the memory-hogging 100,000 LA tunnel to use a synchronized linked list instead.

## Why?
This change maintains the same semantics but eliminates the huge memory overhead that the 100,000 item reserved buffer imposed, particularly in situations where there are many workers in the same process.

I considered doing more stuff to properly exert back-pressure but that would be a much more intrusive change. This is pretty slim.

## Checklist

1. Closes https://github.com/temporalio/sdk-go/issues/2660

2. How was this tested:
Added tests

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Touches core local-activity scheduling concurrency; behavior change is intentional but race/shutdown edge cases warrant review despite new tests.
> 
> **Overview**
> Replaces the **local activity tunnel**’s fixed **100,000-slot buffered channel** with a **mutex-backed FIFO linked list** on `localActivityTask` (`nextQueuedTask`), plus a small `taskReadyCh` to wake pollers.
> 
> **`sendTask` no longer blocks** when the queue is “full”—workflows can keep scheduling local activities while the same goroutine must still drain results—so it avoids the prior **deadlock** when the channel and all execution slots were saturated. Memory use scales with queued work instead of reserving a large buffer up front (notably when many workers share a process).
> 
> Shutdown clears the list via **`clearTasks`**; tests cover **>100k enqueues without a consumer**, **FIFO order**, and stop behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 739d3aaa7dba572999466c9801427bbf7390e524. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,9 @@ to docs, or any other relevant information.
 
 ### Fixed
 
+- Local activity scheduling no longer uses a fixed 100,000-entry task queue. The queue now grows
+  with demand, avoiding both the up-front allocation and a possible worker deadlock when the queue
+  and all local activity execution slots were full.
 - The `PayloadDownloadDuration` and `PayloadUploadDuration` fields on the workflow task duration log
   now report the wall-clock time external storage was in flight. Previously each batch's duration was
   summed, over-reporting the time whenever storage operations ran concurrently.
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -181,6 +181,8 @@ type (
 
 	localActivityTask struct {
 		sync.Mutex
+		// Owned by localActivityTunnel while the task is queued.
+		nextQueuedTask  *localActivityTask
 		workflowTask    *workflowTask
 		activityID      string
 		params          *ExecuteLocalActivityParams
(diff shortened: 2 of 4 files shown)


yuandrew:
bugbot run

yuandrew (approved):
makes sense to me! Yay linked lists 😄

Quinn-With-Two-Ns (approved):

PR #2674, opened 2026-09-09
[Issue 2673]: fix update-with-start unknown update panic

### LINKED ISSUE #2673: Update-with-start panics with `unknown update ... KnownUpdates=[]` when a workflow interceptor awaits a local activity before the workflow function runs
## Expected Behavior

When `temporalClient.UpdateWithStartWorkflow` starts a fresh workflow, an update sent in the same operation should be delivered to the workflow after it registers the matching update handler. The SDK's defaultUpdateHandler already queues updates that arrive before handlers exist. We expect the queued update to remain buffered until the workflow function registers its handler, then be dispatched.

## Actual Behavior

On the first (non-replay) workflow task of a fresh workflow, the SDK rejects the buffered update with:  `unknown update ... KnownUpdates=[]`.

This happens when a workflow inbound interceptor awaits a local activity in `ExecuteWorkflow` before calling `Next.ExecuteWorkflow` (where the user's workflow function registers its update handlers). For context, we have a workflow interceptor that handles authorization that executes a local activity before delegating to the workflow. The current Go SDK logic drains buffered updates as soon as all coroutines are blocked during the first WFT (assuming that by that time the update handlers are registered). However, in our setup, the local activity in the interceptor blocks the root coroutine during the first WFT before the workflow function has run and registered its handlers, so the drain rejects the update prematurely.

  1. Register a workflow inbound interceptor whose `ExecuteWorkflow` awaits a `workflow.ExecuteLocalActivity(...).Get(ctx)` before calling `i.Next.ExecuteWorkflow(ctx, in)`. 
  2. Register an update handler in the workflow
  3. Try `temporalClient.UpdateWithStartWorkflow`

## Specifications

  - Version: v1.38.0 
  - Platform: Linux (self-hosted Temporal server, Datadog fork)

eniko-dif:
## What was changed

Gated `DrainUnhandledUpdates` (the dispatcher's `allBlockedCallback`) on the user workflow function having started, so that buffered updates are only drained/rejected after the workflow has had a chance to register its update handlers.

Concretely, the dispatcher now tracks a `workflowFunctionStarted` flag that is set in two places:

- `workflowEnvironmentInterceptor.ExecuteWorkflow` — the innermost interceptor, right before it invokes the user workflow function via `executeFunction`. By this point all inbound interceptors have finished any pre-`Next.ExecuteWorkflow` work.
- `setUpdateHandler` — as a backup signal, so unit tests that exercise `defaultUpdateHandler`/`DrainUnhandledUpdates` directly (bypassing `ExecuteWorkflow`) still gate the drain correctly.

`ExecuteUntilAllBlocked` now invokes `allBlockedCallback` only when `workflowFunctionStarted` is true:

```go
for !allBlocked || (d.workflowFunctionStarted && d.allBlockedCallback()) {
```

A new `SetWorkflowFunctionStarted()` method was added to the `dispatcher` interface so the interceptor can set the flag without depending on the concrete `dispatcherImpl`.

## Why?

When a workflow inbound interceptor awaits a local activity in `ExecuteWorkflow` before calling `Next.ExecuteWorkflow` (a pattern Datadog uses for authorization checks), update-with-start fails on a fresh workflow with:

```
unknown update <name>. KnownUpdates=[]
```

Root cause: on the first (non-replay) workflow task of a fresh workflow, `TryUse(SDKPriorityUpdateHandling)` returns true, so `defaultUpdateHandler` queues the update rather than spawning `updateRunner`. The root coroutine then blocks on the interceptor's local activity before the user workflow function runs and registers its handlers. When all coroutines are blocked, `ExecuteUntilAllBlocked` invokes `allBlockedCallback` (`DrainUnhandledUpdates`), which unconditionally spawns the queued `updateRunner`. Because `priorityUpdateHandling` is true, `updateRunner` skips the "yield for initial handler registration" guard and rejects immediately — `eo.updateHandlers` is still empty.

The update is rejected by the SDK's own drain logic moments before the workflow function would have registered the handler, even though the local activity (e.g. a NoOp authz check) resolves instantly and the workflow function would have run in the same WFT.

This affects any setup where an inbound interceptor does blocking work before `Next.ExecuteWorkflow`, not just authorization. The previous behavior drained on every all-coroutines-blocked boundary (i.e. every WFT), which was too eager: it rejected updates that were legitimately still waiting for handler registration.

The narrower `workflowFunctionStarted` signal preserves the intent of `DrainUnhandledUpdates` (rejecting genuinely orphaned updates) while not rejecting updates the SDK just queued moments ago. Once the workflow function has started, any update not matching a registered handler is genuinely unknown and is still rejected promptly — including for long-running workflows that never return.

## Checklist

1. Closes [2673](https://github.com/temporalio/sdk-go/issues/2673)

2. How was this tested:
- Added `TestUpdateWithStartWorkflowWithInterceptor` — reproduces the bug using the existing `localActivityInterceptor` (which awaits a local activity before `Next.ExecuteWorkflow`) + `UpdateWithStartWorkflow` against `UpdateEntityWorkflow` (which registers an update handler as the first thing). Fails without the fix (`unknown update update. KnownUpdates=[]`), passes with it.
- Added `TestUpdateWithStartWorkflowWithInterceptorUnknownUpdate` — regression test for the too-broad alternative (gating on root coroutine completion). Sends an update-with-start with a name the workflow never registers against a long-running workflow (`UpdateEntityWorkflow` awaits `counter >= 1`). Must be rejected promptly with `unknown update`, not hang. Passes with `workflowFunctionStarted`; fails (server-side `UnprocessedUpdate` timeout) with a `rootClosed` gate.
- `go test ./internal/` (full unit suite) passes, including `TestDefaultUpdateHandler` which exercises `DrainUnhandledUpdates` directly (this required the `setUpdateHandler` backup signal).
- `go run . integration-test -dev-server -run "TestIntegrationSuite/TestUpdateWithStartWorkflowWithInterceptor"` passes.

3. Any docs updates needed?
- No public API changes. The `dispatcher` interface gained a method but it is internal. The behavior change is a bug fix and is covered by the existing guidance that update handlers should be registered early in the workflow function; this fix makes that guidance hold even when an interceptor blocks before the workflow function runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,14 @@ to docs, or any other relevant information.
 
 ### Fixed
 
+- Update-with-start no longer rejects the buffered update with
+  `unknown update ... KnownUpdates=[]` when a workflow inbound interceptor awaits a
+  local activity before `Next.ExecuteWorkflow`. `DrainUnhandledUpdates` is now gated
+  on the user workflow function having started (all inbound interceptors'
+  pre-`Next.ExecuteWorkflow` work complete), so buffered updates survive the
+  interceptor's initial local activity and are dispatched once the workflow registers
+  its handlers. Genuinely unknown updates are still rejected promptly once the workflow
+  function has started.
 - Local activity scheduling no longer uses a fixed 100,000-entry task queue. The queue now grows
   with demand, avoiding both the up-front allocation and a possible worker deadlock when the queue
   and all local activity execution slots were full.
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -93,6 +93,12 @@ type (
 		// NewCoroutine creates a new coroutine. To be called from within another coroutine.
 		// Used by the interceptors.
 		NewCoroutine(ctx Context, name string, highPriority bool, f func(ctx Context)) Context
+
+		// SetWorkflowFunctionStarted marks that the user workflow function has started
+		// running (all inbound interceptors have finished any pre-Next.ExecuteWorkflow
+		// work). It gates DrainUnhandledUpdates so buffered updates are only rejected
+		// after the workflow has had a chance to register handlers.
+		SetWorkflowFunctionStarted()
 	}
 
 	// Workflow is an interface that any workflow should implement.
@@ -176,6 +182,14 @@ type (
 		// returns true if the callback updated any coroutines state and there may be more work
 		allBlockedCallback func() bool
 		newEagerCoroutines []*coroutineState
+		// workflowFunctionStarted is true once the innermost interceptor has invoked
+		// the user workflow function (so all inbound interceptors have finished any
+		// pre-Next.ExecuteWorkflow work). This variable gates allBlockedCallback
+		// (DrainUnhandledUpdates) so that buffered updates are only drained/rejected
+		// after the workflow function has had a chance to register its handlers,
+		// not merely because all coroutines are temporarily blocked in an interceptor
+		// before the workflow function has run.
+		workflowFunctionStarted bool
 	}
 
 	// WorkflowOptions options passed to the workflow function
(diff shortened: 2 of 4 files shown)


Quinn-With-Two-Ns:
Hi , thank you for the contribution. Let's please discuss in the linked GitHub issue before working on a solution. It isn't clear this is something we can currently support, and if we do we will need to consider history backwards compatibility which would make the solution more complex

PR #2675, opened 2026-09-09
[AI-501] testsuite: run worker plugins in the test environments

### LINKED ISSUE #2480: testsuite: run worker plugins in the test environments
## Problem

Worker plugins (`worker.Options.Plugins`) do not apply in the test environments: `TestWorkflowEnvironment` / `TestActivityEnvironment` never construct a real worker, so no plugin hook (`ConfigureWorker`, `StartWorker`/`RunContextBefore`, `StopWorker`/`RunContextAfter`) ever runs. Anything a plugin contributes — registered workflows/activities, interceptors, converters — must be wired manually in tests.

This splits integration packaging in two: e.g. `contrib/googleadk` ships `NewPlugin` as its standard wiring (#2478), but tests must fall back to `NewActivities` + `Register`, and that manual path has to stay public (and documented) solely for testability. sdk-python does not have this gap — its test environment runs a real worker, so plugins compose into tests transparently.

## Proposal

Let the test environments accept plugins (e.g. honor `SetWorkerOptions(worker.Options{Plugins: ...})` or a dedicated setter) and invoke at minimum `ConfigureWorker` plus the `StartWorker`/`StopWorker` run-context hooks around test execution, with the environment's registry backing the plugin registry. That would let a plugin be the *single* wiring surface for an integration, in production and tests alike — and would unblock deprecating integration-specific manual registration APIs (see discussion on #2478).

DABH:
## What

`TestWorkflowEnvironment` and `TestActivityEnvironment` now run worker plugins set via `SetWorkerOptions(worker.Options{Plugins: ...})`, mirroring a real worker:

- `ConfigureWorker` runs at `SetWorkerOptions` (panics on error, once per environment).
- `StartWorker` runs before and `StopWorker` after each `ExecuteWorkflow`, `ExecuteActivity`, and `ExecuteLocalActivity`, also on panic. The environment is the registry handed to plugins.
- The environment's `Register*` methods fire plugin registry callbacks. Convenience paths (the workflow passed to `ExecuteWorkflow`, `OnActivity`/`OnWorkflow`/`OnNexusOperation` mocks) bypass them.

Drive-by fixes in `AggregatedWorker`: `RegisterDynamicWorkflow` now passes real options to its callback, and `RegisterDynamicActivity` no longer dereferences a nil `OnRegisterDynamicActivity`.

Documented limitations: the environment does not drain in-flight activity goroutines before `StopWorker`, and on `TestActivityEnvironment` each execution is a worker run under the same instance key.

`contrib/opentelemetry-v2`'s test environment now wires its tracing interceptor through `NewPlugin` and `worker.Options.Plugins`, the same path production uses, instead of building the interceptor by hand.

## Why

Closes #2480. Lets a plugin be the single wiring surface for an integration in production and tests. Follow-up after the next SDK release: switch `contrib/googleadk` tests to `NewPlugin` and deprecate `NewActivities`/`Register` (see #2478).

## Testing

Unit tests in `internal/workflow_testsuite_plugin_test.go` and `internal_workers_test.go`, a public-API test in `test/testsuite_plugin_test.go`, `check` and `unit-test` green. Verified end-to-end with googleadk's `NewPlugin` through a local `replace`.

---

> [!NOTE]
> **Medium Risk**
> Changes test-environment behavior (plugins no longer ignored; `SetWorkerOptions` is one-shot when plugins are set) and touches worker registration callbacks, which could affect plugin authors and existing tests that relied on the old semantics.
> 
> **Overview**
> **Test workflow and activity environments now honor `worker.Options.Plugins`**, so integrations can use the same plugin wiring in production and in tests. `ConfigureWorker` runs when `SetWorkerOptions` is called (errors panic; plugins can only be configured once per environment). Each `ExecuteWorkflow`, `ExecuteActivity`, and `ExecuteLocalActivity` brackets execution with `StartWorker` / `StopWorker`, with the environment acting as the plugin registry and firing `Register*` registry callbacks like a real worker (mocks and the workflow passed to `ExecuteWorkflow` still bypass those callbacks).
> 
> **Drive-by worker fixes:** `RegisterDynamicWorkflow` passes the caller’s options to `OnRegisterDynamicWorkflow`, and `RegisterDynamicActivity` invokes `OnRegisterDynamicActivity` instead of incorrectly using `OnRegisterActivity` (which could panic when only the activity callback was set).
> 
> **`contrib/opentelemetry-v2` tests** now attach tracing via `NewPlugin` and `worker.Options.Plugins` instead of manually constructing interceptors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82ad9759b1706ebe41d97aac309f5960fd2cfc20. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/internal_workflow_testsuite.go
```diff
@@ -570,6 +602,45 @@ func (env *testWorkflowEnvironmentImpl) setWorkerOptions(options WorkerOptions)
...
+func (env *testWorkflowEnvironmentImpl) startPluginWorker() error {
+	if len(env.plugins) == 0 {
+		return nil
+	}
+	start := func(context.Context, WorkerPluginStartWorkerOptions) error { return nil }
+	for i := len(env.plugins) - 1; i >= 0; i-- {
+		plugin := env.plugins[i]
+		next := start
+		start = func(ctx context.Context, options WorkerPluginStartWorkerOptions) error {
+			return plugin.StartWorker(ctx, options, next)
```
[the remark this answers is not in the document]
DABH: Good catch, fixed. Plugins now register through a small `testPluginRegistry` wrapper instead of the environment directly: it relaxes the duplicate check for workflows the same way the environment already does for activities, and leaves an already-registered dynamic workflow, dynamic activity, or Nexus service in place on later starts. Callbacks still fire with the plugin's original options. Added `TestActivityEnvPluginRepeatedRegistrations`, which registers all five kinds on every start and runs two executions; it panicked on the duplicate workflow before the change. The activity-env doc now says repeated registrations are tolerated, and keeps the note that a plugin whose stop is terminal still needs a fresh environment per call.

internal/workflow_testsuite.go
```diff
@@ -212,9 +212,19 @@ func (t *TestActivityEnvironment) ExecuteLocalActivity(activityFn any, args ...a
 	return t.impl.executeLocalActivity(activityFn, args...)
 }
 
-// SetWorkerOptions sets the WorkerOptions that will be use by TestActivityEnvironment. TestActivityEnvironment will
-// use options of BackgroundActivityContext, MaxConcurrentSessionExecutionSize, and WorkflowInterceptorChainFactories on the WorkerOptions.
-// Other options are ignored.
+// SetWorkerOptions sets the WorkerOptions used by TestActivityEnvironment. Only
+// BackgroundActivityContext, MaxConcurrentSessionExecutionSize,
+// EnableSessionWorker, DeadlockDetectionTimeout, PreferredVersionProvider,
+// Interceptors, and Plugins are used. Other options are ignored.
```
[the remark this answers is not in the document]
DABH: Agreed, both are read only from workflow code. Removed them from the `TestActivityEnvironment` doc; the list there is now `BackgroundActivityContext`, `MaxConcurrentSessionExecutionSize`, `EnableSessionWorker`, `Interceptors`, and `Plugins`. The workflow-environment doc keeps the full list.

CHANGELOG.md
```diff
@@ -23,6 +23,13 @@ to docs, or any other relevant information.
 
 ### Added
 
+- `TestWorkflowEnvironment` and `TestActivityEnvironment` now run worker plugins set through
```
yuandrew, line 26: Can we shorten this down to a sentence or 2

contrib/opentelemetry-v2/suite_test.go
```diff
@@ -47,16 +46,19 @@ func (s *otelTestSuite) newTestWorkflowEnvironment(
...
+	s.Require().NoError(err)
 
 	var testEnv testsuite.WorkflowTestSuite
 	if len(logger) > 0 {
 		testEnv.SetLogger(logger[0])
 	}
 	env := testEnv.NewTestWorkflowEnvironment()
-	env.SetWorkerOptions(worker.Options{Interceptors: []interceptor.WorkerInterceptor{workerInterceptor}})
+	// The test environment runs worker plugins, so the same plugin wiring
+	// production uses covers these tests.
```
yuandrew: Nit: this comment seems unnecessary

internal/internal_workflow_testsuite.go
```diff
@@ -556,6 +563,31 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(
...
+		if err := plugin.ConfigureWorker(context.Background(), WorkerPluginConfigureWorkerOptions{
+			WorkerInstanceKey:     env.workerInstanceKey,
+			TaskQueue:             env.workflowInfo.TaskQueueName,
+			WorkerOptions:         &options,
+			WorkerRegistryOptions: &pluginRegistryOptions,
+		}); err != nil {
+			panic(err)
+		}
+	}
+	// Only adopt the plugins once every ConfigureWorker succeeded.
```
yuandrew: Is there value in this comment? Seems like it adds little value on top of just reading the code

yuandrew:
Bugbot run

internal/plugin.go
```diff
@@ -116,7 +116,10 @@ type WorkerPlugin interface {
 	ConfigureWorker(context.Context, WorkerPluginConfigureWorkerOptions) error
 
 	// StartWorker is called to start a worker. This is called on Worker.Start
-	// or Worker.Run. Implementers should return an error or invoke next.
+	// or Worker.Run. The test environments in the testsuite package also call
```
yuandrew: What's the value of explicitly describing this here?

yuandrew (commented):
Overall looks good, can we do a round of auditing all comments and remove any that don't add much value to the code? Like if plugin registration works the same way in the testsuite as it does in non-test code, I don't think we need to add explicit comments around it is supported in the testsuite

CHANGELOG.md
DABH: Shortened to two sentences in 08666fa2.

contrib/opentelemetry-v2/suite_test.go
DABH: Removed in 08666fa2.

internal/internal_workflow_testsuite.go
DABH: Removed in 08666fa2, and dropped the remaining "As in NewAggregatedWorker" line in ba903122. The one comment left explains why a second call panics rather than overwriting the plugins.

internal/plugin.go
DABH: Agreed, removed in 08666fa2. `plugin.go` is no longer touched by this PR; ba903122 also reverted a leftover gofmt-only diff there.

DABH:
Done in 08666fa2 and ba903122. Pulled the testsuite mentions out of the plugin docs, the inline comments in `setWorkerOptions` and the otel suite, and a stray gofmt-only diff in `plugin.go`. I left a short paragraph on each `SetWorkerOptions` since that's where behavior actually differs from a real worker (start/stop per execution, callbacks skipping mocks), but happy to cut more.

internal/internal_workflow_testsuite.go
```diff
@@ -570,6 +593,95 @@ func (env *testWorkflowEnvironmentImpl) setWorkerOptions(options WorkerOptions)
...
+// registry, so repeated registrations from a plugin must not panic.
+type testPluginRegistry struct {
+	env *testWorkflowEnvironmentImpl
+}
+
+func (r testPluginRegistry) RegisterWorkflowWithOptions(w any, options RegisterWorkflowOptions) {
+	if r.env.pluginRegistryOptions.OnRegisterWorkflow != nil {
+		r.env.pluginRegistryOptions.OnRegisterWorkflow(w, options)
+	}
+	options.DisableAlreadyRegisteredCheck = true
```
yuandrew: This suppresses first-start registration conflicts, not only repeats from later activity executions. If user setup or another plugin registers the same name, the test passes while a real worker panics. 

I think we need to preserve existing-registration checks and deduplicate only registrations retained from prior runs.

internal/internal_workflow_testsuite.go
```diff
@@ -570,6 +593,95 @@ func (env *testWorkflowEnvironmentImpl) setWorkerOptions(options WorkerOptions)
...
+}
+
+func (r testPluginRegistry) RegisterDynamicActivity(a any, options DynamicRegisterActivityOptions) {
+	if r.env.pluginRegistryOptions.OnRegisterDynamicActivity != nil {
+		r.env.pluginRegistryOptions.OnRegisterDynamicActivity(a, options)
+	}
+	r.env.registry.Lock()
+	registered := r.env.registry.dynamicActivity != nil
+	r.env.registry.Unlock()
+	if !registered {
```
yuandrew: nit: this retains the first dynamic activity after `StopWorker`. If the next `StartWorker` creates a handler with new run-scoped state, execution still uses the stale handler. Maybe we should refresh plugin-owned dynamic registrations each run and cover this case

internal/internal_workflow_testsuite.go
DABH: Good catch. Rather than dedupe retained registrations, the environment now snapshots the registry around the StartWorker chain and removes what plugins added when the worker stops. Each run registers afresh with the normal duplicate checks intact, so a conflict with user setup or another plugin panics as on a real worker. Covered by new duplicate-registration tests for both environments.

internal/internal_workflow_testsuite.go
DABH: Fixed by the same change. Nothing plugin-owned survives StopWorker, so the next StartWorker installs new handlers. Added a test with run-scoped named and dynamic activities asserting each execution uses its own run's handler.

yuandrew (approved):

PR #2680, opened 2026-09-10
Use `ActivitySerializationContext` for standalone activities

GregoryTravis:
Use `ActivitySerializationContext` for activity payloads in `PollActivityResult`, `DescribeActivity` and `ExecuteActivity`.

`PollActivityResult` does not include `activity_type` and `task_queue` as they are not yet readily available.


internal/internal_activity_client.go
```diff
@@ -731,11 +737,18 @@ func (w *workflowClientInterceptor) PollActivityResult(
 		return nil, err
 	}
 
+	// TODO: add `activity_type` and `task_queue` to the `ActivitySerializationContext`
```
jmaeagle99: I don't think this is necessary. We've deprecated / obsoleted these in other languages:
- https://github.com/temporalio/sdk-python/blob/1.32.0/temporalio/converter/_serialization_context.py#L68
- https://github.com/temporalio/sdk-dotnet/blob/1.18.0/src/Temporalio/Converters/ISerializationContext.cs#L60

GregoryTravis: Removed

jmaeagle99 (commented):
- I think there should be some integration tests, similar to `TestSerializationContext_EndToEnd`
- I think we should update `ActivitySerializationContext`'s `WorkflowID` and `WorkflowType` fields to say they are empty for standalone activities.

GregoryTravis:
> I think there should be some integration tests, similar to TestSerializationContext_EndToEnd

I added a test that used a test code that only cares about namespace, since we're deferring figuring out how to plumb the additional fields through for now.

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ to docs, or any other relevant information.
   platforms), and the operating system and architecture. This is sent
   once per worker with the first heartbeat accepted by the server and can be turned off with
   `client.Options.DisableWorkerEnvironmentInfo`.
+- Standalone activity client calls now supply `converter.ActivitySerializationContext` when encoding
+  and decoding activity payloads: `ExecuteActivity` (input and static summary/details),
+  `ActivityHandle.Describe` (input, outcome, heartbeat details, last failure, and static
```
jmaeagle99: I don't think the type that is returned by `ActivityHandle.Describe` (which is `ClientActivityExecutionDescription`) actually provides an input or outcome notion.

jmaeagle99 (approved):

CHANGELOG.md
GregoryTravis: Right, got my PRs out of order in my head.

PR #2681, opened 2026-09-10
💥 Rename Standalone Activity Details to StaticDetails

GregoryTravis:
## What was changed
`client.StartActivityOptions.Details` was renamed to `StaticDetails`, and
`client.ActivityExecutionDescription.GetDetails` was renamed to `GetStaticDetails`.

## Why?
Brings the Standalone Activity user-metadata option in line with `client.StartWorkflowOptions`,
which already names the same field `StaticDetails`.

## Checklist

1. How was this tested:
Tests were updated to use the new names.


internal/internal_activity_client.go
```diff
@@ -353,9 +353,9 @@ func (d *ClientActivityExecutionDescription) GetSummary() (string, error) {
...
 }
 
-// GetDetails returns details of the activity. See ClientStartActivityOptions.Details. Returns empty string if there are no details.
+// GetStaticDetails returns details of the activity. See ClientStartActivityOptions.StaticDetails. Returns empty string if there are no details.
 // Uses the data converter of the client used to make the Describe call. Returns error if data conversion fails.
-func (d *ClientActivityExecutionDescription) GetDetails() (string, error) {
+func (d *ClientActivityExecutionDescription) GetStaticDetails() (string, error) {
 	if d.details != "" {
 		return d.details, nil
 	}
```
maciejdudko, line 361: Nit: we should rename the private field too.

maciejdudko (approved):
Approved with 1 nit. Can be merged as is.

PR #2686, opened 2026-09-13
fix(opentelemetry): keep continue-as-new spans from failing

### LINKED ISSUE #2676: ContinueAsNewError spans marked status:error in contrib/opentelemetry-v2 interceptor
## Expected Behavior
A workflow that completes via workflow.NewContinueAsNewError(...) should not have its OpenTelemetry span marked status:error. Continue-as-new is normal control flow used to avoid unbounded history growth in long-lived workflows, not a failure.

## Actual Behavior
Finish in contrib/opentelemetry-v2/spans.go only special-cases benign ApplicationErrors before calling SetStatus:

```
if opts.Error != nil {
    t.RecordError(opts.Error)
    appError, _ := opts.Error.(*temporal.ApplicationError)
    isBenign := appError != nil && appError.Category() == temporal.ApplicationErrorCategoryBenign
    if !isBenign {
        t.SetStatus(codes.Error, opts.Error.Error())
    }
}
```

ContinueAsNewError is not an ApplicationError, so it always falls through to SetStatus(codes.Error, ...). Every continuation of a self-continuing workflow is recorded as an errored span. In our case this produced ~385k error-status spans over 7 days in APM error tracking for a single workflow, all @error.type: *internal.ContinueAsNewError, none of which were actual failures.

1. Register a workflow that continues itself, e.g.:
```
func LongLivedWorkflow(ctx workflow.Context, iteration int) error {
    if iteration >= maxIterationsBeforeContinue {
        return workflow.NewContinueAsNewError(ctx, LongLivedWorkflow, iteration+1)
    }
    return nil
}
```
2. Wire up the worker with the OTel tracing interceptor:
```
tracer, _ := opentelemetry.NewTracingInterceptor(opentelemetry.TracerOptions{})
worker.Options{Interceptors: []interceptor.WorkerInterceptor{tracer}}
```
3. Run the workflow to completion so it continues-as-new, and inspect the exported RunWorkflow span (e.g. via an in-memory OTel exporter or your tracing backend) — it shows codes.Error status with the ContinueAsNewError recorded, despite the workflow succeeding.

## Specifications

- Version: go.temporal.io/sdk/contrib/opentelemetry-v2 v0.1.0 (commit a08198f, matches main as of this report)
- Platform:  
  - Reproduced locally on macOS (Darwin arm64, go1.26.4)
  - Production instances run in Linux containers (Alpine base, linux/amd64) on AWS
  - Not platform-dependent — the bug is purely in client-side span-status logic and reproduces identically in both environments

jstar0:
## Problem

A workflow that returns `ContinueAsNewError` is exercising normal control flow, but `tracerSpan.Finish` in `contrib/opentelemetry-v2` recorded it as an exception and set the span status to `codes.Error` — `ContinueAsNewError` is not an `ApplicationError`, so only benign application errors were excluded. Every continuation of a self-continuing workflow therefore shows up as an errored span; one production deployment reported ~385k error-status spans over 7 days for a single workflow, all with `@error.type: *internal.ContinueAsNewError` and none of them real failures.

Fixes #2676.

## Fix

`tracerSpan.Finish` now treats `ContinueAsNewError` like benign application errors: it skips both the recorded exception event and the error status, and simply ends the span. Continue-as-new is normal control flow, not a failure.

## Verification

- `go test .` in `contrib/opentelemetry-v2`: full suite passes, including a new `TestSpanErrorStatus/continue-as-new` case that runs a workflow returning `ContinueAsNewError` and asserts the `RunWorkflow` span ends with `codes.Unset` and no recorded exception events (previously `codes.Error` with an exception event).
- `gofmt` clean; `go vet ./...` clean in the module.


brendandesj-koho (approved):

contrib/opentelemetry-v2/spans.go
```diff
@@ -24,13 +26,17 @@ func (t *tracerSpan) Finish(opts *tracing.TracerFinishSpanOptions) {
 	}
 
 	if opts.Error != nil {
```
yuandrew, line 28: We actually have `workflow.IsContinueAsNewError` that we can use to make this a lot cleaner, where we can simply change this line to `if opts.Error != nil && !workflow.IsContinueAsNewError(opts.Error) {`

jstar0: Thanks — I switched the check to `workflow.IsContinueAsNewError` and removed the local `errors.As` logic. `TestIntegrationTestSuite/TestComprehensive` checks the Continue-as-new span remains `codes.Unset` and has no error events; it passes via `go run . unit-test -run TestIntegrationTestSuite/TestComprehensive`, and `go run . check` passes.

PR #2697, opened 2026-09-16
feat: workflow task duration configurable warning threshold

jmaeagle99:
## What was changed

The `[TMPRL1104]` workflow task duration log now uses a single warning threshold, read once from `TEMPORAL_WORKFLOW_TASK_DURATION_WARN_SECONDS` and defaulting to 5 seconds.

## Why?

- Simplify workflow task duration so its only reported when it may be problematic e.g. too long.
- Give customers ability to configure the warning threshold.
- Align with behavior of core-based SDKs.

## Checklist

1. How was this tested: New unit tests


internal/internal_task_pollers.go
```diff
@@ -1090,6 +1088,23 @@ func (wtp *workflowTaskProcessor) errorToFailWorkflowTaskWithCause(taskToken []b
 	return builtRequest
 }
 
+const defaultWFTDurationWarnThreshold = 5 * time.Second
+
+var wftDurationWarnThreshold = sync.OnceValue(func() time.Duration {
+	return parseWFTDurationWarnThreshold(os.Getenv("TEMPORAL_WORKFLOW_TASK_DURATION_WARN_SECONDS"))
```
Quinn-With-Two-Ns, line 1094: So will all SDKs use env variable

jmaeagle99: Yes. sdk-core crate already has this same env var.

PR #2700, opened 2026-09-17
Fix command collection race and integration log spam

yuandrew:
## What was changed
- Synchronized addCommand and getCommands.
- Removed the suite-wide 128-byte payload warning threshold.

TestIntegrationSuite/TestVersionLoopWorkflow exceeded the deadlock timeout while emitting payload warnings. Under worker.FailWorkflow, its coroutine continued adding a version-marker command while the worker collected commands for the failure response.

This fixes that specific race. It does not make commandsHelper fully thread-safe; other deadlock-path interactions require a broader design.

## Why?

[The failing CI job](https://github.com/temporalio/sdk-go/actions/runs/35174876345/job/105054238395?pr=2699) emitted 1,075 TMPRL1103 warnings. The suite-wide threshold caused unrelated tests to exercise payload-warning behavior already covered by dedicated tests

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path workflow command bookkeeping under panic/deadlock handling; locking is limited to add/collect, so other concurrent helper access could still race.
> 
> **Overview**
> Fixes a **race** between workflow command creation and command collection when a deadlocked workflow is failed via **`worker.FailWorkflow`**. The workflow coroutine could still finish an in-flight SDK call (for example adding a **GetVersion** marker) while the worker concurrently walked the command list in **`getCommands`**, which could corrupt internal state or destabilize CI.
> 
> **`commandsHelper`** now uses a **mutex** around **`addCommand`** and **`getCommands`** so those two paths cannot run at the same time. A new stress test exercises concurrent adds and collects. This is intentionally **narrow**—other concurrent uses of the helper are not fully serialized.
> 
> Integration setup also **drops** the suite-wide **128-byte `PayloadSizeWarning`** client option so unrelated integration tests no longer flood logs with **TMPRL1103** warnings and slow runs like **`TestVersionLoopWorkflow`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5f8e2e8936465fcc5fd5f1d1f026f2e0d45cf47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


test/integration_test.go
```diff
@@ -185,9 +185,6 @@ func (ts *IntegrationTestSuite) SetupTest() {
 		options.MetricsHandler = metricsHandler
 		options.TrafficController = trafficController
 		options.Interceptors = clientInterceptors
-		options.PayloadLimits = client.PayloadLimitOptions{
```
jmaeagle99: I would suggest making this change in a separate PR so that its commit is not coupled with the other changes just in case the other changes need reversion.

